PR #1771, opened 2022-11-28
ToricVarieties: Bugfix in is_empty for closed subvarieties

HereAround:
A closed subvariety of a toric variety, defined by an ideal `I`, is empty iff B^l subseteq I for a suitable non-negative integer l.

After discussion with @tthsqe12 , I have modified the code so that (at least to the best of my knowledge) it truly checks this criterion.

### DIFF
--- a/src/ToricVarieties/Subvarieties/properties.jl
+++ b/src/ToricVarieties/Subvarieties/properties.jl
@@ -29,9 +29,6 @@ true
 """
 @attr Bool function is_empty(c::ClosedSubvarietyOfToricVariety)
     B = irrelevant_ideal(toric_variety(c))
-    if (is_one(defining_ideal(c)) || is_one(saturation(defining_ideal(c), B)))
-      return true
-    end
-    return false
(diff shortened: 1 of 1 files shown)


src/ToricVarieties/Subvarieties/properties.jl
```diff
@@ -19,10 +19,19 @@ A closed subvariety of a normal toric variety
...
+true
 ```
 """
 @attr Bool function is_empty(c::ClosedSubvarietyOfToricVariety)
     B = irrelevant_ideal(toric_variety(c))
-    return intersect(radical(c), B) == B
+    if (is_one(defining_ideal(c)) || is_one(saturation(defining_ideal(c), B)))
+      return true
+    end
+    return false
```
lkastner: Why not
```suggestion
    return (is_one(defining_ideal(c)) || is_one(saturation(defining_ideal(c), B)))
```

lkastner (approved):

HereAround:
Thank you @lkastner .

PR #1775, opened 2022-11-29
Bump to 0.11.1

thofma:
Since OSCAR 0.11, there has been a version bump in Hecke from 0.15 to 0.16. So I would like to release OSCAR 0.11.1 to make it not lag behind.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Oscar"
 uuid = "f1435218-dba5-11e9-1e4d-f1a5fab5fc13"
 authors = ["The OSCAR Team <oscar@mathematik.uni-kl.de>"]
-version = "0.11.1-DEV"
+version = "0.11.1"
 
 [deps]
 AbstractAlgebra = "c3fe647b-3220-5bb0-a1ea-a7954cac585d"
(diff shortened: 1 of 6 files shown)


fingolfin:
also good for GAP

fingolfin (approved):

fingolfin (changes requested):
Please use the `etc/update_version.sh` script to ensure the version is updated in all places it appears.

thofma:
ah uups, thanks for the reminder

thofma:
done

fingolfin (approved):

thofma:
Hm, the doctests errors look rather strange, happening because the `save` command returns something different. Are these real errors @benlorenz @antonydellavecchia?

I am inclined to suppress the output via `;` in the doctest.

benlorenz:
I think the number indicates the number of bytes written, and that includes the full version information in the file, and the change `"prerelease": ["DEV"]` to `"prerelease": []` does fit to the five bytes difference.
I would suggest that we drop that return value directly in the save function.

thofma:
Yes, that makes sense. For now I have decorated those `save` calls with `;` to unblock this PR.

PR #1777, opened 2022-11-29
save: don't return number of bytes written

benlorenz:
@antonydellavecchia
I don't think anyone needs this and it mostly causes confusion (or CI issues, see #1775).
It is also not documented and probably not really intended but just an automatic forwarding of the return of `write`.

cc @thofma @lkastner 

PS: Technically I guess this is a breaking change?

### DIFF
--- a/docs/src/PolyhedralGeometry/Polyhedra/serialization.md
+++ b/docs/src/PolyhedralGeometry/Polyhedra/serialization.md
@@ -22,7 +22,7 @@ following way:
 julia> square = cube(2)
 A polyhedron in ambient dimension 2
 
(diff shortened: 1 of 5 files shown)


fingolfin (approved):

thofma:
I can live with `save` returning the number of bytes, but it should not appear in the doctests, which can be achieved also using `;` (but might be harder to spot in future PRs). I don't know if it is a universal convention, but from my limited experience, most `write` implementations in other languages also return the number of bytes. Not sure this is a good reason to keep it though. I never knew what to do with the those numbers.

lkastner (approved):

PR #1780, opened 2022-11-30
When computing matrix group orders, inform GAP about the result

### LINKED ISSUE #1779: `order(G)` for matrix group should set the order in GAP, set `is_finite` etc.
```
julia> K = QQ
Rational Field

julia> M1 = matrix(K, [0 0 1; 1 0 0; 0 1 0])
[0   0   1]
[1   0   0]
[0   1   0]

julia> G = matrix_group(3, K, [ M1 ])
Matrix group of degree 3 over Rational Field

julia> order(G)
3

julia> describe(G)
"a group"
```
This is because only Oscar knows the order at this point, not GAP. Result:

```
julia> has_order(G)
false

julia> has_is_finite(G)
false

julia> is_finite(G)
true

julia> describe(G)
"C3"

```

We should change this and if we managed to compute the order on the OSCAR side, transfer it back to the GAP side.

fingolfin:
Resolves part of #1779, but more should be done

### DIFF
--- a/src/Groups/matrices/MatGrp.jl
+++ b/src/Groups/matrices/MatGrp.jl
@@ -545,7 +545,7 @@ gen(G::MatrixGroup, i::Int) = gens(G)[i]
 ngens(G::MatrixGroup) = length(gens(G))
 
 
-compute_order(G::GAPGroup) = fmpz(GAPWrap.Order(G.X))
+compute_order(G::GAPGroup) = fmpz(GAPWrap.Size(G.X))
 
 function compute_order(G::MatrixGroup{T}) where {T <: Union{nf_elem, fmpq}}
   #=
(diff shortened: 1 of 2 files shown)


src/Groups/matrices/MatGrp.jl
```diff
@@ -561,9 +561,11 @@ function compute_order(G::MatrixGroup{T}) where {T <: Union{nf_elem, fmpq}}
     # The call to `IsHandledByNiceMonomorphism` triggers an expensive
     # computation of `IsFinite` which we avoid by checking
     # `HasIsHandledByNiceMonomorphism` first.
-    return fmpz(GAPWrap.Order(G.X))
+    return fmpz(GAPWrap.Size(G.X))
   else
-    order(isomorphic_group_over_finite_field(G)[1])
+    n = order(isomorphic_group_over_finite_field(G)[1])
+    GAP.Globals.SetSize(G.X, GAP.Obj(n))
```
ThomasBreuer, line 567: I would suggest `set_order(G, n)`.
However, currently `set_order` is wrong; it should call `GAP.Globals.SetSize` with second argument `GAP.Obj(val)` not with `GapObj(val)`.

ThomasBreuer (approved):
Good idea. Either merge the changes as they are or fix also the `set_order` bug.

src/Groups/matrices/MatGrp.jl
fingolfin: Yeah, and I found a few more things. I will now also set up a NiceMonomorphism so that GAP can do even more things

fingolfin: also, if the group is infinite then `isomorphic_group_over_finite_field` also determines this, I'll make sure we also use that information

thofma: In this PR here or in the next one?

fingolfin: The next one, this one is not great but at least fixes one immediate problem. I'll close and reopen to have the tests re-run, then we can merge

PR #1783, opened 2022-12-01
PolyhedralGeometry: Allow solve_mixed to return an iterator

lkastner:


thofma:
Love it! A test would be great.

lkastner:
> Love it! A test would be great.

Unfortunately I cannot test the exact result, as the order may change. I could do some `sort` as it is done in the doctests or is it enough as is? I do compare the `nrows` or `length` though to the correct result.

src/PolyhedralGeometry/solving_integrally.jl
```diff
@@ -1,11 +1,33 @@
...
 
+function solve_mixed(as::Type{SubObjectIterator{PointVector{fmpz}}}, A::fmpz_mat, b::fmpz_mat, C::fmpz_mat, d::fmpz_mat)
+    ncols(A) == ncols(C) || throw(ArgumentError("solve_mixed(A,b,C,d): A and C must have the same number of columns."))
+    nrows(A) == nrows(b) || throw(ArgumentError("solve_mixed(A,b,C,d): A and b must have the same number of rows."))
+    nrows(C) == nrows(d) || throw(ArgumentError("solve_mixed(A,b,C,d): C and d must have the same number of rows."))
+    ncols(b) == 1 || throw(ArgumentError("solve_mixed(A,b,C,d): b must be a matrix with a single column."))
+    ncols(d) == 1 || throw(ArgumentError("solve_mixed(A,b,C,d): d must be a matrix with a single column."))
+    eq = (Matrix{BigInt}(A), vec(Matrix{BigInt}(b)))
+    ineq = (Matrix{BigInt}(-C), vec(Matrix{BigInt}(-d)))
+    P = Polyhedron(ineq, eq)
```
benlorenz:
```suggestion
    P = Polyhedron((-C, vec(-d)), (A, vec(b)))
```
This should avoid a few unnecessary conversion.

benlorenz (approved):

src/PolyhedralGeometry/solving_integrally.jl
lkastner: Thanks.

PR #1784, opened 2022-12-02
Skip nondeterministic tests

thofma:


thofma:
Like this @benlorenz?

benlorenz:
Yes, I think that should work:
> This can be useful for tests that intermittently fail, or tests of not-yet-implemented functionality.

benlorenz (approved):

tthsqe12:
Ah yes, good. I didn't know about test_skiped. All of these tests need to be rewritten.

PR #1785, opened 2022-12-02
Cb/document matrix files

CameronBraunstein:


src/Groups/matrices/MatGrp.jl
```diff
@@ -588,10 +588,27 @@ end
 
 """
     general_linear_group(n::Int, q::Int)
-    general_linear_group(n::Int, F::FqNmodFiniteField)
+    general_linear_group(n::Int, F::Ring)
```
fingolfin: If `F` is a ring now, perhaps rename it to `R` and also adjust the text in line 594 which specifically talks about `F` being a field?

Same applies elsewhere

src/Groups/matrices/MatGrp.jl
```diff
@@ -588,10 +588,27 @@ end
...
     GL = general_linear_group
 
 Return the general linear group of dimension `n` either over the field `F` or the field `GF(q)`.
+
+Currently, this function only supports fields of type FqNmodFiniteField, however it will eventually support any ring.
```
fingolfin: "Any" ring, I am not sure we want to make that promise... I mean, sure, in principle we could create a group object that calls itself `GL(n,R)` for any ring, but its functionality would be extremely limited.
```suggestion
Currently, this function only supports fields of type `FqNmodFiniteField`.
```
Same applies elsewhere

src/Groups/matrices/MatGrp.jl
```diff
@@ -605,10 +622,27 @@ end
 
 """
     special_linear_group(n::Int, q::Int)
-    special_linear_group(n::Int, F::FqNmodFiniteField)
+    special_linear_group(n::Int, F::Ring)
     SL = special_linear_group
 
 Return the special linear group of dimension `n` either over the field `F` or the field `GF(q)`.
+
+Currently, this function only supports fields of type FqNmodFiniteField, however it will eventually support any ring.
```
fingolfin: Same as above, etc. etc

src/Groups/matrices/MatGrp.jl
```diff
@@ -605,10 +622,27 @@ end
...
+
+#Examples
```
fingolfin:
```suggestion
# Examples
```
Same applies elsewhere

src/Groups/matrices/matrix_manipulation.jl
```diff
@@ -138,6 +139,18 @@ end
 
 Return the permutation matrix over the ring `R` corresponding to the sequence `Q` or to the permutation `p`.
 If `Q` is a sequence, then `Q` must contain exactly once every integer from 1 to some `n`.
+
+#Examples
```
fingolfin:
```suggestion
# Examples
```

fingolfin:
@CameronBraunstein ping?

src/Groups/matrices/matrix_manipulation.jl
fingolfin: @CameronBraunstein Unfortunately you did *not* fix this thoroughly in all applicable places *sigh*. I have now done this myself.

fingolfin:
The code was not even runnable (variable `F` was renamed to `R` but not all occurrences in the code were changed). I've fixed this and a bunch of other issues. Assuming tests pass now, I'll merge it.

fingolfin (approved):

PR #1786, opened 2022-12-02
Version 0.11.2

thofma:


thofma:
Releases 0.11.1 and 0.11.0 have, because of the Singular_jll update, broken tests, rendering these releases useless for the downstream tests. Instead of doing the "yank" dance in the registry and another Yggdrasil round, we just release a new Oscar.

benlorenz (approved):

tthsqe12:
Which tests, specifically? I have had many problems with the bad tests in tropical geometry. But, all of the other tests passed when I checked them with the new Singular_jll

tthsqe12:
Nevermind, I see #1784

PR #1787, opened 2022-12-02
Minor adjustment of ideals

### LINKED ISSUE #1778: Addition of empty ideals fails
I had some code where I was adding two ideals `I` and `J` over a polynomial ring `R` which both had an empty set of generators. Then [this](https://github.com/oscar-system/Oscar.jl/blob/master/src/Rings/mpoly.jl#L220) line caused trouble, because there was no element set. 

I tried to produce a minimal example, but failed. The point is that I did not find the constructor which would have an ideal with an empty set of generators as an output. But since this occured in my code, I know that there is a hole somewhere where such things can sneak in.  Either way, I personally think it would make sense to allow ideal with an empty set of generators. 

If necessary, I can upload a branch with the failing example. However, that will then be of more complicated nature. 

Edit: I just saw that the constructor above was called from [this](https://github.com/oscar-system/Oscar.jl/blob/master/src/Rings/mpoly-ideals.jl#L146) line. Most probably you already have the reason for the problem there.

thofma:
- Make `ideals(R, elements)` the primary method
- Do not modify the generators

Should fix #1778.

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -48,10 +48,12 @@ julia> typeof(J)
 MPolyIdeal{MPolyElem_dec{fmpq, fmpq_mpoly}}
 ```
 """
-function ideal(g::Vector{T}) where {T <: MPolyElem}
-  @assert length(g) > 0
-  @assert all(x->parent(x) == parent(g[1]), g)
-  return MPolyIdeal(g)
+function ideal(R::MPolyRing, g::Vector)
+  h = elem_type(R)[R(f) for f = g]
+  #if isempty(g)
+  #  push!(h, R())
+  #end
+  return MPolyIdeal(R, h)
 end
 
 function ideal(I::IdealGens{T}) where {T <: MPolyElem}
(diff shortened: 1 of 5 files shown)


ederc (approved):

src/Rings/mpoly-ideals.jl
```diff
@@ -48,10 +48,12 @@ julia> typeof(J)
 MPolyIdeal{MPolyElem_dec{fmpq, fmpq_mpoly}}
 ```
 """
-function ideal(g::Vector{T}) where {T <: MPolyElem}
-  @assert length(g) > 0
-  @assert all(x->parent(x) == parent(g[1]), g)
-  return MPolyIdeal(g)
+function ideal(R::MPolyRing, g::Vector)
+  h = elem_type(R)[R(f) for f = g]
```
HechtiDerLachs, line 52: @simonbrandhorst pointed out the other day that it might be desirable to *only* accept elements as input which are already elements of `R` and throw an error otherwise. I disagreed, but couldn't point out another instance where implicit coercion was done. Now there's one here which is not due to me! 

Question: Do we have it written down somewhere which convention we want to follow for this? Just so we do not stumble upon this discussion again and again?

thofma: Yes, we do this in lots of different constructs. For example:
```
julia> matrix(QQ, 2, 2, [1, 2, 3, 4])
[1   2]
[3   4]
```
Note that this is mainly convenience for interactive use.

I think this is orthogonal to this PR here.

PR #1795, opened 2022-12-05
Some fixes

HechtiDerLachs:
These are some fixes I came up with today while working on other problems. 

@afkafkafk13 : Can you have a quick look at the changes for the space germ functionality?

### DIFF
--- a/experimental/Schemes/SpaceGerms.jl
+++ b/experimental/Schemes/SpaceGerms.jl
@@ -275,6 +275,25 @@ function issubset(X::AbsSpaceGerm{<:Any, <:MPolyQuoLocalizedRing}, Y::AbsSpaceGe
   return issubset(IY,modulus(OO(X)))
 end
 
+function issubset(X::AbsSpaceGerm{<:Any, <:MPolyLocalizedRing}, Y::AbsSpaceGerm{<:Any, <:MPolyQuoLocalizedRing})
+  R = ambient_coordinate_ring(X)
+  R === ambient_coordinate_ring(Y) || return false
+  point(X) == point(Y) || return false
+  return iszero(modulus(OO(Y)))
+end
+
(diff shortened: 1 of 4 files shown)


src/Rings/mpolyquo-localizations.jl
```diff
@@ -177,7 +177,7 @@ function modulus(L::MPolyQuoLocalizedRing)
 end
 
 ### for compatibility -- also provide modulus in the trivial case
-modulus(R::MPAnyNonQuoRing)=ideal(R,[zero(R)])
+modulus(R::MPAnyNonQuoRing)=ideal(R, elem_type(R)[])
```
afkafkafk13, line 180: Since empty lists work now, this is fine with me.

experimental/Schemes/SpaceGerms.jl
```diff
@@ -303,3 +322,14 @@ function singular_locus(X::AbsSpaceGerm)
   Sgerm = SpaceGerm(S)
   return Sgerm, ClosedEmbedding(SpecMor(Sgerm, X, pullback(inc), check=false), image_ideal(inc), check=false)
 end
+
+function point(X::SpaceGerm{<:Field, <:AbsLocalizedRing{<:Ring, <:RingElem, <:MPolyComplementOfKPointIdeal}})
```
afkafkafk13: Can we move this instance of point up to the two other signatures for point (around line 70 of the same file)?

afkafkafk13 (commented):
Looks good to me. 
Just one thing: Please make sure that the signatures, you now added, end up close to the respective existing methods in the file.

experimental/Schemes/SpaceGerms.jl
HechtiDerLachs: I just saw that the method was already there. I just didn't find it earlier.

afkafkafk13 (approved):

simonbrandhorst:
Could you please add a few tests?

PR #1797, opened 2022-12-06
Clean up printing

HechtiDerLachs:
In this morning's merge I had made a mess with printing of elements of localized rings. This should make the warning go away when starting up Oscar.

@tthsqe12 : Thanks for explaining the expressify stuff (again)!

### DIFF
--- a/src/Rings/localization_interface.jl
+++ b/src/Rings/localization_interface.jl
@@ -277,7 +277,10 @@ function parent(f::AbsLocalizedRingElem)
   error("`parent` is not implemented for the type $(typeof(f))")
 end
 
-expressify(f::AbsLocalizedRingElem; context=nothing) = Expr(:call, ://, expressify(numerator(f), context=context), expressify(denominator(f), context=context))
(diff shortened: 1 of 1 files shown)


afkafkafk13:
Part of this was already merged in #1795 (all but 179d3db). Is this duplicate on purpose?

HechtiDerLachs:
> Is this duplicate on purpose?

No. I was assuming that GH would find out by itself. I just pushed to the same branch. But it's true: This is really only about the very last commit.

afkafkafk13 (approved):
only for the last commit (all others were already merged): 
Change makes sense.

fingolfin:
After discussion with @afkafkafk13 I've used `git rebase --onto=master HEAD^` to drop all but the final commit in this PR, and force-pushed it here. @HechtiDerLachs please check this matches what you want / expected ;-).

HechtiDerLachs:
> please check this matches what you want / expected

Yes, it does. Thank you very much!

PR #1798, opened 2022-12-06
Use sparse matrices in the pre_saturation_data.

HechtiDerLachs:
It turned out that for huge generating sets of ideals, the internal data to translate from one side to the other was eating up memory very badly: I had been using dense matrices for that.

Now I switched to sparse matrices. Unfortunately, due to the restricted functionality for the latter, I had to modify the code to an almost unreadable form. 

If anyone has better ideas on how to solve this, let me know! Otherwise, this should do the job for the time being.

### DIFF
--- a/src/Modules/Posur.jl
+++ b/src/Modules/Posur.jl
@@ -19,11 +19,13 @@ function clear_denominators(A::MatrixType) where {T<:AbsLocalizedRingElem, Matri
   n = ncols(A)
   S = base_ring(A)
   R = base_ring(S)
-  D = zero(MatrixSpace(R, m, m))
+  D = sparse_matrix(R)
+  #D = zero(MatrixSpace(R, m, m))
   B = zero(MatrixSpace(R, m, n))
   for i in 1:m
     d = lcm(vec(denominator.(A[i,:])))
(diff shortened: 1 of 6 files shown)


thofma:
If you could be more specific about the missing functionality, we could do something about it (maybe).

HechtiDerLachs:
> If you could be more specific about the missing functionality, we could do something about it (maybe).

It's not really missing. @fieker explained to me that multiplication `sparse * dense` is only implemented that way round, since otherwise one can not benefit from the storage pattern. But that is against the Oscar covention that I had followed. Hence, I had to store the transpose of the matrix and replace `v*A` by `transpose(mul(A, transpose(v)))` everywhere. 

Do you think there is a solution to this? 

One could probably introduce an internal routine like `_mult_with_transpose(v::MatrixElem, A::SMat)`. But do we really want that?

thofma:
I see. We could store the transpose of `A` internally in `A` itself and use it for `v * A`. I will discuss it with @fieker once I see him again.

simonbrandhorst (approved):
This seems uncontroversial and tests pass. So if you say that it helps, that is good enough for me.

PR #1804, opened 2022-12-08
Fix all warnings again

thofma:


HechtiDerLachs:
Thanks! This seems to be just the right thing -- at least about my parts of the code.

HechtiDerLachs (approved):

PR #1805, opened 2022-12-08
Mpolyquo modules

HechtiDerLachs:
We provide a first implementation of modules over MPolyQuos.


HechtiDerLachs:
Sorry, the "temporary fix" was not supposed to be there. I will remove it, eventually, but let's first see about the tests.

src/Modules/ModuleTypes.jl
```diff
@@ -49,12 +49,15 @@ abstract type AbstractSubQuoElem{T} <: ModuleFPElem{T} end
...
 
 @doc Markdown.doc"""
-    ModuleFPHom{T1, T2}
+    ModuleFPHom{T1, T2, RingMapType}
 
 The abstract supertype for morphisms of finitely presented modules over multivariate polynomial rings .
 `T1` and `T2` are the types of domain and codomain respectively.
+`RingMapType` is a type for a homomorphism of rings ``f : R → S`` whenever the 
+`base_ring` ``R`` of the domain is different from the `base_ring` ``S`` of the codomain 
+and the codomain is considered as an ``R``-module via ``f``.
```
thofma: What is it in the other case?

test/Modules/MPolyQuo.jl
```diff
@@ -22,3 +22,50 @@
...
+F = FreeMod(R, 1)
+FQ = FreeMod(Q, 1)
+f = hom(F, FQ, gens(FQ), phi)
+W, (t,) = PolynomialRing(QQ, ["t"])
+psi = hom(Q, W, [zero(W), W[1]])
+FW = FreeMod(W, 1)
+g = hom(FQ, FW, gens(FW), psi)
+h = compose(f, g) 
+@test h(F[1]) == FW[1]
+end
```
thofma, line 71: Indentation off.

src/Modules/ModuleTypes.jl
HechtiDerLachs: It's `Nothing`. Yes, I should say that.

src/Modules/UngradedModules.jl
```diff
@@ -4985,13 +4993,30 @@ end
...
+             Hecke.MapFromFunc(x->(base_ring_map(g)(base_ring_map(h)(x))), 
+                                   base_ring(domain(h)), 
+                                   base_ring(codomain(g)))
+            )
+
+end
+
 compose(h::ModuleFPHom, g::ModuleFPHom) = h*g
 
+# TODO: We need to also make all of the code below sensitive to base changes!
```
HechtiDerLachs, line 5100: @jankoboehm : I think this is another issue by itself. But should be addressed!

jankoboehm: Think we can postpone until needed.

src/Modules/UngradedModules.jl
```diff
@@ -1261,9 +1276,10 @@ function (==)(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule)
     return false
   end
   #TODO should there be a check for === up to permutation in order to avoid std-computation?
-  M_mod_N = reduce(M, N)
-  N_mod_M = reduce(N, M)
-  return iszero(M_mod_N) && iszero(N_mod_M)
+  # If yes, this could also be incorporated in the `in`-function.
+  all(x->(x in N), gens(M)) || return false
+  all(x->(x in M), gens(N)) || return false
```
simonbrandhorst, line 1281: This does the same but has the advantage that it aborts early?
So this is an attempt at speedup? Can you support this by timing a few examples?

simonbrandhorst (approved):

simonbrandhorst:
Some doctests fail. In most cases just the printing changed. In some cases the output (maybe just a different representative). Please investigate.

src/Modules/UngradedModules.jl
thofma: Could also just be
```julia
return all(x->(x in N), gens(M)) && all(x->(x in M), gens(N))
```

HechtiDerLachs: No. This was to make the code as generic as it was intended to be. `reduce` tried to do groebner basis stuff which it couldn't. 

Edit: This was following a suggestion by @jankoboehm (in oral communication).

HechtiDerLachs:
> 



> Some doctests fail. In most cases just the printing changed. In some cases the output (maybe just a different representative). Please investigate.

Yes, I know. This is, because I changed the homomorphism for the projection to the quotient rings to be a proper homomorphism of rings. I need the `preimage` functionality for it, so I needed to change that. Now it messes up the docstrings.

I just didn't find the time to have my computer run the repair locally (You can not continue programming for one hour if you do that). I will eventually do that tonight.

src/Modules/UngradedModules.jl
```diff
@@ -4543,8 +4590,10 @@ rank   | 0  2  6  6  2
 degree | 4  3  2  1  0
 ```
 """
-function free_resolution(M::SubQuo; ordering::ModuleOrdering = default_ordering(M),
-        length::Int=0, algorithm::Symbol=:fres)
+function free_resolution(M::SubQuo{<:MPolyElem}; 
+    ordering::ModuleOrdering = default_ordering(M),
+    length::Int=0, algorithm::Symbol=:fres
+  )
```
HechtiDerLachs, line 4599: @jankoboehm : I restricted this method to the case where it actually works: Polynomials.

src/Modules/UngradedModules.jl
```diff
@@ -4606,6 +4655,37 @@ function free_resolution(M::SubQuo; ordering::ModuleOrdering = default_ordering(
   return FreeResolution(cc)
 end
 
+function free_resolution(M::SubQuo{T}) where {T<:RingElem}
+  # This generic code computes a free resolution in a lazy way.
```
HechtiDerLachs, line 4662: ...then I implemented another generic method here.

HechtiDerLachs (commented):
@jankoboehm : Some comments on how I re-implemented the free resolution. Please complain or approve. 

(CC @wdecker : You might also have an opinion on this?)

src/Modules/UngradedModules.jl
```diff
@@ -4606,6 +4655,37 @@ function free_resolution(M::SubQuo; ordering::ModuleOrdering = default_ordering(
...
+
+      K, inc = kernel(map(C, i))
+      nz = findall(x->!iszero(x), gens(K))
+      F = FreeMod(R, length(nz))
+      iszero(length(nz)) && set_attribute!(F, :name => "0")
+      phi = hom(F, C[i], iszero(length(nz)) ? elem_type(C[i])[] : inc.(gens(K)[nz]))
+      pushfirst!(C.maps, phi)
+    end
+    return first(C.maps)
+  end
```
HechtiDerLachs: The method does **not** compute the resolution up to a user-specified upper bound, but in a lazy way, incorporated already in the chain complexes.

@fieker: Is this how you intended people to do it? 

I am a bit worried, since I am using the fields of the chain complex directly. But I didn't find any constructor that would accept the `fill` function as an argument. If you have methods for whatever I am doing here: Could you point them out to us?

afkafkafk13:
> Some doctests fail. In most cases just the printing changed. In some cases the output (maybe just a different representative). Please investigate.

I checked them all. @HechtiDerLachs: A direct rebuild of the tests does the job (but no rush to do it right now -- do it anytime it fits in, but before the merge).

src/Rings/MPolyQuo.jl
```diff
@@ -600,13 +610,13 @@ julia> R, (x,) = PolynomialRing(QQ, ["x"]);
 julia> A, p = quo(R, ideal(R, [x^4]));
 
 julia> f = p(x-x^6)
--x^6 + x
+x
```
HechtiDerLachs: The `simplify` command seems to now be carried out automatically. Whoever is responsible for that: Maybe they wish to also adjust the documentation?

src/Rings/solving.jl
```diff
@@ -80,14 +80,20 @@ julia> I = ideal(R, [x1+2*x2+2*x3-1, x1^2+2*x2^2+2*x3^2-x1, 2*x1*x2+2*x2*x3-x2])
...
- [1, 0, 0]
- [1//3, 0, 1//3]
+ERROR: UndefVarError: rational_solutions not defined
+Stacktrace:
+ [1] _rational_solutions(I::MPolyIdeal{fmpq_mpoly}; initial_hts::Int64, nr_thrds::Int64, max_nr_pairs::Int64, la_option::Int64, info_level::Int64, precision::Int64)
+   @ Oscar ~/Oscar.jl/src/Rings/solving.jl:105
+ [2] _rational_solutions(I::MPolyIdeal{fmpq_mpoly})
+   @ Oscar ~/Oscar.jl/src/Rings/solving.jl:93
+ [3] top-level scope
+   @ none:1
```
HechtiDerLachs: This seems wrong. Is the problem somewhere on my side? 

CC: @fieker : Didn't you program this? Did I do anything wrong in my code to mess this up?

fieker:
On Fri, Dec 16, 2022 at 01:17:11AM -0800, Matthias Zach wrote:
> @HechtiDerLachs commented on this pull request.
> 
> 
> 
> > - [1//3, 0, 1//3]
> +ERROR: UndefVarError: rational_solutions not defined
> +Stacktrace:
> + [1] _rational_solutions(I::MPolyIdeal{fmpq_mpoly}; initial_hts::Int64, nr_thrds::Int64, max_nr_pairs::Int64, la_option::Int64, info_level::Int64, precision::Int64)
> +   @ Oscar ~/Oscar.jl/src/Rings/solving.jl:105
> + [2] _rational_solutions(I::MPolyIdeal{fmpq_mpoly})
> +   @ Oscar ~/Oscar.jl/src/Rings/solving.jl:93
> + [3] top-level scope
> +   @ none:1
> 
> This seems wrong. Is the problem somewhere on my side? 
> 
> CC: @fieker : Didn't you program this? Did I do anything wrong in my code to mess this up?
> 
I did a first version, but this one is @ederc it is now hooked up to
msolve

src/Rings/MPolyQuo.jl
```diff
@@ -725,14 +725,7 @@ MPolyQuo{MPolyElem_dec{fmpq, fmpq_mpoly}}
...
   q = MPolyQuo(R, I)
-  function im(a::MPolyElem)
-    parent(a) !== R && error("Element not in the domain of the map")
-    return MPolyQuoElem(a, q)
-  end
-  function pr(a::MPolyQuoElem)
-    return a.f
-  end
-  return q, MapFromFunc(im, pr, R, q)
+  return q, hom(R, q, gens(q))
```
HechtiDerLachs: Now that I look at it again, I am not sure whether this was a good idea. Does the call to `hom` take into account that we only have to map variables to variables? 

@thofma : Do you think we should go back to the old version?

thofma: What is the motivation for changing it (except to clean it up)?

src/Rings/solving.jl
thofma: Can you do a rebase? CI is not running at the moment.

src/Rings/MPolyQuo.jl
HechtiDerLachs: I think I wanted the preimage functionality so that I can take lifts. But that has already been there, so... I don't remember, really. 

I just changed it back, but didn't push it yet. First I run the doctests again.

HechtiDerLachs:
> I did a first version, but this one is @ederc it is now hooked up to msolve

Alright, thanks. I had to do an `]up` before fixing the doctests to update some `msolve` dependency. Maybe something went wrong there. 

@ederc: If you see, what's going wrong here, let us know!

src/Rings/MPolyQuo.jl
HechtiDerLachs: It seems that `simplify` was called by some internal passage through the projection map to the quotient. But now that I changed it back, it's also back to the old version.

src/Rings/solving.jl
HechtiDerLachs: Ok, yes. I didn't notice yet, thanks!

HechtiDerLachs:
> @ederc: If you see, what's going wrong here, let us know!

Sorry! @thofma just pointed out that I needed to do a rebase. Let me try that first and then get back to you, if necessary.

HechtiDerLachs:
The doctests are still failing. I guess this is due to the fact that I have julia 1.8.3 installed on my machine. Could somebody else with the correct julia version eventually run the doctests on their machine and force-push to this branch?

@afkafkafk13 ? @simonbrandhorst ?

lkastner:
> The doctests are still failing. I guess this is due to the fact that I have julia 1.8.3 installed on my machine. Could somebody else with the correct julia version eventually run the doctests on their machine and force-push to this branch?
> 
> @afkafkafk13 ? @simonbrandhorst ?

Ok, I ran `build_doc(doctest=:fix)` with Julia 1.6.7. I squashed the commits fixing the docs together. Hope this is what you had in mind, and also that it works.

src/Modules/mpolyquo.jl
```diff
@@ -0,0 +1,195 @@
...
+  return KK, inc2
+end
+
+function coordinates(
+    v::FreeModElem{T}, 
+    M::SubModuleOfFreeModule{T}, 
+    task::Symbol
+  ) where {T<:MPolyQuoElem}
+  return coordinates(v, M)
+end
```
simonbrandhorst, line 73: why is the `task` ignored?

HechtiDerLachs: Because for generic code, there is no way to make sense of `task`. This is a polynomial internal. But if we don't implement this method, the code doesn't run because dispatch doesn't recognize that we're doing stuff with `MPolyQuo`s.  This should be clarified eventually, either in the documentation of the `SubModuleOfFreeModule`-interface and the methods to be implemented, or in the internal routines. But that can be left for later, the TODOs are documented in the accompanying issue.

Issue #1773, opened 2022-11-29
Integration of bibtool into the CI possible?


HereAround:
"As described in the documentation (https://docs.oscar-system.org/stable/DeveloperDocumentation/documentation/), we use
`bibtool` to standardize the bibkeys and also the bib-entries. Inspired by a discussion with @simonbrandhorst (and also with @lkastner today), I was wondering if we could integrate `bibtool` into the CI.

I guess the dream would be a CI run, that performs as follows:
1. Execute `bibtool`.
2. Check if `docs/oscar_references.bib` changed. Otherwise pass.
3. Apply `bibtool` and check for changes. If application of `bibtool` did not lead to changes, pass.
4. Find out which `bibkeys` changed, replace them (e.g. via `sed`) and add a corresponding commit.

Is something of that sort possible? Maybe @fingolfin knows/has an opinion?

fingolfin:
Checking if running `bibtool` results in a diff, and failing CI in that case? That's fairly easy.

Adding a commit "fixing" things? Very tricky, and possibly confusing, because it'll modify the branch and in the past doing that often resulted in much pain and suffering for non-git-experts. So I don't think it's worth the effort. But of course the generated error should be helpful and point out how to locally address the issue.

HereAround:
My hope was to actually remove the step "apply bibtool" from our (=human) workflow. But yes, I suspected that the fixing part would be tricky, in particular in the CI. Thank you for confirming @fingolfin .

The next best thing that comes to my mind is a fixing script that executes the changes **locally**. This is probably less confusing but still tricky. But I am not sure if this is feasible/efficient. It merely replaces the task "apply bibtool" with "apply new shiny script". In addition, I like to think that we usually add only a few references per PR, so the necessary changes will not take too long by hand.

Maybe the only reasonable adjustment is a CI check to see if running `bibtool` results in a diff?

thofma:
I agree that we should just check for a fixpoint, i.e., `bibtool(bla.bib) == bla.bib` and throw an instructive error in case it fails.

thofma:
Fixed in #1807

PR #1807, opened 2022-12-09
Introduce bibtool CI test

HereAround:
A first proposal following the discussion in https://github.com/oscar-system/Oscar.jl/issues/1773.

cc @fingolfin @thofma

### DIFF
--- a/.github/workflows/BibtoolCI.yml
+++ b/.github/workflows/BibtoolCI.yml
@@ -1,0 +1,40 @@
+name: Bibtool test
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  # group by workflow and ref; the last slightly strange component ensures that for pull
+  # requests, we limit to 1 concurrent job, but for the master branch we don't
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref != 'refs/heads/master' || github.run_number }}
+  # Cancel intermediate builds, but only if it is a pull request build.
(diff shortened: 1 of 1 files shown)


thofma:
Could you add a commit here which triggers a failure? Just for testing purposes.

HereAround:
https://github.com/oscar-system/Oscar.jl/pull/1807/commits/8acb99eb6c6b6cb2596f1a582577b4a6939b07f1 ~~will cause~~ causes a failure.

fingolfin (approved):
Looks good to me, though of course we should wait to see if it detects the failure (and then remove the cause for the failure etc., whatever ;-)

HereAround:
> Looks good to me, though of course we should wait to see if it detects the failure (and then remove the cause for the failure etc., whatever ;-)

Thank you @fingolfin . Yes, I will of course remove this dummy commit. I assume @thofma just wanted to check that this test detects a failure.

HereAround:
Btw. the bibtool test has already detected the failure as introduced in https://github.com/oscar-system/Oscar.jl/commit/8acb99eb6c6b6cb2596f1a582577b4a6939b07f1 for test purposes. This test completes - as we should expect - very quickly (at most a few minutes). It could be even faster if we did not have to install `bibtool` on the server.

thofma:
Yes, looks good. Could you separate the installation step from the actual test? Then there is less output if it fails.

HereAround:
> Yes, looks good. Could you separate the installation step from the actual test? Then there is less output if it fails.

Done. Also removed the dummy commit (I assume we have all seen that this leads to a failure.)

thofma (approved):

PR #1809, opened 2022-12-09
Coherent sheaves

HechtiDerLachs:
This is a draft on coherent sheaves on `AbsCoveredScheme`s. As an example, we implemented twisting sheaves on projective space and the cotangent sheaf of an arbitrary covered scheme. See the test file for samples.


experimental/Schemes/CoherentSheaves.jl
```diff
@@ -0,0 +1,461 @@
...
+      # charts, this might be a non-trivial change of bases, however.
+      dom_sub, inc = sub(F(U), rf)
+      B = [coordinates(e, dom_sub) for e in ambient_representatives_generators(F(U))]
+      images = elem_type(MU)[]
+      for phi in gens(MV)
+        phi_map = element_to_homomorphism(phi)
+        images_f = [sum([B[i][j]*cod_res(phi_map(f[j])) for j in 1:length(f)]) for i in 1:length(B)]
+        psi = hom(F(U), G(U), images_f)
+        push!(images, homomorphism_to_element(MU, psi))
+      end
```
HechtiDerLachs: @jankoboehm : I implemented the restriction of homomorphisms to open subsets here. But I'm not sure, I am using your machinery as intended. In particular the conversion of elements to homomorphisms and back seems to be a bit tedious. Do you provide something like "induced hom from left" and "induced hom from right" for "composition" of Hom-modules with homomorphisms to the domain and from the codomain?

jankoboehm: The construction from what we have is not so difficult and too many shortcuts might in the end be difficult to remember and confusing, but I think it is a good idea to introduce it to provide all the automatisms of Hom modules. We have lift_homomorphism_contravariant and lift_homomorphism_contravariant, though not exactly what you want.

HechtiDerLachs: Ok, I'll have a look. But then you say that my solution is in the scope of how you intended users to handle the modules? 

If not, feel free to propose another solution here!

simonbrandhorst:
Why does this pr change doctests all over the place? This looks fishy.

HechtiDerLachs:
Yes, it's fishy. That is because I had to run the doctests with `:fix` once after changing the type of the map for the projections to quotient rings and that affected a lot of files. These changes have been reverted, but the doctests are still in the old state. Now, when I run them with `:fix` again, I will still mess things up, because I have the wrong julia version installed. I will take care of this, eventually.

HechtiDerLachs:
The other question is: How can I clean up this huge mess in the commit-history? 

The background is that I had to develop two PRs in parallel, this one and the one on modules over `MPolyQuo`s. To this end, I shifted a lot of commits back and forth between the two using `git cherry-pick`. But that makes them appear as different commits in history despite doing the exact same changes. A rebase on master seemed impossible here, so I did a merge. But that leaves us with the duplication mess above. Any suggestions how to avoid this in the future is most welcome (via the usual channels). 

My apologies for what you see above. I guess, we have to squash this one eventually.

simonbrandhorst:
This PR is just open to discuss on coherent sheaves. It is not planned to merge this.

PR #1811, opened 2022-12-10
Borcherds' Method

simonbrandhorst:
Computes automorphism groups of K3 surfaces.

Per @fieker s request on slack: here is the code.


fieker (approved):
not that I understand the material, but it looks like it is well written (and I wanted the stuff to be in as early as possible)

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+    membership_test(g) = is_pm1_on_discr(SS,g)
+  end
+
+  d = exponent(discriminant_group(S))
+  Rdual = dual(R)
+  sv = short_vectors(rescale(Rdual,-1), 2)
+  # not storing the following for efficiency
+  # append!(sv,[(-v[1],v[2]) for v in sv])
+  # but for convenience we include zero
+  T = typeof(sv).parameters[1].parameters[1].parameters[1]
```
fingolfin: Huh, this looks nasty (no offense). What does it do? Perhaps a comment could be added to explain that? It is not immediately obvious to me what type `sv` has and hence what this does.

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+end
+
+################################################################################
+# Chambers
+################################################################################
+
+@doc Markdown.doc"""
+    K3Chamber
+
+The chamber in `S` induced from a Weyl vector in `L`.
```
fingolfin: Huh? What are `S` and what are `L`?

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+    return new()
+  end
+end
+
+export chamber, BorcherdsCtx
+
+@doc Markdown.doc"""
+    chamber(data::BorcherdsCtx, weyl_vector::fmpz_mat, [parent_wall::fmpz_mat, walls::Vector{fmpz_mat}])
+
+Return the chamber with the given weyl vector.
```
fingolfin:
```suggestion
Return the chamber with the given Weyl vector.
```

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+Return the walls of the chamber `D`, i.e. its facets.
+
+The corresponding half space of the wall defined by `v` in `walls(D)` is
+
+```Math
+\{x \in S \otimes \mathbb{R} |  \langle x,v \rangle  \geq 0\}.
+```
+
+`v` is given with respect to the basis of `S` and is primitive in `S`.
+Note that Shimada follows a different convention and takes `v` primitive in `S^\vee`.
```
fingolfin: Purely optional, but:
- "Shimada says" is a bit vague, perhaps this could be turned into a reference?
- perhaps also a reference could be given that *does* follow these conventions? At least it seems a bit odd to me to only explain which convention we do *not* follow ;-)

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+  end
+  return D.walls
+end
+
+weyl_vector(D::K3Chamber) = D.weyl_vector
+
+@doc Markdown.doc"""
+    rays(D::K3Chamber)
+
+Return the rays of the induced chamber `D`.
```
fingolfin: Is `D` the "induced chamber"? Then perhaps just remove the "induced"? Or does the chamber `D` induce some other chamber whose rays are returned? Then this should be explained perhaps?

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+
+They are represented as primitive row vectors with respect to the basis of `S`.
+"""
+function rays(D::K3Chamber)
+  r = reduce(vcat, walls(D), init=zero_matrix(ZZ,0,rank(D.data.SS)))
+  rQ = change_base_ring(QQ, r) * gram_matrix(D.data.SS)
+  C = positive_hull(rQ)
+  Cd = polarize(C)
+  L = rays(Cd)
+  Lq = fmpq_mat[matrix(QQ,1,rank(D.data.SS),i) for i in L]
```
fingolfin, line 309: This function seems to be untested according to Codecov.

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+  # So far we have only O(S)-invariants. There are also ways to produce G-invariants
+  # by working the the images of the rays in the discriminant group and their
+  # G-orbits. Perhaps one has to switch to S^\vee primitive vectors in this case.
+  =#
+  return (m1, m2, m3, m4)
+end
+
+
+"""
+Compute the fingerprint of Plesken-Souvignier and change the basis
```
fingolfin: Do you mean this?
```suggestion
Compute the fingerprint defined by Plesken-Souvignier and change the basis
```

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+matrix and gram matrix accordingly.
+
+It is computed from the walls and the gram matrix of `S`.
+
+a permutation per for the
```
fingolfin:
```suggestion
A permutation `per` for the
```

src/Geometry/K3Auto.jl
```diff
@@ -0,0 +1,2059 @@
...
+  lengths = fmpq[(v*gramS*transpose(v))[1,1] for v in V]
+  D.lengths = lengths
+  gramB = change_base_ring(ZZ, B*gramS*transpose(B))
+  D.gramB = gramB
+  D.B = B
+
+  per = Vector{Int}(undef, n)
+  for i in 1:n
+    per[i] = i
+  end
```
fingolfin, line 410:
```suggestion
  per = collect(1:n)
```

fingolfin (commented):
Thank you! Some nitpicks. These are not meant to block or to annoy, feel free to ignore, but perhaps some are useful anyway?

src/Geometry/K3Auto.jl
simonbrandhorst: Does this differ in performance or in style only?

src/Geometry/K3Auto.jl
simonbrandhorst: Now that the interface for `short_vectors` is finalized I can rewrite this part. Thank you for pointing me to it.

src/Geometry/K3Auto.jl
simonbrandhorst: Me, I follow them :-). But I will add the relevant publications.

PR #1813, opened 2022-12-12
Orthogonal group for `TorQuadMod` extended

StevellM:
Up to now, one could only compute the orthogonal group of a semi-regular `TorQuadMod`. We extend these methods to the case where the gram matrix of the underlying quadratic form is trivial, and also to the non semi-regular case when the radical quadratic splits (which bring us back in some combination of both of the previous cases).


src/Groups/matrices/FiniteFormOrthogonalGroup.jl
```diff
@@ -798,38 +798,135 @@ function _compute_gens(T::TorQuadMod)
...
+  ok, phi = is_isometric_with_isometry(Torth, T)
+  @assert ok # Same module with different basis, somehow
+
+  # We compute the orthoognal group of each (primary) block. Since Torth is made
+  # diagonally, we can generate its orthogonal group by taking some diagonal matrices
+  # from taking the cartesian product on the generators of the blocks. We have nothing
+  # to add since we can't map different primary parts to each other.
+  orth_blocks = _compute_gens_split_degenerate_primary.(domain.(blocks))
+  gensOTorth = fmpz_mat[]
+  for x in Hecke.cartesian_product_iterator(orth_blocks, inplace=false)
```
simonbrandhorst: This looks like you take the direct sums of all generators. 
But this gives too many generators. Just take `diagonal_matrix(id, id, ... id, f, id, ... id)`
Then you get $\sum_i n_i$ generators instead of $\prod_i n_i$ many.

src/Groups/matrices/FiniteFormOrthogonalGroup.jl
```diff
@@ -798,38 +798,135 @@ function _compute_gens(T::TorQuadMod)
...
+  # for the rd, since its quadratic form is trivial, automorphisms are just
+  # automorphisms of the underlying abelian group.
+  gensOrd = fmpz_mat[]
+  OArd = automorphism_group(abelian_group(rd))
+  for f in gens(OArd)
+    push!(gensOrd, matrix(f))
+  end
+  n2 = nrows(gensOrd[1])
+
+  # finally, we have to consider automorphism which maps N into rd: these are well
```
simonbrandhorst, line 899:
```suggestion
  # finally, we have to consider automorphism which map rd into N+rd: these are well
```

simonbrandhorst:
```suggestion
  # finally, we have to consider automorphism which maps N into rd+N: these are well
```

src/Groups/matrices/FiniteFormOrthogonalGroup.jl
```diff
@@ -798,38 +798,135 @@ function _compute_gens(T::TorQuadMod)
...
+  R, psi = hom(abelian_group(N), abelian_group(rd), task = :map)
+  Ntord = fmpz_mat[matrix(psi(f)) for f in gens(R)]
+  Torth, _, _ = Hecke._orthogonal_sum_with_injections_and_projections([rd, N])
+
+  # Same module with different basis
+  ok, phi = is_isometric_with_isometry(Torth, T)
+  @assert ok
+
+  # We have two kind of generators: block diagonal matrices coming from isometries
+  # of N and rd, and those acting identically on N and rd but which send N to rd.
```
simonbrandhorst:
```suggestion
  # of N and rd, and those acting identically on N and rd but which send N to rd+N.
```

simonbrandhorst:
Thank you. I think this looks good in principle.

simonbrandhorst (approved):

PR #1814, opened 2022-12-13
Fix1794

### LINKED ISSUE #1794: _compute_standard_basis_with_transform does not work over number field coefficients
I stumbled upon the following issue when inverting elements in `MPolyQuo`s over number fields. Try the following code:
```julia
P, t = QQ["θ"]
p = t^2+1
kk = splitting_field(p)

R, (x,y,z) = kk["x", "y", "z"]
Oscar._compute_standard_basis_with_transform(ideal(R, [one(R)]))
```
It says
```
ERROR: MethodError: no method matching _compute_standard_basis_with_transform(::MPolyIdeal{AbstractAlgebra.Generic.MPoly{nf_elem}})
```

Edit: The original code calling this would have been something like the following:
```julia
P, t = QQ["θ"]
p = t^2+1
kk = splitting_field(p)

R, (x,y,z) = kk["x", "y", "z"]

Q, _ = quo(R, ideal(R, []))
inv(one(Q))
```

HechtiDerLachs:
This fixes #1794 .

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -777,12 +777,12 @@ zero(Q::MPolyQuo) = Q(0)
 one(Q::MPolyQuo) = Q(1)
 
 function is_invertible_with_inverse(a::MPolyQuoElem)
-  # temporary fix
-  b = a.f
-  R = parent(b)
-  J = ideal(R, vcat([b], gens(modulus(parent(a)))))
-  one(R) in J || return false, zero(a)
-  return true, parent(a)(coordinates(one(R), J)[1])
+ # TODO:
+ # Eventually, the code below should be replaced 
+ # by a call to `coordinates` over the ring `parent(a)`. 
+ # This should then use relative groebner bases and 
+ # make use of the caching of previously computed GBs 
+ # of the modulus of `parent(a)`. 
 
   Q = parent(a)
   I = Q.I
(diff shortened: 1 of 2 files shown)


thofma:
Looks good to me and I think using `coordinates` is the right thing to do as you said.

thofma (approved):

thofma:
Ah, wait, you don't use `coordinates` anymore.

HechtiDerLachs:
> Ah, wait, you don't use `coordinates` anymore.

No, I don't! And for a reason: The current implementation which I now repaired according to @wdecker 's suggestion, is making use of previously computed GBs for the modulus of the quotient ring. I do not know whether calling `coordinates` directly on `a` in an `MPolyQuoIdeal` already does the right thing in the background. In fact, I remember that I had provided another dirty fix for that some weeks ago, so I really expect this to be less efficient. 

The ideals and their Groebner bases are currently undergoing some restructuring process and I am not sufficiently informed about the state of the art at this point. Hence, I preferred to just "leave it as is" now and write a TODO there for when the time's ripe; also given that this is not the scope of my work right now and I don't want to cause any interference with ongoing work of others.

wdecker (approved):
This fixes the function for now. Indeed, we will adjust Gröbner bases for affine algebras at some point and then take a look at this, too.

PR #1818, opened 2022-12-13
Length of modules

HechtiDerLachs:
We compute the length of a finite module over a local ring and an associated decomposition series.

### DIFF
--- a/src/Modules/local_rings.jl
+++ b/src/Modules/local_rings.jl
@@ -1,0 +1,41 @@
+export length
+export decomposition_series
+
+function length(M::ModuleFP{RingElemType}) where {RingElemType<:AbsLocalizedRingElem{<:Any, <:Any, <:MPolyComplementOfPrimeIdeal}}
+  if iszero(M) 
+    set_attribute!(M, :decomposition_series, Vector{elem_type(M)}())
+    return 0
+  end
+  R = base_ring(M)
+  P = prime_ideal(inverted_set(R))
+  F = FreeMod(R, 1)
+  N, _ = quo(F, (R(P)*F)[1])
+  k = 0
+  MM = M
+  p = identity_map(M)
+  NtoM, interp = hom(N, MM)
+  decomp = Vector{elem_type(M)}()
+  while !iszero(NtoM)
+    g = gens(NtoM)
+    i = findfirst(x->!(iszero(x)), g)
+    Msub, inc = sub(MM, [interp(g[i])(N[1])])
+    push!(decomp, preimage(p, inc(Msub[1])))
+    MM, pp = quo(MM, Msub)
+    p = compose(p, pp)
+    NtoM, interp = hom(N, MM)
+    k = k + 1
+  end
+  iszero(MM) || error("module does not have finite length")
+  set_attribute!(M, :decomposition_series, decomp)
+  return k
+end
+
+function decomposition_series(
+    M::ModuleFP{RingElemType}
+  ) where {RingElemType<:AbsLocalizedRingElem{<:Any, <:Any, <:MPolyComplementOfPrimeIdeal}}
+  if !has_attribute(M, :decomposition_series)
+    length(M)
+  end
+  return get_attribute(M, :decomposition_series)::Vector{elem_type(M)}
+end
+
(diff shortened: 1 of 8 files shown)


test/Modules/local_rings.jl
```diff
@@ -0,0 +1,30 @@
...
+
+  N, _ = quo(F, (ideal(L, [x,y,z])*F)[1])
+  @test length(N) == 0
+
+  @test_throws "module does not have finite length" length(F)
+
+  A, _ = quo(R, ideal(R, [x^2, y, z]))
+  W, _ = localization(A, complement_of_ideal(ideal(R, [x,y,z])))
+  F = FreeMod(W, 2)
+  @test_broken length(F) == 4 # See #1806
```
HechtiDerLachs: This should indeed be fixed by that branch.

src/Modules/local_rings.jl
```diff
@@ -0,0 +1,41 @@
...
+  N, _ = quo(F, (R(P)*F)[1])
+  k = 0
+  MM = M
+  p = identity_map(M)
+  NtoM, interp = hom(N, MM)
+  decomp = Vector{elem_type(M)}()
+  while !iszero(NtoM)
+    g = gens(NtoM)
+    i = findfirst(x->!(iszero(x)), g)
+    Msub, inc = sub(MM, [interp(g[i])(N[1])])
```
afkafkafk13, line 21: I know that this remark has nothing to do with this pull-request, but 'interp' is a non-intuitive name....

src/Modules/local_rings.jl
```diff
@@ -0,0 +1,41 @@
...
+    return 0
+  end
+  R = base_ring(M)
+  P = prime_ideal(inverted_set(R))
+  F = FreeMod(R, 1)
+  N, _ = quo(F, (R(P)*F)[1])
+  k = 0
+  MM = M
+  p = identity_map(M)
+  NtoM, interp = hom(N, MM)
```
afkafkafk13, line 21: interp is a non-intuitive name for this map, but I do not have a better name off-hand either.

src/Modules/local_rings.jl
```diff
@@ -0,0 +1,41 @@
...
+    p = compose(p, pp)
+    NtoM, interp = hom(N, MM)
+    k = k + 1
+  end
+  iszero(MM) || error("module does not have finite length")
+  set_attribute!(M, :decomposition_series, decomp)
+  return k
+end
+
+function decomposition_series(
```
afkafkafk13: Do you want a 'composition series' ?  (question on terminology)

src/Modules/local_rings.jl
```diff
@@ -0,0 +1,41 @@
...
+    M::ModuleFP{RingElemType}
+  ) where {RingElemType<:AbsLocalizedRingElem{<:Any, <:Any, <:MPolyComplementOfPrimeIdeal}}
+  if !has_attribute(M, :decomposition_series)
+    length(M)
```
afkafkafk13: Using this sideeffect of length does not seem structurally honest.... Should length not be relying on composition_series in the case of localizations at complements of prime ideals?
(by the way: in what file is the case of localization at MPolyComplementOfKPointIdeal handled? That one should not pass through a composition series, but use combinatorics on the leading module -- and hence not have this sideeffect.)

src/Modules/mpolyquo-localizations.jl
```diff
@@ -98,3 +98,28 @@ function pre_saturated_module(M::SubQuo{T}) where {T<:MPolyQuoLocalizedRingElem}
...
+    T<:MPolyQuoLocalizedRingElem, 
+    DomType<:FreeMod{T},
+    CodType<:FreeMod{T}
+  }
+  S = base_ring(domain(f))
+  A = representing_matrix(f)
+  B, D = clear_denominators(A)
+  Fb = base_ring_module(domain(f))
+  Gb = base_ring_module(codomain(f))
+  GGb, p = quo(Gb, [g*e for g in gens(modulus(underlying_quotient(S))) for e in gens(Gb)])
```
afkafkafk13, line 116: I do not see, where you use GGb later on. You are only aiming for the p here?

HechtiDerLachs: Yes.

afkafkafk13 (approved):

HechtiDerLachs:
Are we good to merge?

PR #1821, opened 2022-12-14
Adds Hilbert driven standard basis computation

RafaelDavidMohr:
This adds the function `_groebner_basis_with_hilbert` and adds calling it as an `:algorithm` option to `standard_basis` and `groebner_basis`. 

This function computes the hilbert series of an ideal (probabilistically over a randomly chosen prime if the characteristic of the underlying ring is zero) and passes it to `std_hilbert`. 

One thing I am unsure about: Should the user be able to pass the weights into `standard_basis` and `groebner_basis` explicitly or should they be extracted out of the underlying polynomial ring?


thofma:
The docstring says `Int`, but the implementation says `Int32`. I would suggest to use `Int` everywhere one the julia side. Since there is already an explicit conversion before calling Singular, this should be fine.

If the Hilbert series is wrong, will the Gröbner/standard basis still be correct?

ederc:
> If the Hilbert series is wrong, will the Gröbner/standard basis still be correct?

The process the Hilbert series is constructed is semicontinuous, so even if the Hilbert series is not correct (due to a bad prime for example) the computation might only be less efficient (more S-polys to be checked w.r.t. the destination ordering), but the resulting basis will still be correct.

src/Rings/groebner.jl
```diff
@@ -137,7 +141,7 @@ negdegrevlex([x, y])
 ```
 """
 function standard_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I)),
-	complete_reduction::Bool = false, algorithm::Symbol = :buchberger)
+	complete_reduction::Bool = false, algorithm::Symbol = :buchberger, weights::Vector{E} = ones(Int32, ngens(base_ring(I)))) where {E <: Integer}
```
ederc: We do not add `weights` to `standard_basis()`, same holds for `groebner_basis()`. We need to compute this data internally.

src/Rings/groebner.jl
```diff
@@ -249,9 +258,9 @@ julia> leading_coefficient(G[8])
 ```
 """
 function groebner_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I)), complete_reduction::Bool=false,
-	algorithm::Symbol = :buchberger)
+	algorithm::Symbol = :buchberger, weights::Vector{E} = ones(Int32, ngens(base_ring(I)))) where {E <: Integer}
```
ederc: See above.

src/Rings/groebner.jl
```diff
@@ -1221,3 +1232,104 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
 	dim(I) == 0 || error("Dimension of ideal must be zero.")
 	I.gb[destination_ordering] = _fglm(G, destination_ordering)
 end
+
+
+# TODO: does the target ordering have to be global?
+@doc Markdown.doc"""
+    _groebner_basis_with_hilbert(I::MPolyIdeal,
```
ederc: Make this an exported function called `groebner_basis_hilbert_driven()`, we will later on adjust calls to `f4` and `fglm` to this naming convention. Here you correctly have the `weights` as parameters, but we do not have them for the general `..._basis()` calls.

src/Rings/groebner.jl
```diff
@@ -1221,3 +1232,104 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+  all(p -> _is_homogeneous_weights(p, weights), gens(I)) || error("I must be given by generators homogeneous with respect to given weights.")
+	isa(coefficient_ring(base_ring(I)), AbstractAlgebra.Field) || error("The underlying coefficient ring of I must be a field.")
+	haskey(I.gb, destination_ordering) && return I.gb[destination_ordering]
+  if isempty(I.gb) && iszero(characteristic(base_ring(I)))  
+    while true
+      p = 0
+      while true
+        p = rand(2^15:2^16)
+        if isprime(p)
+          break
```
ederc:
```suggestion
    r = rand(2^15:2^16)
    p = next_prime(r)
```

Maybe that's more efficient?

src/Rings/groebner.jl
```diff
@@ -1221,3 +1232,104 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
 	dim(I) == 0 || error("Dimension of ideal must be zero.")
 	I.gb[destination_ordering] = _fglm(G, destination_ordering)
 end
+
+
+# TODO: does the target ordering have to be global?
```
ederc: Yes, only global orderings are supported.

RafaelDavidMohr:
Adjusted the code based on your reviews, see the last few commits. If I understood correctly, the `std_hilbert` function in Singular does not work with general matrix orderings as well, so I added a corresponding error message.

src/Rings/groebner.jl
```diff
@@ -1221,3 +1231,112 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+      p = Hecke.next_prime(r)
+        
+      base_field = GF(p)
+      ModP, _ = PolynomialRing(base_field, "x" => 1:ngens(base_ring(I)))
+      I_mod_p_gens = Vector{elem_type(ModP)}(undef, length(gens(I))) 
+      try
+        I_mod_p_gens = [ModP(f) for f in gens(I)]
+        G = f4(ideal(ModP, I_mod_p_gens))
+        break
+      catch
```
thofma: What is the possible error? The error should be checked and rethrown if it does not fit.

RafaelDavidMohr: This error is caused if one of the coefficients of the input polynomials is divisible by our prime `p`. I changed the code to handle specifically only this error (see https://github.com/oscar-system/Oscar.jl/pull/1821/commits/89e6371abc0aabc6da63938f43b2f637fc9c3df6), I don't know if there is a better way to handle this.

thofma: Looks good, thanks.

RafaelDavidMohr:
I've modified `standard_basis` to homogenize the input before running the hilbert driven GB algorithm if it is not already homogeneous and dehomogenize the resulting Gröbner basis to obtain one for the original input. This only works if the `base_ring` of the input is not already a graded ring, I am not sure what to do in this case. See https://github.com/oscar-system/Oscar.jl/pull/1821/commits/5b03dcde8bed103b53cc805747488a0d266e229f and https://github.com/oscar-system/Oscar.jl/pull/1821/commits/1fd6300be3232cc8efc4e61b933fc7b0a340be5e for the corresponding changes.

src/Rings/groebner.jl
```diff
@@ -1221,3 +1248,126 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+Compute a Gröbner basis of `I` with respect to `destination_ordering`
+using a Hilbert Series driven method as follows: If a Gröbner basis
+for `I` is present, compute the Hilbert series of `I` and use it to
+optimize the Gröbner basis computation for `I`
+w.r.t. `destination_ordering`. If no Gröbner basis for `I` is present
+compute the Hilbert series for `I` if the base field of `I` has
+positive characteristic, otherwise compute the Hilbert series for `I`
+modulo a randomly chosen prime. Use the resulting Hilbert series to
+optimize the Gröbner basis computation for `I`
+w.r.t. `target_ordering`.
```
ederc: There is no `target_ordering`

src/Rings/groebner.jl
```diff
@@ -1221,3 +1248,126 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
 	dim(I) == 0 || error("Dimension of ideal must be zero.")
 	I.gb[destination_ordering] = _fglm(G, destination_ordering)
 end
+
+
+@doc Markdown.doc"""
+    groebner_basis_hilbert_driven(I::MPolyIdeal,
+                                  destination_ordering::MonomialOrdering;
```
ederc: Please rename it to `ordering` and make it a keyword argument to coincide with our naming/argument convention.

ederc:
Please add some tests, for example,

```
R,(x,y) = PolynomialRing(QQ, ["x","y"])
I = ideal([x*(x+1), x^2-y^2+(x-2)*y])
standard_basis(I, ordering=lex(R), algorithm=:hilbert)
```

is broken.

src/Rings/groebner.jl
RafaelDavidMohr: What should be the default value for `ordering`?

ederc: I would propose that we do not have default value here. For me it is the same as with `fglm`, we do not know the destination ordering of the GB, this is more of a specialized function where we expect the users to know what they want.

src/Rings/groebner.jl
```diff
@@ -1,7 +1,8 @@
 export  reduce, reduce_with_quotients, reduce_with_quotients_and_unit, f4, fglm,
 		standard_basis, groebner_basis, standard_basis_with_transformation_matrix,
 		groebner_basis_with_transformation_matrix,
-		leading_ideal, syzygy_generators, is_standard_basis, is_groebner_basis
+	leading_ideal, syzygy_generators, is_standard_basis, is_groebner_basis,
+  groebner_basis_hilbert_driven
```
fingolfin: Please don't use tabs (ideally configure your editor to not use tabs in the OSCAR code base -- e.g. by configuring to honor your `.editorconfig` file, see https://editorconfig.org).
```suggestion
		leading_ideal, syzygy_generators, is_standard_basis, is_groebner_basis,
		groebner_basis_hilbert_driven
```

src/Rings/groebner.jl
```diff
@@ -150,6 +153,22 @@ function standard_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_orde
 		end
 	elseif algorithm == :fglm
 		_compute_groebner_basis_using_fglm(I, ordering)
+  elseif algorithm == :hilbert
+    if typeof(base_ring(I)) <: MPolyRing_dec
```
fingolfin: Better:
```suggestion
    if base_ring(I) isa MPolyRing_dec
```
Potentially even better: use a helper function with two methods that does this, but only if it can be reused (so don't bother)

src/Rings/groebner.jl
```diff
@@ -249,7 +269,7 @@ julia> leading_coefficient(G[8])
 ```
 """
 function groebner_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I)), complete_reduction::Bool=false,
-	algorithm::Symbol = :buchberger)
+                        algorithm::Symbol = :buchberger)
```
fingolfin, line 272: Here you de-tab, thank you :-)

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+with respect to the ordering
+deglex([x, y, z])
+```
+"""
+
+function groebner_basis_hilbert_driven(I::MPolyIdeal{P};
+                                       ordering::MonomialOrdering,
+                                       complete_reduction::Bool = false) where {P <: MPolyElem_dec}
+  
+  all(p -> is_homogeneous(p), gens(I)) || error("I must be given by generators homogeneous with respect to its underlying ring.")
```
fingolfin:
```suggestion
  all(is_homogeneous, gens(I)) || error("I must be given by generators homogeneous with respect to its underlying ring")
```
or even just
```suggestion
  all(is_homogeneous, gens(I)) || error("I must be given by homogeneous generators")
```

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+	isa(coefficient_ring(base_ring(I)), AbstractAlgebra.Field) || error("The underlying coefficient ring of I must be a field.")
+  is_global(ordering) || error("Destination ordering must be global.")
+	haskey(I.gb, ordering) && return I.gb[ordering]
+  if isempty(I.gb) && iszero(characteristic(base_ring(I)))  
+    while true
+      r = rand(2^15:2^16)
+      p = Hecke.next_prime(r)
```
fingolfin: Just wondering, but: is there any particular reason to pick a random `r` each time (and thus potentially reusing the same prime, instead of just doing this:
```suggestion
    p = 32771
    while true
      p = Hecke.next_prime(p)
```

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+          ctx = MPolyBuildCtx(ModP)
+          for (c, m) in zip(coefficients(f), exponents(f))
+            push_term!(ctx, base_field(c), m)
+          end
+          I_mod_p_gens[i] = finish(ctx)
+        end
+        G = groebner_assure(ideal(ModP, I_mod_p_gens), default_ordering(ModP))
+        break
+      catch e
+        if e == ErrorException("Unable to coerce")
```
fingolfin, line 1302: Maybe add a comment that explains where this is supposed to be (potentially) thrown? Presumably it is just the conversion/coercion to `ModP`, but not the call to `groebner_assure`? If the `groebner_assure` can also throw, that should be mentioned. If it can't, perhaps move it out of the `try/catch`?

Thing is, I am very slightly concerned about future breakage where the hardcoded  error message changes and we then might have a hard time restoring this, so it's helpful to not what is supposed to throw

Maybe we should add a `CoercionException` or so?

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+"""
+
+function groebner_basis_hilbert_driven(I::MPolyIdeal{P};
+                                       ordering::MonomialOrdering,
+                                       complete_reduction::Bool = false) where {P <: MPolyElem_dec}
+  
+  all(p -> is_homogeneous(p), gens(I)) || error("I must be given by generators homogeneous with respect to its underlying ring.")
+	isa(coefficient_ring(base_ring(I)), AbstractAlgebra.Field) || error("The underlying coefficient ring of I must be a field.")
+  is_global(ordering) || error("Destination ordering must be global.")
+	haskey(I.gb, ordering) && return I.gb[ordering]
```
fingolfin:
```suggestion
  haskey(I.gb, ordering) && return I.gb[ordering]
```

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+  end
+
+  if characteristic(base_ring(I)) > 0 && ordering == default_ordering(base_ring(I))
+    return G
+  end
+
+  singular_assure(G)
+  weights = _extract_weights(base_ring(G))
+  h = Singular.hilbert_series(G.S, weights)
+	singular_assure(I.gens, ordering)
```
fingolfin: Odd formatting follows due to tabs vs spaces

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+function _extract_weights(T::MPolyRing)
+  if typeof(T) <: MPolyRing_dec
+    if !is_z_graded(T)
+      error("Ring must be graded by the Integers.")
+    end
+    return [Int(first(gr_elem.coeff)) for gr_elem in T.d]
+  else
+    return ones(Int, nvars(T))
+  end
+end
```
fingolfin:
```suggestion
function _extract_weights(T::MPolyRing_dec)
  if !is_z_graded(T)
    error("Ring must be graded by the Integers.")
  end
  return [Int(first(gr_elem.coeff)) for gr_elem in T.d]
end

function _extract_weights(T::MPolyRing)
  return ones(Int, nvars(T))
end
```

test/Rings/groebner-test.jl
```diff
@@ -1,3 +1,4 @@
+using Test: Error
```
fingolfin: Why is this needed? If it isn't, please remove again.

src/Rings/groebner.jl
```diff
@@ -1221,3 +1243,124 @@ function _compute_groebner_basis_using_fglm(I::MPolyIdeal,
...
+      try
+        # we have to coerce to ModP like this because the base ring of `I` is
+        # graded
+        for (i, f) in enumerate(gens(I))
+          ctx = MPolyBuildCtx(ModP)
+          for (c, m) in zip(coefficients(f), exponents(f))
+            push_term!(ctx, base_field(c), m)
+          end
+          I_mod_p_gens[i] = finish(ctx)
+        end
```
fingolfin: I have not tested this, but: can this perhaps be simplified to
```suggestion
        for (i, f) in enumerate(gens(I))
          I_mod_p_gens[i] = map_coefficients(base_field, f; parent=ModP)
        end
```
or even
```suggestion
        I_mod_p_gens = [map_coefficients(base_field, f; parent=ModP) for f in gens(I)]
```

src/Rings/groebner.jl
```diff
@@ -322,6 +342,8 @@ function f4(
     return I.gb[ord]
 end
 
+  
+
```
fingolfin:
```suggestion
```

fingolfin (commented):
Thank you for working on this. I have some additional minor feedback

src/Rings/groebner.jl
RafaelDavidMohr: I wrote the error message in this way because the base ring of `I` may have non-standard grading and wanted to make it completely clear that homogeneity is meant with respect to this base ring. Do you nonetheless prefer your version?

src/Rings/groebner.jl
RafaelDavidMohr: I added such a comment, see https://github.com/oscar-system/Oscar.jl/pull/1821/commits/1c944f41fd60c7aeee67eeb3d607ebd763757e83. I'd be happy to also try and add a `CoercionException`.

test/Rings/groebner-test.jl
RafaelDavidMohr: It is not, my language server just keeps adding in these import statements if it thinks something is out of scope... I will pay closer attention to it in the future.

src/Rings/groebner.jl
RafaelDavidMohr: It can! Added in https://github.com/oscar-system/Oscar.jl/pull/1821/commits/ea977e7db3ee648e386f4a30508a24ae9ecab226

src/Rings/groebner.jl
fingolfin: If you prefer to leave the message, that's fine by me -- that's why I made two alternative suggestions. Just apply the first?

src/Rings/groebner.jl
fingolfin: `CoercionException` is way beyond the scope of this PR, it shouldn't be hold up by it :-)

fingolfin (approved):
I've pushed two minor commits. Thanks!

PR #1822, opened 2022-12-14
first step in the Oscarification of pc groups

ThomasBreuer:
- changed the definition of the `PcGroup` type from the GAP type `IsPcGroup` to the union of `IsPcGroup` and `IsPcpGroup` (which now allows for infinite groups),
- introduced some GAP filters for that,
- added `Collector` objects in Oscar from which the corresponding GAP objects are created on demand,
- changed the defaults for `cyclic_group`, `dihedral_group` to return `PcGroup`s also in the infinite case (no longer `FPGroup`s),
- support `abelian_group(PcGroup, ...)` for infinite groups,
- moved some documentation,
- small additions to make this work:
  - `free_group` now admits an optional keyword argument s.t. one can create a group with elements in syllable representation,
  - the syntax `G([1 =>2, 2 => -3])` is now supported for free groups `G`, in order to construct elements of `G`,
  - `PcGroup` does no longer imply finiteness.


src/Groups/GAPGroups.jl
```diff
@@ -151,7 +151,7 @@ false
 """
 @gapattribute isfinite(G::GAPGroup) = GAP.Globals.IsFinite(G.X)::Bool
 
-Base.isfinite(G::PcGroup) = true
+# Base.isfinite(G::PcGroup) = true
```
```suggestion
```

src/Groups/GAPGroups.jl
```diff
@@ -1701,6 +1701,40 @@ function length(g::FPGroupElem)
 end
 
 
+"""
+    (G::FPGroup)(l::AbstractVector{Pair{T, S}}) where {T <: IntegerUnion, S <: IntegerUnion}
+
+Return the element `x` of `G` that is described by `l`
```
fingolfin: The font used by my browser for GitHub render `l` quite similar to `i1j`. Perhaps we could use a slightly more distinctive name here, e.g. `list` or `pairs` or so

src/Groups/GAPGroups.jl
```diff
@@ -1701,6 +1701,40 @@ function length(g::FPGroupElem)
...
+true
+```
+"""
+function (G::FPGroup)(l::AbstractVector{Pair{T, S}}) where {T <: IntegerUnion, S <: IntegerUnion}
+   n = ngens(G)
+   ll = S[]
+   for p in l
+     0 < p.first && p.first <= n || throw(ArgumentError("generator number is at most $n"))
+     push!(ll, p.first)
+     push!(ll, p.second)
```
fingolfin: `ll` is a `Vector{S}` but `first` is of type `T`. I guess in practice `T` and `S` both will be `Int` and this won't matter... But we could also sidestep all of this by changing `ll` into a `Vector{GAP.Obj}` and then doing this:
```suggestion
     push!(ll, GAP.Obj(p.first))
     push!(ll, GAP.Obj(p.second))
```
And then of course later we can just write `GapObj(ll)` w/o a need for recursive conversion.

src/Groups/group_constructors.jl
```diff
@@ -360,7 +385,7 @@ quaternion_group(n::IntegerUnion) = quaternion_group(PcGroup, n)
 function quaternion_group(::Type{T}, n::IntegerUnion) where T <: GAPGroup
   # FIXME: resolve naming: dicyclic vs (generalized) quaternion: only the
   # former should be for any n divisible by 4; the latter only for powers of 2.
-  # see also debate on the GAP side
+  # see also debate on the GAP side (gap-system/gap/issues/2725)
```
fingolfin: Good idea to add this reference, how about turning it into a full link?
```suggestion
  # see also debate on the GAP side (https://github.com/gap-system/gap/issues/2725)
```

src/Groups/pcgroup.jl
```diff
@@ -0,0 +1,345 @@
...
+# Create GAP filters that describe
+# - the union of `IsMultiplicativeElementWithInverseByPolycyclicCollector`
+#   and `IsPcpElement` and
+# - the union of `IsPcGroup` and `IsPcpGroup`.
+# This must be done at runtime.
+function __init_PcGroups()
+  GAP.evalstr("DeclareFilter(\"IsPcElementOrPcpElement\")");
+  GAP.evalstr("InstallTrueMethod(IsPcElementOrPcpElement, IsMultiplicativeElementWithInverseByPolycyclicCollector)");
+  GAP.evalstr("InstallTrueMethod(IsPcElementOrPcpElement, IsPcpElement)");
+  GAP.evalstr("BindGlobal(\"IsPcGroupOrPcpGroup\", IsGroup and CategoryCollections(IsPcElementOrPcpElement))");
```
fingolfin: Could also do tis (but do whatever you prefer, though)
```suggestion
  GAP.evalstr("""
      DeclareFilter("IsPcElementOrPcpElement");
      InstallTrueMethod(IsPcElementOrPcpElement, IsMultiplicativeElementWithInverseByPolycyclicCollector);
      InstallTrueMethod(IsPcElementOrPcpElement, IsPcpElement);
      BindGlobal("IsPcGroupOrPcpGroup", IsGroup and CategoryCollections(IsPcElementOrPcpElement));
      """)
```

src/Groups/pcgroup.jl
```diff
@@ -0,0 +1,345 @@
...
+"""
+function set_conjugate!(c::Collector{T}, j::Int, i::Int, rhs::Vector{Pair{Int, T}}) where T <: IntegerUnion
+  (0 < i && i <= c.ngens) || throw(ArgumentError("the collector has only $(c.ngens) generators not $i"))
+  i < j || throw(ArgumentError("only for i < j, but i = $i, j = $j"))
+  c.conjugates[i,j] = copy(rhs)
+
+  if isdefined(c, :X)
+    if GAP.Globals.IsSingleCollectorRep(c.X)
+      G = c.F
+      GAP.Globals.SetConjugate(c.X, j, i, G(rhs).X)
```
fingolfin, line 217: perhaps we should also wrap `SetCommutator`?

fingolfin (commented):
Some quick remarks. I'd like to have a closer look after I've slept ;-)

fingolfin:
Has a conflict with `src/Oscar.jl`, hence CI does not run

src/Groups/pcgroup.jl
```diff
@@ -0,0 +1,345 @@
...
+    end
+  end
+  GAP.Globals.UpdatePolycyclicCollector(cGAP)
+  
+  return cGAP
+end
+
+
+# Create the collector on the GAP side on demand
+function Base.getproperty(c::GAP_Collector, sym::Symbol)
```
fingolfin: We use `getproperty` like this for some of the other types mostly for legacy reasons, IMHO (namely because we did not provide a good way to access the underlying GAP object).

I think it would be preferable to change code which uses `collector.X` to get the underlying GAP object to use a helper method instead. The canonical idea would be to defined `GAPObj(collector)` for this role (i.e., what the code here does: compute `collector.X` if missing and then return it). Though we could also discuss other names, of course (`underlying_gap_object`?)

src/Groups/pcgroup.jl
```diff
@@ -0,0 +1,345 @@
...
+
+julia> gg = pc_group(c)
+<pc group of size 6 with 2 generators>
+
+julia> describe(gg)
+"S3"
+```
+"""
+function pc_group(c::GAP_Collector)
+  if 0 in c.relorders
```
fingolfin: Could we perhaps instead check the filters of `c.X` to decide which way we go? Then in the future we could offer a way to optionally also create a pcp group even if the collector describes a finite group (I am not sure we *want* that, but at least it would be an option. Actually I can think of one advantage pcp groups have there: e.g. a cyclic group of order 10^6 can be described with one generator as a pcp group, but not as a pc group (at least not if one wants that pc group the useful)

fingolfin (changes requested):
Overall this is a good change and I approve, thank you very much! I left a few minor comments. Also, the merge conflict needs to be resolved

fingolfin (approved):

PR #1825, opened 2022-12-14
Improvements for rational_solutions

thofma:
- Hook up msolve's `_rational_solutions`
- Add more tests over other fields


ederc (approved):
Look's good. Maybe we should put all the `rational_solutions` functionality into `solving.jl` instead of distributing it in `mpoly.jl` and `mpoly-graded.jl`?

thofma:
OK, will do tomorrow.

fingolfin (approved):

PR #1827, opened 2022-12-15
Collect deprecations

### LINKED ISSUE #1567: Collect deprecations in one place and document the process
I added `Deprecations.jl`, but did not want to move all of them in the otherwise unrelated PR. So here is just a reminder to add the following:
- [ ] Would be good to have them in one place (also recording during which release cycle the deprecations took place). 
- [ ] This should get an entry in the developer documentation.

lkastner:
Close #1567

### DIFF
--- a/docs/src/DeveloperDocumentation/styleguide.md
+++ b/docs/src/DeveloperDocumentation/styleguide.md
@@ -209,6 +209,12 @@ The comment about the version number is only necessary if you are the first one
 adding to `Deprecations.jl` after a release, otherwise please add to the
 existing block.
 
+!!! note
+    Please make sure to change to the new function everywhere in the existing
+    OSCAR code base. Even if you think, you were the only one using the
+    function, run a quick `grep` to make sure. When you are done,
+    `Deprecations.jl` should be the only place mentioning `old_function`.
+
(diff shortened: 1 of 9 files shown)


docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -209,6 +209,12 @@ The comment about the version number is only necessary if you are the first one
 adding to `Deprecations.jl` after a release, otherwise please add to the
 existing block.
 
+!!! note
+    Please make sure to change to the new function everywhere in the existing
+    OSCAR code base. Even if you think, you were the only one using the
+    function, run a quick `grep` to make sure. When you are done,
+    `Deprecations.jl` should be the only place mentioning `old_function`.
```
thofma: Maybe mention running julia with `--depwarn=yes` or even `--depwarn=error`.

thofma:
Thanks Lars!

thofma (approved):

lkastner:
Is there something I should write about aliases?

thofma:
In my ideal world we would not need any new aliases :) So from my part there is no need to mention aliases.

PR #1828, opened 2022-12-15
Add is total

tthsqe12:


src/Rings/orderings.jl
```diff
@@ -1394,6 +1394,16 @@ function is_mixed(ord::MonomialOrdering)
...
 end
 
+@doc Markdown.doc"""
+    is_total(ord::MonomialOrdering)
+
+Return `true` if `ord` is total ordering, `false` otherwise.
+"""
+function is_total(ord::MonomialOrdering)
+  m = canonical_matrix(ord)
+  return nrows(m) >= ncols(m)
```
afkafkafk13: ok, because _canonical_matrix does row reductions. Maybe add a comment on this for better readbility in the future.

afkafkafk13 (approved):
please add the comment on canonical_form doing row-reductions in the function is_total. Then this is good to go.

src/Rings/orderings.jl
thofma: I don't think the `canonical_matrix` is row-reduced.

fieker:
On Thu, Dec 15, 2022 at 09:41:05AM -0800, Tommy Hofmann wrote:
> @thofma commented on this pull request.
> 
> 
> 
> > @@ -1394,6 +1394,16 @@ function is_mixed(ord::MonomialOrdering)
>    return false
>  end
>  
> ***@***.*** Markdown.doc"""
> +    is_total(ord::MonomialOrdering)
> +
> +Return `true` if `ord` is total ordering, `false` otherwise.
> +"""
> +function is_total(ord::MonomialOrdering)
> +  m = canonical_matrix(ord)
> +  return nrows(m) >= ncols(m)
> 
> I don't think the `canonical_matrix` is row-reduced.
Well: yes/no: it is not in Gauss-echelon form, but it is unique and can
never have more rows than columns.
It should be mostly - it is unique after all. But not sure if positivity
needs to taken into account as well?

afkafkafk13:
The canoncial matrix is only a permutation of columns away from row echelon form, it does not have zero rows and the matrix can have less rows than columns, if the ordering is not a total ordering. Testing >=  boils down to testing for full rank, which is precisely what we need here.

fieker:
On Thu, Dec 15, 2022 at 11:07:11PM -0800, Anne Frühbis-Krüger wrote:
> > On Thu, Dec 15, 2022 at 09:41:05AM -0800, Tommy Hofmann wrote: @thofma commented on this pull request. > @@ -1394,6 +1394,16 @@ function is_mixed(ord::MonomialOrdering) return false end ***@***.*** Markdown.doc""" + is_total(ord::MonomialOrdering) + +Return `true` if `ord` is total ordering, `false` otherwise. +""" +function is_total(ord::MonomialOrdering) + m = canonical_matrix(ord) + return nrows(m) >= ncols(m) I don't think the `canonical_matrix` is row-reduced.
> > Well: yes/no: it is not in Gauss-echelon form, but it is unique and can never have more rows than columns. It should be mostly - it is unique after all. But not sure if positivity needs to taken into account as well?
> > […](#)
> > -- Reply to this email directly or view it on GitHub: [#1828 (comment)](https://github.com/oscar-system/Oscar.jl/pull/1828#discussion_r1049961014) You are receiving this because you are subscribed to this thread. Message ID: ***@***.***>
> 
> On Thu, Dec 15, 2022 at 09:41:05AM -0800, Tommy Hofmann wrote:
>  ***@***.***(https://github.com/thofma) commented on this pull request.
> 
> 
> 
>  > @@ -1394,6 +1394,16 @@ function is_mixed(ord::MonomialOrdering)
>     return false
>   end
> 
>  ***@***.*** Markdown.doc"""
>  +    is_total(ord::MonomialOrdering)
>  +
>  +Return `true` if `ord` is total ordering, `false` otherwise.
>  +"""
>  +function is_total(ord::MonomialOrdering)
>  +  m = canonical_matrix(ord)
>  +  return nrows(m) >= ncols(m)
> 
>  I don't think the `canonical_matrix` is row-reduced.
> Well: yes/no: it is not in Gauss-echelon form, but it is unique and can
> never have more rows than columns.
> It should be mostly - it is unique after all. But not sure if positivity
> needs to taken into account as well?
> […](https://github.com/oscar-system/Oscar.jl/pull/1828#)
> 
> 
>  --
>  Reply to this email directly or view it on GitHub:
>  [#1828 (comment)](https://github.com/oscar-system/Oscar.jl/pull/1828#discussion_r1049961014)
>  You are receiving this because you are subscribed to this thread.
> 
>  Message ID: ***@***.***>
> 
> The canoncial matrix is only a permutation of columns away from row
> echelon form, it does not have zero rows and the matrix can have less
> rows than columns, if the ordering is not a total ordering. Testing >=
> boils down to testing for full rank, which is precisely what we need
> here.
I wrote the function... but if thats what you need there: good

PR #1829, opened 2022-12-16
added some tests for `complement_class_reps`

ThomasBreuer:
(Somehow this code was forgotten in an earlier pull request.)

### DIFF
--- a/test/Groups/subgroups_and_cosets.jl
+++ b/test/Groups/subgroups_and_cosets.jl
@@ -327,6 +327,32 @@ end
    
 end
 
+@testset "Complement classes" begin
+   # solvable group
+   G = symmetric_group(4)
+   N = pcore(G, 2)[1]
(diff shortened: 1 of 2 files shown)


test/Groups/subgroups_and_cosets.jl
```diff
@@ -327,6 +327,32 @@ end
    
 end
 
+@testset "Complement classes" begin
+   # solvable group
+   G = symmetric_group(4)
+   N = pcore(G, 2)[1]
+   @test length(complement_class_reps(G, N)) == 1
```
fingolfin, line 334: This is also one of the functions we should rename to something like **`representative_of_XXXX** (possibly `complement_class_reps` can stay as an alias for convenience of GAP users who might guess this name?)

fingolfin (approved):

PR #1830, opened 2022-12-16
moved some code

ThomasBreuer:
from `experimental/GModule/Cohomology.jl` to `src/Groups/homomorphisms.jl`

### DIFF
--- a/src/Groups/homomorphisms.jl
+++ b/src/Groups/homomorphisms.jl
@@ -505,7 +505,7 @@ true
 """
 function isomorphism(::Type{T}, G::GAPGroup) where T <: Union{FPGroup, PcGroup, PermGroup}
    # Known isomorphisms are cached in the attribute `:isomorphisms`.
-   isos = get_attribute!(Dict{Type, Any}, G, :isomorphisms)
+   isos = get_attribute!(Dict{Type, Any}, G, :isomorphisms)::Dict{Type, Any}
    return get!(isos, T) do
      fun = _get_iso_function(T)
      f = fun(G.X)::GapObj
@@ -524,7 +524,7 @@ An exception is thrown if `G` is not abelian or not finite.
 """
 function isomorphism(::Type{GrpAbFinGen}, G::GAPGroup)
    # Known isomorphisms are cached in the attribute `:isomorphisms`.
-   isos = get_attribute!(Dict{Type, Any}, G, :isomorphisms)
+   isos = get_attribute!(Dict{Type, Any}, G, :isomorphisms)::Dict{Type, Any}
    return get!(isos, GrpAbFinGen) do
      is_abelian(G) || throw(ArgumentError("the group is not abelian"))
      isfinite(G) || throw(ArgumentError("the group is not finite"))
@@ -557,7 +557,7 @@ An exception is thrown if no such isomorphism exists or if `A` is not finite.
 """
 function isomorphism(::Type{T}, A::GrpAbFinGen) where T <: GAPGroup
    # Known isomorphisms are cached in the attribute `:isomorphisms`.
-   isos = get_attribute!(Dict{Type, Any}, A, :isomorphisms)
+   isos = get_attribute!(Dict{Type, Any}, A, :isomorphisms)::Dict{Type, Any}
    return get!(isos, T) do
      # find independent generators
      if is_diagonal(rels(A))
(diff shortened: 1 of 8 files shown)


src/Groups/homomorphisms.jl
```diff
@@ -616,6 +616,26 @@ function isomorphism(::Type{GrpAbFinGen}, A::GrpAbFinGen)
    end::AbstractAlgebra.Generic.IdentityMap{GrpAbFinGen}
 end
 
+# We need not find independent generators in order to create
+# a presentation of a fin. gen. abelian group.
+function isomorphism(::Type{FPGroup}, A::GrpAbFinGen)
+   # Known isomorphisms are cached in the attribute `:isomorphisms`.
+   isos = get_attribute!(Dict{Type, Any}, A, :isomorphisms)
```
fingolfin: Maybe this for type stability?
```suggestion
   isos = get_attribute!(Dict{Type, Any}, A, :isomorphisms)::Dict{Type, Any}
```

test/Groups/homomorphisms.jl
```diff
@@ -244,6 +244,21 @@ end
       end
    end
 
+   @testset "Infinite GrpAbFinGen to GAPGroup" begin
+      Agens = matrix(ZZ, 2, 2, [2, 3, 0, 0])
+      A = abelian_group(Agens)
+      for T in [FPGroup]
+         iso = @inferred isomorphism(T, A)
+         for x in gens(A)
+            for y in gens(A)
```
fingolfin: Maybe more julia-ish would be this?
```suggestion
         for x in gens(A), y in gens(A)
```

src/Groups/homomorphisms.jl
```diff
@@ -616,6 +616,26 @@ function isomorphism(::Type{GrpAbFinGen}, A::GrpAbFinGen)
...
+   return get!(isos, FPGroup) do
+      G = free_group(ngens(A))
+      R = rels(A)
+      s = vcat(elem_type(G)[i*j*inv(i)*inv(j) for i = gens(G) for j = gens(G) if i != j],
+           elem_type(G)[reduce(*, [gen(G, i)^R[j,i] for i=1:ngens(A) if !iszero(R[j,i])], init = one(G)) for j=1:nrows(R)])
```
fingolfin: Hmmm, I wonder if there was a reason @fieker did not use `prod`?

Also, the `elem_type(G)` shouldn't be necessary?

So perhaps
```suggestion
           [prod([gen(G, i)^R[j,i] for i=1:ngens(A) if !iszero(R[j,i])], init = one(G)) for j=1:nrows(R)])
```
or even
```suggestion
           [prod(gen(G, i)^R[j,i] for i=1:ngens(A) if !iszero(R[j,i]), init = one(G)) for j=1:nrows(R)])
```
And similar for `reduce(+, ...) -> sum` ? And the other `elem_type(G)` above?

fingolfin (approved):

src/Groups/homomorphisms.jl
thofma: I would guess the code predates 1.6. There was no `init` for `prod` before 1.6.

Why remove the `elem_type(G)`? The julia compiler can be very touchy (aka throwing errors) for comprehensions. I have spent days fixing type inference errors which were fixed by simply adding these types.

src/Groups/homomorphisms.jl
```diff
@@ -616,6 +616,26 @@ function isomorphism(::Type{GrpAbFinGen}, A::GrpAbFinGen)
...
+      F, mF = quo(G, s)
+      @assert is_finite(A) == is_finite(F)
```
fieker: maybe turn them into hassert? So that they are not always run? Is IsFinite not potentially a desaster for fp-groups - or is it clear that the presentations here are trivial?

src/Groups/homomorphisms.jl
ThomasBreuer: Good idea (also in the other instances of this code line)

test/Groups/homomorphisms.jl
ThomasBreuer: Good idea (also in other instances of nested loops in test files)

src/Groups/homomorphisms.jl
ThomasBreuer: O.k., I replace `reduce` by `prod` and `sum`, respectively.

Concerning `elem_type(G)`, this information is necessary if the vector in question can be empty. Are there other situations where `elem_type(G)` cannot be omitted?

src/Groups/homomorphisms.jl
ThomasBreuer: @fieker I had kept your assertion that `order(A) == order(F)` holds, but this statement can be evaluated only if the two groups are finite, thus I have restricted the check to finite groups, and then added the additional assertion for checking `is_finite(A) == is_finite(F)`.
This additional assertion is not critical here, already because even the group orders get computed when the original assertion gets executed. However, we can of course remove the assertions.

Would it make sense to document `@hassert` in order to advertise it for Oscar code?

src/Groups/homomorphisms.jl
thofma: I am not talking about the type being inferred as `Vector{Any}`. I am talking about the compiler crashing because things got too complicated. This is not hypothetical, e.g., https://github.com/thofma/Hecke.jl/search?q=compiler&type=commits.

PR #1840, opened 2022-12-23
Coherent sheaves vol. 1

HechtiDerLachs:
This is a cleanup of the first portion of functionality from #1809 . 

We introduce

 * Sheaves of modules
 * Hom-Sheaves
 * pushforwards along closed embeddings

As an example, we provide implementations of the twisting sheaves on projective schemes and computation of the (co-)tangent sheaves.


simonbrandhorst (approved):
Thank you for this PR. Can't say that I understand the details. But it looks reasonable. The code is well written. Tests pass. So it is certainly sufficient for experimental.

PR #1841, opened 2022-12-27
Fix some experimental GAP code

fingolfin:
- `IsSymmetric` is not provided by the GAP library (just some packages),
  so don't try to install methods for it
- `IsMonomial` is a dispatch for multiple other operations, and no methods
  should be installed for it; instead install one for `IsMonomialMatrix`


ThomasBreuer (approved):
Thanks, I agree:
- According to the GAP documentation, `IsSymmetric` is not defined for matrices (but it would make sense to define it for `IsMatrixObj` objects which have a chance to store the value).
- And `IsMonomialMatrix` is the right operation for this method installation.

PR #1843, opened 2022-12-28
Matroid: fix independent set for cycle matroids and a few minor improvements

benlorenz:
- fix `independent_sets` for matroids with UnitRange as ground set type
- improve function signatures for docs
- add a few bool return types

cc: @bschroter


fingolfin (approved):
LGTM

fingolfin:
@bschroter ping?

PR #1845, opened 2022-12-29
Semi-invariants / relative invariants

joschmitt:
Add basic functionality to compute generators of semi-invariants / relative invariants with respect to a character as module over the algebra generated by the primary invariants.

Some caveats:
* I don't need this; I implemented it because Ulrich Thiel was asked whether we have it. That means I don't really know what one wants to do with it. @fieker: I remember you asking me about something in this direction once. Can you use this?
* I barely found literature on this. Basically the only references are Sta79 for some theory and Gat96 for algorithms. I assume that much of this is folklore for the experts. In particular, I find it weird that the semi-invariants are computed over the primary invariants and not the whole invariant ring. However, I don't really have time to think about this in detail right now (or probably for the next months, they want me to write a thesis...). If anybody knows another reference, please say so!
* There is some potential for confusing names here: I followed Sta79 and call generators of the isotypic component of the polynomial ring with respect to a character semi-invariants or relative invariants. If the character is linear, so a representation by itself, this coincides with what is called equivariants. This has nothing to do with co(in)variants as far as I can see.

### DIFF
--- a/src/InvariantTheory/invariant_rings.jl
+++ b/src/InvariantTheory/invariant_rings.jl
@@ -612,17 +612,17 @@ function molien_series(S::PolyRing, I::InvRing, chi::Union{GAPGroupClassFunction
   end
 end
 
-function molien_series(I::InvRing)
-  if !isdefined(I, :molien_series)
+function molien_series(I::InvRing, chi::Union{GAPGroupClassFunction, Nothing} = nothing)
+  if chi === nothing
+    if !isdefined(I, :molien_series)
+      S, t = PolynomialRing(QQ, "t", cached = false)
+      I.molien_series = molien_series(S, I)
+    end
+    return I.molien_series
+  else
     S, t = PolynomialRing(QQ, "t", cached = false)
-    I.molien_series = molien_series(S, I)
+    return molien_series(S, I, chi)
   end
-  return I.molien_series
-end
-
-function molien_series(I::InvRing, chi::GAPGroupClassFunction)
-  S, t = PolynomialRing(QQ, "t", cached = false)
-  return molien_series(S, I, chi)
 end
 
 # There are some situations where one needs to know whether one can ask for the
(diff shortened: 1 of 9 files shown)


thofma:
Looks good from my side, although I did not check all the details.

joschmitt:
Since The Big Renaming is looming: can this be merged? Should I change something?

thofma:
If there are no objections, I will merge tomorrow.

src/InvariantTheory/invariant_rings.jl
```diff
@@ -246,6 +248,93 @@ function reynolds_operator(IR::InvRing, f::MPolyElem)
...
+Reynolds operator projecting onto the isotypic component of the polynomial ring
+with respect to `chi`, that is, the semi-invariants (or relative invariants)
+with respect to `chi`, see [Sta79](@cite).
+It is assumed that `chi` is an irreducible character.
+
+In case `chi` is a linear character, the returned polynomial, say `h`, fulfils
+`h^g = chi(g)h` for all `g` in `group(IR)` (possibly `h` is zero).
+
+!!! note
+    It is implicitly assumed that `coefficient_ring(IR)` contains all character values of `chi`.
```
simonbrandhorst: Is that an expensive check? If it is not too expensive, I would rather add a `check=true` keyword  so that 
the method is safe by default, and if anyone needs full speed an knows what they do, they can disable the check.

joschmitt: There is no actual check. I have to cast the character values (which are `QQAbElem`s) in the coefficient field. This either works or there is an error.

simonbrandhorst: Okay. Then the behavior is fine to me.  I interpreted 

"It is implicitly assumed that `coefficient_ring(IR)` contains all character values of `chi`."

as: if the implicit assumption is not true, than anything may happen including false result. 
So I would be more concrete and reformulate this to:  
"If the `coefficient_ring(IR)` does not contain all character values of `chi`, an error is raised."

joschmitt: Ah, you're right! I changed the wording accordingly. Thank you!

simonbrandhorst (approved):

PR #1846, opened 2023-01-02
Improve coverings, round I

HechtiDerLachs:
In order to better model morphisms of covered schemes and vector bundles/divisors on covered schemes, we need to work on the refinements of coverings. This is a first bit, starting out by introducing `SimplifiedSpec` and `LazyGlueing`. Both will be important in the implementation of resolution of singularities. 

The idea of `LazyGlueing` is to postpone harder computation to the point when it is actually needed. This is achieved by providing a method for the computation and a container for the required data. This way, we should be able to save a lot of space and computation time, especially in blowup constructions.


docs/src/AlgebraicGeometry/Schemes/CoveredSchemes.md
```diff
@@ -53,5 +52,27 @@ An `AbsCoveredScheme` may have different properties such as
     is_smooth(X::AbsCoveredScheme)
 ```
 
+## The modeling of covered schemes and their expected behaviour 
+
+Any `AbsCoveredScheme` may posess several `Covering`s. This is necessary for 
+several reasons; for instance, a morphism $f : X \to Y$ between `AbsCoveredScheme`s 
+will in general only be given on affine patches on a refinement of the `default_covering` of `X`.
```
afkafkafk13, line 59: Do you want to say: 'on the affine patches of some appropriate refinement of the ...'?

docs/src/AlgebraicGeometry/Schemes/CoveredSchemes.md
```diff
@@ -53,5 +52,27 @@ An `AbsCoveredScheme` may have different properties such as
...
+    coverings(X::AbsCoveredScheme)
+```
+Every `AbsCoveredScheme` $X$ has to be modeled using one original `default_covering` $C$, simply 
+to gather the data necessary to fully describe $X$. The `affine_charts` of $X$ return the 
+patches of this covering. For any refinement $D < C$, we require the following to hold: 
+Every element $U$ of the `affine_charts` of $D$ is either 
+
+  * directly an element of the `affine_charts` of $C$;
+  * a `PrincipalOpenSubset` with some ancestor in the `affine_charts` of $C$; 
+  * a `SimplifiedSpec` with some original in the `affine_charts` of $C$.
```
afkafkafk13, line 71: You do not allow SimplifiedSpec of a PrincipalOpenSubset, correct?

experimental/Schemes/LazyGlueing.jl
```diff
@@ -0,0 +1,58 @@
...
+      compute_function::Function, GD::GlueingDataType
+    ) where {GlueingDataType}
+    return new{typeof(X), typeof(Y), GlueingDataType}(X, Y, GD, compute_function)
+  end
+end
+
+patches(G::LazyGlueing) = (G.X, G.Y)
+
+# The underlying_glueing triggers the computation of the glueing on request.
+function underlying_glueing(G::LazyGlueing)
```
afkafkafk13, line 28: stupid question: Is there a general name convention in OSCAR for functions turning lazy objects into explicit objects? If so, use the respective name.

experimental/Schemes/Sheaves.jl
```diff
@@ -164,6 +164,174 @@ end
...
+  ### Checks for open containment.
+  #
+  # We allow the following cases:
+  #
+  #  * U::PrincipalOpenSubset with one ancestor W in the basic charts of X
+  #  * U::SimplifiedSpec with one ancestor W in the basic charts of X
+  #  * U::PrincipalOpenSubset ⊂ V::PrincipalOpenSubset with ambient_scheme(U) === ambient_scheme(V) in the basic charts of X
+  #  * U::PrincipalOpenSubset ⊂ V::PrincipalOpenSubset with ambient_scheme(U) != ambient_scheme(V) both in the basic charts of X
+  #    and U and V contained in the glueing domains of their ambient schemes
+  #  * U::AbsSpec ⊂ U::AbsSpec in the basic charts of X
```
afkafkafk13, line 177: Typo? Are both of them U in this line?

experimental/Schemes/Sheaves.jl
```diff
@@ -164,6 +164,174 @@ end
...
+    Udirect = codomain(inc_U_flat)
+    Vdirect = codomain(inc_V_flat)
+   # some_ancestor(W->any(WW->(WW===W), affine_charts(X)), U) || return false
+   # some_ancestor(W->any(WW->(WW===W), affine_charts(X)), V) || return false
+   # incU, dU = _find_chart(U, default_covering(X))
+   # incV, dU = _find_chart(V, default_covering(X))
+   # A = codomain(incU)
+   # B = codomain(incV)
+   # Udirect = (ambient_scheme(U) === A ? U : PrincipalOpenSubset(codomain(incU), dU))
+   # Vdirect = (ambient_scheme(V) === B ? V : PrincipalOpenSubset(codomain(incV), dV))
```
afkafkafk13, line 198: This block seems obsolete. Is there a reason to leave it in this inactive state?

experimental/Schemes/Sheaves.jl
```diff
@@ -164,6 +164,174 @@ end
...
+  function is_open_func(
+      U::Union{<:PrincipalOpenSubset, <:SimplifiedSpec}, 
+      Y::AbsCoveredScheme
+    )
+    return Y === X && some_ancestor(W->(W in affine_charts(X)), U)
+  end
+  function is_open_func(U::AbsSpec, Y::AbsCoveredScheme)
+    return Y === X && some_ancestor(W->any(WW->(WW===W), affine_charts(X)), U)
+  end
+  function is_open_func(Z::AbsCoveredScheme, Y::AbsCoveredScheme)
```
afkafkafk13, line 220: Put a comment on this case, to make clear why it is considered at all. 
Also mention that this has nothing to do with the case that one is contained in the other by means of containment of atlases.

experimental/Schemes/Sheaves.jl
```diff
@@ -164,6 +164,174 @@ end
...
+      gU = preimage(g, Udirect)
+      is_subset(gU, Vdirect) || return false
+    end
+    return true
+  end
+  function is_open_func(
+      U::Union{<:PrincipalOpenSubset, <:SimplifiedSpec}, 
+      Y::AbsCoveredScheme
+    )
+    return Y === X && some_ancestor(W->(W in affine_charts(X)), U)
```
afkafkafk13, line 215: comparison with == or ===?

experimental/Schemes/Sheaves.jl
```diff
@@ -164,6 +164,174 @@ end
...
+    inc_U_flat = _flatten_open_subscheme(U, default_covering(X))
+    A = ambient_scheme(codomain(inc_U_flat))
+    Udirect = codomain(inc_U_flat)
+    W = ambient_scheme(Udirect)
+    # some_ancestor(W->(W===V), U) && return true
+    # incU, dU = _find_chart(U, default_covering(X))
+    # W = codomain(incU)
+    haskey(glueings(default_covering(X)), (W, V)) || return false # In this case, they are not glued
+    G = default_covering(X)[W, V]
+    #Udirect = (ambient_scheme(U) === W ? U : PrincipalOpenSubset(W, dU))
```
afkafkafk13, line 243: Is there a reason not to clean up the deprecated lines?

simonbrandhorst (approved):

experimental/Schemes/Types.jl
```diff
@@ -513,30 +434,59 @@ identifications given by the glueings in the `default_covering`.
...
         return hom(OV, OU, rho_func.(gens(OV)), check=false)
       else
-        G = default_covering(X)[ambient_scheme(V), U]
+        G = default_covering(X)[W, U]
         W1, W2 = glueing_domains(G)
         f, g = glueing_morphisms(G)
+        g_res = restrict(g, U, V_direct)
+        inc_res = restrict(incV, V, V_direct, check=false)
+        return pullback(compose(g_res, inverse(inc_res)))
+        ### deprecated code below
```
afkafkafk13, line 443: If it is deprecated, why don't you put in in a comment? Is it still in use somewhere?

experimental/Schemes/SimplifiedSpec.jl
```diff
@@ -0,0 +1,159 @@
...
+  return compose(f, h), d
+end
+
+#=
+# This follows U in its ancestor tree up to the point 
+# where a patch W in C is found. Then it recreates U as a 
+# PrincipalOpenSubset UU of W and returns the identification 
+# with UU.
+=#
+function _flatten_open_subscheme(
```
afkafkafk13, line 96: find a better name for this in the future. Flatten only makes sense, if you have the tree structure in mind, which we cannot expect from a user.

afkafkafk13 (changes requested):
Please resolve the issue with the comparison == and === in line 215 Sheaves.jl (see my comment).

Please add tests for lazy glueing.

No need to worry about the function names right now. This should be discussed at some point in the future, before people start using the.

docs/src/AlgebraicGeometry/Schemes/CoveredSchemes.md
HechtiDerLachs: No, I do. The `SimplifiedSpec` does not care about the nature of its origin.

experimental/Schemes/LazyGlueing.jl
HechtiDerLachs: I have not encountered other lazy objects in Oscar, yet. In either case, this function needs to be triggered by `underlying_glueing`, either way. I don't change this, at least not for now.

experimental/Schemes/Sheaves.jl
HechtiDerLachs: I'm afraid I do not fully get what you have in mind here. You are speaking about line 220?

experimental/Schemes/Sheaves.jl
HechtiDerLachs: No. I had just kept them there for comparison while debugging.

experimental/Schemes/SimplifiedSpec.jl
HechtiDerLachs: But this is not a user-facing function. We (the programmers) do have a tree-structure in mind and we know about it.

HechtiDerLachs:
Tests for lazy glueings are inherited from the existing tests: I changed some tests so that the lazy glueings are used internally and hence they should be tested on the fly.

experimental/Schemes/Sheaves.jl
afkafkafk13: I am talking about the function beginning in 220 and only doing the Check X===Y===Z.
So is_open_func is designed to only return true for two AbsCoveredSchemes, if the AbsCoveredSchemes are equal. This is by choice and this is fine, but it could happen that the user expects that it also works if Y and Z are subschemes of X by selection of subsets of the patches -- if e.g. X is the disjoint union of U_1,...,U_4,  and the other two are the disjoint unions of U_1,U_2 and of U_1,...,U_3 respectively.

experimental/Schemes/LazyGlueing.jl
jankoboehm: Free resolutions with homological degree limit are a kind of similar situation, calling _extend_free_resolution.

PR #1851, opened 2023-01-04
milp serialization

antonydellavecchia:
(De)Serialization of MILPs, with test

### DIFF
--- a/src/Serialization/PolyhedralGeometry.jl
+++ b/src/Serialization/PolyhedralGeometry.jl
@@ -46,6 +46,41 @@ function load_internal(s::DeserializerState, ::Type{LinearProgram{T}}, dict::Dic
     return LinearProgram{T}(fr, lp, Symbol(conv))
 end
 
+##############################################################################
+@registerSerializationType(MixedIntegerLinearProgram{fmpq})
+
+function save_internal(s::SerializerState, milp::MixedIntegerLinearProgram)
+    milp_coeffs = milp.polymake_milp.LINEAR_OBJECTIVE
+    coeffs_serialized = Polymake.call_function(Symbol("Core::Serializer"), :serialize, milp_coeffs)
+    coeffs_jsonstr = Polymake.call_function(:common, :encode_json, coeffs_serialized)
+
+    return Dict(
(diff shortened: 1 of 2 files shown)


src/Serialization/PolyhedralGeometry.jl
```diff
@@ -46,6 +46,41 @@ function load_internal(s::DeserializerState, ::Type{LinearProgram{T}}, dict::Dic
...
+function save_internal(s::SerializerState, milp::MixedIntegerLinearProgram)
+    milp_coeffs = milp.polymake_milp.LINEAR_OBJECTIVE
+    coeffs_serialized = Polymake.call_function(Symbol("Core::Serializer"), :serialize, milp_coeffs)
+    coeffs_jsonstr = Polymake.call_function(:common, :encode_json, coeffs_serialized)
+
+    return Dict(
+        :feasible_region => save_type_dispatch(s, milp.feasible_region),
+        :convention => milp.convention,
+        :milp_coeffs => JSON.parse(coeffs_jsonstr)
+    )
```
lkastner: I think you need another address parameter, the integer variables. There can be several milps with the same coefficient vector, but different integer variables.

src/Serialization/PolyhedralGeometry.jl
```diff
@@ -46,6 +46,41 @@ function load_internal(s::DeserializerState, ::Type{LinearProgram{T}}, dict::Dic
...
+        :common,
+        :deserialize_json_string,
+        json(dict[:milp_coeffs])
+    )
+    all = Polymake._lookup_multi(pm_object(fr), "MILP")
+    index = 0
+    for i in 1:length(all)
+        if all[i].LINEAR_OBJECTIVE == milp_coeffs
+            index = i
+            break
```
lkastner: Same here then. And then the check should be
```
if all[i].LINEAR_OBJECTIVE == milp_coeffs && all[i].INTEGER_VARIABLES == intvar
...
```

lkastner (changes requested):

lkastner (approved):
Thank you.

PR #1854, opened 2023-01-05
Add Cohen-Macaulay test for affine algebras

wdecker:


afkafkafk13 (approved):
looks good.
(Remark: Tests were already present before the function was finished. No adjustments to tests necessary.)

PR #1855, opened 2023-01-05
Improve coverings, round II

HechtiDerLachs:
Continuation of #1846 .

### DIFF
--- a/experimental/Schemes/CoherentSheaves.jl
+++ b/experimental/Schemes/CoherentSheaves.jl
@@ -184,7 +184,7 @@ identifications given by the glueings in the `default_covering`.
         UU, _ = glueing_domains(default_covering(X)[U, V])
         psi = OOX(UU, U) # Needs to exist by the checks of is_open_func, even though 
         # in general UU ⊂ U!
-        return hom(MV, MU, [sum([psi(A[i, j]) * MU[j] for j in 1:ngens(MU)]) for i in 1:ngens(MV)], rho)
+        return hom(MV, MU, [sum([psi(A[i, j]) * MU[j] for j in 1:ngens(MU)]) for i in 1:ngens(MV), init=zero(MU)], rho)
       else
         error("invalid input")
       end
@@ -217,7 +217,7 @@ identifications given by the glueings in the `default_covering`.
       # respect to the generators of F(W), so we have to map them manually down.
       # The call to F(W, U) will be handled by the above if-clauses.
       return hom(F(V), F(U), 
-                 [sum([OOX(WW, U)(A[i, j])*F(W, U)(F(W)[j]) for j in 1:ngens(F(W))]) 
+                 [sum([OOX(WW, U)(A[i, j])*F(W, U)(F(W)[j]) for j in 1:ngens(F(W)), init=zero(F(U))]) 
                   for i in 1:ngens(F(V))], 
                  OOX(V, U)
                 )
@@ -250,7 +250,7 @@ identifications given by the glueings in the `default_covering`.
       # respect to the generators of F(W), so we have to map them manually down.
       # The call to F(W, U) will be handled by the above if-clauses.
       return hom(F(V), F(U), 
-                 [sum([OOX(WW, U)(A[i, j])*F(W, U)(F(W)[j]) for j in 1:ngens(F(W))]) 
+                 [sum([OOX(WW, U)(A[i, j])*F(W, U)(F(W)[j]) for j in 1:ngens(F(W)), init=zero(F(U))]) 
                   for i in 1:ngens(F(V))], 
                  OOX(V, U)
                 )
@@ -717,7 +717,7 @@ end
       images = elem_type(MU)[]
       for phi in gens(MV)
         phi_map = element_to_homomorphism(phi)
-        images_f = [sum([B[i][j]*cod_res(phi_map(f[j])) for j in 1:length(f)]) for i in 1:length(B)]
+        images_f = [sum([B[i][j]*cod_res(phi_map(f[j])) for j in 1:length(f), init=zero(G(U))]) for i in 1:length(B)]
         psi = hom(F(U), G(U), images_f)
         push!(images, homomorphism_to_element(MU, psi))
       end
--- a/experimental/Schemes/CoherentSheaves.jl
+++ b/experimental/Schemes/CoherentSheaves.jl
@@ -82,18 +82,39 @@ function _is_open_for_modules(X::AbsCoveredScheme)
     return X === Y === Z
   end
   function is_open_func(U::AbsSpec, V::AbsSpec)
-    U in affine_charts(X) || return false
-    V in affine_charts(X) || return false
-    G = affine_charts(X)[U, V]
+    any(x->x===U, affine_charts(X)) || return false
+    any(x->x===V, affine_charts(X)) || return false
+    G = default_covering(X)[U, V]
     return issubset(U, glueing_domains(G)[1])
   end
-  function is_open_func(U::PrincipalOpenSubset, V::AbsSpec)
-    V in affine_charts(X) || return false
-    ambient_scheme(U) === V && return true
-    W = ambient_scheme(U)
-    W in affine_charts(X) || return false
+  function is_open_func(
+      U::AbsSpec,
+      V::Union{<:PrincipalOpenSubset, <:SimplifiedSpec}
+    )
+    issubset(U, V) && return true
+    any(x->x===U, affine_charts(X)) || return false
+    inc_V_flat = _flatten_open_subscheme(V, default_covering(X))
+    A = ambient_scheme(codomain(inc_V_flat))
+    Vdirect = codomain(inc_V_flat)
+    W = ambient_scheme(Vdirect)
+    haskey(glueings(default_covering(X)), (W, U)) || return false # In this case, they are not glued
+    G = default_covering(X)[W, U]
+    f, g = glueing_morphisms(G)
+    pre_V = preimage(g, V)
+    return is_subset(U, pre_V)
+  end
+  function is_open_func(
+      U::Union{<:PrincipalOpenSubset, <:SimplifiedSpec}, 
+      V::AbsSpec
+    )
+    any(x->x===V, affine_charts(X)) || return false
+    inc_U_flat = _flatten_open_subscheme(U, default_covering(X))
+    A = ambient_scheme(codomain(inc_U_flat))
+    Udirect = codomain(inc_U_flat)
+    W = ambient_scheme(Udirect)
+    haskey(glueings(default_covering(X)), (W, V)) || return false # In this case, they are not glued
     G = default_covering(X)[W, V]
-    return is_subset(U, glueing_domains(G)[1])
+    return is_subset(Udirect, glueing_domains(G)[1])
   end
   return is_open_func
 end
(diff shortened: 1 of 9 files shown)


experimental/Schemes/Types.jl
```diff
@@ -673,12 +616,21 @@ identifications given by the glueings in the `default_covering`.
...
+      haskey(ID, U) && return ID[U]
       V = ambient_scheme(U)
       IV = F(V)::Ideal
       rho = OOX(V, U)
       IU = ideal(OO(U), rho.(gens(IV)))
       return IU
     end
+    function production_func(F::AbsPreSheaf, U::SimplifiedSpec)
+      haskey(ID, U) && return ID[U]
+      V = original(U)
```
afkafkafk13, line 630: Original goes all the way up to the original covering of X, right? Then is is fine.

afkafkafk13 (approved):

experimental/Schemes/Types.jl
HechtiDerLachs: No. It only goes up one step.

afkafkafk13: So we call F(V) one step upstairs, which then calls production_func with the appropriate input and we recurse our way up?

HechtiDerLachs: Yes.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -162,31 +163,373 @@ identifications given by the glueings in the `default_covering`.
...
-      while VV isa PrincipalOpenSubset
-          VV = ambient_scheme(VV)
+      if any(W->(W === U), affine_charts(X)) && any(W->(W === V), affine_charts(X))
+        MV = F(V)
+        MU = F(U)
+        A = MG[(V, U)] # The transition matrix
+        UU, _ = glueing_domains(default_covering(X)[U, V])
+        psi = OOX(UU, U) # Needs to exist by the checks of is_open_func, even though 
+        # in general UU ⊂ U!
+        return hom(MV, MU, [sum([psi(A[i, j]) * MU[j] for j in 1:ngens(MU)]) for i in 1:ngens(MV)], rho)
```
simonbrandhorst: This line will throw an error if `ngens(MU)` is zero, e.g. for the zero module. And maybe it is not type stable.
The reason is that the empty sum is not well defined, because julia does not know which zero element to choose.
You can fix this by providing the keyword  `init=zero(MU)` .

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -162,31 +163,373 @@ identifications given by the glueings in the `default_covering`.
...
+    end
+
+    function restriction_func(F::AbsPreSheaf, V::AbsSpec, U::PrincipalOpenSubset)
+      # If V was not an affine_chart of X, some other function would have 
+      # been triggered. 
+
+      # First the easy case: Inheritance from an ancestor in the tree.
+      if ambient_scheme(U) === V
+        # If the restriction was more complicated than what follows, then 
+        # it would have been cached earlier and this call would not have happened
```
simonbrandhorst, line 221: Relying on caching seems brittle to me. It is necessary?

HechtiDerLachs: Why brittle?

First of all, it is default-stable in the sense that it doesn't produce wrong results in case nothing was cached. 

Second, the call to the `production_func` and `restriction_func` follows a clear protocol (not yet publicly documented, admittedly) which says that things are first looked up in the caches and only then produced if necessary. 

The idea is to use this with ideal sheaves or coherent sheaves which happen to be locally free: We want to ask for things like `locally_trivializing_cover` or `local_complete_intersection_cover`. These have to be implemented in a way so that 
  * the trivializing patches are `PrincipalOpenSubset`s or `SimplifiedSpec`s (or position themselves by other means in the tree-structure) with the results in the desired form on these patches
  * the restriction maps from their parent nodes are also in the prescribed form and, necessarily, cached.
 
Then, for every other patch `W` which sits below a trivializing patch `U`, we expect to also find restrictions which take into account the intermediate trivializations. That is also the reason why we go up recursively along the tree structure for the productions and restrictions (see @afkafkafk13 's comment above) and not straight to the top. 

Does this rule out your concerns?

simonbrandhorst: Brittle in the sense that a change of the caching behavior might affect this function and this could be unexpected. 
But if you say that it is stable by default, I am okay. 
For larger computations one might want to clear the caches from time to time / or allow them to be collected / or disable caching (where feasible) to save memory.

HechtiDerLachs: Changing of the caching behaviour will certainly affect this function and whoever wants to do such things in obligated to also clean up here. 

Turning off caching is not an option here: We need the stability to have the very same object in memory whenever we call `M(U)` for a fixed `U` and a coherent sheaf `M` in order to maintain the element-parent structure in Oscar. 

I put a lot of effort into making the caching garbage collectable in the future. However, that still relies on introducing the `WeakKeyIdDict`s in Oscar at some point: https://github.com/JuliaCollections/DataStructures.jl/pull/402. I hope this will help with your concerns in this regard.

simonbrandhorst (approved):

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -966,7 +966,172 @@ function _pushforward(f::Hecke.Map{<:Ring, <:Ring}, I::Ideal, M::SubQuo)
...
+      for k in j+1:ncols(Ares)
+        push!(img_gens, F[k-1])
+      end
+      res = hom(MU, F, img_gens, OOX(U, V))
+
+      # Since we are messing with the internals of the sheaf, we need 
+      # to leave everything clean. This includes manual caching.
+      add_incoming_restriction!(M, U, F, res)
+      object_cache(M)[V] = F
+      set_attribute!(F, :_presentation_matrix, Asub)
```
HechtiDerLachs, line 1159: @simonbrandhorst : This is an example of where one manually constructs restrictions and is responsible for their cachings.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -622,7 +987,199 @@ function _pushforward(f::Hecke.Map{<:Ring, <:Ring}, I::Ideal, M::SubQuo)
...
+        add_incoming_restriction!(domain(M), U, dom_UV, dom_res)
+        add_incoming_restriction!(domain(M), W, dom_UV, 
+                                  compose(domain(M)(W, U), dom_res))
+        object_cache(domain(M))[UV] = dom_UV
+
+        cod_UV, cod_res = change_base_ring(OOX(V, UV), codomain(M)(V))
+        add_incoming_restriction!(codomain(M), V, cod_UV, cod_res)
+        add_incoming_restriction!(codomain(M), W, dom_UV, 
+                                  compose(codomain(M)(W, V), cod_res))
+        object_cache(codomain(M))[UV] = cod_UV
```
HechtiDerLachs: @simonbrandhorst : This is another instance of where one needs to manually construct restrictions and, hence, also needs to take care of the caching.

HechtiDerLachs:
Sorry about this morning's work almost doubling the size of the PR. Maybe, you want to have a look again. I will not make any further changes now, except repairing the tests if necessary.

HechtiDerLachs:
One loose end that we need to pick up here is the following. When creating the refinements, I do not yet fill out the glueings. The function for that has an empty body. We could do this with `LazyGlueing`s, but @afkafkafk13 suggested the other day to introduce another type like `ImplicitGlueing` for glueings within one chart. I don't really see that this is the way to go, because it produces a lot of overhead. Whether we need/want such things will depend on the use-cases later on. Are we happy with the tree structure and the glueings in the default_covering? Or do we really want to access and use glueings in refinements? 

For the time being, the refinement coverings just skip the glueings and that doesn't break anything, yet. But we should think about these issues.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -622,7 +987,199 @@ function _pushforward(f::Hecke.Map{<:Ring, <:Ring}, I::Ideal, M::SubQuo)
...
+    end
+  end
+  return return_patches
+end
+
+@attr MatrixElem function _presentation_matrix(M::ModuleFP)
+  return matrix(map(presentation(M), 1))
+end
+
+function fill_with_lazy_glueings!(C::Covering, X::AbsCoveredScheme)
```
HechtiDerLachs, line 1196: This is the function I was talking about.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -162,31 +184,373 @@ identifications given by the glueings in the `default_covering`.
...
+        c = coordinates(w)
+        w = sum(sum(OOX(V, U)(c[i])*OOX(WW, U)(A[i, j])*gens_U[j] 
+                    for i in 1:length(gens_V)) 
+                for j in 1:length(gens_U)
+               )
+        push!(img_gens, w)
+      end
+      return hom(F(V), F(U), img_gens, OOX(V, U))
+    end
+    function restriction_func(F::AbsPreSheaf, V::SimplifiedSpec, U::PrincipalOpenSubset)
```
simonbrandhorst, line 429: This seems largely untested. It would be good to reach the code once. If generating an example is too complicated at this point, it can also be done later.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -622,7 +987,199 @@ function _pushforward(f::Hecke.Map{<:Ring, <:Ring}, I::Ideal, M::SubQuo)
...
+        #place in the tree structure.
+        incU = _flatten_open_subscheme(U, W)
+        incV = _flatten_open_subscheme(V, W)
+        UV = intersect(codomain(incU), codomain(incV))
+        push!(patch_list, UV)
+      end
+    end
+  end
+  return patch_list
+end
```
simonbrandhorst: please add a test for this function. And perhaps a few comments for docu?

experimental/Schemes/SimplifiedSpec.jl
```diff
@@ -156,4 +171,155 @@ function _flatten_open_subscheme(
...
+  new_iso_inv = compose(inverse(ident), inverse(iso))
+  set_attribute!(new_iso, :inverse, new_iso_inv)
+  set_attribute!(new_iso_inv, :inverse, new_iso)
+  if W === P
+    return new_iso
+  end
+  return _flatten_open_subscheme(W, P, iso=new_iso)
+end
+
+function _flatten_open_subscheme(
```
simonbrandhorst, line 203: Needs a test.

experimental/Schemes/CoherentSheaves.jl
HechtiDerLachs: I agree that this should be tested. But, as you probably saw already, this still has a lot of code duplication, so it is not unlikely that 'the same' code was tested elsewhere. I want to clean this up eventually, but that requires some decisions on internals that I did not want to make yet. If possible, I would like to postpone writing the tests until things settle a bit more.

experimental/Schemes/CoherentSheaves.jl
HechtiDerLachs: Looking at it now: This function is deprecated. I included its contents directly, so it will be removed.

PR #1857, opened 2023-01-09
more homological algebra functions

wdecker:


src/Modules/homological-algebra.jl
```diff
@@ -1,4 +1,498 @@
...
+function fitting_ideal(F:: FreeMod, i::Int)
+ R = base_ring(F)
+ if !(coefficient_ring(R) isa AbstractAlgebra.Field)
+   throw(ArgumentError("The coefficient ring must be a field."))
+ end
+ if i < rank(F) return ideal(R, [zero(R)]) end
+ return R
+end
+
+function fitting_ideal(M::SubQuo, i::Int)
```
HechtiDerLachs:
```suggestion
function fitting_ideal(M::SubQuo{T}, i::Int) where {T<:MPolyElem}
```
Since the internals are using Singular functionality, this should probably be restricted to the polynomial case; just like the other methods below.

src/Modules/homological-algebra.jl
```diff
@@ -1,4 +1,498 @@
...
+
+function is_flat(F:: FreeMod)
+ R = base_ring(F)
+ if !(coefficient_ring(R) isa AbstractAlgebra.Field)
+   throw(ArgumentError("The coefficient ring must be a field."))
+ end
+ return true
+end
+
+function is_flat(M::SubQuo)
```
HechtiDerLachs:
```suggestion
function is_flat(M::SubQuo{T}) where {T<:MPolyElem}
```
Same as above.

src/Modules/homological-algebra.jl
```diff
@@ -1,4 +1,498 @@
...
+
+julia> non_flat_locus(M)
+ideal(x^3 - y^2)
+```
+"""
+function non_flat_locus(M::ModuleFP)
+ error("not implemented for the given type of module.")
+end
+
+function non_flat_locus(F:: FreeMod)
```
HechtiDerLachs:
```suggestion
function non_flat_locus(F::FreeMod)
```

src/Modules/homological-algebra.jl
```diff
@@ -1,4 +1,498 @@
...
+ R = base_ring(F)
+ if !(coefficient_ring(R) isa AbstractAlgebra.Field)
+   throw(ArgumentError("The coefficient ring must be a field."))
+ end
+ return ideal(R, [one(R)])
+end
+
+function non_flat_locus(M::SubQuo)
```
HechtiDerLachs:
```suggestion
function non_flat_locus(M::SubQuo{T}) where {T<:MPolyElem}
```

src/Modules/homological-algebra.jl
```diff
@@ -1,4 +1,498 @@
...
+   throw(ArgumentError("The coefficient ring must be a field."))
+ end
+ @assert parent(V[1]) == R
+ @assert all(x->parent(x) == R, V)
+ I = ideal(R, V)
+ singular_assure(I)
+ MX = singular_module(F)
+ SG = Singular.Module(base_ring(MX), MX(zero(F)))
+ res = Singular.LibHomolog.isReg(I.gens.S, SG)
+ if res == 1 return true end
```
HechtiDerLachs, line 259:
```suggestion
one(R) in res && return true
```
Just another short way to write the same thing; not necessarily better but closer to what was once suggested to me to avoid the if-clauses.

HechtiDerLachs (approved):
I left some comments with a view towards the problems encountered earlier with non-generic code for modules. 

This looks good to me, but I have one related question: Do the singular routines also digest local orderings? Then it would be nice to also implement all these functions for local rings which are localizations of polynomial rings at rational points.

wdecker:
@HechtiDerLachs 

This looks good to me, but I have one related question: Do the singular routines also digest local orderings? Then it would be nice to also implement all these functions for local rings which are localizations of polynomial rings at rational points.

Yes, I actually wanted to discuss this with you. Let us talk at some point. Also quotient rings may come into play.

PR #1858, opened 2023-01-10
Make struct CycleType immutable

fingolfin:


fieker (approved):

ThomasBreuer (approved):
just for curiosity:
When accessing the internals of a `Pair` object , is it better to call `x[1]` or `x.first`?

PR #1859, opened 2023-01-10
Improve primitive_by_shape and an_sn_by_shape

fingolfin:
Add docstrings, explain why they are correct, add tests

### DIFF
--- a/experimental/GaloisGrp/GaloisGrp.jl
+++ b/experimental/GaloisGrp/GaloisGrp.jl
@@ -1783,35 +1783,91 @@ function order_from_shape(ct::Set{CycleType}, n)
 end
 
 
-#https://mathoverflow.net/questions/286316/recognition-of-symmetric-groups-in-gap
-#show primitivity by finding a permutation which has a p cycle for n/2<p<n−1
+@doc Markdown.doc"""
+  primitive_by_shape(ct::Set{CycleType}, n::Int)
+
+Return `true` if a transitive group $G \leq Sym(n)$ containing permutations
+matching the cycle types in `ct` must be primitive, otherwise return `false`.
+"""
 function primitive_by_shape(ct::Set{CycleType}, n::Int)
-  for p = PrimesSet(div(n, 2)+1, n-2)
-    if any(y->any(x->x[1] % p == 0, y), ct)
+
+  # degree <= 3: always A_n or S_n
+  n <= 3 && return true
+
+  # if n is a prime then any transitive action is primitive
+  is_prime(n) && return true
+
+  # Now use the following observation:
+  #
+  # Lemma: Let G \leq Sym(n) be transitive. If there is \pi \in G with a
+  # non-trivial cycle whose length p is prime and satisfies n/2 < p < n,
+  # then G is primitive.
+  #
+  # Proof: If B is a block of length b, and n = bk, then G is a subgroup of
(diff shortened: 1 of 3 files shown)


experimental/GaloisGrp/GaloisGrp.jl
```diff
@@ -1783,35 +1783,105 @@ function order_from_shape(ct::Set{CycleType}, n)
...
 
 
-#https://mathoverflow.net/questions/286316/recognition-of-symmetric-groups-in-gap
-#show primitivity by finding a permutation which has a p cycle for n/2<p<n−1
+# Given a set of cycle types of a group which acts transitively on n points,
+# this function returns `true` if the group must be primitive. Otherwise
+# `false` is returned (which does *not* imply that the group is imprimitive).
+#
+# See Theorem 10.2.1 on page 225 in Ákos Seress, Permutation group
+# algorithms, Cambridge Tracts in Mathematics, 152.
```
ThomasBreuer: The book is worth to be added to the Oscar References.

experimental/GaloisGrp/GaloisGrp.jl
```diff
@@ -1783,35 +1783,105 @@ function order_from_shape(ct::Set{CycleType}, n)
...
+
   return false
 end
 
+
+@doc Markdown.doc"""
+    an_sn_by_shape(ct::Set{CycleType}, n::Int)
+
+Return `true` if a transitive group $G \leq Sym(n)$ which contains permutations
+matching the cycle types in `ct` contains $Alt(n)$.
```
ThomasBreuer: and `false` otherwise?

experimental/GaloisGrp/GaloisGrp.jl
```diff
@@ -1783,35 +1783,105 @@ function order_from_shape(ct::Set{CycleType}, n)
...
+  # See e.g. Corollary 10.2.2 on page 225 in Ákos Seress, Permutation group
+  # algorithms, Cambridge Tracts in Mathematics, 152.
+  #
+  # Note that this condition is sufficient but not necessary. Indeed for n <= 7
+  # no such p exists, which is why we have to treat these degrees separately above.
+
+  # multiply all candidate primes
+if true
+  lp = prod(fmpz, PrimesSet(div(n, 2)+1, n-3))
+  return any(x->any(y ->gcd(y[1], lp) > 1, x.s), ct)
```
ThomasBreuer: Why first form a product of primes and then check for a nontrivial gcd with this product? Is it not better to test whether one of the primes divides? (And why `if true`?)

ThomasBreuer (approved):
This is a nice improvement.
I have left a few minor comments.

experimental/GaloisGrp/GaloisGrp.jl
fingolfin: done

experimental/GaloisGrp/GaloisGrp.jl
fingolfin: done

experimental/GaloisGrp/GaloisGrp.jl
fingolfin: This was an optimization (?) already present in the code when I started to modify it. But I now did some benchmarks, and at least for small degrees using `is_prime` is actually faster (I first exclude values in the wrong range)

PR #1860, opened 2023-01-11
Minor tweaks for perm groups / cycle types

fingolfin:
- Remove optional return type argument from number_moved_points
- Add degree(g::PermGroupElem)
- Add degree(c::CycleType)
- order(c::CycleType): allow specifying return type
- Remove CycleType() constructor

### DIFF
--- a/src/Groups/perm.jl
+++ b/src/Groups/perm.jl
@@ -523,7 +523,7 @@ julia> cycle_structure(ans)
 ```
 """
 function cycle_structure(g::PermGroupElem)
-    c = GAP.Globals.CycleStructurePerm(GAP.GapObj(g))
+    c = GAP.Globals.CycleStructurePerm(GAP.GapObj(g))::GAP.GapObj
     # TODO: use SortedDict from DataStructures.jl ?
     ct = Pair{Int, Int}[ i+1 => c[i] for i in 1:length(c) if GAP.Globals.ISB_LIST(c, i) ]
     s = degree(CycleType(ct, sorted = true))
--- a/src/Groups/perm.jl
+++ b/src/Groups/perm.jl
@@ -448,9 +448,6 @@ struct CycleType <: AbstractVector{Pair{Int64, Int64}}
     sort!(s, by = x -> x[1])
     return new(s)
   end
-  function CycleType()
-    return new(Pair{Int, Int}[])
-  end
   function CycleType(v::Vector{Pair{Int, Int}}; sorted::Bool = false)
     sorted && return new(v)
     return new(sort(v, by = x -> x[1]))
--- a/src/Groups/perm.jl
+++ b/src/Groups/perm.jl
@@ -497,6 +497,9 @@ function order(c::CycleType)
   return reduce(lcm, map(x->fmpz(x[1]), c.s), init = fmpz(1))
 end
 
+degree(c::CycleType) = mapreduce(x->x[1]*x[2], +, c.s, init = 0)
+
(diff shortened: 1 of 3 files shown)


src/Groups/perm.jl
```diff
@@ -54,8 +54,16 @@ julia> show(Vector(gen(symmetric_group(5), 2)))
...
+    degree(g::PermGroupElem) -> Int
+
+Return the degree of the parent of `g`. This value is always greater or equal number_moved_points
+
+"""
+degree(g::PermGroupElem) = degree(parent(g))
+
+"""
+    moved_points(x::PermGroupElem) -> Int
+    moved_points(G::PermGroup) -> Int
```
ThomasBreuer: `-> Vector{Int}`

src/Groups/perm.jl
```diff
@@ -89,17 +96,12 @@ julia> g = symmetric_group(4);  s = sylow_subgroup(g, 3)[1];
...
-
 julia> number_moved_points(gen(s, 1))
 3
 
 ```
 """
-@gapattribute number_moved_points(x::Union{PermGroupElem,PermGroup}) = fmpz(GAP.Globals.NrMovedPoints(x.X))::fmpz
-
-number_moved_points(::Type{T}, x::Union{PermGroupElem,PermGroup}) where T <: IntegerUnion = T(GAP.Globals.NrMovedPoints(x.X))::T
+@gapattribute number_moved_points(x::Union{PermGroupElem,PermGroup}) = fmpz(GAP.Globals.NrMovedPoints(x.X)::GapInt)
```
ThomasBreuer: If `fmpz` is called here then the docstring must say `-> fmpz` not `-> Int`.
(The changed test indicates that we do not want to call `fmpz`?)

src/Groups/perm.jl
```diff
@@ -485,9 +484,11 @@ function ^(c::CycleType, e::Int)
   return t
 end
 
-function order(c::CycleType)
-  return reduce(lcm, map(x->fmpz(x[1]), c.s), init = fmpz(1))
-end
+order(::Type{T}, c::CycleType) where T = mapreduce(x->fmpz(x[1]), lcm, c.s, init = T(1))
```
ThomasBreuer: If the result shall have the type `T` then better call `T(x[1])` instead of `fmpz(x[1])`.

src/Groups/perm.jl
```diff
@@ -519,10 +520,10 @@ function cycle_structure(g::PermGroupElem)
     c = GAP.Globals.CycleStructurePerm(GAP.GapObj(g))
```
ThomasBreuer: `::GapObj`?

ThomasBreuer (commented):
I have left a few minor comments.

src/Groups/perm.jl
fingolfin: Done

src/Groups/perm.jl
fingolfin: Thanks, fixed

src/Groups/perm.jl
fingolfin: Indeed, good catch! Fixed now

src/Groups/perm.jl
fingolfin: Indeed, that was the intention and then I messed up *sigh*. Once again, thanks for your close attention

ThomasBreuer (approved):
Thanks.

PR #1864, opened 2023-01-15
Fix method invalidations by disabling SLP getindex methods

fingolfin:
I don't think we need these for anything. In any case, they invalidate tons of code in `Base` and stdlibs.

CC @rfourquet


fieker:
if they are not used, I am happy to remove them, but what about fixing? Ia assume the unspecified `x` is the problem?

fingolfin:
Yes, the unspecified type of `x` is the problem. But this is by design, and I see no way to fix this sensibly, as the very idea is to allow "capturing of index access to arbitrary elements by a lazy expression." To be honest, I am also not convinced this is actually useful, and hence don't think it's worthwhile to try fixing it.

It also leads to ambiguities, and so e.g. SLPs can't be used as keys to dictionaries (though I don't think that's a very useful or common operation either, but who knows...)

fieker:
I agree and will approve

fieker (approved):

PR #1867, opened 2023-01-17
Introduce morphism_of_covered_schemes for projective morphisms.

HechtiDerLachs:
Functionality to take a morphism of projective schemes given by means of a homomorphism of graded rings into a corresponding morphism of covered schemes.


HechtiDerLachs:
Turning projective morphisms into morphisms of covered schemes was actually implemented in order to produce some non-trivial examples of pullbacks of ideal sheaves. This is now achieved. 

The key point is not to require that the ideal sheaf is prescribed on every affine chart, but that it can also be defined on the patches of some finer covering. Then it has to be reconstructed on the affine charts by "going upwards" in the tree and gathering the information from the local patches. 

This was just the easiest case for ideal sheaves. For sheaves of modules, this will be more complicated. But I have some prototype in the earlier draft PR that I will now start polishing. This is then also making way to computation of global sections.

HechtiDerLachs:
@simonbrandhorst : If I remember correctly, you had some concerns with the `CartierDivisor`s. Could you have a look whether this is now satisfactory? 

Note that a `CartierDivisor` is not necessarily given by a single function on *every* affine patch, but only on those of the `trivializing_covering` and stuff below those in the refinement tree. Hence the return value of `D(U)` is always a `Vector` of elements, rather than a single element. But this vector is guaranteed to have length one in the trivializing cases.

experimental/Schemes/Auxiliary.jl
```diff
@@ -0,0 +1,47 @@
+@Markdown.doc """
+    pullback(f::CoveredSchemeMorphism)
+
+Return a function `phi` which takes any reasonable 
+argument `v` associated to the `codomain` of `f` and 
+produces ``f*(v)``.
+"""
```
simonbrandhorst: This interface seems to be too generic to me.  Why return a function? 
I would argue for
`pullback(f::CoveredSchemeMorphism, I::IdealSheaf)`
`pullback(f::CoveredSchemeMorphism, F::CoherentSheaf)`
`pullback(f::CoveredSchemeMorphism, f::CoveredRationalFunction)`
which seems more transparent to me. Note that in some cases it is not really clear (at least to me) what the pullback is.

experimental/Schemes/Auxiliary.jl
```diff
@@ -0,0 +1,47 @@
...
+    phi = covering_morphism(f)
+    ID = IdDict{AbsSpec, Ideal}()
+    for U in patches(domain(phi))
+      f_U = phi[U]
+      V = codomain(f_U)
+      pbf = pullback(f_U)
+      ID[U] = ideal(OO(U), pbf.(gens(II(V))))
+    end
+    return IdealSheaf(X, ID, check=false) # TODO: Set to false eventually.
+  end
```
simonbrandhorst: For $f: X \to Y$  and $\mathcal{I} \leq \mathcal{O}_X$ this looks like
$f^{-1} \mathcal{I} \cdot \mathcal{O}_X$ to me, which is in general not the same as the pullback $f^* \mathcal{I}$ (, which may fail to be an ideal sheaf unless $f$ is flat).
See
http://math.stanford.edu/~vakil/216blog/FOAGnov1817public.pdf
16.3.9 for a nice explanation.

experimental/Schemes/Auxiliary.jl
simonbrandhorst: Further I believe that anonymous functions make debugging a bit harder. 
Because in the traceback they only get some number and not a sensible name.

experimental/Schemes/Auxiliary.jl
```diff
@@ -0,0 +1,47 @@
...
+      V = codomain(phi[U])
+      for W in patches(trivializing_covering(C))
+        success, par = _have_common_ancestor(V, W)
+        if success
+          inc_W_flat = _flatten_open_subscheme(W, par)
+          h = complement_equation(codomain(inc_W_flat))
+          UW = PrincipalOpenSubset(U, pullback(phi[U])(OOX(par, V)(h)))
+          triv_dict[UW] = pullback(phi[U])(first(C(V)))
+        end
+      end
```
simonbrandhorst: I have absolutely no clue what happens here. Please provide some comments in the code.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -0,0 +1,40 @@
...
+        C = Covering(collect(keys(D)), IdDict{Tuple{AbsSpec, AbsSpec}, AbsGlueing}())
+        return new{typeof(X)}(X, I, C)
+    end
+end
+
+function (C::CartierDivisor)(U::AbsSpec)
+    return gens(C.I(U))
+end
+
+variety(C::CartierDivisor) = C.X
```
simonbrandhorst: `X` is not necessarily a variety but rather an arbitrary scheme .  So this seems odd to me.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -0,0 +1,40 @@
+export CartierDivisor
+export trivializing_covering
+
+@attributes mutable struct CartierDivisor{
```
simonbrandhorst: A potential problem: 
The way you model a Cartier divisor is basically as Weyl divisor + a cover where it is locally principal.
This has the drawback that we can only write down effective Cartier divisors and the group structure is not defined. Overcoming these obstructions seems awkward to me.
Defining them along the definition as global sections of $\mathcal{K}^* / \mathcal{O}^*$
seems more sustainable in the long run.

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: > This interface seems to be too generic to me. Why return a function?

Why not? Then you can write `pbf = pullback(f)` and then later `pbf(I)` on your ideal sheaf. Just like you would have written it in a mathematics textbook. You get the $f^*$ object right in your hand. Don't you think this is nice?

Debugging: Yes, maybe it's a bit more tedious, but it works. 

Either way: If there are further objections to doing it this way, I mean also from the other Oscar-folks, then I can change this.

experimental/Schemes/CartierDivisor.jl
HechtiDerLachs: Why? Cartier divisors can be defined on arbitrary schemes.

Edit: I was looking at [this reference](https://stacks.math.columbia.edu/tag/01WQ)

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: But this is about the difference of the pullback of `I` seen as a module vs. `I` seen as an ideal. I guess in the latter case we really want $f^{-1}I\cdot \mathscr O_X$, don't we?

experimental/Schemes/CartierDivisor.jl
HechtiDerLachs: Sorry, this is too hard to me to imagine what your concerns are really about at this point. Let's discuss this in person.

experimental/Schemes/Auxiliary.jl
simonbrandhorst: Yes, I think you are doing the right thing. Though I am not so sure about the name "pullback" . 
At least it needs to be documented individually I think.

experimental/Schemes/CartierDivisor.jl
simonbrandhorst: Yes. This command is missleading if X is not a variety. E.g. not over an algebraically closed field, not irreducible or not reduced. Hence I prefer the name scheme.

experimental/Schemes/CartierDivisor.jl
HechtiDerLachs: Alright, I think, I see better now what you mean. 

It is true: This class should probably be called `EffectiveCartierDivisor` rather than `CartierDivisor`. And I remember that I had already pondered about this renaming. In the Stacks project (linked above) they do the same thing: Define effective Cartier divisors and then extend this half group to an honest group by the usual process. 

So overcoming these difficulties is not that difficult in the end: A `CartierDivisor` should then simply be a formal sum of `EffectiveCartierDivisor`s, similar to our current implementation of `WeilDivisor`s. That would also have the benefit that we can absorb the exponents in the coefficients and don't bloat the local equations by exponentiation. 

The question is: Where do we want to go with this? What kind of functionality do we expect? The one I have in mind is intersection theory: I want to turn a `CartierDivisor` into a line bundle `L` and intersect any `WeylDivisor` with the first Chern class of `L`. For this, I need rational sections in a line bundle, right? The good thing is that the intersection is linear, so we can split the problem in taking line bundles of `EffectiveCartierDivisor`s and their rational sections and I think this can be brought to work. 

I don't have a clear idea of how to do things "along the definition as global sections of $$\mathcal K^\star \mathscr O^\star$$, but I see several problems there: We don't have the units of either $$\mathcal K^\star$$ nor $$\mathscr O^\star$$ as an honest group in Oscar. In particular, there is no `quo`-method that we could call on that. Maybe, we can fiddle our way around that, but that would still require introducing another sheaf (which is probably very hard to model) and sections in it. On a first glance, this seems like way more work to do it right.

experimental/Schemes/Auxiliary.jl
```diff
@@ -0,0 +1,175 @@
...
+          # complement equation.
+          h = complement_equation(codomain(inc_W_flat))
+          UW = PrincipalOpenSubset(U, pullback(phi[U])(OOX(par, V)(h)))
+          # Finally, we can add the trivialization on U ∩ f⁻¹(W) to our list.
+          triv_dict[UW] = pullback(phi[U])(first(C(V)))
+        end
+      end
+    end
+    return CartierDivisor(X, triv_dict) # We have to leave the checks back on, because local 
+                                        # equations might pull back to zero or zero divisors.
```
simonbrandhorst: Will the error message be helpful? I.e. that the pullback in this case is not defined? 
Otherwise the user will think it is a bug and blame us.

experimental/Schemes/CartierDivisor.jl
simonbrandhorst: A Cartier divisor could be modeled by giving  $(U_i,f_i)$ where $f_i$ is an element of the total ring of fraction of $\mathcal{O}(U_i)$, the $U_i$ form an open cover of $X$ and $(f_i|U_i \cap U_j) / (f_j | U_i \cap U_j)$ is a unit in $\mathcal{O}(U_i \cap U_j)$. 
I can see your concerns about this approach. 

With the theoretical backing of the stacks project, I can see your approach work too - and probably it is less work. 
Just give the references in the code & in the docu. So thumbs up.

experimental/Schemes/Auxiliary.jl
simonbrandhorst: In mathematics the f^* is either a morphism and has a well defined domain and codomain 
or it is a functor. It has a domain category and a codomain category.
Now this thing is everything in one.  I can see that it is convenient and in math notation we just write f^* (but often also an additional sentence about it.)
But I am not so sure about the implications of this style of coding.

One drawback is that you cannot document its different dispatches separately. 
Any opinions @thofma  @fingolfin ?

simonbrandhorst (approved):

experimental/Schemes/Auxiliary.jl
thofma: Handing around anonymous functions is not so nice for the reasons Simon mentions. Also it makes it rather awkward to write function signatures that accept such objects.

Another way would be returning some new struct and do the call overloading. Then one could document it and still have the `pbf(I)` syntax.

But I also agree with Simon that it is not clear what this actually is mathematically. I have never seen "let g be the pullback of f. Now consider g I" in a textbook (which would be how I read your code snippet with`pbf` above). Probably I should read different textbooks.

HechtiDerLachs:
From my side, I think this PR is now at a good point to stop. The new functionalities are there and they seem to work.

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: I changed this.

experimental/Schemes/CartierDivisor.jl
HechtiDerLachs: Yes, you are right. Sorry, I didn't get it at first.

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: For some reason, I didn't see that the discussion had proceeded here. And I see your points. 

`pbf` is not a rigorous mathematical object, but it is used all over the place by whoever writes down mathematics. In this sense it appears in the textbooks, but as a kind of "joker symbol" which is converted into the right functor implicitly from the context provided by its argument. Since (as far as I understand) we would like to comfort the user as much as possible (or reasonable) with being able to write down their mathematics in the computer as they would write it down on paper, I think that the approach pursued here makes sense. In this sense, it is even close to the textbooks. 

Let me make a proposal: Calling `pullback(f::SomeMorphismType)` returns an object of type `UniversalPullbackSymbol` -- not a rigorous mathematical object, but really a symbol in the sense of what you would write in a paper. It can simply be defined by 
```julia
struct UniversalPullbackSymbol{MorphismType}
  f::MorphismType # one field to store the morphism
end
```
Then, similar to @thofma's suggestion, we can write 
```julia
function (pbf::UniversalPullbackSymbol)(M::Any) 
  return pullback(pbf.f, M)
end
```
to defer it to whatever pullback is defined for `M` along `f`. This means, of course, that the programmer has to implement `pullback(f::MorphismType, O::ObjectType)` for all morphisms and objects as necessary. We would then document these methods.

experimental/Schemes/Auxiliary.jl
```diff
@@ -19,46 +19,111 @@ function pullback(f::CoveredSchemeMorphism)
...
+  function pullback_func(CC::Covering)
+    psi = restrict(f, CC)
+    return domain(psi)
   end
 
   return pullback_func
 end
 
+@Markdown.doc """
+    restrict(f::CoveredSchemeMorphism, DD::Covering)
```
simonbrandhorst, line 100: I would have called it refine. 
But since it is in experimental, I am good to go.

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: I added a note in the docstring of idealsheaves. Making an extra docstring for the pullback didn't seem to be the right thing, since the function is too generic.

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: But you don't refine f, really. This produces a refinement of the `covering_morphism(f)`. But then again, this is not a restriction of `f`, either. Maybe `refine_covering_morphism`?

simonbrandhorst: The `_covering_morphism` does not really add anything new since it is the first argument anyways.
I think we lack the mathematical terminology here. So maybe we are free to use our own terminology. But we should be  consistent. 
General situation: Some $X$ is given on a covering $C$ and we have a refinement $D \to C$. 
We could denote that $X|D$ and say it is the restriction. Because in a sense, its patches are the restrictions. 
The most common example would probably be a section of a sheaf (including a hom sheaf).

simonbrandhorst: So unless we come up with better terminology I am okay with restriction.

experimental/Schemes/Auxiliary.jl
simonbrandhorst: Agreed.

PR #1869, opened 2023-01-19
More on automorphism groups of `ZLat` and `TorQuadMod`

StevellM:
We cover now some corner cases, following recent changes on Hecke:
- isometry group of `ZLat` in the indefinite rank 2 case (using algorithm for binary quadratic forms);
- orthogonal group of `TorQuadMod` which are not semi-regular and which do not split their radical quadratic (using partial isometries).

Minor changes on some existing functions:
- reduce the number of generators used to create orthogonal groups of `TorQuadMod`;
- avoid to compute the full orthogonal group of the discriminant group in `image_in_Oq`;
- use `snf` for abelian groups in the function `_orthogonal_group` for `TorQuadMod` whose underlying abelian group is trivial but not in normal form.

New feature: for embedding of `TorQuadMod` which splits their codomain, we add a function which computes an embedding of the orthogonal groups, by extending by the identity on the complement (in the splitting).

I added also the bump version for Hecke to the last release v0.16.9 otherwise my PR cannot run


simonbrandhorst (approved):

simonbrandhorst:
Tests fail.

simonbrandhorst (approved):

PR #1872, opened 2023-01-20
Added data-structure WeakKeyIdDict

mauro3:
As discussed with @HechtiDerLachs over email, here the PR for adding `WeakKeyIdDict`s.  Feel free to tweak it.

History: This originated in https://github.com/JuliaLang/julia/pull/28161 and transitioned to https://github.com/JuliaLang/julia/pull/28182. Then subsequently I de-coupled it from Julia and made a PR in DataStructures.jl https://github.com/JuliaCollections/DataStructures.jl/pull/402, which thus far has not been merged either.

### DIFF
--- a/src/DataStructures/weakkeyid_dict.jl
+++ b/src/DataStructures/weakkeyid_dict.jl
@@ -1,0 +1,133 @@
+# Weak key dict using object-id hashing/equality
+# see also https://github.com/JuliaCollections/DataStructures.jl/pull/402
+
+# Type to wrap a WeakRef to furbish it with objectid comparison and hashing.
+struct WeakRefForWeakDict
+    w::WeakRef
+    WeakRefForWeakDict(wr::WeakRef) = new(wr)
+end
+WeakRefForWeakDict(val) = WeakRefForWeakDict(WeakRef(val))
+Base.:(==)(wr1::WeakRefForWeakDict, wr2::WeakRefForWeakDict) = wr1.w.value===wr2.w.value
+Base.hash(wr::WeakRefForWeakDict, h::UInt) = Base.hash_uint(3h - objectid(wr.w.value))
+
+"""
+    WeakKeyIdDict([itr])
+`WeakKeyIdDict()` constructs a hash table where the keys are weak
+references to objects, and thus may be garbage collected even when
+referenced in a hash table.
+See [`Dict`](@ref) for further help.
+"""
+mutable struct WeakKeyIdDict{K,V} <: AbstractDict{K,V}
+    ht::Dict{WeakRefForWeakDict,V}
+    lock::ReentrantLock
+    finalizer::Function
+
+    # Constructors mirror Dict's
+    function WeakKeyIdDict{K,V}() where V where K
+        t = new(Dict{WeakRefForWeakDict,V}(), ReentrantLock(), identity)
+        t.finalizer = function (k)
+            # when a weak key is finalized, remove from dictionary if it is still there
+            if islocked(t)
+                finalizer(t.finalizer, k)
+                return nothing
+            end
+            delete!(t, k)
+        end
+        return t
+    end
+end
+function WeakKeyIdDict{K,V}(kv) where V where K
+    h = WeakKeyIdDict{K,V}()
+    for (k,v) in kv
+        h[k] = v
+    end
+    return h
+end
+WeakKeyIdDict{K,V}(p::Pair) where V where K = setindex!(WeakKeyIdDict{K,V}(), p.second, p.first)
+function WeakKeyIdDict{K,V}(ps::Pair...) where V where K
+    h = WeakKeyIdDict{K,V}()
+    sizehint!(h, length(ps))
+    for p in ps
+        h[p.first] = p.second
+    end
+    return h
+end
+WeakKeyIdDict() = WeakKeyIdDict{Any,Any}()
+
+WeakKeyIdDict(kv::Tuple{}) = WeakKeyIdDict()
+Base.copy(d::WeakKeyIdDict) = WeakKeyIdDict(d)
+
+WeakKeyIdDict(ps::Pair{K,V}...)           where {K,V} = WeakKeyIdDict{K,V}(ps)
+WeakKeyIdDict(ps::Pair{K}...)             where {K}   = WeakKeyIdDict{K,Any}(ps)
+WeakKeyIdDict(ps::(Pair{K,V} where K)...) where {V}   = WeakKeyIdDict{Any,V}(ps)
+WeakKeyIdDict(ps::Pair...)                            = WeakKeyIdDict{Any,Any}(ps)
+
+function WeakKeyIdDict(kv)
+    try
+        Base.dict_with_eltype((K, V) -> WeakKeyIdDict{K, V}, kv, eltype(kv))
+    catch e
+        if !Base.isiterable(typeof(kv)) || !all(x->isa(x,Union{Tuple,Pair}),kv)
+            throw(ArgumentError("WeakKeyIdDict(kv): kv needs to be an iterator of tuples or pairs"))
+        else
+            rethrow(e)
+        end
+    end
+end
+
+Base.empty(d::WeakKeyIdDict, ::Type{K}, ::Type{V}) where {K, V} = WeakKeyIdDict{K, V}()
+
+Base.islocked(wkh::WeakKeyIdDict) = islocked(wkh.lock)
+Base.lock(f, wkh::WeakKeyIdDict) = lock(f, wkh.lock)
+Base.trylock(f, wkh::WeakKeyIdDict) = trylock(f, wkh.lock)
+
+function Base.setindex!(wkh::WeakKeyIdDict{K}, v, key) where K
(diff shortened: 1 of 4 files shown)


HechtiDerLachs:
Thank you very much! This looks very good to me!

Unfortunately, the tests fail. And it looks like what you added seems to really interfere with some other internals. On a first glance, I do not see what exactly is going on, but with the help of those people whose code has been broken here, we will manage. 

Other than that: @fingolfin, @thofma : Do you agree on the places where the files and include statements have been put?

mauro3:
Yeah, that looks like a pretty weird error.  There seems to be some odd interaction occurring between the WeakKeyDicts in Hecke and this here...

In https://github.com/mauro3/Oscar.jl/tree/m3/weakkeyid-dict-error there is a slightly more minimal reproducer.  The line
https://github.com/mauro3/Oscar.jl/blob/81d01e3bd9d06f92a3801d8298252db46e96e4a5/test/DataStructures/weakkeyid_dict-test.jl#L12
then leads to the failure when executing 
https://github.com/mauro3/Oscar.jl/blob/81d01e3bd9d06f92a3801d8298252db46e96e4a5/test/Groups/abelian_aut.jl#L56
(this is on my local Julia 1.8.3 install)

Not sure, but might be a Julia bug...
Alternatively, running this code or not could trigger some garbage collection leading to that weak-ref not being defined anymore.  Albeit, sprinkling `GC.gc()` statements around did not reproduce the error.

HechtiDerLachs:
I checked out this pr on my local machine (running on julia 1.8.3).

To me it looked like the error occured in line 58 (not 54) of `test/Groups/abelian_aut.jl`. However, for me the tests seem to just go through! 

I have to say, though, that I added the two lines 
```julia
    @show L.weak_vertices_rev[k]
    @show typeof(L.weak_vertices_rev[k])
```
just before the assertion in line 461 of `src/GrpAb/Lattice.jl` in Hecke.jl (since I was curious what these objects really are). 

But removing the two lines again does not seem to affect the tests (apart from printing stuff): They are still running for me. 

Edit: In order to add those two lines, I had to do `]dev Hecke`. That might have made a difference; looking into it now.

HechtiDerLachs:
Calling `]free Hecke` was complaining:
```
ERROR: Unsatisfiable requirements detected for package Hecke [3e1990a7]:
 Hecke [3e1990a7] log:
 ├─possible versions are: 0.5.0-0.16.6 or uninstalled
 └─restricted to versions 0.16.9-0.16 by Oscar [f1435218] — no versions left
   └─Oscar [f1435218] log:
     ├─possible versions are: 0.11.3 or uninstalled
     └─Oscar [f1435218] is fixed to version 0.11.3-DEV
```
so I had to execute `]up`. Calling `]free Hecke` again after that succeeded. 

Then, I was starting up Oscar on my local branch with this PR and `test/Groups/abelian_aut.jl` went through, again.

HechtiDerLachs:
@thofma : Could you have a look at the tests? I was talking with @fieker today since @simonbrandhorst suggested that this might have been his code that's producing the trouble here. But @fieker said, it would have been most likely you. Can you explain why the `@assert` is there, why it is failing now and whether one can do something about it?

thofma:
When I wrote that code I was not too familiar with the GC. There is probably some lock missing, but the code is a bit complicated. It is the end of the term so I am a bit short on time. Once the term is over, I can say/do more.

HechtiDerLachs:
Ok. Thanks for having a look! Let's just not forget about this.

mauro3:
Also note, there is at least one bug in the PR as it stands: when a key gets GC'ed then the dict-entry is not dropped but lingers as a "nothing" key.  I can look into fixing this, if interest is around.

fingolfin:
Is this based on the code of `WeakKeyDict`? If so, which version? Because `WeakKeyDict` underwent some major revisions in the past few years, including some to fix GC issues, see e.g. https://github.com/JuliaLang/julia/pull/38180

mauro3:
Based on code from 2018, see "History" in the first post.

thofma:
I released a new Hecke with a fix and rebased this PR here. This seems to have fixed the problem. There are now new exciting errors with segmentation faults (https://github.com/oscar-system/Oscar.jl/actions/runs/4191718023/jobs/7267185929).

HechtiDerLachs:
> I released a new Hecke with a fix and rebased this PR here.

Thank you very much for this! Unfortunately, I will be gone for some weeks now and don't have time look into this. But I shall get back to it!

fingolfin:
Actually, we could use this in Nemo and perhaps even in AA.

So, I wonder, why not add it to AA, next to our `WeakValueDict`? Anyone got objections to that?

thofma:
> Actually, we could use this in Nemo and perhaps even in AA.
> 
> So, I wonder, why not add it to AA, next to our `WeakValueDict`? Anyone got objections to that?

I like it!

HechtiDerLachs:
Did this discussion already move to Nemo/AA, then?

thofma:
No, nothing has changed since my comment here: https://github.com/oscar-system/Oscar.jl/pull/1872#issuecomment-1432739565. (Not quite true. There is now also a merge conflict.)

HechtiDerLachs:
I tried again and the current version seems to work well now for the particular setting I have in mind. Here's some sample code: 
```julia
using Test

wkd = WeakKeyDict{AbsSpec, MatrixElem}()
P, _ = QQ["x", "y", "z"]
X = Spec(P)
U = hypersurface_complement(X, P[1])
wkd[X] = P[one(P); P[1]]
wkd[U] = OO(U)[one(OO(U)); gens(OO(U))[1]]

@test haskey(wkd, X)
@test haskey(wkd, U)

U = hypersurface_complement(X, P[1]) # override U
GC.gc() # Call the garbage collector

@test length(keys(wkd)) == 1

X = Spec(P) # override X
U = hypersurface_complement(X, P[1]) # override U again as the previous one still had a reference to the original X

GC.gc() # Call the garbage collector
@test isempty(keys(wkd))
```

Edit: Sorry, my bad. I was using `WeakKeyDict` instead of `oscar.WeakKeyIdDict`. Replacing that accordingly in the above code leads to errors.

HechtiDerLachs:
Here's a new chunk of code against which we can measure: 
```julia
using Test

wkd = Oscar.WeakKeyIdDict{AbsSpec, MatrixElem}()
P, _ = QQ["x", "y", "z"]
X = Spec(P)
U = hypersurface_complement(X, P[1])
wkd[X] = P[one(P); P[1]]
wkd[U] = OO(U)[one(OO(U)); gens(OO(U))[1]]

@test haskey(wkd, X)
@test haskey(wkd, U)

### Delete only one key
U = hypersurface_complement(X, P[1]) # override U
@test !haskey(wkd, U) # check that `==` is not used.
GC.gc() # Call the garbage collector

@test length(keys(wkd)) == 1 # Throws an error: The number of stored keys is 2.
wkd[X] # Does not finish because of some internal lock. 
wkd # Throws a typeassert error. If I remember correctly, the key `X` was just 
    # replaced by `nothing` of type `Nothing` instead of being garbage collected. 
    # However, I can not reproduce this error right now. 

### Delete all keys
X = Spec(P) # override X
U = hypersurface_complement(X, P[1]) # override U again as the previous one still had a reference to the original X

GC.gc() # Call the garbage collector
@test isempty(keys(wkd))
```
It still produces several errors as annotated above.

fingolfin:
I have just rebased this (but did not change anything otherwise). I'll now sync it up with change made to `WeakKeyDict` in recent Julia versions

fingolfin:
@mauro3 I've updated this implementation to match the latest `WeakKeyDict` implementation in Julia 1.6 and later, which is known to fix various GC issues. I also added a comment explaining why we have to use `WeakRefForWeakDict(key)` instead of `key` in a bunch of place, even though `WeakKeyDict` gets away with using `key` (instead of `WeakRef(key)`), because I stumbled over that, and I think it might be helpful for the next person who has to touch this code.

Let's see how it fares... We still may want to move this to `AbstractAlgebra`, though. But since we had issues precisely with the interaction with Oscar, I felt it would be good to first get it working here, before opening a new PR for `AbstractAlgebra`.

fingolfin:
The example by @HechtiDerLachs still fails, there's clearly a bug there:
```
julia> length(keys(wkd))
2

julia> keys(wkd)
KeySet for a Oscar.WeakKeyIdDict{AbsSpec, MatrixElem}(...). Keys:
  Spec of multivariate polynomial ring

julia> collect(keys(wkd))
2-element Vector{AbsSpec}:
    Spec of multivariate polynomial ring
 #undef
```
I'll look into it.

fingolfin:
Ah, fixed. I forgot `Base.` in a few places. Now the example by @HechtiDerLachs works fine for me.

fingolfin:
Nice, all tests pass (at least the ones that pass elsewhere; Julia nightly and 1.10 are as broken as ever).

I'll open a PR for AbstractAlgebra then.

fingolfin:
See https://github.com/Nemocas/AbstractAlgebra.jl/pull/1419

fingolfin:
`WeakKeyIdDict` is now in AbstractAlgebra 0.32.0, which will soon be available in Oscar (after we also update Nemo, Hecke, Singular...)

PR #1874, opened 2023-01-23
Projective bundles

HechtiDerLachs:
We would like to be able to associate to any vector bundle `E`, given as a `CoherentSheaf` on `X`, its projectivization `IP(E)` as a `CoveredProjectiveScheme`. 

Up to now, this is only preliminary work to achieve this, but more is about to follow.


src/Modules/UngradedModules.jl
```diff
@@ -6966,3 +6966,73 @@ function change_base_ring(f::Hecke.Map{DomType, CodType}, M::SubQuo) where {DomT
   map = SubQuoHom(M, MS, gens(MS), f)
   return MS, map
 end
+
```
HechtiDerLachs, line 6991: @jankoboehm : 

Could you have a look at this section? I didn't find that functionality, so I just wrote it up. I hope it goes along with your ideas on the design.

src/Rings/ReesAlgebra.jl
```diff
@@ -0,0 +1,166 @@
...
+
+function is_surjective(f::ModuleFPHom)
+  return iszero(cokernel(f))
+end
+
+function is_isomorphism(f::ModuleFPHom)
+  return is_injective(f) && is_surjective(f)
+end
+
+### Auxiliary deflections for MPolyQuos to make arithmetic work in the Rees algebras
```
HechtiDerLachs, line 127: @wdecker : 

Since you are working on the `MPolyQuo`s these days: For my purposes, it would be nice, if the code on `MPolyQuo`s was so generic that it would also work when the coefficient ring is not a field (resp. Singular-understandable). Do you think, it is possible, to also point out two or three basic methods that must be 
implemented to derive the remaining functionality in a generic way? 

For instance, these few methods below could then be sufficient. Imho that would be great!

HechtiDerLachs:
Since this PR has already become quite big, I open it for review now. Keep in mind, however, that some parts are extracted to #1876 . Hence, feel free to postpone your review until that one's merged.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -761,9 +832,76 @@ underlying_presheaf(M::HomSheaf) = M.M
...
+  function DirectSumSheaf(X::AbsCoveredScheme, summands::Vector{<:AbsCoherentSheaf})
+    all(x->(scheme(x)===X), summands) || error("summands must be defined over the same scheme")
+    OOX = OO(X)
+    all(x->(sheaf_of_rings(x)===OOX), summands) || error("summands must be defined over the same sheaves of rings")
+
+    ### Production of the modules on open sets; to be cached
+    function production_func(FF::AbsPreSheaf, U::AbsSpec)
+      result, inc, pr = direct_sum([F(U) for F in summands]..., task=:both)
+      set_attribute!(result, :inclusions, inc) # TODO: Workaround as long as the maps are not cached.
+      set_attribute!(result, :projections, pr)
```
HechtiDerLachs, line 862: @jankoboehm :
As already indicated in my mail, I didn't find directly how to get my hands on the canonical projections and inclusions of the summands via the caching of canonical maps. If this can already be solved in another way, please let me know! 

Otherwise, this seems to do the job for the time being. So working everything else out until Monday does not seem to be that urgent anymore.

jankoboehm: Would Hecke.canonical_injection(G::ModuleFP, i::Int) and Hecke.canonical_projection(G::ModuleFP, i::Int) do the job?

HechtiDerLachs:
I did a rebase after the merge of #1876 . But the files introduced there still show as new which makes this look bigger than it actually is. If somebody can help me with cleaning this up, please speak up!

simonbrandhorst:
Squash everything in one commit and apply that on master?

simonbrandhorst:
Tests fail.

HechtiDerLachs:
> Squash everything in one commit and apply that on master?

Yes. Everything else will be a mess, again. Unfortunately.

HechtiDerLachs:
> But the files introduced there still show as new which makes this look bigger than it actually is.

Turns out, I was working on the wrong branch locally.

simonbrandhorst:
Can you please split this into some smaller pieces? I am not going to review such big PRs.

simonbrandhorst:
Coverage is only at 77% please add some tests.

HechtiDerLachs:
This has now been split up into three parts. Converting to draft.

afkafkafk13:
Is this PR now obsoleted by the three Projective Bundles PRs? If so, please close.

PR #1876, opened 2023-01-24
computation of kernels for towers of polynomial rings via flattenings

HechtiDerLachs:
This is an attempt to make generic polynomial rings useful over some other coefficient rings which are not fields. 

The idea is to deflect the essential functionality for morphisms/ideals/etc. to their counterparts on the respective flattenings. This is needed for treating the Rees algebras in #1874 .


src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+  return S_flat
+end
+
+@attr function _flatten(
+    S::AbstractAlgebra.Generic.MPolyRing{RingElemType}
+  ) where {RingElemType <: MPolyQuoElem}
+  A = base_ring(S)::MPolyQuo
+  R = base_ring(A)::MPolyRing
+  I = modulus(A)::MPolyIdeal
+  kk = coefficient_ring(R)::Field
```
thofma: Why these type annotations?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+    S::AbstractAlgebra.Generic.MPolyRing{RingElemType}
+  ) where {RingElemType <: Union{<:MPolyElem, <:MPolyQuoElem, <:MPolyLocalizedRingElem, 
+                                 <:MPolyQuoLocalizedRingElem
+                                }
+          }
+  _flatten(S) # Will set the attribute
+  return get_attribute(S, :_orig_to_flat)
+end
+
+@attr function _base_ring_to_flat(
```
thofma: Should not have the `@attr`.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+    S::AbstractAlgebra.Generic.MPolyRing{RingElemType}
+  ) where {RingElemType <: Union{<:MPolyElem, <:MPolyQuoElem, <:MPolyLocalizedRingElem, 
+                                 <:MPolyQuoLocalizedRingElem
+                                }
+          }
+  _flatten(S) # Will set the attribute
+  return get_attribute(S, :_flat_to_orig)
+end
+
+@attr function _orig_to_flat(
```
thofma: Should not have `@attr`.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+                                                              )
+  ambient_ring(U) === domain(phi) || error("multiplicative set does not lay in the domain of the morphism")
+  S = codomain(phi) 
+  Q = ideal(S, phi.(gens(prime_ideal(U))))
+  SU = MPolyComplementOfPrimeIdeal(S, Q, check=check)
+  return SU
+end
+
+### Canonical maps involved in flattening of towers of polynomial rings
+@attr function _flat_to_orig(
```
thofma: Should not have `@attr`.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+function (phi::MPolyAnyMap{<:MPolyRing, <:MPolyRing, Nothing})(U::MPolyPowersOfElement;
+                                                               check::Bool=true
+                                                              )
+  ambient_ring(U) === domain(phi) || error("multiplicative set does not lay in the domain of the morphism")
+  S = codomain(phi) 
+  SU = MPolyPowersOfElement(S, phi.(denominators(U)))
+  return SU
+end
+
+function (phi::MPolyAnyMap{<:MPolyRing, <:MPolyRing, Nothing})(U::MPolyComplementOfPrimeIdeal;
```
thofma: Why is that here?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+
+  set_attribute!(S, :_flat_to_orig, S_flat_to_S)
+  set_attribute!(S, :_orig_to_flat, S_to_S_flat)
+  set_attribute!(S, :_base_ring_to_flat, L_to_S_flat)
+
+  return S_flat
+end
+
+### Transfer of multiplicative sets along ring homomorphisms 
+function (phi::MPolyAnyMap{<:MPolyRing, <:MPolyRing, Nothing})(U::MPolyPowersOfElement;
```
thofma: Why is that here?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
+########################################################################
+# Computation of kernels of homomorphisms of polynomial rings 
+# over coefficient rings which are not fields.
```
thofma:
```suggestion
# over coefficient rings which are polynomial rings are variants thereof.
```

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+                                }
+          }
+  f_flat = _flatten(f)
+  K_flat = kernel(f_flat)
+  K = ideal(domain(f), _flat_to_orig(domain(f)).(gens(K_flat)))
+  #set_attribute!(K, :_flat_counterpart, K_flat) # TODO: Support attribute storage?
+  return K
+end
+
+@attr function _flatten(
```
thofma: Can you construct the return type from the type of `S`? If so, we could add it here for type stability.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+  S_flat_to_S = hom(S_flat, S, vcat(gens(S), S.(gens(R))))
+  
+  set_attribute!(S, :_flat_to_orig, S_flat_to_S)
+  set_attribute!(S, :_orig_to_flat, S_to_S_flat)
+  set_attribute!(S, :_base_ring_to_flat, R_to_S_flat)
+
+  return S_flat
+end
+
+@attr function _flatten(
```
thofma: Same as above regarding type stability.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+    S::AbstractAlgebra.Generic.MPolyRing{RingElemType}
+  ) where {RingElemType <: MPolyQuoLocalizedRingElem}
+  Q = base_ring(S)::MPolyQuoLocalizedRing
+  R = base_ring(Q)::MPolyRing
+  L = localized_ring(Q)::MPolyLocalizedRing
+  A = underlying_quotient(Q)::MPolyQuo
+  U = inverted_set(Q)::AbsMultSet
+  I = modulus(Q)::MPolyLocalizedIdeal
+  kk = coefficient_ring(R)::Field
```
thofma: Why all these type annotations?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+  S_flat_to_S = hom(S_flat, S, vcat(gens(S), S.(gens(A))))
+
+  set_attribute!(S, :_flat_to_orig, S_flat_to_S)
+  set_attribute!(S, :_orig_to_flat, S_to_S_flat)
+  set_attribute!(S, :_base_ring_to_flat, A_to_S_flat)
+
+  return S_flat
+end
+
+@attr function _flatten(
```
thofma: Same as above regarding type stability.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+  ) where {RingElemType <: Union{<:MPolyElem, <:MPolyQuoElem, <:MPolyLocalizedRingElem, 
+                                 <:MPolyQuoLocalizedRingElem
+                                }
+          }
+  _flatten(S) # Will set the attribute
+  return get_attribute(S, :_base_ring_to_flat)
+end
+
+### Computation of induced morphisms on flattened towers of polynomial rings
+@attr function _flatten(
```
thofma: Same as above regarding type stability.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -0,0 +1,247 @@
...
+
+  S_flat = _flatten(S)
+  T_flat = _flatten(T)
+  imgs = _orig_to_flat(T).(f.(gens(S))) # The first half of the images
+  imgs = vcat(imgs, _orig_to_flat(T).(T.(coefficient_map(f).(gens(RS))))) # the ones from the base ring
+  return hom(S_flat, T_flat, imgs)
+end
+
+### Deflecting some of the basic methods for ideals in towers of polynomial rings
+@attr function _flat_counterpart(
```
thofma: Why is this not called `_flatten`?

thofma (commented):
The design makes sense. A left a few minor comments.

In an ideal world, `_flatten(S)` would just return a map that could be used for everything, but we are not there yet. So I would have loved if we could just do
```
function ideal_membership(x, I)
  g = _flatten(parent(x))
  return ideal_membership(g(x), g.(I))
```
I think all the information is in the map and the rest is just fluff. The flattened ring is almost always not that useful for itself. But this is just a minor remark.

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: These came out of writing the code and recalling what these objects would be, respectively, in order not to get too confused. Then, I just left them there, because they don't hurt and probably they even improve readability. 

Would you remove them?

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: The `@attr` is weird in the sense that the body of the function does not really produce the attribute to be cached, but it calls another function which manually caches it. But I didn't find a better way to do this. The map is created together with the flattened ring and, hence, should already be cached there. Moreover, it does not make sense to reproduce this map every time it is being called. 

Another solution would be to cache everything in the attributes of the flattened ring, remove the `@attr` here and write the body as a getter to the respective attribute. Better?

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: This one also gives me a bit of pain, because I don't think there's any such thing in general on the mathematical side. But for some maps and multiplicative sets, one can make sense of what this does. I'm afraid, I have to think about it again to make this more rigorous.

It's here, because it's used in the flattening of polynomial rings with localized rings as coefficient rings.

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: "and variants thereof" -- agreed.

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: Yes, that would be good. It's a bit lengthy, but possible.

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: You're right. It could also just be another method of the same function.

HechtiDerLachs:
> I think all the information is in the map and the rest is just fluff.

That could probably also be addressed by introducing a new type and overloading it's call signature, right?

HechtiDerLachs:
> In an ideal world, `_flatten(S)` would just return a map that could be used for everything, but we are not there yet.

This should bring us a bit closer, however.

src/Rings/MPolyMap/flattenings.jl
thofma: We don't use them for documentation purposes elsewhere, that is why I was wondering.

src/Rings/MPolyMap/flattenings.jl
thofma: I personally would put it next to where the multiplicative sets are or where the poly ring maps are.

src/Rings/MPolyMap/flattenings.jl
thofma: I don't see how the `@attr` adds anything. The `_flatten(S)` call is already cached and the return type is not used. So it is basically a noop. And the `@attr` here just does the `get_attribute`, which is already done in the body of the function.

This can just be written as
```
function _base_ring_to_flat(S)
  _flatten(S) # Will set the attribute or do nothing it if is set
  return get_attribute(S, :_base_ring_to_flat)
end
```

thofma:
Yes, looks good, thanks.

thofma:
Why change all the `Base.in`?

thofma:
And do you want to remove the `_flatten` methods?

HechtiDerLachs:
> Why change all the `Base.in`?

Because it turned out when I want to deflect `ideal_membership` for ideals in towers of polynomial rings, it doesn't find the correct method on the other side unless I do these changes. We probably haven't fully agreed on this, yet, but if one day we write down an oscar-wide `Ideal` interface, it would probably say `ideal_membership` is the atomic thing and `Base.in` calls to that internally.

HechtiDerLachs:
> And do you want to remove the `_flatten` methods?

Already on it.

thofma:
But why the useless indirection via `ideal_membership`? It does add nothing and the name is redundant. One of the arguments is already an ideal.

HechtiDerLachs:
> But why the useless indirection via `ideal_membership`? It does add nothing and the name is redundant. One of the arguments is already an ideal.

I don't understand. I thought `ideal_membership` was *the* well respected method, wasn't it? 

Personally, I don't have a strong opinion on these things. I just need a reliable guidance about which names to use!

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -421,30 +203,82 @@ end
...
-  S = base_ring(I)
-  S_flat = _flatten(S)
-  return ideal(S_flat, _orig_to_flat(S).(gens(I)))
+  parent(x) === base_ring(I) || return base_ring(I)(x) in I
+  phi = flatten(parent(x))
+  return ideal_membership(phi(x), phi(I))
 end
 
-function ideal_membership(
+function coordinates(
```
thofma, line 218: I don't see any example, so can you explain what this does? I am bit confused. More precisely, if `v = coordinates(x, I)`, does `x = sum(x[i] * gens(I)[i])` hold?

thofma:
Don't worry about it. Here was probably not the right place to talk about this function :)

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: Yes, `x = sum(v[i]*gens(I, i))` holds, exactly. The `coordinates` function is also implemented for ideals over polynomial rings, quotient rings, and their localizations. 

Edit: It has first been there for the modules where it received its name, I think. Then it was carried over.

thofma: I was confused for a moment and thought that `coefficients` for things in `R[x]` will be in `R`, but this here is about the ring structure and not the `R`-algebra structure.  Makes sense! Thanks.

thofma:
I am happy with the code. Might be a good idea to add tests to cover all the code, e.g. the `coordinates` function and all the other things that are shown not to be covered (not that many).

HechtiDerLachs:
Does anybody know what's going wrong with the tests?

simonbrandhorst:
Looks like they timed out to me? I would just try to restart them and see if it happens again.

HechtiDerLachs:
Now all the required tests pass. 

@simonbrandhorst : Do you feel like merging this? On the other PR I am now *this* close to reconstructing your K3 with the correct glueings... 

Technically, I can produce its singular version, already. Only the resolution of singularities still seems to take hours. But all that depends on this.

simonbrandhorst (approved):

PR #1878, opened 2023-01-26
Blowup morphism

HechtiDerLachs:
Draft PR for joint work.


experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,100 @@
...
+        multiplicity = length(M)
+        mult_flag = true
+      end
+      ideals_on_patches[U] = J_rad
+    end
+    IE = IdealSheaf(Y, ideals_on_patches)
+    p.exceptional_divisor = WeilDivisor(Y, ZZ, IdDict{IdealSheaf, elem_type(ZZ)}(IE => multiplicity))
+  end
+  return p.exceptional_divisor
+end
```
HechtiDerLachs, line 76: @afkafkafk13 : Is this what you need/had in mind? Feel free to tune and adjust this according to your needs!

test/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,14 @@
+R, (x,y,z) = QQ["x", "y", "z"]
```
HechtiDerLachs, line 1: @afkafkafk13 , @simonbrandhorst : Here's the test file, covering the already working functionality. This can be extended by items on a wishlist: What would you like to be able to write? What's the intended behaviour? We can start with the tests and then work our way up to the actual implementation.

PR #1879, opened 2023-01-26
Projective bundles part I

HechtiDerLachs:


HechtiDerLachs:
Tests for the duality will follow from the rees algebras.

HechtiDerLachs:
@jankoboehm : Please, have a look at the tests. `is_surjective` appears to not work correctly.

experimental/Schemes/ProjectiveModules.jl
```diff
@@ -47,11 +47,14 @@ function is_projective(M::SubQuo; check::Bool=true)
   A = matrix(a)
 
   success, P, k = _is_projective_without_denominators(A)
-  !success && return false, zero_map(Rr, M), zero_map(M, Rr)
+  !success && return false, hom(Rr, M, [zero(M) for x in gens(Rr)]), hom(M, Rr, [zero(Rr) for x in gens(M)])
```
simonbrandhorst, line 50: what is wrong with `zero_map`?

experimental/Schemes/SimplifiedSpec.jl
```diff
@@ -171,6 +172,20 @@ function _flatten_open_subscheme(
   return _flatten_open_subscheme(W, C, iso=new_iso)
 end
 
+function _flatten_open_subscheme(
```
simonbrandhorst, line 175: please document/comment (even if only internal)

src/Modules/UngradedModules.jl
```diff
@@ -5280,23 +5280,45 @@ end
...
 compose(h::ModuleFPHom, g::ModuleFPHom) = h*g
 
-# TODO: We need to also make all of the code below sensitive to base changes!
--(h::ModuleFPHom) = hom(domain(h), codomain(h), [-h(x) for x in gens(domain(h))])
-function -(h::ModuleFPHom, g::ModuleFPHom)
+-(h::ModuleFPHom{D, C, Nothing}) where {D, C} = hom(domain(h), codomain(h), [-h(x) for x in gens(domain(h))])
+-(h::ModuleFPHom{D, C, T}) where {D, C, T} = hom(domain(h), codomain(h), [-h(x) for x in gens(domain(h))], base_ring_map(h))
+
+function -(h::ModuleFPHom{D, C, T}, g::ModuleFPHom{D, C, T}) where {D, C, T}
+  @assert domain(h) === domain(g)
```
simonbrandhorst, line 5287: I think `@req` which is coming from Hecke is the more appropriate macro for input checks
because the @assert macro may be ignored by the compiler in the future. (See its documentation:
```
  │ Warning
  │
  │  An assert might be disabled at various optimization levels. Assert
  │  should therefore only be used as a debugging tool and not used for
  │  authentication verification (e.g., verifying passwords), nor
  │  should side effects needed for the function to work correctly be
  │  used inside of asserts.
```

src/Modules/UngradedModules.jl
```diff
@@ -6966,3 +6988,73 @@ function change_base_ring(f::Hecke.Map{DomType, CodType}, M::SubQuo) where {DomT
...
+end
+
+@Markdown.doc """
+    dual(f::ModuleFPHom; cod::FreeMod)
+
+For a morphism of modules ``f : M → N`` this returns the morphism 
+``fᵀ : N* → M*, φ ↦ (v ↦ φ(f(v)))`` induced on the duals.
+
+The optional argument allows to specify a free module over the 
+base ring of ``f`` for building the duals of ``M`` and ``N``.
```
simonbrandhorst: No clue what this means mathematically. For the duals I know it is always `hom_R(M,R)` where `M` is an `R`-module.

simonbrandhorst:
Please add some tests. E.g. for homomorphisms with base change.

experimental/Schemes/ProjectiveModules.jl
HechtiDerLachs: It didn't exist.

src/Modules/UngradedModules.jl
HechtiDerLachs: Ok, good to know. But I'm not using it, usually.

src/Modules/UngradedModules.jl
HechtiDerLachs: Yes. But on the computer side, this is different. `R` is not just `R`, but one specific free module of rank 1, located somewhere in memory. `dual` returns a map which allows to evaluate stuff and that lands in that very module. So you might want to specify, where you land.

HechtiDerLachs:
> Please add some tests. E.g. for homomorphisms with base change.

I'm sorry, but these tests are implicit in the other parts. I really don't see why I should go through this to pull them out of there manually. In the end, it did already take me at least two hours to separate this into the three parts we have now. Test coverage seems to be low on this one, but, frankly, I'm already doing other people's work here.

experimental/Schemes/SimplifiedSpec.jl
HechtiDerLachs: Please, have a look at line 106.

simonbrandhorst: Okay great. I see that you document some private functions with 
`#comments`. This is good. 
A suggestion you may ignore:
Turn some comments into docstrings. (Since they are private I am okay with ignoring docstring conventions.) Then one can also ask for the docu in the repl. 
Sometimes I use private methods for my purposes (at my own risk). If they are documented, it is helpful.

src/Modules/UngradedModules.jl
```diff
@@ -6966,3 +6988,73 @@ function change_base_ring(f::Hecke.Map{DomType, CodType}, M::SubQuo) where {DomT
...
+  return M_double_dual, psi
+end
+
+@Markdown.doc """
+    dual(f::ModuleFPHom; cod::FreeMod)
+
+For a morphism of modules ``f : M → N`` this returns the morphism 
+``fᵀ : N* → M*, φ ↦ (v ↦ φ(f(v)))`` induced on the duals.
+
+The optional argument allows to specify a free module over the
```
```suggestion
The optional argument allows to specify a free module of rank one over the 
```

simonbrandhorst (approved):

simonbrandhorst:
> > Please add some tests. E.g. for homomorphisms with base change.
> 
> I'm sorry, but these tests are implicit in the other parts. I really don't see why I should go through this to pull them out of there manually. 

O.K.

PR #1880, opened 2023-01-26
rename ChainComplex -> ComplexOfMorphisms

fieker:


fieker:
requires the corresponding Hecke change. Its merges but not "deployed"?

thofma:
We have to release Hecke 0.17 first.

thofma (approved):

PR #1881, opened 2023-01-26
Projective bundles part ii

HechtiDerLachs:
The part about Rees algebras. Will be rebased, once part I is merged. Please review already in parallel by going through the additional commits only.


simonbrandhorst:
The code looks reasonable. But I can't say I understand what is going on.

experimental/Schemes/SimplifiedSpec.jl
```diff
@@ -171,6 +172,20 @@ function _flatten_open_subscheme(
   return _flatten_open_subscheme(W, C, iso=new_iso)
 end
 
+function _flatten_open_subscheme(
```
afkafkafk13: Here I am not sure that I see the purpose of this function. Can you give me a hint?

src/Rings/ReesAlgebra.jl
```diff
@@ -0,0 +1,185 @@
...
+#end
+#
+#function is_surjective(f::ModuleFPHom)
+#  return iszero(cokernel(f))
+#end
+
+function is_isomorphism(f::ModuleFPHom)
+  return is_injective(f) && is_surjective(f)
+end
+
```
afkafkafk13: Do the '#' in front of is_injective and is_surjective mean that this functionality is inherited from somewhere else? They are used in is_isomorphism after all.

test/Modules/UngradedModules.jl
```diff
@@ -1007,4 +1007,43 @@ end
...
+  F2 = FreeMod(R, 2)
+  F2v, ev = oscar.dual(F2, cod=F1)
+  @test ev(F2v[1])(F2[1]) == F1[1] # the first generator
+
+  FF, psi = oscar.double_dual(F2)
+  @test is_injective(psi) 
+  @test_broken is_surjective(psi) # fails! Why?
+  
+  M, inc = sub(F2, [x*F2[1], y*F2[1]])
+  F1 = FreeMod(R, 1)
```
afkafkafk13: That was already defined in 1012.

test/Modules/UngradedModules.jl
```diff
@@ -1007,4 +1007,43 @@ end
...
+  Mv, ev = dual(M, cod=F1)
+  @test ev(Mv[1])(M[1]) == x*F1[1]
+
+  Mvv, psi = oscar.double_dual(M, cod=F1)
+  @test matrix(psi) == R[x; y]
+  
+  ### Quotient rings
```
afkafkafk13: Start a new testset for duals of quotient rings -- allows for more precise error messages from the tests in the overview page of the PR-tests.

experimental/Schemes/CoveredScheme.jl
```diff
@@ -390,8 +390,8 @@ _compose_along_path(X::CoveredScheme, p::Vector{Int}) = _compose_along_path(X, [
    } <: AbsCoveredSchemeMorphism{
                                  DomainType,
                                  CodomainType,
-                                 CoveredSchemeMorphism,
-                                 BaseMorphismType
+                                 BaseMorphismType,
+                                 CoveredSchemeMorphism
```
afkafkafk13: Are you repairing a previous inconsistency or is there some important reason behind the change of order of arguments.

afkafkafk13 (commented):
Looks good to me. But I need to look at the Rees-algebra once again tomorrow morning.

experimental/Schemes/SimplifiedSpec.jl
HechtiDerLachs: It's part of part I. See the discussion there.

src/Rings/ReesAlgebra.jl
HechtiDerLachs: Yes, it turned out that they were already there. Partially disfunctional, though, as it seems. See the tests in part I.

test/Modules/UngradedModules.jl
HechtiDerLachs: Yes.

test/Modules/UngradedModules.jl
HechtiDerLachs:
```suggestion
end 

@testset "duals over quotient rings" begin
```

experimental/Schemes/CoveredScheme.jl
HechtiDerLachs: No. Just repair. It was wrong before.

src/Rings/ReesAlgebra.jl
afkafkafk13: So 129-135 could be removed before merge, right?

src/Rings/ReesAlgebra.jl
```diff
@@ -0,0 +1,185 @@
...
+#
+# Following Eisenbud, Huneke, Ulrich: "What is the Rees algebra 
+# of a module?"; arXiv:math/0209187v1
+########################################################################
+
+@Markdown.doc """
+    rees_algebra(f::ModuleFPHom{<:ModuleFP, <:FreeMod}; check::Bool=true)
+
+For a *versal* (†) morphism ``f : M → F`` of a module ``M`` into a free 
+module ``F`` this computes the Rees algebra of ``M`` according to [1].
```
fingolfin: Add the paper to the bib and use `[EHU03](@cite)` or so?

Is there a reason not to cite the published version, see <https://mathscinet.ams.org/mathscinet-getitem?mr=1937406> reps. <https://www.ams.org/journals/proc/2003-131-03/S0002-9939-02-06575-9/>?

Of course it may still be useful to give that arXiv link for readers w/o a subscription

src/Rings/ReesAlgebra.jl
```diff
@@ -0,0 +1,185 @@
...
+
+!!!Note: if `check` is set to `true`, the method will check the sufficient
```
fingolfin: I don't think this is valid admonition syntax; it must be `!!! note` on a line of its own, the content then follows with indentation, see <https://docs.julialang.org/en/v1.8/stdlib/Markdown/#Admonitions> resp. how we use it elsewhere

src/Rings/ReesAlgebra.jl
```diff
@@ -0,0 +1,185 @@
...
+# Rees algebras of modules 
+#
+# Following Eisenbud, Huneke, Ulrich: "What is the Rees algebra 
+# of a module?"; arXiv:math/0209187v1
+########################################################################
+
+@Markdown.doc """
+    rees_algebra(f::ModuleFPHom{<:ModuleFP, <:FreeMod}; check::Bool=true)
+
+For a *versal* (†) morphism ``f : M → F`` of a module ``M`` into a free
```
fingolfin: Maybe use a footnote reference, see https://docs.julialang.org/en/v1.8/stdlib/Markdown/#Footnote-references

src/Rings/ReesAlgebra.jl
HechtiDerLachs: Citing the published version is even more accurate, because only the published version has the remark about the usage of the double dual for constructing a versal morphism. 

Thanks for the feedback with all the links below! I will look into that later today.

afkafkafk13 (changes requested):
Content is in principle ready for merging. Can you do the remaining small requested changes:
- fingolfin wanted some changes to the docstrings
- afkafkafk13 asked for removing the commented out lines is_surjective/is_injective

afkafkafk13 (approved):

HechtiDerLachs:
Can somebody help me with the bibtool tests? I don't know what's going wrong.

afkafkafk13:
There might be mandatory fields missing.
Try adding the year (2003) and possibly also a URL (https://www.jstor.org/stable/1194470).

HechtiDerLachs:
Tests fail, because due to recent changes, during the construction of quotient rings `Q = R/I` of multivariate rings `R` it is always attempted to compute a groebner basis of `I`. This does not make sense for our types of coefficient rings. I will point out the problem more precisely below.

PR #1882, opened 2023-01-26
Projective bundles part III

HechtiDerLachs:
The real juice: New functionality for coherent sheaves.

Again, this will be rebased, once the other things are merged.


HechtiDerLachs:
Please start reviewing from commit [d37d316](https://github.com/oscar-system/Oscar.jl/pull/1882/commits/d37d31666ae28a2c5abeceab0a7cdbf4bca2e5fb). 

@afkafkafk13 : Have a look at commit [1a1727f](https://github.com/oscar-system/Oscar.jl/pull/1882/commits/1a1727ff88fc0318fef6a33093d34fd7140d5711) and say whether you approve it. 

[2906766](https://github.com/oscar-system/Oscar.jl/pull/1882/commits/29067666f80b130d0dcc7b8d31614dd3bd16420a) is really the main piece, but now all in one file and hence in one go. Direct sums, PullbackSheaf, and projectivization can be handled disjointly. 

[e7fab54](https://github.com/oscar-system/Oscar.jl/pull/1882/commits/e7fab54bfa0efdb49401c8f99d71191c3c0ee591) should be non-controversial.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -761,9 +833,79 @@ underlying_presheaf(M::HomSheaf) = M.M
...
+########################################################################
+@attributes mutable struct DirectSumSheaf{SpaceType, OpenType, OutputType,
+                                          RestrictionType
+                                         } <: AbsCoherentSheaf{
+                                                               SpaceType, OpenType,
+                                                               OutputType, RestrictionType
+                                                              }
+  summands::Vector{AbsCoherentSheaf{SpaceType, OpenType, OutputType, RestrictionType}}
+  #projections::Vector #TODO: Realize as sections in HomSheafs
+  #inclusions::Vector # TODO: same.
```
simonbrandhorst, line 871: why?

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,208 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+#
+# is a local affine representative of the map f. But then the Uᵢ might 
+# not be `affine_charts` of X, anymore. Thus, we can a priori only 
+# construct the modules locally on X and the `production_func` has 
+# to take care of extending them to the `affine_charts` if necessary.
+#
+# Again, it is clear that this can and should be made lazy.
+#                                                                     =#
+#
+# TODO: PullbackSheaf is not yet fully functional.
```
simonbrandhorst, line 1144: document what is missing?

src/Schemes/CoveredSchemes/Objects/Attributes.jl
```diff
@@ -157,13 +157,36 @@ end
   return domain(inc), inc
 end
 
-@attr function singular_locus(X::AbsCoveredScheme)
+@attr function singular_locus(
+    X::AbsCoveredScheme;
+  )
```
simonbrandhorst: This line break seems non-standard. To me it just makes the code harder to parse.

docs/oscar_references.bib
```diff
@@ -1365,3 +1365,13 @@ @Misc{cohomCalg:Implementation
   year          = {2010},
   url           = {https://github.com/BenjaminJurke/cohomCalg}
 }
+
+@Article{EHU03,
```
afkafkafk13: This Bib-Entry is missing the publication year. Didn't you already resolve this Problem in the preceding PR ProjectiveBundlesII ?
(Is this indicating that the PR still needs to be made up-to-date?)

simonbrandhorst: The diff is messed up as well. Rebase on Master?

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -146,6 +148,7 @@ identifications given by the glueings in the `default_covering`.
   ID::IdDict{AbsSpec, ModuleFP} # the modules on the basic patches of the default covering
   OOX::StructureSheafOfRings # the structure sheaf on X
   M::PreSheafOnScheme # the underlying presheaf of modules for caching
+  C::Covering # The covering of X on which the modules had first been described
```
afkafkafk13: you call this 'default covering' a few lines before, maybe you should stay consistent with this to avoid confusion.

HechtiDerLachs: But that would overwrite a function's name with a local variable. So: No, I don't want to do that.

afkafkafk13: I am refering to the help text (after the  #) not to the name of it

simonbrandhorst: I suppose there is a default covering for `X` and the default covering for the sheaf and they can be different?
If that is the case, please clarify at various places in the code which cover you mean.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -158,46 +161,112 @@ identifications given by the glueings in the `default_covering`.
...
+    # 1) Look up whether U is in the list of patches of default_cov.
+    #    If yes, return the value of the dictionary.
+    #
+    # 2) See whether U hangs below some V in the ancestry tree with V 
+    #    in the default_cov; if yes, restrict from V.
+    #
+    # 3) Otherwise, U is covered by patches {Vᵢ}, i ∈ I, from default_cov and 
+    #    contained in one affine chart W of X 
+    #      W ⊃ U ⊃ Vᵢ.
+    #    We distinguish to sub-cases.
```
simonbrandhorst:
```suggestion
    #    We distinguish two sub-cases. 
```

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -158,46 +161,112 @@ identifications given by the glueings in the `default_covering`.
...
+    #    3.1) U appears in the refinement tree T of W whose leafs consist 
+    #    entirely of the Vᵢs. Let T' be the subtree starting from U and 
+    #    let {Vⱼ}, j ∈ J be the leafs of this subtree. We then recursively 
+    #    build up the modules on the nodes in T'.
+    #
+    #    3.2) U does not appear in the refinement tree T above. Then we 
+    #    first have to build the module on W and restrict from there. 
+    #
+    # The point 3) is not implemented, yet. Instead, the current code 
+    # requires to take refinements hanging under nodes in default_cov.
```
simonbrandhorst: Seems reasonable to me. Though I do not understand the last sentence.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -158,46 +161,112 @@ identifications given by the glueings in the `default_covering`.
...
+        return MU
+      end
+      # We are in case 3) above
+      error("production of modules not implemented in this case")
     end
 
     ### Production of the restriction maps; to be cached
     function restriction_func(F::AbsPreSheaf, V::AbsSpec, U::AbsSpec)
+      # Since the other cases are caught by the methods below, both U and V 
+      # must be `affine_chart`s of X with U contained in V along some glueing.
```
simonbrandhorst, line 290: Checking this condition seems fast. Could you add an `@assert` statement for it? 
I am sure it works now. But maybe the code for the other functions below changes in the future and one
accidentally messes this up. An assert statement could safe some headaches then.

simonbrandhorst: (and similar in some cases below)

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
   print(io, "pushforward of $(original_sheaf(M)) along $(map(M))")
 end
 
+########################################################################
+# Pullback of sheaves along morphisms                                  #
+########################################################################
+
+#=
+# Let f : X ↪ Y be a closed embedding with ideal sheaf ℐ on Y and ℳ
```
simonbrandhorst: Why do we need it to be a closed embedding?
The theory works in general of course. So where in practice do we use the property that `f` is a closed embedding?

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+        floc = morphisms(fcov)[U]
+        MU, map = change_base_ring(pullback(floc), M(codomain(floc)))
+        pullbacks[U] = map
+        return MU
+      end
+
+      # If not, check whether we are hanging below such a patch in the 
+      # refinement tree.
+      if some_ancestor(y->any(x->(x===y), patches(domain(fcov))), U)
+        V = __find_chart(U, domain(fcov))
```
simonbrandhorst: To me this use seems like `some_ancestor`  should return
a `bool` and the ancestor (or a dummy in case of false)

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+  function PullbackSheaf(f::AbsCoveredSchemeMorphism, M::AbsCoherentSheaf)
+    X = domain(f)
+    Y = codomain(f)
+    Y === scheme(M) || error("sheaf must be defined over the domain of the embedding")
+    OOY = sheaf_of_rings(M)
+    OOX = OO(X)
+    fcov = covering_morphism(f)::CoveringMorphism
+    CX = domain(fcov)::Covering
+    CY = codomain(fcov)::Covering
+    pullbacks = IdDict{AbsSpec, Hecke.Map}()
```
simonbrandhorst, line 1176: please comment what is stored here

simonbrandhorst: is that the base change map, i.e  $M \to S \otimes M$ $x \mapsto 1_S \otimes x$?

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+    # For U ⊂ V ⊂ X, f : X → Y, M on Y and F = f^*M we do the following. 
+    # Let ϕ be the `covering_morphism` behind f. 
+    #
+    # 1) If both U and V are in `domain(ϕ)` induce the restriction from 
+    #    the one on Y. 
+    # 2) If V is in `domain(ϕ)` and U is a node hanging below V in 
+    #    the refinement tree, restrict from V. 
+    # 3) If V is in `domain(ϕ)` and U is a subset of V, restrict as usual.
+    # 4) If V is a node hanging below some patch in `domain(ϕ)` and 
+    #    U is a subset, restrict as usual.
```
simonbrandhorst: Could you please write up the commutative algebra behind it? i.e. the corresponding commutative diagram? 
That would help me understand what exactly goes on. [reading math is easier than reading  code]

simonbrandhorst: Looks to me like you restrict and then you tensor.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+
+        # case 2)
+        error("restriction map should have been cached by production")
+      end
+      error("case not implemented")
+    end
+    
+    ident = IdDict{AbsSpec, Union{Hecke.Map, Nothing}}()
+
+    Blubber = PreSheafOnScheme(X, production_func, restriction_func,
```
simonbrandhorst, line 1275: :-)

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+
+########################################################################
+# Auxiliary helper functions for PullbackSheaves                       #
+########################################################################
+#=
+# For an 𝒪(U)-module M this constructs an 𝒪(V)-module N where V is 
+# the `ambient_scheme` of U, together with a map ϕ : N → M 
+# over the canonical map 𝒪(V) → 𝒪(U) such that M ≅ image(ϕ).        =#
+
+function _lift_module(M::SubQuo, U::PrincipalOpenSubset)
```
simonbrandhorst: 1. Is this used anywhere? 
2. Can you expand the comment a little? Is $U$ open in $V$? how is $N$ defined?
3. Please make such comments a docstring :-)

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -1193,10 +1546,90 @@ end
...
-  # leading to affine charts of X.
-  return C
+########################################################################
+# Projectivization of vector bundles                                   #
+#
+# To any vector bundle E → X one can associate its projectivization 
+# ℙ(E) → X. In general, E will be given as a locally free 
+# AbsCoherentSheaf.
+########################################################################
+
```
simonbrandhorst, line 1570: Please document.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -1193,10 +1546,90 @@ end
...
+  algebras = IdDict{AbsSpec, Union{MPolyQuo, MPolyRing}}()
+  on_patches = IdDict{AbsSpec, AbsProjectiveScheme}()
+  for U in patches(C)
+    F = E(U)
+    F isa FreeMod || error("modules must locally be free")
+    r = rank(F)
+    if length(var_names) == 0
+      var_names = ["s$i" for i in 0:r-1]
+    end
+    length(var_names) == r || error("number of names for the variables does not coincide with the local rank of the module")
```
simonbrandhorst: If the scheme is not connected, a locally free sheaf could have different ranks in different components? (or is that excluded?) (probably should be)

simonbrandhorst (commented):
`SimplifiedSpec` and `PrincipalOpen` seem to behave very similar. At some point we could think about a  common abstract type.

Many of the comments could/should be docstrings at some point.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -1193,10 +1546,90 @@ end
...
+    PUV, PUVtoP = fiber_product(OX(U, UV), P)
+    QVU, QVUtoQ = fiber_product(OX(V, VU), Q)
+
+    # to construct the identifications of PUV with QVU we need to 
+    # express the generators of I(U) in terms of the generators of I(V)
+    # on the overlap U ∩ V. 
+    !(G isa Glueing) || error("method not implemented for this type of glueing")
+
+    # The problem is that on a SpecOpen U ∩ V
+    # despite I(U)|U ∩ V == I(V)|U ∩ V, we
```
simonbrandhorst, line 1642: what is I? M?

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -520,21 +520,105 @@ function blow_up(
 #  return projective_version, covered_version, projection_map, exc_div
 end
 
+struct ProjectiveGlueingData
```
simonbrandhorst, line 526: A few comments please.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -158,46 +161,112 @@ identifications given by the glueings in the `default_covering`.
...
+    #
+    # 1) Look up whether U is in the list of patches of default_cov.
+    #    If yes, return the value of the dictionary.
+    #
+    # 2) See whether U hangs below some V in the ancestry tree with V 
+    #    in the default_cov; if yes, restrict from V.
+    #
+    # 3) Otherwise, U is covered by patches {Vᵢ}, i ∈ I, from default_cov and 
+    #    contained in one affine chart W of X 
+    #      W ⊃ U ⊃ Vᵢ.
```
afkafkafk13: This text sounds slightly unclear to me. Allow me to ask whether my understanding of it is correct:
It is the same W for all V_i, right?
This is the case where default_cov and W are completely unrelated (i.e. we do have a single affine chart W, but unfortunately it is not compatible with our default covering of X)?

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -947,6 +1089,215 @@ function Base.show(io::IO, M::PushforwardSheaf)
...
+      end
+
+      # We are in case 3)
+      error("case not implemented")
+    end
+
+    ### Restriction for pulled back sheaves of modules
+    #
+    # For U ⊂ V ⊂ X, f : X → Y, M on Y and F = f^*M we do the following. 
+    # Let ϕ be the `covering_morphism` behind f.
```
afkafkafk13: Do not just fix notation here, but write out clearly in mathematical terms, what the goal is, before going into the cases below, which are only of technical relevance.

experimental/Schemes/CoherentSheaves.jl
afkafkafk13: These four cases do not differ in their theory, only in the algorithmic approach taken to handle different settings of the underlying data. I am content with this part of the documentation.

experimental/Schemes/CoherentSheaves.jl
afkafkafk13: As I see this function, it assumes that on each patch U the rank is constant, which should be reasonable even in the case of a non-connected scheme, because it has already computed a trivializing covering. 

The key question is: How do is_locally_free and trivializing_covering behave in the case of a non-connected scheme? A test for this should be added (no matter whether it is forbidden or handled correctly).

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -1193,10 +1546,90 @@ end
...
+    UV, VU = glueing_domains(G)
+    f, g = glueing_morphisms(G)
+
+    PUV, PUVtoP = fiber_product(OX(U, UV), P)
+    QVU, QVUtoQ = fiber_product(OX(V, VU), Q)
+
+    # to construct the identifications of PUV with QVU we need to 
+    # express the generators of I(U) in terms of the generators of I(V)
+    # on the overlap U ∩ V. 
+    !(G isa Glueing) || error("method not implemented for this type of glueing")
```
afkafkafk13, line 1639: I(U) and I(V) are the modulus of OO(X) on the respective patches, correct?

experimental/Schemes/CoherentSheaves.jl
HechtiDerLachs: This will be taken care of in the next PR. If I change this now, there will be a merge conflict.

simonbrandhorst:
Tests fail.

simonbrandhorst (approved):

afkafkafk13:
@simonbrandhorst: only merge after I approved,too. There is something I want to look at in trivializing_covering before merging.

HechtiDerLachs:
Alright. Finally, there was not so much left to do. If the tests pass, this should finally be OK and I'm done with this one.

experimental/Schemes/CoherentSheaves.jl
HechtiDerLachs: See comment above.

afkafkafk13 (approved):

simonbrandhorst:
`some_ancestor` still needs to be replaced by `has_ancestor` everywhere else. Tests fail.

PR #1887, opened 2023-01-30
Projective bundles part iv

HechtiDerLachs:
Continuation of #1882 , mainly to expose some follow-up difficulties for discussion at the moment.


HechtiDerLachs:
I continued playing a little with @simonbrandhorst 's K3 example, trying to achieve the resolution of singularities of the E_8 singularity. To this end, I made the projective glueings lazy which was already a big step. But now the next bottleneck shows up. 

When running the test file in `test/Schemes/K3.jl`, it seems that most of the computation time goes into `ideal_sheaf_of_singular_locus` and into the `lift` commands triggered by `coordinates(a, I)` for ideals `I`.

@afkafkafk13 : Could you have a look at the singular locus computation? If I remember correctly, this particular implementation was due to you and you also have some expertise and experience in tuning this. 

@wdecker : The `coordinates(a, I)` is still implemented via a hack in some file called `missing_functionality.jl`. That was OK, because we didn't have a `lift`-command that was relying on cached Groebner bases at the time. But that means that for every element `a`, this calls to the singular backend and computes a whole new Groebner basis! Hence the question: Do you already have something cooking somewhere that could help us here?

test/Schemes/K3.jl
```diff
@@ -0,0 +1,97 @@
+#@testset "A K3 surface from an elliptic fibration" begin
```
HechtiDerLachs: @afkafkafk13 : This is the test file to check out the bottleneck for the singular locus computation. The full script never finished for me, but one can also run it in parts, say, until the third blowup.

afkafkafk13: I am already in the middle of analyzing this...

afkafkafk13: Can we discuss about this directly (gather/BBB/phone)?

jankoboehm:
You can try to use the respective coordinates function for modules where all the caching is already in place. If computing over a polynomial ring over a field, it computes as standard a liftstd and caches the transformation matrix and Groebner basis, so further lifts are cheap.

HechtiDerLachs:
I pushed tonight's version from experimenting with the stuff. In my opinion, the problem is that the generators of the `modulus` of the affine charts in the blowups just become unreasonably big. I tried to make them smaller with various calls to `radical` (in my experience Singular's radical command usually provides you with a reasonably small set of generators), but it doesn't seem to help. If you call the common jacobian criterion on such sets of generators, you're lost. But it's an E_8 singularity after all, so the computations in the charts shouldn't be too difficult. People can do this by hand!

@afkafkafk13 : Let's see tomorrow how we can either reduce the number and size of generators or use another method to determine the singular locus after all.

simonbrandhorst:
When doing it by hand I made sure the number of variables stays the same. 
Then the modulus is always given by a single equation. There should be no increase in complexity.

afkafkafk13:
@simonbrandhorst: Yes, this is precisely the point which had gone wrong in yesterdays version and which caused the singular locus to be handed larger and larger matrices, which of course took longer and longer. The task is to automatically detect and handle your simplification steps without breaking the glueing.
It is handled properly (but not glued) in the respective Singular libraries. The additional complication arises from the requirement of providing a glueing. This is what we plan to work on today.

jankoboehm:
Is there already the call to presolve.lib to elimintate the linear part and it does not do the job, or is presolve.lib not used yet?

afkafkafk13:
It was already there in a previous version. The simplification of the blowing up clearly only makes sense by using  this elimpart step, we are aware of this. The problem is  not, how to do it, but to make all parts work together....
Matthias and I are on it. The theory is clear and I have done this before, but it needs to be fiddled into the covered schemes environment properly.

jankoboehm:
Elimpart gives the transformation which induces an isomorphism, so you just replace the chart and plug in the isomorphism resp its inverse in the chart change maps.

afkafkafk13:
Again: this is clear!!! Of course, elimpart is central to this and ,of course, it provides the isomorphism.
It is the environment which needs to swallow these steps properly. 
There should not have been a push to this PR yesterday evening at all and this discussion would not have eaten up your time.

HechtiDerLachs:
@afkafkafk13 : `ClosedEmbedding` should now be repaired. In the end, I had already made the necessary changes due to other reasons some minutes before.

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -98,3 +111,76 @@ end
...
+    pbJ = ideal(OO(U), pullback(p_res).(gens(J)))
+    pbJ_sat = saturated_ideal(pbJ)
+    pbJ_sat = saturation(pbJ_sat, ideal(base_ring(pbJ_sat), lifted_numerator.(E(U))))
+    pbJ = ideal(OO(U), [g for g in OO(U).(gens(pbJ_sat)) if !iszero(g)])
+    ID[U] = pbJ
+  end
+
+  I_trans = IdealSheaf(Y, ID, check=true) # TODO: Set to false
+  inc_Z_trans = CoveredClosedEmbedding(Y, I_trans, covering=CY_simp)
+  Z_trans = domain(inc_Z_trans)
```
afkafkafk13, line 116: Why do you create Z_trans, if you do not use it afterwards?

HechtiDerLachs:
The current implementation should now do more or less what we talked about (@afkafkafk13 ). See the `K3.jl` test-file for how to proceed. It seems the equations are now kept simple.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: Maybe this line is superfluous. I don't remember now. But in any case: It's not created in line 156. The creation takes place already in line 155. And I'm pretty sure, I am using that map in what follows. So no severe harm should have been done.

HechtiDerLachs:
I massaged the code a bit further to push the resolution of the E_8 singularity. However, if you run the `K3.jl`-file now, you will see that after the 4-th blowup you find a situation with three singular points deployed over two charts. Blowing these up directly gets us stuck again, because elim-part does not work correctly and the singular locus routine computes with an ideal of >1000 generators of degree ~40 due to the jacobian criterion. 

@afkafkafk13 : What should we do here?

afkafkafk13:
As a rule of thumb: never ever blow up reducible centers in a single step. Decompose, choose one component, store the rest for further use and then blow up in the chosen one. After carrying along the stored components choose one of those and so on...
(Background: The union of several components has significantly more generators than each component itself and thus makes the blowing up significantly more complicated)

HechtiDerLachs:
Ok. It sounds reasonable. Even though the number of generators is 3 despite the two components in each chart, see the final output of the `K3.jl`-file. Anyway, I have little idea how to achieve this effectively here. Could you take over?

simonbrandhorst:
> As a rule of thumb: never ever blow up reducible centers in a single step. Decompose, choose one component, store the rest for further use and then blow up in the chosen one. After carrying along the stored components choose one of those and so on... (Background: The union of several components has significantly more generators than each component itself and thus makes the blowing up significantly more complicated)

What do we do if the center is irreducible but geometrically reducible?

afkafkafk13:
> What do we do if the center is irreducible but geometrically reducible?

Then we use it as it is -- it is irreducible in the current ring. This rule of thumb just states, what to avoid to keep the situation as simple as possible. It is not a strict rule, but does apply here.

HechtiDerLachs:
Just a heads up: I have a local version now which computes the whole resolution successfully in comparibly short time (<10 sec.). Now I'm working on the pullback of the exceptional divisors and the induced projection maps. Unfortunately, this entails a lot of difficulties about refinements and will take some more time.

HechtiDerLachs:
We can now resolve the singularity and look at the pullbacks of the exceptional divisors in all charts. The current implementation is in the updated `K3.jl`. I hope you find this half-way satisfactory.

simonbrandhorst:
I can review this, once it is rebased.

experimental/Schemes/Auxiliary.jl
```diff
@@ -72,9 +89,20 @@ end
...
+  # and pull back along psi.
   phi = covering_morphism(f)
   triv_dict = IdDict{AbsSpec, RingElem}()
-  psi = restrict(f, trivializing_covering(C))
+  # TODO: restrict f to a common_refinement(codomain(phi), trivializing_covering(C))
+  # (to be implemented)
+  # This might be codomain(phi) itself and that case is not covered, yet.
+  E, a, b = common_refinement(codomain(phi), trivializing_covering(C))
+  psi = restrict(f, E)
+  psi = compose(psi, b)
```
simonbrandhorst: I do not understand the TODO. It seems to be done in 103-105?

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
+export BlowupMorphism
+export center, projection, exceptional_divisor
+
+mutable struct BlowupMorphism{
+                              CodomainType<:AbsCoveredScheme
+                             }
+  projective_bundle::CoveredProjectiveScheme 
+  codomain::CodomainType   # i.A. ein CoveredScheme
```
simonbrandhorst:
```suggestion
  codomain::CodomainType   # in general a CoveredScheme
```

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+  center::IdealSheaf      # on codomain
+  projection::AbsCoveredSchemeMorphism
+  domain::AbsCoveredScheme # i.A. ein CoveredScheme
```
simonbrandhorst:
```suggestion
  domain::AbsCoveredScheme 
```

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
+export BlowupMorphism
+export center, projection, exceptional_divisor
+
+mutable struct BlowupMorphism{
```
simonbrandhorst, line 11: document please

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+    p.projection = covered_projection_to_base(p.projective_bundle)
+  end
+  return p.projection
+end
+
+# TODO: Find better name!
+covered_projective_scheme(p::BlowupMorphism) = p.projective_bundle
+
+@Markdown.doc """
+    exceptional_divisor(p::BlowupMorphism)
```
simonbrandhorst, line 52: can this thing have multiplicities?

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+on ``Y`` associated to the (relative) tautological bundle ``𝒪(1)``. 
+
+On a pair of charts ``V → U`` of the `covered_scheme` of the 
+`projection` of ``p`` this returns the pullback of the `i`-th 
+generator of ``ℐ(U)`` when ``V`` is the `i-1`-st canonical chart 
+of the local blowup over ``U``.
+"""
+function exceptional_divisor(p::BlowupMorphism)
+  if !isdefined(p, :exceptional_divisor)
+    error("exceptional divisor needs to be cached during construction")
```
simonbrandhorst, line 65: The 40 lines of code below will never be reached with this error. Typo?

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+      end
+      ideals_on_patches[U] = J_rad
+    end
+    IE = IdealSheaf(Y, ideals_on_patches)
+    p.exceptional_divisor = WeilDivisor(Y, ZZ, IdDict{IdealSheaf, elem_type(ZZ)}(IE => multiplicity))
+  end
+  return p.exceptional_divisor
+end
+
+### Auxiliary function which were not there; to be moved
```
simonbrandhorst: Let's move them right away

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+@attr function radical(I::MPolyLocalizedIdeal)
+  L = base_ring(I)
+  J = pre_saturated_ideal(I)
+  J_rad = radical(J)
+  return ideal(L, [g for g in L.(gens(J_rad))])
+end
+
+@attr function dim(I::MPolyQuoLocalizedIdeal)
+  return dim(pre_image_ideal(I))
+end
```
simonbrandhorst: Is this correct? The Krull dimension of $I/J$ and $I$ can be different. 
Please document `pre_image_ideal`. No clue which morphism the preimage refers to.

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+    strict_transform(p::BlowupMorphism, inc::CoveredClosedEmbedding)
+
+For a `BlowupMorphism` ``p : Y → X`` and a `CoveredClosedEmbedding` 
+``ι : Z ↪ X``, compute the strict transform ``Z'`` of ``Z`` along ``p`` and 
+return a triple ``(Z', j, π)`` containing the `CoveredClosedEmbedding` 
+``j : Z' ↪ Y`` and the induced projection ``π : Z' → Z``.
+
+!!! note The projection is not yet implemented! Instead, `nothing` is returned 
+as the third item.
+"""
```
simonbrandhorst: Is $\pi$ for free? Or would it be better to refer this to another method `restrict(p,Z',Z)`?

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+
+  # compose the covering morphisms
+  p_cov_simp = compose(phi, p_cov)
+  CX = codomain(p_cov)
+  E = exceptional_divisor(p)
+  for U in patches(CY_simp)
+    p_res = p_cov_simp[U]
+    V = codomain(p_res)
+    J = image_ideal(inc)(V)
+    g = maps_with_given_codomain(inc, V)
```
simonbrandhorst: No clue what `maps_with_given_docomain` does. Please document.
What is `g`? Note that `g` is not used below. So why is it assigned?

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+    pbJ = ideal(OO(U), pullback(p_res).(gens(J)))
+    pbJ_sat = saturated_ideal(pbJ)
+    pbJ_sat = saturation(pbJ_sat, ideal(base_ring(pbJ_sat), lifted_numerator.(E(U))))
+    pbJ = ideal(OO(U), [g for g in OO(U).(gens(pbJ_sat)) if !iszero(g)])
+    ID[U] = pbJ
+  end
+
+  I_trans = IdealSheaf(Y, ID, check=false) # TODO: Set to false
```
simonbrandhorst: Set what to false?

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: Looking at the code it seems to me that there is a non-trivial 3rd item now?

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+end
+
+#=
+#  Z' ↪ Y
+#       ↓ f
+#  Z ↪  X
+#
+#  Assuming f(Z') ⊂ Z, we compute and return the restriction f : Z' → Z.
+=#
+function restrict(f::AbsCoveredSchemeMorphism,
```
simonbrandhorst: Document please.

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+  result = Vector{AbsSpecMor}()
+  for U in keys(morphisms(phi))
+    floc = morphisms(phi)[U]
+    codomain(floc) === V || continue
+    push!(result, floc)
+  end
+  return result
+end
+
+#function saturation(I::MPolyLocalizedIdeal, J::MPolyLocalizedIdeal)
```
simonbrandhorst: If you do not need the saturations here, we can make a separate PR and include them there.

experimental/Schemes/Auxiliary.jl
```diff
@@ -40,6 +41,18 @@ function (f_star::UniversalPullbackSymbol)(M::Any)
   return pullback(f_star.f, M)
 end
 
+# We can do the same thing with pushforward and other symbols.
+struct UniversalPushforwardSymbol{MorphismType}
+  f::MorphismType
+end
+
+# The following refers a call of f_star(M) to a bivariate pullback-method.
```
afkafkafk13: please adjust this copy of the pullback comment to the setting of pushforward: f_lower_star, pushforward

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -10,11 +10,59 @@ export strict_transform, total_transform, weak_transform, controlled_transform
...
+
+### getters for the essential functionality
+base_glueing(PG::AbsProjectiveGlueing) = base_glueing(underlying_glueing(PG))
+inclusion_maps(PG::AbsProjectiveGlueing) = inclusion_maps(underlying_glueing(PG))
+glueing_domains(PG::AbsProjectiveGlueing) = glueing_domains(underlying_glueing(PG))
+patches(PG::AbsProjectiveGlueing) = patches(underlying_glueing(PG))
+glueing_morphisms(PG::AbsProjectiveGlueing) = glueing_morphisms(underlying_glueing(PG))
+
+
+mutable struct LazyProjectiveGlueing{
```
simonbrandhorst, line 53: Document please.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -10,11 +10,59 @@ export strict_transform, total_transform, weak_transform, controlled_transform
...
+                                     GlueingType<:AbsGlueing,
+                                     GlueingDataType
+                                    } <: AbsProjectiveGlueing{GlueingType}
+  base_glueing::GlueingType
+  patches::Tuple{AbsProjectiveScheme, AbsProjectiveScheme}
+  compute_function::Function
```
simonbrandhorst, line 59: computes the glueings? Input output specification? Please document.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -10,11 +10,59 @@ export strict_transform, total_transform, weak_transform, controlled_transform
...
+    (base_scheme(X), base_scheme(Y)) == patches(BG) || error("glueing is incompatible with provided patches")
+    return new{typeof(BG), typeof(glueing_data)}(BG, (X, Y), compute_function, glueing_data)
+  end
+end
+
+# Essential getters
+patches(G::LazyProjectiveGlueing) = G.patches
+base_glueing(G::LazyProjectiveGlueing) = G.base_glueing
+
+# Everything else will trigger the computation
```
simonbrandhorst, line 79: ? I do not understand what is going on.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -207,6 +285,31 @@ function blow_up(W::AbsSpec{<:Field, <:RingType}, I::Ideal;
...
+    p_res = p_cov[U]
+    W === codomain(p_res) || error("codomain not correct")
+    ID[affine_charts(Y)[i]] = pullback(p_res)(gens(I)[i])
+  end
+  E = oscar.EffectiveCartierDivisor(Y, ID, trivializing_covering=domain(p_cov), check=false) # TODO: Set to false
+  set_attribute!(Y, :exceptional_divisor, E)
+  set_attribute!(Bl_W, :exceptional_divisor, E)
+  return Bl_W
+
+  # the version below uses the now deprecated affine cone construction
```
simonbrandhorst: deprecated because it is slower?

experimental/Schemes/Auxiliary.jl
```diff
@@ -40,6 +41,18 @@ function (f_star::UniversalPullbackSymbol)(M::Any)
...
+# The latter is what really needs to be implemented by the programmer. 
+# Also, all parent checks, etc. are expected to happen there.
+function (f_lower_star::UniversalPushforwardSymbol)(M::Any)
+  return pushforward(f_lower_star.f, M)
+end
+
 @Markdown.doc """
     pullback(f::CoveredSchemeMorphism)
```
afkafkafk13, line 57: do not forget to add a docstring for pushforward, too

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -225,23 +328,27 @@ function blow_up(W::AbsSpec{<:Field, <:RingType}, I::Ideal;
   IWh = ideal(S, poly_to_homog(P).(lifted_numerator.(gens(K))))
   return subscheme(P, IWh)
 
-  # Deprecated code below
+  # The code below breaks the projective glueings, because we loose control 
+  # over the generators of the ideal corresponding to the variables of the 
+  # projective spaces.
+
```
simonbrandhorst: Best comment the code below, so that we do not spoil our coverage :-). 
But I like that you keep some failed attempts if one can learn from them.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -395,129 +502,107 @@ end
 #  end
 #end
 
+struct CoveredProjectiveGlueingData
```
simonbrandhorst, line 524: Needs docu. No clue what is going on.

simonbrandhorst: This looks similar to something I have seen in in the Projective bundles part III, the projectivization of a vector bundle. 
How is this similar / something different

simonbrandhorst: Okay I see that the code just moved. So I do not look into the details. Nevertheless some docu would be great.

simonbrandhorst: Okay I see that the code just moved. So I do not look into the details. Nevertheless some docu would be great.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -487,3 +487,59 @@ function order_on_divisor(
...
+end
+
+@attr MPolyQuoLocalizedIdeal function radical(I::MPolyQuoLocalizedIdeal)
+  W = base_ring(I)
+  J = pre_image_ideal(I)
+  return ideal(W, [g for g in W.(gens(radical(J))) if !iszero(g)])
+end
+
+@Markdown.doc """
+    smooth_lci_covering(I::IdealSheaf)
```
simonbrandhorst, line 502: No abbreviations please.

experimental/Schemes/Types.jl
```diff
@@ -595,23 +595,10 @@ identifications given by the glueings in the `default_covering`.
     )
     OOX = StructureSheafOfRings(X)
 
+    ### Production of the rings of regular functions; to be cached
     function production_func(F::AbsPreSheaf, U::AbsSpec)
       # If U is an affine chart on which the ideal has already been described, take that.
       haskey(ID, U) && return ID[U]
-      # The ideal sheaf has to be provided on at least one dense
```
simonbrandhorst, line 603: why is this removed?

experimental/Schemes/Types.jl
```diff
@@ -638,6 +625,29 @@ identifications given by the glueings in the `default_covering`.
     function production_func(F::AbsPreSheaf, U::PrincipalOpenSubset)
       haskey(ID, U) && return ID[U]
       V = ambient_scheme(U)
+      # In case the ambient_scheme is leading out of the admissible domain, 
+      # this is a top-chart and we have to reconstruct from below.
```
simonbrandhorst, line 631: what is the admissible domain? I know what an admissible chart is. Is it something else?

src/Schemes/Covering/Objects/Methods.jl
```diff
@@ -226,3 +226,84 @@ function Base.show(io::IO, C::Covering)
...
+  #TODO: How to inherit the glueings?
+  phi = CoveringMorphism(E, C, map_dict_C, check=false)
+  psi = CoveringMorphism(E, D, map_dict_D, check=false)
+  return E, phi, psi
+end
+
+function some_ancestor_in(L::Vector, U::AbsSpec)
+  return some_ancestor(x->any(y->(y===x), L), U)
+end
+
```
simonbrandhorst: comment / remove / separate pr

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -0,0 +1,255 @@
...
+generator of ``ℐ(U)`` when ``V`` is the `i-1`-st canonical chart 
+of the local blowup over ``U``.
+"""
+function exceptional_divisor(p::BlowupMorphism)
+  if !isdefined(p, :exceptional_divisor)
+    error("exceptional divisor needs to be cached during construction")
+    Y = domain(p)
+    pr = projection(p)
+    pr_cov = covering_morphism(pr)
+    IC = center(p)
```
afkafkafk13: .... (and some more lines) 
Am I mistaken or is everything after the error statement in line 58 not reached here. Is this on purpose?

By the way, could we discuss on some of this unreached code at some point.

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: three saturation functions commented out: have they moved somewhere else and can bei removed?

src/Schemes/Covering/Objects/Methods.jl
```diff
@@ -226,3 +226,84 @@ function Base.show(io::IO, C::Covering)
...
+    for U in patches(E)
+      f, _ = _find_chart(U, C)
+      map_dict[U] = f
+    end
+    phi = CoveringMorphism(E, C, map_dict, check=false)
+    return E, phi, psi
+  end 
+  
+  error("case not implemented")
+  # We still need to adjust the code below.
```
afkafkafk13, line 272: You are using error("not implemented") to comment out drafts (here and in at least one other place). Do you really want that draft code in the PR?

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: I thought so too, before I found other instances of code made inactive by a error("not implemented"). It rather looks like missing cleanup to me.

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: pre_image_ideal returns the preimage of I under the canonical projection and localization (see src/Rings/mpolyquo-localizations.jl), so this is precisely the right thing -- but why is it here and not in some Rings/...,jl ? --> different PR?

experimental/Schemes/IdealSheaves.jl
afkafkafk13: @simonbrandhorst: Suggest a better name for a function finding a covering such that the (smooth) locally complete intersection is described by a complete intersection on each patch. This is a vital helper for many practical computations.

experimental/Schemes/CoveredProjectiveSchemes.jl
afkafkafk13: @simonbrandhorst: The essential getters (see above)  return data always present in the glueing, the function below needs to fill in previously lazy data, so they trigger computations and fill the other (lazy) entries. (Did I understand this correctly, Matthias?)

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: Ah right the Krull dimension counts the prime ideal chains starting in I: I< P1 ... < Pn < R
and so quotienting by I is okay. Thx for the reminder :-).

experimental/Schemes/IdealSheaves.jl
simonbrandhorst: `complete_intersection_cover(I::IdealSheaf)`?

simonbrandhorst: Maybe some consistency is in order here as this pattern shows up in several places. 
I suggest the prefix `as_locally_bla` for such things e.g.: `as_locally_complete_intersection`, `as_locally_free_module` 
`as_locally_principal_ideal`. The drawback is that it hides that the functions actually return a cover. But at least it is consistent.  In any case I would strongly argue for a  consistent naming. (whatever it turns out to be)

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: p.exceptional_divisor seems to be set upon creation of p, while the code here is deprecated (error-statement in line 58). This needs to be cleaned up.
Stupid question: How are multiplicities of excpetional divisors in total transforms of subschemes handled?

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: Indeed. Looks like I forgot to remove my TODO.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: In which sense? It's an `EffectiveCartierDivisor`. If you convert this to a `WeilDivisor`, it will have multiplicities in general. But I don't know how to extract the multiplicity from an `EffectiveCartierDivisor` directly. 

What precisely did you have in mind with your question?

afkafkafk13: Let's talk about this in person later.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: No, no typo. As it says, the computation of the divisor has been shifted to another place. I kept the other code behind the error message for the moment in case I still wanted something to draw from. But now this is just one of the sanity checks that you proposed the other day.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: Depending on where they belong, this might be a delicate thing to do. Other people working there might complain about merge conflicts. But I'll look into that.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: This `note` was deprecated. Thanks for catching.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: The internal check. I first leave them on during development to assure that bugs are caught. I just forgot to remove the comment.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: You mean with markdown, or what? I mean: Otherwise, it's written there what we do.

experimental/Schemes/CoveredProjectiveSchemes.jl
HechtiDerLachs: Yes, precisely.

experimental/Schemes/CoveredProjectiveSchemes.jl
HechtiDerLachs: No. Because now we have `RingFlattening` and the complicated detour has become superfluous.

experimental/Schemes/Types.jl
HechtiDerLachs: Because it is taken care of in the respective constructor. This was unacceptably slow and forced computation of glueings for almost no reason.

experimental/Schemes/Types.jl
HechtiDerLachs: An admissible chart is one for which `is_open_func(F)(U)` returns `true`.  This then falls into the general lookup and construction pattern documented already for the modules.

src/Schemes/Covering/Objects/Methods.jl
HechtiDerLachs: Yes, why not? In this case, I started implementing a potentially horribly complicated function from which we only need some very special cases up to now. If your demand goes beyond the implemented levels, you are kindly informed about this. Precisely what we talked about two weeks ago and, I think, what even you suggested to do.

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: Yep.  I like to ask `?bla`  in the repl.
Doesn't hurt to replace the # by triple quotes

experimental/Schemes/Types.jl
simonbrandhorst: Yes. I know that. But then call it an admissible chart and not an admissible domain.
The term "admissible domain" is not yet defined.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: Yes, now I remember talking about this. I think, it's done by now.

HechtiDerLachs:
Up to now tests fail for reasons beyond our reach, right? Who's doing the `gmodule`s?

experimental/Schemes/Types.jl
HechtiDerLachs: A "chart" is a patch of the `default_covering` of a scheme. Here "domain" is referring rather colloquially to the "domain of all admissible input". But I agree that comment has not been very well formulated.

simonbrandhorst:
The failing doctests are our fault.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -10,11 +10,104 @@ export strict_transform, total_transform, weak_transform, controlled_transform
...
+    )
+
+Produce a container `pg` to host a non-computed `ProjectiveGlueing` of ``X`` with ``Y``.
+
+The arguments consist of 
+
+ * the patches ``X`` and ``Y`` to be glued;
+ * a glueing `BG` of the `base_scheme`s of ``X`` and ``Y`` over which the 
+   `ProjectiveGlueing` to be computed sits;
+ * a function `compute_function` which takes a single single argument `glueing_data`
```
simonbrandhorst:
```suggestion
 * a function `compute_function` which takes a single argument `glueing_data` 
```

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -10,11 +10,104 @@ export strict_transform, total_transform, weak_transform, controlled_transform
...
+
+Produce a container `pg` to host a non-computed `ProjectiveGlueing` of ``X`` with ``Y``.
+
+The arguments consist of 
+
+ * the patches ``X`` and ``Y`` to be glued;
+ * a glueing `BG` of the `base_scheme`s of ``X`` and ``Y`` over which the 
+   `ProjectiveGlueing` to be computed sits;
+ * a function `compute_function` which takes a single single argument `glueing_data` 
+   of arbitrary type and actually carries out the computation;
```
simonbrandhorst, line 43: Does the compute function compute a single glueing of one chart 
or of all charts at once? Or is the output a ProjectiveGlueing? please document accordingly.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -395,129 +507,120 @@ end
 #  end
 #end
 
+# This is a sample for how to use LazyProjectiveGlueings. 
+# Originally, we probably had some body of a double for-loop iterating over 
+# pairs of patches (P, Q) that we need to glue. We take that body and move 
+# it to an external function (here _compute_projective_glueing). Then we 
+# go through all the local variables in the body of the for-loop which 
+# are needed for the actual computation and create a tailor-made struct to 
+# hous them. We add an extraction section in the beginning of the compute
```
simonbrandhorst:
```suggestion
# house them. We add an extraction section in the beginning of the compute 
```

simonbrandhorst (approved):

simonbrandhorst:
Since this is the last PR of a long line I would be glad about some more test coverage.
"45.50% of diff hit"

simonbrandhorst:
`strict_transform(p::BlowupMorphism, inc::CoveredClosedEmbedding)` would be worth a test.

HechtiDerLachs:
> Since this is the last PR of a long line I would be glad about some more test coverage.

That's not true. The last one is #1918 . But I'll have a look at the tests.

simonbrandhorst:
There is always a next PR :D

HechtiDerLachs:
> But I'll have a look at the tests.

Let's start by actually integrating those we already have. 

Edit: We're already busting testing time. Now comes the non-optimized K3 example. At some point, we probably have to do some triage there and remove some again.

> There is always a next PR :D

Yes, but not in this "line" you mentioned.

experimental/Schemes/CoveredProjectiveSchemes.jl
HechtiDerLachs: One glueing. Why should it be taken from within the double for-loop otherwise? I think, it's already spelled out.

simonbrandhorst: What are the charts? X and Y? but these are projective schemes? 
The docu is too fuzzy for me. Further when reading this I was not aware of any for loop whether double or not. 
Of course you know because you bisected the old code and sewed it into this "Frankenstein" ;-). 
However I don't know where the pieces come from.  Anyways if this is rather internal, I can live with never touching it again.

HechtiDerLachs:
We are in a bit of a messy situation here: I don't seem to get the `K3.jl` file running again. Whatever goes wrong there now is fixed in #1918 , but I don't remember in detail how I resolved the issues. 

What I can try: Rebase #1918 on this one and then one can hopefully take it from there.

simonbrandhorst:
Works for me as long as the tests will be there in the near future.

simonbrandhorst:
I set to auto merge. Just make sure the tests pass.

PR #1890, opened 2023-01-30
Morphisms between GAP groups and abelian groups

### LINKED ISSUE #1884: Morphisms between GAP groups and abelian groups
Would be great if we could make the following work:
```julia
julia> G = SymmetricGroup(3)
Sym( [ 1 .. 3 ] )

julia> A = abelian_group([2]);

julia> h = hom(G, A, [sign(g) == 1 ? zero(A) : A[1] for g in gens(G)])
```
(My end goal is to compute `kernel(h)`.)

CC: @fingolfin @ThomasBreuer

ThomasBreuer:
resolves  #1884

- added type `GroupIsomorphismFromFunc` (wrapping `MapFromFunc`), with methods analogous to those for `MapFromFunc`
- added `hom(G::GAPGroup, A::GrpAbFinGen[, gensG::Vector], imgs::Vector)`, returning an `AbstractAlgebra.Generic.CompositeMap`
- added `kernel(comp::AbstractAlgebra.Generic.CompositeMap{T, GrpAbFinGen})` for `T <: GAPGroup`, delegating to its two maps
- added `kernel(f::GroupIsomorphismFromFunc)`, returning the trivial subgroup of `domain(f)`

The idea behind the proposed changes is that `MapFromFunc` objects are suitable for mapping individual elements, but they do not know about structural properties such as injectivity and surjectivity.
W.r.t. functions such as `kernel` (or a function that asks for the image or peimage of a subgroup), the map has to know whether it respects addition/multiplication.

(Probably this approach is not general enough.)

### DIFF
--- a/src/Groups/homomorphisms.jl
+++ b/src/Groups/homomorphisms.jl
@@ -637,10 +637,10 @@ end
 ####
 mutable struct GroupIsomorphismFromFunc{R, T} <: Map{R, T, Hecke.HeckeMap, MapFromFunc}
     map::MapFromFunc{R, T}
+end
 
-    function GroupIsomorphismFromFunc{R, T}(f, g, D::R, C::T) where {R, T}
-      return new(MapFromFunc(f, g, D, C))
-    end
+function GroupIsomorphismFromFunc{R, T}(f, g, D::R, C::T) where {R, T}
+  return GroupIsomorphismFromFunc{R, T}(MapFromFunc(f, g, D, C))
 end
 
 function GroupIsomorphismFromFunc(f, g, D, C)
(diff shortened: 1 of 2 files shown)


thofma:
Looks good to me. Not sure if the `...Isomorphism` scales to other situations, but I am happy with it.

(As an alternative one could have done something like

```julia
mutable struct MorFromGAPGroupToAbGroup{T}
  G::T
  A::GrpAbFinGen
  ...
```
)

thofma (approved):

ThomasBreuer:
Perhaps we should discuss the alternatives; tomorrow afternoon in gather.town?

thofma:
I wrote you a message on slack.

thofma:
We will go with this for now and will have to think about the overall design and challenges later.

PR #1892, opened 2023-01-31
Revision of quotient rings

ederc:
As discussed with @wdecker this PR is a revision of quotient ring business. Main changes are:

- We only allow a fixed monomial ordering per quotient.
- When generating `R/I` we directly compute the Gröbner basis of `I` in order efficiently work in those rings.
- If an operation has as a side effect the simplification of an object in `R/I` we directly cache this simplification.


src/Rings/MPolyQuo.jl
```diff
@@ -126,23 +151,34 @@ Quotient of Multivariate Polynomial Ring in x, y, z over Rational Field by ideal
...
-# TODO rename oscar_assure to oscar_assure!
 function oscar_assure(a::MPolyQuoIdeal)
-  isdefined(a, :I) && return
+  if isdefined(a.gens.gens, :O)
+      return a.gens.gens.O
+  end
   r = base_ring(base_ring(a))
-  a.I = ideal(r, r.(gens(a.SI)))
+  a.gens.O = [r(g) for g = gens(a.gens.S)]
 end
```
thofma: If it is there, we return `a.gens.gens.O`. If not we set `a.gens.O` and return `a.gens.O`. Is this on purpose?

src/Rings/MPolyQuo.jl
```diff
@@ -300,16 +382,17 @@ ideal(y)
...
+    singular_assure(a)
+    singular_assure(b)
+    return MPolyQuoIdeal(base_ring(a), Singular.quotient(a.gens.S, b.gens.S))
 end
 (::Colon)(a::MPolyQuoIdeal, b::MPolyQuoIdeal) = quotient(a, b)
 
 # TODO: replace by a more efficient method!
-@attr function is_prime(I::MPolyQuoIdeal)
-  return is_prime(saturated_ideal(I))
+@attr function isprime(I::MPolyQuoIdeal)
```
thofma: Why the change to `isprime`?

src/Rings/MPolyQuo.jl
```diff
@@ -804,7 +843,7 @@ end
 zero(Q::MPolyQuo) = Q(0)
 one(Q::MPolyQuo) = Q(1)
 
-function is_invertible_with_inverse(a::MPolyQuoElem)
+function isinvertible_with_inverse(a::MPolyQuoElem)
```
thofma: Why the change of name?

src/Rings/MPolyQuo.jl
```diff
@@ -830,10 +862,10 @@ function is_invertible_with_inverse(a::MPolyQuoElem)
   return false, a
 end
 
-is_unit(a::MPolyQuoElem) = is_invertible_with_inverse(a)[1]
+isunit(a::MPolyQuoElem) = isinvertible_with_inverse(a)[1]
```
thofma: Why change the name?

src/Rings/MPolyQuo.jl
```diff
@@ -937,10 +960,8 @@ end
 
 #TODO: find a more descriptive, meaningful name
 function _kbase(Q::MPolyQuo)
-  I  = Q.I
-  GI = groebner_assure(I)
-  singular_assure(GI)
-  s = Singular.kbase(GI.S)
+    G = Q.I.gb[Q.ordering].gens.S
+  s = Singular.kbase(G)
```
thofma, line 952: Indentation.

thofma (commented):
Why mix 2 space and 4 space indentation?

I have some concerns about the function `simplify(a)` changing the generators of `a`. I agree that this does not change the ideal itself, but I think we always considered the given generators as part of the datum that should not be changed? I think people are lying on this, when working with `coordinates` for example:

```
julia> Qx, (x, y) = QQ["x", "y"]
(Multivariate Polynomial Ring in x, y over Rational Field, fmpq_mpoly[x, y])

julia> I = ideal(Qx, [x, y, x^2])
ideal(x, y, x^2)

julia> coordinates(x + y + x^2, I)
[x + 1   1   0]
```

Allowing functions to modify generators of ideals will break this. (This is an example for ideals in ordinary polynomial rings, but the interface should be uniform.)

wdecker:
@thofma  I think the docu text (which is old) might be misleading here: This simplify operation does NOT change the generators. It changes the representation of the generators in the original ring. So the coordinates argument does not apply.

src/Rings/MPolyQuo.jl
ederc: Reversed

src/Rings/MPolyQuo.jl
ederc: Reversed

src/Rings/MPolyQuo.jl
ederc: Reversed

src/Rings/MPolyQuo.jl
ederc: Fixed accordingly.

src/Rings/MPolyQuo.jl
ederc: Fixed, put the indention back to the old 2-spaces style. Should be unified now.

thofma:
I see, thanks for the explanation. I misread the documentation.

fieker (approved):

PR #1893, opened 2023-01-31
ToricVarieties: Bugfix in vanishing sets

HereAround:
It holds $h^1(P^2, L ) = 0$ for every line bundle $L$ on the projective space $P^2$. This extreme case that for every line bundle the cohomology in question vanishes, was not yet handeled correctly when computing vanishing sets in OSCAR. I provide a bug fix and have added a test that will detect this failure in the future.

(I also noticed a docstring, which appears in vacuum. I.e. the corresponding method does not exist in that particular file but was (?) moved to another file in the past. This obselete docstring should have no effect and so - presumably - can be removed. I have done so.)

### DIFF
--- a/src/ToricVarieties/cohomCalg/auxiliary.jl
+++ b/src/ToricVarieties/cohomCalg/auxiliary.jl
@@ -95,14 +95,14 @@ function contributing_denominators(variety::AbstractNormalToricVariety)
     # we execute cohomCalg with homogeneous variable names "x1", "x2" and so on
     # -> translate back into actually chosen variable names
     gens = Hecke.gens(cox_ring(variety))
-    for i in 1:length(contributing_monomials)
-        if contributing_monomials[i] == [""]
-            contributing_monomials[i] = String[]
+    for cm in contributing_monomials
+        if cm == [""]
+            empty!(cm)
         else
-            for j in 1:length(contributing_monomials[i])
-                m = contributing_monomials[i][j]
+            for j in 1:length(cm)
+                m = cm[j]
                 present_variables = [occursin("x$k", m) for k in 1:ngens(cox_ring(variety))]
-                contributing_monomials[i][j] = string(prod(gens[k]^present_variables[k] for k in 1:ngens(cox_ring(variety))))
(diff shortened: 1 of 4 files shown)


src/ToricVarieties/cohomCalg/auxiliary.jl
```diff
@@ -96,10 +96,14 @@ function contributing_denominators(variety::AbstractNormalToricVariety)
     # -> translate back into actually chosen variable names
     gens = Hecke.gens(cox_ring(variety))
     for i in 1:length(contributing_monomials)
-        for j in 1:length(contributing_monomials[i])
-            m = contributing_monomials[i][j]
-            present_variables = [occursin("x$k", m) for k in 1:ngens(cox_ring(variety))]
-            contributing_monomials[i][j] = string(prod(gens[k]^present_variables[k] for k in 1:ngens(cox_ring(variety))))
+        if contributing_monomials[i] == [""]
+            contributing_monomials[i] = String[]
```
fingolfin:
```suggestion
            empty!(contributing_monomials[i])
```

src/ToricVarieties/cohomCalg/auxiliary.jl
```diff
@@ -96,10 +96,14 @@ function contributing_denominators(variety::AbstractNormalToricVariety)
     # -> translate back into actually chosen variable names
     gens = Hecke.gens(cox_ring(variety))
     for i in 1:length(contributing_monomials)
```
fingolfin: off-topic, but: why don't you do this instead:
```suggestion
    for cm in contributing_monomials
```
and then replace each occurrence of `contributing_monomials[i]` inside the loop by just `cm`?

fingolfin (approved):

lkastner (approved):

lkastner:
Modulo comment by Max

HereAround:
Thank you for the suggestion @fingolfin. I added those changes in https://github.com/oscar-system/Oscar.jl/pull/1893/commits/a210944767a4ffe28123f422869954bdfbd4954c.

PR #1894, opened 2023-01-31
minor cleanups for singular_locus

afkafkafk13:
- removed some unreached lines (leftovers from previous changes)
- changed singular_locus_with_decomposition to always return empty list, if it is the empty set


afkafkafk13:
@HechtiDerLachs: My changes are minor, independent of the blow-up changes and make sense also outside the blow-up context. So I put them in a separate PR

src/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -635,20 +634,19 @@ function singular_locus_reduced(X::AbsSpec{<:Field, <:MPAnyQuoRing})
...
 
 # internal workhorse, not user-facing
 function _singular_locus_with_decomposition(X::AbsSpec{<:Field, <:MPAnyQuoRing}, reduced::Bool=true)
   I = saturated_ideal(modulus(OO(X)))
-  result = typeof(X)[]
+  empty = typeof(X)[]
+  result = empty
 
 # equidimensional decompositon to allow Jacobi criterion on each component
   P = []
```
HechtiDerLachs:
```suggestion
  P = Ideal[]
```

src/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -672,7 +670,7 @@ function _singular_locus_with_decomposition(X::AbsSpec{<:Field, <:MPAnyQuoRing},
     minvec = minors(M, n-d)
     J = ideal(R, minvec)
     JX = ideal(OO(X),minvec)
-    one(OO(X)) in JX && return result
+    one(OO(X)) in JX && return empty
```
HechtiDerLachs, line 673: This is the line that usually kills everything: If the ideal of minors is just too big, we have no chance to go through with this. But there's no way avoiding it, is there? I mean: The Jacobi-criterion is just what it is, right?

HechtiDerLachs (approved):
Removing some deprecated code. No change in algorithms or functionality. If the tests pass, this should be OK.

src/Schemes/AffineSchemes/Objects/Attributes.jl
afkafkafk13: Yes, indeed. this is a Groebner Basis computation which can easily get out of hand.
We can get around the Jacobian criterion completely in the application K3.jl, if we use a Hironaka-style invariant for resolution -- but for the time being (until the other line of development has reached this), there is no way around the Jacobian criterion.

simonbrandhorst (approved):

PR #1810, opened 2022-12-09
Algebraic solving updates

ederc:
Bumps `AlgebraicSolving.jl` to version 0.3.0 which provides `msolve_jll.jl` resp. `msolve` version 0.4.6.

1. Fixes #1743 (adds it also as a test).
2. Adds internal function `_rational_solutions` as discussed in #1744.

Ping: @thofma

PR #1897, opened 2023-02-01
Fixes tests for msolve v0.4.9

### LINKED ISSUE #1896: CI failing due to msolve update
CI is currently failing on master due to the update of msolve to 0.4.9 earlier today (cc @ederc), the same error triggers with the latest Oscar release:
```
solving: Test Failed at ~/.julia/packages/Oscar/Iei9b/test/Rings/solving-test.jl:11
  Expression: res[1] == sols
   Evaluated: Vector{fmpq}[[5416829397//8589934592, 2708414699//8589934592, -2844258330290649520990905062759917788583//21778071482940061661655974875633165533184], [1, 0, 0], [1945971683//8589934592, 972985841//8589934592, 744426424910260862653434112767010536665//2722258935367507707706996859454145691648], [2863311531//8589934592, 0, 3629678580490010276942662479272194255531//10889035741470030830827987437816582766592]] == Vector{fmpq}[[744483363399261433351//1180591620717411303424, 372241681699630716673//1180591620717411303424, -154187553040555781639//1180591620717411303424], [1, 0, 0], [71793683196126133110381699745//316912650057057350374175801344, 71793683196126133110381699745//633825300114114700748351602688, 173325283664805084153412401855//633825300114114700748351602688], [196765270119568550571//590295810358705651712, 1//590295810358705651712, 196765270119568550571//590295810358705651712]]
Stacktrace:
 [1] macro expansion
   @ julia-1.8.5/share/julia/stdlib/v1.8/Test/src/Test.jl:464 [inlined]
 [2] macro expansion
   @ ~/.julia/packages/Oscar/Iei9b/test/Rings/solving-test.jl:11 [inlined]
 [3] macro expansion
   @ julia-1.8.5/share/julia/stdlib/v1.8/Test/src/Test.jl:1363 [inlined]
 [4] top-level scope
   @ ~/.julia/packages/Oscar/Iei9b/test/Rings/solving-test.jl:2
Test Summary: | Pass  Fail  Total  Time
solving       |   10     1     11  2.4s
```

Maybe we can change the testcase so that it only compares these rational approximations (I am somewhat guessing here) up to some given precision?

ederc:
msolve v0.4.9 has a new internal dynamic precision adjustment if the default or given precision is not sufficient to solve correctly. Thus the tests need to be adjusted correspondingly.

Fixes issue #1896 .


fingolfin:
If I understand @benlorenz right on #1896 then this also affects the current OSCAR *release*.

But what is the effect? From the description here it sounds as if it just solves more things correctly? As in, the tests relied on some incorrect (?) behavior?

Then this fix here could be OK.

Or does it break something for real?, in the last releast? Then it is a semver violation in msolve, and strictly speaking that msolve version ought to be yanked from the registry and re-release as 0.5.0

fingolfin:
Also, it seems to me that msolve needs to get a CI setup similar to that which we use for Singular.jl, GAP.jl, Nemo.jl etc etc, where we test changes also against the Oscar test suite (for Oscar master resp. the last Oscar release) to detect early if a change made there would require changes on the Oscar side (and hence would warrant a semver change)

benlorenz:
I am not aware of any failure other than these numbers being slightly off, but this does affect the release as well, which for example would break julia's PkgEval for Oscar (if it weren't failing anyway most of the time because our tests take too long).

I am still not really sure if comparing these large fractions exactly is a good idea for stability. Is this a best possible approximation with the given precision and is this really stable or can this change again slightly in future updates?

Would it make sense to check it like this instead:
```julia
prec=32
for g in gens(I)
   for sol in real_solutions(I; precision=prec)[1]
      @test abs(evaluate(g, sol)) < fmpq(2)^-prec
   end
end
```

For the doctests a simpler [example](https://github.com/oscar-system/Oscar.jl/pull/1810/files#diff-c9c99755dd53d2a2fea5b36f750b0b2cbf69ea67eb134142b37934ce44b20f7fR35-R39) like from #1810 would work, or something else where we have an exact solution?

afkafkafk13:
My opinion might not be relevant here, but to me (as someone not directly involved in the use of msolve) the situation looks as follows:
- we use some '3rd party'-package (i.e. not directly in Oscar) for solving and rely on its output
- the testing example in question has been chosen to show the full power of the tool, not to test correct interfacing to it

Would it make sense to have 
a) some testing example with exact fractions to test interfacing to msolve for this function for something which does not have precision issues on any sensible configuration (no prescribed precision, exact solution of simple, but not trivial structure)
b) some testing example using the full power, but using prescribed precision and only testing up to it?

ederc:
> I am not aware of any failure other than these numbers being slightly off, but this does affect the release as well, which for example would break julia's PkgEval for Oscar (if it weren't failing anyway most of the time because our tests take too long).
> 
> I am still not really sure if comparing these large fractions exactly is a good idea for stability. Is this a best possible approximation with the given precision and is this really stable or can this change again slightly in future updates?
> 
> Would it make sense to check it like this instead:
> 
> ```julia
> prec=32
> for g in gens(I)
>    for sol in real_solutions(I; precision=prec)[1]
>       @test abs(evaluate(g, sol)) < fmpq(2)^-prec
>    end
> end
> ```
> 
> For the doctests a simpler [example](https://github.com/oscar-system/Oscar.jl/pull/1810/files#diff-c9c99755dd53d2a2fea5b36f750b0b2cbf69ea67eb134142b37934ce44b20f7fR35-R39) like from #1810 would work, or something else where we have an exact solution?

I took both of your sugggestions with slight changes to fit to msolve's behaviour.

ederc:
> If I understand @benlorenz right on #1896 then this also affects the current OSCAR _release_.
> 
> But what is the effect? From the description here it sounds as if it just solves more things correctly? As in, the tests relied on some incorrect (?) behavior?
> 
> Then this fix here could be OK.
> 
> Or does it break something for real?, in the last releast? Then it is a semver violation in msolve, and strictly speaking that msolve version ought to be yanked from the registry and re-release as 0.5.0

Nothing breaks, it is just a different way of doing the real root isolation, so you get correct (in the sense of precision) but maybe slightly different results. There is no semver violation from my point of view, it is just that the tests were not good. I changed this now as suggested by @benlorenz: From msolve's point of view we can assume that the precision of the results must be correct, but we cannot enforce a specific representation of these results w.r.t. the given precision.

test/Rings/solving-test.jl
```diff
@@ -6,8 +6,16 @@
...
+    for p in prec
+        res = real_solutions(I, precision=p)
+        sols = res[1]
+        for g in gens(I)
+            for s in sols
+                @test abs(evaluate(g, s)) < fmpq(2)^(-p+1)
+            end
+        end
+    end
     @test res[1] == sols
```
afkafkafk13: outdated line? you are setting sols to res[1] just above it inside the new loop.

afkafkafk13 (approved):
Looks fine to me now: 
- both exact functionality and handling of precision are tested 
- tests do no longer seem prone to permitted variations inside msolve.

benlorenz (approved):
Thank you very much for the quick fixes!

PR #1895, opened 2023-01-31
fix typo/ julia change: ' -> transpose

fieker:

Issue #1888, opened 2023-01-30
frequent CI failures with Julia nightly and 1.9 on macOS due to timeout


fingolfin:
Looking at e.g. < https://github.com/oscar-system/Oscar.jl/actions/runs/4026539430> it exceeds our self-imposed 2 hour timeout limit.

In contrast, Julia nightly on Linux runs the test suite in ~1 hour; while Julia 1.6 on macOS require roughly 1 hour 25 minutes (which is about the same as Julia 1.6 on Linux in that CI run).

- [this was the last "good" CI run on master](https://github.com/oscar-system/Oscar.jl/actions/runs/3985968642/jobs/6833977117) where it took ~1 hour on macOS (again similar to Ubuntu)
- [first bad CI run on master](https://github.com/oscar-system/Oscar.jl/actions/runs/4007934647/jobs/6881451218)

So the failure started with our commit c6ece85768abdbfac416a23fdf0155861ec61b9f but due to the nature of the failure, I don't think that commit is at fault, it seems more likely that a change on the Julia side is responsible?

The last "good" run used
> Julia Version 1.10.0-DEV.418
> Commit 7630606f16a (2023-01-22 05:37 UTC)

The first "bad" run used
> Julia Version 1.10.0-DEV.430
> Commit 9b1ffbbe1bc (2023-01-25 12:52 UTC)

Changes on the Julia side: https://github.com/JuliaLang/julia/compare/7630606f16a...9b1ffbbe1bc

Of course there could also be changes in one of our dependencies (one could try to extract this from the CI logs above). It might also be interesting to find out which tests got slower... A quick check shows that e.g. the `AbelianClousre` (sic!) went from 52s to 81.3s and `MPolyAnyMap/MPolyRing` went from  4m04.7s to 6m47.2s. But why, and why only on macOS + Julia nightly?

benlorenz:
I just thought I could re-run the last good job on github actions to see if the newer julia version would cause this to timeout as well but this has installed an even older julia version now:
```
Julia Version 1.10.0-DEV.387
Commit e1fc4824b92 (2023-01-18 03:54 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.5.0)
  CPU: 3 × Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, ivybridge)
[...]

PR #1909, opened 2023-02-08
ToricVarieties: Several bugfixes

HereAround:
Frederik Witt reported the following bug to me:

```julia
|julia> H2=hirzebruch_surface(2);

|julia> transpose(matrix(ZZ, rays(H2)))
[1   0   -1    0]
[0   1    2   -1]

|julia> matrix(map_from_character_lattice_to_torusinvariant_weil_divisor_group(H2))
[0   1    2   -1]
[1   0   -1    0]
```
The two matrices should be identical, but apparently are not. The rows are exchanged.

In the constructor for the Hirzebruch surface, an attribute was set incorrectly. This is the root cause for this behavior. This PR fixes this in the constructor for the projective space, Hirzebruch and del Pezzo surfaces.

While investigating this issue, I also noticed a similar glitch. For a smooth toric variety, we proceed as follows:
1. Set `map_from_torusinvariant_weil_divisor_group_to_class_group`.
2. Set `map_from_cartier_divisor_group_to_picard_group` as being identical to `map_from_torusinvariant_weil_divisor_group_to_class_group`.
 
It then follows, that Picard group and Class group of the smooth variety in question are identical (as in are the exact same object in the memory of the computer). To me this seems odd, even though Picard and Class group are isomorphic in such a setting. So I have modified the code such that they are isomorphic but not identical.

Another bug that Frederik Witt reported was the following:
```julia
|julia> P2 = projective_space(NormalToricVariety, 2);

|julia> td = ToricDivisor(P2,[1,-1,0]);

|julia> is_effective(td)
true
```
A toric Weil divisor is effective iff all of its coefficients are non-negative. So `td` is **not** effective, in contrast to the output by OSCAR. I have done the following:
- I modified the existing `is_effective` function of a toric divisor: It now checks if all coefficients are non-negative. If so, it returns `true` and otherwise `false`.
- I have added a new function `is_linearly_equivalent_to_effective_toric_divsior`: This constructs the corresponding toric divisor in `Polymake` and then checks if this divisor is linearly equivalent to an effective toric divisor. (This is the behavior of `is_effective` before this PR.)
- I have added a new function `is_effective` for toric divisor classes: It finds a toric divisor in the divisor class at hand and then checks if this divisor is linearly equivalent to an effective toric divisor.

Tests and documentation for the new functions have been added. Those for the old functions have been modified.


src/ToricVarieties/ToricDivisorClasses/properties.jl
```diff
@@ -1,2 +1,30 @@
...
+
+julia> tdc2 = ToricDivisorClass(P2, [-1])
+A divisor class on a normal toric variety
+
+julia> is_effective(tdc2)
+false
+```
+"""
+@attr Bool is_effective(tdc::ToricDivisorClass) = is_effective(toric_divisor(tdc))
+export is_effective
```
fingolfin:
```suggestion
export is_effective

```
Please end files with a newline (if you configure your editor to honor our `.editorconfig` this should be automatic, see <https://editorconfig.org>).

src/ToricVarieties/ToricDivisors/properties.jl
```diff
@@ -67,24 +67,55 @@ export is_basepoint_free
...
+
+julia> td2 = ToricDivisor(P2, [1,2,3])
+A torus-invariant, non-prime divisor on a normal toric variety
+
+julia> is_effective(td2)
 true
 ```
 """
-@attr Bool is_effective(td::ToricDivisor) = pm_tdivisor(td).EFFECTIVE
+@attr Bool is_effective(td::ToricDivisor) = all(c -> (c >= 0), coefficients(td))
```
fingolfin, line 91: So what did the old code compute then?!

test/ToricVarieties/divisor_classes.jl
```diff
@@ -2,19 +2,24 @@
...
+    
     @testset "Basic properties" begin
         @test is_trivial(toric_divisor(DC2)) == false
         @test rank(parent(divisor_class(DC2))) == 2
         @test dim(toric_variety(DC2)) == 2
+        @test is_effective(DC7) == true
+        @test is_effective(DC8) == false
     end
-
+
```
fingolfin:
```suggestion

```

fingolfin (commented):
Looks good to me, a few very minor suggestions

src/ToricVarieties/ToricDivisors/properties.jl
HereAround: It checked whether the toric divisor `td` is linearly equivalent to an effective divisor. In other words, it checked whether there exists a principal divisor `p` s.t. `td + p` is effective.

src/ToricVarieties/ToricDivisorClasses/properties.jl
HereAround: Ok. Thank you for the pointer. I shall honor that in the future/configure my editor accordingly.

HereAround:
The error in the required `tests / test (~1.8.0-0, x64, ubuntu-latest) (pull_request)` stems (if I read the log correctly) from the following:

`Examples.factor_absolute: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/ModStdQt-test.jl:17`

For `tests / test (~1.9.0-0, x64, ubuntu-latest) (pull_request)`, the failure seems to stem from

`Experimental LocalH2: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/gmodule-test.jl:30`

Both files have not been touched in this PR and all tests (except for the time-out on macOS - cf. https://github.com/oscar-system/Oscar.jl/issues/1888) succeeded yesterday. I am not sure why I am seeing this error.

Any suggestions @thofma and @fieker? Is this possibly related to https://github.com/oscar-system/Oscar.jl/issues/1912?

benlorenz:
> `Examples.factor_absolute: Error During Test at ...`

The `factor_absolute` failure comes from [this PR](https://github.com/oscar-system/Oscar.jl/pull/1895) that was merged despite this failing (but it doesn't always fail).

> `Experimental LocalH2: Error During Test at ...`
> Any suggestions @thofma and @fieker? Is this possibly related to #1912?

The `Local H2` error looks very similar to the one I reported in #1912, yes.

Both errors will probably go away when re-running the tests.

HereAround:
> > `Examples.factor_absolute: Error During Test at ...`
> 
> The `factor_absolute` failure comes from [this PR](https://github.com/oscar-system/Oscar.jl/pull/1895) that was merged despite this failing (but it doesn't always fail).
> 
> > `Experimental LocalH2: Error During Test at ...`
> > Any suggestions @thofma and @fieker? Is this possibly related to #1912?
> 
> The `Local H2` error looks very similar to the one I reported in #1912, yes.
> 
> Both errors will probably go away when re-running the tests.

Thank you for elaborating. Let me rerun the tests...

benlorenz:
> Let me rerun the tests...

A quick tip for the next time: Instead of re-opening the ticket you can also go to the `Checks` tab, select the `Run tests` job on the left, then on the top-right `Cancel` (if e.g. nightly is still running) and finally `Re-run` just the failed jobs. This way you can just avoid running all of them again.
In that view you can also re-run single workflows by hovering over the job and clicking that refresh icon that will appear, see [here](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs#re-running-a-specific-job-in-a-workflow). (This works only once all jobs have finished)

PS: I have re-run way too many jobs to debug that nightly issue...

HereAround:
> > Let me rerun the tests...
> 
> A quick tip for the next time: Instead of re-opening the ticket you can also go to the `Checks` tab, select the `Run tests` job on the left, then on the top-right `Cancel` (if e.g. nightly is still running) and finally `Re-run` just the failed jobs. This way you can just avoid running all of them again. In that view you can also re-run single workflows by hovering over the job and clicking that refresh icon that will appear, see [here](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs#re-running-a-specific-job-in-a-workflow). (This works only once all jobs have finished)
> 
> PS: I have re-run way too many jobs to debug that nightly issue...

Thank you for the pointer. I shall remember that now. Indeed, this is the better option. (I actually recalled that option, but could not find the corresponding button a few minutes ago... I must have overlooked it. Thank you for telling me where to look.)

HereAround:
Indeed, after rerunning the tests, all checks pass (other than the time-out for the macOS test). So this is ready for review @lkastner and @fingolfin . Thank you!

src/ToricVarieties/ToricDivisors/properties.jl
```diff
@@ -67,24 +67,55 @@ export is_basepoint_free
...
+
+julia> is_effective(td)
+false
+
+julia> is_linearly_equivalent_to_effective_toric_divisor(td)
+true
+```
+"""
+@attr Bool is_linearly_equivalent_to_effective_toric_divisor(td::ToricDivisor) = pm_tdivisor(td).EFFECTIVE
+export is_linearly_equivalent_to_effective_toric_divisor
```
lkastner, line 116: Do we really want this? Can't one just use the class construction? I mean we could end up with tons of `is_linearly_equivalent_to_*` functions. Is this something we want to avoid or would it be ok?

lkastner (approved):
Presetting all those properties for e.g. Hirzebruch surfaces always makes me a bit nervous, since it essentially makes it useless to have tests on these. I am a bit afraid that at some point we will run into situations where we have a bug, but it won't manifest for such standard constructions, even though it should, since we are assigning so many properties manually. Is it just me? Is there some kind of OSCAR policy regarding this?

Aside from that, great work.

src/ToricVarieties/ToricDivisors/properties.jl
HereAround: I thought this might be convenient, but on second thought you might be right. As I see it (now), we could remove this.

HereAround:
> Presetting all those properties for e.g. Hirzebruch surfaces always makes me a bit nervous, since it essentially makes it useless to have tests on these. I am a bit afraid that at some point we will run into situations where we have a bug, but it won't manifest for such standard constructions, even though it should, since we are assigning so many properties manually. Is it just me? Is there some kind of OSCAR policy regarding this?
> 
> Aside from that, great work.

I agree that this could happen. The first bug above was sort-of similar: Issue for the default/standard constructor `hirzebruch_surface()` and no issue if you constructed it by hand from a fan.

About feeling uneasy: Personally, I am not too worried. Originally, the intention was to preset properties for speed-up. But then, my feeling is that we do not gain too much from this (but I have not done any benchmarking) and so would be happy to remove them.

PR #1916, opened 2023-02-09
PolyhedralGeometry: add {load,save}_{lp,mps} to save (MixedInteger)LinearPrograms

### LINKED ISSUE #1852: Save MILP to .lp file
We can create a MixedIntegerLinearProgram object in Oscar by
`milp = MixedIntegerLinearProgram(Polytope, objective_function; convention = :min)`.

If I want to save this object as a standard .lp file, currently I have to go through a rather sinuous way, i.e., by running
`Polymake.polytope.poly2lp(Oscar.pm_object(feasible_region(milp)),milp.polymake_milp,0,"filename.txt")`. 
However, this line of code is not ideal. It needs to compute the feasible region of a MILP so it can be slow. And if the MILP is infeasible, this code results in an error.

I wonder whether it is possible to create a more direct way of saving LP and MILP objects as .lp files. Thanks!

@benlorenz

benlorenz:
these formats are used by many commercial LP and ILP solvers


benlorenz:
Note that currently loading an mps file will always return a maximization problem as the objective sense is not stored in the mps file (and does not seem to be a standard how to handle this).
Maybe we could add some functions to our LP and MILP objects to switch the convention (both changing it or creating a copy should be fine as the polymake object is also not really fixed apart from an attachment).

lkastner (approved):

antonydellavecchia (approved):

PR #1918, opened 2023-02-09
Concrete intersection theory

HechtiDerLachs:
This is a preview on the functionality required for the investigation of exceptional divisors on resolutions of singularities.


HechtiDerLachs:
New contributions compared to #1887 start from [c7e433f](https://github.com/oscar-system/Oscar.jl/pull/1918/commits/c7e433f522096cc92ff315f22d2ee3fbe96d3f30).

HechtiDerLachs:
Up to the self-intersection you can now get to the automatically generated intersection matrix. See `K3.jl`. 

I'm a bit worried because we only have 8 divisors, not 9. Probably there is still some non-transverse intersection that we need to also blow up? Also, the intersection multiplicity of two divisors is either 0 or 2, but never 1. That's wrong, isn't it?

For me, the script runs for about 15 minutes. 

Edit: Sorry! I was looking at the extended Dynkin diagram. 8 divisors is the correct number, then.

HechtiDerLachs:
@afkafkafk13 : @simonbrandhorst and I have been discussing this last friday, but didn't reach a conclusion. 

According to @simonbrandhorst , the intersection matrix of the exceptional divisors is obtained by intersecting the "reduced" exceptional divisors. Now some of them have higher multiplicities. And that's also the reason why we get the entries equal to `2` off the diagonal. 

Problem is: I know how to make sense of a reduced WeilDivisor. This should probably just be called `support` or something. But what is a "reduced Cartier divisor"? We can not intersect two Weil divisors, because our resolution is only proper over the base, but not compact. To produce a reduced Cartier divisor, I would need something like a reduced equation. But we don't have `gcd` in quotient rings, so taking `gcd` with the derivative won't work. 

How exactly is this intersection theoretic issue resolved in theory and what do you want to have implemented?

afkafkafk13:
> @afkafkafk13 : @simonbrandhorst and I have been discussing this last friday, but didn't reach a conclusion.
> 
> According to @simonbrandhorst , the intersection matrix of the exceptional divisors is obtained by intersecting the "reduced" exceptional divisors.

This is correct. It is obtained by intersection of the reduced irreducible curves on the resolved surface which correspond to the exceptional divisor of the blow-up process. By the simple normal crossing property of the resolution, we also know that (reasoning over the complex numbers) all intersections between the irreducible reduced curves will either be empty or of multiplicity 1; our goal is to just compute the dual graph of the resolution (with a vertex corresponding to an exceptional component and an edge corresponding to the intersection of the two components represented by its vertices) -- note that this graph need not be a tree in general. 
In the presence of QQ-irreducible and CC-reducible curves, we see the sum of the multipliities of the intersections (counted over CC) as the multiplicity counted over QQ and need to amend the computed matrix by introducing separate row/column pairs for the CC-components, if we are interested in the usual intersection matrix.
 
> How exactly is this intersection theoretic issue resolved in theory and what do you want to have implemented?

First of all: The user wants this for surfaces, not in general.
What the user expects, is the intersection matrix of the exceptional curves on the resolved surface over CC, because these are the intersection matrices that are usually considered in literature. The rare case that some really wants the intersection properties over the true groundfield QQ can be left for later, when the need arises.

Do not think too complicated in the sense of general intersection theory. We are in a very particular setting because of the origin of the divisors from a resolution of a surface.  If you pay attention to only use (QQ-)irreducible centers, you can rely on having  (QQ)-irreducible excpetional divisors from the individual blow-ups and you can keep track of the decompostion of the exceptional divisor of the whole process without decomposing. CC-reducible QQ-components can already be detected at the level of the respective center in the resolution process, if need be.

In general: Do not call this intersection theory, call it dual graphs of resolution -- to keep it well separated from the general case.

afkafkafk13:
One more remark: We are talking about CC-point-counting of a QQ-irreducible point (even in ambient_scheme), so there is no harm to be expected here.

HechtiDerLachs:
Thanks for the explanations. 

I still see some open points: 
 * We need to provide a method to check that a configuration of divisors is snc. The non-reduced divisors here are **not** snc, because they have non-trivial multiplicities. They might locally be of the form `x*y^2`. 
 * The problem of counting over QQ rather than CC does not play a role in this particular example. The multiplicities are determined as lengths of modules over localized rings at prime ideals. Does this give you the right thing in general? 
 * I called this "intersection theory" because I made an effort to provide the minimum requirements on the framework for that to host the particular special case we have here in a way that it has a future. This is a PR on the general pattern of intersection theory with your example of a resolution as a particular use-case for my proof of concept. There is no separation of these things here in this PR. Of course, we could implement special case by special case, but I am not. The later clean up of such messes tends to exceed the initial overhead you have when doing things right from the beginning. 

I don't see how to detect squares in `MPolyQuo`s. I mean: if our local equation for the union of divisors is `x*y^2` then how do we factor `y^2`? Abstractly we know that the local ring is factorial because it is smooth. But concretely, we have a localization of an `MPolyQuo` so we can not expect factorization to be there. How do we resolve this?

HechtiDerLachs:
A summary of today's discussion: 

We have special requirements for investigating resolutions of (isolated) surface singularities and harvesting the resolution graph. To this end, we plan to implement the following special functionality: 

 * A getter `exceptional_divisor_as_weil_divisor` to benefit from extra context in the conversion `EffectiveCartierDivisor -> WeilDivisor`. 
 * `strict_transform(p::BlowupMorphism, D::WeilDivisor)` to have something like "pullback" for them. Probably, this needs to be extended to `controlled_transform` and the like. 
 * For two *compact* `WeilDivisor`s `C` and `D` on a *smooth* surface in *sufficiently general position*, meaning the support of their intersection is zero-dimensional, we implement Proposition 1.4, p.360 from Hartshorne: The intersection multiplicity at the various points is just the length of their quotient algebras. In particular, this means we want to be able to intersect two `WeilDivisor`s in general position on a smooth surface. 
 * To count the local multiplicities we try to go through the charts, counting only contributions located at relative horizons, but avoiding the primary-decomposition-matching procedure. In particular this requires the `LazyGlueing`s involved to provide their `complement_equations` without computing the whole glueing.

HechtiDerLachs:
Rebased and with tests included. However, I assume that the tests run for too long, right now. So take this as a proof of concept. 

Maybe, we can once verify that all tests are running and then disable the `K3.jl`-file for the time being? I could then look into providing smaller tests when I'm back and the PR could eventually get merged in the meantime. This might also be good for the FTheory people to have a running prototype for intersection theory? (CC: @HereAround ).

experimental/Schemes/Sheaves.jl
```diff
@@ -77,7 +77,7 @@ production_func(F::PreSheafOnScheme) = F.production_func
 restriction_func(F::PreSheafOnScheme) = F.restriction_func
 
 ### Production and caching of the values of F on admissible open sets
-function (F::PreSheafOnScheme{<:Any, OpenType, OutputType})(U::T; cached::Bool=true, check::Bool=true) where {OpenType, OutputType, T<:OpenType}
+function (F::PreSheafOnScheme{<:Any, OpenType, OutputType})(U::T; cached::Bool=true, check::Bool=false) where {OpenType, OutputType, T<:OpenType}
```
HechtiDerLachs, line 80: This is a bit delicate. I had the impression that the code would just be running too slow if this was left on `true`. But already, it showed in other discussions that leaving out these sanity checks this immediately leads to difficult bugs. I don't know what's better at the moment.

experimental/Schemes/Sheaves.jl
```diff
@@ -98,7 +98,7 @@ For a `F` produce (and cache) the restriction map `F(U) → F(V)`.
 """
 function restriction_map(F::PreSheafOnScheme{<:Any, OpenType, OutputType, RestrictionType},
     U::Type1, V::Type2;
-    check::Bool=true
+    check::Bool=false
```
HechtiDerLachs, line 101: Same here.

src/Schemes/CoveredSchemes/Objects/Attributes.jl
```diff
@@ -178,7 +178,7 @@ end
   covering = (has_attribute(X, :simplified_covering) ? simplified_covering(X) : default_covering(X))
   for U in covering
     _, inc_sing = singular_locus(U)
-    D[U] = image_ideal(inc_sing)
+    D[U] = radical(image_ideal(inc_sing))
```
afkafkafk13, line 181: Yes, the radical definitely makes sense. 

Just another question (not a todo for this PR, but a '#TODO' to put in the code to not forget it in the future: Should we also provide ideal_sheaf_of_singular_locus_reduced in the future? For many (most) applications -- like the one of @simonbrandhorst has in mind -- this would suffice and it internally avoids the full primary decomposition and relies on the significantly cheaper computation of minimal associated primes.

experimental/Schemes/SimplifiedSpec.jl
```diff
@@ -119,6 +119,7 @@ function _flatten_open_subscheme(
       f
     end
   )
+  any(W->(W===U), patches(C)) && return iso # If U is already in C do nothing.
   has_ancestor(W->any(WW->(WW === W), patches(C)), U) || error("patch not found")
   W = ambient_scheme(U)
```
afkafkafk13, line 124: I seem to remember this added line from another PR. Has this one not been properly rebased yet?

experimental/Schemes/Sheaves.jl
simonbrandhorst: There is a difference between input checks and sanity checks. 
For sanity checks you use `@hassert` with the proper level. Then it is easy to turn them off. 
For input checks of user facing functions there is the `check=true` keyword.

src/Schemes/CoveredSchemes/Objects/Attributes.jl
simonbrandhorst: So you mean that would be faster than `reduced_scheme(singular_locus(X))?` 
I don't like these long names for functions which do everything at once. Usually it is more user friendly to have fewer functions. (unless there is a significant speed difference)

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -76,23 +76,44 @@ For a `BlowupMorphism` ``p : Y → X`` and a `CoveredClosedEmbedding`
 ``ι : Z ↪ X``, compute the strict transform ``Z'`` of ``Z`` along ``p`` and 
 return a triple ``(Z', j, π)`` containing the `CoveredClosedEmbedding` 
 ``j : Z' ↪ Y`` and the induced projection ``π : Z' → Z``.
+
+!!! note The projection is not yet implemented! Instead, `nothing` is returned 
+as the third item.
```
simonbrandhorst:
```suggestion
```

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -101,21 +122,94 @@ function strict_transform(p::BlowupMorphism, inc::CoveredClosedEmbedding)
...
+    if length(gens(pbJ))==0 
+        error("output of saturation invalid; faulty exceptional divisor?")
+    end
     ID[U] = pbJ
   end
 
-  I_trans = IdealSheaf(Y, ID, check=false)
-  inc_Z_trans = CoveredClosedEmbedding(Y, I_trans, covering=CY_simp, check=false)
-  inc_cov = covering_morphism(inc_Z_trans)
+  I_trans = IdealSheaf(Y, ID, check=false) # TODO: Set to false
```
simonbrandhorst:
```suggestion
  I_trans = IdealSheaf(Y, ID, check=false) 
```

test/Schemes/K3.jl
```diff
@@ -48,86 +49,180 @@
...
+    for j in 1:8
+      if j == i 
+        # avoid self intersection for now
+        A[i, i] = -ZZ(2)
+        continue
+      end
+      A[i, j] = integral(intersect(F[i], E[j]))
+    end
+  end
+  @test A == ZZ[-2 0 0 0 2 0 0 0; 0 -2 0 0 0 0 0 2; 0 0 -2 0 2 2 0 0; 0 0 0 -2 0 2 2 2; 2 0 2 0 -2 0 0 0; 0 0 2 2 0 -2 0 0; 0 0 0 2 0 0 -2 0; 0 2 0 2 0 0 0 -2]
```
simonbrandhorst:
```suggestion
  # Compute the intersection matrix:
  # A = zero(MatrixSpace(ZZ, 8, 8))
  # E = [E1res, E2res, E3res, E41res, E42res, E51res, E52res, E53res]
  # F = weil_divisor.(E)   # too slow to be a test
  # for i in 1:8
  #   for j in 1:8
  #    if j == i 
  #      # avoid self intersection for now
  #      A[i, i] = -ZZ(2)
  #      continue
  #    end
  #    A[i, j] = integral(intersect(F[i], E[j]))
  #  end
  #end
  #@test A == ZZ[-2 0 0 0 2 0 0 0; 0 -2 0 0 0 0 0 2; 0 0 -2 0 2 2 0 0; 0 0 0 -2 0 2 2 2; 2 0 2 0 -2 0 0 0; 0 0 2 2 0 -2 0 0; 0 0 0 2 0 0 -2 0; 0 2 0 2 0 0 0 -2]
```

simonbrandhorst (approved):

simonbrandhorst:
This still needs some work. But I think that it is okay for experimental.

src/Schemes/CoveredSchemes/Objects/Attributes.jl
afkafkafk13: No, these are two different things:
- here the radical is applied to the ideal sheaf of the singular locus
  (I sure hope that the routines are aware that in each chart the radical has already been computed and hence just assign a flag instead of recomputing the radical!!!)
- the other thing is that in the computation of the singular locus, we can make use of the fact that we want the singular locus of the corresponding reduced scheme, if this is the case.

As I see it, you are not dealing with non-reduced schemes most of the time -- nor am I -- so we need a way to make use of this (bypassing the more expensive part of equidimensional_decomposition_weak and only using equidimensional_decomposition_radical in the internal decomposition inevitably happening inside singular_locus).

HereAround:
> Rebased and with tests included. However, I assume that the tests run for too long, right now. So take this as a proof of concept.
> 
> Maybe, we can once verify that all tests are running and then disable the `K3.jl`-file for the time being? I could then look into providing smaller tests when I'm back and the PR could eventually get merged in the meantime. This might also be good for the FTheory people to have a running prototype for intersection theory? (CC: @HereAround ).

Thank you @HechtiDerLachs . We will definitely have a look at this. (I also cc @apturner).

PR #1924, opened 2023-02-10
fix trivia - and rename

fieker:


examples/PerfectPowers.jl
```diff
@@ -302,13 +316,12 @@ function ispower_bernstein(a::fmpz)
   c = [x for x = c if !isone(x)]
   #not sure still necessary
   @time c = Hecke.coprime_base(c) 
-  @show length(c), length(Set(c))
   k = [valuation(a, p) for p = c]
   @show no_p
   return gcd(k)*f
```
fingolfin, line 333: In the  base cases, this function seems to return a tuple `(n,x)` such that the `x^n == a`. But here a single integer (?) is returned? How does that fit together?

examples/PerfectPowers.jl
```diff
@@ -229,13 +229,27 @@ function _root_exact(a::fmpz, p::Int, extra_s::Int = 5, extra_w::Int = 1)
   return D
 end
 
-function ispower_bernstein(a::fmpz)
+function is_power_bernstein(a::fmpz)
```
fingolfin, line 244: Perhaps also add at least a trivial comment saying what this function does / what its return type is? Bonus points if it is a doc string with a jldoctest ;-)

fingolfin (approved):
The changes seems sensible, but I don't understand what `ispower_bernsteinz returns (see comments)

examples/PerfectPowers.jl
```diff
@@ -229,13 +229,30 @@ function _root_exact(a::fmpz, p::Int, extra_s::Int = 5, extra_w::Int = 1)
...
 #https://cr.yp.to/lineartime/powers2-20060914-ams.pdf
+  if isone(a)
+    return (0, a)
+  end
   k, a = remove(a, fmpz(2))
+  if isone(a)
+    return k
+    return (k, fmpz(2))
```
fingolfin: Huh, two in a row? Also, in line 235 it still returns a pair? I am confused :-)

PR #1927, opened 2023-02-12
Fix finite orthogonal group

simonbrandhorst:
The result was wrong for odd lattices. This should fix it.


thofma (approved):
If it works, it works.

PR #1931, opened 2023-02-13
NAIVE impl of exterior_algebra (via PBM alg)

JohnAAbbott:
Main question is whether the user interface to the constructor is as desired.
Speed is not great currently (but I'm working on it)

There are some tests in Oscar.jl/test/Experimental/ExteriorAlgebra-test.jl


experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
```diff
@@ -0,0 +1,95 @@
+export exterior_algebra_naive;  # constructor
+export ContainsDuplicates;   # export just for developing/testing
+
+# Auxiliary function -- worth keeping?
+
+function ContainsDuplicates(L)  # L must be indexable (1:n), and its elems must be sortable
```
HechtiDerLachs: Function names should be snake case; i.e. this should be `contains_duplicates`.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
```diff
@@ -0,0 +1,95 @@
...
+  n = length(L);
+  if (n < 2)  return false;  end;
```
HechtiDerLachs: Could also be 
```
n < 2 && return false
```

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
```diff
@@ -0,0 +1,95 @@
...
+  if (n == 2)  return (L[1] == L[2]);  end;
+  sorted = sort(L);
+  for i=1:n-1  if (sorted[i] == sorted[i+1])  return true;  end;  end;
+  return false;
```
HechtiDerLachs: Despite being correct, this is probably not the fastest way to do it. 
I would suggest something like 
```
for i in 1:n-1
  any(y->(y==L[i]), L[i+1:end]) && return true
end
return false
```

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
```diff
@@ -0,0 +1,95 @@
...
+julia> x2*x1
+-x1*x2
+
+julia> ExtAlg,(x,y) = exterior_algebra_naive(QQ, ["x","y"]);
+
+julia> y*x
+-x*y
+```
+"""
+function exterior_algebra_naive(K::Ring, NumVars::Int64;  var_prefix="e")
```
HechtiDerLachs:
```suggestion
function exterior_algebra_naive(K::Ring, NumVars::Int;  var_prefix="e")
```
I think it's good to be more generic here.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
```diff
@@ -0,0 +1,95 @@
...
+
+# exterior_algebra constructor:
+#  args should be underlying CoeffRing and number of indets
+#  optional keyword arg: indet_prefix (must be roman-alphabetic"
+
+
+# Returns 2 components: ExtAlg, list of the gens/variables in order (e1,..,en)
+# [[depends on what pbw_algebra returns!]]
+@doc Markdown.doc"""
+    exterior_algebra_naive(CoeffRing, NumVars)
```
HechtiDerLachs:
```suggestion
    exterior_algebra_naive(coeff_ring::Ring, num_vars::Int)
```
Why not annotate the types? Also, it might be necessary to match the docstring namings with the actual code below for the documenter to find it. But there I'm not sure.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
```diff
@@ -0,0 +1,95 @@
...
+  end;
+  return exterior_algebra_naive(K, (1:NumVars) .|> (k -> "$var_prefix$k"));
+end
+
+function exterior_algebra_naive(K::Ring, xs::Union{AbstractVector{<:AbstractString}, AbstractVector{Symbol}, AbstractVector{Char}})
+    fn_name = "exterior_algebra_naive";
+    NumVars = length(xs)
+    if (NumVars == 0)  throw(ArgumentError("$fn_name: (arg2) no variables/indeterminates given"));  end;
+    if (ContainsDuplicates(xs))  throw(ArgumentError("$fn_name: (arg2) variable names must be distinct"));  end;
+    R, indets = PolynomialRing(K, xs);  # ordering??
```
HechtiDerLachs: This is a polynomial ring in Oscar. Orderings are not part of their data structure officially. Internally, there might be something stored to fix the order of printing for elements, though.

HechtiDerLachs (commented):
Looks good! 

I left some minor comments where I thought, it might be helpful to get more aquinted with our conventions in julia/Oscar.

thofma:
Looks good so far. Would be best to follow https://docs.oscar-system.org/stable/DeveloperDocumentation/styleguide/, in particular
- the name `ContainsDuplicates` should be snake case (by the way, it would be shorter to do `length(unique(L)) == length(L)`,
- not explicitly mentioned, but for readability would be great to not nest blocks in one line, e.g.
```
for x in bl
  if y
    blub
  end
end
```
instead of
```
for x in bl; if y; blub; end; end;
```
- The `exterior_algebra` constructors should behave like the polynomial ring constructors, so no `var_name` keyword argument.
- Is there a specific reason for disallowing non-ascii variable names? I think we allow this everywhere else.
- No superfluous `;` 

(P.S. One can also do `&&` for short-circuit evaluation.)

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
JohnAAbbott: Thanks!  It seems that the Julia function `allunique` actually does what I need, so `ContainsDuplicates` will be deleted.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
JohnAAbbott: Thanks!  It is a question of style.  I prefer to write explicitly `if... end` because I find it easier to read.  I presume that compilation will produce the same code in both cases.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
JohnAAbbott: Not sure.  Assuming `sort` is "soft-linear" then sorting and checking adjacent elements should be "soft-linear" (at the cost of making copies of the vector elements.  Your loop appears at first sight to be quadratic (but maybe Julia can compile it cleverly), so ought to be slower when the vector contains enough elements...

JohnAAbbott: Anyway, thanks for the suggestion.  I shall take note of the `any` function which will surely be useful at some point.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
JohnAAbbott: Thanks! Good point!  I plan to accept this suggestion after my next push.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
JohnAAbbott: Thanks!  Yes, I can annotate the formal parameters in the docstring.
At the moment I have a (mild) preference for CamelCase for **non-exported** parts of the code -- simply because that is what I am used to.  If there are objections, I can adjust my preferred style.  If you have a persuasive argument why I should change my style, let me know... I'm willing to listen!

JohnAAbbott:
Thanks for the feedback!  It seems that the Julia function `allunique` can replace `ContainsDuplicates`.  I'm glad we agree that code readability is important (even if we might not agree on all the details).
Maybe you are right about allowing non-ascii names; it does seem to be an artificial restriction.

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
JohnAAbbott: Oops!  I thought this could be done automatically, but evidently not after pushing.  Sorry about that.  Anyway, I have adopted your suggestion -- thanks!

experimental/ExteriorAlgebra/ExteriorAlgebra_naive.jl
fieker: try `any(isequal(L[i]), L[i+1:end])` don't know about speed but more julia...
`any(isequal(a), b)` is equivalent to `any(x->x==a, b)`

JohnAAbbott: In the end I used the built in Julia function `Base.allunique` which seems to do what I needed.

wdecker (approved):

PR #1935, opened 2023-02-14
Integrating JuLie into Oscar

ClaudiaHeYun:
Three files from JuLie are included: partitions, schur_polynomials, tableaux. Their tests and documentations are included too. We stopped importing partitions from Hecke. Right now there is a problem with `YoungTableau`, which uses partitions from AbstractAlgebra. For example,
```
p = Partition([4,3,2])
yt = YoungTableau(p)
```
will create an error. This can be fixed by specifying `p = AbstractAlgebra.general.Partition([4,3,2])`. Eventually it would be nice to merge Tableau and YoungTableau.


HechtiDerLachs:
@micjoswig : Please, have a look whether this includes all the functionality you had asked for. 
@ulthiel: We did some small changes to the best of our knowledge to make your files coherent with the Oscar conventions. In particular, we modified the introductory comments in many files to give credit. Please, have a look and let us know whether you approve. Moreover, I will point out one more TODO in the code which I would like you to look at.

experimental/JuLie/schur_polynomials.jl
```diff
@@ -0,0 +1,304 @@
...
+  #end
+
+  n = length(x)
+  @assert n > 0 "number of variables must be > 0"
+  # TODO: The next line suggested that the case n<=0 is invalid input. But is this 
+  # really the case? I see some comment lines below which have probably been there 
+  # at some point to catch such boundary cases. 
+  R = x[1].parent # Multi-polynomialring
+  @assert all(y->parent(y)===R, x) "input elements must have the same parent"
+  S = base_ring(R) # Integer Ring
```
HechtiDerLachs, line 151: This is the TODO I was talking about: I added an assertion for `n>0`. @ulthiel: Is this correct?

ulthiel:
Excellent work. I will look into the code later (soon) but before I forget: Tom Schmit (who was my student) should be credited in the files schur_polynomials and tableaux. This seems to be missing in my repository, I don't know why. Could you please add this?

experimental/JuLie/schur_polynomials.jl
micjoswig: The JuLie functions that are made available in this pull request are precisely the ones that we need now. Well done!

Concerning the conflict with AbstractAlgebra, I would like to suggest to switch YoungTableau to work with JuLie's partitions.  Unless there are further ramifications.

micjoswig:
@HechtiDerLachs If I see this correctly the tests fail because there is a JuLie test file missing?

experimental/JuLie/schur_polynomials.jl
ClaudiaHeYun: I just noticed that there are other functions than `YoungTableau` that depend on partitions, for example, characters of irreducible representations of symmetric groups, see [here](https://docs.oscar-system.org/stable/AbstractAlgebra/ytabs/#Young-Diagrams-and-Young-Tableaux). But all of them come with AbstractAlgebra. I wonder whether AbstractAlgebra also wants to be a self-contained system and therefore is unwilling to adopt code from JuLie?

ClaudiaHeYun:
@ulthiel Yes! I will add Tom's name to the files.

HechtiDerLachs:
I fixed the tests [here](https://github.com/oscar-system/Oscar.jl/tree/julie_yun). However, something went wrong and I somehow couldn't push to this pr. I'll see tomorrow how to resolve this.

experimental/JuLie/schur_polynomials.jl
thofma: We can safely ignore AbstractAlgebra. Just put everything that is clashing with AbstractAlgebra into the exclude list for the import statements.

HechtiDerLachs: I think the printing for the Tableaus from AA is pretty nice. But making it work with the partitions from JuLie would require changes in AA. And we don't import JuLie into AA so the required types are not available there. I do not see how we can resolve this, so: yes, it seems there are further ramifications here.

test/runtests.jl
```diff
@@ -31,6 +31,7 @@ include("Experimental/ModStdQt-test.jl")
 include("Experimental/ModStdNF-test.jl")
 include("Experimental/MPolyRingSparse-test.jl")
 include("Experimental/MatrixGroups-test.jl")
+include("Experimental/JuLie_test.jl")
```
benlorenz:
```suggestion
include("Experimental/JuLie_tests.jl")
```

docs/src/Combinatorics/partitions.md
```diff
@@ -0,0 +1,36 @@
...
+    ascending_partitions
+```
+
+## Restricted partitions
+
+```@docs
+    partitions(::Integer, ::Integer, ::Integer, ::Integer)
+    partitions(::Integer, ::Integer)
+    num_partitions(::Integer, ::Integer)
+    partitions(::Array{Integer,1}, ::Integer, ::Array{Integer,1}, ::Integer)
```
fingolfin: Please don't use `Array{Integer,1}`, use `Vector{Integer}` instead (they are semantically the same, but the latter is the default name in code and when printing these types since Julia 1.6).

(This is not a blocker for this PR, it could be changed later, but it's good to be aware of this)

docs/src/Combinatorics/partitions.md
```diff
@@ -0,0 +1,36 @@
+# Partitions
+
+```@docs
+    Partition
+```
+
+## Unrestricted partitions
+
+```@docs
+    partitions(::Integer)
```
fingolfin: Meta-question: what is `Integer` here? The Julia type `Base.Integer`?

Can one also pass the OSCAR integer type `fmpz` to these functions?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -184,6 +184,53 @@ end
 
 However, as always, rules sometimes should be broken.
 
+## Optional arguments for parents of return values
+
+Several objects in Oscar have `parent`s, e.g. polynomials, group elements, ...
```
fingolfin: Please use `OSCAR` instead of `Oscar` whenever talking about the computer algebra system (while `Oscar` is reserved for the Julia package itself; but this is a subtle distinction, if in doubt, write `OSCAR`).

(Again, not a blocker, but I hope changing it won't be hard)

experimental/JuLie.jl
```diff
@@ -0,0 +1,14 @@
...
+import Nemo:libflint
+
+include("JuLie/partitions.jl")
+include("JuLie/schur_polynomials.jl")
+include("JuLie/tableaux.jl")
+
+end
+
+using .JuLie
+export Partition, partitions, ascending_partitions, dominates, conjugate, getindex_safe, num_partitions, schur_polynomial, Tableau, shape, semistandard_tableaux, is_standard, is_semistandard, standard_tableaux, schensted, hook_length, hook_lengths, num_standard_tableaux, reading_word, weight, bump!
```
fingolfin: Please ensure files end in a newline (e.g. by configuring your editor to honor our `.editorconfig`, see https://editorconfig.org).

Like all my other comments: not a blocker

experimental/JuLie/partitions.jl
```diff
@@ -0,0 +1,675 @@
...
+      # because here k could be zero and then we access
+      # a[k].
+      # That's why I do a while true and check k > 0 here.
+      if k == 0
+        break
+      else
+        x = a[k] + 1
+      end
+    end
+    return P
```
fingolfin, line 397: So if `alg` is an unsupported value, we just run into an error about P being unassigned here? Perhaps better to add an explicit `error` inside an `else` block above?

fingolfin (commented):
A few quick remarks. Unfortunately I am busy with other stuff right now, I can look some more tomorrow. Or else you merge and we address these later *shrug*

fingolfin:
Also, lest I forget this: Thank you very much! Overall it looks really good, my remarks are all just tiny nitpicks.

HechtiDerLachs:
I don't know what to do to make the tests run. If I run them locally, it's going just fine. But it seems we need to do something special due to our wrapping up things in a submodule. Can anybody help? 

(CC: @fieker , @fingolfin )

fingolfin:
@HechtiDerLachs and how do you run them locally? Using `Pkg.test`?

HechtiDerLachs:
> and how do you run them locally? Using `Pkg.test`?

`Oscar.test_module("Experimental/JuLie_tests", false)`

test/Experimental/JuLie/schur_polynomials.jl
```diff
@@ -0,0 +1,70 @@
+import .JuLie: schur_polynomial_cbf
```
fingolfin: This is where the error occurs. I don't see how this can work for you locally? Do you perhaps mean
```suggestion
import Oscar.JuLie: schur_polynomial_cbf
```

HechtiDerLachs: Alright. Honestly, I don't know what precisely either one of the two statements is doing. But let's try it out.

experimental/JuLie/schur_polynomials.jl
micjoswig: > I think the printing for the Tableaus from AA is pretty nice.

Printing the tableaux should not stop merging.  Adding this (later) is an easy exercise.  For once I would not even mind (mild) duplication of code; it's a few lines only.

micjoswig:
@HechtiDerLachs @ClaudiaHeYun please complete this work, which is almost done, by taking care of the relevant comments, such as: unimporting from AbstractAlgebra suggested by @thofma , adding that explicit error demanded by @fingolfin .

docs/src/Combinatorics/partitions.md
ClaudiaHeYun: Yes, it's `Base.Integer`. Passing `fmpz` results in an error. Should we do something about this?

fingolfin (changes requested):
Please remove the `.DS_Store` files. You may wish to consider adding those to a global git ignore file (e.g. I have this in my `~/.gitconfig`:
```
[core]
	excludesfile = ~/.config/git/ignore
```
and then in `~/.config/git/ignore` I have many things, including:
```
.DS_Store
*~
```

docs/src/Combinatorics/partitions.md
fingolfin: It would be great if `IntegerUnion` was accepted, i.e. both `fmpz` (= `ZZRingElem`) and `Base.Integer`

fingolfin: (But this is not a requirement for merging, just something that would be nice to have eventually -- if we don't do it in this PR, someone should open an issue to remind us to do it)

experimental/JuLie/tableaux.jl
```diff
@@ -0,0 +1,717 @@
...
+
+function schensted(sigma::Perm{T}) where T<:Integer
+  return schensted(sigma.d)
+end
+
+
+@Markdown.doc """
+    bump!(Tab::Tableau, x::Int)
+
+Inserts the integer `x` into the tableau Tab according to the bumping
```
fingolfin:
```suggestion
Inserts the integer `x` into the tableau `Tab` according to the bumping
```
(strictly speaking the argument should be called `tab`, i.e., in all lowercase. Only types use upper case. But again, that's not a hard requirement).

experimental/JuLie/tableaux.jl
```diff
@@ -0,0 +1,717 @@
...
+      h = div(h,ZZ(hook_length(lambda,i,j)))
+    end
+  end
+  return h
+end
+
+
+@Markdown.doc """
+    schensted(sigma::Vector{Integer})
+    schensted(sigma::Perm{T})
```
fingolfin, line 630: Note that `Perm{T}` is from AbstractAlgebra, but we don't use it in OSCAR -- we have our own permutation type here, `PermGroupElem`

That permutation type definitely should be supported here and in other places taking permutations. (To be clear: I am OK with supporting `Perm{T}`, too, at least for now)

experimental/JuLie/tableaux.jl
```diff
@@ -0,0 +1,717 @@
...
+The Robinson–Schensted correspondence is a bijection between
+permutations and pairs of standard Young tableaux of the same shape. For
+a permutation sigma (given as an array), this function performs the
+Schnested algorithm and returns the corresponding pair of standard
+tableaux (the insertion and recording tableaux).
+
+# Example
+```julia-repl
+julia> P,Q = schensted([3,1,6,2,5,4]);
+julia> P
```
fingolfin: Please don't use `julia-repl`, but rather use `jldoctest`, which then are executed as part of our CI test suite. That ensures that examples always stay up-to-date, and also gives us some basic test covered. 

When doing that, it is important that every `julia>` prompt is preceded by an empty line. So:
```suggestion
```jldoctest
julia> P,Q = schensted([3,1,6,2,5,4]);

julia> P
```

experimental/JuLie/tableaux.jl
```diff
@@ -0,0 +1,717 @@
...
+
+
+
+
+@doc Markdown.doc"""
+    num_standard_tableaux(lambda::Partition)
+
+Returns the number $f^λ$ of standard tableaux of shape ``λ`` using the hook length formula
+
+$f^λ = \frac{n!}{\prod_{i,j} h_λ(i,j)},$
```
fingolfin: for a stand-alone formula, maybe use display math, i.e.:
```suggestion
$$f^λ = \frac{n!}{\prod_{i,j} h_λ(i,j)},$$
```

experimental/JuLie.jl
```diff
@@ -0,0 +1,14 @@
+module JuLie
+
+using ..Oscar
+using Markdown
+import Nemo:libflint
```
fingolfin:
```suggestion
import Nemo: libflint
```
However, I would prefer if we could just write `Nemo.libflint` instead of `libflint`

experimental/JuLie/tableaux.jl
ClaudiaHeYun: I changed the arguments to have lowercase t.

docs/src/Combinatorics/partitions.md
ClaudiaHeYun: I fixed `partitions` to accept `fmpz` integers. However, there are a number of other functions that take integers as inputs. Should we modify everything so that they all accept `IntegerUnion`?

ClaudiaHeYun: I changed all functions in `partitions.jl` to accept `IntegerUnion`. I didn't do that for `schur_polynomials.jl` or `tableaux.jl`. Hope this doesn't break anything... 😬

experimental/JuLie/schur_polynomials.jl
ulthiel: > This is the TODO I was talking about: I added an assertion for `n>0`. @ulthiel: Is this correct?

Hm. There is this commented code block
```
  #if isempty(x) #this event is handled in the calling methods
   #	if sum(lambda)==0
   #	return 1
   #	else
   #	return 0
   #	end
   #end
```

which makes sense to me. Certainly, below, if n=0 then x[1] won't work. I don't know why this block above is commented. It's been too long.

ulthiel: I think these exceptions are handled in the function `schur_polynomial` already. But I think this should also be done in these extra functions (cbf and combinat) even though they are probably not exported (!?). This extra check should not be a tremendous overhead and makes them useable on their own. That would be my understanding trying to understand the code again.

ulthiel: > > I think the printing for the Tableaus from AA is pretty nice.
> 
> Printing the tableaux should not stop merging. Adding this (later) is an easy exercise. For once I would not even mind (mild) duplication of code; it's a few lines only.

Maybe this should not be pretty printed by default (think of a huge tableaux).

docs/src/Combinatorics/partitions.md
ulthiel: > I fixed `partitions` to accept `fmpz` integers. However, there are a number of other functions that take integers as inputs. Should we modify everything so that they all accept `IntegerUnion`?

That was my problem as well and I didn't really know what to do about it.

fieker:
On Wed, Mar 08, 2023 at 12:51:30AM -0800, Ulrich Thiel wrote:
> @ulthiel commented on this pull request.
> 
> 
> 
> > @@ -0,0 +1,36 @@
> +# Partitions
> +
> ***@***.***
> +    Partition
> +```
> +
> +## Unrestricted partitions
> +
> ***@***.***
> +    partitions(::Integer)
> 
> > I fixed `partitions` to accept `fmpz` integers. However, there are a number of other functions that take integers as inputs. Should we modify everything so that they all accept `IntegerUnion`?
> 
> That was my problem as well and I didn't really know what to do about it.
IntegerUnion is the way to go. You always want to allow Integers since
people type
partitions(4)
and not
partitions(ZZRingElem(4)) or ZZ(4)
However, programmantic, most functions in Oscar return ZZRingElem =
fmpz, so they should be valid input

micjoswig:
@fingolfin you had requested changes; to me it seems that these have been addressed. Please merge now or explain what is still missing.

ulthiel:
> IntegerUnion is the way to go. You always want to allow Integers since people type partitions(4) and not partitions(ZZRingElem(4)) or ZZ(4) However, programmantic, most functions in Oscar return ZZRingElem = fmpz, so they should be valid input
> […](#)

Ok. But if an fmpz becomes a loop variable (e.g., for i=1 to n when n::fmpz) this may be not so good. Anyways, such stuff shouldn't stop merging, can be adapted later.

micjoswig:
> @fingolfin you had requested changes; to me it seems that these have been addressed. Please merge now or explain what is still missing.

Sorry; now I see it.

@ClaudiaHeYun this is about the .DS_Store files.  You added that entry to .gitignore; but the files are still there.  Please apply `git rm` to each of them.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -189,6 +189,53 @@ end
...
+      A::MatrixElem
+    )
+    coefficient_ring(P) === base_ring(A) || error("coefficient rings incompatible")
+    return characteristic_polynomial(A, ring=P)
+  end
+```
+This allows for two different entry points for the ring `P` of the output: 
+Once as the first argument of a method of `characteristic_polynomial` with 
+an extended signature, and second as an optional keyword argument for `ring` in 
+the original method. This should be the general rule for such implementations
```
fingolfin: You are editing the OSCAR styleguide here. Either this *is* the general rule within OSCAR, or it isn't; the "should" makes no sense.

Whether this *is* the rule is debatable, though. Do we really want to force all code to offer both these variants? Why? I.e., why two ways to achieve the same thing? 

Also, perhaps `ring` is not always the best name for the argument, and from the text it sounds this is mandatory.

Indeed, as far as I can tell, the "real" `charpoly` and `minpoly` do not support the keyword argument variant.

My recommendation: move this addition to the styleguide into a separate PR (it doesn't really belong into this PR anyway). That way, we can discuss it there, and don't hold up this PR here.

experimental/Experimental.jl
```diff
@@ -27,6 +29,7 @@ include("Schemes/CoherentSheaves.jl")
 include("Schemes/LazyGlueing.jl")
 include("Schemes/CartierDivisor.jl")
 include("Schemes/Auxiliary.jl")
+
```
fingolfin:
```suggestion
```

experimental/JuLie/partitions.jl
```diff
@@ -0,0 +1,669 @@
...
+
+# Examples
+```julia-repl
+julia> P=Partition([3,2,1]) #The partition 3+2+1 of 6
+[3, 2, 1]
+julia> sum(P) #The sum of the parts.
+6
+julia> P[1] #First component
+3
+julia> P=Partition(Int8[3,2,1]) #Same partition but using 8 bit integers
```
fingolfin: All examples should be (if at all possible) using `jldoctest`, not `julia-repl`, to ensure they stay valid and are tested as part of our CI tests.
````suggestion
```jldoctest
julia> P=Partition([3,2,1]) #The partition 3+2+1 of 6
[3, 2, 1]

julia> sum(P) #The sum of the parts.
6

julia> P[1] #First component
3

julia> P=Partition(Int8[3,2,1]) #Same partition but using 8 bit integers
Int8[3, 2, 1]
````

File: experimental/JuLie/partitions.jl [hunk 17]
```diff
@@ -0,0 +1,669 @@
...
+    if dif < 0
+      return false
+    end
+  end
+  return true
+end
+
+
+@Markdown.doc """
+    conjugate(lambda::Partition{T}) where T<:IntegerUnion
```
fingolfin, line 688: The following is a purely informal remark, I don't expect you to do anything about it:

Note that in Julia (and OSCAR) this is normally called `conj`. However, I've added a suggestion to #1826 that we should allow `conjugate` as a synonym for `conj`.

fingolfin (commented):
A couple more comments.

I will shortly push several commits to this PR to address some of my own remarks. Please make sure to *pull* before (force) pushing anything to this PR / branch, to avoid loosing my changes

experimental/JuLie/schur_polynomials.jl
```diff
@@ -0,0 +1,304 @@
...
+
+If neither `R` nor `x` are given, the Schur polynomial will be over `PolynomialRing(ZZ,["x1","x2",...,"xn"])`.
+
+# Example
+```julia-repl
+julia> R,x = PolynomialRing(ZZ, ["a","b","c"])
+(Multivariate Polynomial Ring in a, b, c over Integer Ring, fmpz_mpoly[a, b, c])
+julia> schur_polynomial(Partition([2,1]),[x[1],x[2]])
+a^2*b + a*b^2
+julia> schur_polynomial(Partition([2,1]),R)
```
fingolfin: This example does not actually work (argument order is switched), which highlights the importance of using `jldoctest`

docs/src/Combinatorics/partitions.md
fingolfin: Ideally, any place that would accept a BigInt should also accept an `fmpz` / `ZZRingElem`.

Of course there are places where really only a "small" integer makes sense, in those case staying with `Integer` is fine.

fingolfin:
> > IntegerUnion is the way to go. You always want to allow Integers since people type partitions(4) and not partitions(ZZRingElem(4)) or ZZ(4) However, programmantic, most functions in Oscar return ZZRingElem = fmpz, so they should be valid input
> > […](#)
> 
> Ok. But if an fmpz becomes a loop variable (e.g., for i=1 to n when n::fmpz) this may be not so good. 

If changing `Integer` to `IntegerUnion` allows a loop variable to become an `ZZRingElem`, then the code already allowed it to become a `BigInt, which is equally bad. There are various ways to address this, but of course it depends on the code.

E.g. if the code really only makes sense for small inputs (e.g. `factorial(n)` with `n >= 2^64` is not something we can actually compute), then one can write code like this:
```julia
my_factorial(n::T) where T <: IntegerUnion = my_factorial(T, Int(n))
function my_factorial(::Type{T}, n::Int) where T <: IntegerUnion
  @req n >= 0 "n must be non-negative"
  x = T(1)
  for i in 1:n
    x = mul!(x, x, i)
  end
  return x
end
```



> Anyways, such stuff shouldn't stop merging, can be adapted later.

Agreed

experimental/JuLie/partitions.jl
```diff
@@ -0,0 +1,727 @@
...
+I don't see a significant speed difference to the descending encoding:
+```julia-repl
+julia> @btime partitions(Int8(90));
+3.376 s (56634200 allocations: 6.24 GiB)
+
+julia> @btime ascending_partitions(Int8(90),alg="ks");
+3.395 s (56634200 allocations: 6.24 GiB)
+
+julia> @btime ascending_partitions(Int8(90),alg="m");
+3.451 s (56634200 allocations: 6.24 GiB)
```
fingolfin, line 296: Interesting! With which Julia version, OS and hardware did you obtain these benchmark results?

On my M1 MacBook with Julia 1.9.0-rc1, `ascending_partitions` is actually twice as fast as `partitions`:
```
julia> @btime partitions(Int8(90));
  8.959 s (56634195 allocations: 4.53 GiB)

julia> @btime ascending_partitions(Int8(90),alg="ks");
  4.546 s (56634195 allocations: 4.53 GiB)

julia> @btime ascending_partitions(Int8(90),alg="m");
  4.616 s (56634195 allocations: 4.53 GiB)
```

experimental/JuLie/partitions.jl
```diff
@@ -0,0 +1,727 @@
...
+
+
+@Markdown.doc """
+    ascending_partitions(n::IntegerUnion;alg="ks")
+
+Instead of encoding a partition of an integer ``n ≥ 0`` as a *descending*
+sequence (which is our convention), one can also encode it as an *ascending*
+sequence. In the papers Kelleher & O'Sullivan (2014) and Merca (2012) it is
+said that generating the list of all ascending partitions is more efficient
+than generating descending ones. To test this, I have implemented the
```
fingolfin, line 275: This docstring uses "I" quite a bit: "I have", "I am", .... For an OSCAR user it won't be clear to whom that "I" refers. So we should rewrite it, either using "we"; or using "U. Thiel"

@ulthiel any preferences?

experimental/JuLie/partitions.jl
```diff
@@ -0,0 +1,727 @@
...
+  else
+    error("alg must be either ks or m")
+  end
+
+end
+
+
+
+@Markdown.doc """
+    partitions(m::IntegerUnion, n::IntegerUnion, l1::IntegerUnion, l2::IntegerUnion; z=0)
```
fingolfin: The type for `z` here should match the declaration below

experimental/JuLie/partitions.jl
```diff
@@ -0,0 +1,727 @@
...
+A list of all partitions of an integer ``m ≥ 0`` into ``n ≥ 0`` parts with
+lower bound ``l1 ≥ 0`` and upper bound ``l2 ≥ l1`` for the parts. There are
+two choices for the parameter z:
+* z=0: no further restriction (*default*);
+* z=1: only distinct parts.
+The partitions are produced in *decreasing* order.
+
+The algorithm used is "parta" in [RJ76](@cite), de-gotoed from old ALGOL code by E. Thiel!
+"""
+function partitions(m::IntegerUnion, n::IntegerUnion, l1::IntegerUnion, l2::IntegerUnion; z::IntegerUnion=0)
```
fingolfin: If `z` only allows for two choices, there is no point in allowing it to be `IntegerUnion`, it can go back to being `Int`.

However, why isn't this a `Bool` with a speaking name? E.g.
```suggestion
function partitions(m::IntegerUnion, n::IntegerUnion, l1::IntegerUnion, l2::IntegerUnion; only_distinct_parts::Bool = false)
```

experimental/JuLie/partitions.jl
fingolfin: Oh! With Julia 1.8.5, it is *much* faster and looks more like your numbers:
```julia-repl
julia> @btime partitions(Int8(90));
  2.123 s (56634195 allocations: 4.53 GiB)

julia> @btime ascending_partitions(Int8(90),alg="ks");
  2.478 s (56634195 allocations: 4.53 GiB)

julia> @btime ascending_partitions(Int8(90),alg="m");
  2.432 s (56634195 allocations: 4.53 GiB)
```
So there seems to be a major performance regression in Julia 1.9 :-(. Wonder if we can pinpoint it...

experimental/JuLie/tableaux.jl
ulthiel: I find it debatable whether a permutation really needs to be an element of a group. I feel there should be both (a permutation as an element of a group and a permutation just by itself) as separate types.

But again, such discussions should not stop any of the real work.

experimental/JuLie/partitions.jl
ulthiel: This is weird. I believe my old benchmarks were done even with Julia 1.7.

experimental/JuLie/partitions.jl
ulthiel: Well, this was my personal documentation, so of course I used "I". I don't know. It's maybe also not so important.

ulthiel: Noticing that the ascending_partitions stuff is in the pull request as well: I later changed this to ascending_compositions (which was part of my composition generator) because it's technically not a partition by definition. Since the composition stuff is currently not in the pull request (and not in JuLie) I'm unsure what should happen with this functionality. It should definitely be there eventually.

experimental/JuLie/tableaux.jl
fingolfin: Sure. And note that I did not ask for `Perm{T}` support to be dropped, but rather for support for `PermGroupElem` to be added, as that is the kind of permutation most OSCAR users will encounter.

But this is not a blocker for this PR.

micjoswig:
It seems that the test (1.6, x64, ubuntu-latest) was killed only because some test cases took too long?

If you agree, merge this PR now.  And create a new issue to reduce the time for testing.  At the end of the test output the biggest offenders are listed.  Add them explicitly to the new issue.

@fingolfin @benlorenz

fingolfin:
The tests of this PR are not slow, I already verified this.

I already spent some effort last week to clean up this PR and fix a bunch of issues with it. It's unfortunate that nobody else really had timte to carefully review this code.

The remaining blocker for me is the part where it changes the developer documentation. I will remove this now from the PR and adjust some more things. If someone wants to re-submit it in a separate PR, they are welcome, we can then discuss it there.

experimental/JuLie/partitions.jl
fingolfin: Of course it made perfect sense to use "I" while this was in JuLie, and I am not criticizing this. I am merely pointing out that it doesn't make sense anymore in the context of Oscar, because most readers won't know whom "I" refers to. Moreover, it differs from the style we used elsewhere. Hence it should be changed. I can easily change it now, of course, but as a courtesy I thought I'd ask you for your preferences...?

Regarding your second reply: Note that this is the docstring for `ascending_partitions` we are discussing here. So you are saying you renamed it later anyway, and moved it to some other module... So should we just delete it from this PR for now and it'll be added back later?

fingolfin (approved):

PR #1937, opened 2023-02-14
Improvements for toric geometry

HereAround:
This PR aims for the following:

* Remove method `is_linearly_equivalent_to_effective_divisor`.
* Add optional boolean `set_attributes` to all constructors of toric varieties. By default, our constructors will set (a possibly large number of) attributes. If one passes `set_attributes = false` as last argument, no attributes are set. The advantage of this being switched off is to avoid possible inconsistencies, possibly at the price of slightly less performance.
* Document this switch.
* Improve the tests: Based on this switch, we loop over the possibility to either have all attributes settings for toric variety constructors switched off or all of them turned on. This allows to organize our tests more efficiently and to test the code more thoroughly. It may come at the price of somewhat longer tests.
* To balance the effect of such longer tests, I have taken the liberty to reduce the tests for direct products, as this seems to be the test in the toric universe that takes longest (cf. https://github.com/oscar-system/Oscar.jl/actions/runs/4178595546 - based on these results, the direct product tests in the toric geometry are also among the top 10 most consuming tests in the OSCAR).
* Fix the deprecation warnings in the tests (cf. https://github.com/oscar-system/Oscar.jl/issues/1947).

I wonder if this switch `set_attributes` should be added to **all** constructors in the toric universe. If so, it might be best to address this in a separate PR.

cc @lkastner


HereAround:
@lkastner : Would be great if we could merge this one quickly, at the very last before the big renaming. (For instance, to remove the deprecation warnings during testing.)

cc @fingolfin

test/ToricVarieties/line_bundles.jl
```diff
@@ -1,36 +1,46 @@
...
+    @testset "Basic attributes" begin
+        if set_attributes
+            @test degree(l) == -6
+            @test degree(l^(-1)) == 6
+            @test degree(l*l) == -12
+        else
+            @test degree(l) == 10
+            @test degree(l^(-1)) == -10
+            @test degree(l*l) == 20
+        end
```
lkastner, line 35: We should avoid having too many of these blocks. Could you maybe add a short description as a comment above that the rays have different order and hence the divisors are different?

lkastner (approved):

test/ToricVarieties/line_bundles.jl
HereAround: I send you more details in slack. Maybe best to discuss there.

Long story short: You are right, I need to explain this in more detail. I shall do so in a separate PR towards the end of the upcoming week, i.e. once the big renaming is complete.

HereAround: With that said, I will merge this PR now.

PR #1942, opened 2023-02-14
Tests: create github action summary with run-times and compile-times per test-file

benlorenz:
The code to for the compile time is inspired by `base/timing.jl` since @timed doesn't return those values and uses several internal functions, currently works for julia 1.9 and nightly.

I changed code-coverage to run only for julia 1.6, I think this makes sense as this is where the doctests run as well. Not sure if there is any reason to run code-coverage for other versions as well.
The reason is that enabling code-coverage breaks the compilation time stats on julia 1.9 (and probably nightly as well).

Currently this sorts by run-time and prints the top 50 files, the compilation and recompilation times are printed separately; In contrast to julia printing total time + percentages in `@time`. This can of course be adjusted.
_Edit:_ GC-time is currently still included in the main run-time, but that could be changed easily.

To make this work I created a custom `include` function that replaces the `Base.include` and takes all the measurements, it also ignores any file that itself includes other files like `PolyhedralGeometry/runtests.jl` which just pulls in more files. (This heuristic is not really perfect but does work fine currently)

Tables in the github actions summary are currently generated only for 1.9 and nightly (because for older versions there would just be one timing).

And if you want to have a look at the timings for one test file just search through the test-log, the times are printed after each include block.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -60,6 +60,8 @@ jobs:
         run: echo "OMP_NUM_THREADS=1" >> $GITHUB_ENV
       - name: "Run tests"
         uses: julia-actions/julia-runtest@latest
+        with:
+          coverage: ${{ matrix.julia-version == '1.6' }}
       - name: "Process code coverage"
         uses: julia-actions/julia-processcoverage@v1
         with:
--- a/Project.toml
+++ b/Project.toml
@@ -41,7 +41,9 @@ julia = "1.6"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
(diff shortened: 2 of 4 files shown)


thofma:
What is the logic to distinguish test files that only include other files and those which also have tests on their own?

benlorenz:
> What is the logic to distinguish test files that only include other files and those which also have tests on their own?

The custom include function sets a global flag to true before calling the real include, if this is still true after the include block it collects the stats and sets that flag to false.

We can also avoid this hackish override for the include and just use something like a custom
`timed_include(...)` but for a first version of this I wanted to avoid touching too many files.

thofma (approved):
Thanks for the explanation. I like it. Code coverage only where we do doctests is also a good idea (we are doing the same in Hecke).

PR #1949, opened 2023-02-16
is... ---> is_...

wdecker:
I only addressed cases showing up in the commutative algebra docu.
I hope that does not cause any confusion.


wdecker:
The last commit removes some duplicate text from my previous PR and adds one further remark. As soon as the tests are green again, this can be merged.

src/Rings/mpoly-ideals.jl
```diff
@@ -907,11 +907,11 @@ ideal(x^2)
 julia> J = ideal(R, [x, y])^2
 ideal(x^2, x*y, y^2)
 
-julia> issubset(I, J)
+julia> is_subset(I, J)
 true
 ```
 """
-function Base.issubset(I::MPolyIdeal{T}, J::MPolyIdeal{T}) where T
+function is_subset(I::MPolyIdeal{T}, J::MPolyIdeal{T}) where T
```
afkafkafk13, line 914: Question: Does this behave well, if a user mistakenly writes issubset(I,J) instead of is_subset(I,J)? 
                   (i.e. does it correctly give an error message or does it simply do something unexpected)

src/Rings/MPolyQuo.jl
```diff
@@ -638,14 +638,14 @@ ideal(x^3*y^4 - x + y, x*y^2 + x*y)
...
+julia> is_subset(a,b)
 false
 
-julia> issubset(b,a)
+julia> is_subset(b,a)
 true
 ```
 """
-function Base.issubset(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}) where T
+function is_subset(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}) where T
```
afkafkafk13, line 648: same comment as in the other occurrence

src/Rings/mpoly-ideals.jl
fingolfin: They are aliases of each other and behave 100% identical

fingolfin (approved):

PR #1953, opened 2023-02-17
Tests for projective bundles iv

HechtiDerLachs:
@simonbrandhorst noted that test coverage on the diff of #1887 was only 47%. That was simply, because I didn't include the tests. But then, I found them not running anymore, but now I finally made them work again. Small change, great effect. And hard to catch.


simonbrandhorst:
Thanks for catching this!

simonbrandhorst (approved):

PR #1958, opened 2023-02-18
Fix coxring gradings & renaming some constructors

HereAround:
@lkastner Here is a proposal to streamline the attribute setting.

This works by always (= for Hirzebruch surfaces, del Pezzo surfaces and (weighted) projective spaces) fixing a grading of the Cox ring. As a consequence, many of the if/else statements that I recently introduced in the tests for the toric functionality could be removed.

@fingolfin In addition, this renames the following constructors:
* `AffineNormalToricVariety` -> `affine_normal_toric_variety`,
* `NormalToricVariety` -> `normal_toric_variety`.

As @fingolfin pointed out: For algebraic cycles, we currently issue `export p`. This should be `export polynomial` and is fixed in https://github.com/oscar-system/Oscar.jl/pull/1958/commits/7342b2109e12d9fc54c60b9f3e9af04e2db35d9f.

### DIFF
--- a/src/ToricVarieties/NormalToricVarieties/constructors.jl
+++ b/src/ToricVarieties/NormalToricVarieties/constructors.jl
@@ -274,6 +274,16 @@ function projective_space(::Type{NormalToricVariety}, d::Int; set_attributes::Bo
     pm_object = Polymake.fulton.NormalToricVariety(Oscar.pm_object(f))
     variety = NormalToricVariety(pm_object)
     
+    # make standard choice for weights of the cox ring
+    set_attribute!(variety, :torusinvariant_cartier_divisor_group, free_abelian_group(d+1))
+    set_attribute!(variety, :torusinvariant_weil_divisor_group, free_abelian_group(d+1))
+    set_attribute!(variety, :class_group, free_abelian_group(1))
+    set_attribute!(variety, :picard_group, free_abelian_group(1))
+    weights = matrix(ZZ,hcat([1 for i in 1:d+1]))
+    set_attribute!(variety, :map_from_torusinvariant_weil_divisor_group_to_class_group, hom(torusinvariant_weil_divisor_group(variety), class_group(variety), weights))
+    set_attribute!(variety, :map_from_torusinvariant_cartier_divisor_group_to_picard_group, hom(torusinvariant_cartier_divisor_group(variety), picard_group(variety), weights))
+    set_attribute!(variety, :map_from_torusinvariant_cartier_divisor_group_to_torusinvariant_weil_divisor_group, hom(torusinvariant_cartier_divisor_group(variety), torusinvariant_weil_divisor_group(variety), identity_matrix(ZZ,d+1)))
+    
     if set_attributes
         set_attribute!(variety, :is_affine, false)
         set_attribute!(variety, :is_projective, true)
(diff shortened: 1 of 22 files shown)


src/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -463,6 +472,38 @@ function del_pezzo_surface(b::Int; set_attributes::Bool = true)
...
+        weights = matrix(ZZ, [1 1; 1 1; 1 0; 0 -1])
+        set_attribute!(variety, :map_from_torusinvariant_weil_divisor_group_to_class_group, hom(torusinvariant_weil_divisor_group(variety), class_group(variety), weights))
+        set_attribute!(variety, :map_from_torusinvariant_cartier_divisor_group_to_picard_group, hom(torusinvariant_cartier_divisor_group(variety), picard_group(variety), weights))
+        set_attribute!(variety, :map_from_torusinvariant_cartier_divisor_group_to_torusinvariant_weil_divisor_group, hom(torusinvariant_cartier_divisor_group(variety), torusinvariant_weil_divisor_group(variety), identity_matrix(ZZ,4)))
+    end
+    if b == 2
+        set_attribute!(variety, :torusinvariant_cartier_divisor_group, free_abelian_group(5))
+        set_attribute!(variety, :torusinvariant_weil_divisor_group, free_abelian_group(5))
+        set_attribute!(variety, :class_group, free_abelian_group(3))
+        set_attribute!(variety, :picard_group, free_abelian_group(3))
```
fieker, line 491: I think you shoult be able to do:
`set_attribute!(varienty, :torus=>1, :paul=>true, :free=>A)`

HereAround:
Converted to a draft until the tests pass again. Chances are that after the renaming, we see some failures.

src/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -177,21 +208,28 @@ Set `P` to be a square.
...
+    variety = normal_toric_variety(normal_fan(P))
     if set_attributes
         set_attribute!(variety, :polyhedron, P)
     end
     return variety
 end
 
 
+function NormalToricVariety(P::Polyhedron; set_attributes::Bool = true)
+    Base.depwarn("'NormalToricVariety(P::Polyhedron; set_attributes::Bool = true)' is deprecated, use"*
```
fingolfin: Maybe better to move these deprecations to `src/Deprecations.jl`, together with a comment on the version (look at the existing content of the file to see what I mean). That way, it is easier for us to remove it in the future.

src/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -274,6 +319,16 @@ function projective_space(::Type{NormalToricVariety}, d::Int; set_attributes::Bo
     pm_object = Polymake.fulton.NormalToricVariety(Oscar.pm_object(f))
     variety = NormalToricVariety(pm_object)
     
+    # make standard choice for weights of the cox ring
+    set_attribute!(variety, :torusinvariant_cartier_divisor_group, free_abelian_group(d+1))
+    set_attribute!(variety, :torusinvariant_weil_divisor_group, free_abelian_group(d+1))
+    set_attribute!(variety, :class_group, free_abelian_group(1))
+    set_attribute!(variety, :picard_group, free_abelian_group(1))
+    weights = matrix(ZZ,hcat([1 for i in 1:d+1]))
```
fingolfin: Just FYI you could also do this:
```suggestion
    weights = matrix(ZZ,ones(Int,d+1,1))
```
It requires fewer allocations:
```
julia> @btime ones(Int,d+1,1);
  120.773 ns (1 allocation: 112 bytes)

julia> @btime hcat([1 for i in 1:d+1]);
  135.216 ns (4 allocations: 288 bytes)
```

src/ToricVarieties/NormalToricVarieties/constructors.jl
fingolfin: Yes this can be done. I don't think it'll save much, but then perhaps it does... one would have to benchmark.

But these are slow constructors anyway, so  maybe not a reason to worry much

fingolfin (approved):
Looks good to me, I have one change suggestion (moving the deprecations to a different file) but that could also be done in a future PR.

src/ToricVarieties/NormalToricVarieties/constructors.jl
HereAround: Done. I updated the PR just now.

src/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -347,7 +350,13 @@ function weighted_projective_space(::Type{NormalToricVariety}, w::Vector{T}; set
     tr,_ = pseudo_inv(lattice_gens)
     ray_gens = ray_gens * transpose(tr)
     mc = IncidenceMatrix(subsets(Vector{Int}(1:length(w)), length(w)-1))
-    variety = NormalToricVariety(PolyhedralFan(ray_gens, mc; non_redundant=true ))
+    variety = normal_toric_variety(PolyhedralFan(ray_gens, mc; non_redundant=true ))
+    
+    # make standard choice for the weights of the cox ring
+    set_attribute!(variety, :torusinvariant_weil_divisor_group, free_abelian_group(length(w)))
+    set_attribute!(variety, :class_group, free_abelian_group(1))
+    weights = matrix(ZZ,hcat([w[i] for i in 1:length(w)]))
```
fingolfin, line 358:
```suggestion
    weights = matrix(ZZ,hcat(w))
```
That's shorter *and* allocs less / is faster (not that this matter much here).

I think we could improve this in a bunch of places outside of your code, too (don't worry about that, though, I just wanted to point out that multiple people, including me until today, seem not to be aware of this...)

HereAround: Thx for the pointer. We should look into this eventually. I put it in a separate issue (https://github.com/oscar-system/Oscar.jl/issues/1969), so we do not forget about this.

PR #1959, opened 2023-02-19
add Koszul complex

wdecker:


wdecker:
Unfortunately, the printing of complexes is still a mess.

fieker (approved):

PR #1961, opened 2023-02-20
Rework of Hilbert-driven GB computation

RafaelDavidMohr:
As discussed with @ederc and @wdecker this is a rework of the hilbert driven GB computation to

1. First try to find weights w.r.t. which the input ideal is homogeneous in `standard_basis` before homogenizing.
2. Let the user pass weights themselves into `groebner_basis_hilbert_driven`

### DIFF
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -1362,3 +1362,26 @@ function _extend_mon_order(ordering::MonomialOrdering,
   m_hom[2:end, 2:end] = m
   return matrix_ordering(homogenized_ring, m_hom)
 end
+
+# compute weights such that F is a homogeneous system w.r.t. these weights
+function _find_weights(F::Vector{P}) where {P <: MPolyElem}
+  nrows = maximum(length, F) - 1
+  ncols = ngens(parent(first(F)))
+  mat_space = MatrixSpace(ZZ, nrows, ncols)
+
+  # TODO: maybe find a simpler way to build these matrices
+  exponent_diffs = [[i > length(e) ? zeros(Int, ncols) : e[i] - e[1] for i in 1:nrows]
+                    for e in (collect).((exponents).(F))]
+  matrices = [mat_space(permutedims(reduce(hcat, e))) for e in exponent_diffs]
+
+  # vector in the kernel of M should lie in the kernel of all of the matrices
+  M = sum([rand(-1000:1000)*m for m in matrices])
+  KM = kernel(M)[2]
+  k = sum([rand(-1000:1000)*v for v in eachrow(collect(KM))])
+
+  if all(ent -> ent != 0, k) && all(iszero, [mat * k for mat in matrices])
+    return (Int).(k)
+  else 
+    return zeros(Int, ncols)
+  end
+end
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -1255,24 +1255,25 @@ end
 @doc Markdown.doc"""
     groebner_basis_hilbert_driven(I::MPolyIdeal{P};
                                   ordering::MonomialOrdering,
-                                  complete_reduction::Bool = false) where {P <: MPolyDecRingElem}
+                                  complete_reduction::Bool = false,
+                                  weights::Vector{Int} = ones(ngens(base_ring(I)))) where {P <: MPolyElem}
 
 
 Compute a Gröbner basis of `I` with respect to `ordering` using a
 Hilbert Series driven method as follows: If a Gröbner basis for `I` is
-present, compute the Hilbert series of `I` and use it to optimize the
-Gröbner basis computation for `I` w.r.t. `ordering`. If no Gröbner
-basis for `I` is present compute the Hilbert series for `I` if the
-base field of `I` has positive characteristic, otherwise compute the
-Hilbert series for `I` modulo a randomly chosen prime. Use the
-resulting Hilbert series to optimize the Gröbner basis computation for
-`I` w.r.t. `ordering`.
+present, compute the Hilbert series of `I` (w.r.t. `weights`) and use
+it to optimize the Gröbner basis computation for `I`
+w.r.t. `ordering`. If no Gröbner basis for `I` is present compute the
+Hilbert series for `I` if the base field of `I` has positive
+characteristic, otherwise compute the Hilbert series for `I` modulo a
+randomly chosen prime. Use the resulting Hilbert series to optimize
+the Gröbner basis computation for `I` w.r.t. `ordering`.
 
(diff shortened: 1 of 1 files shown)


src/Rings/groebner.jl
```diff
@@ -1362,3 +1362,26 @@ function _extend_mon_order(ordering::MonomialOrdering,
...
+  exponent_diffs = [[i > length(e) ? zeros(Int, ncols) : e[i] - e[1] for i in 1:nrows]
+                    for e in (collect).((exponents).(F))]
+  matrices = [mat_space(permutedims(reduce(hcat, e))) for e in exponent_diffs]
+
+  # vector in the kernel of M should lie in the kernel of all of the matrices
+  M = sum([rand(-1000:1000)*m for m in matrices])
+  KM = kernel(M)[2]
+  k = sum([rand(-1000:1000)*v for v in eachrow(collect(KM))])
+
+  if all(ent -> ent != 0, k) && all(iszero, [mat * k for mat in matrices])
```
fieker: for the fun of it: there is also `any`: `all(ent -> end != 0, ..)` should be (I think) `!any(iszero, ...)`

ederc:
I do not see where you use `_find_weights()` in `standard_basis()`. The same holds for the weights as arguments for the hilbert driven GB algorithm. I assume this is still work in progress, just wanted to point this out.

wdecker:
Let us agree on what we wish to do: Here is a suggestion (very quickly written):

1.  Given an ideal $I$ of a multivariate polynomial ring $R$ over a field and a slow `destination_ordering`,
check whether $I$ is homogeneous with respect to standard $\mathbb Z$-grading on `R`. If so, set `o = degrevlex`.
2. If not, check whether there exists a $\mathbb Z$-grading on $R$ with positive weights such that $I$ is homogeneous
with respect to this grading. If so, let `o` be the corresponding weight ordering. If not, go to step 5.
3.  Compute a Gröbner basis of $I$ with respect to `o`. Use this Gröbner basis to compute the Hilbert function of `R/I`.
4. Compute a Gröbner basis with respect to the `destination_ordering`,  but skip all Buchberger tests in a given (weighted)
degree as soon as the leading terms found so far  account for the Hilbert function in that degree. Return the computed basis.
5. Homogenize I with respect .., extend the `destination_ordering` to a block ordering go through steps 3,4 and return the dehomogenized GB.

Allow more entries in `groebner_basis_hilbert_driven` (like the Hilbert-function).

RafaelDavidMohr:
> Let us agree on what we wish to do: Here is a suggestion (very quickly written):
> 
> 1. Given an ideal I of a multivariate polynomial ring R over a field and a slow `destination_ordering`,
>    check whether I is homogeneous with respect to standard Z-grading on `R`. If so, set `o = degrevlex`.
> 2. If not, check whether there exists a Z-grading on R with positive weights such that I is homogeneous
>    with respect to this grading. If so, let `o` be the corresponding weight ordering. If not, go to step 5.
> 3. Compute a Gröbner basis of I with respect to `o`. Use this Gröbner basis to compute the Hilbert function of `R/I`.
> 4. Compute a Gröbner basis with respect to the `destination_ordering`,  but skip all Buchberger tests in a given (weighted)
>    degree as soon as the leading terms found so far  account for the Hilbert function in that degree. Return the computed basis.
> 5. Homogenize I with respect .., extend the `destination_ordering` to a block ordering go through steps 3,4 and return the dehomogenized GB.
> 
> Allow more entries in `groebner_basis_hilbert_driven` (like the Hilbert-function).

Yes, this is precisely my plan, maybe I should have made that clearer in the intial description, sorry. For now I just wanted to start the pull request so that people know this is happening.

RafaelDavidMohr:
> I do not see where you use `_find_weights()` in `standard_basis()`. The same holds for the weights as arguments for the hilbert driven GB algorithm. I assume this is still work in progress, just wanted to point this out.

Yep, still work in progress!

src/Rings/groebner.jl
```diff
@@ -154,15 +154,32 @@ function standard_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_orde
...
-    if base_ring(I) isa MPolyDecRing
-      J, target_ordering  = I, ordering
+    if all(_is_homogeneous, gens(I))
+      J, target_ordering, weights, hn  = I, ordering, ones(Int, ngens(base_ring(I))), nothing
     else
-      gens_hom = homogenization(gens(I), "w")
-      J = ideal(parent(first(gens_hom)), gens_hom)
-      target_ordering = _extend_mon_order(ordering, base_ring(J))
+      weights = _find_weights(gens(I))
+      if !iszero(weights[1])
```
ederc, line 161: `elseif` instead of `else`and `if`?

src/Rings/groebner.jl
```diff
@@ -1362,3 +1365,74 @@ function _extend_mon_order(ordering::MonomialOrdering,
...
+    L = LinearProgram(Pol, obj_func)
+    m, v = solve_lp(L)
+    if isnothing(v)
+      Pol_new = intersect(Pol, Polyhedron(ei*K, [1]))
+      L = LinearProgram(Pol_new, obj_func)
+      v = optimal_vertex(L)
+    end
+    pos_vec += K*(v.p)
+  end
+  ret = (Int).(lcm((denominator).(pos_vec)) .* pos_vec)
```
ederc, line 1435: Shouldn't we be careful with the possible degrees? Like say degree should be `<2^16` for faster Singular computations?

src/Rings/groebner.jl
```diff
@@ -154,15 +154,32 @@ function standard_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_orde
   elseif algorithm == :fglm
     _compute_groebner_basis_using_fglm(I, ordering)
   elseif algorithm == :hilbert
-    if base_ring(I) isa MPolyDecRing
-      J, target_ordering  = I, ordering
+    if all(_is_homogeneous, gens(I))
```
ederc: Wouldn't `_find_weights(gens(I))` find this, too, returning `ones(...)` as weights? If so we could make the code more compact.

src/Rings/groebner.jl
RafaelDavidMohr: `Singular.std_hilbert` accepts `Int32` weights, is this the only restriction we should enforce on the weights? I have no idea if the speed of the computation is otherwise affected by the size of the weights.

src/Rings/groebner.jl
RafaelDavidMohr: See https://github.com/oscar-system/Oscar.jl/pull/1961/commits/24612827533bfb9bbd822383a032936d1e9843ed

ederc (approved):

PR #520, opened 2021-07-03
IntersectionTheory

jieaosong:
Here is a first draft for moving IntersectionTheory to Oscar. Things that already work can be found in the test file (no blowup yet).

I used Oscar's `MPolyRing_dec` and `MPolyQuo` for the Chow rings, but it becomes a bit hard to keep track of the differences between the two (like `R.R` vs `R.R.R` or `x.f` vs `x.f.f`...). So I'm a bit keen to create a custom ring structure, with an optional field `R.I` for the ideal.

PR #1986, opened 2023-02-25
IntersectionTheory again

thofma:
This is a revised version of #520. This is now in the same state as #520, which means that still many files are not included and the corresponding tests disabled (these were already disabled in #520).


micjoswig:
The code for grassmannian and flag looks quite naive to me, i.e., not obviously useful in the sense that it can be used for any nontrivial computation.

@dcorey2814  Probably Schubert calculus should be scrutinized, too.

dcorey2814:
What sort of computations did you have in mind? I was able to reproduce the computation on p. 227 of "3264 and all that" by Eisenbud and Harris. There, they compute degrees of top Chern classes of certain vector bundles on Gr(2,n+1) in order to count the number of lines on a hypersurface of projective space. The largest number they compute is  

4798492409653834563672780605191070760393640761817269985515 = lines in a general deg 37 hypersurface of P^20

and the whole thing took ~10 seconds (in the text, they say that Macaulay2 takes ~120 seconds)

But maybe it's fast since the Chow ring has only 2 generators, so maybe for higher dimensional subspaces this slows down?

wdecker:
To give a nontrivial degree  3 example, once all files Are incorporated, you can easily recompute the famous number of twisted cubics on a general quintic in P4 (and much More).

Von meinem iPhone gesendet

wdecker:
@thofma  Is this good to go?

micjoswig:
I had a look at some of the code, and we just had a discussion with @benlorenz @lkastner @antonydellavecchia .

We concluded that, in its current state, the code does not meet the quality requirements for OSCAR yet.  The documentation is incomplete, many tests are not running, the design does not take full advantage of OSCAR etc.

We need to set general rules for onboarding code like this.  Let's talk about this in person tomorrow.

thofma:
I just adjusted the code to make it run on the current Oscar version. As I told Wolfram, I am not finished with this.

thofma:
@wdecker I think we can put this in for now and improve/change it once you started playing with it.

experimental/IntersectionTheory.jl
lgoettgens, line 1: I think this file is no longer needed due to the new experimental structure.

wdecker (approved):
I will gradually add documention to this before more of the functionality will be made available.

HereAround:
Can https://github.com/oscar-system/Oscar.jl/pull/520 be closed now?

PR #1926, opened 2023-02-11
Experimental module `SymInt`

StevellM:
For my research, I wrote this package. It has been suggested to me to push what I have on the experimental of Oscar (with tests and docs) for the following reasons:
- there are some ideas of what could be needed regarding group algebra modules and representation theory, for later implementation into the `Oscar/src` and extension of what is already available (in this module everything is for finite groups and "big enough" fields of characteristic zero);
- to make referencing easier (if one day someone is interested about all this...);
- to prevent breaking changes when I would have to re-use, improve or extend my codes.

All this code was originally written for the functions defined in `K3Models.jl`: only some parts of it were essential but I wrote more for debugging, testing and working with my outputs. It is important to note that:
- I am not an absolute expert about the mathematical tools I have used. Some definitions might be incomplete or biased, the codes could be more efficient and/or less restrictive, and some namings are not part of any reliable corpus. However, for my purpose, everything was fast enough and easy to work with;
- What I call `elevators` are just some `solve_mixed` in disguise, which for my purpose needed to be handled differently;
- I could be sloppy sometimes with the definitions I make and the actual maths definitions: again, at the time of writing the codes, it was easier for me to follow and understand what I was doing (and avoid names with 150 characters)
- The `K3Models.jl` file is not tested: they are new codes but basically huge shortcuts regarding what I was doing previously;
- The base of representation theory is wrapped from the `repsn` package and some other parts could have been taken from another gap package that I discovered recently. I have not tried to see whether my code was faster (which I doubt) and I decided to keep my code instead of reconstructing considering this other gap package;
- Many parts of this code, I guess, will be replaced in the time by a dedicated Oscar code. I will help to delete and replace my parts of codes with the most reliable implemented code in due time (and make the tests pass).

I have tried to cover most of the code several times by the tests, on my local machine it takes around 4 minutes to test everything. I can still shorten them if that is too much for a module.

This is the final version of my code that I told you about @fingolfin , @ThomasBreuer and @fieker . Let me know if they are really big/controversial things about it. It is not meant to be pure final Oscar code, so I do not know whether we should discuss about deep design or namings. Though, I will follow what you suggest me on that point.

Issue #1875, opened 2023-01-23
Kernel and rationality of characters


thofma:
Would be great if we had `kernel(x)` or `is_rational(x)`, where `x` is a character of a finite group.

Edit: `is_abelian(x)` would also be nice, but I am not sure if you would endorse it.

thofma:
ping @ThomasBreuer

joschmitt:
This was fixed by #1989 as it seems. Please reopen if there is something missing.

ThomasBreuer:
@joschmitt Thanks for spotting this.
Yes, I think this issue can be closed.
(I do not remember what the intended meaning of `Is_abelian` for characters had been, but we had agreed on leaving this out.)

PR #1989, opened 2023-02-27
added `center`, `det`, `is_rational`, `kernel`, `order` for group characters

ThomasBreuer:
This addresses issue #1875 and some comments about pull request #1926.

### DIFF
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -24,6 +24,7 @@ export
     class_lengths,
     class_multiplication_coefficient,
     class_parameters,
+    class_positions_of_center,
     class_positions_of_kernel,
     class_positions_of_pcore,
     decomposition_matrix,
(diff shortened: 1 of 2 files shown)


src/Groups/group_characters.jl
```diff
@@ -1207,7 +1229,7 @@ Base.iterate(chi::GAPGroupClassFunction, state = 1) = state > length(chi.values)
 
 @doc Markdown.doc"""
     degree(::Type{T} = QQFieldElem, chi::GAPGroupClassFunction)
-           where T <: Union{IntegerUnion, ZZRingElem, mpq, QQAbElem}
+           where T <: Union{IntegerUnion, ZZRingElem, QQFieldElem, QQAbElem}
```
thofma, line 1214: Not related to this PR, but why is this not `ZZRingElem` by default?

src/Groups/group_characters.jl
```diff
@@ -1378,6 +1425,116 @@ function class_positions_of_kernel(list::Vector{T}) where T
...
+    class_positions_of_center(chi::GAPGroupClassFunction)
+
+Return the array of those integers `i` such that `chi[i]` is `chi[1]` times
+a root of unity.
+
+# Examples
+```jldoctest
+julia> println(class_positions_of_center(character_table("2.A5")[2]))
+[1, 2]
+
```
thofma:
```suggestion
```

thofma:
Thanks!

src/Groups/group_characters.jl
ThomasBreuer: O.k.

src/Groups/group_characters.jl
ThomasBreuer: `GAPGroupClassFunction` objects are more general than (virtual) characters, one can think of scalar multiples of differences of characters. One example where rational multiples occur in practice are the coefficient vectors of idempotents in the centre of the group algebra.
Of course one can argue that calling `degree` for such a class function does not make sense. Or one can argue that one could consider arbitrary cyclotomics for scalar multiplication of class functions.

fingolfin (approved):

Issue #1981, opened 2023-02-24
Mention rules for types names in dev docs


fingolfin:
We came up with these provision rules during the workshop.

"Someone" should put this into the Oscar dev manual. (Maybe sort the list of approved abbreviations alphabetically?)

- types for rings/groups/ideals/modules/... end with `Ring`/`Group`/`Ideal`/`Module`/...
- types for element should have the same name as the type of the parent with `Elem` added
   - exception: `MatrixSpace` elements end with `Matrix`
- We abbreviate certain parts of type names, according to a fixed set of substitutions; further abbreviations should be carefully decided upon
- every abbreviation must be unique; e.g. `Abs` stands for `Absolute`, and so must not be used for e.g. `Abstract`
- list of approved abbreviations
    - quotient -> `Quo`
    - subquotient -> `Subquo` 
    - localized -> `Loc`
    - decorated -> `Dec`
    - absolute -> `Abs`
        - abstract -> `Abstract`
    - relative -> `Rel`
    - polynomial -> `Poly`
    - multivariate polynomial -> `MPoly`
    - matrix -> `Matrix` or `Mat` ??
    - group -> `Group`
    - ring ->`Ring`
    - ideal -> `Ideal`
    - module -> `Module`
- If a type comes in sparse and dense variants, then call the dense type `T` and the sparse on `SparseT`

fingolfin:
Apparently was resolved in 2023 by @lkastner and we forgot to close this :-)

PR #1990, opened 2023-02-27
Rename MPolyRingSparse -> SparseMPolyRing

lgoettgens:
Rename `MPolyRingSparse` according to https://github.com/oscar-system/Oscar.jl/issues/1981#issue-1598784722
> * If a type comes in sparse and dense variants, then call the dense type `T` and the sparse on `SparseT`

Closes #2405 (alternative to this PR).

### DIFF
--- a/experimental/SparseMPolyRing/SparseMPolyRing.jl
+++ b/experimental/SparseMPolyRing/SparseMPolyRing.jl
@@ -3,7 +3,7 @@ This is an approach to create a multivariate polynomial ring implementation
 where the size of a monomial is independent of the number of variables in the ring.
 Thus this implementation is to be used with low degree polynomials over rings
 with lots of variables.
-This strongly follows Generic.MPolyRingSparse from AbstractAlgebra.
+This strongly follows Generic.MPolyRing from AbstractAlgebra.
 """
 
 import Base: deepcopy_internal, divrem, gcd, hash, isless, isone, iszero, 
@@ -24,11 +24,11 @@ import AbstractAlgebra: base_ring, change_base_ring, change_coefficient_ring,
 
 import AbstractAlgebra.Generic: ordering
 
-export MPolyRingSparse, MPolySparse, PolynomialRingSparse
+export SparseMPolyRing, SparseMPoly, sparse_polynomial_ring
 
 ###############################################################################
 #
-#   MPolyRingSparse / MPolySparse
+#   SparseMPolyRing / SparseMPoly
 #
 ###############################################################################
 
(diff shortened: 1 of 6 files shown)


fingolfin (commented):
Thanks. Makes sense to me. I'll leave it to @fieker to confirm whether we want to settle on that rule or not.

If we do, this PR needs an entry in `src/Aliases.jl` for backwards compatibility.

fingolfin:
Actually, on second thought, this name makes no sense: all MPolyRing are sparse by design.

This ring seems to be about "being even sparser", so it is a "sparse sparse polynomial ring", which isn't a great name. I recommend we come up with some alternative name.

A comment in the code says:
> This strongly follows Generic.MPolyRingSparse from AbstractAlgebra.

But there is no Generic.MPolyRingSparse in AA?

lgoettgens:
> A comment in the code says:
> 
> > This strongly follows Generic.MPolyRingSparse from AbstractAlgebra.
> 
> But there is no Generic.MPolyRingSparse in AA?

That was a typo in the previous version. I fixed it to
`This strongly follows Generic.MPolyRing from AbstractAlgebra.`

> Actually, on second thought, this name makes no sense: all MPolyRing are sparse by design.
> 
> This ring seems to be about "being even sparser", so it is a "sparse sparse polynomial ring", which isn't a great name. I recommend we come up with some alternative name.

The main application was for use cases, where one has many (thousands) of variables, but it is known that all occurring polynomials have a low total degree. In this case, the `Generic.MPolyRing` from AA wastes a lot of space of just storing zero exponents for all variables not occurring in a monomial.
A different name is fine with me :)

experimental/SparseMPolyRing/SparseMPolyRing.jl
```diff
@@ -24,11 +24,11 @@ import AbstractAlgebra: base_ring, change_base_ring, change_coefficient_ring,
 
 import AbstractAlgebra.Generic: ordering
 
-export MPolyRingSparse, MPolySparse, PolynomialRingSparse
+export SparseMPolyRing, SparseMPoly, sparse_polynomial_ring
```
fingolfin, line 27: If you want to follow the new naming scheme, it should be at least this:
```suggestion
export SparseMPolyRing, SparseMPolyRingElem, sparse_polynomial_ring
```

fingolfin: @thofma @fieker would be good to reach an agreement what to do with this PR resp. this type.... any alternative proposals, or shall we stick with the suggested new names above? I am somewhat wary of `sparse_polynomial_ring`, I think it may confuse people / lead them to incorrect conclusions about MPoly(Ring)

thofma: Can we discuss it before/after the next meeting?

fingolfin: sure

lgoettgens: Was there already a conversation about this?

fingolfin: nope, at least not with me

fieker:
This discussion wasn't yet. Status:
  - we do not have a dense mpoly-type, they are always sparse
  - if you want a dense type it would be the recursive univariate one - but no mpoly interface
  - there are also the SLPolys
  - Bill was dreaming of a dense mpoly type at some time... as well as a sparse univariate one

lgoettgens:
If there is no strong endorsement for this type, I would suggest to remove it (instead of renaming it).
In recent benchmarks in my use-case project, it wasn't any better (as one should have expected) and thus I reverted to using the "usual" mpolys there again. 

The corresponding PR to remove `MPolyRingSparse` is #2405.

PR #1992, opened 2023-02-27
added index fix for Grassmann Pluecker Relations

antonydellavecchia:
index shift when the default ring is used


benlorenz:
I think we can try to adjust the conversion methods here to `1:nvars`:
https://github.com/oscar-system/Oscar.jl/blob/5a46876b4520bb454b88041de5e8a453bffc8792/src/Polymake/polymake_to_oscar.jl#L11-L19

That would avoid adding extra code here and I think it makes more sense for a polymake ->  OSCAR conversion.

I tried this locally and got no errors from tests or doctests.

benlorenz (approved):

PR #1996, opened 2023-02-28
Remove invalid root test

fingolfin:
The function `root` is for computing exact roots, and by default throws an
exception if the input is not a perfect power (except that this was
disabled in Nemo for some time, and has only been re-enabled in the most
recent version).

For integer truncated roots, use `iroot`.


fingolfin:
This causes failures in the OscarCI tests at e.g. https://github.com/Nemocas/Nemo.jl/actions/runs/4288062105/jobs/7469638459

thofma:
There are more coming from Hecke, I am working on it.

fingolfin:
@thofma should we merge this one here anyway?

PR #1997, opened 2023-02-28
Update URLs to www.oscar-system.org

### LINKED ISSUE #1792: Move OSCAR homepage from oscar.computeralgebra.de to www.oscar-system.org ?
Right now the official OSCAR homepage is https://oscar.computeralgebra.de 

Some time ago I already registered https://www.oscar-system.org and set it up to forward to https://oscar.computeralgebra.de; also https://oscar-system.org forwards to it.

Based on some informal polling on the Slack, it seems I am not alone in preferring https://www.oscar-system.org -- e.g. it seems "more international" (`.de` instead of `.org)`.

Actually I already changed our documentation to be served under https://docs.oscar-system.org/ (instead of https://oscar-system.github.io/Oscar.jl/) and so far at least nobody complained ;-).

I thus propose we make https://www.oscar-system.org the new canonical homepage URL, and let https://oscar-system.org and https://oscar.computeralgebra.de redirect to it. (That is, *all existing links* would keep working, it is just a matter what URL is shown in the end in the address bar, and of course we'd also change our links to the website in our READMEs, manuals etc.)

I'll wait and here if @fieker @micjoswig @wdecker have any objections, or if they are OK with it.

fingolfin:
This is our new primary website. The old will keep working as a redirect (it may take some time for http://oscar.computeralgebra.de to redirect due to DNS caches and whatnot)

Also replace hardcoded Slack Join links by https://oscar-system.org/slack so that we only have to update the join URL in one place. Also, this URL is reasonably easy to remember and give to others.

Resolves #1792

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ julia> using Oscar
 
 However, some of OSCAR's components have additional requirements.
 For more detailed information, please consult the [installation
-instructions](https://oscar.computeralgebra.de/install/) on our website.
+instructions](https://www.oscar-system.org/install/) on our website.
 
 ## Contributing to OSCAR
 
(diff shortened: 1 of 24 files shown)


fingolfin:
I invite people to update the URL in any other places they can think of

PR #1999, opened 2023-02-28
First draft of how `experimental` should be used.

lkastner:
After a discussion during the retreat I wrote down my notes as a draft on how the `experimental` folder should be used for adding new features. The main point is that code in `experimental` is allows to have its interface changed.

Also we should write something about that code from `src` may never call code from `experimental`.

### DIFF
--- a/docs/src/DeveloperDocumentation/documentation.md
+++ b/docs/src/DeveloperDocumentation/documentation.md
@@ -11,10 +11,12 @@ Exceptions to this philosophy are the developer and general pages.
     https://docs.oscar-system.org/previews/PR<prnumber>/
     where you insert the number of your PR for `prnumber`. Alternatively you
     can look at the github actions tab of your PR and click the details link
-    next to the documentation action. These things only work if there are no
-    conflicts with the master branch and the documentation action is
-    successful. You can still build the documentation locally with the commands
-    described below.
+    next to the `documenter/deploy` action. There are a few conditions for this 
+    to work:
+    - No conflicts with the master branch.
+    - Documentation action is successful, i.e. no doctest errors.
+    - The branch for the PR is in the main `oscar-system/Oscar.jl` repository.
+    You can still build the documentation locally with the commands described below.
 
 ## Docstrings of exported functions
 
--- a/docs/src/DeveloperDocumentation/documentation.md
+++ b/docs/src/DeveloperDocumentation/documentation.md
@@ -5,6 +5,17 @@ information as possible into the docstrings and only use the doc pages for
 collecting this information and provide some additional general context.
 Exceptions to this philosophy are the developer and general pages.
 
+!!! note "Previewing the documentation"
+    Once you have created a pull request it is possible to preview the
+    documentation on github using the link
+    https://docs.oscar-system.org/previews/PR<prnumber>/
+    where you insert the number of your PR for `prnumber`. Alternatively you
+    can look at the github actions tab of your PR and click the details link
+    next to the documentation action. These things only work if there are no
+    conflicts with the master branch and the documentation action is
+    successful. You can still build the documentation locally with the commands
+    described below.
+
 ## Docstrings of exported functions
 
 Exported function should have docstrings, which look like
--- a/docs/src/DeveloperDocumentation/documentation.md
+++ b/docs/src/DeveloperDocumentation/documentation.md
@@ -5,18 +5,6 @@ information as possible into the docstrings and only use the doc pages for
 collecting this information and provide some additional general context.
 Exceptions to this philosophy are the developer and general pages.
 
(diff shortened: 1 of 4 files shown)


docs/src/DeveloperDocumentation/new_developers.md
```diff
@@ -27,6 +27,8 @@ on these things online.
 4. Please also read our page on [Documenting OSCAR code](@ref).
 5. Look at existing code that does similar things to your project to get an
    idea of what OSCAR code should look like. Try to look at multiple examples.
+6. If you are plan to implement a new feature from scratch, please also read
```
fingolfin:
```suggestion
6. If you plan to implement a new feature from scratch, please also read
```
or
```suggestion
6. If you are planning to implement a new feature from scratch, please also read
```

docs/src/DeveloperDocumentation/experimental.md
```diff
@@ -0,0 +1,25 @@
+# Implementing features from scratch
+
+For new features whose interface naturally will change a lot during the
```
fingolfin:
```suggestion
For new features whose interface is likely to change a lot during the
```

docs/src/DeveloperDocumentation/experimental.md
```diff
@@ -0,0 +1,25 @@
...
+implementation phase, Oscar.jl has a staging area, the `experimental` folder.
+
+1. The new feature is implemented in the `experimental` folder.
+2. Additionally to the original authors, a maintainer is assigned to guide the
```
fingolfin: This would be something new, so far we haven't done it. I see some merit (we want to ensure some forward progress happens and that someone takes responsibility). But for e.g. people who are already active (e.g. @fieker has quite some in `experimental`, I am not sure assigning a second internal person would be helpful?

micjoswig: I just wanted to write: We need a second person only for newbies.

But maybe that's not true.  Maybe even OSCAR veterans could occasionally discuss why some of their code is sitting in experimental for a quite a while?

The main problem with the current version of `experimental` is that the user cannot really distinguish from the actual core.

fieker (approved):

thofma (approved):
Thanks Lars!

docs/src/DeveloperDocumentation/experimental.md
fingolfin: Indeed. Some options come to mind:
- maybe accessing code in experimental should be "guarded" and require `using Oscar.Experimental`, or require writing `Experimantal.FOO` (so `Experimental.galois_group(...)`
  - pro: obviously experimental
  - con: less discoverable
- or maybe the primary functions should have a `@warn_experimental()` call in them, which prints a warning about them being experimental (possibly only the first time they are called)
- ... other ideas?

simonbrandhorst (approved):

docs/src/DeveloperDocumentation/documentation.md
```diff
@@ -5,6 +5,17 @@ information as possible into the docstrings and only use the doc pages for
...
+!!! note "Previewing the documentation"
+    Once you have created a pull request it is possible to preview the
+    documentation on github using the link
+    https://docs.oscar-system.org/previews/PR<prnumber>/
+    where you insert the number of your PR for `prnumber`. Alternatively you
+    can look at the github actions tab of your PR and click the details link
+    next to the documentation action. These things only work if there are no
+    conflicts with the master branch and the documentation action is
+    successful. You can still build the documentation locally with the commands
+    described below.
```
benlorenz: There are two documenter actions after a successful run, `Documentation / deploy_docs (pull_request)` will have the logs and `documenter/deploy` points to the docs, so I made this more precise.
Also there are a few more conditions for this to work:
```suggestion
    next to the `documenter/deploy` action. There are a few conditions for this 
    to work:
    - No conflicts with the master branch.
    - Documentation action is successful, i.e. no doctest errors.
    - The branch for the PR is in the main `oscar-system/Oscar.jl` repository.
    You can still build the documentation locally with the commands described below.
```

benlorenz (approved):

docs/src/DeveloperDocumentation/experimental.md
lkastner: For the discoverability: I still think, code should not spend too much time in `experimental`, then this would matter less. I don't know whether we want to provide a concrete time frame though.

simonbrandhorst: Code in experimental will be buggy because it is less tested and not battle proof yet. So the naive user should not use it at all. Certainly not to prove a theorem. It is like a preprint which has not been peer reviewed yet - use at your own risk.

Issue #1930, opened 2023-02-13
Orders of matrix groups over arbitrary number field


joschmitt:
**Is your feature request related to a problem? Please describe.**
One cannot compute the order of a matrix group over a field that is not a cyclotomic field:
```
julia> Qx, x = QQ["x"];

julia> K, a = number_field(x^2 + x + 1, "a");

julia> M = matrix(K, 1, 1, [ a ]);

julia> G = matrix_group(M);

julia> order(G)
ERROR: Number field over Rational Field with defining polynomial x^2 + x + 1 is not a cyclotomic field
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] _iso_oscar_gap(FO::AnticNumberField)
...
```

**Describe the solution you'd like**
Make this possible, if it is feasible. Especially since the number field above is exactly `CyclotomicField(3)`, OSCAR just doesn't know about it.

**Describe alternatives you've considered**
If this is not possible/feasible, provide a more informative error message, like "Group orders can only be computed over cyclotomic fields..." + possible information, how to solve the problem.

**Additional context**
I'm not sure how much of an "issue" this actually is since there will always be a big enough cyclotomic field for a given finite group. Feel free to close this issue, if you don't think it is relevant.

fingolfin:
Thanks! It's a good suggestion

fieker:
a pure gap-interface problem as this works:
```
julia> Oscar.isomorphic_group_over_finite_field(G)
(Matrix group of degree 1 over Finite field of degree 2 over F_5, Map from
G to Matrix group of degree 1 over Finite field of degree 2 over F_5 defined by a julia-function with inverse)
```
Same group as in the example above...

joschmitt:
You are right!
The problem appears to be the access to `G.X` in `compute_order`.
```
julia> G.X
ERROR: Number field over Rational Field with defining polynomial x^2 + x + 1 is not a cyclotomic field
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] _iso_oscar_gap(FO::AnticNumberField)
    @ Oscar ~/.julia/dev/Oscar/src/GAP/iso_oscar_gap.jl:249
...
```

fieker:
On Tue, Feb 14, 2023 at 06:05:28AM -0800, Johannes Schmitt wrote:
> You are right!
> The problem appears to be the access to `G.X` in `compute_order`.
> ```
> julia> G.X
[...]

PR #2003, opened 2023-03-02
provide `iso_oscar_gap` and `iso_gap_oscar` for number fields

ThomasBreuer:
- provide `iso_oscar_gap` and `iso_gap_oscar` for number fields that are not of cyclotomic type (addresses the problem from #1930, but does not address the questions raised in its discussion)

- provide docstrings for `Oscar.iso_oscar_gap` and `Oscar.iso_gap_oscar` (had been suggested by Meinolf; it is not clear to me where in the manual these docstrins should appear)

- removed a workaround for certain finite fields, since the problem on the GAP side has been fixed (a test and an assertion make sure that the problem will not come up again)

- extended the tests

- replaced `GAP.Globals` access by `Oscar.GAPWrap` access

### DIFF
--- a/src/GAP/gap_to_oscar.jl
+++ b/src/GAP/gap_to_oscar.jl
@@ -44,7 +44,7 @@ function (F::FinField)(x::GAP.FFE)
         # FFE in GAP only exist for "small" characteristic, so we know the int
         # value fits into an Int; telling Julia about this via a type assertion
         # results in slightly better code
-        val = GAPWrap.INT_FFE_DEFAULT(x)::Int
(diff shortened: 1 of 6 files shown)


fingolfin (approved):

fingolfin:
Thank you.

I would prefer to not add `Oscar.iso_oscar_gap` and `Oscar.iso_gap_oscar` to the manual at this time. Perhaps at some point we'll decide to add a section (or section*s*) for "internal functions", or on one low-level details for the GAP-Oscar bridge, then this would fit it.

PR #2004, opened 2023-03-02
Combinatorics: Add sanity check to add_edge!(Graph)

lkastner:
Previously we would get a crash from polymake, if the source or target of the edge to be added did not exist. Now we just throw an Oscar error.

### DIFF
--- a/src/Combinatorics/Graphs.jl
+++ b/src/Combinatorics/Graphs.jl
@@ -77,9 +77,7 @@ function Graph{T}(nverts::Int64) where {T <: Union{Directed, Undirected}}
     return Graph{T}(pmg)
 end
 
-function _has_node(G::Graph, node::Int64)
-    return 0<node && node<=nv(G)
-end
+_has_node(G::Graph, node::Int64) = 0 < node <= nv(G)
 
 @doc Markdown.doc"""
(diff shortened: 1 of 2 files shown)


afkafkafk13 (approved):
Yes, this is a useful sanity check.

src/Combinatorics/Graphs.jl
```diff
@@ -77,6 +77,9 @@ function Graph{T}(nverts::Int64) where {T <: Union{Directed, Undirected}}
     return Graph{T}(pmg)
 end
 
+function _has_node(G::Graph, node::Int64)
+    return 0<node && node<=nv(G)
+end
```
benlorenz:
```suggestion
_has_node(G::Graph, node::Int64) = 0 < node <= nv(G)
```

test/Combinatorics/Graph.jl
```diff
@@ -112,4 +112,9 @@
         @test length(connected_components(g)) == 1
         @test diameter(g) == 3
     end
+
+    @testset "errors" begin
+        g = graph{Undirected}(1)
```
benlorenz:
```suggestion
        g = Graph{Undirected}(1)
```

Issue #1987, opened 2023-02-26
One finite field to rule them all


thofma:
As many of us are painfully reminded from time to time, we have four (actually five) different finite field types. This is not only a burden for developers, but also for users. Due to heroic efforts of the flint maintainers, there is actually one finite field type (used to be known as `fq_default`) that we could use instead. Although it is already hooked up in Nemo, so far we did not adjust the other packages. I hope that we can do the transition as soon as possible (also in view of *the* book). I am opening this issue to track progress and coordinate our efforts:

- [x] Adjust Hecke (WIP, https://github.com/thofma/Hecke.jl/pull/989)
- [x] Tweak the Singular interface (`singular_coeff_ring` produces a generic `n_Field{}` object, which is bad for performance).
- [x] Adjust the GAP interface (I did some small `iso_oscar_gap` tests and it seems to work.)
- [x] #2897
- [x] <s> Get https://github.com/flintlib/flint2/pull/1261 into our flint2 binaries</s> (I worked around it)
- [x] #2192
- [x] Introduce `Native.GF` and `Native.FiniteField` in Nemo and use it in Hecke where necessary.

Once we have this, we can

- [ ] Make the new type the default. This means making `GF` and `FiniteField` always construct the new type.

P.S.: At the moment, to construct the new type, one can use `Nemo._FiniteField` and `Nemo._GF`. 

Feel free to add more things to the list.

fingolfin:
See also issue #867

ThomasBreuer:
@thofma Concerning the GAP interface:
`Nemo._GF(p)` and `Nemo._GF(p, d)` return objects of type `FqField`, and this type is supported by `iso_oscar_gap`.

`iso_gap_oscar` chooses an Oscar field by calling `GF(p)` or `GF(p, d)`.
Thus everything will be fine as soon as `GF` will return the default fields,
but currently a field of the type `Nemo.fpField` or `fqPolyRepField` is used.
Wouldn't it be better to change `iso_gap_oscar` already now to choose the Oscar field by calling `Nemo._GF`?
Then there will be at most a name change in the end, but no change in the behaviour.

thofma:
It would be an interesting experiment to see what breaks when one changes `iso_gap_oscar` as you suggested. Would be great if we could make a draft PR with that change so we can see the CI run.

ThomasBreuer:
@thofma O.k., I will do this.

thofma:
[...]

PR #2008, opened 2023-03-03
let `iso_gap_oscar` choose `Nemo._GF`

ThomasBreuer:
see #1987 for the background

### DIFF
--- a/src/GAP/iso_gap_oscar.jl
+++ b/src/GAP/iso_gap_oscar.jl
@@ -82,9 +82,9 @@ function _iso_gap_oscar_field_finite(FG::GAP.GapObj)
      if p < ZZRingElem(2)^64
        p = UInt64(p)
      end
-     FO = GF(p)
+     FO = Nemo._GF(p)
    else
-     FO = GF(p, d)
+     FO = Nemo._GF(p, d)
    end
 
    finv, f = _iso_oscar_gap_field_finite_functions(FO, FG)
--- a/src/GAP/iso_gap_oscar.jl
+++ b/src/GAP/iso_gap_oscar.jl
@@ -76,16 +76,7 @@ function _iso_gap_oscar_residue_ring(RG::GAP.GapObj)
 end
 
 function _iso_gap_oscar_field_finite(FG::GAP.GapObj)
-   p = characteristic(FG)  # of type `ZZRingElem`
-   d = GAPWrap.DegreeOverPrimeField(FG)
-   if d == 1
-     if p < ZZRingElem(2)^64
-       p = UInt64(p)
-     end
-     FO = Nemo._GF(p)
-   else
-     FO = Nemo._GF(p, d)
-   end
+   FO = Nemo._GF(characteristic(FG), GAPWrap.DegreeOverPrimeField(FG))
 
    finv, f = _iso_oscar_gap_field_finite_functions(FO, FG)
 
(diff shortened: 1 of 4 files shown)


ThomasBreuer:
That's funny: The test failure means that the finite field that is now chosen by `iso_gap_oscar` behaves better than the previously chosen one; thus I have to change a test in order to really get an error.

fingolfin:
A few doctests need to be adjusted

thofma:
There is no need for the "dance" with `p < 2^64` and `degree == 1` or not. `_GF` and `_FiniteField` do this automatically.

One other thing. The new finite fields come also in a relative version, so that the use of `degree(F)` etc in the `_iso_oscar_gap_field_finite_functions` a bit problematic is:
```julia
julia> F = Hecke.Nemo._GF(2, 2);

julia> Fx, x = F["x"]
(Univariate Polynomial Ring in x over Finite field of degree 2 over F_2, x)

julia> f = x^3 + x + 1
x^3 + x + 1

julia> FF, = Hecke.Nemo._FiniteField(f)
(Finite field of degree 6 over F_2 (over Finite field of degree 2 over F_2), o)

julia> degree(FF)
3

julia> absolute_degree(FF)
6
```
I just noticed that `modulus` is doing the wrong thing:
```julia
julia> modulus(FF)
T^6 + T^4 + T^3 + T + 1

julia> defining_polynomial(FF)
x^3 + x + 1
``` 
(I will fix this in Nemo)

fingolfin:
Woa, `modulus` for finite fields is really weird. As @ThomasBreuer just pointed out to me: it is not really clearly defined what it is (as such: I am happy that @thofma says with confidence it "is doing the wrong thing" as I am not sure what it is supposed to do). But there are more issues. Let's look at one of them, from Nemo:
```
@doc Markdown.doc"""
    modulus(k::fqPolyRepField, var::String="T")

Return the modulus defining the finite field $k$.
"""
function modulus(k::fqPolyRepField, var::String="T")
    p::Int = characteristic(k)
    Q = polynomial(GF(p), [], var)
    P = ccall((:fq_nmod_ctx_modulus, libflint), Ref{fpPolyRingElem},
                 (Ref{fqPolyRepField},), k)
    ccall((:nmod_poly_set, libflint), Nothing,
          (Ref{fpPolyRingElem}, Ref{fpPolyRingElem}),
          Q, P)

    return Q
end
```
So "the modulus defining the finite field $k$ is unclear. I guess a hint that this is a polynomial $f$ such that $parent(f)/(f)$ is isomorphic to $k$  would be helpful (assuming that what it is). In so far it seems consistent:
```
julia> f1 = modulus(FF)
T^6 + T^4 + T^3 + T + 1

julia> parent(f1)
Univariate Polynomial Ring in T over Galois field with characteristic 2

julia> base_ring(parent(f1))
Galois field with characteristic 2

julia> order(coefficient_ring(parent(f1)))
2
```
and
```
julia> f2 = defining_polynomial(FF)
x^3 + x + 1

julia> parent(f2)
Univariate Polynomial Ring in x over Finite field of degree 2 over F_2

julia> order(coefficient_ring(parent(f2)))
4
```
at least seem consistent to me?


But then also `var::String="T"`  raises eyebrows: Why aren't symbols also allowed? And  shouldn't it allow passing in an existing polynomial ring (instead or in addition of the name of a variable)?

ThomasBreuer:
@thofma Thanks for your comments.

Concerning `_iso_oscar_gap_field_finite_functions`, I have now added a check that the Oscar field is an extension of a prime field. (The analogous check for the GAP field had been already there.)

thofma (approved):

thofma:
Thanks @ThomasBreuer

PR #2001, opened 2023-03-01
Properties covered schemes

afkafkafk13:
add some standard properties of covered schemes as attributes
(work in progress, in parts untested)

PR #2009, opened 2023-03-03
Properties covered schemes (cleaned up version of #2001)

afkafkafk13:
add some standard properties of covered schemes as attributes

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
@@ -661,8 +661,11 @@ is_smooth(X::AbsSpec{<:Field, <:MPolyLocRing}) = true
 
 Return the boolean value whether an affine scheme `X` is irreducible.
 
+!!! note
+    Irreducibility is checked over the (computable) base field of the affine scheme as specified upon creation of the ring, not over the algebraic closure thereof.
+
 """
-@attr function is_irreducible(X::AbsSpec)
+@attr function is_irreducible(X::AbsSpec{<:Field, <:MPolyAnyRing})
   !is_empty(X) || return false
   !get_attribute(X, :is_integral, false) || return true 
                                            ## integral = irreducible + reduced
@@ -672,11 +675,10 @@ end
 @doc Markdown.doc"""
    is_integral(X::AbsSpec)
 
-Return the boolean value whether an affine scheme `X` is integral,
-i.e. irreducible and reduced.
+Return the boolean value whether an affine scheme `X` is integral, i.e. irreducible and reduced.
 
 """
-@attr function is_integral(X::AbsSpec)
+@attr function is_integral(X::AbsSpec{<:Field, <:MPolyAnyRing})
   !is_empty(X) || return false
   if has_attribute(X,:is_reduced) && has_attribute(X,:is_irreducible)
      return get_attribute(X,:is_reduced) && get_attribute(X,:is_irreducible)
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
@@ -664,11 +664,9 @@ Return the boolean value whether an affine scheme `X` is irreducible.
 """
 @attr function is_irreducible(X::AbsSpec)
   !is_empty(X) || return false
-  if get_attribute(X, is_integral, false)
-## integral = irreducible + reduced
-     return true
-  end 
-  return length(minimal_primes(modulus(OO(X))))==1
+  !get_attribute(X, :is_integral, false) || return true 
+                                           ## integral = irreducible + reduced
+  return (length(minimal_primes(modulus(OO(X)))) == 1)
 end
 
 @doc Markdown.doc"""
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
@@ -659,7 +659,7 @@ is_smooth(X::AbsSpec{<:Field, <:MPolyLocRing}) = true
 @doc Markdown.doc"""
    is_irreducible(X::AbsSpec)
 
-Return the boolean value whether an affine scheme `X` is irreducible.
+Return whether the affine scheme `X` is irreducible.
 
 !!! note
     Irreducibility is checked over the (computable) base field of the affine scheme as specified upon creation of the ring, not over the algebraic closure thereof.
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
@@ -666,7 +666,7 @@ Return the boolean value whether an affine scheme `X` is irreducible.
   !is_empty(X) || return false
   !get_attribute(X, :is_integral, false) || return true 
                                            ## integral = irreducible + reduced
-  return (length(minimal_primes(modulus(OO(X)))) == 1)
+  return (length(minimal_primes(saturated_ideal(modulus(OO(X))))) == 1)
 end
 
 @doc Markdown.doc"""
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
@@ -700,5 +700,5 @@ Return the boolean value whether an affine scheme `X` is connected.
 ## note for future implementation: expensive property
 ## 1) do primary decomposition
 ## 2) check connectedness of lowest two layers of the intersection lattice
-  return(get_attribute(X,:is_connected))
+  return get_attribute(X,:is_connected)
 end
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
@@ -551,14 +551,13 @@ end
 ##############################################################################
 # (6) Check, if scheme is reduced
 ##############################################################################
-# TODO: projective schemes, covered schemes
+# TODO: projective schemes
 
 @doc Markdown.doc"""
    is_reduced(X::AbsSpec{<:Field, <:MPolyAnyRing})
 
-Return the boolean value whether a scheme `X` is reduced.
+Return the boolean value whether an affine scheme `X` is reduced.
 
-Currently, this command is available for affine schemes and space germs.
 """
 @attr function is_reduced(X::AbsSpec{<:Field, <:MPAnyQuoRing})
   I = saturated_ideal(modulus(OO(X)))
@@ -575,15 +574,12 @@ end
 # The routine checks whether the module for the cotangent sheaf Ω¹(X)
 # is locally free over 𝒪(X) and returns `true` if this is the case. 
 ########################################################################
-# TODO: Covered schemes, projective schemes
 # TODO: is_regular using Hironaka's criterion
 
 @doc Markdown.doc"""
     is_smooth(X::AbsSpec{<:Field, <:MPolyAnyRing})
 
 Return whether a scheme `X` is smooth.
-
-Currently this command is available for affine schemes and space germs.
 
 Note that smoothness and regularity do not coincide over non-perfect fields. 
 Smoothness implies regularity, but regular non-smooth schemes exist.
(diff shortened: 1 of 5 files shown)


src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -654,6 +650,53 @@ end
...
+   is_irreducible(X::AbsSpec)
+
+Return the boolean value whether an affine scheme `X` is irreducible.
+
+"""
+@attr function is_irreducible(X::AbsSpec)
+  !is_empty(X) || return false
+  !get_attribute(X, :is_integral, false) || return true 
+                                           ## integral = irreducible + reduced
+  return (length(minimal_primes(saturated_ideal(modulus(OO(X))))) == 1)
```
simonbrandhorst, line 672: What if the base ring is not an integral domain? or a field? is this correct? throw an error? 
Looks to me like we should be able to ask a ring whether it is an integral domain. 
Then the ring should know best what to do.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -654,6 +650,53 @@ end
...
+@doc Markdown.doc"""
+   is_integral(X::AbsSpec)
+
+Return the boolean value whether an affine scheme `X` is integral,
+i.e. irreducible and reduced.
+
+"""
 @attr function is_integral(X::AbsSpec)
-  return Oscar._is_integral_domain(OO(X))
+  !is_empty(X) || return false
```
simonbrandhorst, line 682: for speed only?

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -1,29 +1,112 @@
...
-  return is_connected(X) && all(U->(is_integral(U)), affine_charts(X))
+  !is_empty(X) || return false
+  return is_reduced(X) && is_irreducible(X)
 end
 
+# auxilliary function for connectedness of glueing graph
+#      do not confuse with connectedness of the scheme
+# Note: This does not work with glueing_graph, because empty patches
+#      need to be ignored without changing the covering
+@attr function is_connected_glueing(X::AbsCoveredScheme)
```
simonbrandhorst, line 73: please provide a docstring including a definition of this property.

src/AlgebraicGeometry/Schemes/Covering/Objects/Methods.jl
```diff
@@ -68,6 +69,25 @@ function update_glueing_graph(C::Covering)
...
 
+## prune the covering by throwing away empty charts and then compute
+## the glueing graph afterwards 
+## (relevant data for connectedness of glueing)
+function pruned_glueing_graph(C::Covering)
+  v = findall(U->!is_empty(U), C.patches)
+  m = length(v)
+  gt = Graph{Undirected}(m)
+  for (X, Y) in keys(glueings(C))
+    i = findfirst(==(C[X]),v)
```
simonbrandhorst, line 80:
```suggestion
    i = findfirst(===(C[X]),v)
```
? so that no expensive computation is triggered? not sure if this could trigger one or not.

src/AlgebraicGeometry/Schemes/Covering/Objects/Methods.jl
```diff
@@ -68,6 +69,25 @@ function update_glueing_graph(C::Covering)
...
+    !isnothing(i) || continue
+    j = findfirst(==(C[Y]),v)
+    !isnothing(j) || continue
+    (U, V) = glueing_domains(C[X,Y])
+    is_dense(U) && add_edge!(gt, i, j)
+    is_dense(V) && add_edge!(gt, j, i)
```
simonbrandhorst, line 86: what happens here? shouldn't we just extract the subgraph consisting of non-empty glueings? 
Then why do we ask for something to be dense. Probably I do not understand the purpose of this function.

simonbrandhorst (approved):
O.k. once the comments are addressed. And some of the functions are restricted to schemes over fields only.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
afkafkafk13: We have to look at two different settings:
A) For the given ring OO(X) minimal_primes/primary_decomposition/is_integral_domein is implemented, then the problem is solved on the algebraic side, i.e. we obtain a correct answer for the given ring. In particular, the answer will be over the given field k for k[x_1,..,x_n]/I and not over the algebraic closure, but this is the way it should be for a scheme.
B) For the given ring OO(X) these functions are not available, then the algebra side will throw an error.

But I agree that I should put a comment in the documentation pointing out the subtlety of the underlying field.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
afkafkafk13: No, the point was to use the function is_prime, which is officially available, instead of an internal one which calls is_prime anyway.

By the way, the respective function in src/Rings/... stayed untouched.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
afkafkafk13: I did not intend to publish this function, it is more for internal use.

src/AlgebraicGeometry/Schemes/Covering/Objects/Methods.jl
afkafkafk13: No, C[X] is an int64.

src/AlgebraicGeometry/Schemes/Covering/Objects/Methods.jl
afkafkafk13: This line is from the original version, which I only changed to handle empty patches properly. I understand it as a sanity check that we are really glueing along opens.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
simonbrandhorst: The docstring is not for the user. It is for me so that I may understand what this function does. 
You can also do that as comments in the code, but a docstring is accessible from the repl and comments are not.

src/AlgebraicGeometry/Schemes/Covering/Objects/Methods.jl
simonbrandhorst: okay.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
afkafkafk13: - now restricted to schemes of MPolyAnyRing
- docstring added

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
afkafkafk13: docstring done

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -654,6 +650,55 @@ end
...
+
+Return the boolean value whether an affine scheme `X` is connected.
+
+"""
+@attr function is_connected(X::AbsSpec)
+  has_attribute(X,:is_connected) || error("not implemented yet")
+## note for future implementation: expensive property
+## 1) do primary decomposition
+## 2) check connectedness of lowest two layers of the intersection lattice
+  return(get_attribute(X,:is_connected))
```
fingolfin:
```suggestion
  return get_attribute(X,:is_connected)
```

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -654,6 +650,55 @@ end
...
+# Connectedness                                                   #
+###################################################################
+@doc Markdown.doc"""
+   is_connected(X::AbsSpec)
+
+Return the boolean value whether an affine scheme `X` is connected.
+
+"""
+@attr function is_connected(X::AbsSpec)
+  has_attribute(X,:is_connected) || error("not implemented yet")
```
fingolfin: But this line is *only* executed when the `is_connected` attribute is not set. So you might as well do this:
```suggestion
  error("not implemented yet")
```

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -1,29 +1,122 @@
...
 
+# auxilliary function for connectedness of glueing graph
+#      do not confuse with connectedness of the scheme
+# Note: This does not work with glueing_graph, because empty patches
+#      need to be ignored without changing the covering
+@doc Markdown.doc"""
+   is_connected_glueing(X::AbsCoveredScheme)
+
+Return the boolean value whether the glueing graph of the default
+covering of the scheme X is connected. .
```
fingolfin:
```suggestion
covering of the scheme X is connected.
```

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -654,6 +650,55 @@ end
...
+###################################################################
+# Irreducibility and Integrality                                  #
+#    integral iff (irreducible && reduced)                        #
+#    integral = OO(X) is integral domain                          #
+#    irreducible = nilradical of OO(X) is prime                   #
+###################################################################
+@doc Markdown.doc"""
+   is_irreducible(X::AbsSpec)
+
+Return the boolean value whether an affine scheme `X` is irreducible.
```
fingolfin: This is rather subjective but "the boolean value" seems a bit redundant. Also, we are talking about a concrete object. So I'd use this:
```suggestion
Return whether the affine scheme `X` is irreducible.
```
That said, this is merely a suggestion, and also not about anything important. So I am totally fine with merging this as-is. We can at some future point decide if we want to have someone go over all our docstrings to make this kind of thing a bit more consistent across various parts of the documentation.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -1,29 +1,122 @@
...
 @attr function is_smooth(X::AbsCoveredScheme)
   if !isdefined(X, :coverings) 
     return true
   end
-  # TODO: This can be optimized! We only need to check the rank of 
+  # TODO: Can this be optimized? We only need to check the rank of
   # the jacobian matrices where we haven't checked in another chart before. 
+  # This is a tradeoff between cost of Jacobian criterion and cost of
+  # optimizing the use of the covering
   return all(x->is_smooth(x), affine_charts(X))
```
fingolfin, line 41: More idiomatic would be this:
```suggestion
  return all(is_smooth, affine_charts(X))
```

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -1,29 +1,122 @@
...
+@doc Markdown.doc"""
+   is_irreducible(X::AbsCoveredScheme)
+
+Return the boolean value whether a covered scheme `X` is irreducible.
+
+"""
+@attr function is_irreducible(X::AbsCoveredScheme)
+  is_connected_glueing(X) || return false
+  !is_empty(X) || return false
+  v=findall(U->!is_empty(U), affine_charts(X))  ## only check non-empty patches
```
fingolfin, line 120:
```suggestion
  v=findall(!is_empty, affine_charts(X))  ## only check non-empty patches
```

src/AlgebraicGeometry/Schemes/Covering/Objects/Methods.jl
```diff
@@ -68,6 +69,25 @@ function update_glueing_graph(C::Covering)
   return gg
 end
 
+## prune the covering by throwing away empty charts and then compute
+## the glueing graph afterwards 
+## (relevant data for connectedness of glueing)
+function pruned_glueing_graph(C::Covering)
+  v = findall(U->!is_empty(U), C.patches)
```
fingolfin, line 76:
```suggestion
  v = findall(!is_empty, C.patches)
```

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -1,29 +1,122 @@
...
+   is_empty(X::AbsCoveredScheme)
+
+Return the boolean value whether a covered scheme `X` is empty.
+
+"""
 @attr function is_empty(X::AbsCoveredScheme)
   if !isdefined(X, :coverings) 
     return true
   end
   return all(x->isempty(x), all_patches(default_covering(X)))
```
fingolfin, line 21:
```suggestion
  return all(isempty, all_patches(default_covering(X)))
```

fingolfin (approved):
Some minor nitpicks, feel free to ignore

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
afkafkafk13: If someone (or some other function) set the attribute, we want to return it. So the line is precisely, what was intended.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
afkafkafk13: Ok, I will keep this in mind for next time, but not change it in this PR (also holds for the analogous suggestions below).

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
fingolfin: If someone else sets the attribute, this code won't be reached. That's the point pf `@attr` !

afkafkafk13: Thanks for that pointer. This is precisely, what was intended.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
fieker, line 21: `all(isempty, all_patches...)`
Isempty is already a function

afkafkafk13: Ah, that explanation helps. But there are so many of these in the file (and many not even touched by this PR) that I shall do these changes in a separate PR, which does not introduce new functionality.

PR #2010, opened 2023-03-03
Stop using @gapwrap

fingolfin:
No deep reason other than that anybody trying to understand this bit of the code would be confused as to why `@gapwrap` exists in addition to the `GAPWrap` module (the answer of course is "historic reasons").


src/GAP/wrappers.jl
```diff
@@ -175,6 +177,8 @@ GAP.@wrap PrimePGroup(x::GapObj)::GapInt
 GAP.@wrap Range(x::GapObj)::GapObj
 GAP.@wrap ReduceCoeffs(x::GapObj, y::GapObj)
 GAP.@wrap RelatorsOfFpGroup(x::GapObj)::GapObj
+GAP.@wrap SetMaximalAbelianQuotient(x::Any, y::Any)::Bool
+GAP.@wrap SetSize(x::Any, y::Any)::Bool
```
ThomasBreuer: `::Nothing` not `::Bool`

src/Groups/GAPGroups.jl
```diff
@@ -192,8 +192,8 @@ end
 
 order(x::Union{GAPGroupElem, GAPGroup}) = order(ZZRingElem, x)
 
-@gapwrap has_order(G::GAPGroup) = GAP.Globals.HasSize(G.X)::Bool
-@gapwrap set_order(G::GAPGroup, val::T) where T<:IntegerUnion = GAP.Globals.SetSize(G.X, GAP.Obj(val))
+has_order(G::GAPGroup) = GAPWrap.HasSize(G.X)::Bool
+set_order(G::GAPGroup, val::T) where T<:IntegerUnion = GAPWrap.SetSize(G.X, GAP.Obj(val))::nothing
```
ThomasBreuer: The types of the results have been defined in the `GAP.@wrap` calls, I think they need not be specified here.
(In any case, write `::Nothing` not `::nothing`.)

ThomasBreuer (changes requested):
This is in principle o.k., just there are a few typos.

ThomasBreuer (approved):

PR #2013, opened 2023-03-06
Fix warning about coordinate_ring

fingolfin:
Fixes this warning:

    WARNING: using Geometry.coordinate_ring in module Oscar conflicts
    with an existing identifier.

### DIFF
--- a/src/AlgebraicGeometry/Miscellaneous/basics.jl
+++ b/src/AlgebraicGeometry/Miscellaneous/basics.jl
@@ -9,7 +9,7 @@ import Oscar: weights
 import AbstractAlgebra, Nemo
 import Base: ==, show, hash
 
-export proj_space, normalize!, coordinate_ring
+export proj_space, normalize!
 
 struct ProjSpc{T}  <: Oscar.ProjSpc{T}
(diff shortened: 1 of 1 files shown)


benlorenz (approved):

fingolfin:
I just discussed this with @afkafkafk13: it seems the `coordinate_ring` in question, in file `src/AlgebraicGeometry/Miscellaneous/basics.jl`, is for type `ProjSpc` and goes back to 56d4403336c20afcd18d4efd1f1f4230e89845bd by @fieker. It is only used in code in the experimental dir (in `experimental/PlaneCurve`).

This method apparently yields something different mathematically than `coordinate_ring(X::AbsSpec)`, and so "merging" the two as this PR currently does is probably a bad idea. In general we need to think hard about `ProjSpc` in general...

But for now what this all means is that this PR should instead drop the `export` of `coordinate_ring`, so that the two functions remain separate and the warning is fixed. Then someone else (@HechtiDerLachs perhaps?) can later take care of properly addressing the projective space code (whatever that means :joy:).


Also CC @simonbrandhorst

afkafkafk13 (approved):

PR #2015, opened 2023-03-07
Minor changes/extensions in toric geometry

HereAround:
* Typo in one example (constructor called with argument `p::Polyhedron` rather than just `p`).
* For convenience, provide constructor `rational_equivalence_class` which directly accepts a polynomial in the Chow ring as argument.


afkafkafk13 (approved):

HereAround:
Thank you @afkafkafk13 .

Issue #1971, opened 2023-02-23
New rule: `export` only one identifier at a time


fingolfin:
I propose the following new rule for our code (and the dev docs): "Every `export` statement must export exactly one identifier. If multiple export statements appear in sequence, they must be sorted alphabetically.".

So this:
```julia
export foo, bar,
    qux
```
would become
```julia
export bar
export foo
export qux
```

If this seems to harsh, then I propose this relaxation: "Every `export` statement must be on a single line." Then we'd have:
```julia
export foo, bar, qux
```
or
```julia
export foo, bar
export qux
```

Rationale: Earlier today I spent 10 minutes trying to track down the place where a function was exported. And that wasn't the first time I was in this situation. I could not just do something like `git grep export.*SYMBOL` to find it because `export` and `SYMBOL` did not appear on the same line. (Yes, one can use fancier search methods to still find it, and I eventually did, but I a tempted to believe that many people would have had an even harder time than me, as they are not as proficient with regex and grep and similar tools.)

So one of the rules above would make such searches much easier. I can't think of a major downside other than subjective ones... ("This wastes spaces/lines, my computer is almost out of spare newlines! All those extra bytes will flood my hard disk! What should we do with all those commas this frees up?"). Likely that just means my imagination is too limited as usual, so I'll wait to hear what objections / better alternatives people have.

(I'd suggest a similar rules for imports, but we already moved most imports into a single file, so it's not that bad.)

thofma:
Pretty please, let us do this. (Maybe also exports only in one place, but I do not want to sidetrack this issue here.)

ThomasBreuer:
As an experiment, I have changed the `export` statements in the `src/Groups` files (see pull request #2007).
I am in favour of the relaxed version which admits several exported names on one line;
this (together with the rule that the names should be ordered alphabetically) seems to be natural in the situation of properties together with tester and setter, as in `export isfinite, has_is_finite, set_is_finite`.

Another place where several names on one line make sense is in `export GL, GO, GU, SL, SO, Sp, SU`.
(I am not sure whether we want to keep these names at all.)

HereAround:
I collected the exports for `AlgebraicGeometry` and `Combinatorics` in https://github.com/oscar-system/Oscar.jl/pull/2016.

HereAround:
[...]

PR #2016, opened 2023-03-07
Collect export

HereAround:
Cf. https://github.com/oscar-system/Oscar.jl/issues/1971. This is just a start, by trying to achieve this for all the code in `AlgebraicGeometry` and (since it turned out to be simple enough) also `Combinatorics`.


afkafkafk13:
I understand the intention of this cleanup, but now the export statements of several Geometry subpackages ended up all in one file without the information which package / subdirectory each originates from. For the future this is good, but:

Did you find any name clashes etc while doing this? 

If so, we should save this information to resolve the name clashes (by changing names or by seeing that the functions have the same purpose in different settings).

HereAround:
Yes, functions with the same name but different signature do exist. I had a little script do (most of) the work in this PR. Among others, this sorted the exports alphabetically and remove duplicates. So I cannot give you a complete list of "clashes". At least not now. Still, you can rest assured that `dim` is part of that list.

I am skeptical towards keeping track of "clashes":
1. Such a list is likely outdated the moment I create it.
2. Someone has to maintain it.
3. There are other ways to find out about such "clashes", e.g. just run `git grep dim` or `git grep "dim("` or `git grep "function dim"` etc. within `OSCAR`. This seems much more convenient for all parties involved.

Besides, most likely not all of these "clashes" do need attention. An instance that would require attention -- most likely the ones that you are thinking of -- is for instance if we ever intend to have `hirzebruch_surface` or `del_pezzo_surface` as schemes. The toric counterparts already exist. The protoype for handling this is `projective_space`. We can then easily mimic this procedure.

Let me ping @fingolfin who originally raised this issue, and @thofma who also seemed interested.

Note also that @ThomasBreuer accomplished a similar task for `src/Groups` in https://github.com/oscar-system/Oscar.jl/pull/2007

afkafkafk13:
Actually, I did not think of a list to keep updated forever. I thought more of a 'heads up' to the teams so that everyone is reminded of this, has a short look at the differences and proceeds as they see appropriate (i.e. ignoring the majority which is without problems like 'dim', taking care of the few bad ones). 

My idea was in the sense of: While doing a 'uniq' on the export list anyway, let us know about the 'diff' before and after, because this is just one additional command -- assuming that you were scripting this.

HereAround:
> Actually, I did not think of a list to keep updated forever. I thought more of a 'heads up' to the teams so that everyone is reminded of this, has a short look at the differences and proceeds as they see appropriate (i.e. ignoring the majority which is without problems like 'dim', taking care of the few bad ones).
> 
> My idea was in the sense of: While doing a 'uniq' on the export list anyway, let us know about the 'diff' before and after, because this is just one additional command -- assuming that you were scripting this.

My script is very elementary and quite a few tasks need to be performed by hand. Since you seem to care deeply, I can try to give you a list. But this will take time. Maybe I get to it late in the evening today.

afkafkafk13 (approved):

HereAround:
For the record, @afkafkafk13 and I talked yesterday and agreed that we are happy to proceed *without* the list of duplicate exports.

PR #2018, opened 2023-03-08
Basic (and not so basic) functionality for linear quotients

joschmitt:
Provide some functionality for linear quotients, that is, spectra of invariant rings by finite groups, most prominently a function to compute their Cox rings and the Cox ring of a QQ-factorial terminalization using an algorithm by Ryo Yamagishi.
This is code I wrote as part of my PhD project and I think it is better kept in OSCAR/experimental than on a shelf in the graduate school office...
I hope that I can work further on this in the next months to make it stable enough to move it out of experimental:

- The interface for the linear quotients themselves should probably be aligned to the affine schemes (because that's what a linear quotient is).
- There are some functions in `LinearQuotients/misc.jl`, which don't have anything particular to do with linear quotients, so one can hopefully move them somewhere else.
- I would like to improve Yamagishi's algorithm, but that's mostly an open research problem.


joschmitt:
Just to clarify, I would like to have this reviewed (and possibly merged) like this. I plan to work further on this, but not in this PR.
However, there probably will be a small merge conflict with #2025 .

thofma:
Who would be a good candidate to review this (apart from the technical side)?

joschmitt:
> Who would be a good candidate to review this (apart from the technical side)?

Well, I guess one can file it under algebraic geometry / commutative algebra.

HereAround:
First things first: To me, the code looks good. I have no complaints, but...

We had issues with the CI over the last couple of hours. Hence, I re-triggered the runs. The failures are (most likely) not linked to this PR. Please rebase to the current master. I would expect that then all tests pass. I will wait until they pass for a more detailed review.

Apparently singularities matter to this PR. So let me ping our "local" expert @afkafkafk13. She should also be able to give pointers towards future integration within the (affine) schemes.

@lkastner: What about the (toric) `cyclic_quotient_singularities`? My recollection is, that in the early days of coding the toric varieties, you mentioned that more code on the subject would eventually arrive. Certainly, you did not have this PR in mind, but still - there is a connection/overlap, right?

HereAround (approved):
@joschmitt I am not an expert on the subject. Still, to the extend that I can tell, this PR is good to go. Thank you!

(Nitpicks, per usual, include higher code coverage and documentation. The latter in particular once this migrates into `/src`. But then, you knew this already...)

I suggest we wait a bit more (at most until early next week) to see if @afkafkafk13 or @lkastner have more substantial feedback.

experimental/LinearQuotients/cox_rings.jl
```diff
@@ -0,0 +1,606 @@
...
+function cox_ring_of_qq_factorial_terminalization(L::LinearQuotient; verbose::Bool = false)
+  # TODO: can we handle this without error (in a type-stable way)?
+  @assert !has_terminal_singularities(L) "Variety is already terminal"
+
+  G = group(L)
+  is_subgroup_of_sl(G) || throw(Hecke.NotImplemented())
+
+  # Compute the Cox ring of L itself
+  RVG, RVGtoR = cox_ring(L)
+  verbose && @info "Computed K[V]^[G, G]"
```
fingolfin, line 301: I guess this would be a good use case for the `@vprint` macro (I am not asking for this to be changed in this PR! but you may want to look into that for future work on this or other code)

joschmitt: I forgot/didn't know about this. Thank you!

joschmitt:
> @joschmitt I am not an expert on the subject. Still, to the extend that I can tell, this PR is good to go. Thank you!
> 
> (Nitpicks, per usual, include higher code coverage and documentation. The latter in particular once this migrates into `/src`. But then, you knew this already...)
> 
> I suggest we wait a bit more (at most until early next week) to see if @afkafkafk13 or @lkastner have more substantial feedback.

Sounds good! Thanks!

afkafkafk13:
I did not go into the details of the background, because this has probably already been done for the referee reports of the thesis. 
The code itself looks good to me. 

For the future, you might want to rethink which parts of the documentation should move to some .md file and which should stay as comment in the file.

afkafkafk13 (approved):

joschmitt:
> I did not go into the details of the background, because this has probably already been done for the referee reports of the thesis.

Well, I did not submit the thesis quite yet... I would like to "publish" the code like this, so that in theory the referees have the possibility to look at it.

afkafkafk13:
@joschmitt: I resolved one trivial merge conflict in the bib-file before enabling automerge.

joschmitt:
Thank you!

lkastner:
Ok, since this already has been merged, giving my detailed remarks does not really make sense anymore.

However, the code is not clean. Please read the [style guide](https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/) and adjust accordingly. Functions are too long (`cox_ring` seems to be more than 100 lines) and nesting level is too high, so it will be very hard for someone else than the original author to maintain.

Furthermore there are some useful Julia commands to replace some loops. To my mind come immediately:
```
all, filter, findfirst
```
E.g. a for loop that just checks whether all elements of a list satisfy something and if not, sets a bool to false can and should be replaced by `all`.

joschmitt:
I agree that `cox_ring` is horrible to read, but I don't see a way to split it up. (When I started programming, I learned the rule that no function, or in emergencies no major block, should be longer than the height of the screen; at least the emergency-variant of that is fulfilled.)
In what other way do I violate the style guide? If you insist, I can clear up the `for i = 1:n` and my odd habit of adding spaces in array brackets (`[ 1, 2 ]`).

I went through the code and replaced loops by `all`, `any`, etc. as you suggested; I will push this in the next weeks, hopefully with more fundamental changes/additions.
I personally find several short lines easier to read as one long line, so I often forget about those functions.

lkastner:
For experimental everything is fine, so it can be published there as is. I hope that is ok for your thesis plans.

However at some point this code should be migrated to `src` and maybe also be maintained by someone else. I suggested the `all` etc loops, since probably Julia can optimize them in a better way than the for loops (@fingolfin will correct me if I am wrong). Some other things:
- There is a very high nesting level in some places. I have seen at least four `for` loops in each other.
- There is direct access to members via the `object.property_name` notation. There is no point on this in the style guide, but as far as I know it is generally discouraged, again @fingolfin correct me if I am wrong. Otherwise I'll add this to the style guide.
- You already have comments like
  ```
  # Find a basis of common eigenvectors for the action of gensA on the degree d
  ```
  in the code. Would it not make sense e.g. to move the content of the loop to a different function called `_basis_of_common_eigenvectors`? Or some similar name. It is the same problem in papers, if the proof is too long it becomes unreadable and needs to be broken down into lemmata.

PR #2023, opened 2023-03-09
unified behaviour of blow_up (in scheme context)

afkafkafk13:
blow_up now always returns a BlowupMorphism -- independent of whether it was called with a pair of AbsSpec and Ideal or a pair of AbsCoveredScheme and Ideal Sheaf.

Old behaviour is still needed as internal functionality on charts and has been renamed.


afkafkafk13:
@fingolfin or @thofma: Do I interpret the CI-run output correctly, if I deduce that the problem is not in my courtyard?

thofma:
Yes, it is not related to the changes here.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -581,6 +576,30 @@ function _compute_projective_glueing(gd::CoveredProjectiveGlueingData)
...
+    X::AbsSpec{<:Any, <:MPolyAnyRing},
+    I::MPolyAnyIdeal)
+
+  R = OO(X)
+  R == base_ring(I) || error("I is not an ideal in the coordinate ring of X")
+  ## prepare trivially covered scheme and ideal sheaf on it as input for blow_up(I::IdealSheaf)
+  C = Covering([X])
+  XX = CoveredScheme(C)
+  Isheaf = ideal_sheaf(XX,X,gens(I))
+  return(blow_up(Isheaf))
```
fingolfin:
```suggestion
  return blow_up(Isheaf)
```

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -581,6 +576,30 @@ function _compute_projective_glueing(gd::CoveredProjectiveGlueingData)
...
+    blow_up(X::AbsSpec,I::Ideal)
+
+Return the blow-up morphism of blowing up X at I in OO(X).
+"""
+function blow_up(
+    X::AbsSpec{<:Any, <:MPolyAnyRing},
+    I::MPolyAnyIdeal)
+
+  R = OO(X)
+  R == base_ring(I) || error("I is not an ideal in the coordinate ring of X")
```
fingolfin: Maybe this:
```suggestion
  @req R == base_ring(I) "I must be an ideal in the coordinate ring of X"
```

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -581,6 +576,30 @@ function _compute_projective_glueing(gd::CoveredProjectiveGlueingData)
   return ProjectiveGlueing(G, PUVtoP, QVUtoQ, fup, gup, check=false)
 end
 
+
+@Markdown.doc """
+    blow_up(X::AbsSpec,I::Ideal)
```
fingolfin:
```suggestion
    blow_up(X::AbsSpec, I::Ideal)
```

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -581,6 +576,30 @@ function _compute_projective_glueing(gd::CoveredProjectiveGlueingData)
...
+
+Return the blow-up morphism of blowing up X at I in OO(X).
```
fingolfin:
```suggestion
Return the blow-up morphism of blowing up `X` at `I` in `OO(X)`.
```

experimental/Schemes/CartierDivisor.jl
```diff
@@ -269,3 +269,20 @@ end
...
+  # If there is a simplified covering, use it!
+  covering = C.C
+  n = npatches(covering)
+  i=1
+  println("Effective Cartier Divisor on Covered Scheme with ",n," Charts:\n")
+
+    ID = IdDict{AbsSpec, Ideal}()
+
+  for U in patches(covering)
+    println("Chart ",i,":\n\t", I(U),"\n")
```
fingolfin: I would recommend using interpolation more, and avoid `\n` in favor of additional `println` calls:
```suggestion
    println("Chart $i:")
    println("\t$(I(U)))   # resp. with spaces as we discussed
    println()
```
This way it is obvious that (at least) three lines are printed

experimental/Schemes/IdealSheaves.jl
```diff
@@ -698,3 +700,21 @@ end
...
+  i=1
+  println("Ideal Sheaf on Covered Scheme with ",n," Charts:\n")
+
+
+  ID = IdDict{AbsSpec, Ideal}()
+
+  for U in patches(covering)
+    println("Chart ",i,":\n\t", I(U),"\n")
+    i=i+1
+  end
```
fingolfin: You could use `enumerate` to make this more elegant:
```suggestion
  for (i,U) in enumerate(patches(covering))
    println("Chart ",i,":\n\t", I(U),"\n")
  end
```

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -278,3 +278,100 @@ function maps_with_given_codomain(phi::CoveringMorphism, V::AbsSpec)
...
+        ideal in Localization of Quotient of Multivariate Polynomial Ring in (s1//s0), (s2//s0), x, y, z over Rational Field by ideal(-(s1//s0)*z + (s2//s0)*y, (s2//s0)*x - z, (s1//s0)*x - y) at the multiplicative set powers of QQMPolyRingElem[1] generated by [x]
+
+Chart 2:
+        ideal in Localization of Quotient of Multivariate Polynomial Ring in (s0//s1), (s2//s1), x, y, z over Rational Field by ideal((s2//s1)*y - z, -(s0//s1)*z + (s2//s1)*x, -(s0//s1)*y + x) at the multiplicative set powers of QQMPolyRingElem[1] generated by [y]
+
+Chart 3:
+        ideal in Localization of Quotient of Multivariate Polynomial Ring in (s0//s2), (s1//s2), x, y, z over Rational Field by ideal(-(s1//s2)*z + y, -(s0//s2)*z + x, -(s0//s2)*y + (s1//s2)*x) at the multiplicative set powers of QQMPolyRingElem[1] generated by [z]
+```
+"""
+function show_details(Bl::BlowupMorphism; more::Bool = false)
```
fingolfin: Maybe consider adding an optional `io::IO` argument, just like `show`, so that one can conveniently print this data into strings, files, etc.
```suggestion
function show_details(Bl::BlowupMorphism; more::Bool = false)
  show_details(stdout, Bl; more)
end

function show_details(io::IO, Bl::BlowupMorphism; more::Bool = false)
```
and then of course change all calls to `println` to `println(io, ...)`

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -278,3 +278,100 @@ function maps_with_given_codomain(phi::CoveringMorphism, V::AbsSpec)
...
+  n1 = npatches(C1)
+
+## data of the blowing up itself
+  ED = exceptional_divisor(Bl)
+  C_X0 = Bl.center
+
+## create the output
+  println("Blow up of a Covered Scheme with ",n0," Charts at a ",dim(C_X0),"-dimensional Center leading to a Covered Scheme with ",n1," Charts.\n")
+
+  if(more)
```
fingolfin:
```suggestion
  if more
```

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -278,3 +278,101 @@ function maps_with_given_codomain(phi::CoveringMorphism, V::AbsSpec)
...
+
+julia> A3 = Spec(R);
+
+julia> I = ideal(R,[x,y,z]);
+
+julia> bl = blow_up(A3,I)
+BlowupMorphism{CoveredScheme{QQField}}(relative projective scheme over covered scheme with 1 affine patches in its default covering, covered scheme with 1 affine patches in its default covering, sheaf of ideals on covered scheme with 1 affine patches in its default covering, #undef, #undef, Oscar.EffectiveCartierDivisor{CoveredScheme{QQField}}(covered scheme with 3 affine patches in its default covering, sheaf of ideals on covered scheme with 3 affine patches in its default covering, Covering with 3 patches, #undef))
+
+julia> show_details(bl)
+Blow up of a Covered Scheme with 1 Charts at a 0-dimensional Center leading to a Covered Scheme with 3 Charts.
```
simonbrandhorst: It looks to me like BlowupMorphism is missing a `Base.show` method. Could you please provide it? 
Its output should be similar to the output of `show_details(bl,more=false)`. (If the dimension of the center is not known, we can omit it and not trigger its computation)

simonbrandhorst: It seems to me that your show details functions (with `more=false`) should rather be the way we print stuff, because surely they are more useful than what we have.

experimental/Schemes/IdealSheaves.jl
afkafkafk13: thank you. Changes made accordingly at all occurrences of this PR

experimental/Schemes/CartierDivisor.jl
afkafkafk13: thank you. Changes made accordingly at all occurrences of this PR

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -236,21 +236,16 @@ function empty_covered_projective_scheme(R::T) where {T<:AbstractAlgebra.Ring}
...
 
-@Markdown.doc """
-    blow_up(W::AbsSpec, I::Ideal)
 
-Return the blowup of ``W`` at the ideal ``I``; this is a `ProjectiveScheme` 
-with `base_scheme` ``W``.
-"""
-function blow_up(W::AbsSpec, I::Ideal; var_name::String="s")
-  error("method `blow_up` not implemented for arguments of type $(typeof(W)) and $(typeof(I))")
+function blow_up_chart(W::AbsSpec, I::Ideal; var_name::String="s")
```
simonbrandhorst, line 249: Even though it is internal please document it. Not for the user but for us.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -581,6 +576,30 @@ function _compute_projective_glueing(gd::CoveredProjectiveGlueingData)
   return ProjectiveGlueing(G, PUVtoP, QVUtoQ, fup, gup, check=false)
 end
 
+
+@Markdown.doc """
+    blow_up(X::AbsSpec, I::Ideal)
+
+Return the blow-up morphism of blowing up `X` at `I` in `OO(X)`.
```
simonbrandhorst:
```suggestion
Return the blow-up morphism of blowing up ``X`` at ``I`` in ``OO(X)``.
```

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -581,6 +576,30 @@ function _compute_projective_glueing(gd::CoveredProjectiveGlueingData)
...
+  C = Covering([X])
+  XX = CoveredScheme(C)
+  Isheaf = ideal_sheaf(XX,X,gens(I))
+  return blow_up(Isheaf)
+end
+
+@Markdown.doc """
+    blow_up(I::IdealSheaf)
+
+Return the blow-up morphism of blowing up of the underlying scheme of I  at I.
```
simonbrandhorst:
```suggestion
Return the blow-up morphism of blowing up of the underlying scheme of ``I``  at ``I``.
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -698,3 +700,22 @@ end
...
+  X = scheme(I)
+
+  # If there is a simplified covering, use it!
+  covering = (has_attribute(X, :simplified_covering) ? simplified_covering(X) : default_covering(X))
+  n = npatches(covering)
+  i=1
+  println("Ideal Sheaf on Covered Scheme with ",n," Charts:\n")
+
+
+  ID = IdDict{AbsSpec, Ideal}()
```
simonbrandhorst:
```suggestion
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -698,3 +700,22 @@ end
...
   return IdealSheaf(X, ID, check=false)
 end
+
+function show_details(I::IdealSheaf)
+  X = scheme(I)
+
+  # If there is a simplified covering, use it!
+  covering = (has_attribute(X, :simplified_covering) ? simplified_covering(X) : default_covering(X))
+  n = npatches(covering)
+  i=1
```
simonbrandhorst:
```suggestion
```

test/AlgebraicGeometry/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -13,19 +13,19 @@
   @test codomain(p) === covered_scheme(P)
 end
 
-@testset "blowups II" begin
+@testset "testing Oscar.blow_up_chart" begin
```
simonbrandhorst:
```suggestion
@testset "Oscar.blow_up_chart" begin
```

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: thank you, optional argument added as suggested.

experimental/Schemes/CoveredProjectiveSchemes.jl
afkafkafk13: documentation added

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: ok. Then the short printing will go into Base.show and the long one into show_details. Does this suit you?

experimental/Schemes/IdealSheaves.jl
afkafkafk13: indeed.... (also at another occurrence -- I changed both of them)

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: Yes. Thank you. Though `IdealSheaf` already has a print method. Did you delete the old one?

afkafkafk13: I just moved it to a comment. (If Matthias would decide to revert this change, I would not object) 
My idea is that Cartier (and later also Weil) divisors print via base.show in a similar way as IdealSheafs and that the user gets some information on the size of the covering of the underlying scheme.

simonbrandhorst (approved):

PR #2024, opened 2023-03-09
Collect more export statements

HereAround:
Another attempt towards https://github.com/oscar-system/Oscar.jl/issues/1971.

I collect export statements in the following locations in `/src`:
* Invariant Theory,
* Polyhedral Geometry,
* Modules,
* Number theory,
* Serialization, and
* Ring.

This means, that all export statements in `/src` are now combined in `/src/Exports.jl` but the ones in:
* `/src/Groups`,
* in any (sub)module defined in `/src`,

After this PR, there will be close to 1000 exports collected in the `/src/Exports.jl`.

Some of the exported names make more, some less sense to me. To name just one instance that I found surprising: We export `f4`. This is/was exported in `/src/Rings/groebner.jl`. Maybe this is a famous operations/function/property that I just do not know about?

In any case, there is definitely a chance for a clash. Having e.g. a polynomial ring with indeterminates `f1`, `f2`, `f3`, `f4` seems not completely impossible.

The following were exported multiple times (relative to the clean-up/collection earlier this week):
* Localization,
* base_ring,
* codim,
* codomain,
* codomain_type,
* coefficients,
* cokernel,
* coordinates,
* degree,
* denominator,
* dim,
* divexact,
* domain,
* domain_type,
* exponents,
* f_vector,
* facets,
* fraction,
* gens,
* grading_group,
* groebner_basis,
* h_vector,
* hilbert_function,
* hilbert_polynomial,
* hilbert_series,
* hilbert_series_expanded,
* hilbert_series_reduced,
* ideal,
* image,
* images,
* intersect,
* inverse,
* inverted_set,
* is_bijective,
* is_complete,
* is_homogeneous,
* is_injective,
* is_isomorphism,
* is_normal,
* is_primary,
* is_prime,
* is_reduced,
* is_regular,
* is_simple,
* is_simplicial,
* is_smooth,
* is_subset,
* is_surjective,
* is_trivial,
* is_unit,
* is_very_ample,
* issubset,
* jacobi_ideal,
* jacobi_matrix,
* length,
* modulus,
* morphism_type,
* ngens,
* numerator,
* nvertices,
* parent,
* polynomial_ring,
* preimage,
* presentation,
* product,
* quo,
* quotient,
* radical,
* reduce,
* reduce_fraction,
* represents_element,
* saturated_ideal,
* simplify,
* standard_basis,
* visualize.


HereAround:
The (earlier) failures of this PR seem to be because of https://github.com/oscar-system/Oscar.jl/issues/2027#issuecomment-1463050570. I just checked the changes locally without https://github.com/oscar-system/Oscar.jl/pull/1988, and the tests completed successfully.

afkafkafk13:
Thank you for the list: Fortunately, these are all very canonical in their meaning and (as far as I checked the list) no bad clashes seem to arise, only the good kind (type dispatch...).

Concerning f4: You are perfectly right about it. It is a famous name in the Groebner bases  community, standing for Faugere's F4 algorithm. However, it does not make any sense when typeset in lower case. Maybe @ederc could come up with a slightly longer name at some point (naive suggestion: f4_algorithm) -- just to avoid problems with variable naming.

HereAround:
This is not urgent. Yet, since this PR touched many files, it might be good to merge this sooner than later to avoid conflicts. @fingolfin could you please take a look?

afkafkafk13 (approved):
This looks good to go. 
(I checked the moved exports, but might have missed a few because I prepared input by hand and then scripted the test)

PR #2025, opened 2023-03-09
Improve matrix_group()

fingolfin:
- match the `MatrixGroup` constructor functionality
- replace all uses of `MatrixGroup()` in docstrings and examples by `matrix_group`
- fix a bunch of places that directly set the `gens` of a matrix group
- make it more prominent in the documentation
- improve input validation in `MatrixGroup` constructor; e.g. we now check the inputs are square, and optionally, that they are invertible over their base rings
- those input validation checks do not use `@assert` anymore, but rather are control by a new `check` argument to `MatrixGroup` and `matrix_group`


src/Groups/matrices/MatGrp.jl
```diff
@@ -911,19 +919,23 @@ const SU = special_unitary_group
...
-   if T<:MatElem      # if T <: MatrixGroupElem, we can already assume that det(V[i]) != 0 
-      for v in V @assert det(v) !=0 "The matrix is not invertible" end
-   end
-   return MatrixGroup(nrows(V[1]),base_ring(V[1]),V)
+function matrix_group(m::Int, R::Ring, V::AbstractVector{T}; check::Bool=true) where T<:Union{MatElem,MatrixGroupElem}
+   return MatrixGroup(m, R, V)
 end
 
-matrix_group(V::T...) where T<:Union{MatElem,MatrixGroupElem} = matrix_group(collect(V))
+matrix_group(m::Int, R::Ring, V::T...; check::Bool=true) where T<:Union{MatElem,MatrixGroupElem} = matrix_group(m, R, collect(V); check)
```
joschmitt, line 919: Could we change the ordering of the ring and the degree, so `matrix_group(::Ring, ::Int, ::T...)` to align this with `matrix`, `matrix_algebra` and `matrix_space`? (I appreciate that this might be a bigger change, I just wanted to ask for opinions.)

fingolfin: On the other hand it would then clash with eg `GL(n,R)` and *that* is a book convention I wouldn't want to break with.

of course we could also support both orders...

fingolfin: (Overall I agree with you, though)

experimental/Matrix/matrix.jl
```diff
@@ -49,69 +49,69 @@ end
...
+          end
+     end
+     if K isa ZZRing
+        K = QQ
+     end
+
+     Fq, matrices_Fq, OtoFq = Oscar.good_reduction(matrices, 2)
+
+     ele = matrices_Fq[1]
+     hom = Oscar._iso_oscar_gap(ele.base_ring)
```
ThomasBreuer, line 82: Better use `Oscar.iso_oscar_gap` instead of `Oscar._iso_oscar_gap`, then the map will be cached in the ring if possible.

src/Groups/matrices/MatGrp.jl
```diff
@@ -41,13 +41,18 @@ MatrixGroup(m::Int, F::Ring) = MatrixGroup{elem_type(F), dense_matrix_type(elem_
 
 # build a MatrixGroup given a list of generators, given as array of either MatrixGroupElem or AbstractAlgebra matrices
 # WARNING: if the elements of V have type MatElem, it does not check whether the determinant is nonzero
```
ThomasBreuer: This warning is no longer true.

ThomasBreuer (approved):
Looks nice (just two minor comments).

experimental/Matrix/matrix.jl
fingolfin: Note that I merely changed the indentation of this code.

src/Groups/matrices/MatGrp.jl
fingolfin: Indeed!

src/Groups/matrices/MatGrp.jl
fingolfin: Perhaps I can discuss this with @fieker tomorrow and we'll resolve it one way or another.

(In theory we could actually offer both orders, the type system would make it possible. Not sure that's a good idea, though....)

src/Groups/matrices/MatGrp.jl
```diff
@@ -911,19 +918,23 @@ const SU = special_unitary_group
...
 end
 
-matrix_group(V::T...) where T<:Union{MatElem,MatrixGroupElem} = matrix_group(collect(V))
+matrix_group(m::Int, R::Ring, V::T...; check::Bool=true) where T<:Union{MatElem,MatrixGroupElem} = matrix_group(m, R, collect(V); check)
+
+matrix_group(V::AbstractVector{T}; check::Bool=true) where T<:Union{MatElem,MatrixGroupElem} = matrix_group(nrows(V[1]), base_ring(V[1]), V; check)
+
+matrix_group(V::T...; check::Bool=true) where T<:Union{MatElem,MatrixGroupElem} = matrix_group(collect(V); check)
 
 matrix_group(m::Int, F::Ring) = MatrixGroup{elem_type(F), dense_matrix_type(elem_type(F))}(m, F)
```
joschmitt: I introduced those `matrix_group(m::Int, F::Ring)` and `matrix_group(m::Int, F::Ring, V::AbstractVector{T})` in #2018 . They can be removed again.

ThomasBreuer:
@fingolfin the failing test is due to the reordered signatures for `matrix_group`.
Now the `matrix_group(V::AbstractVector{T})` line in `src/Groups/matgroup.md` should be replaced by a `matrix_group(m::Int, R::Ring, V::AbstractVector{T}; check::Bool=true)` line.

docs/src/Groups/matgroup.md
```diff
@@ -16,6 +16,7 @@ Pages = ["matgroup.md"]
 # Matrix groups
 
 ```@docs
+matrix_group(V::AbstractVector{T}) where T<:Union{MatElem,MatrixGroupElem}
```
ThomasBreuer:
```suggestion
matrix_group(m::Int, R::Ring, V::AbstractVector{T}; check::Bool=true) where T<:Union{MatElem,MatrixGroupElem}
```

ThomasBreuer:
The reason for the new test failure seems to be the problem from issue #2081.

PR #2028, opened 2023-03-10
Revert "let `iso_gap_oscar` choose `Nemo._GF` (#2008)"

### LINKED ISSUE #2027: CI tests broken, failures related to GAP code
I am seeing these warnings when loading OSCAR now (possibly since merging PR #1988 by @ThomasBreuer ?)
<s>
```
WARNING: Method definition julia_to_gap(Base.Set{T}) where {T} in module GAP at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/julia_to_gap.jl:163 overwritten in module Oscar at /Users/mhorn/Projekte/OSCAR/Oscar.spielwiese/src/GAP/oscar_to_gap.jl:52.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition julia_to_gap(Base.Set{T}, Base.IdDict{Any, Any}) where {T} in module GAP at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/julia_to_gap.jl:163 overwritten in module Oscar at /Users/mhorn/Projekte/OSCAR/Oscar.spielwiese/src/GAP/oscar_to_gap.jl:52.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition kwcall(Any, typeof(GAP.julia_to_gap), Base.Set{T}) where {T} in module GAP at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/julia_to_gap.jl:163 overwritten in module Oscar at /Users/mhorn/Projekte/OSCAR/Oscar.spielwiese/src/GAP/oscar_to_gap.jl:52.
  ** incremental compilation may be fatally broken for this module **

WARNING: Method definition kwcall(Any, typeof(GAP.julia_to_gap), Base.Set{T}, Base.IdDict{Any, Any}) where {T} in module GAP at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/julia_to_gap.jl:163 overwritten in module Oscar at /Users/mhorn/Projekte/OSCAR/Oscar.spielwiese/src/GAP/oscar_to_gap.jl:52.
  ** incremental compilation may be fatally broken for this module **
```
</s>

UPDATE: sorry, those warnings came from my Oscar looking at a dev version of GAP.jl. But the CI failures are real, see e.g. https://github.com/oscar-system/Oscar.jl/actions/runs/4379175069

fingolfin:
This reverts commit e2134dc4a22bd6c987fdef58bdd2a1e9a17a8445 from PR #2008.

I've chosen to revert this instead of reverting PR #1988 because...
- it is much smaller, thus removing it has less impact, and adding it back again later should be easier
- it was an experiment to begin with
- I want to verify that it is really the combination of these two PRs that causes the breakage, and not something else

Assuming this works and is merged, I'll open a new PR which adds these changes back so that we can take our time to analyze the problem.

Hopefully fixes #2027

### DIFF
--- a/src/GAP/iso_gap_oscar.jl
+++ b/src/GAP/iso_gap_oscar.jl
@@ -76,7 +76,16 @@ function _iso_gap_oscar_residue_ring(RG::GAP.GapObj)
 end
 
 function _iso_gap_oscar_field_finite(FG::GAP.GapObj)
-   FO = Nemo._GF(characteristic(FG), GAPWrap.DegreeOverPrimeField(FG))
+   p = characteristic(FG)  # of type `ZZRingElem`
+   d = GAPWrap.DegreeOverPrimeField(FG)
+   if d == 1
+     if p < ZZRingElem(2)^64
+       p = UInt64(p)
+     end
+     FO = GF(p)
+   else
+     FO = GF(p, d)
+   end
 
    finv, f = _iso_oscar_gap_field_finite_functions(FO, FG)
 
--- a/src/GAP/iso_oscar_gap.jl
+++ b/src/GAP/iso_oscar_gap.jl
@@ -74,16 +74,14 @@ function _iso_oscar_gap_field_finite_functions(FO::Union{FqPolyRepField, FqField
    p = characteristic(FO)
    d = degree(FO)
 
-   if degree(FO) != absolute_degree(FO) ||
-      ! GAPWrap.IsPrimeField(GAPWrap.LeftActingDomain(FG))
-     # The Oscar field or the GAP field is not an extension of the prime field.
+   # Compute the canonical basis of `FG`.
+   if ! GAPWrap.IsPrimeField(GAPWrap.LeftActingDomain(FG))
+     # The GAP field is not an extension of the prime field.
      # What is a reasonable way to compute (on the GAP side) a polynomial
      # w.r.t. the prime field, and to decompose field elements w.r.t.
      # the corresponding basis?
      error("extensions of extension fields are not supported")
    end
-
-   # Compute the canonical basis of `FG`.
    basis_FG = GAPWrap.Basis(FG)
    # Test that we do not run into the problem from
    # https://github.com/gap-system/gap/issues/4694.
--- a/src/Groups/libraries/atlasgroups.jl
+++ b/src/Groups/libraries/atlasgroups.jl
@@ -22,7 +22,7 @@ julia> atlas_group("A5")  # alternating group A5
 Group([ (1,2)(3,4), (1,3,5) ])
 
 julia> atlas_group(MatrixGroup, "A5")
-Matrix group of degree 4 over Finite field of degree 1 over F_2
+Matrix group of degree 4 over Galois field with characteristic 2
 
 julia> atlas_group("M11")  # Mathieu group M11
 Group([ (2,10)(4,11)(5,7)(8,9), (1,4,3,8)(2,5,6,9) ])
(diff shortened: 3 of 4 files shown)


thofma (approved):

ThomasBreuer (approved):
Yes, good idea.
(In fact I cannot reproduce the problems from #2027 locally yet.)

fingolfin:
Unfortunately this does *not* help. So I've now opened #2029 to try to revert #1988 after all.

But I wonder now if this is even the problem: we also updated Nemo and AA yesterday. Could those updates somehow have caused the problems? We could check the CI logs to see when exactly these failures occurred for the first time.

@ThomasBreuer why are there no `julia_to_gap` methods for finite fields? I assume that means our code doesn't need them. So why are they are now being invoked (which fails as no such methods exist), any idea?

fingolfin:
The other PR also doesn't help. So perhaps it is the AA and/or Nemo update? Or something else? I guess I first need to understand what's going on better *sigh*

fingolfin:
I believe the root cause is this Nemo PR https://github.com/Nemocas/Nemo.jl/pull/1406 by @thofma. We are doing this to map from an OSCAR FFE to a GAP FFE:
```julia
     f = function(x)
       v = [GAP.Obj(coeff(preimage(emb, x), i)) for i in 0:(d - 1)]
       return sum([v[i]*basis_FG[i] for i in 1:d])
     end
```
And `coeff(...)` use to return an `ZZRingElem`; now it returns a finite field element.

fingolfin:
OK, with this PR, at least I get *different* errors than with master.

fingolfin:
I cannot reproduce the error in this PR, which is this:
```
(p, d) = (2, 1): Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/GAP/iso_oscar_gap.jl:64
  Test threw exception
  Expression: f(a * b) == f(a) * f(b)
  UndefKeywordError: keyword argument recursive not assigned
```

However, if I do `using Oscar, Test ; include("test/GAP/iso_oscar_gap.jl")` then no such error.

Perhaps this is a matter of caching (?) or something else in the preceding tests that affects global state and thus causes the failure?

fingolfin:
Trying to deal with this. To get the coeffs as integers instead of FFEs, I could replace `coeff` by `Nemo._coeff` but that is undocumented and I am not sure it does the right thing in all cases. I could also use `lift(ZZ, coeff(x, i))`, but (a) this is more expensive (not that I care at this point), and (b)  this code then breaks if `coeff(x,i)` returns a `XX` (say it is from another ring). Perhaps we should add a method `lift(::ZZRing, x::ZZRingElem) = x` ?

ThomasBreuer:
@fingolfin Concerning your questions

> why are there no `julia_to_gap` methods for finite fields?
> I assume that means our code doesn't need them.

`iso_oscar_gap` (or `iso_gap_oscar`) is provided for finite fields, the returned mapping then deals with the conversion of the elements of the finite fields, w.r.t. the fields stored as `domain` and `codomain` of the mapping.
One could think of `julia_to_gap` (or rather `GAP.Obj`) methods for elements of finite *prime* fields, based on the correspondence with integers, but there was indeed no need for that.

> So why are they are now being invoked (which fails as no such methods exist), any idea?

As you wrote above, `Nemo.coeff(x::FqFieldElem, n::Int)` used to return an integer if `parent(x)` is an extension of the prime field, but now it returns an element of `base_field(parent(x))`.

@thofma We can get back to the old behaviour by calling `Nemo._coeff(x, n)`. Do you recommend a different solution?

PR #2032, opened 2023-03-10
Resolve duplicate  _is_homogeneous(f::MPolyElem)

fingolfin:


afkafkafk13 (approved):
It  definitely makes sense to compute the (most common) standard graded case separately and to do it in groebner.jl and not hidden within the tropical stuff.

PR #2037, opened 2023-03-10
Use macro `@req` in more places

fingolfin:


HereAround (approved):
Thank you @fingolfin . This looks good to me.

PR #2039, opened 2023-03-12
Revise `schur_polynomial`

fingolfin:
- remove `schur_polynomial(lambda::Partition{T}, x::Vector{fmpz_mpoly})` as it
  was broken (the vector `x` was ignored in several cases)
- remove optional keyword argument `ring` (specify the ring explicitly)
- change default for `n` from `sum(lambda)` to `length(lambda)`
- if `n` exceed `nvars(R)`, throw an exception instead of silently truncating `n`
- various other minor tweaks

### DIFF
--- a/experimental/JuLie/schur_polynomials.jl
+++ b/experimental/JuLie/schur_polynomials.jl
@@ -10,34 +10,30 @@
 export schur_polynomial
 
 @doc Markdown.doc"""
-    schur_polynomial(lambda::Partition{T}, n::Int=sum(lambda); ring::FmpzMPolyRing) where T<:Integer
-    schur_polynomial(R::FmpzMPolyRing, lambda::Partition{T}, n::Int=sum(lambda)) where T<:Integer
-    schur_polynomial(lambda::Partition{T}, x::Vector{fmpz_mpoly}) where T<:Integer
+    schur_polynomial(lambda::Partition{T}, n::Int=length(lambda)) where T<:Integer
+    schur_polynomial(R::FmpzMPolyRing, lambda::Partition{T}, n::Int=length(lambda)) where T<:Integer
 
-Returns the Schur polynomial ``s_λ(x_1,x_2,...,x_n)`` in n variables, as a
-Multivariate Polynomial.
+Returns the Schur polynomial ``s_λ(x_1,x_2,...,x_n)`` in `n` variables.
 
-If neither `R` nor `x` are given, the Schur polynomial will be over
-`PolynomialRing(ZZ,["x1","x2",...,"xn"])`.
+If `R` is not given, the Schur polynomial will be over `polynomial_ring(ZZ,n)`.
 
 # Examples
 ```jldoctest
-julia> R,x = PolynomialRing(ZZ, ["a","b","c"])
-(Multivariate Polynomial Ring in a, b, c over Integer Ring, ZZMPolyRingElem[a, b, c])
-
-julia> schur_polynomial(Partition([2,1]),[x[1],x[2]])
-a^2*b + a*b^2
+julia> R,x = polynomial_ring(ZZ, ["a","b","c"]);
 
 julia> schur_polynomial(R, Partition([2,1]))
+a^2*b + a*b^2
+
+julia> schur_polynomial(R, Partition([2,1]), 3)
 a^2*b + a^2*c + a*b^2 + 2*a*b*c + a*c^2 + b^2*c + b*c^2
 
 julia> schur_polynomial(Partition([2]))
-x1^2 + x1*x2 + x2^2
+x1^2
 ```
 
 # Algorithm
 We use two different Algorithms, depending on the size of the input.
-The Combinatorial Algorithm is used for Partitions of small Integers, or if
+The Combinatorial Algorithm is used for Partitions of small integers, or if
 ``n ≥ 10``. In the other cases we use Cauchy's bialternant formula.
 
 **Combinatorial Algorithm**
@@ -58,13 +54,7 @@ x_1^{λ_n} & x_2^{λ_n} & … & x_n^{λ_n}
 \end{vmatrix}
 ```
 """
-function schur_polynomial(lambda::Partition{T}, n::Int=sum(lambda);
-    ring::AbstractAlgebra.Ring=(n<=0 ? ZZ : begin # the ring for the return value
-                                  x = [string("x",string(i)) for i=1:n]
-                                  PolynomialRing(ZZ, x)[1]
-                                end
-                               )
-  ) where T<:Integer
+function schur_polynomial(lambda::Partition{T}, n::Int=length(lambda)) where T<:Integer
   @req n >= 0 "n >= 0 required"
   if n==0 || n < length(lambda)
     if isempty(lambda)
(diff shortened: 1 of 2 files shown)


test/Experimental/JuLie/schur_polynomials.jl
```diff
@@ -13,12 +12,12 @@ import Oscar.JuLie: schur_polynomial_cbf
...
 	@test schur_polynomial(S, Partition([])) == 1
 
-	# schur_polynomial(λ, R, n)
+	# schur_polynomial(R, λ, n)
 	@test schur_polynomial(S, Partition([1]), 3) == x[1] + x[2] + x[3]
 	@test schur_polynomial(S, Partition([1,1]), 3) == x[1]*x[2] + x[1]*x[3] + x[2]*x[3]
 	@test schur_polynomial(S, Partition([2]), 3) == x[1]^2 + x[2]^2 + x[1]*x[2] + x[1]*x[3] + x[2]*x[3] + x[3]^2
 	@test schur_polynomial(S, Partition([1,1,1]), 3) == x[1]*x[2]*x[3]
-	@test schur_polynomial(S, Partition([1,1,1]), 4) == x[1]*x[2]*x[3]
+	#@test schur_polynomial(S, Partition([1,1,1]), 4) == x[1]*x[2]*x[3]
```
HereAround: Was this test disabled on purpose? If so, why?

fingolfin: Ahhh, good point, thanks for reminding me, I wanted to ask @ulthiel about this:
This is a partition of 3, with three components, and we gave a polynomial ring with three variables. Yet we asked for a Schur polynomial in four variables... which the code silently truncated to 3 variables. Which I find problematic, I'd rather have an error. But perhaps there is a rationale for this I am simply not aware. In that case we can allow this again, and extend the docstrings to describe explain this behavior.

experimental/JuLie/schur_polynomials.jl
```diff
@@ -10,24 +10,18 @@
 export schur_polynomial
 
 @doc Markdown.doc"""
-    schur_polynomial(lambda::Partition{T}, n::Int=sum(lambda); ring::FmpzMPolyRing) where T<:Integer
-    schur_polynomial(R::FmpzMPolyRing, lambda::Partition{T}, n::Int=sum(lambda)) where T<:Integer
-    schur_polynomial(lambda::Partition{T}, x::Vector{fmpz_mpoly}) where T<:Integer
```
fingolfin, line 15: @ulthiel I removed `schur_polynomial(lambda, x)` because it was buggy (the content of `x` was ignored in many cases, it only worked right when `x == gens(R)`.

If supporting this for an arbitrary vector of polynomials is important to you, we can of course also keep it but then should fix it. I can look into doing that but I did not want to expend the effort if nobody then intends to use it ;-).

experimental/JuLie/schur_polynomials.jl
```diff
@@ -58,32 +52,21 @@ x_1^{λ_n} & x_2^{λ_n} & … & x_n^{λ_n}
...
 ```
 """
-function schur_polynomial(lambda::Partition{T}, n::Int=sum(lambda);
-    ring::AbstractAlgebra.Ring=(n<=0 ? ZZ : begin # the ring for the return value
-                                  x = [string("x",string(i)) for i=1:n]
-                                  PolynomialRing(ZZ, x)[1]
-                                end
-                               )
-  ) where T<:Integer
+function schur_polynomial(lambda::Partition{T}, n::Int=sum(lambda)) where T<:Integer
```
fingolfin: @ulthiel actually, why is the default value for `n` here `sum(lambda)`? I would have expected `length(lambda)`... 

In the definition of Schur polynomial I am familiar with, then given a partition of length $n$ I'd expect the Schur polynomial to be in $n$ indeterminates. So it is not even clear to me what happens when `n > length(lambda)`. Perhaps there is a more general definition which allows this?

What am I missing?

experimental/JuLie/schur_polynomials.jl
ulthiel: I have to remember what was being done here. I believe that passing the vector should return the evaluated Schur polynomial in this vector. This should be faster than first computing the Schur polynomial (in the indeterminates) and then evaluating. So I believe there's good reason for this function in principle.

test/Experimental/JuLie/schur_polynomials.jl
ulthiel: Okay, so, this starts from the combinatorial formula for Schur polynomials: $s_\lambda = \sum_T x^T$ where the sum is over all semistandard Young tableaux of shape $\lambda$ and $x^T = x_1^{t_1} \cdots x_n^{t_n}$, where $t_i$ is the multiplicity of $i$ in $T$. So, the number of variables you need (at minimum) equals the maximum value of the entries in T. But semistandard Tableaux are always relative to an alphabet, usually 1 up to some maximal number. But this maximal number can be chosen to be anything from 1 to infinity (see also `semistandard_tableaux` and also [Sage](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/tableau.html#sage.combinat.tableau.SemistandardTableaux)). So, the n can be any number >= 1 from the combinatorial viewpoint. If n is not specified then it should default the _length_ of $\lambda$; not the sum as in the code, which would be the default for listing semistandard tableaux, that's where this typo may come from.

Example: 
```
julia> semistandard_tableaux(Partition([3,2]), 2)
2-element Vector{Tableau{Int64}}:
 [[1, 1, 1], [2, 2]]
 [[1, 1, 2], [2, 2]]

julia> schur_polynomial(Partition([3,2]),2)
x1^3*x2^2 + x1^2*x2^3
```

```
julia> semistandard_tableaux(Partition([2,1]), 3)
8-element Vector{Tableau{Int64}}:
 [[1, 1], [2]]
 [[1, 1], [3]]
 [[1, 2], [2]]
 [[1, 2], [3]]
 [[1, 3], [2]]
 [[1, 3], [3]]
 [[2, 2], [3]]
 [[2, 3], [3]]

julia> schur_polynomial(Partition([2,1]),3)
x1^2*x2 + x1^2*x3 + x1*x2^2 + 2*x1*x2*x3 + x1*x3^2 + x2^2*x3 + x2*x3^2
```

experimental/JuLie/schur_polynomials.jl
ulthiel: See my comment above. True, the default should be length(lambda) and this is wrong several times.

test/Experimental/JuLie/schur_polynomials.jl
ulthiel: I am not entirely sure though whether the implementation of the bialternant formula works in this combinatorial generality though I have to check this.

ulthiel: Yes, I think also in the bialternant formula you can take any number of variables, so this should be okay. But you are correct about `schur_polynomial(S, Partition([1,1,1]), 4)` when S is a polynomial ring in 3 variables. This should not happen. It needs to have at least n variables.

experimental/JuLie/schur_polynomials.jl
ulthiel: Yes, `schur_polynomial(lambda, x)` as it is in the code is useless and can be removed. In principle there would be a purpose (when x is a vector of elements of a ring) but this needs to be implemented separately I think. Now there's still `schur_polynomial_cbf(lambda, x)` where the x is technically not necessary as well and should better have the same signature as the other functions (needs some rewriting of the code).

ulthiel: Okay, an idea: what comes after the comment "#version of the function semistandard_tableaux(shape::Vector{T}, max_val=sum(shape))" in `schur_polynomial` is actually an interator over semistandard tableaux. This should (later) go into an actual separate iterator. Then this function looks cleaner. Moreover, one can reuse it do implement `schur_polynomial` with argument a vector x of ring elements which returns the Schur polynomial evaluated at this vector but which should be faster than first building a large polynomial and then evaluating. One application I have in mind is the q-analog where one evaluates at 1,q,q^2,... (but there's a closed formula for the result actually).

ulthiel: But maybe first merge with your fixes and then all this can be done later.

fingolfin:
@ulthiel @HereAround I've update the PR now, hopefully addressing all comments (well, except the one of adding back support for a list `x` of variables; that should come in a future PR, probably not by me)

ulthiel (approved):
Looks good to me, thanks for the revision!

ulthiel:
@fingolfin Your last push (faaf58d) seems to involve nothing related to Schur polynomials. Is this correct?

fingolfin:
IIRC the last push was just a rebase to resolve a minor conflict somewhere.

PR #2040, opened 2023-03-13
Update experimental SymInt

StevellM:
Several changes:
- Minor fixes of some function (which were not tested, or for which the result could be wrong) - namely `matrix_representation`, `is_projective` and `is_similar`;
- Adapt to the new naming conventions (shorter type names, provide constructor functions, ...);
- Remove `determinant` and `center_of_character` to use the recently added `det` and `center` respectively;
- Add `export` in front of each exported function;
- Reduce the number of tests and their complexity (as you requested @fingolfin );
- Add more documentation for the types/structures.


fingolfin:
Thanks! Tests are down to 11 seconds, plus 30 seconds compilation -- not much to be done about the latter, I guess :-/. Still, a lot better than what it was before (e.g. I just saw a `master` CI run where it was 150+50 !)

fingolfin (approved):

PR #2041, opened 2023-03-13
Changes related to recent Hecke release

StevellM:
Some functions are deprecated after the last Hecke release, namely direct sums of lattice-related objects, so some codes had to be adapted.

I have also fixed of a function which was flawed because of the use of the wrong method inside.

@simonbrandhorst


fingolfin:
Which version of Hecke introduced these methods? 0.18? Or was it a later one (in that case we also need to adjust `Project.toml`)

StevellM:
The very last one, so 0.18.2.

StevellM:
But everything seems to run without conflict right?

fingolfin:
My point is: if there are changes in here which require hecke 0.18.2, then this must also change `Project.toml` accordingly. Of course by default if you install Oscar from scratch, it will also install the latest Hecke from scratch. But there are all kinds of scenarios where an older Hecke might be accidentally used with a new Oscar. The only thing protecting us against that are accurate version restrictions in `Project.toml`

fingolfin (commented):
Apparently requires a `Project.toml` update

fingolfin (approved):

PR #2045, opened 2023-03-13
Final stab towards organizing export statements

### LINKED ISSUE #1971: New rule: `export` only one identifier at a time

HereAround:
Hopefully the final stab at fixing #1971.

This PR achieves the following:
1. Moves all exports in `src/Groups/` to `src/Exports.jl`.
2. Moves a few left over exports from other places within `src` to `src/Exports.jl`.
3. Removes outcommented export statements in `src`.
4. Orginizes the exports in all submodules of `src` according to the rule proposed in https://github.com/oscar-system/Oscar.jl/issues/1971.
5. Apply the export rule to all of `experimental`. Crucially, the experimental exports were **not** moved to `src/Exports.jl`.
6. Explain the export rule in the style guide.

Therefore, after this PR, the export statements within `src` should all  be collected in `src/Exports.jl`. The only exception are (sub)modules. Also, all the code in the experimental folder should satisfy the export rule.

During this clean-up, I noticed that the following commands were exported multiple times:
* `automorphism_group`,
* `sub`,
* `ambient_ring`,
* `codomain`,
* `cokernel`,
* `complement`,
* `degree`,
* `direct_product`,
* `domain`,
* `embedding`,
* `gens`,
* `image`,
* `inverted_set`,
* `is_bijective`,
* `is_conjugate`,
* `is_injective`,
* `is_isomorphic`,
* `is_left`,
* `is_regular`,
* `is_right`,
* `is_simple`,
* `is_surjective`,
* `kernel`,
* `maximal_subgroup_reps`,
* `ngens`,
* `order`,
* `orthogonal_group`,
* `quo`,
* `representative`,
* `representative_action`,
* `right_coset`,
* `right_cosets`,
* `right_transversal`,
* `slope`,
* `unitary_group`.

Resolves #1971.


src/Exports.jl
```diff
@@ -398,26 +523,92 @@ export groebner_basis_hilbert_driven
...
+export has_derived_series
+export has_derived_subgroup
 export has_edge
+export has_exponent
+export has_fitting_subgroup
+export has_frattini_subgroup
+export has_gens
+export has_hall_system
+export has_is_abelian
+export has_is_almostsimple
```
fingolfin: I would strongly prefer to keep the has_ and set_ function together on the same line with the corresponding getters. There is no drawback for this in terms of finding where an export is.

fingolfin (changes requested):

src/Exports.jl
HereAround: I have rebased and added this change in https://github.com/oscar-system/Oscar.jl/pull/2045/commits/9448c2826223fd76f8783fb7b0bab98e4d8a1b33.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -187,6 +187,9 @@ end
   to understand; it is also more difficult to see all the code at once. Consider
   splitting the function into multiple ones, if it is sensibly possible.
 
+- Every export statement must export exactly one identifier. If multiple
```
fingolfin: I guess that's not accurate anymore with the `has_`/`set_` revision. It is also IMHO not necessary; the crucial bit is that every `export` statement should be on a single line, so that one can conveniently grep for exports.

How about this:
```suggestion
- Every export statement must be confined to a single line; the intention is to
  make it easy to use tools like `git grep` to find exports. In general it is
  recommended export exactly one identifier per export statement. Exceptions may
  be made for certain tightly related identifiers, e.g. `is_finite`, `set_is_finite`
  and `has_is_finite` could be put on a single line. In general if multiple
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -187,6 +187,9 @@ end
...
+  export statements appear in sequence, they must be sorted alphabetically.
```
fingolfin, line 195: I think it's a good idea to explain the *why* behind some of our rules. If we can't explain a rule, then it's probably not a good rule. Knowing the motivation for a rule can also make it easier to accept it psychologically. Plus it helps to decide when it might be sensible to break a rule, or to revise it. 
```suggestion
  export statements appear in sequence, they must be sorted alphabetically.
  The purpose of this is to make it easy to decide where to put additional
  exports, and also to detect duplicate exports (which are not harmful by
  themselves, but if we decide to no longer export an identifier, having
  multiple exports can lead to mistakes such as removing only one of them.
```

experimental/GModule/Cohomology.jl
```diff
@@ -2055,8 +2055,17 @@ function Oscar.direct_product(C::GModule...; task::Symbol = :none)
   end
 end
 
-export GModule, gmodule, word, fp_group, confluent_fp_group, induce,
-       action, cohomology_group, extension, is_coboundary, pc_group
+export GModule
```
fingolfin, line 2058: So it seems we are sorting uppercase before lowercase? Then the styleguide should perhaps mention that, too?

src/Aliases.jl
```diff
@@ -76,7 +76,6 @@
 
 # Allow backwards compatibility after removal of Oscar.Graphs module.
 Graphs = Oscar
```
fingolfin: This is of course not your mistake at all, but since i am seeing this:
```suggestion
const Graphs = Oscar
```
CC @lkastner

src/Exports.jl
```diff
@@ -398,26 +516,91 @@ export groebner_basis_hilbert_driven
 export groebner_basis_with_transformation_matrix
 export groebner_polyhedron
 export group
+export gset
 export h_star_polynomial
 export h_vector
 export halfspace_matrix_pair
+export hall_subgroup
+export hall_subgroup_reps
+export has_center, set_center, center
```
fingolfin: Apologies for requesting yet another change, but, I'd really prefer if this was as it was before, i.e.:
```suggestion
export center, has_center, set_center
```

Because the relevant bit is `center`, and `set_` / `has_` just qualify it.

I do not wish to cause you excessive work. It should be easy to change this with a simple regex search&replace, so I am happy to make the change myself; just let me know (I prefer to not make changes to your PR without coordinating them with you!)

fingolfin (commented):
Thank you for working on this!

src/Exports.jl
HereAround: Please feel free to change this (you are most likely quicker than I) and also incorporate the above suggestions - they all look good to me.

My rational to put the `has_` at the beginning was to ensure that all triples (`has_`, `set_`, something) appear grouped together in `Exports.jl`. Maybe this is useful if one wants to inspect the list in the future?

E.g. there are pairs (`has_`, `set_`) with the getter missing (e.g. `export has_exponent, set_exponent`), which one might want to add in the future. But then, of course one can also find them if these pairs are "somewhere" in the `Exports.jl` file...

Your call. Please change it to a form that you like.

fingolfin (approved):

src/Exports.jl
fingolfin: done. actually we do have `exponent`, we just forgot to export it, as it's a method for `Base.exponent` and thus always available in the REPL. But when importing `Oscar` into a module, this could lead to confusing problems. So I've added that now. I think that also shows that it's valuable to group these together, so that we notice such problems.

PR #2048, opened 2023-03-14
Jaa/exterior algebra

JohnAAbbott:
`exterior_algebra` now uses the Singular implementation to do the actual work; I had to revise `PBWAlgebraQuo.jl`
to make everything work properly.
I have left the implmenetation of `exterior_algebra_PBWAlgQuo` in the file, though I am unsure whether it could ever be useful.


experimental/ExteriorAlgebra/ExteriorAlgebra.jl
```diff
@@ -29,12 +34,15 @@ for its second argument, and this function becomes redundant.
...
+##??? return string.(xs);  # shorter but equivalent ???
    NumVars = length(xs);
 # This commented out code appears to be a bit slower.
 ##   L = [""  for _ in 1:NumVars];
 ##   for i = 1:NumVars   L[i] = string(xs[i]);  end;
    L = String[];
-   for var in xs   push!(L, string(var));  end;
+    for var in xs
+        push!(L, string(var));
+    end;
```
fingolfin: I understand that you are used to using semicolons in your code. But this is highly non-idiomatic for Julia, and I think it would be best to start adapting to the realities of this language and break with this habit here.

experimental/ExteriorAlgebra/ExteriorAlgebra.jl
```diff
@@ -68,56 +100,74 @@ julia> e2*e1
...
 julia> y*x
 -x*y
 ```
 """
-function exterior_algebra(K::Field, NumVars::Int)
+function exterior_algebra(K::Field, numVars::Int)
   fn_name = "exterior_algebra";
-  if (NumVars < 1)  throw(ArgumentError("$fn_name ctor: (arg2) NumVars must be strictly positive"));  end;
-  return exterior_algebra(K, (1:NumVars) .|> (k -> "e$k"));
+  if (numVars < 1)  throw(ArgumentError("$fn_name ctor: (arg2) must be strictly positive, but numVars=$numVars"));  end;
```
fingolfin: Idiomatic Julia would be:
```suggestion
  numVars >= 1 || throw(ArgumentError("numVars must be strictly positive, but is $numVars"))
```
In Oscar you can also use the `@req` macro:
```suggestion
  @req numVars >= 1 "numVars must be strictly positive, but is $numVars"
```
Using `fn_name` is not necessary because the stacktrace already pinpoints where the error comes from:
```
julia> exterior_algebra(QQ, 0)
ERROR: ArgumentError: exterior_algebra ctor: (arg2) NumVars must be strictly positive
Stacktrace:
 [1] exterior_algebra(K::QQField, NumVars::Int64)
   @ Oscar ~/Projekte/OSCAR/Oscar.spielwiese/experimental/ExteriorAlgebra/ExteriorAlgebra.jl:82
 [2] top-level scope
   @ REPL[15]:1
```

And if the docstring for `exterior_algebra` already mentions `NumVars` then this should be sufficient to inform the user which argument the error is about, no need to say `arg2`

src/Rings/PBWAlgebraQuo.jl
```diff
@@ -1,5 +1,57 @@
...
+#   EXERCISE: sit down and draw a box-and-arrow diagram of values and references!!
+
+
+
 mutable struct PBWAlgQuo{T, S} <: NCRing
-  I::PBWAlgIdeal{0, T, S}
+    I::PBWAlgIdeal{0, T, S}
+    sring::Singular.PluralRing{S}  # For ExtAlg this is the Singular impl; o/w same as I.basering.sring
+
+##    # THIS DOES NOT WORK: no idea why!  [try Julia manual "parametric constructors"]
```
fingolfin: When you say "this does not work", what do you mean exactly? I.e., what does not work? Do you get a syntax error, or a runtime error, or...  (and which exactly)?

src/Rings/PBWAlgebraQuo.jl
```diff
@@ -1,5 +1,57 @@
...
+end
+
+# For backward compatibility: ctor with 1 arg:
+#    uses "default" arith impl -- namely that from basering!
+function PBWAlgQuo(I::PBWAlgIdeal{0, T, S})  where {T, S}
+    return PBWAlgQuo{T, S}(I, I.basering.sring)
+end
+
+
+function have_special_impl(Q::PBWAlgQuo) :: Bool
```
fingolfin:
```suggestion
function have_special_impl(Q::PBWAlgQuo)
```
The type indicator is redundant here, Julia can infer that `!=` will produce a `Bool`

src/Rings/PBWAlgebraQuo.jl
```diff
@@ -57,46 +109,47 @@ end
 ####
 
 function ngens(Q::PBWAlgQuo)
-  return ngens(base_ring(Q))
+  return ngens(base_ring(Q))  # EQUIV  ngens(Q.sring)  ???
 end
 
 function gens(Q::PBWAlgQuo)
-  return elem_type(Q)[PBWAlgQuoElem(Q, x) for x in gens(base_ring(Q))]
+  return elem_type(Q)[PBWAlgQuoElem(Q, PBWAlgElem(Q.I.basering,x)) for x in gens(Q.sring)]
```
fingolfin, line 110: Meta note: if at all possible, we should move away from directly accessing struct members like `Q.sring` or `Q.I.basering`; instead accessors should be used (and added if necessary). That way, it becomes easier to change internals of structs in the future.

src/Rings/PBWAlgebraQuo.jl
```diff
@@ -57,46 +109,47 @@ end
...
 end
 
 function one(Q::PBWAlgQuo)
-  return PBWAlgQuoElem(Q, one(base_ring(Q)))
+  return PBWAlgQuoElem(Q, PBWAlgElem(Q.I.basering, one(Q.sring)))
 end
 
-function simplify!(a::PBWAlgQuoElem)
+function simplify(a::PBWAlgQuoElem)
+  if (have_special_impl(parent(a)))  return a;  end; # short-cut for exterior algebras
```
fingolfin: Idiomatic:
```suggestion
  have_special_impl(parent(a)) && return a   # short-cut for exterior algebras
```

fingolfin: If this is a shortcut for exterior algebras, then does that mean `have_special_impl` really checks for exterior algebras? Then perhaps that would suggest that `is_exterior_algebra_impl` or so might be a better name for that function?

src/Rings/PBWAlgebraQuo.jl
```diff
@@ -215,29 +278,38 @@ function (Q::PBWAlgQuo{T, S})(c::T) where {T, S}
...
 
 function (Q::PBWAlgQuo)(c::IntegerUnion)
-  return PBWAlgQuoElem(Q, base_ring(Q)(c))
+  return PBWAlgQuoElem(Q, PBWAlgElem(base_ring(Q), Q.sring(c)))
 end
 
 function (Q::PBWAlgQuo)(a::PBWAlgQuoElem)
-  parent(a) == Q || error("coercion impossible")
-  return a
+    if (parent(a) != Q)
```
fingolfin:
```suggestion
    if parent(a) != Q
```
and similarly elsewhere. Well, or just use `@req` here

fingolfin (commented):
Thanks John! I left some quick remarks, all about superficial things only :-)

src/Rings/PBWAlgebraQuo.jl
JohnAAbbott: Currently the only special case is exterior algebras.  I don't know if further special cases may come along.
The "special" implementation just means that arithmetic is delegated directly to Singular, rather than indirectly via `PBWAlgebra`.  The `PBWAlgebra` multiplication does not automatically reduce the result (because this might be unnecessary or quite costly in some circumstances).  I prefer to leave the name as is in case more special cases are needed.

experimental/ExteriorAlgebra/ExteriorAlgebra.jl
JohnAAbbott: I can remove them.  I found them a useful visual clue.

src/Rings/PBWAlgebraQuo.jl
JohnAAbbott: Mmm.  If I recall correctly, I failed to find a combination of constructor declaration, and constructor call which worked -- Julia would report that no matching method existed, and would then printout what looked to me like two identical function signatures, claiming that they were different.  The only loss is that an argument sanity check is not performed.  I can just delete the commented-out code.

src/Rings/PBWAlgebraQuo.jl
JohnAAbbott: OK, fair point.  This'll take some work.

fingolfin: sure! I should clarify that I did not mean to suggest you should do it in this PR

src/Rings/PBWAlgebraQuo.jl
fingolfin: OK, thanks for the explanation. Maybe you could add a comment to the function with just that explanation? It might help the next person reading that code :-)

docs/src/NoncommutativeAlgebra/PBWAlgebras/quotients.md
```diff
@@ -39,15 +39,17 @@ quo(A::PBWAlgRing, I::PBWAlgIdeal)
...
 
 $\langle e_1^2,\dots, e_n^2\rangle.$
 
 ```@docs
-    exterior_algebra(K::Ring, xs::Union{AbstractVector{<:AbstractString}, 
-                                    AbstractVector{Symbol}, AbstractVector{Char}})
+    exterior_algebra(K::Field, numVars::Int)
+    exterior_algebra(K::Field, listOfVarNames::Union{AbstractVector{<:AbstractString},
+                                                     AbstractVector{Symbol},
+                                                     AbstractVector{Char}})
```
fingolfin: I am now trying to figure out what the problem with the docs are.

But as another meta note (i.e.: not a change request): there are now a bunch of places what want to process "specifications of variable names": this code here, but also polynomial rings, various field constructors (those usually only take a single variable, though), and I recently saw a bunch more in various pull requests (I think one was/is in the one on blow ups). Many of those only support strings, or only support symbols; they all support different subsets and extensions.

Perhaps there is a chance here to extract some logic in a way so that it can be easily reused across multiple different implementations? I'll think a bit more about it (and I think you had a relevant GitHub Discussion were I should probably post this. I'll try to find it later)

fingolfin: Pushed a fix. Explanation can be found in the commit message.

experimental/ExteriorAlgebra/ExteriorAlgebra.jl
```diff
@@ -68,102 +78,139 @@ julia> e2*e1
...
-    SINGULAR_PTR = Singular.libSingular.exteriorAlgebra(Singular.libSingular.rCopy(P.ptr));
-    ExtAlg = Singular.create_ring_from_singular_ring(SINGULAR_PTR);
-    return ExtAlg, gens(ExtAlg);
-end;
 
+# ---------------------------------
+# -- Method where caller specifies just number of variables
 
+function exterior_algebra(K::Field, numVars::Int)
+    if (numVars < 1)
```
fingolfin:
```suggestion
    if numVars < 1
```

experimental/ExteriorAlgebra/ExteriorAlgebra.jl
```diff
@@ -68,102 +78,139 @@ julia> e2*e1
...
-# Exterior algebra implementation as a quotient of a PBW algebra;
-# PREFER exterior_algebra over this slow implementation
+#---------------------------------
+# Method where caller specifies name of variables.
+
+function exterior_algebra(K::Field, listOfVarNames::Union{AbstractVector{<:AbstractString},
+                                                          AbstractVector{Symbol},
+                                                          AbstractVector{Char}})
+    numVars = length(listOfVarNames)
+    if (numVars == 0)
```
fingolfin:
```suggestion
    if numVars == 0
```

fingolfin (approved):

PR #2051, opened 2023-03-15
Don't throw strings or ErrorException

fingolfin:
Instead use

### DIFF
--- a/src/PolyhedralGeometry/LPFiles.jl
+++ b/src/PolyhedralGeometry/LPFiles.jl
@@ -141,7 +141,7 @@ function load_mps(file::String)
     Polymake.attach(milp, "convention", "max")
     return MixedIntegerLinearProgram(Polyhedron(poly), milp, :max)
   else
-    throw(ErrorException("load_mps: cannot find LP or MILP subobject in polymake object"))
+    error("load_mps: cannot find LP or MILP subobject in polymake object")
   end
 end
 
@@ -164,7 +164,7 @@ function load_lp(file::String)
     Polymake.attach(milp, "convention", String(convention))
     return MixedIntegerLinearProgram(Polyhedron(poly), milp, convention)
   else
-    throw(ErrorException("load_lp: cannot find LP or MILP subobject in polymake object"))
+    error("load_lp: cannot find LP or MILP subobject in polymake object")
   end
 end
 
--- a/src/PolyhedralGeometry/triangulations.jl
+++ b/src/PolyhedralGeometry/triangulations.jl
@@ -31,8 +31,8 @@ function topcom_regular_triangulations(pts::AbstractCollection[PointVector]; ful
     end
     wait(task)
     if !success(proc)
-        error = eof(err) ? "unknown error" : readchomp(err)
-        throw("Failed to run TOPCOM: $error")
+        msg = eof(err) ? "unknown error" : readchomp(err)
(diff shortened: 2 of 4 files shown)


benlorenz:
I do see that throwing strings is a bit weird but what is the reason to avoid `throw(ErrorException(...))`, apart from this being longer?

JohnAAbbott:
Is there some "positive" guidance about the usual/preferred way to report errors in Oscar code?
Given my background in C++, throwing an exception springs to mind.
In CoCoALib we throw only "CoCoA" exceptions so they can be easily distinguished from standard C++ exceptions.
Is there a similar policy to throw preferably "Oscar" exceptions?  Or is it preferred to re-use the existing Julia exceptions?

fieker:
On Wed, Mar 15, 2023 at 08:35:55AM -0700, JohnAAbbott wrote:
> Is there some "positive" guidance about the usual/preferred way to report errors in Oscar code?
> Given my background in C++, throwing an exception springs to mind.
> In CoCoALib we throw only "CoCoA" exceptions so they can be easily distinguished from standard C++ exceptions.
> Is there a similar policy to throw preferably "Oscar" exceptions?  Or is it preferred to re-use the existing Julia exceptions?
> 
using error is throwing an exception.
Apart from special situations we do not have special Oscar exceptions,
the vast majority is already covered by julia (and/or is illegal input)

HereAround:
> I do see that throwing strings is a bit weird but what is the reason to avoid `throw(ErrorException(...))`, apart from this being longer?

In extending, do you also suggest to avoid `throw(ArgumentError(...))`? If so, why?

thofma:
The call `throw(ArgumentError(...))` should not change.

But I support changing `throw(ErrorException(...))` to `error(...)` because it is much shorter.

JohnAAbbott:
Error exceptions serve 2 purposes:
(1)  if of the correct type, they can be caught and handled inside a calling function
(2) if not caught/handled, they should communicate useful information to the user (e.g. in an interactive session)

Is there a description of all/most situations in Oscar where we expect to catch and handle exceptions?
For exceptions which almost certainly won't be caught/handled, we could simply call Julia's `error` function, I guess.
My experience in CoCoA is that it is surprisingly tricky to make a good design for exceptions -- what we implemented in CoCoA did/does not work well.  For instance, it is not clear to me what extra useful information is embodied in an `ArgumentError` compared to an `ErrorException`... perhaps there is a distinction in randomized algorithms?

fieker:
On Thu, Mar 16, 2023 at 03:53:27AM -0700, JohnAAbbott wrote:
> Error exceptions serve 2 purposes:
> (1)  if of the correct type, they can be caught and handled inside a calling function
> (2) if not caught/handled, they should communicate useful information to the user (e.g. in an interactive session)
> 
> Is there a description of all/most situations in Oscar where we expect to catch and handle exceptions?
Outside very specific methods, we do not expect to see exceptions at
all.
The places that come to my mind are
 - loss of precision in real computations in Hecke (and here we throw a
   couple of specialized exception objects that are caught)
 - in modular methods, choice of a bad prime/ evaluation point. Again in
   the modular_proj (that handles the evluation) a special exception
   is thrown and caught
 - in Oscar there is some code in experimentatl/ModStd I think   
 - a possibly very bad example otherwise is in flint:
   function det(a::ZZModMatrix)

fieker (approved):

PR #2053, opened 2023-03-15
Add {characteristic,minimal}_polynomial aliases

### LINKED ISSUE #2026: `characteristic_polynomial` vs `charpoly`
We have a function `charpoly` with many methods -- and then we have `characteristic_polynomial`, with one method, for matroids. This is confusing.

Some possible solutions:
1. rename `characteristic_polynomial` to `charpoly`
2. the same but also add `@alias characteristic_polynomial charpoly`
3. rename `charpoly` to `characteristic_polynomial` in AA and add `@alias charpoly characteristic_polynomial` there

See also issue #1826, which in fact mentions `charpoly`; however I decided to open a dedicated issue because because we already are using both names with separate methods, which is IMHO worse than just having a name that is not 100% perfect.

fingolfin:
This resolves some confusion: there was a single `characteristic_polynomial` method, with a docstring that gave no hint about `charpoly`. It also prepares us for a potential future renaming of these functions.

As a natural extension also add `minimal_polynomial` as alias for `minpoly`.

Resolve #2026.

### DIFF
--- a/src/Combinatorics/Matroids/properties.jl
+++ b/src/Combinatorics/Matroids/properties.jl
@@ -912,7 +912,6 @@ function characteristic_polynomial(M::Matroid)
     R, q = polynomial_ring(ZZ, 'q')
     return (-1)^M.pm_matroid.RANK*tutte_polynomial(M)(1-q,0)
 end
-charpoly(M::Matroid) = characteristic_polynomial(M)
 
 @doc Markdown.doc"""
     reduced_characteristic_polynomial(M::Matroid)
--- a/src/Exports.jl
+++ b/src/Exports.jl
@@ -276,7 +276,6 @@ export character_lattice
 export character_parameters
 export character_table
 export character_to_rational_function
-export characteristic_polynomial
 export characteristic_subgroups, has_characteristic_subgroups, set_characteristic_subgroups
 export charpoly
 export chip_firing_move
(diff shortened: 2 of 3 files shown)


src/Combinatorics/Matroids/properties.jl
```diff
@@ -912,7 +912,6 @@ function characteristic_polynomial(M::Matroid)
     R, q = polynomial_ring(ZZ, 'q')
     return (-1)^M.pm_matroid.RANK*tutte_polynomial(M)(1-q,0)
 end
-charpoly(M::Matroid) = characteristic_polynomial(M)
```
afkafkafk13, line 915: Just above the deleted line, the characteristic_polynomial is defined by means of a method for this particular context (and below it the reduced_characteristic_polynomial for the same context). Have you tested that characteristic_polynomial for matroids is still correctly accessible/exported after the change? (This instance of characteristic_polynomial seems to be in the flavor of a chromatic polynomial, not of a characteristic polynomial in other contexts)

afkafkafk13: It is being tested in the CI by test/Combinatorics/Matroids/Matroids.jl. So everything should be fine.

afkafkafk13 (approved):

PR #2060, opened 2023-03-15
Simplify a hcat invocation

### LINKED ISSUE #1969: Code improvement (among others for ToricVarieties)
As pointed out by @fingolfin, we can replace  

`weights = matrix(ZZ,hcat([w[i] for i in 1:length(w)]))`

by

`weights = matrix(ZZ,hcat(w))`.

In @fingolfin 's words (cf. https://github.com/oscar-system/Oscar.jl/pull/1958)

> That's shorter and allocs less / is faster (not that this matter much here).
> 
> I think we could improve this in a bunch of places outside of your code, too...

fingolfin:
Resolves #1969

(I thought there'd be other places where this applies, but I've just checked and can't find any, so it seems I have been imagining things)


HereAround:
> (I thought there'd be other places where this applies, but I've just checked and can't find any, so it seems I have been imagining things)

Also I would have expected this to be applicable in other places. Surprising.

Thank you @fingolfin !

HereAround (approved):

PR #2062, opened 2023-03-16
Remove uses of Oscar.Graphs

fingolfin:


lkastner (approved):
Thanks, sorry I didn't catch these.

Issue #2049, opened 2023-03-15
Support mapping `quadratic_field` elements to GAP and back


fingolfin:
It would be great if we could support mapping elements of quadratic number fields to/from GAP. But the following right now gives an error:
```julia
F, sqrt3 = quadratic_field(3);
sigma = matrix(F, [1 0 0 ; 0 -1 0 ; 0 0 -1])
tau = matrix(F, [1//2 -sqrt3//2 0 ; sqrt3//2 1//2 0 ; 0 0 1])
G = matrix_group(sigma, tau)
order(G)
```
I guess we can just ask GAP to what element we should map the square root:
```
gap> Sqrt(3);
-E(12)^7+E(12)^11
```

Indeed, for now I can use this workaround:
```
F, z12 = CyclotomicField(12)
sqrt3 = sqrt(F(3))
```

ThomasBreuer:
Is the question to define a default conversion between quadratic number fields in Oscar and quadratic subfields of cyclotomic fields in GAP, via `Oscar.iso_oscar_gap`?
On the one hand, GAP's support for quadratic number fields as subfields of cyclotomic fields is better than its support for number fields via `AlgebraicExtension`.
On the other hand, computations with square roots represented as cyclotomics are not efficient.

Do we want to offer both conversions for quadratic (and cyclotomic) fields in Oscar to GAP, with either an `AlgebraicExtension` field or a field of cyclotomics on the GAP side, perhaps via an optional argument with value `IsAlgebraicExtension` or `IsCyclotomicCollection`?

fingolfin:
Honestly I don't care about the result as long as my example code above works. I guess on the long run we'd want that code to work by using `OscarMatrixObj` group elements on the GAP side, but as you know, we are still not there.

So, if you can rig something up to make that example work, I am fine no matter what it uses on the GAP side.

fingolfin:
Actually, I take that back: of course there is the code by @danielrademacher in `experimental/Matrix/matrix.jl` and `experimental/Matrix/matrix.g` which *does* work by wrapping. And it seems to work in this example just fine:
```
julia> F, sqrt3 = quadratic_field(3);

julia> sigma = matrix(F, [1 0 0 ; 0 -1 0 ; 0 0 -1]);

julia> tau = matrix(F, [1//2 -sqrt3//2 0 ; sqrt3//2 1//2 0 ; 0 0 1]);

[...]

PR #2072, opened 2023-03-16
natural embedding of quadratic fields into `QQAbField`

ThomasBreuer:
- moved `(K::QQAbField)(a::nf_elem)` from `experimental/GModule/GModule.jl` to `src/Rings/AbelianClosure.jl`

- extended the function to the case of elements `a` in quadratic number fields (addresses issue #2049, should resolve this issue together with #2056)

### DIFF
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -952,12 +952,6 @@ function gmodule(K::AnticNumberField, M::GModule{<:Any, <:Generic.FreeModule{nf_
   return gmodule(F, group(M), [hom(F, F, map_entries(K, mat(x))) for x = M.ac])
 end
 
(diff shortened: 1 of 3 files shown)


thofma:
Looking at https://github.com/oscar-system/Oscar.jl/pull/2072/files, a lot of the changed code is not called during the tests?

ThomasBreuer:
@thofma Thanks for the observation.

I do not understand what is going on.

The test logs contain `Test Summary` lines for `AbelianClousre` (yes, there is a typo ...) with 3293 total tests.
When I run the tests in `test/Rings/AbelianClosure.jl` manually in my local version then the summary lists 3287 total tests, and several hundreds of them belong to the code from this pull request (I have checked this by inserting `print` statements.)

fingolfin (approved):

Issue #2066, opened 2023-03-16
Weirdness with polytopes over a quadratic field


fingolfin:
Consider this input:
```
julia> F, t = quadratic_field(3)
(Real quadratic field defined by x^2 - 3, sqrt(3))

julia> a = diagonal_matrix(F.([1, -1, -1]))
[1    0    0]
[0   -1    0]
[0    0   -1]

julia> b = matrix(F, [1//2 -t//2 0
                      t//2  1//2 0
                         0     0 1])
[        1//2   -1//2*sqrt(3)   0]
[1//2*sqrt(3)            1//2   0]
[           0               0   1]

julia> G = matrix_group(a, b)
Matrix group of degree 3 over F

julia> o = orbit(G, *, F.([1,1,1]));

julia> vert = matrix(F, collect(o));

julia> ph = convex_hull(vert)
Polyhedron in ambient dimension 3
```
I think the last command is slightly buggy, because the output is this:
```
julia> typeof(ph)
Polyhedron{QQFieldElem}
```
I have to use the semi-undocumented optional first argument to `convex_hull` (it is documented only in the barest sense: the docstring mentions it, but provides no other information);
```julia-repl
julia> ph2 = convex_hull(nf_elem, vert)
Polyhedron in ambient dimension 3 with nf_elem type coefficients

julia> typeof(ph2)
Polyhedron{nf_elem}
```

Why do I even have to provide this? Can't it automatically determine it?

Anyway, I did not even notice the problem for quite a time, because everything worked, e.g. visualization. But then I tried this:

```
julia> ehrhart_polynomial(ph)
polymake:  WARNING: rule libnormaliz.ehrhartpoly: H_STAR_VECTOR, HILBERT_SERIES : RAYS | INPUT_RAYS failed: Some error in the normaliz input data detected: Generator degrees are too huge, h-vector would contain more than 10^6 entires. at /Users/mhorn/.julia/artifacts/e75f8bd18c56df0983c9054335c32294bb1eebac/share/polymake/bundled/libnormaliz/apps/polytope/rules/main.rules line 58.
ERROR: Can't locate object method "code" via package "1" (perhaps you forgot to load "1"?) at /Users/mhorn/.julia/artifacts/e75f8bd18c56df0983c9054335c32294bb1eebac/share/polymake/perllib/Polymake/Core/BigObjectType.pm line 1023.
```

[...]

PR #2077, opened 2023-03-17
Polyhedron: add and use is_lattice_polytope check

benlorenz:
enforce lattice property for several properties that depend on this

this fixes some parts of #2066.

cc: @lkastner

### DIFF
--- a/src/PolyhedralGeometry/Polyhedron/properties.jl
+++ b/src/PolyhedralGeometry/Polyhedron/properties.jl
@@ -509,7 +509,7 @@ julia> lattice_volume(C)
 8
 ```
 """
-lattice_volume(P::Polyhedron{QQFieldElem})::ZZRingElem = (pm_object(P)).LATTICE_VOLUME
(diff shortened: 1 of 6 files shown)


src/PolyhedralGeometry/Polyhedron/properties.jl
```diff
@@ -832,10 +833,11 @@ Compute the $h^*$ polynomial of `P`.
 # Examples
 ```jldoctest
 julia> c = cube(3)
-A polyhedron in ambient dimension 3
+Polyhedron in ambient dimension 3
 
 julia> h_star_polynomial(c)
 x^3 + 23*x^2 + 23*x + 1
+```
```
fingolfin, line 840: Off-topic, but: That's the second or third time I've recently seen this kind of issue (missing or misplaced triple backquote). I totally can see this happening to me as well, and I wonder if there is some tool or check we can activate to find these systematically?

lkastner (approved):

PR #2080, opened 2023-03-17
ToricVarieties: Support for immaculate line bundles

HereAround:


fingolfin (approved):
Looks good to me from a purely technical POV (I can't comment on the mathematical content).

Thanks!

src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl
```diff
@@ -65,14 +65,15 @@ Normal, non-affine, smooth, projective, gorenstein, fano, 2-dimensional toric va
...
-2
+julia> cohomology_indices(vs[3])
+1-element Vector{Int64}:
+ 2
 ```
 """
-function cohomology_index(tvs::ToricVanishingSet)
-    return tvs.i::Int
+function cohomology_indices(tvs::ToricVanishingSet)
+    return tvs.cis::Vector{T} where {T <: IntegerUnion}
```
afkafkafk13: Is the abreviation 'cis' really standard or should we better use a longer/clearer form? (Keeping in mind that not only toric geometers are using Oscar).
I agree, however, that 'cis' is ok as a shorthand for internal use only, if it is well hidden behind a getter function.

afkafkafk13 (commented):
Looks good to me, but we should wait for @lkastner to approve it.

src/AlgebraicGeometry/ToricVarieties/cohomCalg/special_attributes.jl
```diff
@@ -12,12 +12,56 @@ Compute the vanishing sets of an abstract toric variety `v` by use of the cohomC
...
+x₂ ≦ -2
+
+julia> print_constraints(polyhedra(ilb)[3])
+-x₂ ≦ -1
+x₁ - x₂ ≦ -2
+
+julia> print_constraints(polyhedra(ilb)[4])
+x₁ - x₂ ≦ -2
+x₁ ≦ -3
+"""
```
benlorenz:
````suggestion
x₁ ≦ -3
```
"""
````

File: src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl (hunk 1 above)
Comment (HereAround) [reply src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl]: This is a purely internal variable, which is never revealed to a user. It is (my) abbreviation for `cohomology_indices`. When writing this out, the code becomes clunky and hard to read.

**Comment by HereAround** [comment]
Thank you @benlorenz !

File: src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl (hunk 1 above)
Comment (afkafkafk13) [reply src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl]: Then I am fine with it.

Comment (HereAround) [reply src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl]: Thank you.

Comment (thofma) [inline src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl]:
```suggestion
    return tvs.cis
```
Don't see why it should be necessary.

thofma:
Could the field
```
cis::Vector{T} where {T <: IntegerUnion}
```
be just
```
cis::Vector{ZZRingElem}
```
with the appropriate conversion when initializing? This would be a bit cleaner and friendlier for the compiler, but maybe there is a performance reason to store it as `Int` if possible?

src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl
HereAround: I specified the return value, because I believed that this allows `Julia` to optimize the performance. Is this wrong? Or is the return value inferred differently here?

HereAround:
> Could the field
> 
> ```
> cis::Vector{T} where {T <: IntegerUnion}
> ```
> 
> be just
> 
> ```
> cis::Vector{ZZRingElem}
> ```
> 
> with the appropriate conversion when initializing? This would be a bit cleaner and friendlier for the compiler, but maybe there is a performance reason to store it as `Int` if possible?

You are right. This indeed looks cleaner. Thank you for the pointer.

As far as I can tell, there is no performance reason to store this as `Int`.

At least within the toric varieties, `IntegerUnion` is used frequently. Would you suggest that we change all of them to `ZZRingElem`?

src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl
thofma: The field of `tvs.cis` is defined as `Vector{<: IntegerUnion}` already in the definition of the structure. So julia will always know what type `.cis` is (or more precisely, it will know that `Vector{<: IntegerUnion}` is a supertype of `.cis`).

On the other hand, since this is not a concrete type, it is not that useful. Accessing `.cis` the compiler still does not know what it is exactly (could be for example `Vector{Int}`, `Vector{Union{Int, ZZRingElem}}` or `Vector{ZZRingElem}`) , so the performance will be bad.

thofma:
I think we allow `IntegerUnion` at many places for the convenience of the user and the programmer (for example, in your case when constructing the object), but internally, this will often be stored just as a `ZZRingElem`.

HereAround:
> I think we allow `IntegerUnion` at many places for the convenience of the user and the programmer (for example, in your case when constructing the object), but internally, this will often be stored just as a `ZZRingElem`.

Ok. That makes sense.

Hopefully not too far into the future, I shall have the time to take a look at all of the toric varieties and check if and where we could replace `IntegerUnion` by `ZZRingElem`. For that reason, I opened a corresponding issue: https://github.com/oscar-system/Oscar.jl/issues/2086

lkastner:
I don't really understand what is going on but that might be due to covid. I see that `IntegerUnion` makes sense in the places it currently exists for the toric varieties. But for the cohomology indices I thought this is like an index in some chain complex. Is that wrong? Do we really expect to perform so well here that we leave the realm of just `Int`? Until reading this discussion to me these indices were something completely different than say the coefficients of a divisor. And I still think that we don't have to be consistent in all places, i.e. we don't have to use `IntegerUnion` or `fmpz` or whichever everywhere. It should be angemessen.

src/AlgebraicGeometry/ToricVarieties/cohomCalg/special_attributes.jl
```diff
@@ -12,12 +12,57 @@ Compute the vanishing sets of an abstract toric variety `v` by use of the cohomC
...
+
+julia> print_constraints(polyhedra(ilb)[4])
+x₁ - x₂ ≦ -2
+x₁ ≦ -3
+```
+"""
+@attr function immaculate_line_bundles(variety::AbstractNormalToricVariety)
+    denominator_contributions = reduce(vcat, contributing_denominators(variety))
+    list_of_polyhedra = Polyhedron{QQFieldElem}[turn_denominator_into_polyhedron(variety, m) for m in denominator_contributions]
+    return ToricVanishingSet(variety, list_of_polyhedra, [i for i in 0:dim(variety)])
```
fingolfin: Could also do this (but feel free to leave it as it is)
```suggestion
    return ToricVanishingSet(variety, list_of_polyhedra, collect(0:dim(variety)))
```

HereAround:
> I don't really understand what is going on but that might be due to covid. I see that `IntegerUnion` makes sense in the places it currently exists for the toric varieties. But for the cohomology indices I thought this is like an index in some chain complex. Is that wrong? Do we really expect to perform so well here that we leave the realm of just `Int`? Until reading this discussion to me these indices were something completely different than say the coefficients of a divisor. And I still think that we don't have to be consistent in all places, i.e. we don't have to use `IntegerUnion` or `fmpz` or whichever everywhere. It should be angemessen.

I simply did not think long about `Int` vs. `IntegerUnion` vs `fmpz`/did not think it matter that much to people/for performance. But it appears that I was too naive...

My last rebase turned the cohomology indices into type `Int`.

src/AlgebraicGeometry/ToricVarieties/cohomCalg/VanishingSets/attributes.jl
```diff
@@ -65,14 +65,15 @@ Normal, non-affine, smooth, projective, gorenstein, fano, 2-dimensional toric va
...
+julia> cohomology_indices(vs[3])
+1-element Vector{Int64}:
+ 2
 ```
 """
-function cohomology_index(tvs::ToricVanishingSet)
-    return tvs.i::Int
+function cohomology_indices(tvs::ToricVanishingSet)
+    return tvs.cis
 end
```
lkastner:
```suggestion
cohomology_indices(tvs::ToricVanishingSet) = tvs.cis
```

HereAround:
@lkastner and I are discussing more on this in slack, as there exists similar (if not identical) functionality in `Polymake`. Hence, back to draft until we have evaluated the situation.

fingolfin:
What's the status of this PR?

HereAround:
There is functionality in `Polymake` that can compute immaculate bundles on toric spaces with different technology/algorithms. When @lkastner and I last discussed this matter, we agreed that it would be cool to support both variants/approaches. As such, it was agreed that @lkastner would adjust the corresponding `Polymake` functionality and report back once it is ready.

fingolfin:
@lkastner do you have any estimate when you might have time to get back to this?

Maybe a better (as in: less stressful for you, and overall saving time) approach would be to get this PR ready as-is, with one implementation; and then we can add the second Polymake based implementation whenever @lkastner has time for it?

HereAround:
As suggested by @fingolfin, it might be less stressful for @lkastner, if we merge this PR and eventually (whenever this may be) extend it with the corresponding Polymake functionality. Consequently, I have just rebased this PR.

lkastner (approved):

PR #2084, opened 2023-03-19
Style guide on optional arguments for parents of return values

HechtiDerLachs:
Re-import of the changes in #1935 . 

We had discussed this with some people at the last Polymake workshop and this was the outcome of our joint conclusions. Now, as it seems, this does not yet satisfy everybody and was eventually removed from #1935 . 

This is to hopefully address everybody else's concerns and doubts.

### DIFF
--- a/docs/src/DeveloperDocumentation/styleguide.md
+++ b/docs/src/DeveloperDocumentation/styleguide.md
@@ -194,6 +194,63 @@ end
 
 However, as always, rules sometimes should be broken.
 
+## Optional arguments for parents of return values
+
+Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, ... 
+Whenever a function creates such objects from an input which does not involve 
+the output's parent, we strongly recommend that 
+the user should have the possibility to pass on this 
+parent as a keyword argument under the name `parent`.
+Beyond that you can make more entry points for such parents available for 
+the user's convenience.
+
+Let's see an example. Say, you want to implement the characteristic 
+polynomial of a matrix. You could do it as follows:
+```julia
+  function characteristic_polynomial(A::MatrixElem)
+    kk = base_ring(A)
+    P, x = kk["x"]
+    AP = change_base_ring(P, A)
+    return det(AP - x*one(AP))
+  end
+```
+You can see that the polynomial ring `P`, i.e. the parent of the output, 
+is newly created in the body of the function. In particular, calling this 
+function two times on two different matrices `A` and `B` might produce 
+incompatible polynomials `p = det(A - x*one(A))` and `q = det(B - x*one(B))` 
+with different parents. Calling `p + q` will result in an error. 
+
+To solve this, we should have implemented the function differently:
+```julia
+  # Implementation of the recommended keyword argument signature:
+  function characteristic_polynomial(
+      A::MatrixElem;
+      parent::AbstractAlgebra.Ring=polynomial_ring(base_ring(A), :t)[1]
+    )
+    AP = change_base_ring(parent, A)
+    x = first(gens(ring))
+    return det(AP - x*one(AP))
+  end
+
+  # Optional second signature to also allow for the specification of the 
+  # output's parent as the first argument:
+  function characteristic_polynomial(
+      P::PolyRing,
+      A::MatrixElem
+    )
+    coefficient_ring(P) === base_ring(A) || error("coefficient rings incompatible")
+    return characteristic_polynomial(A, parent=P)
+  end
+```
+In fact this now allows for two different entry points for the parent ring `P` 
+of the output: First as the required `parent` keyword argument and second 
+as the first argument of a method of `characteristic_polynomial` with 
+an extended signature. Note that within the scope of the first method's body 
+the OSCAR function `parent` is necessarily overwritten by the name of the 
+keyword argument. Hence to call the actual parent of any other object, you 
+must then use `Oscar.parent`. E.g. to get the `MatrixSpace` of the 
+matrix `A`, write `Oscar.parent(A)`.
+
 
 ## Documentation
 


thofma:
No doubt that we should have this functionality, but should it be a keyword argument called `ring`? There at least two other convention that are commonly used in OSCAR land.

HechtiDerLachs:
> There at least two other convention that are commonly used in OSCAR land.

Then we were probably unaware of these other conventions. Could you point me towards some samples?

thofma:
1. For `map_coefficients`, the keyword argument is `parent`.
2. For `charpoly` of a matrix, the polynomial ring is an optional first argument.

HechtiDerLachs:
Alright, thanks. The optional first argument is supported in the style guide. Calling an optional argument `parent` overrides the `parent` function for the body of that function, doesn't it? I don't see what's wrong with `ring`, really. Another example from today: A function which constructs an ideal `my_precious_ideal(param1, param2; ring::Ring=QQ["x", "y"])`. Calling the optional argument `parent` simply doesn't make sense here. Is the problem that the styleguide sounds as if it was mandatory to call it "ring"? Then I could change that to explain that the programmer is free to choose their preferred name for the keyword, but it should be there and detectable by the user.

thofma:
No, the programmer should not be free to choose what he wants. We want consistency.

thofma:
Any opinions @fieker @fingolfin?

fieker:
A bad example: the way to crate ideals is essentially via `ideal(Ring, elems...)` or `ideal(elems...)`
Otherwise, I prefer `parent` in most situations, as it is generic and not tied to rings.
There is even a 3rd convention: passing the ring via an element (I don't like it, but check `cyclo` for a (bad) example).
Explicitly: `minpoly` / `charpoly`: I like one without the polynomial ring - as this reflects most of the casual use. Then passing a poly ring in as either 1st or via optional `parent` can both be defendend: `parent` via precedence (`map_coefficients`) and via 1st argument via considering a ring as a function or a target and following julia.
I can see a 3rd: passing an element in and expecting the polynomial to be evaluated there (I think used in `swinnerton_dyer`) which can be used here if the element is `x`...
However, none of those examples should use `ring`

simonbrandhorst:
@fieker  please state explicitly what should be changed so that this PR can move forward.

thofma:
Call it `parent` instead of `ring`.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+is newly created in the body of the function. In particular, calling this 
+function two times on two different matrices `A` and `B` might produce 
+incompatible polynomials `p = det(A - x*one(A))` and `q = det(B - x*one(B))` 
+with different parents. Then `p + q` will produce an error. 
+
+To solve this, we should have implemented the function differently:
+```julia
+  function characteristic_polynomial(
+      A::MatrixElem;
+      ring::AbstractAlgebra.Ring=(base_ring(A)["t"])[1]
```
simonbrandhorst:
```suggestion
      parent::AbstractAlgebra.Ring=(base_ring(A)["t"])[1]
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+    )
+    AP = change_base_ring(ring, A)
+    x = first(gens(ring))
+    return det(AP - x*one(AP))
+  end
+  function characteristic_polynomial(
+      P::AbstractAlgebra.Ring,
```
simonbrandhorst:
```suggestion
      parent::AbstractAlgebra.Ring,
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+incompatible polynomials `p = det(A - x*one(A))` and `q = det(B - x*one(B))` 
+with different parents. Then `p + q` will produce an error. 
+
+To solve this, we should have implemented the function differently:
+```julia
+  function characteristic_polynomial(
+      A::MatrixElem;
+      ring::AbstractAlgebra.Ring=(base_ring(A)["t"])[1]
+    )
+    AP = change_base_ring(ring, A)
```
simonbrandhorst:
```suggestion
    AP = change_base_ring(parent, A)
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+    x = first(gens(ring))
```
simonbrandhorst, line 231:
```suggestion
    x = first(gens(parent))
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+    return det(AP - x*one(AP))
+  end
+  function characteristic_polynomial(
+      P::AbstractAlgebra.Ring,
+      A::MatrixElem
+    )
+    coefficient_ring(P) === base_ring(A) || error("coefficient rings incompatible")
+    return characteristic_polynomial(A, ring=P)
```
simonbrandhorst:
```suggestion
    return characteristic_polynomial(A, parent=P)
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+      parent::AbstractAlgebra.Ring,
+      A::MatrixElem
+    )
+    coefficient_ring(P) === base_ring(A) || error("coefficient rings incompatible")
+    return characteristic_polynomial(A, parent=P)
+  end
+```
+This allows for two different entry points for the ring `P` of the output: 
+Once as the first argument of a method of `characteristic_polynomial` with 
+an extended signature, and second as an optional keyword argument for `ring` in
```
simonbrandhorst:
```suggestion
an extended signature, and second as an optional keyword argument for `parent` in 
```

simonbrandhorst (approved):

simonbrandhorst:
I replaced `ring` by `parent`.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
 However, as always, rules sometimes should be broken.
 
 
+## Optional arguments for parents of return values
+
+Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, ...
```
fingolfin, line 199: You could also link to the relevant section of the AbstractAlgebra documentation here?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
 However, as always, rules sometimes should be broken.
 
 
+## Optional arguments for parents of return values
+
+Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, ... 
+Whenever a function creates such objects from an input which does not involve 
+the output's parent, the user should have the possibility to pass on this
```
```suggestion
the output's parent, the caller should have the possibility to specify this 
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
 
 
+## Optional arguments for parents of return values
+
+Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, ... 
+Whenever a function creates such objects from an input which does not involve 
+the output's parent, the user should have the possibility to pass on this 
+parent as an optional argument. 
+
+Let's see an example. Say, you want to implement the characteristic
```
fingolfin, line 207:
```suggestion
As an example, suppose you want to add a function computing the characteristic 
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
 
+## Optional arguments for parents of return values
+
+Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, ... 
+Whenever a function creates such objects from an input which does not involve 
+the output's parent, the user should have the possibility to pass on this 
+parent as an optional argument. 
+
+Let's see an example. Say, you want to implement the characteristic 
+polynomial of a matrix. You could do it as follows:
```
fingolfin, line 208:
```suggestion
polynomial of a matrix (OSCAR of course already has this, but we will ignore this
for the moment for the sake of this example). You could implement it as follows:
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+```julia
+  function characteristic_polynomial(A::MatrixElem)
+    kk = base_ring(A)
+    P, x = kk["x"]
+    AP = change_base_ring(P, A)
+    return det(AP - x*one(AP))
+  end
+```
+You can see that the polynomial ring `P`, i.e. the parent of the output,
```
fingolfin, line 217:
```suggestion
The polynomial ring `P`, i.e. the parent of the output, 
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+    return det(AP - x*one(AP))
+  end
+```
+You can see that the polynomial ring `P`, i.e. the parent of the output, 
+is newly created in the body of the function. In particular, calling this 
+function two times on two different matrices `A` and `B` might produce 
+incompatible polynomials `p = det(A - x*one(A))` and `q = det(B - x*one(B))` 
+with different parents. Then `p + q` will produce an error. 
+
+To solve this, we should have implemented the function differently:
```
fingolfin, line 223:
```suggestion
To overcome this, the function should allow specifying the parent
of the return value:
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+Whenever a function creates such objects from an input which does not involve 
+the output's parent, the user should have the possibility to pass on this 
+parent as an optional argument. 
+
+Let's see an example. Say, you want to implement the characteristic 
+polynomial of a matrix. You could do it as follows:
+```julia
+  function characteristic_polynomial(A::MatrixElem)
+    kk = base_ring(A)
+    P, x = kk["x"]
```
fingolfin, line 212:
```suggestion
    P, x = kk[:x]
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+is newly created in the body of the function. In particular, calling this 
+function two times on two different matrices `A` and `B` might produce 
+incompatible polynomials `p = det(A - x*one(A))` and `q = det(B - x*one(B))` 
+with different parents. Then `p + q` will produce an error. 
+
+To solve this, we should have implemented the function differently:
+```julia
+  function characteristic_polynomial(
+      A::MatrixElem;
+      parent::AbstractAlgebra.Ring=(base_ring(A)["t"])[1]
```
```suggestion
      parent::AbstractAlgebra.Ring=(base_ring(A)[:t])[1]
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+    )
+    AP = change_base_ring(parent, A)
+    x = first(gens(parent))
```
fingolfin: Ironically, in this example it seems to me that specifying the intended generator `x` would be even nicer... ;-)

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
 However, as always, rules sometimes should be broken.
 
 
+## Optional arguments for parents of return values
+
+Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, ... 
+Whenever a function creates such objects from an input which does not involve 
+the output's parent, the user should have the possibility to pass on this 
+parent as an optional argument.
```
fingolfin: So here it specifically talks about an "optional" argument. But then in the example it shows *two* methods for specifying the caller, only one involving an optional argument. Nothing seems to resolve this?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -190,6 +190,54 @@ end
...
+    )
+    coefficient_ring(P) === base_ring(A) || error("coefficient rings incompatible")
+    return characteristic_polynomial(A, parent=P)
+  end
+```
+This allows for two different entry points for the ring `P` of the output: 
+Once as the first argument of a method of `characteristic_polynomial` with 
+an extended signature, and second as an optional keyword argument for `parent` in 
+the original method. This should be the general rule for such implementations 
+within OSCAR.
```
fingolfin: I still disagree with this: I don't want two ways to do this in general. There should be one way, and that should be implemented consistently (instead of some code implementing one, some the other, some both)

fingolfin (changes requested):
I still don't agree with this proposal as currently phrased.

docs/src/DeveloperDocumentation/styleguide.md
fieker: both are terrible - and should be replaced by `polynomial_ring(kk, cached = false)`. Within the system, `kk[:x]` and variants should never be used as they create global cached objects that clutter up the memory.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: No clue how to do that.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: Agreed. It is better to be consistent and have only one variant.  I have no preference here.

simonbrandhorst: Rethinking I would prefer 
`f(parent, input)`
`f(input)`
over 
`f(input; parent)`
`f(input)`
This matches our parent first philosophy for e.g. `ideal` and soon `factor`.
@fingolfin @thofma @fieker  any opinion?

fingolfin: @fieker just pointed out an issue with this approach: some functions like `map_coefficients` have another "canonical" first argument, namely a function to apply to all arguments. Here is one declaration:
```julia
map_coefficients(f, p::PolyRingElem{<: RingElement};
  cached::Bool=true, parent::PolyRing)
```

fingolfin:
We discussed this some more. We think this text should be less about *demanding* something and more about suggesting a best practice.

The only variant that (as far as we could tell) can always be provided is the `parent` keyword. So we can mention this as "recommended" way to do it. There should also be a note pointing out a common gotcha: if `parent` is an argument name, then to access the `parent()` function one needs to write `Oscar.parent()` instead.

We can also mention the other two patterns:
- the parent as an (optional?) first argument
- passing in a ring element (or multiple) and then the polynomial/series/whatever is evaluated at those.

... and then say that anything else is suspicious ;-).

simonbrandhorst:
Looks reasonable to me now. Can this be merged?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -194,6 +194,63 @@ end
...
+```
+In fact this now allows for two different entry points for the parent ring `P` 
+of the output: First as the required `parent` keyword argument and second 
+as the first argument of a method of `characteristic_polynomial` with 
+an extended signature. Note that within the scope of the first method's body 
+the OSCAR function `parent` is necessarily overwritten by the name of the 
+keyword argument. Hence to call the actual parent of any other object, you 
+must then use `Oscar.parent`. E.g. to get the `MatrixSpace` of the 
+matrix `A`, write `Oscar.parent(A)`.
+
```
fingolfin, line 253: @fieker mentioned a third option that is sometimes used: passing in a ring element (or at least a polynomial) and then the output has the same parent resp. is evaluated at that input ring element. E.g.:
```
@doc raw"""
    cyclotomic(n::Int, x::ZZPolyRingElem)

Return the $n$th cyclotomic polynomial, defined as
$$\Phi_n(x) = \prod_{\omega} (x-\omega),$$ where $\omega$ runs over all the
$n$th primitive roots of unity.
"""
function cyclotomic(n::Int, x::ZZPolyRingElem)
```

fingolfin (approved):
Looks good to me.

docs/src/DeveloperDocumentation/styleguide.md
fingolfin: Well...
```suggestion
Several objects in OSCAR have `parent`s, e.g. polynomials, group elements, and so on.
See [here](https://docs.oscar-system.org/stable/Nemo/developer/parents/) for more information.
```

PR #2085, opened 2023-03-20
Fix serialization so that there is no conflict with Hecke update to `defining_polynomial`

antonydellavecchia:
fixes serializtion so that there is no conflict with hecke update to defining polynomial


fingolfin:
Typo: "serializtion" -> "serialization" (in commit message and PR title)

fingolfin (approved):

thofma:
Thanks, that fixes the problem!

PR #2092, opened 2023-03-20
Towards moving FTheoryTools into OSCAR's experimental folder

HereAround:
Subject to various discussions and decisions, and so a draft-draft as of right now.

(Cf. https://github.com/Julia-meets-String-Theory/FTheoryTools.jl)

cc @afkafkafk13 @simonbrandhorst @HechtiDerLachs @apturner


experimental/FTheoryTools/.github/workflows/CI.yml
fingolfin, line 1: I realize you've been straight on importing all FTheoryTools files here. But there is no reason to keep this `.github` dir, is there?

simonbrandhorst:
We should mention the conflict in conventions regarding the projectivization of a vector bundle.

HechtiDerLachs:
> We should mention the conflict in conventions regarding the projectivization of a vector bundle.

Could you briefly summarize your lengthy discussion on Slack in this regard?

apturner:
As mentioned at [The Stacks Project page on projective bundles](https://stacks.math.columbia.edu/tag/01OA) (search for "Warning"), there are two opposite conventions in common use when defining the projectivization, and we seem to be using the opposite convention in our code to that of Oscar. This is evident when comparing our code (or our documentation pages) to the K3.jl example.

We both agree that for the K3 example the projective coordinates [x : y : z] of the ambient space should respectively transform as sections of $\mathcal{O}(4), \mathcal{O}(6), \mathcal{O}$. The difference is that in your convention, this means that the ambient bundle is $\mathbb{P}(\mathcal{O}(-4) \oplus \mathcal{O}(-6) \oplus \mathcal{O})$, while in our convention, it is $\mathbb{P}(\mathcal{O}(4) \oplus \mathcal{O}(6) \oplus \mathcal{O})$.

HechtiDerLachs:
Alright, makes sense. Yes, I implemented the projectivization of vector bundles so that the homogeneous coordinates of the fibers are polynomial functions on the fibers of the original bundle. So locally on the base, the homogeneous coordinates are a dual basis of a trivialization of the original vector bundle. This is the most geometric way to look at these things for me.

experimental/FTheoryTools/.github/workflows/CI.yml
HereAround: I wanted to migrate the complete `FTheoryTools` into the experimental folder first, then strip if off all obsolete files and finally integrate it. This is exactly what I have just done.

HereAround:
I have spread-out and integrated the code of the F-theory tools: https://github.com/oscar-system/Oscar.jl/pull/2092/commits/b5f322d3353bd4f8f1d12232067c21d1291ff2b3.

Among others, this step required to rename a few functions so that they align with the status of OSCAR after the big renaming.

The tests of F-theory tools take, as of the latest changes, up to 30 minutes (cf. https://github.com/Julia-meets-String-Theory/FTheoryTools.jl/actions/workflows/CI.yml). This is not acceptable in OSCAR. Currently, OSCAR's test suite already takes an excessive amount of time and some of the tests tend to time-out after 2 hours. Consequently, I have *not* activated the FTheoryTools tests. If desired in the future/the tests run faster, this can be changed by activating one line in `test/runtests.jl`. Note however that we execute doctests in the documentation of FheoryTools.

I hope that after my changes, the code aligns with the demands for `OSCAR/experimental`. If so, I would suggest to reserve other (minor) changes for future PRs.

For completeness, I have added a comment on the projectivization conventions in https://github.com/oscar-system/Oscar.jl/pull/2092/commits/24b64b0534e8edb8c2802e57062d2301101af5bc.

docs/oscar_references.bib
```diff
@@ -288,6 +288,19 @@ @Book{CLS11
   url           = {https://doi.org/10.1090/gsm/124}
 }
 
+@Book{CLS11*1,
+  author        = {Cox, David A. and Little, John B. and Schenck, Henry K.},
+  title         = {Toric varieties},
+  mrnumber      = {2810322},
```
fingolfin: Duplicate, this book is already in (as CLS11)

experimental/FTheoryTools/TateModels/attributes.jl
```diff
@@ -0,0 +1,348 @@
...
+Return the Tate section ``a_6``.
+
+```jldoctest
+julia> t = global_tate_model(test_base())
+Global Tate model over a concrete base
+
+julia> tate_section_a6(t);
+```
+"""
+@attr MPolyElem{fmpq} tate_section_a6(t::GlobalTateModel) = t.a6
```
fingolfin: `MPolyElem` is the old name, from before the big renaming. I think you should run the renaming script from https://github.com/oscar-system/Oscar.jl/wiki/The-Big-Renaming on this PR before we merge it. (Beware: that script can make some unwanted changes; I usually use `git add -p` resp. `git checkout -p` to selectively add/discard the changes it makes. A bit tedious but still only takes a few minutes.

fingolfin: Second issue: `MPolyElem{fmpq}` is not a concrete type, so it doesn't give type stability to prescribe this type. Instead this should use `QQMPolyRingElem`:
```suggestion
@attr QQMPolyRingElem tate_section_a6(t::GlobalTateModel) = t.a6
```
There are bunch of other places with the same problem.

experimental/FTheoryTools/WeierstrassModels/constructors.jl
```diff
@@ -0,0 +1,175 @@
...
+```
+"""
+function global_weierstrass_model(poly_f::MPolyElem{fmpq}, poly_g::MPolyElem{fmpq}, auxiliary_base_ring::MPolyRing, d::Int)
+    if (parent(poly_f) != auxiliary_base_ring) || (parent(poly_g) != auxiliary_base_ring)
+        throw(ArgumentError("All Weierstrass sections must reside in the provided auxiliary base ring"))
+    end
+    if d <= 0
+        throw(ArgumentError("The dimension of the base space must be positive"))
+    end
+    if length([string(k) for k in gens(auxiliary_base_ring)]) < d
```
fingolfin: Huh... isn't this the same as
```suggestion
    if ngens(auxiliary_base_ring) < d
```

experimental/FTheoryTools/WeierstrassModels/constructors.jl
```diff
@@ -0,0 +1,175 @@
...
+        throw(ArgumentError("We expect at least as many base variables as the desired base dimension"))
```
fingolfin: Could use the `@req` macro here and elsewhere

experimental/FTheoryTools/auxiliary.jl
```diff
@@ -0,0 +1,319 @@
...
+    return [a1, a2, a3, a4, a6]
+end
+
+function _tate_polynomial(base::AbstractNormalToricVariety, S::MPolyDecRing{fmpq, FmpqMPolyRing})
+    (a1, a2, a3, a4, a6) = _tate_sections(base)
+    return _tate_polynomial(S, [a1, a2, a3, a4, a6])
+end
+
+function _tate_polynomial(ais::Vector{<:MPolyElem{fmpq}}, S::MPolyDecRing{fmpq, FmpqMPolyRing})
+    x = gens(S)[length(gens(S))-2]
```
fingolfin: Don't use `gens(S)` as a cheap disposable like you would do in GAP: in fact, for certain rings, *every* call to `gens` allocates a new fresh deepcopy of the generating set. This function calls `gens(S)` eight times... 

Besides, to get generator `i`, you can also use `gen(S, i)` which again is in general more efficient than `gens(S)[i]`

experimental/FTheoryTools/TateModels/constructors.jl
```diff
@@ -0,0 +1,201 @@
...
+    if length(ais) != 5
+        throw(ArgumentError("We expect exactly 5 Tate sections"))
+    end
+    if any(k -> parent(k) != auxiliary_base_ring, ais)
+        throw(ArgumentError("All Tate sections must reside in the provided auxiliary base ring"))
+    end
+    if d <= 0
+        throw(ArgumentError("The dimension of the base space must be positive"))
+    end
+    if length([string(k) for k in gens(auxiliary_base_ring)]) < d
```
fingolfin: Again, isn't this just
```suggestion
    if ngens(auxiliary_base_ring) < d
```

experimental/FTheoryTools/TateModels/constructors.jl
```diff
@@ -0,0 +1,201 @@
...
+        throw(ArgumentError("We expect at least as many base variables as the desired base dimension"))
+    end
+
+    # convert Tate sections into polynomials of the auxiliary base
+    auxiliary_base_space = _auxiliary_base_space([string(k) for k in gens(auxiliary_base_ring)], d)
+    S = cox_ring(auxiliary_base_space)
+    ring_map = hom(auxiliary_base_ring, S, [gens(S)[i] for i in 1:length(gens(S))])
```
fingolfin: Why not just this?
```suggestion
    ring_map = hom(auxiliary_base_ring, S, gens(S))
```
and indeed, I think even this might work (but I might be wrong, please verify
```suggestion
    ring_map = hom(auxiliary_base_ring, S)
```

experimental/FTheoryTools/TateModels/constructors.jl
```diff
@@ -0,0 +1,201 @@
...
+end
+
+
+################################################
+# 4: Display
+################################################
+
+function Base.show(io::IO, t::GlobalTateModel)
+    if base_fully_specified(t)
+        join(io, "Global Tate model over a concrete base")
```
fingolfin: Why `join`? I think this should be changed to `print` here and in most (all?) other places

experimental/FTheoryTools/auxiliary.jl
```diff
@@ -0,0 +1,319 @@
...
+    zray[ncols(base_rays)+2] = -3
+
+    # Construct the rays of the toric ambient space
+    ambient_space_rays = hcat([r for r in base_rays], [-2 for i in 1:nrows(base_rays)], [-3 for i in 1:nrows(base_rays)])
+    ambient_space_rays = vcat(ambient_space_rays, transpose(xray), transpose(yray), transpose(zray))
+
+    # Construct the incidence matrix for the maximal cones of the ambient space
+    ambient_space_max_cones = []
+    for i in 1:nrows(base_cones)
+        push!(ambient_space_max_cones, [k for k in hcat([b for b in base_cones[i,:]], [1 1 0])])
```
fingolfin, line 39: WHy is the `k for k` and `b for b` necessary here?

I'd suggest to look at all the `hcat` calls in this PR, I suspect many of them could be optimized

docs/oscar_references.bib
HereAround: Corrected. Thank you!

experimental/FTheoryTools/WeierstrassModels/constructors.jl
HereAround: Applied everywhere (unless I missed an instance...). Thank you for the suggestion.

experimental/FTheoryTools/TateModels/constructors.jl
HereAround: I have tried to optimize the construction of all ring maps in FTheoryTools.

I received an error from your last suggestion, i.e. without `gens(S)`, but the suggestion above works quite well. In fact, in most places we need a subset of `gens(S)` for these maps. Unless I am mistaken, there is only instance where once could hope to get away without specifying the images. So maybe not a big drawback.

experimental/FTheoryTools/TateModels/constructors.jl
HereAround: I guess at the time of coding this, I was thinking of displaying some attributes in `Base.show`. However, as of now, this never happened. So `print` it is - at least for now.

experimental/FTheoryTools/auxiliary.jl
HereAround: Thank you for explaining that for certain rings, every call to gens allocates a new fresh deepcopy. I was absolutely not aware of this.

I have reduced the number of call to `gens`. This should be an improvement. There might be room for further optimization. You can have a look later. I hope to push a new polished version of this PR hopefully within the hour (but more likely, it will be the hour after that...)

experimental/FTheoryTools/auxiliary.jl
HereAround: I would be happy if it can be optimized. Thing is, I do not know how. Similarly, I do not know how to avoid the `k for k` and `b for b` here. For instance:
```julia
push!(ambient_space_max_cones, hcat(base_cones[i,:], [1 1 0]))
```
gives an `ArgumentError: number of columns of each array must match`.

Best to illustrate with an example what we are actually trying to achieve:

```julia
|julia> base = projective_space(NormalToricVariety,3)
Normal, non-affine, smooth, projective, gorenstein, fano, 3-dimensional toric variety without torusfactor

|julia> base_cones = matrix(ZZ, ray_indices(maximal_cones(base)))
[1   1   1   0]
[0   1   1   1]
[1   0   1   1]
[1   1   0   1]
```
Our goal is to make 3 copies of the first row. To the first copy, we append `[1 1 0]`, to the second copy `[1 0 1]` and to the third `[0 1 1]`. These rows - each consisting of 7 integer entries - are then made the first three rows of a new matrix. We repeat this step for all rows of the above matrix. At the end of all this, we should have constructed the following desired matrix:

```julia
[1   1   1   0   1   1   0 ]
[1   1   1   0   1   0   1 ]
[1   1   1   0   0   1   1 ]
[0   1   1   1   1   1   0 ]
...
[1   1   0   1   0   1   1 ]
```
Ultimately, for our purposes, the zeros do not matter (and this is not an allowed input for other functions). So we opt for the incidence matrix of this instead. This is what is achieved with the following lines of code:

```julia
|julia> ambient_space_max_cones = []
Any[]

julia> for i in 1:nrows(base_cones)
           push!(ambient_space_max_cones, [k for k in hcat([b for b in base_cones[i,:]], [1 1 0])])
           push!(ambient_space_max_cones, [k for k in hcat([b for b in base_cones[i,:]], [1 0 1])])
           push!(ambient_space_max_cones, [k for k in hcat([b for b in base_cones[i,:]], [0 1 1])])
       end
julia> ambient_space_max_cones = IncidenceMatrix(vcat(ambient_space_max_cones...))
12×7 IncidenceMatrix
[1, 2, 3, 5, 6]
[1, 2, 3, 5, 7]
[1, 2, 3, 6, 7]
[2, 3, 4, 5, 6]
[2, 3, 4, 5, 7]
[2, 3, 4, 6, 7]
[1, 3, 4, 5, 6]
[1, 3, 4, 5, 7]
[1, 3, 4, 6, 7]
[1, 2, 4, 5, 6]
[1, 2, 4, 5, 7]
[1, 2, 4, 6, 7]
```
(Yes, the very first line is bad since the type is not specified. Yet another thing that can be improved. But one thing at a time.)

Maybe there is a better way to achieve this?

experimental/FTheoryTools/TateModels/attributes.jl
HereAround: I have applied the renaming script. This seemed to worked quite well. (I will push the changes shortly and we can see if the tests pass. I very much do hope they do.)

I have not yet worked on the second issue that you raise. The renaming script replaced the above line with
```julia
@attr MPolyRingElem{QQFieldElem} tate_section_a6(t::GlobalTateModel) = t.a6
```
I asume you propose to remove `QQFieldElem`, i.e. modify it further to look like this:
```julia
@attr MPolyRingElem tate_section_a6(t::GlobalTateModel) = t.a6
```
Correct?

HereAround:
@fingolfin Thank you for the many suggestions towards improving this code. I have applied all, but two, of your suggestions. See above for details.

Also, I have activated those tests for `FTheoryTools`, that I anticipate to complete quickly (roughly a couple of minutes).

fingolfin (approved):
Looks good now. Yes, there are a few outstanding discussions, but I  just talked to @HereAround about them, and we can deal with them post-merge.

PR #2094, opened 2023-03-20
Add OscarInterface GAP package

fingolfin:
Its primary purpose: allowing us to write GAP packages that depend on OSCAR, resp. which behave differently under OSCAR.

Secondly, we also collect a bunch of GAP glue code that was scattered around in here.

More code may be added in the future.


gap/OscarInterface/PackageInfo.g
```diff
@@ -0,0 +1,99 @@
...
+  ArchiveURLSubset := ["doc"],
+  HTMLStart := "doc/chap0_mj.html",
+  PDFFile   := "doc/manual.pdf",
+  SixFile   := "doc/manual.six",
+  LongTitle := "GAP interface to OSCAR",
+),
+
+Dependencies := rec(
+  GAP := ">= 4.12",
+  NeededOtherPackages := [ ],
```
ThomasBreuer: Logically, JuliaInterface will be a needed package.
In practice, it is clear that JuliaInterface will be available when OscarInterface gets loaded.
I am not sure whether JuliaInterface should be listed here.

ThomasBreuer (approved):
Looks good.

What triggers that OscarInterface gets loaded?
I thought that the `__init__` function in `src/Oscar.jl` should do this, and then the various `__init*` functions that are called there for initializing GAP related stuff could perhaps be collected in OscarInterface.
Also the GAP code from `Oscar.jl/gap/misc.g` could be moved there.

src/Oscar.jl
```diff
@@ -198,6 +198,7 @@ function __init__()
         (GAP.Globals.IsSubgroupFpGroup, FPGroup),
     ])
     __GAP_info_messages_off()
+    GAP.Globals.SetPackagePath(GAP.Obj("OscarInterface"), GAP.Obj(joinpath(@__DIR__, "..", "gap", "OscarInterface")))
```
fingolfin, line 204: @ThomasBreuer `OscarInterface` is loaded here

gap/OscarInterface/PackageInfo.g
fingolfin: I'll add it

src/Oscar.jl
fingolfin: Ah, I just noticed I forgot the `LoadPackage` *sigh*. Fixed now.

fingolfin:
I've migrated a bunch of GAP code into `OscarInterface` now.

gap/OscarInterface/gap/OscarInterface.gi
```diff
@@ -0,0 +1,45 @@
+InstallOtherMethod( ImagesRepresentative,
```
ThomasBreuer, line 1:
```suggestion
DeclareAttribute( "JuliaData", IsObject );

InstallOtherMethod( ImagesRepresentative,
```

fingolfin: I forgot to `git add gap/OscarInterface/gap/OscarInterface.gd`

src/Oscar.jl
```diff
@@ -198,17 +198,20 @@ function __init__()
...
+    GAP.Globals.BindGlobal(GapObj("Oscar"), Oscar)
+    GAP.Globals.SetPackagePath(GAP.Obj("OscarInterface"), GAP.Obj(joinpath(@__DIR__, "..", "gap", "OscarInterface")))
+    GAP.Globals.LoadPackage(GAP.Obj("OscarInterface"))
     withenv("TERMINFO_DIRS" => joinpath(GAP.GAP_jll.Readline_jll.Ncurses_jll.find_artifact_dir(), "share", "terminfo")) do
       GAP.Packages.load("browse"; install=true) # needed for all_character_table_names doctest
     end
     GAP.Packages.load("ctbllib")
     GAP.Packages.load("forms")
     GAP.Packages.load("wedderga") # provides a function to compute Schur indices
     GAP.Packages.load("repsn")
```
fingolfin, line 212: I guess these could be turned into dependencies for OscarInterface. However, I still want to migrate to `GAP_pkg_repsn` etc. and I am not sure this would be the right thing then... In any case, it could still be done later, if we decide we want to do it after all.

ThomasBreuer: We could turn each such call into a suggested package, but I am not sure that this is a good idea.
As far as I see, it would not be possible to load OscarInterface with the OnlyNeeded option, such this change would be safe.
(I think the call `GAP.Packages.load("ferret"; install=true)` in `src/NumberTheory/GaloisGrp/GaloisGrp.jl` cannot be expressed in terms of GAP's dependencies.)

ThomasBreuer (approved):

Issue #2043, opened 2023-03-13
JuLie: bounds error in partitions function


benlorenz:
**Describe the bug**
BoundsError in partitions function. This appeared in the CI on master https://github.com/oscar-system/Oscar.jl/actions/runs/4398899970/jobs/7703066287#step:6:1134 :
```
combinatorics/partitions.jl: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/JuLie/partitions.jl:5
  Got exception outside of a @test
  BoundsError: attempt to access 3-element Vector{Int8} at index [4]
  Stacktrace:
    [1] setindex!
      @ ./array.jl:969 [inlined]
    [2] partitions(mu::Vector{Integer}, m::Int64, v::Vector{Integer}, n::Int64)
      @ Oscar.JuLie ~/work/Oscar.jl/Oscar.jl/experimental/JuLie/partitions.jl:586
    [3] macro expansion
      @ ~/work/Oscar.jl/Oscar.jl/test/Experimental/JuLie/partitions.jl:230 [inlined]
...
Test Summary:               | Pass  Error  Total   Time
combinatorics/partitions.jl |   26      1     27  18.8s
ERROR: LoadError: Some tests did not pass: 26 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/JuLie/partitions.jl:5
```

The tests seem to use random input but the code cannot deal with some of the generated input values.

**To Reproduce**

One example that triggers this would be:
```julia
using Oscar
# Julia code
julia> partitions(Integer[1, 4] , 17 , Integer[1, 4] , 3)
ERROR: BoundsError: attempt to access 3-element Vector{Int8} at index [4]
Stacktrace:
 [1] setindex!
   @ ./array.jl:969 [inlined]
 [2] partitions(mu::Vector{Integer}, m::Int64, v::Vector{Integer}, n::Int64)
   @ Oscar.JuLie /home/datastore/lorenz/software/julia/Oscar.jl/experimental/JuLie/partitions.jl:586
 [3] top-level scope
   @ REPL[53]:1
```

The code from the test file can be used to gather more failing examples:
```julia
julia> while true
         check = true
         N = 0:20
         for n in N
            for k = 1:n+1
               for i = 1:20
                  v = convert(Array{Integer,1}, rand(1:20,i))
                  unique!(v)
                  sort!(v)
[...]

PR #2095, opened 2023-03-21
Warham preparations

HechtiDerLachs:
Some functionality required by @wdecker .


simonbrandhorst:
There are conflicts.

experimental/Schemes/ProjectiveSchemes.jl
```diff
@@ -1014,3 +1014,29 @@ of ``X`` and ``Y``, respectively.
...
+function degree(P::AbsProjectiveScheme{<:Field})
+  return degree(affine_algebra(P))
+end
+
+function arithmetic_genus(P::AbsProjectiveScheme{<:Field})
+  h = hilbert_polynomial(P)
+  return (-1)^dim(P) * (first(coefficients(h)) - 1)
+end
+
+@attr function is_smooth(P::AbsProjectiveScheme)
```
fingolfin: Ideally specify the return type to allow calling code to be type stable
```suggestion
@attr Bool function is_smooth(P::AbsProjectiveScheme)
```

fingolfin: Also holds for the other `@attr` in here. But this is not a blocker, just a quick observation.

test/AlgebraicGeometry/Schemes/ProjectiveSchemes.jl
```diff
@@ -267,3 +267,48 @@ end
...
+  Q, _ = quo(S, I)
+  Y = ProjectiveScheme(Q)
+  @test affine_algebra(Y) === Q
+  @test dim(Y) == 2
+  @test degree(Y) == 4
+  @test is_smooth(Y)
+  @test arithmetic_genus(Y) == 1
+
+
+  Base.:/(x::RingElem, y::RingElement) = divexact(x, y; check=true)
```
fingolfin: This doesn't really belong in a test. Of course the next AA release will also contain this, so if you can wait for that (I hope we can release it this week, perhaps already tomorrow) ... Alternatively, adjust the function below and replace coefficients like `F(99)/52` by `F(99)//F(52)`

HechtiDerLachs:
@wdecker : I expect the smoothness test in your example to succeed within 3-4 hours. This is still way too much to include it in the test suite, but we should let it run on the compute servers and compare to the jacobian criterion. 

Edit: The bottleneck seems to really be an `is_zero` applied to matrices. If we could reduce the overhead involving conversion to the singular side, etc. that might speed up things significantly.

HechtiDerLachs:
@ClaudiaHeYun : There is some weird failure of the MacOS test here that might be of interest to you.

benlorenz:
> @ClaudiaHeYun : There is some weird failure of the MacOS test here that might be of interest to you.

That looks like https://github.com/oscar-system/Oscar.jl/issues/2043

ulthiel:
> That looks like #2043

I am working on a PR fixing this.

experimental/Schemes/ProjectiveSchemes.jl
```diff
@@ -1014,3 +1014,30 @@ of ``X`` and ``Y``, respectively.
   ff = CoveredSchemeMorphism(X, Y, phi)
   return ff
 end
+
+# Basic functionality required for Warham
+@attr Integer function dim(P::AbsProjectiveScheme{<:Field})
```
simonbrandhorst: There is no benefit for type stability in providing an abstract type. 
It is not always an int?

HechtiDerLachs: Who knows... It can change in the future. Maybe one day people even return `ZZRingElem`. 

Edit: No, seriously, I had to simply guess most of these return types. This is difficult to say. 

Edit: It might even be a `Singular` integer, because people get the idea to return the result directly from within singular. But Ok: We can just claim what we want to have here.

experimental/Schemes/ProjectiveSchemes.jl
```diff
@@ -1014,3 +1014,30 @@ of ``X`` and ``Y``, respectively.
   ff = CoveredSchemeMorphism(X, Y, phi)
   return ff
 end
+
+# Basic functionality required for Warham
+@attr Int64 function dim(P::AbsProjectiveScheme{<:Field})
```
simonbrandhorst:
```suggestion
@attr Int function dim(P::AbsProjectiveScheme{<:Field})
```
Concrete and does not depend on the architecture.

simonbrandhorst (approved):

experimental/Schemes/ProjectiveSchemes.jl
fingolfin: This serves multiple purposes:
- it documents the expected return type ;-)
- if some method returns something else, e.g. a Singular integer, we get an error, which is *good*
- and lastly, type stability.

So yeah, `Int` sounds good to me. If someone truly needs to work with schemes in dimension > 2^63, I think we have other problems to solve first to make that possible...

experimental/Schemes/ProjectiveModules.jl
```diff
@@ -178,63 +182,68 @@ function _is_projective_without_denominators(A::MatElem; unit::RingElem=one(base
...
-  for ((u, i, j), (_, Q, k), (Rloc, inc), lambda) in zip(entry_list, sub_results, sub_localizations, c)
-    # Lift the matrix Q to an n×n-matrix over R
-    Qinc = zero(matrix_space(R, n, n))
-    for r in 1:j-1
-      for s in 1:j-1
-        Qinc[r, s] = preimage(inc, Q[r, s])
-      end
-      for s in j+1:n
-        Qinc[r, s] = preimage(inc, Q[r, s-1])
+    result = zero(matrix_space(R, n, n))
```
fingolfin, line 188: Rule of thumb: `matrix_space` really only exists to have a parent for matrices, there is rarely a valid reason to use it. E.g. here it's better to do this:
```suggestion
    result = zero_matrix(R, n, n)
```
Same elsewhere.

This is no reason to hold up merging of this PR, I just wanted to make you aware of this :-)

test/Experimental/PlaneCurve-test.jl
```diff
@@ -337,12 +337,12 @@ end
     S, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
     T, _ = grade(S)
     C = Oscar.ProjPlaneCurve(T(y^2 * z - x^3 - x * z^2))
-    @test (@inferred Oscar.arithmetic_genus(C)) == 1
-    @test (@inferred Oscar.geometric_genus(C)) == 1
+    @test (Oscar.PlaneCurveModule.arithmetic_genus(C)) == ZZ(1)
+    @test (Oscar.geometric_genus(C)) == ZZ(1)
```
fingolfin, line 341: With the specification of return values for the various `@attr`, it should be possible to add the `@inferred` back in here. Moreover, it is perfectly fine to compare a `ZZRingElem` to an `Int`. So basically it should be possible to restore this whole test file? I hope?

fingolfin: (again: not a blocker, but maybe address in a future PR or at least keep in mind for future changes)

PR #2096, opened 2023-03-21
Fix decomposition of affine varieties into its components.

HechtiDerLachs:
@wdecker : This is the fix for your mail.


simonbrandhorst (approved):
Thank you for the fix.

Issue #1507, opened 2022-08-11
Revise serialization code to avoid constructing a full internal tree representation


fingolfin:
(I've been alluding to this for quite some time now but somehow never got around to creating the prototype I wanted to do; so for now I'll just try to write up the idea so I have a place to refer to should the need arise.)

Right now our serialization code builds a tree (or really, directed graph if you consider backrefs as edges) representing the data to be serialized, then (de)serializes that as `JSON`. This is quite simple to do, of course, but it can become a serious bottleneck for big data, and even for small data it can be noticeable if one wants to use it for IPC (interprocess communication). Of course the JSON encoding/decoding adds to that, for IPC it might be nice to be able to use something like protobuf instead.

I think we can solve this by restructuring our serialization API a bit. If done right, then we should be able separate the "high level" serialization code which knows about e.g. fmpz_mat, from the "low level" code which turns it into actual bytes (be it by creating an in-memory tree and serializing it as JSON, or be it by directly writing bytes into a memory buffer), without making it harder to write the serialization code. But it will require changes, so I'd rather do this sooner than later, otherwise the amount of code we need to change will grow a lot.

I don't claim to have a fully fledged plan for this, but here is a sketch of how this roughly works in other serialization frameworks I've worked on:
- let `s` denote the active `(De)SerializerState`; those states now may contain a stack tracking in some what the "active" object
- before and after calling `save_internal`, we update this stack "suitably" to track the active object; that may e.g. just mean pushing a new `Dict` onto the stack; and at the end, pop'ing the top `Dict` and adding it to the previous one
- then instead of writing say
```julia
function save_internal(s::SerializerState, K::SimpleNumField)
    return Dict(
        :def_pol => save_type_dispatch(s, defining_polynomial(K)),
        :var => save_type_dispatch(s, var(K)) 
    )
end
```
you might do something like
```julia
function save_internal(s::SerializerState, K::SimpleNumField)
    save_type_dispatch(:def_pol, save_type_dispatch(s, defining_polynomial(K)))
    save_type_dispatch(:var, save_type_dispatch(s, var(K)))
end
```
[...]

PR #2102, opened 2023-03-22
Adv/file formatting

antonydellavecchia:
Reformatting polynomial serialization for better efficiency
 the terms of a polynomial are now stored as
   terms = [(exponent, coeff), ..., (exponent, coeff)]
 where exponent is either an int or a vector of ints, and coeff is either a string
 or is given as a vector of terms. This results in a nested structure, the parents
 for the coefficients are found in the parents vector. The parent of the
coefficient can be determined based on the depth level of the given coefficient
 This new format also allows us to improve serialization of vectors of ring
elements, since the parents vector can be brought to the root, and we use
 the vector entries to store the terms


src/Serialization/main.jl
```diff
@@ -53,14 +53,21 @@ end
 # registerSerializationType is a macro to ensure that the string we generate
 # matches exactly the expression passed as first argument, and does not change
 # in unexpected ways when import/export statements are adjusted.
-macro registerSerializationType(ex::Any, str::Union{String,Nothing} = nothing)
+# It also sets the value of serialize_with_id, which determines
+# whether or not the type can be back referenced.
+macro registerSerializationType(
+    ex::Any,
+    uses_id::Bool = false,
+    str::Union{String,Nothing} = nothing)
```
fingolfin: Instead of forcing calls that pass a value for `str` to also pass one for `uses_id`, even if it is false, I think you can use "macro overloading: I.e. have two macros with the same name but different argument lists:
```julia-repl
julia> macro foo1(ex::Any, uses_id::Bool = false, str::Union{String,Nothing} = nothing)
         return :(:bar, $ex, $uses_id, $str)
       end
@foo1 (macro with 3 methods)

julia> @foo1 1+2
(:bar, 3, false, nothing)

julia> macro foo1(ex::Any, str::Union{String,Nothing})
         return :(:quux,$ex,false,$str)
       end
@foo1 (macro with 4 methods)

julia> @foo1 1+2
(:bar, 3, false, nothing)

julia> @foo1 1+2 false
(:bar, 3, false, nothing)

julia> @foo1 1+2 "house"
(:quux, 3, false, "house")
```

fingolfin: Oh, and those two macro implementations can share code: just move the bulk of the macro implementation into a regular function, then invoke it from both macro variants

antonydellavecchia:
closes https://github.com/oscar-system/Oscar.jl/issues/2190

src/Serialization/main.jl
```diff
@@ -53,16 +54,34 @@ end
...
+    if str === nothing
+      str = string(ex)
+    end
+    return esc(
+        quote
+            registerSerializationType($ex, $str)
+            encodeType(::Type{<:$ex}) = $str
+            # use_id is necessary for now since has_elem_basic_encoding is false for all types
+            # and not just parent types, this will have to be reworked in future
+            serialize_with_id(obj::T) where T <: $ex = !has_elem_basic_encoding(obj) && $uses_id
```
```suggestion
            serialize_with_id(obj::T) where T <: $ex = $uses_id && !has_elem_basic_encoding(obj)
```

src/Serialization/main.jl
```diff
@@ -73,36 +92,117 @@ function decodeType(input::String)
...
+        ref_dict = s.refs[Symbol(id)]
+        ref_dict[:id] = id
+        loaded_ref = load_unknown_type(s, ref_dict)
     end
+    return loaded_ref
+end
 
-    result = Dict{Symbol, Any}(:type => encodeType(T))
+function save_as_ref(s::SerializerState, obj::T) where T
+    if has_elem_basic_encoding(obj) && s.depth != 0
```
fingolfin: Maybe add a comment here. 
```suggestion
    # For rings whose elements have basic encoding (e.g. Z/7Z), we don't bother
    # storing backrefs to the ring, we just create a new instance of the ring.
    if has_elem_basic_encoding(obj) && s.depth != 0
```

Also: when can it happens that `s.depth == 0`? That would mean we invoke `save_as_ref` as the "top level", no? And anyway, how can we then store a backref if we created all those rings without an id ?!?!?

Maybe we can either give an example (in a comment, and/or a test); or else argue that it cannot happen (and then we just put in an `@assert s.depth != 0` or something like this)

src/Serialization/main.jl
```diff
@@ -73,36 +92,117 @@ function decodeType(input::String)
...
+    return Dict{Symbol, Any}(
+        :type => backref_sym,
+        :id => string(ref),
+    )
+end
+
+function save_type_dispatch(s::SerializerState, obj::T) where T
+    # this is used when serializing basic types like "3//4"
+    # when s.depth == 0 file should know it belongs to QQ
+    if s.depth != 0 && has_basic_encoding(obj)
```
fingolfin, line 218: So depth 0 is somewhat special. Maybe we need some test cases just for it, to make sure things work right?

src/Serialization/Fields.jl
```diff
@@ -1,3 +1,36 @@
...
+        return Any[]
+    end
+    base = parent(defining_polynomial(parent_ring))
+    parents = get_parents(base)
+    push!(parents, parent_ring)
+    return parents
+end
+
+function get_parents(parent_ring::T) where T <: Union{NfAbsNS, NfRelNS}
+    ngens = length(gens(parent_ring))
```
fingolfin: Strictly speaking instead of `length(gens(x))` one should do `ngens(x)`, *but* (a) I have not verified it exists in those case (I mean: it *should* but it might be one of the many places that we need to fix...); and (b) of course then the local variable shouldn't be called `ngens` anymore but e.g. `n`

src/Serialization/main.jl
```diff
@@ -53,16 +54,34 @@ end
 # registerSerializationType is a macro to ensure that the string we generate
 # matches exactly the expression passed as first argument, and does not change
 # in unexpected ways when import/export statements are adjusted.
+# It also sets the value of serialize_with_id, which determines
```
fingolfin, line 65: Actually it is more complex now, because also `has_elem_basic_encoding` plays a role... 

I am wondering now if we should rethink the relation between `has_elem_basic_encoding` and `serialize_with_id` and this macro... Like, when I specify that "elements have basic encoding", then we automatically don't store the id. Only if element do *not* have basic encoding is it even an option to store the id. Perhaps we can come up with a slightly less complex way to specify all of this?

src/Serialization/main.jl
fingolfin: Actually, now I wonder: shouldn't this use `serialize_with_id` here? I am confused, sorry :-)

src/Serialization/Fields.jl
```diff
@@ -410,34 +526,25 @@ end
 
 function save_internal(s::SerializerState, c::acb)
     return Dict(
-        :parent => save_type_dispatch(s, parent(c)),
-        :real => save_type_dispatch(s, real(c)),
-        :imag => save_type_dispatch(s, imag(c))
+        :vector => save_internal(s, [real(c), imag(c)])[:vector],
```
fingolfin: I am slighty concerned that we kind of hard-code here the requirement of first producing nested dictionaries, and then serializing them at the end. I still hope that we can arrange the code in a way that allows for other serialization formats which are e.g. more suitable for inter process communication (see issue #1507).

Of course we are already constructing a `Dict` in this function! But this wouldn't be too hard to change (see sketch in #1507). But if we now rely on the return value to be dict...

So maybe we can extract part of function which serializes the `Vector` into a new helper which is called from here (say `save_vector` or so...) and then just also call this from `save_internal(s, ::Vector,...)`

src/Serialization/containers.jl
```diff
@@ -3,77 +3,115 @@
...
     if is_basic_serialization_type(T)
         d[:entry_type] = encodeType(T)
     end
+
+    if !(T <: RingElem)
+        return d
+    end
+    entry_parent = parent(vec[1])
+    if has_elem_basic_encoding(entry_parent)
+        d[:parent] = save_type_dispatch(s, entry_parent)
```
fingolfin: Is this implicitly relying on the fact that `serialize_with_id` will return `false` for this parent? Or what's going on here?

It seems that `has_elem_basic_encoding(X)` has two meanings at once:
1. the element of `X` have a "basic encoding"
2. `X` itself can be encoded in a "basic" fashion, namely we don't need to encode any of its baserings etc. ?

I think it might help to separate these two concepts into two traits (even if in the end they will always have equal value)

src/Serialization/Rings.jl
```diff
@@ -1,3 +1,19 @@
+################################################################################
+# Utility functions for ring parent tree
+
+# builds parent tree
+function get_parents(parent_ring::Ring)
+    base = base_ring(parent_ring)
+    if has_elem_basic_encoding(base)
+        return Any[parent_ring]
+    end
```
fingolfin: I don't quite understand what's going on here. Why isn't it
```suggestion
    if has_elem_basic_encoding(parent_ring)
        return Any[]
    end
    base = base_ring(parent_ring)
```

fingolfin: ... and then it would be basically identical to this one from above (so that method could be removed)
```julia

 function get_parents(parent_ring::T) where T <: Union{FracField,
                                                       AbstractAlgebra.Generic.RationalFunctionField,
                                                       AbstractAlgebra.Generic.LaurentSeriesField}
     if has_elem_basic_encoding(parent_ring)
         return Any[]
     end

     base = base_ring(parent_ring)
     parents = get_parents(base)
     push!(parents, parent_ring)
     return parents
 end
```

src/Serialization/main.jl
```diff
@@ -73,36 +92,117 @@ function decodeType(input::String)
...
+
+has_basic_encoding(obj::T) where T = is_basic_serialization_type(T)
+
+function has_basic_encoding(obj::T) where T <: RingElem
+    return has_elem_basic_encoding(parent(obj))
+end
+
+has_elem_basic_encoding(obj::T) where T = false
+has_elem_basic_encoding(obj::ArbField) = true
+has_elem_basic_encoding(obj::AcbField) = true
```
fingolfin: I think before we merge this PR, these definitions should be merged next to the corresponding `@registerSerializationType` calls.

src/Serialization/main.jl
```diff
@@ -73,36 +92,117 @@ function decodeType(input::String)
...
+is_basic_serialization_type(::Type{T}) where T <: Number = isconcretetype(T)
+
+# ATTENTION
+# We need to distinguish between data with a globally defined normal form and data where such a normal form depends on some parameters.
+# In particular, this does NOT ONLY depend on the type; see, e.g., FqField.
+function has_elem_basic_encoding(obj::T) where T <: Ring
+    return is_basic_serialization_type(elem_type(obj))
+end
+
+has_basic_encoding(obj::T) where T = is_basic_serialization_type(T)
```
fingolfin, line 168: We need some comments that explain the (intended) semantics for `has_elem_basic_encoding` and `has_basic_encoding` and in particular clarify how they differ from `is_basic_serialization_type` ?

src/Serialization/Rings.jl
```diff
@@ -88,161 +95,185 @@ end
...
+    parents = []
+    encoded_terms = []
 
     for i in 1:length(p)
-        term = Dict(
-            :exponent => save_type_dispatch(s, exponent_vector(p, i)),
-            :coeff => save_type_dispatch(s, coeff(p, i))
-        )
-        push!(terms, term)
+        if coeff == parent_ring(0)
```
fingolfin:
```suggestion
        if iszero(coeff(p, i))
```
Does this ever happen?

src/Serialization/Rings.jl
```diff
@@ -88,161 +95,185 @@ end
...
+            continue
+        end
+        encoded_coeff = save_internal(s, coeff(p, i))
+        if has_elem_basic_encoding(base)
+            push!(encoded_terms,  (exponent_vector(p, i), encoded_coeff))
+        else
+            parents = encoded_coeff[:parents]
+            push!(encoded_terms, (exponent_vector(p, i), encoded_coeff[:terms]))
+        end
```
fingolfin: Before the loop:
```julia
    parents = save_parents_internal(s, SOMETHING)
```
then here
```suggestion
        # no need to serialize the parents here, we already did;
        # in this case, the serialization of the coefficient does not
        # need to store the parents; and also it can directly store (return) the terms, instead of putting them into a sublist
        encoded_coeff = save_internal(s, coeff(p, i); omit_parents=true)
        push!(encoded_terms, (exponent_vector(p, i), encoded_coeff))
```

src/Serialization/Rings.jl
```diff
@@ -88,161 +95,185 @@ end
...
+    base = base_ring(parent_ring)
+    polynomial = MPolyBuildCtx(parent_ring)
+    for term in terms
+        e, coeff = term
+        if length(parents) == 1
+            coeff_type = elem_type(base)
+            if has_elem_basic_encoding(base)
+                c = load_type_dispatch(s, coeff_type, coeff, parent=base)
+            else
+                c = base(c)
```
fingolfin: Maybe this?
```suggestion
                c = base(coeff)
```

fingolfin: Or if you can't get here, maybe add an `@assert` or so?

src/Serialization/Rings.jl
```diff
@@ -88,161 +95,185 @@ end
...
+    return parent_ring(loaded_terms)
+end
 
-    return R(coeffs)
+function load_terms(s::DeserializerState, parents::Vector, terms::Vector,
+                    parent_ring::Union{MPolyRing, UniversalPolyRing})
+    base = base_ring(parent_ring)
+    polynomial = MPolyBuildCtx(parent_ring)
+    for term in terms
+        e, coeff = term
```
fingolfin:
```suggestion
    for e, coeff in terms
```

src/Serialization/main.jl
antonydellavecchia: this now checks if serialize with id is false and returns the encoded obj, storing of any  parent types now go through a save_ref, maybe this means the function deserves a name change to save_parent

src/Serialization/Fields.jl
antonydellavecchia: save_internal for vector now has an include_parents flag

src/Serialization/Rings.jl
antonydellavecchia: this never happens and was removed

src/Serialization/Rings.jl
antonydellavecchia: resolved by adding an include_parents flag for save_internal of any ring elements, we use the get_parent_refs function now if the flag is true and returns a dict with a parents key, otherwise just returns the terms as a vector

src/Serialization/Rings.jl
antonydellavecchia: line 208 could never be reached

src/Serialization/main.jl
antonydellavecchia: I've opted to keep these two ideas seperate to avoid confusion or possible inconsistences in the future. i.e. Objects whose elements have basic encodings may also be stored with an id. Having elements with basic encoding just means that the object has no base_ring, and represents a sort of atomic fundamental type at least within oscar.

antonydellavecchia: I've also removed the dependancy to the has_elem_basic_encoding

src/Serialization/containers.jl
antonydellavecchia: I've added a note above has_elem_basic_encoding definition, to clarify what is meant and how it is used

src/Serialization/main.jl
antonydellavecchia: tests have been added

antonydellavecchia:
closes https://github.com/oscar-system/Oscar.jl/issues/2191

antonydellavecchia:
closes https://github.com/oscar-system/Oscar.jl/issues/2193

antonydellavecchia:
closes https://github.com/oscar-system/Oscar.jl/issues/2513
closes https://github.com/oscar-system/Oscar.jl/issues/2514

src/Serialization/Fields.jl
```diff
@@ -241,37 +398,49 @@ end
...
+                    parent_ring::FracField)
+    num_coeff, den_coeff = terms
+    loaded_num = load_terms(s, parents[1:end - 1], num_coeff, parents[end - 1])
+    loaded_den = load_terms(s, parents[1:end - 1], den_coeff, parents[end-1])
+    return  parent_ring(loaded_num) // parent_ring(loaded_den)
+end
 
-    return R(num) // R(den)
+function load_internal(s::DeserializerState,
+                       ::Type{<: FracElem},
```
fingolfin:
```suggestion
                       ::Type{<:FracElem},
```

src/Serialization/Fields.jl
```diff
@@ -18,23 +56,33 @@ function load_internal(s::DeserializerState, ::Type{Nemo.fpField}, dict::Dict)
...
-        :data => Nemo.data(elem)
-    )
+function save_internal(s::SerializerState, elem::fpFieldElem; include_parents::Bool=true)
+    if include_parents
+        return Dict(
+            :parent => save_as_ref(s, parent(elem)),
+            :class_rep => string(elem)
+        )
+    end
+    return string(elem)
```
fingolfin, line 66: This pattern repeats over and over again. I think there is an opportunity to factor out this duplication (possibly in a future PR)...

Is ther already a trait function `is_FOO` which returns true for all types that honor `include_parents` (so like `fpFieldElem` here)? If not, we could add one, that would help with this refactoring. 

And then I think we could handle it uniformly at the top level (just like you already handle it uniformly when loading, I think), by editing `save_type_dispatch`: just replace  `foo = save_internal(s, obj)` by something like
```julia
   if is_type_storing_parents(typeof(obj))
      p = save_as_ref(s, parent(obj))
      foo = Dict(:parent => p, :data => save_internal(s, obj))
   else
      foo = save_internal(s, obj)
   end
```
Of course if instead of the generic `:data` we want something more bespoke like `:class_rep`, that's also possible (e.g. add a function ```
tag_for_serialization_data(::Any) = :data
tag_for_serialization_data(::fpFieldElem) = :class_rep
```
and use that appropriately.

**Review by benlorenz — APPROVED** [review]
This is already a great improvement and implements many of the requested features. I did not see any major issues in the code and I think to move forward we should merge this as soon as possible and defer more changes to future PRs.

PS: nightly failures are unrelated.

File: src/Serialization/main.jl [hunk 18]
```diff
@@ -81,36 +103,103 @@ function decodeType(input::String)
...
+# In particular, this does NOT ONLY depend on the type; see, e.g., FqField.
+
+# Objects have a elements that have a basic encoding are fundamental objects. They are "atomic"
+# with regards to the available types in Oscar (maybe in general), they are base cases when
+# a parent tree is constructed via recursion, and they can be decoded from a string or array of
+# strings provided their parent is known. All types with basic serialization have a basic encoding
+# has_elem_basic_encoding also deals with basic parametrized types.
+
+function has_elem_basic_encoding(obj::T) where T <: Ring
+    return is_basic_serialization_type(elem_type(obj))
```
fingolfin, line 165: The relation between `is_basic_serialization_type`, `has_elem_basic_encoding` and `has_basic_encoding` still is confusing to me; and the latter two names are so similar that I keep misreading one for the other. But I don't want to hold up this PR for that, I'll try and see if I can come up with better alternatives, this can still be changed in the future.

src/Serialization/main.jl
```diff
@@ -81,36 +103,103 @@ function decodeType(input::String)
...
+the vector.
+
+# Examples
+
+```jldoctest
+julia> is_basic_serialization_type(ZZRingElem)
+true
+```
+"""
+is_basic_serialization_type(::Type) = false
```
fingolfin, line 135: I am still not happy with `is_basic_serialization_type` as a name but I don't want to hold up this PR further, so I'll think about a better name later.

src/Serialization/main.jl
```diff
@@ -61,16 +62,37 @@ end
 # registerSerializationType is a macro to ensure that the string we generate
 # matches exactly the expression passed as first argument, and does not change
 # in unexpected ways when import/export statements are adjusted.
+# It also sets the value of serialize_with_id, which determines
+# whether or not the type can be back referenced.
```
fingolfin, line 66:
```suggestion
# whether or not the type can be back referenced.
# If omitted, the default is that no back references are allowed.
```

src/Serialization/main.jl
```diff
@@ -61,16 +62,37 @@ end
...
+                                   uses_id::Bool,
+                                   str::Union{String,Nothing} = nothing)
+    if str === nothing
+      str = string(ex)
+    end
+    return esc(
+        quote
+            registerSerializationType($ex, $str)
+            encodeType(::Type{<:$ex}) = $str
+            # There exists types where equality cannot be discerned from the serialization
```
```suggestion
            # There exist types where equality cannot be discerned from the serialization
```

src/Serialization/main.jl
```diff
@@ -61,16 +62,37 @@ end
...
+            # these types require an id so that equalities can be forced upon load.
```
fingolfin, line 79: I don't understand the preceding two lines. Perhaps an example would be illustrative.. or perhaps two? one for when this is needed and one when not?

src/Serialization/main.jl
```diff
@@ -61,16 +62,37 @@ end
...
+            # The ids are only necessary for parent types, checking for element type equality
+            # can be done once the parents are known to be equal.
+
+            serialize_with_id(obj::T) where T <: $ex = $uses_id
```
fingolfin, line 91: So is the "id" the same as the "uuid"?

src/Serialization/main.jl
```diff
@@ -81,36 +103,103 @@ function decodeType(input::String)
...
+is_basic_serialization_type(::Type{Bool}) = true
+is_basic_serialization_type(::Type{Symbol}) = true
+# this deals with int32, int64 etc.
+is_basic_serialization_type(::Type{T}) where T <: Number = isconcretetype(T)
+
+# ATTENTION
+# We need to distinguish between data with a globally defined normal form and data where such a normal form depends on some parameters.
+# In particular, this does NOT ONLY depend on the type; see, e.g., FqField.
+
+# Objects have a elements that have a basic encoding are fundamental objects. They are "atomic"
```
fingolfin: "Objects have a elements" ??

Is this only talking about "things that have elements" (like rings, groups, ...) or about anything that has basic serialization?

src/Serialization/main.jl
```diff
@@ -81,36 +103,103 @@ function decodeType(input::String)
...
+# with regards to the available types in Oscar (maybe in general), they are base cases when
+# a parent tree is constructed via recursion, and they can be decoded from a string or array of
+# strings provided their parent is known. All types with basic serialization have a basic encoding
+# has_elem_basic_encoding also deals with basic parametrized types.
```
fingolfin, line 151: So... "basic encoding" = "basic serialization"? If so, perhaps we should align the names? If not, i.e. there is a difference here: what is it?

fingolfin (commented):
I think there are great improvements here, but there is still quite a bit to be done. Unfortunately I don't have time right now to give this the full review it deserves.

Since @micjoswig and @antonydellavecchia want to work on this, feel free to merge it.

The one drawback is that then it gets harder to review these changes  (unless one goes back to this PR), but so be it.

src/Serialization/main.jl
antonydellavecchia: yes

src/Serialization/main.jl
antonydellavecchia: i've added some more details describing this to the code. Basically the difference is that encodings rely on a parent.
so some basic encodings use basic serialization, elements of ZZ for example, however, numbers of type Int64 dont really have an encoding since they don't really have a parent, same for the type Symbol, Bool. We also have ZZ/nZZ whose elements have a basic encoding but not a basic serialization, we cannot just store "6" and know it represents "6" in ZZ/7ZZ, we need to know the parent

PR #2104, opened 2023-03-22
Provides Examples for the study of surfaces

wdecker:
This requires a thorough introduction which I will write as part of a subsequent PR.


docs/doc.main
```diff
@@ -194,7 +194,8 @@
          "AlgebraicGeometry/Curves/para_rational_curves.md",
      ],
      "Surfaces" => [
-        "AlgebraicGeometry/Surfaces/K3Surfaces.md"
+        "AlgebraicGeometry/Surfaces/K3Surfaces.md",
+	"AlgebraicGeometry/Surfaces/SurfacesP4.md",
```
fingolfin:
```suggestion
        "AlgebraicGeometry/Surfaces/SurfacesP4.md",
```

wdecker:
Why was this test cancelled?

thofma:
Took too long. Can be ignored here.

docs/src/AlgebraicGeometry/Surfaces/SurfacesP4.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Rational Surfaces
+
+#### A Rational Surface with $d=3$, $\pi=0$
+
+```@docs
+cubic_scroll()
+```
+
+#### A Rational Surface with $d=4$, $\pi=0$
```
fingolfin, line 26: Wouldn't it make more sense to move this description into the respective docstrings, so that REPL users can also see it?

(I am fine either way, just wondering)

fingolfin (approved):
Preview: https://docs.oscar-system.org/previews/PR2104/AlgebraicGeometry/Surfaces/SurfacesP4/

Fine by me as a start, can be refined later.

docs/src/AlgebraicGeometry/Surfaces/SurfacesP4.md
wdecker: But this information is in the respective doc strings. I do not understand?

fingolfin:
One more remarks (not a change request, let's merge this as-is for now): the new file is 1.2 MB, most of which is taken for polynomials. Just parsing it is quite some work for Julia.

I think these ideals (resp. the schemes) would be an ideal candidate for our serialization framework. Perhaps someone (@antonydellavecchia ? @lkastner ?) is interested in experimenting with serialization some of these (either the ideals or the full Scheme objects). In particular I wonder how the resulting JSON file size compares to the Julia code; and how fast loading is. (Another question is whether each surface must be stored in a separate file, or whether it would be possible to store them in one file but only decode the parts we actually need)

But again, that's future work, not meant to block this PR.

PR #2105, opened 2023-03-22
improve `experimental/Matrix`

ThomasBreuer:
Up to now, the stored `NiceMonomorphism` did not support computing images and preimages.
Now a `GroupHomomorphismByFunction` is used, which uses the known reduction from characteristic zero to the finite field; this yields the computation of images.
For preimages, the current implementation is just a workaround for the problem described at https://github.com/gap-system/gap/issues/5422, the special GAP code can be removed as soon as this problem is solved on the GAP side.

By the way, using tha name `matrix_group` for an Oscar function that returns a GAP object looks confusing.


fingolfin (approved):

fingolfin:
> By the way, using `matrix_group` for a function that returns a GAP object looks confusing.

Agreed. Feel free to change it (rename it, or have it return a "proper" OSCAR matrix group?)

PR #2106, opened 2023-03-22
Lazy toric schemes

HechtiDerLachs:
If our marriage of toric stuff with the general schemes is to compete in any way to the purely toric world, we need to postpone creating the algebraic objects until needed. This is a first step towards this end.

### DIFF
--- a/experimental/Schemes/ToricSchemes/AffineToricSchemes/Attributes.jl
+++ b/experimental/Schemes/ToricSchemes/AffineToricSchemes/Attributes.jl
@@ -34,7 +34,6 @@ function underlying_scheme(X::ToricSpec)
   end
   return X.X
 end
-export underlying_scheme
 
 
 @doc Markdown.doc"""
@@ -59,7 +58,6 @@ Normal, affine toric variety
 ```
 """
 affine_normal_toric_variety(X::ToricSpec) = X.antv
-export affine_normal_toric_variety
 
 
 @doc Markdown.doc"""
@@ -84,7 +82,6 @@ Polyhedral cone in ambient dimension 2
 ```
 """
 cone(X::ToricSpec) = cone(affine_normal_toric_variety(X))
-export cone
 
 
 @doc Markdown.doc"""
@@ -117,7 +114,6 @@ function dual_cone(X::ToricSpec)
   end
   return X.dual_cone
 end
-export dual_cone
 
 
 @doc Markdown.doc"""
@@ -158,5 +154,4 @@ function hilbert_basis(X::ToricSpec)
   end
   return X.hb
 end
-export hilbert_basis
 
--- a/experimental/Schemes/ToricSchemes/AffineToricSchemes/Attributes.jl
+++ b/experimental/Schemes/ToricSchemes/AffineToricSchemes/Attributes.jl
@@ -18,7 +18,7 @@ julia> affine_toric_scheme = ToricSpec(antv)
 Spec of an affine toric variety with cone spanned by RayVector{QQFieldElem}[[1, 0], [0, 1]]
 
 julia> underlying_scheme(affine_toric_scheme)
-Spec of Quotient of Multivariate Polynomial Ring in x1, x2 over Rational Field by ideal()
+Spec of Quotient of Multivariate Polynomial Ring in x1, x2 over Rational Field by ideal(0)
 ```
 """
 function underlying_scheme(X::ToricSpec)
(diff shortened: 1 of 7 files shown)


fingolfin:
A doctest is failing (in experimental/Schemes/ToricSchemes/AffineToricSchemes/Attributes.jl line 10ff ?)

experimental/Schemes/ToricSchemes/AffineToricSchemes/Attributes.jl
```diff
@@ -135,5 +152,11 @@ julia> hilbert_basis(affine_toric_scheme)
...
 ```
 """
-hilbert_basis(X::ToricSpec) = X.hb
+function hilbert_basis(X::ToricSpec)
+  if !isdefined(X, :hb)
+    X.hb = matrix(ZZ, hilbert_basis(dual_cone(X)))
+  end
+  return X.hb
+end
 export hilbert_basis
```
simonbrandhorst: All exports are at the top of the file and sorted alphabetically now.

experimental/Schemes/ToricSchemes/NormalToricSchemes/Attributes.jl
```diff
@@ -18,7 +18,112 @@ julia> underlying_scheme(toric_scheme)
...
+        # Step 1: Find the cone of the intersection of X and Y.
+        facet = intersect(CX, CY)
+        # harvest the dual cone
+        dual_facet = polarize(facet)
+        if !(dim(facet) == dim(CX) - 1)
+          continue
+        end
+
+        # Step 2: Extract the candidates for the localization element.
+        candidates = dual_facet.pm_cone.HILBERT_BASIS_GENERATORS[2]
```
simonbrandhorst, line 52: are you aware of the `lineality_space` of a cone?

experimental/Schemes/ToricSchemes/NormalToricSchemes/Attributes.jl
```diff
@@ -18,7 +18,112 @@ julia> underlying_scheme(toric_scheme)
...
+        AX = transpose(hilbert_basis(X))
+        Id = identity_matrix(ZZ, ncols(AX))
+        sol = solve_mixed(AX, vmat, Id, zero(vmat))
+        fX = prod([x^k for (x, k) in zip(gens(ambient_coordinate_ring(X)), sol)])
+        U = PrincipalOpenSubset(X, OO(X)(fX))
+        
+        AY = transpose(hilbert_basis(Y))
+        Id = identity_matrix(ZZ, ncols(AY))
+        sol = solve_mixed(AY, -vmat, Id, zero(vmat))
+        fY = prod([x^(k>0 ? 1 : 0) for (x, k) in zip(gens(ambient_coordinate_ring(Y)), sol)])
```
simonbrandhorst, line 78: why are `fY` and `fX` computed differently? I would have expected it to be symmetric.

simonbrandhorst (changes requested):

experimental/Schemes/ToricSchemes/NormalToricSchemes/Attributes.jl
HechtiDerLachs: I am in theory. However, this part of the code had been worked out jointly with @lkastner in Blaubeuren and unfortunately I am still not in the position to start walking on my own two feet here. So I prefer not to change it right now.

experimental/Schemes/ToricSchemes/NormalToricSchemes/Attributes.jl
HechtiDerLachs: It probably isn't symmetric because the joint faces of the two cones and their lineality spaces have a relative orientation with the two cones. Then we start with the one for `X` choosing such orientations, so that no negative signs can occur there. For `Y` it seems that we can run into problems, then. But this is just a gut-feeling right now. I would have to look through it again. 

The thing is: It's the same code as before and it has already been reviewed once. So I would just keep it as is for now.

simonbrandhorst (approved):

HereAround (approved):
Thank you @HechtiDerLachs !

HechtiDerLachs:
We're waiting for proceedings in https://github.com/oscar-system/Oscar.jl/pull/2163. Hopefully we do not cross the time limits for 1.6 any more. @HereAround is on it.

HereAround (commented):
I see that you are adding new exports into `src/Exports.jl`. However, this here is experimental code, and so the interface, exports etc. will likely change in the future. To facilitate those changes, I would suggest to keep the exports in `experimental/Schemes/ToricSchemes` until we migrate this code into `OSCAR/src`.

HereAround:
> We're waiting for proceedings in #2163. Hopefully we do not cross the time limits for 1.6 any more. @HereAround is on it.

Completed.

PR #2107, opened 2023-03-22
Abstract types for varieties

simonbrandhorst:
We fix our terminology.

In a nutshell:
Affine[/Projective]AlgebraicSet :=  affine [projective] scheme which is geometrically reduced and of finite type over a field
Affine[/Projective / Covered]Variety:= (affine/projective/separated) geometrically integral scheme of finite type over a field 

Curves, Surfaces, Threefolds and ToricVarieties are varieties. 

The plane curves should be integrated into this framework.
There should also be enough room for elliptic curves. @fieker 

@afkafkafk13, @wdecker 
Once we migrate AbsProjectiveScheme out of experimental more will follow.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/Varieties/Types.jl
+++ b/src/AlgebraicGeometry/Schemes/Varieties/Types.jl
@@ -1,0 +1,78 @@
+################################################################################
+#
+# Abstract types for varieties
+#
+################################################################################
+
+@doc Markdown.doc"""
+    AbsAffineVariety <: AbsSpec{<:Field}
+
+An affine, geometrically integral scheme of finite type over a field.
+"""
+abstract type AbsAffineVariety{BaseField<:Field, RingType} <:AbsSpec{BaseField, RingType}# where {BaseField <:Field, RingType}
+end
+
+@doc Markdown.doc"""
+    AbsProjectiveVariety <: AbsProjectiveScheme{<:Field}
+
+A projective variety over a field.
+
+That is a projective, geometrically integral scheme over a field.
+"""
+abstract type AbsProjectiveVariety{BaseField<:Field, GradedRingType<:Ring} <: AbsProjectiveScheme{BaseField, GradedRingType} end
+
+@doc Markdown.doc"""
+    AbsCoveredVariety <: Scheme{<:Field}
+
+A separated, geometrically integral scheme of finite type over a field.
+
+Note that we allow reducible varieties.
+"""
+abstract type AbsCoveredVariety{BaseField<:Field} <: AbsCoveredScheme{BaseField} end
+
+################################################################################
+#
+# Abstract types for curves
+#
+################################################################################
+
+@doc Markdown.doc"""
+    AbsAffineCurve <: AbsAffineVariety
+
+A curve in affine space.
+
+An affine curve is an affine variety of dimension one.
+"""
+abstract type AbsAffineCurve{BaseField<:Field, RingType<:Ring} <: AbsAffineVariety{BaseField, RingType} end
+
+@doc Markdown.doc"""
+    AbsProjectiveCurve <: AbsProjectiveVariety
+
+A projective curve embedded in an ambient projective space.
+"""
+abstract type AbsProjectiveCurve{BaseField<:Field, RingType<:Ring} <: AbsProjectiveVariety{BaseField, RingType} end
+
+
+@doc Markdown.doc"""
+    AbsCoveredCurve
+
+A curve represented in terms of a covering.
+
+A curve is a geometrically integral scheme of finite type over a field which is
+of dimension one.
+"""
+abstract type AbsCoveredCurve{BaseField} <: Scheme{BaseField} end
+
+#=
+################################################################################
+# Plane curves
+################################################################################
+@doc Markdown.doc"""
+    ProjectivePlaneCurve
+
+A plane projective curve embedded in an ambient projective space.
+"""
+mutable struct PlaneProjectiveCurve{BaseField<:Field, GradedRingType<:Ring} <: AbsProjectiveCurve{BaseField, GradedRingType}
+  X::ProjectiveScheme{BaseField, GradedRingType}
+end
+=#
--- a/src/AlgebraicGeometry/Schemes/Varieties/Types.jl
+++ b/src/AlgebraicGeometry/Schemes/Varieties/Types.jl
@@ -1,3 +1,19 @@
+@doc Markdown.doc"""
+    AbsAffineAlgebraicSet <: AbsSpec
+
+An affine, geometrically reduced scheme of finite type over a field.
+"""
+abstract type AbsAffineAlgebraicSet{BaseField<:Field, RingType} <:AbsSpec{BaseField, RingType}# where {BaseField <:Field, RingType}
+end
+
+@doc Markdown.doc"""
+    AbsProjectiveAlgebraicSet <: AbsProjectiveScheme
+
+An affine, geometrically reduced scheme of finite type over a field.
+"""
+abstract type AbsProjectiveAlgebraicSet{BaseField<:Field, RingType} <:AbsProjectiveScheme{BaseField, RingType}# where {BaseField <:Field, RingType}
+end
+
 ################################################################################
 #
 # Abstract types for varieties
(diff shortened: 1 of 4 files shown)


afkafkafk13:
@simonbrandhorst : I definitely agree with fixing this terminology -- and the type definitions do precisely this. Do you want a review from me?

simonbrandhorst:
Yes please.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
```diff
@@ -0,0 +1,94 @@
...
+"""
+abstract type AbsProjectiveVariety{BaseField<:Field, GradedRingType<:Ring} <: AbsProjectiveAlgebraicSet{BaseField, GradedRingType} end
+
+@doc Markdown.doc"""
+    AbsCoveredVariety <: Scheme
+
+A separated, geometrically integral scheme of finite type over a field.
+
+Note that we allow reducible varieties.
+"""
```
wdecker: What is meant here: The variety is (geometrically) integral, but may be reducible?

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
```diff
@@ -0,0 +1,94 @@
...
+
+An affine, geometrically reduced scheme of finite type over a field.
+"""
+abstract type AbsAffineAlgebraicSet{BaseField<:Field, RingType} <:AbsSpec{BaseField, RingType}# where {BaseField <:Field, RingType}
+end
+
+@doc Markdown.doc"""
+    AbsProjectiveAlgebraicSet <: AbsProjectiveScheme
+
+An affine, geometrically reduced scheme of finite type over a field.
```
afkafkafk13: Copy&Paste error: A **projective**, geometrically...

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
```diff
@@ -0,0 +1,94 @@
...
+That is a projective, geometrically integral scheme over a field.
+"""
+abstract type AbsProjectiveVariety{BaseField<:Field, GradedRingType<:Ring} <: AbsProjectiveAlgebraicSet{BaseField, GradedRingType} end
+
+@doc Markdown.doc"""
+    AbsCoveredVariety <: Scheme
+
+A separated, geometrically integral scheme of finite type over a field.
+
+Note that we allow reducible varieties.
```
afkafkafk13: ??? geometrically integral is equivalent to geometrically reduced and geometrically irreducible???
What did you really want to state with this comment?

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
```diff
@@ -0,0 +1,94 @@
...
+abstract type AbsProjectiveCurve{BaseField<:Field, RingType<:Ring} <: AbsProjectiveVariety{BaseField, RingType} end
+
+
+@doc Markdown.doc"""
+    AbsCoveredCurve
+
+A curve represented in terms of a covering.
+
+A curve is a geometrically integral scheme of finite type over a field which is
+of dimension one.
```
afkafkafk13: This sentence is somehow hard to read. Maybe
"A curve is a geometrically integral scheme of dimension 1 which is of finite type over a field"

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
```diff
@@ -0,0 +1,94 @@
...
+#=
+################################################################################
+# Plane curves
+################################################################################
+@doc Markdown.doc"""
+    ProjectivePlaneCurve
+
+A plane projective curve embedded in an ambient projective space.
+"""
+mutable struct PlaneProjectiveCurve{BaseField<:Field, GradedRingType<:Ring} <: AbsProjectiveCurve{BaseField, GradedRingType}
```
afkafkafk13: Mismatch of word order in the typename.

afkafkafk13 (commented):
I left you some comments. 

Only one of them is critical (geometrically integral), the others are of the quality of typos.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
```diff
@@ -0,0 +1,94 @@
+@doc Markdown.doc"""
+    AbsAffineAlgebraicSet <: AbsSpec
```
fingolfin, line 2: Per the naming conventions outlined in #1981 (not (yet) in the dev docs), `Abs` should always stand for "Absolute", and `Abstract` should not be abbreviated.

I don't mean this as a blocker for this PR, but we need to resolve this and either change this code (could be in a future PR), or adjust that rule.

CC @fieker

wdecker:
One question is, what does it mean on the practical side, that we work with these types. E.g., I might start with a (complicated) construction yielding some scheme which may or may not be geometrically integral. So then I better do not use the restricted type for the result even if I expect the result to be geometrically integral. So what are the typical situations for using these types? And do we then leave it to the user to make sure that his constructed objects have the desired properties?

afkafkafk13:
From my point of view, these types are for non-specialists. Algebraic geometers will know their way around schemes and not be scared by them. But beginners and users of algebraic geometry (as opposed to algebraic geometers) will need some more basic objects to use for their rather restricted purposes. As soon as they outgrow it, there are always schemes.

For these more restricted types, we should provide a optional argument which allows switching tests of the properties on and off. What the argument should default to, is then a question to be discussed.

simonbrandhorst:
The properties required are always tested. The constructors will have a `check` keyword. 
If you disable the checks, then you compute at your own risk. For instance if you ask for a function field, 
this does not make sense if the variety is not integral. And if it is not geometrically integral then it does not make sense to base change it naively to the algebraic closure.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
simonbrandhorst: We have tons of abstract types. The names will grow very long.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
simonbrandhorst: Thanks, it is a leftover of an older version.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
simonbrandhorst: leftover from an old version. Thanks for catching.

simonbrandhorst: I deleted the "Note..." .

afkafkafk13:
You are right: check by default and allow bypass of the check for the experienced user.

afkafkafk13 (approved):

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
simonbrandhorst: So some abbreviation would be desirable for me. But we should rename this consistently in the schemes folder and not mix `Abstract` and `Abs` in the same ecosystem or even in the same file.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
afkafkafk13, line 1: one nitpick: 
'Varieties' does not belong on the same level as Objects and Morphisms. I understand, why you put them below Schemes (as they belong to the ecosystem), but in @HereAround 's philosophy they would end up either as neighbouring directory of Schemes or as simple separate file in Schemes/Objects.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
simonbrandhorst: > From my point of view, these types are for non-specialists. Algebraic geometers will know their way around schemes and not be scared by them. But beginners and users of algebraic geometry (as opposed to algebraic geometers) will need some more basic objects to use for their rather restricted purposes. As soon as they outgrow it, there are always schemes.
> 
> For these more restricted types, we should provide a optional argument which allows switching tests of the properties on and off. What the argument should default to, is then a question to be discussed.

In some cases I do not want to work with schemes myself and stay in safe world of varieties,
e.g. when I am considering function fields, rational equivalence and Weil divisors. 
Having the varieties **geometrically** integral assures that one can interpret the results more easily after a base change to an algebraically closed field. 
A question for @wdecker we have absolute primary decomposition over `QQ`. Therefore we can check if a variety is geometrically integral over `QQ`. Can we also do this over a finite field? 
Say I am integral but not geometrically integral. Is there a way to detect this (without computing the absolute primary decomposition)? In Oscar? For geometrically reduced this is no problem because (reduced + over a perfect field) => geometrically reduced.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
simonbrandhorst: Since they rely on schemes they have to be included only after the schemes. 
This was the easiest way to do it. But feel free to change it if you care enough.

afkafkafk13: No, I won't change this. I can live with it. I am looking for stuff by git grep anyway.

src/AlgebraicGeometry/Schemes/Varieties/Types.jl
wdecker: We currently can check this ONLY over QQ.

PR #2112, opened 2023-03-23
Implementing facet_indices(vertices(P)) with P polytope

### LINKED ISSUE #2089: Polyhedron: add `facet_indices(vertices(P))`
This should work similar to `vertex_indices(facets(P))`.

Both should be added to the documentation.

cc: @micjoswig

marabelotti:
Draft for issue #2089


fingolfin:
Welcome @marabelotti and thanks for the PR. If this is still a draft, perhaps consider marking it as such. You can read [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request) how to do that.

Perhaps you could also edit the title of this PR to be a bit more meaningful and indicate what it is about (self-contained, without invoking the issue number)?

Thanks again!

lkastner:
This looks good, thank you. I have some requests though:
- `facet_indices` also makes sense in `facet_indices(rays(Cone))` right? It would be nice if that worked too. Are there maybe other things missing too for symmetry reasons?
- Just to confirm: This does the right thing also if there are rays or lineality?
- There are a bunch of these functions that kind of get autogenerated in this way. Unfortunately they are not documented, as @benlorenz mentioned. Currently the only solution to have them properly documented I see is to add a section in the documentation with examples. Otherwise one would have to undo these loops. What do you think?

src/PolyhedralGeometry/iterators.jl
```diff
@@ -252,7 +252,7 @@ Base.size(iter::SubObjectIterator) = (iter.n,)
 ################################################################################
 
 # Incidence matrices
-for (sym, name) in (("ray_indices", "Incidence Matrix resp. rays"), ("vertex_indices", "Incidence Matrix resp. vertices"), ("vertex_and_ray_indices", "Incidence Matrix resp. vertices and rays"))
+for (sym, name) in (("facet_indices", "Incidence matrix resp. facets"), ("ray_indices", "Incidence Matrix resp. rays"), ("vertex_indices", "Incidence Matrix resp. vertices"), ("vertex_and_ray_indices", "Incidence Matrix resp. vertices and rays"))
     M = Symbol(sym)
     _M = Symbol(string("_", sym))
```
fingolfin, line 257: Since @lkastner asked about documenting these auto-generated methods: you can in principle assign docstrings to these as usual (the docstrings of course have to follow a fix pattern, with `$sym` and `$name` interpolated in as necessary). We do that

micjoswig:
>     * `facet_indices` also makes sense in `facet_indices(rays(Cone))` right? It would be nice if that worked too. Are there maybe other things missing too for symmetry reasons?
> 
>     * Just to confirm: This does the right thing also if there are rays or lineality?

@marabelotti please take care of these

micjoswig (approved):

PR #2113, opened 2023-03-23
Add serialization for universal polynomials

SoongNoonien:
Working with the universal polynomials I noticed that they can't be serialized. This should be fixed now.

### DIFF
--- a/src/Serialization/Rings.jl
+++ b/src/Serialization/Rings.jl
@@ -50,6 +50,8 @@ end
 
 @registerSerializationType(MPolyRing)
 
+@registerSerializationType(UniversalPolyRing)
+
 @registerSerializationType(QQMPolyRing)
 @registerSerializationType(QQPolyRing)
 @registerSerializationType(ZZMPolyRing)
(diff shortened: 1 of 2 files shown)


src/Serialization/Rings.jl
```diff
@@ -60,35 +62,40 @@ end
...
     end
 
     return polynomial_ring(base_ring, symbols, cached=false)[1]
 end
 
 ################################################################################
-# Multivariate Polynomials
+# Multivariate and Universal Polynomials
 @registerSerializationType(QQMPolyRingElem)
 @registerSerializationType(ZZMPolyRingElem)
```
fingolfin, line 92: This is *not a review of this PR* but rather a separate question for @antonydellavecchia : I am actually surprised that we list `QQMPolyRingElem`, `ZZMPolyRingElem`, etc. separately here. Shouldn't these all be serialized as `MPolyRingElem`, with the `parent` and or `base_ring` determining the full type?

fingolfin (approved):
Thank you! This looks good to me, but it would be good if @antonydellavecchia had a look, too, before we merge this

src/Serialization/Rings.jl
antonydellavecchia: it does seem like these are no longer needed, I'll make a seperate pull requests and get rid of some unnecessary registration

antonydellavecchia:
also looks good to me

PR #2114, opened 2023-03-23
Towards Migrating projective schemes out of experimental

HechtiDerLachs:
In particular, we can finally create quotient rings of polynomial rings over arbitrary base rings (thanks to the changes entailed by #1876). That allows for a significant maturation of the projective schemes.

### DIFF
--- a/src/Rings/MPolyMap/flattenings.jl
+++ b/src/Rings/MPolyMap/flattenings.jl
@@ -8,8 +8,8 @@
 # To the outside, this is primarily the map identifying S with its 
 # flattened version.
 ########################################################################
-mutable struct RingFlattening{TowerRingType<:MPolyRing, FlatRingType<:Ring, 
-                              CoeffRingType<:Ring
+mutable struct RingFlattening{TowerRingType<:Union{MPolyRing, MPolyQuoRing}, 
+                              FlatRingType<:Ring, CoeffRingType<:Ring
                              } <: Hecke.Map{TowerRingType, FlatRingType, 
                                             SetMap, RingFlattening
                                            }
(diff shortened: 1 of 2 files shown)


simonbrandhorst:
Is this ready for review?

HechtiDerLachs:
> Is this ready for review?

As a first part, yes. But we opened the pandora's box here. This will become a big PR, not only, because stuff was copy-pasted. I'm still working on it and I'm far from half way through.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -121,6 +121,132 @@ mutable struct RingFlattening{TowerRingType<:MPolyRing, FlatRingType<:Ring,
...
+      S::MPolyQuoRing{RingElemType}
+    ) where {RingElemType <: MPolyRingElem{<:MPolyQuoLocRingElem}}
+    P = base_ring(S)::MPolyRing
+    Q = base_ring(P)::MPolyQuoLocRing
+    R = base_ring(Q)::MPolyRing
+    L = localized_ring(Q)::MPolyLocRing
+    A = underlying_quotient(Q)::MPolyQuoRing
+    U = inverted_set(Q)::AbsMultSet
+    I = modulus(Q)::MPolyLocalizedIdeal
+    kk = coefficient_ring(R)::Field
```
simonbrandhorst: I do not really understand what happens here. Also `A`, `U` and  `I` are not used anywhere. This looks fishy. Please check.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -121,6 +121,132 @@ mutable struct RingFlattening{TowerRingType<:MPolyRing, FlatRingType<:Ring,
...
+  end
+
+  function RingFlattening(
+      S::MPolyQuoRing{RingElemType}
+    ) where {RingElemType <: MPolyRingElem{<:MPolyLocRingElem}}
+    P = base_ring(S)::MPolyRing
+    L = base_ring(P)::MPolyLocRing
+    R = base_ring(L)::MPolyRing
+    U = inverted_set(L)::AbsMultSet
+    kk = coefficient_ring(R)::Field
```
simonbrandhorst: Are they used anywhere?

src/Rings/MPolyMap/flattenings.jl
simonbrandhorst: It would be nice to write a small comment on what the objects are. It is hard for me to understand this. 
Like ` (U^-1QQ[x])[y] / J -> U^{-1}QQ[x,y]/J`

fingolfin:
I am confused by the disparity between the description of this PR and its content: it says it moves something "out of experimental", yet it does not touch "experimental".

Which is fine by me, as the action described in the title also seems problematic to me: having one PR move something out of experimental and at the same time modify it besides trivial changes necessary to accommodate the move sounds to me as something asking for trouble: "it was working before we moved it, so something in the move broke it" is the kind of problem that's really no fun to deal with. So once you actually move things out, I'd recommend to separate "moving" from "modifying".

src/Rings/MPolyMap/flattenings.jl
HechtiDerLachs: What happened here: I first gathered all information I thought would be useful to untangle things. Then I implemented 2-3 different versions of how to achieve it and in the end I didn't use all the data from the beginning. Cleaned it up now.

HechtiDerLachs:
From my side, this is ready. As agreed with @simonbrandhorst , the actual migration will then happen in different PRs.

simonbrandhorst:
It works if it works.

fingolfin (approved):

PR #2121, opened 2023-03-24
Fix promotion for quotient rings

### LINKED ISSUE #2119: Promote rules for quotient rings seem buggy
I'm having issues with the following situation:
```julia
R, (x,y) = QQ["x", "y"]
I = ideal(R, x)
Q, _ = quo(R, I)
P, (u, v) = Q["u", "v"]
J = ideal(P, u)
A, _ = quo(P, J)
x*u # is ok
Q(x)*u # is ok
Q(x)*A(u) # fails
A(x)*u # is ok
```
The multiplication is not carried out because of failure of promotion in the marked cases. I tried fiddling around with the promote rules myself a little bit, but failed. Can somebody help me with this? E.g. by providing some hotfix? Or is the problem on my side and due to some basic misunderstanding of things?

Once this is running, I would like to take this even further also involving graded rings and their quotients. But one thing at a time.

thofma:
Closes #2119

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -507,13 +507,13 @@ function check_parent(a::MPolyQuoRingElem, b::MPolyQuoRingElem)
   return true
 end
 
-+(a::MPolyQuoRingElem, b::MPolyQuoRingElem) = check_parent(a, b) && MPolyQuoRingElem(a.f+b.f, a.P)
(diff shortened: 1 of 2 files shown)


HechtiDerLachs (approved):
This looks good and in fact similar to some things I already tried. Thanks for resolving this. Unfortunately, I am still facing problems with graded rings. But this will eventually become another issue as I need more time to produce a minimal example.

PR #2122, opened 2023-03-24
caching for properties of ideals

simonbrandhorst:
We cache for instance is an ideal is prime. But also the radical idea.
@fingolfin  how can I make `@attr function radical(I::MPolyIdeal)` type stable?

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -572,7 +572,7 @@ julia> minpoly(a)
 x^2 + 1
 ```
 """
-@attr function absolute_primary_decomposition(I::T) where {T <: MPolyIdeal{<:MPolyRingElem{QQFieldElem}}}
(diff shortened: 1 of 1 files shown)


thofma:
`@atttr T radical(I::T) where {T <: MPolyIdeal}` should work

simonbrandhorst:
Thanks works.

src/Rings/mpoly-ideals.jl
```diff
@@ -569,7 +572,7 @@ julia> minpoly(a)
 x^2 + 1
 ```
 """
-@attr function absolute_primary_decomposition(I::MPolyIdeal{<:MPolyRingElem{QQFieldElem}})
+@attr function absolute_primary_decomposition(I::T) where {T <: MPolyIdeal{<:MPolyRingElem{QQFieldElem}}}
```
thofma: `T` not used yet.

simonbrandhorst: Removed the `T`. But it is not type stable anyways.

wdecker (approved):

Issue #161, opened 2020-09-17
GAP groups: bring GAP groups and Hecke.jl groups (`GrpAbFinGen`)  together (e.g. providing homomorphisms, aligning names)


fingolfin:

fingolfin:
@ThomasBreuer : so @fieker and me discussed this a bit tomorrow. Here's my current idea (which we can still revise): we add a new function `isomorphism(T, G)` (name could clearly be improved). It computes an isomorphism from the group `G` into an isomorphic group `H` of type `T`. So you might call it as `isomorphism(GrpAbFinGen, G)` where `G` is any abelian GAP group; or conversely e.g. `isomorphism(PermGroup, G)`.

So this also replaces `isomorphic_perm_group`, `isomorphic_fp_group`, `isomorphic_pc_group`. We would then `@deprecate` those.

The isomorphic group then can be obtained via `codomain(iso)`. But in addition, we should allow writing `T(G)` in accordance with the usual Julia conventions; so e.g. `GrpAbFinGen(someAbelianPermGroup)` as shorthand for `codomain(isomorphism(someAbelianPermGroup))` (though of course more efficient implementations may be supplied).

fingolfin:
Some more thoughts on how things map:
- `Hecke.id(G)` is `Oscar.one(G)`
- `Hecke.conjugancy_classes` is a typo (PR submitted)
- `Hecke.psylow_subgroup` is `Oscar.sylow_subgroup`
  - and the one in Oscar should accept an `fmpz` (or ideally `Union{fmpz,Base.Integer}`) as the prime
- `Hecke.commutator_subgroup` is `Oscar.derived_subgroup` (we probably should accept both names?)
- `derived_series`
- ...

ThomasBreuer:
Concerning `Union{fmpz,Base.Integer}`, this occurs already frequently; should we define a name for it? (`OscarInteger`?)

Concerning `commutator_subgroup`, I had added a comment to `src/Groups/GAPGroups.jl`:

    In the literature, the name commutator subgroup is often used as a synonym of derived subgroup.
    GAP defines `CommutatorSubgroup( G, H )` for arbitrary groups `G`, `H` (in the same family);
    thus the result is in general not contained in `G` or `H`,  and we do not know into which group the result should be embedded.
    Is this function useful at all? (The name `Commutator*Subgroup*` is irritating, isn't it?)

fingolfin:
[...]

PR #2123, opened 2023-03-24
support `GrpAbFinGen` groups in group functions

ThomasBreuer:
- added "group function" methods for `GrpAbFinGen` groups (I am not sure how much of this should better be moved to Hecke)

- changed `GroupConjClass` to be an abstract type, defined `GAPGroupConjClass` and `GrpAbFinGenConjClass`

- renamed `radical_subgroup` to `solvable_radical` (the name has meanwhile been corrected in GAP)

- support prescribed type of integer for the return value of `is_pgroup_with_prime`, `length(::GroupConjClass)`, `number_conjugacy_classes`

- fixed `hall_subgroup_reps`

- added `images` to compute the image of a subgroup under a homomorphism

- added some doctests, improved docstrings

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -542,7 +542,7 @@ function Base.rand(C::GroupConjClass{S,T}) where S where T<:GAPGroupElem
    return Base.rand(Random.GLOBAL_RNG, C)
 end
 
-function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S where T<:GAPGroupElem
+function Base.rand(rng::Random.AbstractRNG, C::GAPGroupConjClass{S,T}) where S where T<:GAPGroupElem
    return group_element(C.X, GAP.Globals.Random(GAP.wrap_rng(rng), C.CC)::GapObj)
 end
 
@@ -607,7 +607,7 @@ end
 Return the subgroup conjugacy class `cc` of `H` in `G`, where `H` = `representative`(`cc`).
 """
 function conjugacy_class(G::T, g::T) where T<:GAPGroup
-   return GroupConjClass(G, g, GAP.Globals.ConjugacyClassSubgroups(G.X,g.X)::GapObj)
+   return GAPGroupConjClass(G, g, GAP.Globals.ConjugacyClassSubgroups(G.X,g.X)::GapObj)
 end
 
 function Base.rand(C::GroupConjClass{S,T}) where S where T<:GAPGroup
@@ -629,7 +629,7 @@ julia> G = symmetric_group(3)
 Sym( [ 1 .. 3 ] )
 
 julia> conjugacy_classes_subgroups(G)
-4-element Vector{GroupConjClass{PermGroup, PermGroup}}:
+4-element Vector{GAPGroupConjClass{PermGroup, PermGroup}}:
  Group(()) ^ Sym( [ 1 .. 3 ] )
  Group([ (2,3) ]) ^ Sym( [ 1 .. 3 ] )
  Group([ (1,2,3) ]) ^ Sym( [ 1 .. 3 ] )
@@ -684,7 +684,7 @@ Return the vector of all conjugacy classes of maximal subgroups of G.
 julia> G = symmetric_group(3);
 
 julia> conjugacy_classes_maximal_subgroups(G)
-2-element Vector{GroupConjClass{PermGroup, PermGroup}}:
+2-element Vector{GAPGroupConjClass{PermGroup, PermGroup}}:
  Group([ (1,2,3) ]) ^ Sym( [ 1 .. 3 ] )
  Group([ (2,3) ]) ^ Sym( [ 1 .. 3 ] )
 
@@ -692,7 +692,7 @@ julia> conjugacy_classes_maximal_subgroups(G)
 """
 function conjugacy_classes_maximal_subgroups(G::GAPGroup)
   L = Vector{GapObj}(GAP.Globals.ConjugacyClassesMaximalSubgroups(G.X)::GapObj)
-  return [GroupConjClass(G, _as_subgroup_bare(G, GAP.Globals.Representative(cc)), cc) for cc in L]
+  return [GAPGroupConjClass(G, _as_subgroup_bare(G, GAP.Globals.Representative(cc)), cc) for cc in L]
 end
 
 """
@@ -935,7 +935,7 @@ end
 
 
 # START iterator
-Base.IteratorSize(::Type{<:GroupConjClass}) = Base.SizeUnknown()
+Base.IteratorSize(::Type{<:GAPGroupConjClass}) = Base.SizeUnknown()
 
 Base.iterate(cc::GAPGroupConjClass) = iterate(cc, GAPWrap.Iterator(cc.CC))
 
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1408,10 +1408,10 @@ julia> prime_of_pgroup(UInt16, quaternion_group(8))
 0x0002
 
 julia> prime_of_pgroup(symmetric_group(1))
-ERROR: only supported for non-trivial p-groups
+ERROR: ArgumentError: only supported for non-trivial p-groups
 
 julia> prime_of_pgroup(symmetric_group(3))
-ERROR: only supported for non-trivial p-groups
+ERROR: ArgumentError: only supported for non-trivial p-groups
 
 ```
 """
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1,5 +1,3 @@
-@alias is_finiteorder isfiniteorder
-
 # TODO: as soon as GAP packages like `polycyclic` or `rcwa` are loaded,
 # the custom group types and isos they define should be added to the arrays
 # _gap_group_types resp. _iso_function
@@ -58,21 +56,23 @@ function (G::GAPGroup)(x::BasicGAPGroupElem{T}) where T<:GAPGroup
 end
 
 """
-    isfinite(G::GAPGroup) -> Bool
+    is_finite(G::GAPGroup) -> Bool
 
 Return `true` if `G` is finite, and `false` otherwise.
 
 # Examples
 ```jldoctest
-julia> isfinite(symmetric_group(5))
+julia> is_finite(symmetric_group(5))
 true
 
-julia> isfinite(free_group(2))
+julia> is_finite(free_group(2))
 false
 
 ```
 """
-@gapattribute isfinite(G::GAPGroup) = GAP.Globals.IsFinite(G.X)::Bool
+@gapattribute is_finite(G::GAPGroup) = GAP.Globals.IsFinite(G.X)::Bool
+
+# Base.is_finite(G::PcGroup) = true
 
 """
     is_finiteorder(g::GAPGroupElem) -> Bool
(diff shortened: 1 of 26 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -994,13 +1046,17 @@ the intersection of all maximal subgroups of `G`.
...
 Return the solvable radical of `G`, i.e.,
 the largest solvable normal subgroup of `G`.
 """
-@gapattribute radical_subgroup(G::GAPGroup) = _as_subgroup(G, GAP.Globals.RadicalGroup(G.X))
-#T wrong name, already in GAP!
+@gapattribute solvable_radical(G::GAPGroup) = _as_subgroup(G, GAP.Globals.SolvableRadical(G.X))
+
+@deprecate radical_subgroup solvable_radical
+@deprecate has_radical_subgroup has_solvable_radical
+@deprecate set_radical_subgroup set_solvable_radical
```
thofma: I think there is a file where we collect them?

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+   else
+     return false
+   end
+end
+
+function representative_action(G::GrpAbFinGen, H::GrpAbFinGen, K::GrpAbFinGen)
+   if is_subgroup(H, K)[1] && is_subgroup(K, H)[1]
+     return true, zero(G)
+   else
+     return false, nothing
```
thofma, line 170:
```suggestion
     return false, zero(G)
```
is there harm in doing that?

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+solvable_radical(G::GrpAbFinGen) = (G, identity_map(G))
+
+
+################################################################################
+#
+# Sylow & Hall Subgroups
+#
+################################################################################
+
+#TODO: how to compute complements?
```
thofma, line 269: Complements are already in `Hecke`. Probably `has_complex(U, G)` or something like that.

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+function _prime_of_pgroup(G::GrpAbFinGen)
+  flag, p, e = is_prime_power_with_data(order(G))
+  if !flag
+    error("only supported for non-trivial p-groups")
+  end
+  return p
+end
+
+function prime_of_pgroup(::Type{T}, G::GrpAbFinGen) where T <: IntegerUnion
+  return ZZRingElem(_prime_of_pgroup(G))
```
thofma:
```suggestion
  return T(_prime_of_pgroup(G))
```

thofma (commented):
Looks good! I left a few remarks.

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+#
+# Sylow & Hall Subgroups
+#
+################################################################################
+
+#TODO: how to compute complements?
+# complement_class_reps(G::T, N::T) where T <: GrpAbFinGen
+# complement_system(G::GrpAbFinGen)
+
+function sylow_subgroup(G::GrpAbFinGen, p::IntegerUnion)
```
fingolfin, line 273: Note that there also is `function psylow_subgroup(G::GrpAbFinGen, p::IntegerUnion, to_lattice::Bool = true)`

thofma: Yes, we should consolidate this

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+################################################################################
+#
+# Specific Subgroups
+#
+################################################################################
+
+fitting_subgroup(G::GrpAbFinGen) = (G, identity_map(G))
+
+function frattini_subgroup(G::GrpAbFinGen)
+   is_finite(G) || throw(ArgumentError("G is not finite"))
```
fingolfin: Could also use the `@req` macro:
```suggestion
   @req is_finite(G) "G is not finite"
```

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+# x^n for GrpAbFinGenElem -> n*x
+
+function comm(x::GrpAbFinGenElem, y::GrpAbFinGenElem)
+  x.parent == y.parent || error("Elements must belong to the same group")
+  zero(parent(x))
+end
+
+has_gens(G::GrpAbFinGen) = true
+
+#TODO: improve?
```
fingolfin: Maybe `snf` can be used, even to get a minimal generating set, see http://www.thofma.com/Hecke.jl/stable/abelian/structural/#snf-Tuple{GrpAbFinGen}

src/Groups/GAPGroups.jl
```diff
@@ -102,7 +102,28 @@ positive integer `n` such that `x^n` is the identity of `G`.
...
+
+julia> order(g)
+6
+
+julia> order(gen(g, 1))
+3
+
+julia> g = free_group(1);
+
+julia> isfinite(g)
```
fingolfin: Meta-question: should we say `is_finite` here, to be more consistent with the rest? (I have no strong opinion on this, just noticed the contrast to the next input line)

src/Groups/GAPGroups.jl
fingolfin: Indeed, `src/Deprecations.jl`

We also have some in `src/Groups/sub.jl` (I think they may predate `src/Deprecations.jl`)

src/Groups/GAPGroups.jl
```diff
@@ -1098,8 +1154,7 @@ function hall_subgroup_reps(G::GAPGroup, P::AbstractVector{<:IntegerUnion})
    end
 end
 
-hall_subgroups_representatives(G::GAPGroup, P::AbstractVector{<:IntegerUnion}) = hall_subgroup_reps(G,P)
-# use @alias?
+@alias hall_subgroups_representatives hall_subgroup_reps
```
fingolfin: Should this be in `src/Aliases.jl` perhaps?

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+
+Base.eltype(::Type{GrpAbFinGenConjClass{T,S}}) where {T,S} = S
+
+Base.rand(C::GrpAbFinGenConjClass) = representative(C)
+
+Base.rand(rng::Random.AbstractRNG, C::GrpAbFinGenConjClass) = representative(C)
+
+number_conjugacy_classes(G::GrpAbFinGen) = ZZRingElem(order(G))
+
+number_conjugacy_classes(::Type{T}, G::GrpAbFinGen) where T <: IntegerUnion = T(order(G))
```
fingolfin:
```suggestion
number_conjugacy_classes(::Type{T}, G::GrpAbFinGen) where T <: IntegerUnion = order(T, G)
```

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+
+conjugacy_class(G::GrpAbFinGen, g::GrpAbFinGenElem) = GrpAbFinGenConjClass(G, g)
+
+Base.eltype(::Type{GrpAbFinGenConjClass{T,S}}) where {T,S} = S
+
+Base.rand(C::GrpAbFinGenConjClass) = representative(C)
+
+Base.rand(rng::Random.AbstractRNG, C::GrpAbFinGenConjClass) = representative(C)
+
+number_conjugacy_classes(G::GrpAbFinGen) = ZZRingElem(order(G))
```
fingolfin:
```suggestion
number_conjugacy_classes(G::GrpAbFinGen) =order(ZZRingElem, G)
```

src/Groups/GrpAb.jl
```diff
@@ -12,3 +12,348 @@ function restrict_codomain(f::GrpAbFinGenMap)
...
+# conjugacy_classes_maximal_subgroups(G::GrpAbFinGen)
+# maximal_subgroup_reps(G::GrpAbFinGen)
+# low_index_subgroup_reps(G::GrpAbFinGen, n::Int)
+
+conjugate_group(G::GrpAbFinGen, x::GrpAbFinGenElem) = G
+
+Base.:^(G::GrpAbFinGen, x::GrpAbFinGenElem) = G
+
+function is_conjugate(G::GrpAbFinGen, H::GrpAbFinGen, K::GrpAbFinGen)
+   if is_subgroup(H, K)[1] && is_subgroup(K, H)[1]
```
fingolfin: Should this also have a test whether `H` and `K` are actually subgroups of `G`?

Out of curiosity, does `H == K` also work? With all those magic embeddings, this is not clear to me.

src/Groups/GrpAb.jl
ThomasBreuer: Currently the `representative_action` documentation promises `nothing`, and the method for `GAPGroup` does this.
We can change the definition and the methods such that always the identity element of the group is returned instead of `nothing`.

src/Groups/GrpAb.jl
ThomasBreuer: What about supporting the name `sylow_subgroup` in Hecke?
(And I think the `to_lattice` hint should be added to all functions that call `sub` for their return values.)

thofma: `sylow_subgroup` sounds better. I will add it in Hecke.

src/Groups/GrpAb.jl
ThomasBreuer: No, `H == K` is not implied by `is_subgroup(H, K)[1] && is_subgroup(K, H)[1]`.

src/Groups/GrpAb.jl
ThomasBreuer: There is `has_complement`, which computes one complement if it exists. `complement_class_reps` would have to compute all complements.

thofma: @fieker I remember you played with complements. Do you by accident know how to "efficiently" compute all representatives?

src/Groups/GrpAb.jl
fingolfin: Yeah I guessed as much. Very confusing :-(

ThomasBreuer: For group theoretic questions, it is usual to construct subgroups in different ways and then ask whether they are equal. From this point of view, the setup for `GrpAbFinGen` is indeed strange.
Here is an example.
```
julia> g = abelian_group([2,3]);  # cyclic group of order 6

julia> s1, emb1 = sylow_subgroup(g, 3);  # the unique subgroup of order 3

julia> s2 = low_index_subgroup_reps(g, 2)[2]; # another way to get the subgroup of order 3

julia> flag2, emb2 = is_subgroup(s2, g); flag2
true

julia> order(s1) == order(s2)
true

julia> is_subgroup(s1, s2)[1] && is_subgroup(s2, s1)[1]  # we can check containment ("magically", w.r.t. the embeddings into g)
true

julia> s1 == s2  # the two groups are not regarded as equal
false
```
The point seems to be that `s1` and `s2` are regarded as independent *groups*, and that only `emb1` and `emb2` interpret them as *subgroups* of `g`. Perhaps the cleanest way to ask for the equality as subgroups of `g` would be something like `is_equal_image(emb1, emb2)`.

ThomasBreuer:
@thofma @fingolfin How shall we proceed with this pull request?
Eventually part of  it should be moved to Hecke, but are there reasons not to merge it into Oscar?

thofma:
Looks good to me. @fingolfin can merge if he is happy

fingolfin (approved):
Thank you!

fingolfin:
Can we close #161 now? Maybe @ThomasBreuer can have a look, and either close it, or perhaps summarize in a comment or a new issue what is still missing?

PR #2126, opened 2023-03-24
Change gitfan from polymake to Oscar types

jankoboehm:
This is the first round of changes to adapt from polymake types to Oscar types. Everything works except the last step to integrate fans with groups actions. Some int types should be updated still.


fingolfin:
The error I see in the logs is this:
```
Examples.GITFans, the example: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/GITFans-test.jl:1
  ArgumentError: Unrecognized argument type: Cone{QQFieldElem}.
  You need to convert to polymake compatible type first.
  Stacktrace:
    [1] convert_to_pm_type(T::Type)
      @ Polymake ~/.julia/packages/Polymake/A0Z5B/src/convert.jl:64
    [2] convert(#unused#::Type{Polymake.PolymakeType}, x::Cone{QQFieldElem})
      @ Polymake ~/.julia/packages/Polymake/A0Z5B/src/convert.jl:29
    [3] _broadcast_getindex_evalf
      @ ./broadcast.jl:670 [inlined]
```

jankoboehm:
The last step with the new group action on fans has to be still put together properly.

experimental/GITFans/GITFans.jl
```diff
@@ -225,15 +225,15 @@ and returns `true` if the two objects are considered as equal,
...
+function orbit(point, generators, action)
     orb = [point]
     # Note that in Julia (like in GAP),
     # the 'for' loop runs also over entries that get added
     # inside the loop.
     for b in orb
         for g in generators
             c = action(b, g)
-            if ! any(i->compare_func(i, c), orb)
+            if ! any(i-> i == c, orb)
```
fingolfin: Perhaps this could also be changed?
```suggestion
            if !(c in orb)
```

experimental/GITFans/GITFans.jl
```diff
@@ -225,15 +225,15 @@ and returns `true` if the two objects are considered as equal,
 and `false` otherwise.
 The elements of the returned array are in general not sorted.
 """
-function orbit(point, generators, action, compare_func)
+function orbit(point, generators, action)
```
fingolfin: Can't this perhaps be replaced by a call to our "standard" `orbit` function? Probably need the subgroup generated by those generators, but otherwise...?

experimental/GITFans/GITFans.jl
```diff
@@ -112,22 +112,22 @@ such that the matrix `S` has full rank and such that `I` is
...
                is_monomial_free(I, setdiff(1:nr_variables, i))
-                cone = Polymake.polytope.Cone(INPUT_RAYS = current_mat)
-                cone.FACETS
-                cone.RAYS
-                if ! any(j -> Polymake.polytope.equal_polyhedra(j, cone),
+                cone = positive_hull(current_mat)
+                facets(cone)
+                rays(cone)
+                if ! any(j -> j == cone,
                       collector_cones)
```
fingolfin, line 139: Maybe this would also work now in the much nicer new setup? It would arguably easier to understand, I think
```suggestion
                if !(cone in collector_cones)
```

experimental/GITFans/GITFans.jl
```diff
@@ -112,22 +112,22 @@ such that the matrix `S` has full rank and such that `I` is
...
             current_mat = Q[i,:]
             if rank(current_mat) == projected_dimension &&
                is_monomial_free(I, setdiff(1:nr_variables, i))
-                cone = Polymake.polytope.Cone(INPUT_RAYS = current_mat)
-                cone.FACETS
-                cone.RAYS
-                if ! any(j -> Polymake.polytope.equal_polyhedra(j, cone),
+                cone = positive_hull(current_mat)
+                facets(cone)
+                rays(cone)
```
fingolfin, line 137: Why are these two lines needed? For their side effects, I assume, but what are the relevant side effects? Or can they be removed?
```suggestion
```

experimental/GITFans/GITFans.jl
```diff
@@ -336,7 +335,7 @@ function compute_bit_list(orbits, point)
         current_orbit = orbits[current_orbit_nr]
         for current_cone_nr in 1:length(current_orbit)
             current_cone = current_orbit[current_cone_nr]
-            if Polymake.polytope.contains(current_cone, point)
+            if contains(current_cone, point)
```
fingolfin: Dunno if this is supported, but if it is, IMHO it would be easier to read
```suggestion
            if point in current_cone
```

experimental/GITFans/GITFans.jl
```diff
@@ -416,16 +423,21 @@ function fan_traversal(orbit_list, q_cone, perm_actions)
 
         # note that we run also over elements added inside the loop
         current_cone_list = cones_from_bitlist(orbit_list, current_hash)
-        intersected_cone = Polymake.polytope.intersection(current_cone_list...)
-        facets = intersected_cone.FACETS
+        intersected_cone = current_cone_list[1]
+        for i in current_cone_list
+            intersected_cone = intersect(intersected_cone, i)
+        end
+        # TODO extend intersect to more than two cones
```
fingolfin: I think thanks to @lkastner this could now be changed to this?
```suggestion
        intersected_cone = intersect(current_cone_list...)
```
BTW according to Codecov, this code is not executed as part of CI?

experimental/GITFans/GITFans.jl
```diff
@@ -386,28 +385,36 @@ function get_neighbor_hash(orbits, facet_point, inner_normal_vector)
...
-    fan_traversal(orbit_list, q_cone, perm_actions)
+    fan_traversal(orbit_list, q_cone::Cone, perm_actions)
 
 Return the pair `(hash_list, edges)` where `hash_list` is an array that
 encodes orbit representatives of the maximal cones of the GIT fan described
 by `orbit_list`, `q_cone`, and `perm_actions`,
 and `edges` encodes the `Set` of edges of the incidence graph of the orbits.
 """
-function fan_traversal(orbit_list, q_cone, perm_actions)
+function fan_traversal(orbit_list, q_cone::Cone, perm_actions)
```
fingolfin: I wonder why Codecov reports that this function is not being called? There seems to be an explicit call to it in `test/Experimental/GITFans-test.jl` ?!

fingolfin (commented):
Some remarks. Not sure if some of this is covered by "The last step with the new group action on fans has to be still put together properly", but I guess it doesn't hurt to mention these?

experimental/GITFans/GITFans.jl
lkastner: I'll add a comment there, same one I write here:
When comparing two cones it is fastest to solve a bunch of LP's to figure out whether the rays of one are contained in the other and vice versa. It would be even easier to just do some matrix multiplication if one already has the facets and rays. However to get the facets and rays is very hard compared to the LP's. But in this case we are comparing cones a lot, so the cost of the LP's outweigh the heavy lifting of computing rays and facets.

lkastner: I think in this case @benlorenz had something even faster in mind.

experimental/GITFans/GITFans.jl
lkastner: Yeah, that was the intention. It turned out to be harder than expected, since instead of being a `Vector{Cone{T}}`, the variable `current_cone_list` only holds a `Vector{Any}`. Nevertheless, it works now.

experimental/GITFans/GITFans.jl
fingolfin: Thank you for the explanation and the new comment!

experimental/GITFans/GITFans.jl
```diff
@@ -112,22 +112,30 @@ such that the matrix `S` has full rank and such that `I` is
 monomial-free (see [`is_monomial_free`](@ref)) w.r.t. the variables `x_i`
 for which the `i`-th row of `Q` is not contained in `S`.
 """
-function orbit_cones(I::Oscar.MPolyIdeal, Q::Matrix{Int}, G::PermGroup = symmetric_group(1))
+function orbit_cones(I::MPolyIdeal, Q::Matrix{Int}, G::PermGroup = symmetric_group(1))
     nr_variables, projected_dimension = size(Q)
 
-    collector_cones = []
+    collector_cones = Vector{Cone{QQFieldElem}}([])
```
fingolfin:
```suggestion
    collector_cones = Vector{Cone{QQFieldElem}}()
```
or
```suggestion
    collector_cones = Cone{QQFieldElem}[]
```

experimental/GITFans/GITFans.jl
```diff
@@ -416,19 +438,14 @@ function fan_traversal(orbit_list, q_cone, perm_actions)
...
+        ic_facets = -linear_inequality_matrix(facets(intersected_cone))
+        facet_points = [Vector{T}([relative_interior_point(f)...]) for f in faces(intersected_cone, dim(intersected_cone)-1)]
 
         neighbor_hashes = []
         for i in 1:length(facet_points)
-            if any(x->x==0, q_cone_facets_converted*facet_points[i])
-                continue
-            end
-            push!(neighbor_hashes, get_neighbor_hash(orbit_list, facet_points[i], facets[i, :]))
+          !any(f->contains(f, facet_points[i]), faces(q_cone, dim(q_cone)-1)) || continue
```
fingolfin, line 426: Could also write it like this:
```suggestion
          !any(contains(facet_points[i]), faces(q_cone, dim(q_cone)-1)) || continue
```
and if #2142 is resolved then could even write
```suggestion
          !any(contains(facet_points[i]), facets(Cone, q_cone)) || continue
```

experimental/GITFans/GITFans.jl
```diff
@@ -416,19 +438,14 @@ function fan_traversal(orbit_list, q_cone, perm_actions)
...
         current_cone_list = cones_from_bitlist(orbit_list, current_hash)
-        intersected_cone = Polymake.polytope.intersection(current_cone_list...)
-        facets = intersected_cone.FACETS
-        facets = convert(Matrix{Rational{BigInt}},facets)
-        facet_points = Vector{Rational{BigInt}}[
-                           Polymake.polytope.facet(intersected_cone, i-1).REL_INT_POINT
-                           for i in 1:size(facets, 1)]
+        intersected_cone = intersect(current_cone_list)
+        ic_facets = -linear_inequality_matrix(facets(intersected_cone))
+        facet_points = [Vector{T}([relative_interior_point(f)...]) for f in faces(intersected_cone, dim(intersected_cone)-1)]
```
fingolfin:
```suggestion
        facet_points = [T[relative_interior_point(f)...] for f in faces(intersected_cone, dim(intersected_cone)-1)]
```
if #2142 is resolved then could even write
```suggestion
        facet_points = [T[relative_interior_point(f)...] for f in facets(Cone,intersected_cone)]
```

experimental/GITFans/GITFans.jl
```diff
@@ -351,8 +365,8 @@ Return the array of all cones at `true` positions in the bitsets,
 where `cone_list` is an array of arrays of cones,
 and `bit_list_tuple` is an array of bitsets.
 """
-function cones_from_bitlist(cone_list, bit_list_tuple)
-    return_list = Any[]
+function cones_from_bitlist(cone_list::Vector{Vector{Cone{T}}}, bit_list_tuple) where T
+  return_list = Vector{Cone{T}}([])
```
fingolfin:
```suggestion
  return_list = Cone{T}[]
```

experimental/GITFans/GITFans.jl
```diff
@@ -362,7 +376,7 @@ function cones_from_bitlist(cone_list, bit_list_tuple)
 end
 
 
-hash_to_cone(orbit_list, hash) = Polymake.polytope.intersection(
+hash_to_cone(orbit_list, hash) = intersect(
     cones_from_bitlist(orbit_list, hash)...)
```
fingolfin:
```suggestion
hash_to_cone(orbit_list, hash) = intersect(cones_from_bitlist(orbit_list, hash)...)
```

experimental/GITFans/GITFans.jl
```diff
@@ -386,28 +400,36 @@ function get_neighbor_hash(orbits, facet_point, inner_normal_vector)
...
 by `orbit_list`, `q_cone`, and `perm_actions`,
 and `edges` encodes the `Set` of edges of the incidence graph of the orbits.
 """
-function fan_traversal(orbit_list, q_cone, perm_actions)
+function fan_traversal(orbit_list::Vector{Vector{Cone{T}}}, q_cone::Cone{T}, perm_actions) where T
     # the induced actions on each of the orbits
     generators_new_perm = rewrite_action_to_orbits(perm_actions)
 
-    q_cone_facets_converted = convert(Matrix{Rational{BigInt}}, q_cone.FACETS)
+    q_cone_facets_converted = matrix(QQ, q_cone.pm_cone.FACETS)
```
fingolfin: I think this is not used anymore (good!)
```suggestion
```

experimental/GITFans/GITFans.jl
lkastner: This requires more work than suggested in #2142:
```
julia> c = positive_hull([1 0; 0 1])
Polyhedral cone in ambient dimension 2

julia> fc = facets(Cone{QQFieldElem}, c)
2-element SubObjectIterator{Cone{QQFieldElem}}:
 Polyhedral cone in ambient dimension 2
 Polyhedral cone in ambient dimension 2

julia> [rays_modulo_lineality(f) for f in fc]
2-element Vector{NamedTuple{(:rays_modulo_lineality, :lineality_basis), Tuple{SubObjectIterator{RayVector{QQFieldElem}}, SubObjectIterator{RayVector{QQFieldElem}}}}}:
 (rays_modulo_lineality = [[1, 0]], lineality_basis = [[0, 1]])
 (rays_modulo_lineality = [[0, 1]], lineality_basis = [[1, 0]])
```
As you can see, these "facets" are just the linear halfspaces as a different type. I was just as confused as you and happy that I found a workaround for now. I am currently discussing with @benlorenz what to do.

experimental/GITFans/GITFans.jl
lkastner: Do you have a preference? I'd then change it consistently.

fingolfin: No strong preference. I slightly prefer `Cone{QQFieldElem}[]` because it is shorter, but perhaps it is also harder for someone who does not know this feature? I.e., `Vector{Int}()` is kinda obvious, `Int[]` maybe less so?

Maybe do a poll on Slack? 😛

experimental/GITFans/GITFans.jl
fingolfin: Thank you for looking into this :-)

experimental/GITFans/GITFans.jl
lkastner: I added that to our todolist, but I think it makes sense to do this in a different PR.

fingolfin:
There's a lot of improvements here. Any reason not to merge this right now? We can still make further improvements in a new PR afterwards, right?

@jankoboehm if you agree, please remove the draft status, please until tomorrow.

fingolfin (approved):

lkastner (approved):

experimental/GITFans/GITFans.jl
fingolfin: OK, I see:
```julia-repl
julia> q_cone = cone_from_inequalities([-1 0 0; 0 -1 0; 0 0 -1])

julia> f1 = faces(q_cone, dim(q_cone)-1);

julia> f2 = facets(Cone{QQFieldElem}, q_cone);

julia> [rays_modulo_lineality(f) for f in f1]
3-element Vector{NamedTuple{(:rays_modulo_lineality, :lineality_basis), Tuple{SubObjectIterator{RayVector{QQFieldElem}}, SubObjectIterator{RayVector{QQFieldElem}}}}}:
 (rays_modulo_lineality = [[0, 1, 0], [0, 0, 1]], lineality_basis = [])
 (rays_modulo_lineality = [[1, 0, 0], [0, 0, 1]], lineality_basis = [])
 (rays_modulo_lineality = [[1, 0, 0], [0, 1, 0]], lineality_basis = [])

julia> [rays_modulo_lineality(f) for f in f2]
3-element Vector{NamedTuple{(:rays_modulo_lineality, :lineality_basis), Tuple{SubObjectIterator{RayVector{QQFieldElem}}, SubObjectIterator{RayVector{QQFieldElem}}}}}:
 (rays_modulo_lineality = [[1, 0, 0]], lineality_basis = [[0, 1, 0], [0, 0, 1]])
 (rays_modulo_lineality = [[0, 1, 0]], lineality_basis = [[1, 0, 0], [0, 0, 1]])
 (rays_modulo_lineality = [[0, 0, 1]], lineality_basis = [[1, 0, 0], [0, 1, 0]])
```

So yeah, these are different. Thanks also for the explanation on gather.

PR #2127, opened 2023-03-24
Project.toml: allow Singular 0.18.x

fingolfin:
It is identical to 0.17 except some functions were removed (which we don't use)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -36,7 +36,7 @@ Polymake = "0.9.0"
 Preferences = "1"
 RandomExtensions = "0.4.3"
 RecipesBase = "1.2.1"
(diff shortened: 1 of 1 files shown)


thofma (approved):

fingolfin:
Since this just adds to the compat range, I don't think we have to wait for full CI to complete. I verified the syntax is correct (by loading the package locally). So I'll just go ahead and merge it as-is.

thofma:
> Since this just adds to the compat range, I don't think we have to wait for full CI to complete. I verified the syntax is correct (by loading the package locally). So I'll just go ahead and merge it as-is.

Famous last words

Issue #2099, opened 2023-03-22
is_zero is slow for quotient rings


HechtiDerLachs:
I was observing that `is_zero` is rather slow for quotient rings. For instance, try the following code:
```julia
R, (x,y,z) = QQ["x", "y", "z"]

I = ideal(R, [x^4 + y*z^7 - 35*x^3*z, x^29 - 15 + z- y*z^3])

Q, _ = quo(R, I)

a = x
b = Q(x)

@time for i in 1:1000;  iszero(a); end
@time for i in 1:1000;  iszero(b); end
```
I get `0.000083 seconds` vs. `0.116045 seconds`. Moreover, an analysis with `@profview` indicates that for every call to `is_zero`, a groebner basis for `I` is looked up in a dictionary. This lookup seems to first create an integer matrix for the given ordering and then these matrices are compared. In various other instances (which I can unfortunately not reproduce here) this has lead to significant time consumption in my computations (At least it showed up to consume much time in the `ProfileView` analysis). My general feeling is that this comes to bear when we need to check that many polynomials are zero modulo an ideal, for instance when dealing with matrices over quotient rings or polynomial rings over quotient rings. 

Any ideas on improvements for this?

thofma:
Can you quantify what "much time" means for the lookup? 99 percent? 50 percent? 10 percent?

HechtiDerLachs:
Often it looks close to 95% of the time consumed for the `is_zero` test. 

Edit: On my machine, this can even be observed for the above example when using `@profview`.

thofma:
I see various (easy) ways to improve it. I will open a PR soonish.

thofma:
Do you have a non-trivial example (but which does not run for hours)?

fingolfin:
A side remark for @HechtiDerLachs: I recommend using the `BenchmarkTools` package for doing such timings. So instead of applying `@time` to a loop, you can do this:
```julia-repl
julia> using BenchmarkTools

julia> @btime iszero(a);
  41.961 ns (0 allocations: 0 bytes)

julia> @btime iszero(b);
  40.708 μs (675 allocations: 20.06 KiB)
```
So that shows a factor 1000 in performance, and a lot of extra allocations.

BTW you can use the `@benchmark` macro to get a super nifty visualization of the distribution of the timings for several runs.

Now, I thought that surely `a in I` (surely equivalent to `is_zero(b)`) might be faster, but to my surprise it is even worse:
```julia-repl
[...]

PR #2133, opened 2023-03-24
Ideal membership

ederc:
This is a part of a fix for #2099: `ideal_membership` now uses the first cached Gröbner basis for an ideal, otherwise it first computes a Gröbner basis w.r.t. the given monomial ordering.

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -910,7 +910,7 @@ end
     ideal_membership(f::T, I::MPolyIdeal{T}; ordering::MonomialOrdering = default_ordering(base_ring(I))) where T
 
 Return `true` if `f` is contained in `I`, `false` otherwise. Uses a cached Gröbner basis for `I` or, if no such cached
-Gröbner basis exists, computes first a Gröbner basis for `I` w.r.t. the monomial ordering `ordering.
+Gröbner basis exists, computes first a Gröbner basis for `I` w.r.t. the monomial ordering `ordering`.
 Alternatively, use `f in I`.
 
 # Examples
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -907,9 +907,11 @@ end
 #######################################################
 
 @doc Markdown.doc"""
-    ideal_membership(f::T, I::MPolyIdeal{T}) where T
+    ideal_membership(f::T, I::MPolyIdeal{T}; ordering::MonomialOrdering = default_ordering(base_ring(I))) where T
 
-Return `true` if `f` is contained in `I`, `false` otherwise. Alternatively, use `f in I`.
+Return `true` if `f` is contained in `I`, `false` otherwise. Uses a cached Gröbner basis for `I` or, if no such cached
+Gröbner basis exists, computes first a Gröbner basis for `I` w.r.t. the monomial ordering `ordering.
+Alternatively, use `f in I`.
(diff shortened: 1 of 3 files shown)


src/Rings/mpoly-ideals.jl
```diff
@@ -933,7 +935,11 @@ false
 ```
 """
 function ideal_membership(f::T, I::MPolyIdeal{T}; ordering::MonomialOrdering = default_ordering(base_ring(I))) where T
-  GI = standard_basis(I, ordering=ordering, complete_reduction=false)
+  if isempty(I.gb)
+    GI = standard_basis(I, ordering=ordering, complete_reduction=false)
+  else
+    GI = first(I.gb)[2]
```
fingolfin, line 941: Ah but this now ignores `ordering`? Is this temporary?

Moving forward, we discussed a bit last week, and it seems that a lot of the problems could be addressed if only `MonomialOrdering` cached its hash value and perhaps a few other things, *and* if the the default ordering was actually stored in the ring, so that calling `default_ordering(base_ring(I))` twice returns an identical ordering. Does that sound sensible?

afkafkafk13: @fingolfin has a point here -- though not the one he intended to make:
If we were only considering global orderings, we would be free to take any ordering and a standard basis w.r.t. it to perform the test. So if a standard basis would have already been computed, we should not recompute. In this case, @ederc would be right.
However: As we also allow **local** and **mixed** orderings, we need to make sure that the given ordering and the ordering of the stored standard basis have 1>x_i for exactly the same variables x_i. This might require checking more than just the first standard basis in the list for a suitable ordering to appear (think of deliberately computing ideal membership in the polynomial ring and then in its localization at the origin to spot a difference!)

simonbrandhorst: It sounds like these orderings model different rings and different mathematics. 
Mixing them in the same parent seems unnatural to me. Or at least the output should be an ideal over a the new appropriate parent.

afkafkafk13: Yes, indeed. We are suffering from a design difference between Singular and Oscar here: 
- Singular controls global and local computations by means of the monomial ordering -- and it is (and needs to be) possible to specify local orderings for the Singular side also from Oscar (because we need this functionality). 
- Oscar deliberately makes a (mathematically well founded) distinction between Polynomial Rings and their localization, but relies on functionality of Singular for computations in the latter. 

The challenge here is to make sure that two currently unrelated choices do not lead to contradictions: 
- the expectations of the (programming) user, who might have deliberately specified a global or a local ordering to pass to the backend
- the returned result, which currently happens to rely on the (inherited and thus uncontrolled) choice of the ordering of the first precomputed standard basis.

ederc:
OK, let's forget this PR. I think @fingolfin's proposal is the cleanest solution to #2099 .

src/Rings/mpoly-ideals.jl
fingolfin: @afkafkafk13 @simonbrandhorst what you are discussing to me sounds as if there should be a `default_ordering(::Ideal)` instead of (or perhaps in addition to) `default_ordering(::MPolyRing)` ? And then various places that currently do `default_ordering(base_ring(I))` should instead use `default_ordering(I)`?

Anyway, this seems completely orthogonal to the optimizations we need? I suggest you open a separate issue for that, if this is something you think would be valuable.

PR #2139, opened 2023-03-26
Run the renaming script again

fingolfin:


simonbrandhorst:
Thank you, and sorry it will take a little while until the old names fade and the new names come naturally.

simonbrandhorst (approved):

fingolfin:
No worries, I expected this to happen, and it happened to multiple people; there are even pending PRs with this "problem". As you say, it'll just take some time for this to fade. No harm there, other than perhaps confusing some people who only know the new names (= a tiny minority at this point).

One day we might turn those `@alias` invocations into `@deprecate` at which point we'll weed out whatever remains.

PR #2144, opened 2023-03-27
Migration projective schemes ii

HechtiDerLachs:
Continuation of #2114 .

PR #2147, opened 2023-03-27
Migration projective schemes iii

HechtiDerLachs:
Remake of #2144 . Probably due to website issues with GH I couldn't push to that PR anymore. This should now be complete and ready for review.


simonbrandhorst:
Tests fail.

simonbrandhorst:
Tests still fail.

HechtiDerLachs:
> Tests still fail.

Indeed. But this code is not due to me. I will have a look tomorrow.

simonbrandhorst:
@StevellM wrote it.

StevellM:
My bad, I did not know the code for `affine_cone` was experimental... before it was not returning the map, but now yes. So the line of code where I use it (twice) now is not behaving as it should (in `experimental/SymmetricIntersections/SymmetricGrassmannians.jl` line 605). I do not know if I can myself make the fix on this PR... I guess, just asking for the first output for each should be enough:
From
```
J = modulus(OO(intersect(affine_cone(Grtn), affine_cone(PV))))
```
to
```
J = modulus(OO(intersect(affine_cone(Grtn)[1], affine_cone(PV)[1])))
```
?

src/Rings/mpoly.jl
```diff
@@ -1147,3 +1147,20 @@ end
 function _is_integral_domain(R::MPolyRing) 
   return true
 end
+
+@doc Markdown.doc"""
+    total_degree(f::MPolyRingElem, w::Vector{Int})
```
afkafkafk13, line 1152: total_degree for polynomials without specified weight exists somewhere else. Wouldn't it be a good idea to have both methods in the same place?

simonbrandhorst (approved):
Thank you for polishing this Matthias.

afkafkafk13 (approved):
Good to have this out of experimental. Thank you Matthias.

src/Rings/mpoly.jl
wdecker: So far we had two identfiers here: total_degree = the usual total degree.
                                                        degree = degree of an element of a graded polynomial ring
Both are in accordance with the use in the literature and are introduced in the docu. 

Question to everybody: Shouldn`t we use something like weighted_degree here and stick to total_degree as it was?

afkafkafk13: I agree, weighted_degree avoids ambiguities, is in accordance with the use in literature (in particular in the weighted local setting for (semi-)quasihomogeneous singularities) and is a logical choice, when users look for it with <Tab>-completion.

HechtiDerLachs:
> My bad, I did not know the code for `affine_cone` was experimental

No worries! I'm happy to see you're using it. And you're right with your suggestion for correction. I already made the changes and I will push them, once I get the other tests also back running. Thanks for having a look!

src/Rings/MPolyMap/MPolyAnyMap.jl
```diff
@@ -256,3 +256,8 @@ function morphism_type(::Type{D}, ::Type{C}, f::Type{F}) where {D <: Union{MPoly
 end
 
 morphism_type(::D, ::C, ::F) where {D <: Union{MPolyRing, MPolyQuoRing}, C <: NCRing, F} = morphism_type(D, C, F)
+
+function (f::MPolyAnyMap{<:MPolyRing, <:AbstractAlgebra.NCRing})(I::MPolyIdeal)
+  return ideal(codomain(f), [f(g) for g in gens(I)])
+end
+
```
HechtiDerLachs, line 263: @thofma : Are you OK with this? I vaguely recall that you once had some ideas on how ideals should be handled. The functionality here has been there before, but now I moved it. Might be a good point to ask for your approval.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -8,7 +8,7 @@
 # To the outside, this is primarily the map identifying S with its 
 # flattened version.
 ########################################################################
-mutable struct RingFlattening{TowerRingType<:Union{MPolyRing, MPolyQuoRing}, 
+@attributes mutable struct RingFlattening{TowerRingType<:Union{MPolyRing, MPolyQuoRing},
```
HechtiDerLachs, line 11: I found that the `show` function was throwing an error unless I make this attributable. Probably this has to do with the implementation of `show` for `Hecke.Map`s. Is it intended that I get an error? I 'solved' the problem by adding the `@attributes` macro. But that's not really how one should do it, I guess...

src/Rings/MPolyQuo.jl
```diff
@@ -89,6 +89,10 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
   return MPolyQuoRingElem(Base.deepcopy_internal(a.f, dict), a.P, a.simplified)
 end
 
+function is_zero(a::MPolyQuoRingElem)
+  return a.f in modulus(parent(a))
+end
```
HechtiDerLachs: I needed to re-implement this, because the previous generic code was trying to compute a groebner/standard basis. But my rings don't have a singular side, so I put this here. Is it OK? @ederc, @wdecker

src/Rings/MPolyQuo.jl
```diff
@@ -701,7 +709,7 @@ julia> f
 x^3 + x
 ```
 """
-function simplify(f::MPolyQuoRingElem)
+function simplify(f::MPolyQuoRingElem{T}) where {S<:Union{FieldElem, ZZRingElem}, T<:MPolyElem{S}}
```
HechtiDerLachs, line 711: This restriction was also necessary, because the body of the function is using the singular side. Should this be too restrictive, speak up!

src/Rings/MPolyQuo.jl
```diff
@@ -713,6 +721,26 @@ function simplify(f::MPolyQuoRingElem)
   return f::typeof(f)
 end
 
+# Extra method for quotients of graded rings. 
+# TODO: Could this be simplified if the type-parameter signature of decorated rings 
+# was consistent with the one for polynomial rings? I.e. if the first type parameter 
+# was the one for the coefficient rings and not the one for the underlying polynomial ring?
+function simplify(f::MPolyQuoRingElem{MPolyDecRingElem{T}}) where {T<:MPolyElem{<:FieldElem}}
```
HechtiDerLachs: Probably, I should then also add the new functionality with `f.simplified` here, right? @thofma

src/Rings/MPolyQuo.jl
```diff
@@ -713,6 +721,26 @@ function simplify(f::MPolyQuoRingElem)
...
+  g  = f.f
+  f.f = OR(reduce(SR(g), G))
+  return f::elem_type(R)
+end
+
+# By default this does nothing. Simply because there's no general rule what to 
+# do when groebner bases can not be assumed to make sense.
+function simplify(f::MPolyQuoRingElem)
+  return f
+end
```
HechtiDerLachs: Is this OK? @wdecker @ederc

src/Rings/mpoly-graded.jl
```diff
@@ -2424,6 +2424,12 @@ function AbstractAlgebra.promote_rule(::Type{MPolyDecRingElem{S, T}}, ::Type{U})
   end
 end
 
+# For some reason the following is necessary to remove ambiguities
+function AbstractAlgebra.promote_rule(::Type{MPolyDecRingElem{S, T}}, ::Type{MPolyDecRingElem{S, T}}) where {S, T}
+  return MPolyDecRingElem{S, T}
+end
```
HechtiDerLachs, line 2430: Unfortunately, I didn't write down a minimal example for this. But if you too think this is weird and would like to investigate this, remove the method and run the scheme's test code.

src/Rings/mpoly-graded.jl
```diff
@@ -2507,3 +2513,9 @@ end
 Return the ideal in the underlying ungraded ring.
 """
 forget_grading(I::MPolyIdeal{<:MPolyDecRingElem}) = forget_decoration(I)
+
+gens(A::MPolyDecRing, i::Int) = A[i]
+
+### This is a temporary fix that needs to be addressed in AbstractAlgebra, issue #1105.
+# TODO: This still seems to be not resolved!!!
```
HechtiDerLachs, line 2520: See comment. Removing this still breaks the code.

src/Rings/mpoly.jl
HechtiDerLachs: I'm fine with calling it whatever name you find suitable. Let me know when you reach a conclusion.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1682,3 +1682,31 @@ end
   return dim(pre_image_ideal(I))
 end
 
+### Hack for a detour to speed up mapping of elements 
+# This is terribly slow in all kinds of quotient rings 
+# because of massive checks for `iszero` due to memory 
+# management.
```
HechtiDerLachs, line 1688: Maybe with the latest changes along #2099 these hacks can now be removed?

HechtiDerLachs:
Sorry about the size. But this sure was a hell of a lot of work. Especially debugging all the promote rules and coercions.

src/Rings/mpoly.jl
simonbrandhorst: We seem to agree on `weighted_degree`.

src/Rings/MPolyQuo.jl
simonbrandhorst: I seem to recall a recent pull request on this by @thofma ?

simonbrandhorst: Ah but this concerned rings which have a singular side?

src/Rings/MPolyQuo.jl
```diff
@@ -713,6 +721,26 @@ function simplify(f::MPolyQuoRingElem)
   return f::typeof(f)
```
simonbrandhorst:
```suggestion
  return f
```

src/Rings/MPolyQuo.jl
thofma: This is overwriting/conflicting with the `iszero` method below. This will basically make the zero check as slow as before.

`MPolyQuoRing` assumes that there is some kind of unique representative that we can compute (a la `simplify`).

src/Rings/MPolyQuo.jl
HechtiDerLachs: I also found this weird. But isn't this due to recent changes by @thofma ?

src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/attributes.jl
```diff
@@ -73,8 +73,14 @@ julia> exponents(cc)
...
 """
-exponents(c::CohomologyClass) = matrix(ZZ, [k for k in AbstractAlgebra.exponent_vectors(polynomial(c).f)])
+function exponents(c::CohomologyClass) 
+  simplify!(c)
+  matrix(ZZ, [k for k in AbstractAlgebra.exponent_vectors(polynomial(c).f)])
+end
 
+function simplify!(c::CohomologyClass)
+  c.p = simplify(c.p)
+end
```
HechtiDerLachs, line 83: @HereAround : I made an attempt to repair the stuff here. For some reason, the polynomials were not brought to normal form automatically anymore. Do you agree with the solution here?

src/Rings/MPolyQuo.jl
simonbrandhorst: Yea, but it can be removed (or stay, whatever).

src/Rings/MPolyQuo.jl
HechtiDerLachs: > `MPolyQuoRing` assumes that there is some kind of unique representative that we can compute (a la `simplify`).

I do not think, this is a reasonable assumption.

HechtiDerLachs: > Ah but this concerned rings which have a singular side?

The opposite: Rings which don't have a singular side. Because ideal membership can be deflected via `RingFlattening` while standard basis computation and reduction to normal form can not.

thofma: Then maybe `MPolyQuoRing` is not the right thing to use. Without `simplify`, it does not satisfy the ring interface, since arithmetic is not working with the current implementation.

thofma: I don't like that we remove the `simplify` from some functions to make them unnecessarily slower (which there were complaints about), but keep them for others. It is just an inconsistent mess.

HechtiDerLachs: > Then maybe `MPolyQuoRing` is not the right thing to use. Without `simplify`, it does not satisfy the ring interface, since arithmetic is not working with the current implementation.

I'm afraid, I'm missing some discussions that you have had with others to reach these conclusions. So I don't really understand what the problem is. 

What it looks like from my perspective: Most probably, `simplify` (or `simplify!`) is supposed to pick a distinguished representative for every element. This is usually achieved by choice of a monomial ordering when working over a reasonable field of coefficients. 

The arithmetic in quotient rings requires an equality check for elements (`==`). If a normal form is provided via `simplify`, one can use that to decide equality. But mathematically, the equality check is independent of that and can also be carried out by testing ideal membership (as done in the generic code I propose here). For the coefficient rings I am interested in there probably is no reasonable meaning of `simplify` and even if there was, we would certainly lack the singular backend to implement it. 

So far, we wanted to model the mathematical objects in Oscar with no additional "non-mathematical" information and requirements; the prominent example being that polynomial rings do not have an ordering. In this sense it seems to me that requiring `simplify` to be implemented for quotient rings in general is breaking with this convention. Or am I understanding something wrong?

HechtiDerLachs: > I don't like that we remove the `simplify` from some functions to make them unnecessarily slower (which there were complaints about), but keep them for others. It is just an inconsistent mess.

I see the point about the inconsistency. But I am keeping the simplify functionality for all the relevant cases where it actually has a computational backend. And there your recent changes are an improvement! But why not make a different dispatch for the exotic remaining ones? That method could also make use of the `a.simplified` field for the `iszero` check but using ideal membership instead. But it would have to be well documented how the value of this field must be interpreted in this general case. 

Be aware that my last commits were taking care of the duplicate method. This looks different now.

HechtiDerLachs: I just pushed an example of what I had in mind with my last message. Now the `iszero` is implemented generically.

thofma: Every model assumes something. So far the `MPolyQuoRing` was assuming, that one has a way of obtaining a unique representative (as is evident by the usage of `simplify`). This makes them performant in many situations. Is this the only way to model quotient rings? Surely no, there is also a model for quotients of Euclidean domains, which is even faster, but of course assumes more. I don't see why this is breaking with the convention of modelling mathematics. We are not claiming to model everything with the same underlying representation. For a specific ring you might get a other implementation. But a quotient ring constructor will yield a quotient ring.

Please correct me if I am wrong, but it seems that you want yet another model for a quotient, which is the most generic one and at the same time less performant one. One models `R/I` by working in `R` with the additional `==(x, y) = (x - y) in I` method. It does not matter whether it is a quotient of a multivariate polynomial ring.

What I am not comfortable with is changing `MPolyQuoRing` to be less and less useful for the rings it was intended to by making it more and more generic. Thanks for the latest commit. I am happy with using dispatch for now. It will be a bit awkward if one wants to change all the functions that currently use `simplify`, but this should not matter for now.

src/Rings/MPolyQuo.jl
```diff
@@ -505,7 +505,13 @@ parent(a::MPolyQuoRingElem) = a.P
 
 zero(R::MPolyQuoRing) = MPolyQuoRingElem(zero(base_ring(R)), R, true)
 
-iszero(a::MPolyQuoRingElem) = iszero(simplify(a).f)
+iszero(a::MPolyQuoRingElem{T}) where {S<:Union{FieldElem, ZZRingElem}, T<:MPolyElem{S}} = iszero(simplify(a).f)
+iszero(a::MPolyQuoRingElem{<:MPolyDecRingElem{<:FieldElem}}) = iszero(simplify(a).f)
+
+function is_zero(a::MPolyQuoRingElem)
+  return iszero(simplify(a).f)
```
thofma, line 512:
```suggestion
  return a.f in modulus(parent(a))
```
or something like that?

src/Rings/MPolyMap/MPolyAnyMap.jl
thofma: Looks good. The general machinery has not been implemented.

src/Rings/mpoly-graded.jl
thofma: This is fine.

src/Rings/MPolyQuo.jl
HechtiDerLachs: I see your point about making some basic assumptions when designing a new type. Hence, I also understand objections to my overriding the generic methods for `iszero` and `simplify`, especially right after you had polished them. Still, it would be nice if I could make use of the existing types and accompanying infrastructure for my purposes. Maybe, it is then better that I do special method dispatch for my particular cases and leave the generic code as it was. I'll try to be concise and explain what I do in comments.

thofma: I think your changes are fine. I am happy if we add it in this form.

src/Rings/mpolyquo-localizations.jl
simonbrandhorst: No clue. Who hacked them?

HechtiDerLachs: It was me. But I didn't try it out again. Either way: this can wait until later.

src/Rings/MPolyQuo.jl
```diff
@@ -682,7 +682,7 @@ function Base.:(==)(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}) where T
 end
 
 @doc Markdown.doc"""
-    simplify(f::MPolyQuoRingElem)
+    simplify(f::MPolyQuoRingElem{T}) where {S<:Union{FieldElem, ZZRingElem}, T<:MPolyElem{S}}
```
HechtiDerLachs, line 685: I know we want to avoid having these type annotations in the docstrings. But in the previous version, the docstring was not found. Since this tends to be a trial-and-error-game, I put it here for now. If it works, I advocate for keeping it that way as an exception.

HechtiDerLachs:
Same status as in #2106 .

PR #2148, opened 2023-03-27
Printing guidelines

simonbrandhorst:
The new printing guidelines.
Feel free to suggest improvements.


docs/src/DeveloperDocumentation/printing.md
```diff
@@ -0,0 +1,638 @@
...
+  with defining polynomial 9x^4 - 5x^3 + 3x^2 - 3x + 1
+    over
+    Number field of degree 2
+    with defining polynomial x^2+1
+    over
+      QQ
+```
+```
+Number field of degree 2 over Number field
+```
```
afkafkafk13: Am I misunderstanding the example or shouldn't the outermost be printed instead of the innermost here. 
Then it would read '...of degree 10...' in :oneline.

simonbrandhorst: Right. Typo.

docs/src/DeveloperDocumentation/printing.md
```diff
@@ -0,0 +1,638 @@
+# Printing in Oscar
+
+## The 2+1 new print modes
+We propose two user print modes `:details` and `:oneline` and one internal
+print mode `:supercompact`. The latter is for use during recursion,
```
lkastner:
```suggestion
Oscar has two user print modes `:details` and `:oneline` and one internal
print mode `:supercompact`. The latter is for use during recursion,
```

lkastner: Since these are instructions for developers, I think it is better not to propose, since this might suggest that it is ok to deviate without asking.

docs/src/DeveloperDocumentation/printing.md
```diff
@@ -0,0 +1,638 @@
...
+julia> PD[1][1]
+Ideal of Multivariate Polynomial Ring in 3 variables over GF(2)
+generated by
+ z + 1
+ x^2 + y^2
+```
+
+## Styleguide
+### :details
+```
```
lkastner: I don't like having a separate style guide section here. Can there maybe be an executive summary of this in the style guide, but without examples, and then a link to this exhaustive guide?

docs/src/DeveloperDocumentation/printing.md
```diff
@@ -0,0 +1,638 @@
...
+  ```
+  is easier to read than
+  ```
+  Structure over bla over blub in 5 generators in 6 generators in 7 generators with 25 relations
+  ```
+
+
+# Questions, discussion, TODOs
+
+## Grammar
```
lkastner: Will all this stuff also be published?

docs/src/DeveloperDocumentation/printing.md
```diff
@@ -0,0 +1,638 @@
...
+```julia
+The hyperplane of ℝ² described by x₂ = 0
+```
+
+# Technical aspects
+
+## How julia prints things
+
+There are lots of open issues regarding with the printing in julia.
+Tommy is too lazy to find those.
```
lkastner:
```suggestion

There are lots of open issues regarding with the printing in julia.
```

docs/src/DeveloperDocumentation/printing.md
```diff
@@ -0,0 +1,638 @@
...
+
+Here is an alternative version, which needs to be used in case the detailed
+printing does not contain newlines.
+
+```julia
+function Base.show(io::IO, R::NewRing)
+    if get(io, :supercompact, false)
+      print(io, "supercompact printing of newring")
+    else
+      println(io, "I am a new ring and always print in one line: )
```
lkastner:
```suggestion
      println(io, "I am a new ring and always print in one line" )
```

docs/src/DeveloperDocumentation/printing.md
simonbrandhorst: I will remove it by the end of this week.

docs/src/DeveloperDocumentation/printing.md
simonbrandhorst: Works for me. 
The detailed part is intended for the discussions and as a reference when we start implementing the new guidelines.

simonbrandhorst: @lkastner I have added a summary to the styleguide. Including one example. Does that work for you?

simonbrandhorst:
@thofma  new version includes the latex printing #2161

simonbrandhorst:
No clue why the  docs do not build. Locally on my machine they do.

thofma:
https://github.com/oscar-system/Oscar.jl/actions/runs/4553567269/jobs/8030280443?pr=2148#step:6:134: 
```
┌ Error: 'Matrix-groups' is not unique in src/Groups/intro.md.
└ @ Documenter.CrossReferences ~/.julia/packages/Documenter/H5y27/src/Utilities/Utilities.jl:32
```
The `Matrix groups` heading we introduce here makes the heading non-unique and the reference fails.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,189 @@ existing block.
...
+
+Here is an alternative version, which needs to be used in case the detailed
+printing does not contain newlines.
+
+```julia
+function Base.show(io::IO, R::NewRing)
+    if get(io, :supercompact, false)
+      print(io, "supercompact printing of newring")
+    else
+      println(io, "I am a new ring and always print in one line" )
```
joschmitt: If there is a `println` it will not print on one line, will it?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,189 @@ existing block.
...
+      print(IOContext(io, :supercompact => true), base_ring(R))
+    end
+end
+```
+
+#### Not working as expected
```
joschmitt: I don't really understand from this, what is not working as expected here and why. Also, is the "Alternative" section above a solution to this problem or is it unrelated?

joschmitt (approved):
Looks good, thank you!

docs/src/DeveloperDocumentation/styleguide.md
thofma: Yes, should be `print`.

docs/src/DeveloperDocumentation/styleguide.md
thofma: It should be explained by
> Then the following will *not* be used for array/tuple printing.
> It will be used for `print(io, R::NewRing)` though.

Can you elaborate what is not clear?

joschmitt: You are right, all the information is there, but it took me quite a while to put it together in the right order.
Maybe there could be a short explanation "Then the following will *not* be used for array/tuple printing, BECAUSE ..." to make it clearer, why the "Not working as expected"-version fails.
Also, in the "Alternative" section, it wasn't immediately clear to me that the `else` case in fact covers both detailed and oneline printing (so that the function replaces both `show` functions from before).

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,189 @@ existing block.
...
+    println(io, "I print with newlines")
+    print(io, R.base_ring)
+end
+```
+
+```julia
+function Base.show(io::IO, R::NewRing)
+    if get(io, :supercompact, false)
+        print(io, "supercompact printing of newring ")
+        print(io, R.base_ring) # this prints the :oneline version of newring
```
joschmitt: Doesn't this violate the "No nested printing" rule?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,189 @@ existing block.
...
+- Only the first word is capitalized e.g. `Polynomial ring`
+- one should use `:supercompact` for nested printing in compact
+- nested calls to `:oneline` (if you think them really necessary) should be at the end,
+  so that one can read sequentially. Calls to `:supercompact` can be anywhere.
+#### Super compact printing
+- no nested printing
+- a single term or a symbol/letter mimicking mathematical notation
+- should usually only depend on the type and not of the type parameters or of
+  the concrete instance - exceptions of this rule are possible e.g. for `GF(2)`
+- no nested printing. In particular variable names and `base_ring` must not be displayed.
```
joschmitt, line 277: Nitpick: rules 1 and 4 are "no nested printing"

simonbrandhorst: Thanks.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: Indeed, thank you. I removed it.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: I added a few explanations.

joschmitt: Thank you!

lkastner (approved):
Thank you. I will look into `src/PolyhedralGeometry` to adjust accordingly.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+end
+```
+
+
+##### Detailed printing in a single line.
+
+This version needs to be used in case the detailed
+printing does not contain newlines.
+Then detailed and one line printing agree.
+The if clause takes care of supercompact printing as well.
```
```suggestion
The `if` clause takes care of supercompact printing as well.
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+#### Mockup
+
+##### Detailed printing with a new line.
+
+```julia
+struct NewRing
+  base_ring
+end
+```
+This implements the detailed printing. Note that the new line is needed for
```
fingolfin: "The" new line? There are two. "The new lines" maybe? I am not sure I understand what is being said here, though, hence I can't make a suggestion for improvement

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+this to work. The last print statement must not add a new line.
```
fingolfin: What does "this" refer to? What is working / not working?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+  end
+end
+```
+
+
+##### Detailed printing in a single line.
+
+This version needs to be used in case the detailed
+printing does not contain newlines.
+Then detailed and one line printing agree.
```
fingolfin, line 347:
```suggestion
Then `detailed` and `one line` printing agree.
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+    print(io, "supercompact printing of newring")
+  else
+    # nested printing allowed, preferably supercompact
+    println(io, "I am a new ring and always print in one line" )
+    print(IOContext(io, :supercompact => true), base_ring(R))
+  end
+end
+```
+
+##### The following is not working as expected and should not be used
```
fingolfin, line 382: I don't understand this "counterexample". What exactly causes it to be broken? It looks identical to "good" example except that the `text/plain` output function does not contain newlines.

Does that mean there is code somewhere which detects if `show` methods issue newlines, and then does something different based on that? Is this described somewhere? I couldn't find it when I just skimmed this document.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+
+```julia
+function Base.show(io::IO, ::MIME"text/plain", R::NewRing)  # do not implement me like this
+  print(io, "I am a new ring with a detailed printing of one line")
+end
+```
+
+Then the following will *not* be used for array/tuple printing.
```
fingolfin, line 394: I guess this is the core bit I don't understand: why is the following not used for array/tuple printing when the previous method only prints a single newline and doesn't perform recursion?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+It will be used for `print(io, R::NewRing)` though.
+
+```julia
+function Base.show(io::IO, R::NewRing)
+  if get(io, :supercompact, false)
+    print(io, "supercompact printing of newring")
+  else # this is what we call one line
+    print(io, "one line printing of newring with ")
+    print(IOContext(io, :supercompact => true), "supercompact ", R.base_ring)
```
fingolfin, line 403:
```suggestion
    print(IOContext(io, :supercompact => true), "supercompact ", base_ring(R))
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+  end
+end
+```
+This example illustrates the unexpected behavior.
+```julia
+julia> R = NewRing(1)
```
fingolfin, line 409:
```suggestion
julia> R = NewRing(1);
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+function Base.show(io::IO, R::NewRing)
+  if get(io, :supercompact, false)
+    print(io, "supercompact printing of newring")
+  else # this is what we call one line
+    print(io, "one line printing of newring with ")
+    print(IOContext(io, :supercompact => true), "supercompact ", R.base_ring)
+  end
+end
+```
+This example illustrates the unexpected behavior.
```
fingolfin, line 407: I really like that we show the "expected output" here; perhaps we could do the same in the two "good" examples that come before?

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+julia> [R,R]  # one line printing is ignored
+2-element Vector{NewRing}:
+ I am a new ring with a detailed printing of one line
+ I am a new ring with a detailed printing of one line
+
+julia> print(Base.stdout, R)
+one line printing of newring with supercompact QQ
+```
+
+## Latex and unicode printing
```
```suggestion
## LaTeX and Unicode printing
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+
+### Latex output
+Some types support latex output.
```
```suggestion
### LaTeX output
Some types support LaTeX output.
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+julia> show(stdout, "text/latex", x^2 + 2x + x^10)
+x^{10} + x^{2} + 2 x
+
+julia> show(stdout, "text/latex", Qx[x x^2; 1 1])
+\begin{array}{cc}
+x & x^{2} \\
+1 & 1
+\end{array}
+```
+To support latex output for your type implement
```
```suggestion
To support LaTeX output for your type implement
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,213 @@ existing block.
...
+```
+Base.show(io::IOContext, "text/latex")
```
```suggestion
Base.show(io::IOContext, ::MIME"text/latex")
```

fingolfin (commented):
Thanks, I (still) like the proposal, these are just some nitpicks, and possibly me being dense, sorry about that

thofma:
@fingolfin I suggest reading https://hackmd.io/cwOoBrTaReihNtmFjSdtTg#Technical-aspects that will answer your questions about newlines.

Edit: We did not add this part of the working document to the documentation on purpose.

docs/src/DeveloperDocumentation/styleguide.md
thofma: This part is semi-true. Since we funnel almost everything through the `expressify` system, this is taken care of by default. There is no need to implement this, say, for a new multivariate polynomial type.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: It is just how julia works. Nothing on the Oscar side here.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: O.K. for the sake of setting an example I can implement `base_ring(R::NewRing)`

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: I trust your word on this. I did not try it out.

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: What should I put instead?

thofma: Not sure, lets keep it. We can correct it if it pops up somewhere.

docs/src/DeveloperDocumentation/styleguide.md
fingolfin: OK, but I still find this confusing when looked at in isolation (i.e. when one gets here via a link and not via reading the whole section sequentially). So I would just add a little bit extra (repeating a few things that were said before)
```suggestion
##### The following is not working as expected and should not be used

This example does not work correctly because the `detailed` printing does not
include a newline, which is expected by the Julia printing system. To correctly
support single line `detailed` printing, read the preceding section.
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,249 @@ existing block.
...
+1 & 1
+\end{array}
+```
+
+```
+Base.show(io::IOContext, ::MIME"text/latex")
+```
+
+### Unicode printing
+Per default output should be ANSI only (no Unicode). Implementors of
```
```suggestion
Per default output should be ASCII only (no Unicode). Implementors of
```

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,249 @@ existing block.
...
+`Base.show` and related functions can branch on the output of
+`Oscar.is_unicode_allowed()` to display objects using non-ASCII characters.
+This will then be used for users which enabled Unicode using
+`allow_unicode(true)`. Note that
+
+- there must be a default ANSI only output, since this is the default setting
```
```suggestion
- there must be a default ASCII only output, since this is the default setting
```
There are many ANSI encodings, ASCII is just one of them

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -241,3 +217,249 @@ existing block.
...
+
+base_ring(R::NewRing) = R.base_ring
+```
+This implements the detailed printing.
+Note that the new line is needed for this to work
+because of how the julia printing system works.
+The last print statement must not add a new line.
+
+```julia
+function Base.show(io::IO, ::MIME"text/plain", R::NewRing)
```
fingolfin, line 309: @simonbrandhorst suggested that we turn these running examples into doctests, which I think is a good idea. But of course we can do this *after* merging this PR

fingolfin (approved):
I am happy with one minor addition

docs/src/DeveloperDocumentation/styleguide.md
simonbrandhorst: I did not like the `julia>` prompt in the examples. It messes with being able to copy paste it in a `.jl` file.
Hence I did not convert the to doctests for the time being.

PR #2150, opened 2023-03-27
Make MPolyQuoRingElem great again

### LINKED ISSUE #2099: is_zero is slow for quotient rings

thofma:
Store the information whether an element is already simplified. Also
optimize the `zero` and `is_zero` methods. Maybe closes #2099?

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -71,10 +71,11 @@ default_ordering(Q::MPolyQuoRing) = default_ordering(base_ring(Q))
 mutable struct MPolyQuoRingElem{S} <: RingElem
   f::S
   P::MPolyQuoRing{S}
+  simplified::Bool
 
-  function MPolyQuoRingElem(f::S, P::MPolyQuoRing{S}) where {S}
+  function MPolyQuoRingElem(f::S, P::MPolyQuoRing{S}, simplified = false) where {S}
     @assert parent(f) === base_ring(P)
-    return new{S}(f, P)
+    return new{S}(f, P, simplified)
   end
 end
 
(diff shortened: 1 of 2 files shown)


simonbrandhorst (approved):

ederc (approved):

src/Rings/MPolyQuo.jl
```diff
@@ -696,13 +702,15 @@ x^3 + x
...
+  f.simplified && return f
   R  = parent(f)
   OR = oscar_origin_ring(R)
   SR = singular_origin_ring(R)
   G  = singular_origin_groebner_basis(R)
   g  = f.f
   f.f = OR(reduce(SR(g), G))
-  return f::elem_type(R)
+  f.simplified = true
+  return f::typeof(f)
```
fingolfin, line 713: Isn't this 100% equivalent to `return f`?

thofma: Ah yes, not sure what I was thinking. There was some type annotation before and I thought I could improve it, but it makes obviously no sense.

PR #2151, opened 2023-03-27
Avoid creating a matrix space just to create a zero or identity matrix

fingolfin:


thofma:
They should be forbidden to construct.

thofma (approved):

PR #2156, opened 2023-03-28
Adapt mpoly_type methods to AA 0.29

fingolfin:


thofma:
What is with the ones that are kept?

fingolfin:
Those are not yet in Nemo, as I only just now added them via https://github.com/Nemocas/Nemo.jl/pull/1434 -- we could release Nemo 0.33.7 now to add them. Or we wait a bit, merge this PR first, then I'll update https://github.com/Nemocas/Nemo.jl/pull/1408, we merge *that*, and then release Nemo 0.33.7 -- and *then* remove the final `mpoly_type` methods here... Phew. Things are so much easier when everything is spread across four packages.

thofma (approved):

PR #2157, opened 2023-03-28
Give experimental a more canonical structure

### LINKED ISSUE #2146: Structure in experimental
I already talked about this to @fingolfin and @HereAround . The folder structure for experimental stuff is a bit confusing. There is `experimental` for source code, separate from `src`, while `test/Experimental` contains the tests and `docs/src/Experimental` contains the docs. Max proposed the following structure, which I like very much (if it is possible):
```
experimental/PACKAGE/{src,test,docs}
```
Then one could still have `experimental/test.jl` and `experimental/experimental.jl` for the code inclusions, not sure about the docs. In case there is not too much discussion, I would go ahead and implement these things in the lie algebra PRs #2145 and #2115, of course while coordinating with the respective authors. I don't think it is necessary to change this for the remaining things, but I'd also adapt the developer docs and then the old style should slowly fade out.

Maybe this also makes things easier for #2011.

lkastner:
We discussed in #2146 that the structure of new additions to `experimental` should become more canonical, also in light of migrating things out of `experimental`, e.g. to `src`. This PR contains the following:
- Every package in experimental should have the structure `experimental/PACKAGE_NAME/{src,docs,test}`
- These folders should have certain files such that they are automatically recognized by loading scripts
- All this is explained in the `intro.md` in `docs/src/experimental/`
- `PlaneCurve` has been modified to become a template for now


Project.toml
```diff
@@ -8,6 +8,7 @@ AbstractAlgebra = "c3fe647b-3220-5bb0-a1ea-a7954cac585d"
 AlgebraicSolving = "66b61cbe-0446-4d5d-9090-1ff510639f9d"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 GAP = "c863536a-3901-11e9-33e7-d5cd0df7b904"
+Glob = "c27321d9-0574-5035-807b-f59d2c89b15c"
```
fieker: Is glob stable? Supported? Core Julia?

fingolfin: What is "Core Julia"? It is not a stdlib if that's what you mean, rather https://github.com/vtjnash/Glob.jl . It is stable in that its interface has been basically unchanged for the past couple years. Its author is one of the primary Julia developer. I just asked very recently if he could make a new release and he did so immediately.

Anyway, we could also rewrite the code to just use built-in Julia stuff, it'd not be hard. But maybe before we think about that, we should first see if the rest of this PR is something we want to go forward with

docs/make_work.jl
```diff
@@ -24,11 +25,47 @@ using Documenter, DocumenterCitations
 # Remove the module prefix
 Base.print(io::IO, b::Base.Docs.Binding) = print(io, b.var)
 
+# This is needed to make Documenter.jl follow symlinks, which it is also
+# supposed to be able to do, since there has been a PR about this since 2017:
```
fingolfin: I don't understand: yes there is this PR, but it's not merged, so in how is Documenter "supposed to be able to do" this?

docs/make_work.jl
```diff
@@ -24,11 +25,47 @@ using Documenter, DocumenterCitations
...
+# https://github.com/JuliaDocs/Documenter.jl/pull/552`:
```
```suggestion
# https://github.com/JuliaDocs/Documenter.jl/pull/552:
```

docs/make_work.jl
```diff
@@ -24,11 +25,47 @@ using Documenter, DocumenterCitations
...
+# The Julia docs have the same issue, see:
+# https://github.com/JuliaLang/julia/blob/master/doc/make.jl#L19
+Base.walkdir(str::String) = Base.walkdir(str; follow_symlinks=true)
```
fingolfin, line 33: Again I am confused: the code in the Julia repo you link to is about copying files; but the line above is type piracy, to change the default value of `follow_symlinks` from `false` to `true`. So perhaps:
> We monkey-patch Base.walkdir to use `true` as default value for `follow_symlinks` (normally `false` is the default), in order to "trick" the Documenter code into following those symlinks.

docs/src/experimental/intro.md
```diff
@@ -0,0 +1,58 @@
...
+`experimental` is supposed to be a staging area for new features whose
+interface is still to be stabilized. Also code is allowed to reside in
+`experimental` while it is brought up to Oscar standard.
+
+## Structure
+To look at a sample for the structure for a new project in `experimental` have
+a look at `experimental/PlaneCurve`. The general structure is
+```
+experimental\
+   PACKAGE_NAME\
```
fingolfin: Hmm, maybe this?
```suggestion
experimental/
   PACKAGE_NAME/
```

experimental/Experimental.jl
```diff
@@ -1,7 +1,12 @@
+import Glob
+for fname in Glob.glob("*/src/main.jl", "experimental")
+  fname = SubString(fname, 14, length(fname))
```
fingolfin: I dislike magic constants like the 14 here. Where does it come from? `length("experimental/") == 13`... 

Anyway, isn't this code executed within `experimental anyway? That would suggest
```suggestion
for fname in Glob.glob("*/src/main.jl")
```

fingolfin (commented):
Thanks for working on this!

docs/make_work.jl
lkastner: Well, if even the Julia people say "until 522 is merged". That seemed to me like even they agree with this sentiment.

docs/src/experimental/intro.md
lkastner: Thanks. But I wanted to ask you anyway: How did you get it to look so nice in #2146 ? I don't recognize all symbols there.

docs/src/experimental/intro.md
```diff
@@ -12,6 +12,13 @@ The code in `experimental` is supposed to be mathematically correct,
 be stabilized. Also code is allowed to reside in `experimental` while it is
 brought up to Oscar standard.
 
+!!! danger "Dependencies"
+    - Code from `src` must never use code from `experimental`
+    - Say there are two packages `A` and `B` in `experimental`, and `B` depends
+      on `A`. That means that `B` cannot be moved to `src` before `A`. Worse:
+      If `A` gets abandoned, `B` might share that fate. So please consider
+      carefully in such situations.
```
lgoettgens: Please note, that this currently is the case. Cf. [binomial_ideals](https://github.com/oscar-system/Oscar.jl/blob/22a59d64709d4ba0ddb01824daf78c8dc7c86671/src/Oscar.jl#L515-L516)

lkastner: That is not the only case. Nevertheless we want to avoid this in the future, since now the intention of `experimental` has become clearer. But as discussed in #2146 I will leave older packages alone for the time being.

fieker:
On Wed, Mar 29, 2023 at 01:43:23AM -0700, Lars Kastner wrote:
> @lkastner commented on this pull request.
> 
> 
> 
> > @@ -12,6 +12,13 @@ The code in `experimental` is supposed to be mathematically correct,
>  be stabilized. Also code is allowed to reside in `experimental` while it is
>  brought up to Oscar standard.
>  
> +!!! danger "Dependencies"
> +    - Code from `src` must never use code from `experimental`
> +    - Say there are two packages `A` and `B` in `experimental`, and `B` depends
> +      on `A`. That means that `B` cannot be moved to `src` before `A`. Worse:
> +      If `A` gets abandoned, `B` might share that fate. So please consider
> +      carefully in such situations.
> 
> That is not the only case. Nevertheless we want to avoid this in the future, since now the intention of `experimental` has become clearer. But as discussed in #2146 I will leave older packages alone for the time being.
> 
I'm not condoning or defending it, but a very likely source for
contamination is going to be code in experimental that extends
infrastructure that should be in one of the lower packages...
It would be introduced to avoid having to juggle pull requests and
releases of n>1 packages. Certainly intended only for a short period of
time...

lkastner:
> I'm not condoning or defending it, but a very likely source for contamination is going to be code in experimental that extends infrastructure that should be in one of the lower packages... It would be introduced to avoid having to juggle pull requests and releases of n>1 packages. Certainly intended only for a short period of time...

Yes, but you are also more experienced and in some sense not bound by these rules (same for Max, Ben and Tommy). And in my experience rules that are formulated as suggestions or proposals tend to be broken immediately. If someone hasn't been working with Oscar for a long time they will not know which things they can rely on in experimental.

lkastner:
Ok, so after discussion with @benlorenz I made it such that an error is thrown for new packages if they are missing the `experimental/PKG/src/PKG.jl` or `experimental/PKG/test/runtests.jl`. For the docs we are more forgiving. The other packages being worked on could then decide to add themselves to the list of old experimental packages to be ignored or adapt to the new structure.

benlorenz (approved):

docs/make_work.jl
lkastner: That makes more sense, thanks.

docs/src/experimental/intro.md
fingolfin: I used the `tree` command line tool (see <http://mama.indstate.edu/users/ice/tree/> but I installed it via Homebrew, it is also available for most Linux distros)

docs/doc.main
```diff
@@ -99,7 +99,7 @@
         "Hecke/orders/frac_ideals.md",
      ],
      "NumberTheory/abelian_closure.md",
-     "Experimental/galois.md",
+     "Experimental_old/galois.md",
```
fingolfin: I didn't see `docs/src/Experimental_old` mentioned in the PR description, what is the idea here: to rename the existing `docs/src/Experimental` to `docs/src/Experimental_old` now and then gradually migrate things back to `Experimental` as they are adapted to the new scheme?

Why is this beneficial to us? Just to make it "obvious" what was migrated, and what not (yet?)? But wouldn't that solve itself over time anyway, and be kinda "obvious" anyway ("all files other than `intro.md` in this directory should be migrated away" isn't too hard to explain) ?


If we do it this way, we should be careful to check if there are in-flight PRs that change things in `docs/src/Experimental`, else they might get broken and adapting them may be non-trivial for non-git-wizards.

But perhaps no such PRs exists? I did not check.

docs/make_work.jl
```diff
@@ -24,11 +24,55 @@ using Documenter, DocumenterCitations
...
+# We monkey-patch Base.walkdir to use true as default value for follow_symlinks
+# (normally false is the default), in order to "trick" the Documenter code into
+# following those symlinks.
+# See also:
+# https://github.com/JuliaDocs/Documenter.jl/pull/552
+# https://github.com/JuliaLang/julia/blob/master/doc/make.jl#L19
+Base.walkdir(str::String) = Base.walkdir(str; follow_symlinks=true)
+
+# When we read a `doc.main` from an experimental package, we need to equip all
+# its entries with a prefix to fix with out docs.
```
fingolfin: I don't understand, what does "to fix with out docs" mean?

src/Oscar.jl
```diff
@@ -28,6 +28,32 @@ const jll_deps = String["Antic_jll", "Arb_jll", "Calcium_jll", "FLINT_jll", "GAP
...
 
+# We read experimental and filter out all packages that follow our desired
+# scheme. Remember those packages to avoid doing this all over again for docs
+# and test.
+# We don't want to interfere with existing stuff in experimental though.
+const expdir = joinpath(@__DIR__, "../experimental")
+const oldexppkgs = [
+  "ExteriorAlgebra", "GaloisGrp", "GITFans", "GModule", "JuLie",
+  "LinearQuotients", "Matrix", "ModStd", "MPolyRingSparse",
+  "Rings", "Schemes", "SymmetricIntersections", "FTheoryTools"
```
fingolfin: Super unimportant nitpick suggestion:

maybe put one per line so that whenever we convert one to the new style, we just have to delete a line. Also sort alphabetically?
```suggestion
  "ExteriorAlgebra",
  "FTheoryTools",
  "GaloisGrp",
  "GITFans",
  "GModule",
  "JuLie",
  "LinearQuotients",
  "Matrix",
  "ModStd",
  "MPolyRingSparse",
  "Rings",
  "Schemes",
  "SymmetricIntersections",
```

src/Oscar.jl
```diff
@@ -28,6 +28,32 @@ const jll_deps = String["Antic_jll", "Arb_jll", "Calcium_jll", "FLINT_jll", "GAP
...
+  if !isfile(joinpath(expdir, pkg, "src", "$pkg.jl"))
+    error("experimental/$pkg is incomplete: $pkg/src/$pkg.jl missing.")
+  end
+  if !isfile(joinpath(expdir, pkg, "test", "runtests.jl"))
+    error("experimental/$pkg is incomplete: $pkg/test/runtests.jl missing.")
+  end
+end
+
+# force trigger recompile when folder changes
+include_dependency("../experimental")
```
fingolfin, line 64: I didn't realize `include_dependency` works for directories. Its doc string talks specifically about files. Did you verify this works cross platform? (I assume the goal is to detect if new subdirectories were added).

benlorenz: Yes the goal is to detect new subdirectories, otherwise adding a new feature into experimental won't trigger precompilation and thus it won't be loaded. (Until something else in Oscar is modified)

`include_dependency` basically just records path + mtime and I did verify that the mtime for a directory does work as expected on macOS, in addition to Linux where I experimented some more with making sure precompilation is triggered properly for new folders.

docs/doc.main
lkastner: You are right, the renaming is a rudiment of a different approach I tried. I'll undo it.

docs/make_work.jl
lkastner: It's two typos, it should say "to fit with our docs". Sorry.

fingolfin (approved):

Issue #2158, opened 2023-03-28
CI failure due to serialization of universal polynomials


lgoettgens:
Since the merge of #2113, there are CI failures on master, e.g. [https://github.com/oscar-system/Oscar.jl/actions/runs/4547519262/jobs/8017499717](https://github.com/oscar-system/Oscar.jl/actions/runs/4547519262/jobs/8017499717).

As far as I can tell, the test sets for serialization of universal polynomials over {Rationals, Integers modulo 6,Finite Field} error, the other cases pass.

thofma:
@SoongNoonien any idea what could be wrong?

SoongNoonien:
As the tests worked with all base rings as I implemented the serialization this may be a new issue with the universal polynomials in general similar to https://github.com/oscar-system/Oscar.jl/issues/2076. So me guess is that the error occurs while creating the universal polynomial ring. I'll try to reproduce it here to make some further tests.

fingolfin:
I've reverted PR #2113 for now to make the tests pass again. This looks like one of those unfortunate cases were changes "cross", i.e., the CI tests passed separately, but not in combination. Perhaps we should try the beta "merge queue" feature which is meant to prevent exactly this kind of problem (well, up to a degree -- if one part of the changes is that a different package made a new release, then we can't fully avoid this)

fingolfin:
@SoongNoonien can you resubmit your changes in a new PR? (Basically you can use `git revert` my revert commit; just edit the commit message to match what you had before). Alternatively I can do that for you.

lkastner:
cc @antonydellavecchia

benlorenz:
Apart from the serialization commit which just seems to be the only code in Oscar that uses this feature, the failure seems to be caused by this AbstractAlgebra change (https://github.com/Nemocas/AbstractAlgebra.jl/pull/1282, @mgkurtz) which was included starting in version 0.28.4:
```
8d769fa9d3fe0477b8cd3f5629302a7cb9d9aeec is the first bad commit
commit 8d769fa9d3fe0477b8cd3f5629302a7cb9d9aeec
Author: Markus Kurtz <kurtz@mathematik.uni-kl.de>
Date:   Mon Mar 20 20:43:46 2023 +0100

    Move `polynomial_ring` code to abstract level (#1282)
```

Starting from that commit we get this failure in Oscar:
```julia
julia> using Oscar
 -----    -----    -----      -      -----   
[...]

PR #2160, opened 2023-03-29
Add back serialization for universal polynomials

SoongNoonien:
@fingolfin is this what you meant in #2158?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 
 [compat]
-AbstractAlgebra = "0.29"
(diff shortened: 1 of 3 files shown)


SoongNoonien:
I think this wont fix https://github.com/oscar-system/Oscar.jl/issues/2158. As the problem seems to be in another package.

SoongNoonien:
@HereAround I think the new title is not correct. I meant "readd" in terms of "to add again".

HereAround:
> @HereAround I think the new title is not correct. I meant "readd" in terms of "to add again".

Ohh. Ok. Sorry then.

fingolfin:
This is what I had in mind, yes, thank you. And of course the CI will fail here, as we did not yet fix the underlying issue. But once it is fixed, we can merge this again. By having the PR open now, we won't forget about it :-)

thofma:
I pushed the fix in AA and made a new release. Everything should be green now.

fingolfin (approved):

Issue #1929, opened 2023-02-13
Reduce runtimes for tests


HereAround:
By now, the CI runs for OSCAR are arguably pretty long. Here are the timings from the last run in https://github.com/oscar-system/Oscar.jl/pull/1909:

* [Bibtool test](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120006/jobs/7166788522): 28s.
* [Cleanup Doc Preview / cleanup-doc-preview](https://github.com/oscar-system/Oscar.jl/actions/runs/4144119591/jobs/7166787676): 8s.
* [Documentation / deploy_docs](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120005/jobs/7166788539): 30m.
* [Run tests / test (1.6, x64, ubuntu-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166788558): 73m.
* [Run tests / test (~1.7.0-0, x64, ubuntu-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166788692): 53m.
* [Run tests / test (~1.8.0-0, x64, ubuntu-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166788824): 48m.
* [Run tests / test (~1.9.0-0, x64, ubuntu-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166788973): 59m.
* [Run tests / test (nightly, x64, ubuntu-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166789116): 74m.
* [Run tests / test (1.6, x64, macOS-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166789300): 91m.
* [Run tests / test (nightly, x64, macOS-latest)](https://github.com/oscar-system/Oscar.jl/actions/runs/4144120007/jobs/7166789441): Cancelled after 120m.

As I see it, it would be great to reduce the runtimes (quicker to debug). Maybe we add this to the agenda on the [OSCAR developer meeting, February 21 - 23](https://oscar.computeralgebra.de/meetings/2023-02/)? (@fieker , @wdecker)

Here are a few thoughts after a brief conversation with @fieker:

### 1. Optimizing the tests:
[...]

PR #2167, opened 2023-03-30
Some renaming & avoid running doctests on two different CI jobs

fingolfin:


benlorenz:
This would remove the doctests from the coverage measurement, which might not even be that bad, as I would prefer to have everything covered by proper tests. But coverage will probably drop a bit, I am somewhat interested in how much.

And it would remove macOS from the doctests, maybe we can add macOS to the docs job here but run the deploy step only on ubuntu (currently it is in the same actions step as the build but maybe we can add a flag for that)?

thofma:
Yes, this is the question. Are you proposing to not count the doctests towards coverage @fingolfin? I know that some people only write doctests and no tests.

simonbrandhorst:
Why should doctests not be proper tests? I think they should count for coverage, e.g. for the show methods.

benlorenz:
> Why should doctests not be proper tests? I think they should count for coverage, e.g. for the show methods.

For the show methods it does make sense.
For other code having only doctests means that it is only run with julia 1.6 and not with any more recent version.

lgoettgens:
Just for reference, this has been suggested in https://github.com/oscar-system/Oscar.jl/issues/1929#issuecomment-1489284942.

fingolfin:
I did not realize that this removes doctests from coverage computations, so that was not an intended side effect.

I am relatively open either way, i.e.:
1. dropping the coverage intentionally
2. keeping it by not applying this PR
3. keeping it by applying this PR after augmnting it to enable coverage tracking for the documentation CI workflow

The argument by @benlorenz as to why 1. might even be a good thing has some merit. Unfortunately, I don't see how we could use doctest to get coverage "only for `show` methods".

On the other hand, in practice it is relatively rare for us to get failures that affects only some Julia versions; and when they happen, then usually it relates to our bridges to GAP/Polymake/Singular. I don't recall any issues of that kind that would have been found by doctests only (then again, such issues might already be there, and we don't know about them yet, precisely because we don't run doctests...)

In an ideal world, yes, we'd also test everything in a regular test world, but that seems like a lot of effort for a relatively small gain?
(Side note: I wish I had a way to specify `@test`s for a function right next to its code, instead of artificially placing it into some file in `test/`)

I'd be more interested to find out if it would be possible to have the doctests run in e.g. Julia 1.7, 1.8 and 1.9 simultaneously ... or maybe 1.8+1.9 ... or are there really differences between *each* Julia version in our printing? One would hope that some of these things might eventually stabilize...

PR #2169, opened 2023-03-31
Raise CI timeout from 120 to 150 minutes

fingolfin:
Of course we want to make our tests faster again, but in the meantime
it's a nuisance if CI tests randomly fail after being 95% complete. So
let's be pragmatic and raise this limit, at least for the time being


thofma:
Feels like the US debt ceiling.

simonbrandhorst (approved):

PR #2171, opened 2023-03-31
Run doctests in Julia >= 1.7

fingolfin:
It seems that with the minor adjustments in this PR, doctests pass
in Julia 1.7, 1.8 and 1.9 (but not anymore in 1.6).

If this works, and also with Julia nightly, that may be a good way
to address the problem that currently doctests are run only in Julia
1.6 and thus cannot catch regressions caused by new Julia versions.

The downside is that this means all CI tests will be even slower...

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -32,14 +32,19 @@ jobs:
           - x64
         os:
           - ubuntu-latest
+        doctest:
+          - 'no'
+          - 'only'
         include:
           # Add a few macOS jobs (not too many, the number we can run in parallel is limited)
           - julia-version: '1.8'
             julia-arch: x64
             os: macOS-latest
+            doctest: 'yes'
           - julia-version: 'nightly'
             julia-arch: x64
             os: macOS-latest
+            doctest: 'yes'
 
     steps:
       - uses: actions/checkout@v3
@@ -59,10 +64,23 @@ jobs:
         # restrict number of openMP threads on macOS due to oversubscription
         run: echo "OMP_NUM_THREADS=1" >> $GITHUB_ENV
       - name: "Run tests"
+        if: matrix.doctest != 'only'
         uses: julia-actions/julia-runtest@latest
         with:
+          annotate: ${{ matrix.julia-version == '1.8' }}
           coverage: ${{ matrix.julia-version == '1.8' }}
           depwarn: error
+      - name: "Run doctests"
+        if: matrix.doctest != 'no'
+        run: |
+          julia --project=docs --color=yes -e '
+            using Pkg
+            Pkg.develop(PackageSpec(path=pwd()))
+            Pkg.instantiate()
+            using Documenter
+            using Oscar
+            DocMeta.setdocmeta!(Oscar, :DocTestSetup, :(using Oscar); recursive = true)
+            doctest(Oscar)'
       - name: "Process code coverage"
         if: matrix.julia-version == '1.8'
         uses: julia-actions/julia-processcoverage@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,14 +27,11 @@ jobs:
           - '1.8'
           - '~1.9.0-0'
           - 'nightly'
-        julia-arch:
-          - x64
         os:
           - ubuntu-latest
         include:
           # Add a few macOS jobs (not too many, the number we can run in parallel is limited)
           - julia-version: 'nightly'
-            julia-arch: x64
             os: macOS-latest
 
     steps:
@@ -47,7 +44,6 @@ jobs:
         uses: julia-actions/setup-julia@v1
         with:
           version: ${{ matrix.julia-version }}
-          arch: ${{ matrix.julia-arch }}
       - name: "Build package"
         uses: julia-actions/julia-buildpkg@latest
       - name: "limit OpenMP threads"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,8 +24,8 @@ jobs:
       matrix:
         julia-version:
           - '1.6'
-          - '~1.7.0-0'
-          - '~1.8.0-0'
+          - '1.7'
+          - '1.8'
           - '~1.9.0-0'
           - 'nightly'
         julia-arch:
@@ -34,7 +34,7 @@ jobs:
           - ubuntu-latest
         include:
           # Add a few macOS jobs (not too many, the number we can run in parallel is limited)
-          - julia-version: '1.6'
+          - julia-version: '1.8'
             julia-arch: x64
             os: macOS-latest
           - julia-version: 'nightly'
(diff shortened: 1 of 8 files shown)


fingolfin:
Ah, my local tests only tested the OSCAR doctests. While the `Documentation` CI job also tests AbstractAlgebra doctests, and hence fails... I think that's actually an argument for *not* running doctests in the documentation CI job (we don't want to run AA tests there, this just causes us grief)

.github/workflows/CI.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: "Run tests"
         uses: julia-actions/julia-runtest@latest
         with:
-          coverage: ${{ matrix.julia-version == '1.6' }}
+          coverage: ${{ matrix.julia-version == '1.8' }}
           depwarn: error
       - name: "Process code coverage"
```
benlorenz, line 125:
```suggestion
      - name: "Process code coverage"
        if: matrix.julia-version == '1.8'
```
Can you add this to the upload step further down as well?
(This might avoid a bunch of the edits for the codecov comments that often initially claim coverage -75%)

lgoettgens:
Another upside of this would be the possibility to debug docs and doctests in a recent julia version, i.e. no need to switch sessions all the time, right? If yes, I am in favor of this!

.github/workflows/CI.yml
```diff
@@ -61,12 +61,14 @@ jobs:
...
+        if: matrix.julia-version == '1.8
```
benlorenz:
```suggestion
        if: matrix.julia-version == '1.8'
```

.github/workflows/CI.yml
```diff
@@ -61,12 +61,14 @@ jobs:
...
         uses: julia-actions/julia-processcoverage@v1
         with:
           directories: src,experimental
       - name: "Upload coverage data to Codecov"
+        if: matrix.julia-version == '1.8
```
benlorenz:
```suggestion
        if: matrix.julia-version == '1.8'
```

fingolfin:
Hmm, I think the dev docs might mention 1.6 as needed for docs, too. Will look when I am at a computer

fingolfin:
I'd be happy to go this way. Question is how much slower the tests really get by running the doctests.

benlorenz:
The doctests currently take between 20 and 30 minutes.

I would still prefer to create separate jobs for the doctests.
There was the issue with the coverage upload for the doctests but I think it should also be possible to just upload the coverage results from the doctest job independently and the results will be merged anyway?
This would probably speed up all the linux tests since there are usually some spare runners. For the macos tests it might slow it down slightly since we need to install Oscar each time but the runners are busy most of the time. (We could even do the split only for linux)

Otherwise I am also happy to go forward with this.

fingolfin:
> I would still prefer to create separate jobs for the doctests.

Ah, right, I forgot about that -- and I agree with it!

But I won't do it tonight ;-)

fingolfin:
(and yeah perhaps we don't split on macOS)

thofma:
Will the coverage be merged? If so, I am happy with splitting it off.

lgoettgens:
> Will the coverage be merged? If so, I am happy with splitting it off.

According to the [codecov docs](https://docs.codecov.com/docs/merging-reports), yes.

Additionally, there is a parameter that prevents the codecov comment until a certain number of uploads happened (see [here](https://docs.codecov.com/docs/notifications#preventing-notifications-until-after-n-builds). This would be very useful to not get notifications after the doctest action succeded, but only after the normal tests did as well.

thofma:
I guess this is only waiting for the separate jobs for the doctests?

lgoettgens:
Since we now have 3? (ubuntu-1.8-no, ubuntu-1.8-only, macos-1.8-yes) jobs uploading coverage to codecov, I would like to have
```
codecov:
    notify:
        after_n_builds: 3
```
in the `.codecov.yml` to only get the codecov comments, after all three jobs have finished. (see [codecov docu](https://docs.codecov.com/docs/notifications#preventing-notifications-until-after-n-builds))

.github/workflows/CI.yml
```diff
@@ -59,14 +64,29 @@ jobs:
...
+        if: matrix.doctest != 'only'
         uses: julia-actions/julia-runtest@latest
         with:
-          coverage: ${{ matrix.julia-version == '1.6' }}
+          annotate: ${{ matrix.julia-version == '1.8' }}
+          coverage: ${{ matrix.julia-version == '1.8' }}
           depwarn: error
+      - name: "Run doctests"
+        if: matrix.doctest != 'no'
+      - run: |
```
benlorenz:
```suggestion
        run: |
```

fingolfin:
This is in principle ready, but I am not sure whether we really need to / want to run the doctests in four Julia versions. Indeed, perhaps time to remove the 1.7 tests altogether? It makes sense to test...
- on 1.6 as that's LTS;
- on nightly so we get early warning about regression;
- on 1.8 as that's the current stable release
- on 1.9rc as that's the upcoming release.

But there seems little reason to test in 1.7.

Also once 1.9.0 is out, we might want to disable the 1.8 tests.

.github/workflows/CI.yml
```diff
@@ -61,12 +61,65 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version:
+          - '1.7'
+          - '1.8'
+          - '~1.9.0-0'
+          - 'nightly'
+        os:
+          - ubuntu-latest
```
benlorenz, line 81:
```suggestion
          - ubuntu-latest
        include:
          - julia-version: '1.8'
            os: macOS-latest
```
I would prefer to keep at least one macos doctest job, I think there is quite a bit of code that is only covered by doctests. Would this require anything other than this `include` section (and another of the very few workers ...)?

benlorenz (approved):
Apart from my comment regarding the macos doctests this is fine from my point of view, thanks for working on this!
And I agree that we can/should drop 1.7 from both tests and doctests.

.github/workflows/CI.yml
fingolfin: I think this patch is sufficient, we'll try. 

But using 50% more macOS workers is painful; I've had multiple occasions in the past weeks were macOS runners were still playing catch up with our various PRs and master updated in the evening after I was already at home :-(.

Originally I tried to do what we discussed above and the macOS runners just run both regular tests and doctests. This worked but I felt the result was simply to hard to understand, debug, and modify. So now I've split the job matrices for the two, and I think the result is quite a bit better, esp. since the job name now makes it clear whether regular tests or doctests are run.

Anyway, in this new setup, it wasn't too bad to make it a bit more ugly again and modify the new macOS doctes runner to also run regular tests... and then drop the "regular test runner" for mac.

Let's see if that works (it rather late, so I wouldn't be surprised if I screwed it up again *sigh*)

fingolfin:
@benlorenz looking OK to you? Then we could squash-merge it, and afterwards adjust the list of CI tests required by the branch protection rules.

benlorenz (approved):
Looks good, yes!

There was a crash in the doctests on nightly on the first try but it went through after a restart. The codecov notification count can be added in another PR.

Issue #2172, opened 2023-03-31
Free resolution broken in some cases


jankoboehm:
In the following code the `free_resolution` command gives an error.

```
R, (x,y,z) = polynomial_ring(QQ, ["x", "y", "z"])
Z = abelian_group(0)
F = free_module(R, 3)
G = free_module(R, 2)
V = [y*G[1], (x+y)*G[1]+y*G[2], z*G[2]]
a = hom(F, G, V)
Mk = kernel(a)
Mk1 = Mk[1]
fr = free_resolution(Mk1)
p = presentation(Mk1)
map(p,1)
```

This is due to the generators of Mk1 not having relations and the implementation just trying to compute on.

lgoettgens:
This has been resolved by #2173.

PR #2173, opened 2023-03-31
fix broken free resolution

jankoboehm:
Aims to fix free resolution command if there are no relations, issue #2172.


ederc (commented):
It would be good to have a test added for this situation, maybe just a stripped down version of #2172 .

jankoboehm:
Busy with that.

simonbrandhorst (approved):

PR #2177, opened 2023-03-31
Fix type stability for empty iterators

thofma:


thofma:
https://github.com/thofma/Hecke.jl/pull/1024 is failing because of empty iterators it seems. I hope this fixes it.

fingolfin (approved):

PR #2178, opened 2023-03-31
More cleanup for ProjectiveScheme

simonbrandhorst:
`graded_coordinate_ring` is now called `homogeneous_coordinate_ring` 
and some further renamings. Further cleanup will come in later in other pull requests.

### DIFF
--- a/experimental/Schemes/CoherentSheaves.jl
+++ b/experimental/Schemes/CoherentSheaves.jl
@@ -677,7 +677,7 @@ function twisting_sheaf(IP::ProjectiveScheme{<:Field}, d::Int)
 
   X = covered_scheme(IP)
   MD = IdDict{AbsSpec, ModuleFP}()
-  S = graded_coordinate_ring(IP)
+  S = homogeneous_coordinate_ring(IP)
   n = ngens(S)-1
   for i in 1:n+1
     U = affine_charts(X)[i]
@@ -763,7 +763,7 @@ end
   R = OO(X)
   P = base_ring(R)
   F = FreeMod(R, ["d$(x)" for x in symbols(P)])
-  rels, _ = sub(F, transpose(change_base_ring(R, jacobi_matrix(gens(modulus(R))))))
+  rels, _ = sub(F, transpose(change_base_ring(R, jacobi_matrix(base_ring(modulus(R)), gens(modulus(R))))))
   M, _ = quo(F, rels)
   return M
 end
@@ -772,7 +772,7 @@ end
   R = OO(X)
   P = base_ring(R)
   F = FreeMod(R, ["d$(x)" for x in symbols(P)])
-  rels, _ = sub(F, transpose(change_base_ring(R, jacobi_matrix(gens(modulus(R))))))
+  rels, _ = sub(F, transpose(change_base_ring(R, jacobi_matrix(base_ring(modulus(R)), gens(modulus(R))))))
   M, _ = quo(F, rels)
   return M
 end
@@ -1623,9 +1623,9 @@ function projectivization(E::AbsCoherentSheaf;
   # prepare for the projective glueings
   for (U, V) in keys(glueings(C))
     P = on_patches[U]
-    SP = graded_coordinate_ring(P)
+    SP = homogeneous_coordinate_ring(P)
     Q = on_patches[V]
-    SQ = graded_coordinate_ring(Q)
+    SQ = homogeneous_coordinate_ring(Q)
     G = C[U, V]
     UV, VU = glueing_domains(G)
     f, g = glueing_morphisms(G)
@@ -1659,8 +1659,8 @@ function projectivization(E::AbsCoherentSheaf;
     B = [coordinates(E(V, UV)(w)) for w in gens(E(V))]
     #A = [coordinates(OX(U, VU)(f), I(VU)) for f in gens(I(U))] # A[i][j] = aᵢⱼ
     #B = [coordinates(OX(V, UV)(g), I(UV)) for g in gens(I(V))] # B[j][i] = bⱼᵢ
-    SQVU = graded_coordinate_ring(QVU)
-    SPUV = graded_coordinate_ring(PUV)
+    SQVU = homogeneous_coordinate_ring(QVU)
+    SPUV = homogeneous_coordinate_ring(PUV)
     # the induced map is ℙ(UV) → ℙ(VU), tⱼ ↦ ∑ᵢ bⱼᵢ ⋅ sᵢ 
     # and ℙ(VU) → ℙ(UV), sᵢ ↦ ∑ⱼ aᵢⱼ ⋅ tⱼ 
     fup = ProjectiveSchemeMor(PUV, QVU, hom(SQVU, SPUV, pullback(f), [sum([B[j][i]*SPUV[i] for i in 1:ngens(SPUV)]) for j in 1:length(B)]))
--- a/src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
@@ -1,4 +1,4 @@
-export graded_coordinate_ring
+export homogeneous_coordinate_ring
 
 ########################################################################
 # Interface for abstract projective schemes                            #
@@ -20,13 +20,13 @@ function ambient_coordinate_ring(P::AbsProjectiveScheme)
 end
 
 @doc Markdown.doc"""
-    graded_coordinate_ring(P::AbsProjectiveScheme)
+    homogeneous_coordinate_ring(P::AbsProjectiveScheme)
 
 On a projective scheme ``P = Proj(S)`` for a standard 
 graded finitely generated algebra ``S`` this returns ``S``.
 """
-function graded_coordinate_ring(P::AbsProjectiveScheme)
-  return graded_coordinate_ring(underlying_scheme(P))
+function homogeneous_coordinate_ring(P::AbsProjectiveScheme)
+  return homogeneous_coordinate_ring(underlying_scheme(P))
 end
 
 @attr AbsSpec function base_scheme(P::AbsProjectiveScheme)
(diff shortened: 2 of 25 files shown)


src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
```diff
@@ -326,8 +289,8 @@ end
     
 Return a `CoveredScheme` ``X`` isomorphic to `P` with standard affine charts given by dehomogenization. 
 
-Use `dehomogenize(P, U)` with `U` one of the `affine_charts` of ``X`` to 
-obtain the dehomogenization map from the `graded_coordinate_ring` of `P` 
+Use [`dehomogenization_map(::ProjectiveScheme, ::AbsSpec)`](@ref) with `U` one of the `affine_charts` of ``X`` to
```
fingolfin: Maybe this will fix the doctest error?
```suggestion
Use [`dehomogenization_map(::AbsProjectiveScheme, ::AbsSpec)`](@ref) with `U` one of the `affine_charts` of ``X`` to
```

simonbrandhorst: Unfortunately not, and neither 5 other things I tried. Since this wasting too much of my time, I disable the reference.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -763,7 +763,7 @@ end
   R = OO(X)
   P = base_ring(R)
   F = FreeMod(R, ["d$(x)" for x in symbols(P)])
-  rels, _ = sub(F, transpose(change_base_ring(R, jacobi_matrix(gens(modulus(R))))))
+  rels, _ = sub(F, transpose(change_base_ring(R, jacobi_matrix(base_ring(modulus(R)), gens(modulus(R))))))
```
HechtiDerLachs, line 766: How come these changes were necessary? Did the syntax for `jacobi_matrix` change?

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
```diff
@@ -123,46 +123,55 @@ end
...
+### TODO: Replace by the map of generators.
+@doc Markdown.doc"""
+    homogeneous_coordinates_on_affine_cone(X::ProjectiveScheme)
+
+On ``X ⊂ ℙʳ_A`` this returns a vector with the homogeneous
 coordinates ``[s₀,…,sᵣ]`` as entries where each one of the 
 ``sᵢ`` is a function on the `affine cone` of ``X``.
 """
-function homogeneous_coordinates(P::ProjectiveScheme)
+function homogeneous_coordinates_on_affine_cone(P::ProjectiveScheme)
```
HechtiDerLachs, line 157: This function will most likely never be used in actual code: Its name alone fills one line. Consider use cases like `[ do_something(x) for x in homogeneous_coordinates_on_affine_cone(P)]`: It's a mess. But I don't have a better solution, either.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -1,48 +1,60 @@
-function dehomogenize(
+@doc Markdown.doc"""
+    dehomogenization_map(X::AbsProjectiveScheme, U::AbsSpec)
+
+Return the restriction morphism from the graded coordinate ring of ``X`` to `𝒪(U)`.
+"""
+function dehomogenization_map(
```
HechtiDerLachs, line 6: We still need to cache this. It's created all over again, each time the user asks for it. But we can't use `@attr` because of the two arguments.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -51,15 +63,15 @@ end
...
 ``P``, return a method ``h`` for the homogenization of elements 
-``a ∈ 𝒪(U)``. 
+``a ∈ 𝒪(U)``.
 
-That means ``h(a)`` returns a pair ``(p, q)`` representing a fraction 
-``p/q ∈ S`` of the `ambient_coordinate_ring` of ``P`` such 
-that ``a`` is the dehomogenization of ``p/q``.
+This means ``h(a)`` returns a fraction ``p/q`` of the
+`homogeneous_coordinate_ring` ``S`` of ``P`` such that ``a`` is the dehomogenization
+of ``p/q``.
```
HechtiDerLachs: I think this is now wrong after our last discussions yesterday, isn't it?

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -122,14 +130,14 @@ function homogenize(P::AbsProjectiveScheme{<:Any, <:MPolyDecRing}, U::AbsSpec)
     if deg_a > 0
       return (pp, qq*ss^deg_a)
     elseif deg_a <0
-      return (pp * ss^(-deg_a), qq)
+      return F(pp * ss^(-deg_a), qq)
```
HechtiDerLachs: This really looks like it needs cleanup/reversion.

src/Exports.jl
```diff
@@ -406,6 +406,7 @@ export degree
 export degrevlex
 export dehomogenization
 export dehomogenize
```
HechtiDerLachs: Remove the previous export?

src/Rings/mpoly.jl
```diff
@@ -742,15 +742,20 @@ function jacobi_ideal(f::MPolyRingElem)
...
 
 @doc Markdown.doc"""
-    jacobi_matrix(g::Vector{<:MPolyRingElem})
+    jacobi_matrix([R::MPolyRing,] g::Vector{<:MPolyRingElem})
 
-Given an array ``g=[f_1,...,f_m]`` of polynomials over the same base ring,
+Given an array ``g=[f_1,...,f_m]`` of polynomials over the same base ring `R`,
 this function returns the Jacobian matrix ``J=(\partial_{x_i}f_j)_{i,j}`` of ``g``.
 """
 function jacobi_matrix(g::Vector{<:MPolyRingElem})
```
HechtiDerLachs, line 750: The signature of the actual method did not change.

HechtiDerLachs (commented):
Mainly renaming. I left some comments where I think more polishing or cleaning is necessary.

experimental/Schemes/CoherentSheaves.jl
simonbrandhorst: I introduced them when trying to get projective space working without a modulus. 
Some 0 element vector got stuffed into `jacobi_matrix` and that caused an error.
But then I decided to keep it for another ticket. However the change there seemed sensible, so I kept it.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
simonbrandhorst: Yea, I agree. Since it is for internal use only,  we can name it whatever we/you like.

simonbrandhorst: On the other hand we are not using it either. Should we remove it?

src/Rings/mpoly.jl
simonbrandhorst: This should be the correct signature for optional arguments.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
simonbrandhorst: o.k. never mind we seem to use them.

src/Exports.jl
simonbrandhorst: The name is also used by others?  @lkastner  is that part of your interface?
```
  [1] dehomogenize(vec::AbstractVector) in Oscar at /home/simon/.julia/dev/Oscar/src/PolyhedralGeometry/helpers.jl:178
  [2] dehomogenize(mat::AbstractMatrix) in Oscar at /home/simon/.julia/dev/Oscar/src/PolyhedralGeometry/helpers.jl:179
```

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -7,13 +7,22 @@ function dehomogenization_map(
     X::AbsProjectiveScheme{<:Ring}, 
     U::AbsSpec
   )
+  if !isdefined(X, :dehomogenization_cache)
+    X.dehomogenization_cache = WeakKeyDict()
+  end
+  cache = X.dehomogenization_cache
+  if haskey(cache, U)
+    return cache[U]
```
HechtiDerLachs, line 15: Problem here: `WeakKeyDict` is using `==` for comparison of keys. We need to wait for the `WeakKeyIdDict`.

simonbrandhorst: o.k. now we check first that it `===` one of the charts. That should be good enough since none of the standard charts 
is `==`.  Alternatively we can just make it an `IdDict`

src/Exports.jl
benlorenz: These dehomogenize functions were not really meant to be part of the public interface (for polyhedral geometry).

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
HechtiDerLachs: I think we should make it an `IdDict` for the time being and leave a TODO to change it eventually. Everything else has proved to introduce hard-to-find bugs.

simonbrandhorst: o.k. Ready for merge from my side.

HechtiDerLachs (approved):

PR #2181, opened 2023-03-31
Adjusted FTheoryTools to new structure of experimental

lkastner:
While I did this, I noticed that the transition had an unexpected pitfall in the docs. It comes from when one migrates from the old structure to the new one and there still is a `docs/src/Experimental/PACKAGE_NAME` folder. The doc generating script tries to set a symlink but fails, so the docs files cannot be reached. Hence I tweaked the docs generating script to give a helpful warning. It looks like this now:
```
build_doc()
[...]
┌ Warning: Documentation for FTheoryTools errored with the following message:
│ >> Documentation file Experimental/FTheoryTools/introduction.md does not exist. <<
│ Possibly the symlink 
│ >> /home/lars/repos/Oscar.jl/docs/src/Experimental/FTheoryTools <<
│ could not be set, because the directory already exists. Please consider
│ removing it. The documentation of FTheoryTools will be empty until this is
│ resolved.
└ @ Main.BuildDoc ~/repos/Oscar.jl/docs/make_work.jl:73
```
What do you think? Would you prefer a hard error? Maybe there is a nice way to check whether this is the correct symlink? @fingolfin @benlorenz


fingolfin:
Which code sets that symlink, if any? That code could check if there is already a directory "in the way" and error out.

Also note that `islink` plus `readlink` can be used to verify that a given path is a symbolic link  with the correct target.

docs/make_work.jl
```diff
@@ -34,9 +34,14 @@ Base.walkdir(str::String) = Base.walkdir(str; follow_symlinks=true)
 
 # When we read a `doc.main` from an experimental package, we need to equip all
 # its entries with a prefix to fit with our docs.
-setup_experimental_prefix(docs::String, prefix::String) = prefix * docs
-setup_experimental_prefix(docs::Pair{String, Vector{T}}, prefix::String) where T = Pair{String, Vector{T}}(docs[1], setup_experimental_prefix(docs[2], prefix))
-setup_experimental_prefix(docs::Vector{T}, prefix::String) where T = T[setup_experimental_prefix(entry, prefix) for entry in docs]
+function setup_experimental_prefix(Oscar::Module, docs::String, prefix::String)
```
fingolfin: Meta comment: I find `setup_experimental_prefix` rather opaque. I have no idea what typical inputs might be and what it does, and the comment before is not really helping me either.

Perhaps we could have some example calls / outputs in a comment before it. Perhaps it would also help if this code was a single plain function instead of being "cleverly" spread over three methods.

docs/src/Experimental/intro.md
```diff
@@ -21,7 +21,7 @@ brought up to Oscar standard.
 
 ## Structure
 To look at a sample for the structure for a new project in `experimental` have
-a look at `experimental/PlaneCurve`. The general structure is
+a look at `experimental/{FTheoryTools, PlaneCurve}`. The general structure is
```
fingolfin: I would not change this. Or is the plan to keep adding each experiment that gets converted? That's not helpful.

If you want to be prepared for a future in which `experimental/PlaneCurve` is gone, one could describe it generically: "look at subdirectories of `experimental` which contain a `src` subdirectory" or so.

docs/make_work.jl
```diff
@@ -56,8 +61,20 @@ function setup_experimental_package(Oscar::Module, package_name::String)
...
+    return result
+  catch exception
+    msg = sprint(showerror, exception)
+    @warn("Documentation for $package_name errored with the following message:
+>> $msg <<
+Possibly the symlink 
+>> $symlink_src <<
+could not be set, because the directory already exists. Please consider
+removing it. The documentation of $package_name will be empty until this is
+resolved.")
```
fingolfin: How about instead something like this in the code which creates the symlink right now:
```julia
  symlink_link = joinpath(Oscar.oscardir, "docs", "src", "Experimental", package_name)
  symlink_target = "../../../experimental/$package_name/docs/src"
  if ispath(symlink_link)
    if !islink(symlink_link) || readlink(symlink_link) != symlink_target
      error("TODO")
    end
  else
    symlink(symlink_target, symlink_link)
  end
```

docs/src/Experimental/intro.md
lkastner: The plan was just to provide more than one example for now, especially since FTheoryTools is a little bit more involved. I plan to remove these specific examples once these procedures have been established.

docs/make_work.jl
lkastner: I added some examples what these methods accomplish together. I could also turn it into one function, but I would make it recursive (which it also is now) and it would use a bunch of `typeof` calls. Once you look at the examples, let me know what you think. Maybe these are already fine as an explanation.

docs/make_work.jl
lkastner: Thanks, I adjusted accordingly. The error message now looks like:
```
ERROR: /home/lars/repos/Oscar.jl/docs/src/Experimental/FTheoryTools already exists, but is not a symlink to ../../../experimental/FTheoryTools/docs/src
Please investigate the contents of /home/lars/repos/Oscar.jl/docs/src/Experimental/FTheoryTools,
optionally move them somewhere else and delete the directiory once you are done.
```
What do you think? A warning would be softer and make sure things succeed, but on the other hand many users would probably be confused why their docs don't show up, so a hard error might be the way to go.

.gitignore
```diff
@@ -2,6 +2,6 @@
 /docs/src/AbstractAlgebra/
 /docs/src/Hecke/
 /docs/src/Nemo/
-/docs/src/experimental/
+/docs/src/Experimental/
```
lgoettgens, line 5: This is a pitfall for everyone still using the "old" experimental structure. They probably miss adding their docs files to git due to this. Unfortunately, I have no good alternative (apart from everyone re-structuring their experimental code).

lkastner: It is only until the old stuff is gone and it is detected by github actions.

test/runtests.jl
```diff
@@ -112,7 +112,6 @@ include("Experimental/JuLie-test.jl")
 include("Experimental/SymmetricIntersections-test.jl")
 include("Experimental/ExteriorAlgebra-test.jl")
 include("Experimental/LinearQuotients/runtests.jl")
-include("Experimental/FTheoryTools/runtests.jl")
```
HereAround, line 115: How will the tests for `FTheoryTools` be executed in the future? Is this in preparation of a possible split of the CI-jobs into `core` and `experimental`?

HereAround (approved):
Thank you very much for working on this @lkastner . ~~I left one minor question above. Other than that I am very happy with this PR!~~ Thank you.

test/runtests.jl
lgoettgens: [`test/runtests.jl`](https://github.com/oscar-system/Oscar.jl/blob/6cf96ad70320645093eacdcd8ae9d1cef22cf7e4/test/runtests.jl#L101-L103) calls [`experimental/runtests.jl`](https://github.com/oscar-system/Oscar.jl/blob/6cf96ad70320645093eacdcd8ae9d1cef22cf7e4/experimental/runtests.jl) which in turn calls the testfile of FTheoryTools based on some regex.

PR #2183, opened 2023-04-01
Add basics for quantum analogs

ulthiel:
Added basic functions for quantum analogs that I had implemented in my JuLie package: quantum integers, factorials, and binomials. They work also with specializations in any (commutative) ring.


experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+################################################################################
+# Quantum integers
+################################################################################
+@doc Markdown.doc"""
+    quantum_integer(n::IntegerUnion, q::RingElem)
+    quantum_integer(n::IntegerUnion, q::Integer)
+    quantum_integer(n::IntegerUnion)
+    
+Let ``n ∈ ℤ`` and let ``ℚ(𝐪)`` be the fraction field of the polynomial ring ``ℤ[𝐪]`` in
+one variable ``𝐪``. The **quantum integer** ``[n]_𝐪 ∈ ℚ(𝐪)`` of ``n`` is defined as
```
fingolfin: But `quantum_integer(5,1)` is not in ℚ(𝐪) ... Also this is quite technical. Can't we just write this:
```suggestion
For an integer `n` and a ring element `q`, the **quantum integer** ``[n]_q`` is defined as
```

fingolfin: In general I would suggest to minimize the use of Unicode. Most of it seems not necessary here. E.g. why does `q` have to be in (faux) bold face?

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+[n]_𝐪 = \sum_{i=0}^{n-1} 𝐪^i ∈ ℤ[𝐪] \quad \text{if } n ≥ 0
+```
+and
+```math
+[n]_𝐪 = -𝐪^{n} [-n]_𝐪 \quad \text{for any } n ∈ ℤ \;,
+```
+hence
+```math
+[n]_𝐪 = - \sum_{i=0}^{-n-1} 𝐪^{n+i} ∈ ℤ[𝐪^{-1}] \quad \text{ if } n < 0 \;.
+```
```
fingolfin: For a mathematical text, it is cool to derive these properties and use them to justify the definition. I think this text would fit perfectly into `quantum_analogs.md`.

But for a docstring, I'd keep it much simpler and to the point, and I would also avoid excessive use of LaTeX and math formulas which are hard to read in a REPL anyway.

So how about just stating the concrete definition, and point to the manual for the justification? E.g. like this
````suggestion
``[n]_q = 1 + q + q^2 + ... + q^{n-1}`` if ``n \geq 0`, in particular ``[0]_q = 1``. For negative `n` we use the identity
```math
[n]_q = -q^n [-n]_q
```
which requires that `q` is invertible, and thus obtain
```math
[n]_q = - \sum_{i=0}^{-n-1} q^{n+i} ∈ ℤ[q^{-1}] .
```
````

File: experimental/JuLie/quantum_analogs.jl [hunk 3]
```diff
@@ -0,0 +1,295 @@
...
+```math
+[n]_𝐪 ∈ ℤ[𝐪,𝐪^{-1}] ⊂ ℚ(𝐪) \quad \text{ for any } n ∈ ℤ \;.
+```
+Now, for an element ``q`` of a ring ``R`` we define ``[n]_q ∈ R`` as the specialization of
+``[n]_𝐪`` in ``q`` using the two equations above—assuming that ``q`` is invertible in ``R``
+if ``n<0``. Note that for ``q=1`` we obtain
+```math
+[n]_1 = n \quad \text{for for any } n ∈ ℤ \;,
+```
+so the quantum integers are "deformations" of the usual integers.
```
fingolfin: All of this sounds OK to me for a description in the reference manual, but isn't it a bit overkill for the docstring? Esp. since it is kinda hard to read this in a terminal?

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+
+# Functions
+* `quantum_integer(n::IntegerUnion,q::RingElem)` returns ``[n]_q`` as an element of ``R``,
+  where ``R`` is the parent ring of ``q``. 
+* `quantum_integer(n::IntegerUnion,q::Integer)` returns ``[n]_q``. Here, if ``n >= 0`` or
+  ``q = ± 1``, then ``q`` is considered as an element of ``ℤ``, otherwise it is taken as an
+  element of ``ℚ``.
```
fingolfin: So then it is not type stable.

How about instead negative `n` leads to an error if `q` is an integer; if a user wants that, they need to to call e.g. `quantum_integer(-5, QQ(3))` ?

But perhaps this case is needed so frequently that the inconvenience outweighs the type instability?

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+			for i = 0:n-1
+				z += q^i
+			end
+		else
+			for i = 0:-n-1
+				z -= q^(n+i)
+			end
+		end
+		return z
+	end
```
fingolfin: Alternative implementation:
```suggestion
    R = parent(q)
    isone(q) && return R(n)
    n >= 0 && return sum(i -> q^i, 0:n-1)
    return -sum(i -> q^(n+i), 0:-n-1)
```
Of course this leaves potential for optimization, because we compute 1, q, q^2, q^3, ... "naively", i.e. to compute, say, q^40, we don't use that we already computed q^39 before! I am not sure if we already have a function doing that "neatly"

Oh, and if `q` is an integer, and we care about efficiency, then I think the last line would be better off if it used recursion (i.e.: first sum powers of an integer, then perform a single division at the end; instead of summing powers of a fraction). So:
```suggestion
    R = parent(q)
    isone(q) && return R(n)
    n < 0 && return -q^n * quantum_integer(-n, q)
    return sum(i -> q^i, 0:n-1)
```

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+# Quantum binomials
+################################################################################
+
+@doc Markdown.doc"""
+    quantum_binomial(n::IntegerUnion, q::RingElem)
+    quantum_binomial(n::IntegerUnion, k::IntegerUnion, q::Integer)
+    quantum_binomial(n::IntegerUnion, k::IntegerUnion)
+
+Let ``k`` be a non-negative integer and let ``n ∈ ℤ``. The **quantum binomial**
+``\begin{bmatrix} n \\ k \end{bmatrix}_𝐪 \in ℚ(𝐪)`` is defined as
```
fingolfin: Perhaps use `\binom`, it is standard and the resulting TeX is more readable in the REPL. Also of course my comments from above also apply here (so I'd replace `𝐪` by `q`)
```suggestion
``\binom{n}{k}_𝐪 \in ℚ(𝐪)`` is defined as
```

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+```math
+\begin{bmatrix} n \\ k \end{bmatrix}_𝐪 ≔ \frac{[n]_𝐪!}{[k]_𝐪! [n-k]_𝐪!} = \frac{[n]_𝐪 [n-1]_𝐪⋅ … ⋅ [n-k+1]_𝐪}{[k]_𝐪!}
+```
+Note that the first expression is only defined for ``n ≥ k`` since the quantum factorials
+are only defined for non-negative integers—but the second  expression is well-defined for
```
fingolfin:
```suggestion
are only defined for non-negative integers, but the second expression is well-defined for
```

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+Now, for an element ``q`` of a ring ``R`` we define ``\begin{bmatrix} n \\ k
+\end{bmatrix}_q`` as the specialization of ``\begin{bmatrix} n \\ k
+\end{bmatrix}_{\mathbf{q}}`` in ``q``, where ``q`` is assumed to be invertible in ``R`` if
+``n < 0``.
+
+Note that for ``q=1`` we obtain
+```math
+\begin{bmatrix} n \\ k \end{bmatrix}_1 = {n \choose k} \;,
+```
+hence the quantum binomial coefficient is a "deformation" of the usual binomial coefficient.
```
fingolfin: All of this IMHO belongs into the `.md` file but not the docstring.

test/Experimental/JuLie-test.jl
```diff
@@ -1,3 +1,4 @@
 include("JuLie/tableaux.jl")
 include("JuLie/schur_polynomials.jl")
 include("JuLie/partitions.jl")
+include("JuLie/quantum_analogs.jl")
```
fingolfin, line 4: please always end files with a trailing new line (and ideally configure your editor to do that automatically, e.g. by making it honor our `.editoconfig` file, as described in the dev docs -- assuming you are using an editor with that capability)
```suggestion
include("JuLie/quantum_analogs.jl")

```

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+
+# Functions
+
+The functions `quantum_binomial` work analogously to [`quantum_integer`](@ref).
+
```
fingolfin: This section doesn't really add anything, esp. if the text before it were shortened to just describe what the function does?
```suggestion
```

experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,295 @@
...
+julia> quantum_binomial(17,10,i)
+0
+```
+
+# References
+1. [Con00](@cite)
+"""
+function quantum_binomial(n::IntegerUnion, k::IntegerUnion, q::RingElem)
+
+	k >= 0 || throw(ArgumentError("k ≥ 0 required"))
```
fingolfin, line 259:
```suggestion

	@req k >= 0 "k >= 0 required"
```

test/Experimental/JuLie/quantum_analogs.jl
```diff
@@ -0,0 +1,70 @@
...
+	@test quantum_binomial(0,0) == 1
+	@test quantum_binomial(1,0) == 1
+	@test quantum_binomial(1,1) == 1
+	@test quantum_binomial(2,1) == 1+q
+	@test quantum_binomial(3,1) == 1+q+q^2
+	@test quantum_binomial(3,2) == 1+q+q^2
+	@test quantum_binomial(4,2) == (1+q^2)*(1+q+q^2)
+	@test quantum_binomial(6,3) == (1+q^2)*(1+q^3)*(1+q+q^2+q^3+q^4)
+
+end
```
fingolfin, line 70:
```suggestion
end

```

fingolfin (commented):
Thank you, this looks really good!

I still have a bunch of remarks, but most of these are on the cosmetically level, or optional, or both.

ulthiel:
Thanks, @fingolfin. I had a discussion with @fieker: the function quantum_integer(n) may be problematic as is minimal_polynomial without passing the ring. He also suggested to implement a structure for quantum integers (so, basically the [ring of quantum integers](https://arxiv.org/pdf/math/0204006.pdf)). I think this would be the way to go. I never thought about that before. That's why I changed the PR to draft mode.

thofma:
Just wanted to say that I was in need of quantum binomial coefficients today and the changes here would have come in handy.

ulthiel:
> Just wanted to say that I was in need of quantum binomial coefficients today and the changes here would have come in handy.

Will do the update in the next couple of weeks. (I personally decided against implementing a dedicated ring as mentioned above for now, I hope you didn't need that particular feature...).

fingolfin:
@ulthiel any news on this?

ulthiel:
> @ulthiel any news on this?

@FabianMaeurer will take this up.

fingolfin:
Any news on this? Does this PR need to stay open for the work planned by @FabianMaeurer ?

ulthiel:
If Claus would not have asked this question on Slack, this would have been finished already... :-)

I could offer to finish as planned by myself originally, i.e. ignoring Claus but at least addressing the comments above.

fingolfin:
@ulthiel what would you prefer? have this as-is (modulo some tweaks), so that you can just use it? Or would you prefer a rewrite? Or perhaps you don't care either way because you don't use it?

ulthiel:
Have it as is, because that's exactly why I've written it. :-) Improvements welcome of course.

fingolfin:
@ulthiel then by all means, do it! I was ready to merge your PR (modulo very minor nitpicks) a year ago, and always assumed you stopped updating it because *you* wanted to make fundamental changes :-)

fingolfin:
@ulthiel let's talk one of these days about getting this ready, perhaps someone else can help finish it?

fingolfin:
I've discussed this PR this morning with @ulthiel and based on that I've now rebased it -- doing this I noticed that a bit more work was needed than I expected, and as a result I've now moved it into a new `experimental/QuantumAnalogs/` directory. But the final place for this probably should be in `src/Combinatorics`.

Unfortunately this move resulted in the existing comments on this file to loose their place, i.e. they are no longer attached to the relevant code :-(. Since that was the case now anyway, I've decided to also perform the changes from tabs to spaces (I didn't do those at first in the hopes that the code review comments would stay attached). I also performed some other "obvious" tweaks.

But I did not e.g. touch the unicode, or move anything to the `.md` file. @ulthiel I am happy to help with that, too, but it's the kind of change I wouldn't want to make w/o coordinating with you.

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,295 @@
...
+    return R(n)
+  else
+    z = zero(R)
+    if n >= 0
+      for i = 0:n-1
+        z += q^i
+      end
+    else
+      for i = 0:-n-1
+        z -= q^(n+i)
```
fingolfin:
```suggestion
      for i = n:-1
        z -= q^i
```

fingolfin: OK while this is IMHO slghtly easier for understanding what the code does, the current code matches the explanation in the docstring, so better to leave it the way it is.

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,295 @@
...
+    end
+    return z
+  end
+end
+
+function quantum_integer(n::IntegerUnion, q::Integer)
+  if n >= 0 || q == 1 || q == -1
+    return quantum_integer(n,ZZ(q))
+  else
+    return quantum_integer(n,QQ(q))
```
fingolfin: This function is type unstable

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,295 @@
...
+  end
+
+end
+
+function quantum_binomial(n::IntegerUnion, k::IntegerUnion, q::Integer)
+  if n > 0
+    return quantum_binomial(n,k,ZZ(q))
+  else
+    return quantum_binomial(n,k,QQ(q))
+  end
```
fingolfin: Also type unstable

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,279 @@
...
+    quantum_binomial(n::IntegerUnion, k::IntegerUnion, q::Integer)
+    quantum_binomial(n::IntegerUnion, k::IntegerUnion)
+
+Let ``k`` be a non-negative integer and let ``n ∈ ℤ``. The **quantum binomial**
+``\begin{bmatrix} n \\ k \end{bmatrix}_𝐪 \in ℚ(𝐪)`` is defined as
+```math
+\begin{bmatrix} n \\ k \end{bmatrix}_𝐪 ≔ \frac{[n]_𝐪!}{[k]_𝐪! [n-k]_𝐪!} = \frac{[n]_𝐪 [n-1]_𝐪⋅ … ⋅ [n-k+1]_𝐪}{[k]_𝐪!}
+```
+Note that the first expression is only defined for ``n ≥ k`` since the quantum factorials
+are only defined for non-negative integers—but the second  expression is well-defined for
```
fingolfin:
```suggestion
are only defined for non-negative integers, but the second expression is well-defined for
```

docs/oscar_references.bib
```diff
@@ -1731,6 +1745,18 @@ @Article{Jow11
...
+@Book{KC02,
+  author        = {Kac, Victor and Cheung, Pokman},
+  title         = {Quantum calculus},
+  mrnumber      = {1865777},
+  series        = {Universitext},
+  publisher     = {Springer-Verlag, New York},
+  pages         = {x+112},
+  year          = {2002},
+  doi           = {10.1007/978-1-4613-0071-7},
+  url           = {https://doi.org/10.1007/978-1-4613-0071-7}
```
fingolfin:
```suggestion
```

docs/oscar_references.bib
```diff
@@ -719,6 +719,20 @@ @Book{Coh93
...
+  title         = {A {$q$}-analogue of {M}ahler expansions. {I}},
+  mrnumber      = {1770929},
+  journal       = {Adv. Math.},
+  fjournal      = {Advances in Mathematics},
+  volume        = {153},
+  number        = {2},
+  pages         = {185--230},
+  year          = {2000},
+  doi           = {10.1006/aima.1999.1890},
+  url           = {https://doi.org/10.1006/aima.1999.1890}
```
fingolfin:
```suggestion
```

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,279 @@
...
+  else
+    for i = 0:-n-1
+      z -= q^(n+i)
+    end
+  end
+  return z
+end
+
+function quantum_integer(n::IntegerUnion, q::Integer)
+  if n >= 0 || q == 1 || q == -1
```
fingolfin:
```suggestion
  # this method is not type stable, but that is acceptable as it is only
  # intended for convenience during interactive use
  if n >= 0 || q == 1 || q == -1
```

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,279 @@
...
+
+  z = zero(R)
+  for i = 0:n-k
+    z += q^i * quantum_binomial(i+k-1,k-1,q)
+  end
+  return z
+end
+
+function quantum_binomial(n::IntegerUnion, k::IntegerUnion, q::Integer)
+  if n > 0
```
fingolfin:
```suggestion
  # this method is not type stable, but that is acceptable as it is only
  # intended for convenience during interactive use
  if n > 0
```

fingolfin:
I've revised the PR now, and from my point of view this is ready for review and merging. @ulthiel I'll remove the "Draft" flag now, based on our discussion last week. Please holler if you have concerns.

docs/oscar_references.bib
```diff
@@ -719,6 +719,19 @@ @Book{Coh93
   doi           = {10.1007/978-3-662-02945-9}
 }
 
+@Article{Con00,
+  author        = {Conrad, Keith},
+  title         = {A {$q$}-analogue of {M}ahler expansions. {I}},
```
lgoettgens, line 736:
```suggestion
  title         = {A $q$-analogue of Mahler expansions. I},
```

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,206 @@
...
+  for i = 0:n-1
+    # at this point qi = q^i
+    z = add!(z, qi)
+    qi = mul!(qi, q)
+  end
+  return z
+end
+
+function quantum_integer(n::IntegerUnion)
+  R,q = laurent_polynomial_ring(ZZ, "q")
```
lgoettgens:
```suggestion
  R,q = laurent_polynomial_ring(ZZ, "q"; cached=false)
```
This shouldn't cache the poly ring by our style guide. However, for interactive use, it is incredibly unuseful if this creates new Laurent rings every time you call this function, as the results are not compatible

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,206 @@
...
+
+  z = one(R)
+  for i = 1:n
+    z *= quantum_integer(i,q)
+  end
+  return z
+end
+
+function quantum_factorial(n::IntegerUnion)
+  R,q = laurent_polynomial_ring(ZZ, "q")
```
lgoettgens: as above

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,206 @@
...
+
+  z = zero(R)
+  for i = 0:n-k
+    z += q^i * quantum_binomial(i+k-1,k-1,q)
+  end
+  return z
+end
+
+function quantum_binomial(n::IntegerUnion, k::IntegerUnion)
+  R,q = laurent_polynomial_ring(ZZ, "q")
```
lgoettgens: as above

experimental/QuantumAnalogs/test/quantum_analogs.jl
```diff
@@ -0,0 +1,70 @@
+@testset "combinatorics/quantum_analogs.jl" begin
+  R,q = laurent_polynomial_ring(ZZ, "q")
+
+	# Test if everything specializes to usual stuff at q=1
```
lgoettgens:
```suggestion
  # Test if everything specializes to usual stuff at q=1
```
there are some tabs left in this file

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
thofma: Related to this is the introduction of a parent (quantum integer ring). This would in my opinion be the cleaner solution, but means more work.

fingolfin: I think the idea of a "parent quantum integer ring" sounds intriguing at first but in practice it wasn't so clear how it should work and who would be using it. I'd rather net let this PR be delayed over this.

fingolfin: And yeah I deliberately did not add the `cached=false` because it seems extremely unhelpful for its intended use.

fingolfin (approved):
I am happy with this but I also rewrote quite some bits of it, so ideally this is also reviewed and approved by others.

Also perhaps @ulthiel could briefly comment to say if he is OK with all this.

ulthiel:
Yes, very good, thanks! The tests (from Conrad etc.) still pass, so this should be ok then. @thofma The quantum integer ring idea is what stalled this PR (good idea, but someone needs to do it, and everything was basically working already...).

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,208 @@
...
+"""
+function quantum_factorial(n::IntegerUnion, q::RingElement)
+  @req n >= 0 "n >= 0 required"
+
+  R = parent(q)
+  isone(q) && return R(factorial(n))
+
+  z = one(R)
+  for i = 1:n
+    z *= quantum_integer(i,q)
```
lgoettgens:
```suggestion
    z = mul!(z, quantum_integer(i,q))
```

experimental/QuantumAnalogs/src/QuantumAnalogs.jl
```diff
@@ -0,0 +1,208 @@
...
+  R = parent(q)
+  isone(q) && return R(binomial(n,k))
+  k == 0 && return one(R)
+  k == 1 && return quantum_integer(n,q)
+  n < 0 && return (-1)^k * q^(div(-k*(k-1),2) + k*n) * quantum_binomial(k-n-1,k,q)
+  n < k && return zero(R)
+
+  z = zero(R)
+  for i = 0:n-k
+    z += q^i * quantum_binomial(i+k-1,k-1,q)
```
lgoettgens:
```suggestion
    z = addmul!(z, q^i, quantum_binomial(i+k-1,k-1,q))
```

lgoettgens:
@felix-roehrich has some objections to this and will write them here in his own comment and will join Triage tomorrow

felix-roehrich:
The definitions here for quantum integers is not consistent with the literature I know. For me a quantum integer is (if `n >= 0`)
```
[n]_q = q^(-n+1) + ... + q^(n-1) =  (q^n - q^-n)/(q-q^-1)
```
Accordingly, you get different factorials and also binomials.

The QuaGroup package currently uses quantum integers as I defined here, though there it is called Gaussian integer/factorial /binomial.

From I a bit of further search, there seem to be 3 popular variants:
```
[n]_q = (q^n - q^-n)/(q - q^-1)
{n}_q = (q^n-1)/(q-1)
<n>_q = (q^(n/2)-q^(-n/2))(q^(1/2)-q^(1/2))
```

fingolfin:
@felix-roehrich thank you for pointing this out. Many objects we use have multiple non-equivalent definitions. Obviously we can pick at most one in each case. Our goal is to follow a specific text book or paper so at least we have some internal consistency and users have a precise reference to follow. E.g. this PR provides two such references.

Can you point us to a book or two using "your" definition? That would make an evaluation easier.

BTW you state that QuaGroup does not even use the term "quantum integer" which of course somewhat weakens your position (and "Gaussian integer" also has a different definition which arguably is much more well-known).

fingolfin:
Some quick research:
- [Wikipedia](https://en.wikipedia.org/wiki/Q-analog) calls "our" values here "classical q-analogs" and the "quantum integer" in particular a "q-number" or "q-bracket", and then "q-factorial" and "q-binomial". The latter is also called "Gaussian binomial (coefficient)", see also <https://en.wikipedia.org/wiki/Gaussian_binomial_coefficient>
- MathWorld agrees with the definitions for [q-analogs](https://mathworld.wolfram.com/q-Analog.html), in particular the names [q-bracket](https://mathworld.wolfram.com/q-Bracket.html), [q-factorial](https://mathworld.wolfram.com/q-Factorial.html), [q-binomial coefficient](https://mathworld.wolfram.com/q-BinomialCoefficient.html)
- SageMath calls these [$q$-analogues](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/q_analogues.html#module-sage.combinat.q_analogues), and has similar names, see e.g. [`sage.combinat.q_analogues.q_binomial`](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/q_analogues.html#sage.combinat.q_analogues.q_binomial)
- another paper that uses "our" definitions is [this one](https://hal.science/hal-00875306v2/document)

All of the above "agree" with our definitions, but most suggest different names for the functions.

But then there is also this:
- SageMath *also* has a seperate page for ["$q$-numbers"](https://doc.sagemath.org/html/en/reference/algebras/sage/algebras/quantum_groups/q_numbers.html#sage.algebras.quantum_groups.q_numbers.q_factorial), and e.g. [`sage.algebras.quantum_groups.q_numbers.q_int`](https://doc.sagemath.org/html/en/reference/algebras/sage/algebras/quantum_groups/q_numbers.html#sage.algebras.quantum_groups.q_numbers.q_int) uses the definition @felix-roehrich  prefers(?), and then [`sage.algebras.quantum_groups.q_numbers.q_binomial`](https://doc.sagemath.org/html/en/reference/algebras/sage/algebras/quantum_groups/q_numbers.html#sage.algebras.quantum_groups.q_numbers.q_binomial) uses that -- but it also comes with a warning:
  > This is not the “usual” $q$-binomial but a variant useful for quantum groups. For the version used in combinatorics, see [`sage.combinat.q_analogues`](https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/q_analogues.html#module-sage.combinat.q_analogues).
- Magma seems to call these values "Gaussian" number/factorial/binomial, just like QuaGroup, see [here](https://magma.maths.usyd.edu.au/magma/handbook/text/1216)
- I did not find a reference to "quantum binomial" or "q-binomial" in the Magma manual

felix-roehrich:
References:
Jantzen, Lectures on Quantum Groups
Lusztig, Introduction to Quantum Groups

Both use the definition I have given here.

> BTW you state that QuaGroup does not even use the term "quantum integer" which of course somewhat weakens your position (and "Gaussian integer" also has a different definition which arguably is much more well-known).

You said yourself that Gaussian integer has already a different meaning and also that Gaussian binomial is also used for "your" definition, so I don't think that weakens my point. The main reason to point this out is, that this will (probably) have an impact on further constructions like quantum groups, where we have already different q-analogues than the one in this PR.

micjoswig:
I think this code is very useful.  And I also think that it is essentially fine; all the remaining issues can be solved via adjusting the docs IMHO.  Of course, we need to state clearly which convention we are following.  It does not hurt to add a word of warning that other conventions may exist.

fingolfin:
I discuss this some more this afternoon with @felix-roehrich 

Of course in the end we can pick one definition, and use that and document it as such.

But then the other definition is also useful, and we'd also like to make it available, and also need to use names for it...

SageMath solves this by having e.g. two `q_binomial` functions, inside different namespaces / modules. We could do that (e.g. `Combinatorics.q_binomial` and `QuantumGroups.q_binomial`) but I am not a fan of that.

We could have `q_binomial` and `q_binomial2` -- works but also not nice.

(BTW, I keep writing `q_binomial` because its shorter and what Sage and Mathematic seem to use but I am happy to stick with `quantum_binomial` -- I just don't want to write both each time)

What makes more sense to me is to allow specifying a convention via an additional
argument. This argument can be made optional, with a default value of our choice.

So, we could have something like this:

```
q_binomial(5, 3, q)          # work like it does now
q_binomial(:combi, 5, 3, q)  # same
q_binomial(:group, 5, 3, q)  # use quantum group convention
```

---

We could go one step further, and pick up the idea of having a "quantum number ring" -- or rather, have two of them. Some pseudo code:

```
mutable struct QuantumNumberRing{Kind} <: Ring
  R::LaurentPolynomialRing 
  # alternatively these could also be singleton types,
  # with a globally shared LaurentPolynomialRing instances,
  # but that's an implementation detail
end

# The following names are already optimal. Nobody can come up with better
# names. Don't even try. This paragraph may contain traces of sarcasm.
const QuantumCombinatoricsRing = QuantumNumberRing{:combi}
const QuantumGroupsRing = QuantumNumberRing{:group}

# now define +, *, - etc. by passing them through
...
```
and then finally, we have

```
function q_number(::QuantumCombinatoricsRing, v::Integer, q::RingElement)
  ... return as in current code in this PR ...
end

function q_number(::QuantumGroupsRing, v::Integer, q::RingElement)
   ... return as in the quantum group definition ...
end
```


To use this, you have two options: use the default values, or use
an explicit quantum (number) ring. This returns the ring and
the generator:

```
R, q = quantum_number_ring(:group)
#R, q = quantum_number_ring(:combi)
```
In many ways this would behave like a Laurent polynomial ring

The higher functions like `q_binomial` can be used like this:
```
# allow omitting `R`
q_binomial(n, k, q::QuantumNumberElem) = quantum_binomial(parent(q), n, k, q)

# possibly specify default convention
q_binomial(n, k, q::RingElement) = quantum_binomial(quantum_number_ring(:combi), n, k, q)

# allow omitting `q`
q_binomial(R::QuantumNumberRing, n, k) = quantum_binomial(R, n, k, gen(R))
```

Then one can do all of these
```
R, q = quantum_number_ring(:group)
q_binomial(5, 3, q)
q_binomial(5, 3, q^2)
```

---


But after writing the above, I am not sure I see what this would add over just using
a plain Laurent polynomial ring, plus an optional specifier for `q_binomial`
to indicate which variant we want.

One feature for down the line @felix-roehrich expressed interested in is taking a
Laurent ring element and expressing it as a quantum binomial/factorial/integer,
or possibly a sum/product/etc. therefore. But this could also be done by a helper
function without a dedicated ring, I think..


Hmmm.

ulthiel:
Hm, so first of all, I know the books by Jantzen and Lusztig of course. I think, at least in oral discussions, I always pointed out that there are different definitions of quantum integers and one needs to be very precise about definitions (also about extensions to negatives etc.). I think at least in my original PR I had a lengthy documentation writing out all details. Especially, I gave a reference to the book "Quantum Calculus" by Kac & Cheung (2002). You can see in the first page of the introduction the definition I am using (factorials and binomials come in the subsequent chapters). 

The quantum integers you mention (and the ones from Jantzen and Lusztig) are what I would call "symmetric quantum integers" in line with Chapter 26 in loc. cit. ("Symmetric quantum calculus"). Originally, I planned to implement the symmetric versions as well but then the "quantum integer ring" discussion started and I did not follow up.

So, my suggestion would be to implement the symmetric versions as well. Both versions are useful. In the end it's a matter of taste: you can also define quantum_integer to be the symmetric ones and have asymmetric_quantum_integer with the other ones.

But you cannot say that what I have implemented is not consistent with the (other) literature when I was very precise in writing out all definitions and giving references for everything.

micjoswig:
I agree with @ulthiel : The current version is meaningful, useful and in line with some portion of the literature.

I also agree with @felix-roehrich : the other version is useful, too.

Since we have one working, correct and useful version available, this should go into the master asap. Concerning the interface, I agree with @fingolfin .  Except, instead of "Combinatorial" and "Group" as distinctions, I suggest to take @ulthiel 's "Asymmetric" and "Symmetric".  The reason is that there is one book (Kac & Cheung) which mentions them both, and they use those names.

On a personal note, in my work the asymmetric ones that we now have, are the more common ones; of course, this is biased.  So I would even be happy to make them the default.

ulthiel:
The non-symmetric ones are maybe more for the combinatorialist and number theorist, and the symmetric ones for the representation-theorist. But this is also not true: the non-symmetric ones can also used in, e.g., nicely writing down idempotents in the Hecke algebra (see https://arxiv.org/pdf/1609.03506, page 35). In the end, only two things are important in my opinion: 1) everything needs to be consistent and well-documented+referenced; b) both versions should be implemented. I gave you the non-symmetric ones.

fingolfin:
Thanks @ulthiel @micjoswig for the comments. Just to be clear, the comment on "consistency" was not meant at all to suggest that your code is not self-consistent and consistent with the literature reference here (it certainly is) -- I just talked with Felix about this, communication on the Internet is hard :-).

He showed me an example of code he currently has that shows why has an interest in having a compact way to evaluate "symmetric" q-binomials at different $q$. I've asked him to add it as a comment here.

Anyway, one solution to go forward would be indeed to merge this PR as-is, and then later also add "quantum number rings" as suggested above, which however would be sort of separate. The interface then we even could use `factorial` and `binomial` (instead of `q_factorial` or `quantum_factorial`).

I'd like to mull this a little bit more, and hopefully merge it during or after the next triage (this PR has been sitting here for two years now, and after I resurrected it last July it again sat without anyone seeming to care for the past 6 months, so I don't believe there is urgency to merge it RIGHT NOW IMMEDIATELY WITHOUT DELAY ;-).

felix-roehrich:
As Max said above, I would like to have a more concise way of writing q-analogues (when used with quantum groups). Here is some sample code to illustrate the issue:
```
U = quantum_group(:B, 3)
qi = quantum_parameters(U)
F = gens(U)
f = F[7]^2*F[9]^2*F[1]^4/(quantum_factorial(2, qi[7]))*quantum_factorial(2, qi[9]))*quantum_factorial(4, qi[1]))
```

micjoswig:
> As Max said above, I would like to have a more concise way of writing q-analogues (when used with quantum groups). Here is some sample code to illustrate the issue:
> 
> ```
> U = quantum_group(:B, 3)
> qi = quantum_parameters(U)
> F = gens(U)
> f = F[7]^2*F[9]^2*F[1]^4/(quantum_factorial(2, qi[7]))*quantum_factorial(2, qi[9]))*quantum_factorial(4, qi[1]))
> ```

I agree that this would be nice to have.  Is suitable code available? For if it is not, then we should proceed now with merging the existing code by @ulthiel after a final decision by @fingolfin concerning the interface.

fingolfin:
Perhaps we can have a zoom call with @ulthiel @felix-roehrich and myself (and anyone else interested in actively using q-analogs) to settle on which user interface we want. Defaulting to the "combinatorics" meaning is fine, but we need to decide how to get "the other".

Triage had some mild preference for using rings (those who spoke up).

Also `quantum_binomial` vs. `q_binomial`... Actually isn't the `q` in q-anlog historically not for "quantum" but for "q" as in "the order of a finite field"? That would suggest using the `q_` prefix and not `quantum_`.

ulthiel:
Sure. I don't have any strong opinions on anything.

About the q/quantum: I think you're right.

felix-roehrich:
I am still experimenting with the usage in my quantum groups implementation, but it seems that caching wouldn't work with a ring and needs to be directly on the quantum group. Furthermore, it seems the only use case for "symmetric" version is the quantum groups implementation.
The consensus seems to be to default to the combinatorial version, so I think it is fine to merge this; it is only experimental.

fingolfin:
The one thing we discussed and agreed on with various people but is *not* documented here (oops) is several of us come to the conclusion that we should call these "$q$-analogs" instead of "quantum analogs" because that's what most of the literature seems to do and the "q" originally wasn't for "quantum" either. Not even the literature linked in this PR calls them "quantum binomials" etc., and `q_binomial` is more in line with what other software uses.

So I just updated this PR to that convention. I am also now building the manual locally and will check it for content and formatting issues etc.

micjoswig:
> So I just updated this PR to that convention. I am also now building the manual locally and will check it for content and formatting issues etc.

Thx.  Looks good to me.

fingolfin:
There was an issue with the .bib file. Otherwise all looks good to me. Once CI passes this is fine to merge from my POV.

PR #2185, opened 2023-04-02
Serialization: remove eval fallback in decodeType

fingolfin:
This is a security nightmare. It also should not be necessary anymore.

Or maybe it *is* still necessary for something, then this will help us
find out for what, and then we can get rid of that, until this PR here
can be merged.


antonydellavecchia (approved):
Agreed, no longer necessary.

PR #2188, opened 2023-04-02
add `Oscar.test_experimental_module`

lgoettgens:
While changing my experimental module to the structure from #2146, I noticed that `Oscar.test_module` feels kind of weird to use due to the fixed prefix `test/`. Thus I attempted to add an alternative `Oscar.test_experimental_module` and put it in the new experimental docs as well.

cc @lkastner

### DIFF
--- a/src/Oscar.jl
+++ b/src/Oscar.jl
@@ -401,40 +401,55 @@ end
 
 
 @doc raw"""
-    test_module(x, new::Bool = true)
+    test_module(file::AbstractString; new::Bool = true)
 
-Run the Oscar tests in the file `x.jl` where `x` is a string.
+Run the Oscar tests in the file `test/<file>.jl` where `file` may be a path.
 
-If `x == "all"` run the entire testsuite.
+The optional parameter `new` takes the values `false` and `true` (default). If
+`true`, then the tests are run in a new session, otherwise the currently active
+session is used.
+
+For experimental modules, use [`test_experimental_module`](@ref) instead.
+"""
+function test_module(file::AbstractString; new::Bool=true)
+  julia_exe = Base.julia_cmd()
+  rel_test_file = normpath("test", "$file.jl")
+  test_file = joinpath(oscardir, rel_test_file)
+
+  if new
+    cmd = "using Test; using Oscar; Hecke.assertions(true); include(\"$test_file\");"
+    @info("spawning ", `$julia_exe -e \"$cmd\"`)
+    run(`$julia_exe -e $cmd`)
+  else
+    @info("Running tests for $rel_test_file in same session")
+    try
+      include(test_file)
+    catch e
+      if isa(e, LoadError)
+        println("You need to do \"using Test\"")
+      else
+        rethrow(e)
+      end
+    end
+  end
+end
+
+@doc raw"""
+    test_experimental_module(project::AbstractString; file::AbstractString="runtests", new::Bool = true)
+
+Run the Oscar tests in the file `experimental/<project>/test/<file>.jl`
+where `file` may be a path.
+The default is to run the entire test suite of the module `project`.
 
 The optional parameter `new` takes the values `false` and `true` (default). If
 `true`, then the tests are run in a new session, otherwise the currently active
 session is used.
 """
-function test_module(x::AbstractString, new::Bool = true)
-   julia_exe = Base.julia_cmd()
-   if x == "all"
-     test_file = joinpath(oscardir, "test/runtests.jl")
-   else
-     test_file = joinpath(oscardir, "test/$x.jl")
-   end
-
-   if new
-     cmd = "using Test; using Oscar; Hecke.assertions(true); include(\"$test_file\");"
-     @info("spawning ", `$julia_exe -e \"$cmd\"`)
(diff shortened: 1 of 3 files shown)


src/Oscar.jl
```diff
@@ -404,38 +404,60 @@ end
...
+"""
+function test_experimental_module(
+  project::AbstractString, x::AbstractString, new::Bool=true
+)
+  if x == "all"
+    test_file = "../experimental/$project/test/runtests"
+  else
+    test_file = "../experimental/$project/test/$x"
+  end
+  test_module(test_file, new)
```
lkastner: I am not really sure, what the convention is, but one alternative could be to put a `;` behind `project::...` to mark the other two parameters as optional. Also I'd prefer a better name for `x`, maybe `testfile` or `testscript` to keep it short. What do you think?

lkastner (approved):
Many changes seem to fix the indentation, is that correct? It would probably be good to do such a thing in a separate PR (although one can turn seeing these changes off for the review).

lgoettgens:
> Many changes seem to fix the indentation, is that correct? It would probably be good to do such a thing in a separate PR (although one can turn seeing these changes off for the review).

My bad. I forgot to remove them from the diff before committing. Should be fixed now.

src/Oscar.jl
lgoettgens: I was thinking about that as well. Then `x::AbstractString="runtests"` would be easier to understand than the default `"all"`. To keep consistency with `test_module`, I would then do the same there, i.e. making the path `x` and `new` kwargs. To keep the name short (it is a kwarg now), I would suggest renaming `x` to `file` in `test_module` and `test_experimental_module`.
Is that what you had in mind?

lkastner: Funny, I was just wondering, who named it `x` in `test_module`, and git blame give:
```
50176b21c43 (Lars Kastner           2022-09-28 10:20:44 +0200 414) function test_module(x::AbstractString, new::Bool = true)
```
Joke aside:
I agree that consistency is a great thing, and to rename there as well. Do we need a deprecation then? I'd say no, since this does not affect functionality of Oscar, but is a developer tool.

lgoettgens: I would agree. Nobody should ever use this in code.

lgoettgens: Is this now working, as you would expect it?

src/Oscar.jl
```diff
@@ -402,40 +402,56 @@ end
...
+For experimental modules, use [`test_experimental_module`](@ref) instead.
 """
-function test_module(x::AbstractString, new::Bool = true)
-   julia_exe = Base.julia_cmd()
-   if x == "all"
-     test_file = joinpath(oscardir, "test/runtests.jl")
-   else
-     test_file = joinpath(oscardir, "test/$x.jl")
-   end
+function test_module(;file::AbstractString="runtests", new::Bool=true)
```
lkastner: For `test_module` I'd make the `file` argument mandatory, since for the entire testsuite one could just use
```
]test Oscar
```
Sorry if this was unclear. Do you think that is too inconsistent?

lkastner (approved):
Feel free to merge once you have decided on my last comment.

fingolfin:
Unfortunately there are conflicts now

src/Oscar.jl
lgoettgens: Hmm, I see your point. So maybe make it mandatory, but still a kwarg? Or a plain mandatory arg?

lgoettgens:
> Unfortunately there are conflicts now

I'll try to rebase tomorrow.

src/Oscar.jl
lkastner: I thought plain, like `project` for the `test_experimental_module`.

lgoettgens:
I rebased, fixed the conflicts, changed `file` of `Oscar.test_module` back to an arg, adapted the docs accordingly, and squashed everything. This should be good to go now according to @lkastner's comment.

src/Oscar.jl
```diff
@@ -401,40 +401,55 @@ end
...
+    cmd = "using Test; using Oscar; Hecke.assertions(true); include(\"$test_file\");"
+    @info("spawning ", `$julia_exe -e \"$cmd\"`)
+    run(`$julia_exe -e $cmd`)
+  else
+    @info("Running tests for $rel_test_file in same session")
+    try
+      include(test_file)
+    catch e
+      if isa(e, LoadError)
+        println("You need to do \"using Test\"")
```
fingolfin, line 429: Note a change request (but perhaps someone would like to follow this up in another PR or an issue): I don't like this error message. It can be widely misleading.

Instead of *guessing* that perhaps `using Test` is missing, we can do much better: We could *first* check whether it is available, even before the `include`, and refuse to even do the `include` if it isn't. 

My second concern is: we are `include`ing the test file into the Oscar namespace. But I don't think that's how tests are run when one uses `Pkg.test` and thus also not how CI runs it; and also not how this function does it when `new` is specified. Rather, the `include` is always done from the REPL.

So perhaps instead of `include(test_file)` we should be doing `Base.include(Base.Main, test_file)` ?

PR #2194, opened 2023-04-03
remove duplicate `experimental` docs

lgoettgens:
I noticed the file `docs/src/DeveloperDocumentation/experimental.md` just being a subset of the information in the new `docs/src/Experimental/intro.md`. Thus is propose to remove the former and redirect all links to the latter instead.

cc @lkastner


lkastner (commented):
I was aware of this, I was just wondering what to do. We will definitely need some hints at this intro in the developer docs, otherwise it might be overlooked. I will push a few pointers onto this PR, so please hold.

lkastner (approved):
I rebased onto master and modified the comments about the structure a bit, mentioning the README.md. Hope this is ok for you @lgoettgens .

fingolfin (approved):

lgoettgens:
Looks good to me. Thanks for editing @lkastner!

PR #2196, opened 2023-04-03
Replace '@doc Markdown.doc' by '@doc raw'

fingolfin:
As far as I can tell, there is no functional difference.

We could now also remove all remaining references to 'Markdown', but that
might break various in-flight PRs. So I only removed some, and also kept
Markdown in Project.toml. We can fully remove it in a future PR.


fieker:
if there is no functional difference, then why do we want to do that?
It used to be that """ without Markdown actually is using a different markdown

fingolfin:
> if there is no functional difference, then why do we want to do that?

Do what? Apply the changes in this PR? Multiple reasons:
- because `@doc raw"""` is shorter than `@doc Markdown.doc"""`
- it is also clearer what it does -- `raw_str` is documented, `Markdown.doc_str` is not
- it allows us to drop `using Markdown` in a ton of places, and indeed remove `Markdown` from `Project.toml`
- it gets rid of a "nonsense idiom" that survives purely through "copy&paste code I don't understand but that seems to work" (and note how this even "mutates" over time to other forms like `@Markdown.doc """` and others)

> It used to be that """ without Markdown actually is using a different markdown

Perhaps it did in some Julia 0.x version, but not anymore as far as I can tell. All it gave us in effect was the same as `raw`, that is: don't treat `$` as interpolation, and allow writing TeX like `\mathbb` without escaping it to `\\mathbb`

If there are other differences, I'd love to hear them. I think @HechtiDerLachs reported that he conducted experiments on this recently?

fieker:
even shorter, julia suggests just using """ should be the default - unless ones wants to be fancy...

fingolfin:
Yes, I'd prefer this even simpler form. But we often *are* fancy as described above. So going beyond what is in this PR would require inspecting the content of the docstrings

micjoswig (approved):
I am fine with that suggestion.  I would be similarly happy with any other solution which makes things more uniform and does not break our current docs.

Once everyone is happy, please add a comment in the style guide.

lkastner:
> Once everyone is happy, please add a comment in the style guide.

~~Please also modify the page https://docs.oscar-system.org/dev/DeveloperDocumentation/documentation/~~
My bad, I thought this was an issue and not a PR...

HechtiDerLachs:
> If there are other differences, I'd love to hear them.

Nothing that we systematically kept track of. It was as you said: We copy-pasted from various other places and simply tried to get our docstring to run. Doing so, we observed that depending on the introductory clause several latex expressions would work and others not. We eventually settled for some version that we found half way satisfactory, but not knowing what was going on behind the scenes. And, frankly, not even wanting to know. I mean: In the end we simply wanted to carry over Uli's JuLie code and not take a lesson in the history of Markdown macros that we could eventually extract from tedious and lengthy investigations. Sorry to disappoint you here. But personally I am 100% in favor of streamlining this! Thanks for your effort.

fingolfin:
@HechtiDerLachs thanks for the clarification. And I am not at all disappointed :-) I just thought I should ask in case you made some relevant observations that I missed :-).

BTW I should explain that I tested this PR by building the docs with it, and then diffing the result against docs built from master. The only diffs were in the source links, which is natural, because those include the SHA1 of the commit the docs were built from. After removing this, it turned out that also the line ranges in those source links differed, but the change is an improvement (IMHO): right now our sources links (e.g. on https://docs.oscar-system.org/stable/Groups/basics/) just linke to the top line of the docstring, like this one: https://github.com/oscar-system/Oscar.jl/blob/250b6338c6022c6d105de4b0a73d3aa089af5bd3/src/Groups/types.jl#L17

With this patch, the links highlights the full docstring, like this one https://github.com/oscar-system/Oscar.jl/blob/250b6338c6022c6d105de4b0a73d3aa089af5bd3/src/Groups/types.jl#L17-L29

So that's a (very minor) improvement. I observed no negative regressions

lkastner (approved):

PR #2197, opened 2023-04-03
Join the plane curves documentation back together

### LINKED ISSUE #2189: Leftovers from #2157 in `docs/src/Experimental/`
In `/docs/src/Experimental/`, there are some files about curves `{divisors,elliptic_curves,introduction,non_plane_curves}.md`. I suppose, they have been forgotten in #2157 when turning `PlaneCurve` into the new `experimental` structure.

cc @lkastner

fingolfin:
With the recent restructuring of `experimental`, the Markdown files for PlaneCurve were accidentally split into two. This joins them back together.

Note that the PlaneCurves docs were disabled before, and this enables them again, albeit under the "Experimental" umbrella. Not sure we want that, perhaps we should still keep them completely out of the manual -- @wdecker shall decide.


wdecker:
@fingolfin : I understand that @simonbrandhorst will work on the design of curves in Oscar, in particular he will check what should be integrated from the old code. So probably we should wait a little bit with including the docu again. Simon?

simonbrandhorst:
It can stay disabled for now.

experimental/PlaneCurve/docs/doc.main
```diff
@@ -1,3 +1,9 @@
 [
-   "plane_curves.md",
+    "Plane Curves" => [
+        "introduction.md",
+        "plane_curves.md",
+        "divisors.md",
+        "elliptic_curves.md",
+        "non_plane_curves.md",
+    ],
 ]
```
lkastner: If this should be disabled, the easiest way should be to remove `doc.main` or maybe rename it to something like `doc.main.disabled`?

lkastner (approved):
If the docs for `PlaneCurve` are disabled -- for me this means they are invisible -- then maybe we should also remove the hint at `PlaneCurve` from `docs/src/Experimental/intro.doc`?

experimental/PlaneCurve/docs/doc.main
fingolfin: Hmm, is that easier than commenting out the code here? Commenting out also means that it is easier to "rediscover" these docs and to notice that they are deactivated; and by adding a comment we can even explain why...

lkastner: If commenting out works. It should though, since I only copied the other code for reading the `doc.main` from elsewhere.

fingolfin:
Hrm, this is the second time I see a CI job fail because the "Build packages" phase takes long (this time almost 87 minutes, the other time it was 44 minutes), while normally it is 2-3 minutes

PR #2200, opened 2023-04-03
FTheoryTools: Use @vprint and add README.md

### LINKED ISSUE #2180: A lot of infos during tests about "AUXILIARY ambient space"
There are quite are lot of messages of the form 
```
[ Info: Base space was not fully specified. Returning AUXILIARY ambient space.
```
during the tests. Is there a way to disable this? See https://github.com/oscar-system/Oscar.jl/actions/runs/4577059265/jobs/8082008614#step:6:1194

Seem to be coming from the FTheoryTools. CC: @HereAround

HereAround:
This PR has two goal:
1. Fix https://github.com/oscar-system/Oscar.jl/issues/2180.
2. Add file `README.md` in Oscar.jl/experimental/FTheoryTools to comply with the new structure for experimental code.

cc @apturner

### DIFF
--- a/experimental/FTheoryTools/README.md
+++ b/experimental/FTheoryTools/README.md
@@ -1,0 +1,59 @@
+# Welcome to FTheoryTools
+
+## Goal
+
+We aim to automate a number of recurring and (at least in part) tedious computations in F-theory model building
+as described in large detail in [Wei18](@cite). Specifically, we focus on the following setups:
+* 4d F-theory compactifications,
+* defined by a global and singular Weierstrass model as codimension 1 locus of a toric ambient space $Y$,
+* which can be crepantly resolved.
+Some of the techniques/algorithms extend naturally to more general settings. For example, it is not at all necessary to restrict to 4-dimensional settings (or alternatively, base spaces of dimension 3). Indeed, the current implementation does allow for *arbitrary* base dimension. For more extensions that we might address in the future, please take a look at the section "possible future extensions" below.
+
+We aim for the following workflow:
+* User input:
+    * Weierstrass polynomial $P_W$,
+    * Data defining the toric ambient space $Y$ (if applicable),
+    * Choice of resolved phase (if applicable),
+    * Generating sections (for $\operatorname{U}(1)$ symmetries).
+* Output:
+    * Singular loci in codimension 1, 2, and 3,
+    * Defining data of resolved geometry,
+    * (Pictures of) fibre diagrams of resolved fibre over the originally singular loci, including intersections of $\operatorname{U}(1)$-sections,
+    * Gauge group,
+    * Topological data (e.g., Euler number).
+
+
+## Status
+
+This project just began, and is therefore in its experimental stage. Upcoming tasks include, but are not limited, to the following:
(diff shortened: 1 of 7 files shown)


experimental/FTheoryTools/src/TateModels/attributes.jl
```diff
@@ -114,7 +114,7 @@ Normal toric variety without torusfactor
 ```
 """
 @attr AbstractNormalToricVariety function toric_base_space(t::GlobalTateModel)
-    base_fully_specified(t) || @info("Base space was not fully specified. Returning AUXILIARY base space.")
+    base_fully_specified(t) || @vprint :GlobalTateModel 1 "Base space was not fully specified. Returning AUXILIARY base space."
```
fingolfin: I think you may also need to use `add_verbosity_scope(:GlobalTateModel)` as described in the `@vprint` docstring?

HereAround: Got it. Thank you!

src/Oscar.jl
```diff
@@ -249,6 +249,10 @@ function __init__()
 
     add_verbose_scope(:K3Auto)
     add_assert_scope(:K3Auto)
+    
+    add_verbose_scope(:GlobalTateModel)
+    add_verbose_scope(:GlobalWeierstrassModel)
```
HereAround, line 254: @thofma Initially, I tried to add these lines at the very top of `Oscar.jl/experimental/FTheoryTools/src/FTheoryTools.jl`. This is the top-level src-file of `FTheoryTools` which includes all other files. So this seemed a natural place, given that `FTheoryTools` is experimental.

Still, if added at the very top of `Oscar.jl/experimental/FTheoryTools/src/FTheoryTools.jl`, I received an `ErrorException` showing the error message "Not a valid symbol" when running the tests. Why is that?

thofma: They need to be added very early in the loading of a package, so must go into `__init__`. If `FTheoryTools` would have its own module, you could put it in the `__ini__` function of that module, like here:

https://github.com/oscar-system/Oscar.jl/blob/915aeb9de5c3ffe97ffcbeca20b6fbca8e61d22c/experimental/GModule/Cohomology.jl#L9-L15

But just putting those in the `__init__` of Oscar itself is also fine.

HereAround: Thank you. That makes sense.

experimental/FTheoryTools/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Contact
+
+Please direct questions about this part of OSCAR to the following people:
+* [Martin Bies](https://martinbies.github.io/),
+* [Andrew Turner](https://apturner.net/).
+
+You can ask questions in the [OSCAR Slack](https://www.oscar-system.org/community/#slack).
+
+Alternatively, you can [raise an issue on github](https://www.oscar-system.org/community/#how-to-report-issues).
```
fingolfin:
```suggestion
Alternatively, you can [raise an issue on GitHub](https://github.com/oscar-system/Oscar.jl).
```

fingolfin (approved):

thofma (approved):
Looks, good thanks!

PR #2202, opened 2023-04-03
Extend support for IncidenceMatrix

alexej-jordan:
The functionality of `IncidenceMatrix` is mainly imported from `Polymake.jl`. This PR aims to better document this type within `OSCAR` and to add convenient access to rows and columns in specific format.


lkastner:
Please have a look at #2260 . Maybe you could add a constructor here for the following data:
```
julia> c = positive_hull([1 0 0; 1 1 0; 1 1 1; 1 0 1])

julia> Polymake.polytope.faces_of_dim(c.pm_cone, 2)
pm::Array<pm::Set<long, pm::operations::cmp>>
{0 1}
{0 3}
{2 3}
{1 2}


julia> IncidenceMatrix(Polymake.polytope.faces_of_dim(c.pm_cone, 2))
ERROR: MethodError: no method matching IncidenceMatrix{Polymake.NonSymmetric}(::Polymake.ArrayAllocated{Polymake.Set{Int64}})
Closest candidates are:
  IncidenceMatrix{Polymake.NonSymmetric}() at ~/.julia/packages/CxxWrap/rxpzr/src/CxxWrap.jl:623
  IncidenceMatrix{Polymake.NonSymmetric}(::SparseArrays.AbstractSparseMatrix) at ~/.julia/packages/Polymake/A0Z5B/src/incidencematrix.jl:32
  IncidenceMatrix{S}(::UndefInitializer, ::Integer, ::Integer) where S<:Union{Polymake.NonSymmetric, Polymake.Symmetric} at ~/.julia/packages/Polymake/A0Z5B/src/incidencematrix.jl:3
  ...
Stacktrace:
 [1] IncidenceMatrix(x::Polymake.ArrayAllocated{Polymake.Set{Int64}})
   @ Polymake ~/.julia/packages/Polymake/A0Z5B/src/incidencematrix.jl:68
 [2] top-level scope
   @ REPL[35]:1
```
I can add a `Polymake.to_one_based_indexing` if that makes it easier. Probably the constructor should also take the number of columns, as this might not really follow from the output of `faces_of_dim`, definitely not if we have a polytope with rays.

alexej-jordan:
@lkastner the latest commit contains a suggestion on how to receive `IncidenceMatrix` objects from `SubObjectIterator`s. The handling would then look like this:
```
julia> PF = face_fan(cube(3))
Polyhedral fan in ambient dimension 3

julia> IncidenceMatrix(maximal_cones(PF))
6×8 IncidenceMatrix
[1, 3, 5, 7]
[2, 4, 6, 8]
[1, 2, 5, 6]
[3, 4, 7, 8]
[1, 2, 3, 4]
[5, 6, 7, 8]
```
This might not be exactly what you suggested, but it is easy to extend (same idea as for `matrix_for_polymake`: just add one definition for `_incidence_matrix` for each applicable access function) and should not cause noticeable overhead (because we mostly just work with references). Also, writing `IncidenceMatrix(maximal_cones(PF))` seems more intuitive for me than `maximal_cones(IncidenceMatrix, PF)`, because here `IncidenceMatrix` is the whole container whereas `maximal_cones(Cone, PF)` specifies single elements.

benlorenz:
We discussed this and decided that we should support both variants, i.e. `maximal_cones(IncidenceMatrix,x)` and `IncidenceMatrix(maximal_cones(x))`.

docs/src/PolyhedralGeometry/intro.md
```diff
@@ -110,6 +110,26 @@ Type                                   | An `AffineHyperplane` corresponds to...
...
+IncidenceMatrix
+```
+
+From the example it can be seen that this type supports `julia`'s matrix functionality. There are also functions to retrieve specific rows or columns as a `Vector` or a `Set` over the non-zero indices.
+
+```@docs
+row_as_set
+row_as_vector
+column_as_set
+column_as_vector
```
benlorenz:
```suggestion
column
row
```

alexej-jordan:
@lkastner regarding your request for `IncidenceMatrix(Polymake.polytope_faces_of_dim(...))`:
why do you need it for this `Polymake.jl` function? adding a constructor to `Polymake.jl` which can take `Set`s (and probably an `Int` for size) should be no problem, but is there a reason you do not use `IncidenceMatrix(faces(c, 2))`? with the code on this PR, this syntax also works:
```
julia> c = positive_hull([1 0 0; 1 1 0; 1 1 1; 1 0 1])

julia> faces(c, 2)
4-element SubObjectIterator{Cone{QQFieldElem}}:

julia> IncidenceMatrix(faces(c, 2))
4×4 IncidenceMatrix
[1, 2]
[1, 4]
[3, 4]
[2, 3]

```

docs/src/PolyhedralGeometry/intro.md
alexej-jordan: Thanks for pointing this out! I still changed this a bit so only the specific methods for `IncidenceMatrix` are included. This might help avoid future conflicts for such a generic funtion name.

lkastner:
> @lkastner regarding your request for `IncidenceMatrix(Polymake.polytope_faces_of_dim(...))`: why do you need it for this `Polymake.jl` function? adding a constructor to `Polymake.jl` which can take `Set`s (and probably an `Int` for size) should be no problem, but is there a reason you do not use `IncidenceMatrix(faces(c, 2))`? with the code on this PR, this syntax also works:

I just didn't know this works, probably I tried something similar and then got frustrated. Sorry.

benlorenz (approved):

PR #2206, opened 2023-04-04
fix Oscar.test_experimental_module

lgoettgens:
Unfortunately, #2188 contained a syntax error. This should be fixed with this change.

cc @lkastner


lkastner (approved):

lkastner:
Basically my fault... :sweat_smile:

PR #2207, opened 2023-04-04
LieAlgebras: More Lie algebra constructions from/to GAP

lgoettgens:
Allow the transformation of fin. dim. Lie algebras from/to GAP. This will extend `Oscar.iso_oscar_gap` and `Oscar.iso_gap_oscar`.
This enables us to, in the future, use GAP to compute some stuff, similar to the group situation.


lgoettgens:
I have implemented `_iso_oscar_gap` and `_iso_gap_oscar` methods for Lie algebras, similar to the already existing ones in `src/GAP/`. But I am unsure, where to put them. Since this is part of `experimental`, I do not know if they should be called from `Oscar.iso_oscar_gap` and `Oscar.iso_gap_oscar` and if they should be put into `experimental` or `src/GAP/`.

In contrast to the already existing implementations, this has an additional layer of caching. The idea behind that is as follows: I would like all Lie algebras to have the possibility to have a corresponding GAP Lie algebra with the iso as an attribute (-> thus using attribute storage). This would not need the additional layer of caching due to the `@attr` e.g. here. https://github.com/oscar-system/Oscar.jl/blob/be51cf81768af3d8bcb2c184cb9b2d68b23a855e/src/GAP/iso_oscar_gap.jl#L386

For the converse direction (GAP -> Oscar), I would like the codomain `LO` of the return value of `_iso_gap_oscar(LG)` to know about the already created isomorphic GAP object `LG` and the iso (which is just the inverse of the return value). This for me doesn't seem possible by hacking into the cache used by `@attr` without relying on the way how the key gets created inside AbstractAlgebra. Ideas on how to handle that in a better way are welcome.

lgoettgens:
This is ready to review, but **NOT** ready to merge yet. The decision on how to integrate `_iso_oscar_gap` and `_iso_gap_oscar` has to be done and implemented first.
For a review, I would ping @fingolfin and @ThomasBreuer. (I unfortunately cannot use the gh tool to ask for a review.)

lgoettgens:
Rebased on top of #2262 to avoid conflicts.

fingolfin:
> The decision on how to integrate `_iso_oscar_gap` and `_iso_gap_oscar` has to be done and implemented first.

I've read through what you've written several times now, and I honestly don't understand what the problem is / which decision has to be made. The only thing I see listed as needing any kind of work or decision is this:

> I would like the codomain LO of the return value of _iso_gap_oscar(LG) to know about the already created isomorphic GAP object LG and the iso (which is just the inverse of the return value). 

But I don't understand what the problem there is, isn't this just a matter of calling `set_attribute!` on the codomain (which is an attribute storing Julia object, after all) to store the iso you just constructed?


Isn't it more the converse which is a problem: namely, ensure that the *domain* of `_iso_gap_oscar` (which is a GAP object) only ever gets assigned a single unique Julia object (and then likewise for the *codomain* of `_iso_oscar_gap`)? To solve that, one would have to store the isomorphism inside the GAP object (which is no problem for those which are attribute storing in the GAP sense, which is related to but different from the OSCAR attribute storage system)

lgoettgens:
> > The decision on how to integrate `_iso_oscar_gap` and `_iso_gap_oscar` has to be done and implemented first.
> 
> I've read through what you've written several times now, and I honestly don't understand what the problem is / which decision has to be made. 

I should have written a bit more about that, sorry. To work as expected, the methods `_iso_oscar_gap` and `_iso_gap_oscar` introduced in this PR would have to be hooked up into `src/GAP/iso_gap_oscar.jl` and `src/GAP/iso_oscar_gap.jl`, e.g. called from https://github.com/oscar-system/Oscar.jl/blob/e92603653dd1f327e24ee524de51c26eec896b36/src/GAP/iso_gap_oscar.jl#L34

This would add `experimental/LieAlgebras/` as a dependency of `src/GAP/`. Moving the functionality from this PR into the two files in `src/GAP/` would make this much more complicated to de-tangle if ever wished so in the future. Not moving it, would make the inclusion order in `Oscar.jl` much more complicated. In particular, `experimental/LieAlgebras/` would have to be included in-between different files from `src/GAP/` (at least in my thoughts).

I hope the above helps a bit to understand my problems and thought process.


> > I would like the codomain LO of the return value of _iso_gap_oscar(LG) to know about the already created isomorphic GAP object LG and the iso (which is just the inverse of the return value).
> 
> But I don't understand what the problem there is, isn't this just a matter of calling `set_attribute!` on the codomain (which is an attribute storing Julia object, after all) to store the iso you just constructed?

There is no problem per se. I just added some justification on why I used `get_attribute/set_attribute!` instead of a simpler `@attr`.

> Isn't it more the converse which is a problem: namely, ensure that the _domain_ of `_iso_gap_oscar` (which is a GAP object) only ever gets assigned a single unique Julia object (and then likewise for the _codomain_ of `_iso_oscar_gap`)? To solve that, one would have to store the isomorphism inside the GAP object (which is no problem for those which are attribute storing in the GAP sense, which is related to but different from the OSCAR attribute storage system)

Yes, indeed. However, I would not see that as a problem for `iso_oscar_gap` and `iso_gap_oscar` for Lie algebras, but for other types as well. Thus, I would propose a general solution in `src/GAP/` for this.

fingolfin:
> To work as expected, the methods `_iso_oscar_gap` and `_iso_gap_oscar` introduced in this PR would have to be hooked up into `src/GAP/iso_gap_oscar.jl` and `src/GAP/iso_oscar_gap.jl`,

Could you please elaborate? I.e. what expectation does this meet? Resp. what undesirable things happen if we don't "hook it up"?

ThomasBreuer:
Currently `iso_gap_oscar(D::GapObj)` is one Julia function that checks the GAP type of `D` and then decides which Julia function will do the work. In order to be able to add more possible `D` in `Oscar.jl/experimental` code (or in private code), one possibility would be to delegate to GAP's method selection; then the GAP methods would call the Julia functions in question. I can provide a pull request for that.

lgoettgens:
> In order to be able to add more possible `D` in `Oscar.jl/experimental` code (or in private code), one possibility would be to delegate to GAP's method selection; then the GAP methods would call the Julia functions in question. I can provide a pull request for that.

That sound like a good approach. Thanks for working on that @ThomasBreuer!

lgoettgens:
For my convenience, this now includes the changes of #2309.

experimental/LieAlgebras/test/iso_oscar_gap-test.jl
```diff
@@ -0,0 +1,106 @@
...
+        special_linear_lie_algebra(RO, 3),
+        special_orthogonal_lie_algebra(RO, 4),
+      ]
+
+      @testset for LO in lie_algebras
+        iso = Oscar.iso_oscar_gap(LO)
+        @test domain(iso) == LO
+        LG = codomain(iso)
+        @test GAPWrap.IsLieAlgebra(LG)
+        # @test codomain(Oscar.iso_gap_oscar(LG)) === LO
```
lgoettgens: @fingolfin I guess you meant something like this in the last paragraph of https://github.com/oscar-system/Oscar.jl/pull/2207#issuecomment-1518424941?
It's not working right now. To remember a future pr, I added it in a comment for now.

lgoettgens:
This pr now allows the use of the "usual" functions `Oscar.iso_gap_oscar` and `Oscar.iso_oscar_gap` for Lie algebras as well. From my pov, this is ready to go now (it has already been rebased on master after the merge of #2309).

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -195,19 +195,20 @@ end
...
+#   Isomorphism to GAP accessors
 #
 ###############################################################################
 
-function _gap_object(L::LieAlgebra{C}) where {C<:RingElement}
-  # later change to storing an isomorphism instead
-  get_attribute!(L, :gap_object) do
-    gap_lie_algebra_by_struct_consts(L)
-  end
+# currently, the user should only use _iso_oscar_gap(LO) to get the isomorphism
```
ThomasBreuer: Why?
The idea in `src/GAP/iso_oscar_gap.jl` is that
- `iso_oscar_gap` deals with setting and fetching the attribute value, and calls `_iso_oscar_gap`, and
- the methods for `_iso_oscar_gap` just have to do the work, they need not care about attribute values.

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -216,34 +217,15 @@ end
...
-  return L
+function lie_algebra(
+  gapL::GAP.GapObj,
+  s::Vector{<:VarName}=[Symbol("x_$i") for i in 1:GAPWrap.Dimension(gapL)];
+  cached::Bool=true,
+)
+  @req GAPWrap.IsLieAlgebra(gapL) "gapL must be a Lie algebra."
+  if GAPWrap.IsLieObjectCollection(gapL)
+    return codomain(_iso_gap_oscar_linear_lie_algebra(gapL, s; cached))
+  else
```
ThomasBreuer: Perhaps check whether `gapL` is finite dimensional; in principle, one can construct free Lie algebras and finitely presented Lie algebras in GAP, which are (currently) not supported here.

experimental/LieAlgebras/src/iso_gap_oscar.jl
```diff
@@ -0,0 +1,84 @@
+###############################################################################
+#
+#   Lie algebras
+#
+###############################################################################
+
+function _iso_gap_oscar_lie_algebra(F::GAP.GapObj)
+  if GAPWrap.IsLieObjectCollection(F)
+    return _iso_gap_oscar_linear_lie_algebra(F)
+  else
```
ThomasBreuer: see the comment for `lie_algebra`

experimental/LieAlgebras/test/iso_gap_oscar-test.jl
```diff
@@ -0,0 +1,110 @@
...
+      ]
+
+      @testset for LG in lie_algebras
+        iso = Oscar.iso_gap_oscar(LG)
+        @test domain(iso) == LG
+        LO = codomain(iso)
+        @test LO isa AbstractLieAlgebra
+        @test codomain(Oscar.iso_oscar_gap(LO)) === LG
+
+        @test iso == Oscar.iso_gap_oscar(LG) # test caching
```
ThomasBreuer: use `===`?

experimental/LieAlgebras/test/iso_gap_oscar-test.jl
```diff
@@ -0,0 +1,110 @@
...
+          ),
+        ),
+      ]
+
+      @testset for LG in lie_algebras
+        iso = Oscar.iso_gap_oscar(LG)
+        @test domain(iso) == LG
```
ThomasBreuer: use `===`?

experimental/LieAlgebras/test/iso_gap_oscar-test.jl
```diff
@@ -0,0 +1,110 @@
...
+          GAP.Obj("basis"),
+        ),
+      ]
+
+      @testset for LG in lie_algebras
+        iso = Oscar.iso_gap_oscar(LG)
+        @test domain(iso) == LG
+        LO = codomain(iso)
+        @test LO isa LinearLieAlgebra
+        @test codomain(Oscar.iso_oscar_gap(LO)) === LG
```
ThomasBreuer: Perhaps put the tests into a function, then the code duplication can be avoided.

ThomasBreuer (approved):
I have added some minor comments.

experimental/LieAlgebras/src/LieAlgebra.jl
lgoettgens: I tried to explain that in https://github.com/oscar-system/Oscar.jl/pull/2207#issuecomment-1497783766.

My problem is that, in some instances, I would like to set the iso by myself instead of using the "standard" one. For example, when I use the constructor `lie_algebra(R::Ring, dynkin::Tuple{Char,Int})`, internally `GAP.Globals.SimpleLieAlgebra` gets called, and the result gets translated to Oscar via `iso_gap_oscar`. However, the knowledge about this GAP Lie algebra then gets lost. I would instead prefer to save (the inverse of) this iso into the Oscar Lie algebra here.
This didn't seem possible to me without relying on the internals of AbstractAlgebra on how to name attributes for `@attr`.

If this behavior is not liked or there are some other ideas on how to handle that, comments are very welcome.

lgoettgens: I just found a naming guarantee of the attribute behind `@attr` in https://github.com/Nemocas/AbstractAlgebra.jl/blob/76a85e28b137ec0b97e2131970a51d443dc3dce7/src/Attributes.jl#L284. With this, I can remove that additional layer of caching and just set the iso myself in the `@attr` storage.

lgoettgens: This has been addressed in 2e499c4a6c4e92603813d667ef66dfc1828e2a38.

lgoettgens:
Is this PR still waiting for something?

ThomasBreuer:
@lgoettgens No. Sorry for not merging it earlier.

PR #2208, opened 2023-04-04
Projective varieties

simonbrandhorst:
A first implementation for projective algebraic sets and projective varieties. 
Included the necessary cleanup of `ProjectiveScheme`.
Docu is still missing.


src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -729,7 +729,7 @@ That is if ``X`` is integral when base changed to any field extension of ``k``.
 """
 @attr Bool function is_geometrically_integral(X::AbsSpec{<:Field,<:MPolyAnyRing})
   is_integral(X) || return false
-  throw(NotImplementedError("absolute primary decomposition is currently only available over the rationals"))
+  throw(NotImplementedError(:is_geometrically_integral, "absolute primary decomposition is currently only available over the rationals"))
```
HechtiDerLachs, line 732: I do not quite understand this. Why does this function check integrality, but then nevertheless throws an error?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
```diff
@@ -0,0 +1,57 @@
...
+otherwise take on faith that ``I`` is radical.
+"""
+function vanishing_locus(I::MPolyIdeal{<:MPolyDecRingElem}; check::Bool=true)
+  if check
+    Irad = radical(I)
+  else
+    Irad = I
+  end
+  X = ProjectiveScheme(base_ring(Irad),Irad)
+  return ProjectiveAlgebraicSet(X, check=check)
```
HechtiDerLachs, line 39: Maybe turn off checks here since they have happened above?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+vanishing_locus(p::MPolyDecRingElem, check::Bool=true) = vanishing_locus(ideal(parent(p),p), check=check)
```
HechtiDerLachs:
```suggestion
vanishing_locus(p::MPolyDecRingElem; check::Bool=true) = vanishing_locus(ideal(parent(p),p), check=check)
```

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
```diff
@@ -0,0 +1,57 @@
...
+
+@doc raw"""
+    set_theoretic_intersection(X::AbsProjectiveAlgebraicSet, Y::AbsProjectiveAlgebraicSet) -> AbsProjectiveAlgebraicSet
+
+Return the set theoretic intersection of `X` and `Y` as a `ProjectiveAlgebraicSet`.
+
+This is the reduced subscheme of the scheme theoretic intersection.
+"""
+function set_theoretic_intersection(X::AbsProjectiveAlgebraicSet, Y::AbsProjectiveAlgebraicSet)
+  throw(NotImplementedError(:set_theoretic_intersection,X))
```
HechtiDerLachs, line 55: Is it intended that this function is not implemented? What about the implementation below?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Methods.jl
```diff
@@ -0,0 +1,52 @@
...
+########################################################
+
+@doc raw"""
+    irreducible_components(X::ProjectiveAlgebraicSet) -> Vector{ProjectiveVariety}
+
+Return the irreducible components of `X` defined over the same base field.
+
+Note that even if `X` is irreducible, there may be several geometric irreducible components.
+"""
+function irreducible_components(X::ProjectiveAlgebraicSet)
```
HechtiDerLachs: Missing `Abs`?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Methods.jl
```diff
@@ -0,0 +1,52 @@
...
+  I = defining_ideal(X)
+  J = minimal_primes(I)
```
HechtiDerLachs: Why only minimal primes? Do we not allow components of different dimension?

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -1,3 +1,57 @@
...
+# Printing and IO
+################################################################################
+function Base.show(io::IO, )
+  print(io, "subscheme of ℙ^$(relative_ambient_dimension(P))_{$(base_ring(P))} defined as the zero locus of  $(defining_ideal(P))")
+end
+
+function Base.show(io::IO, ::MIME"text/plain", P::AbsProjectiveScheme{<:Any, <:MPolyQuoRing})
+  println(io, "Projective scheme")  # at least one new line is needed
+  println(io, "  over ", base_ring(P))
+  println(io, "defined by")
```
HechtiDerLachs:
```suggestion
  println(io, "  defined by")
```

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -1,3 +1,57 @@
...
+    print(io, "Projective scheme in ")
+    print(IOContext(io, :supercompact => true), ambient_space(P), " over ", base_ring(P))
+  end
+end
+
+# Projective space
+function Base.show(io::IO, ::MIME"text/plain", P::AbsProjectiveScheme{<:Any, <:MPolyDecRing})
+  println(io, "Projective space of dimension $(relative_ambient_dimension(P))")  # at least one new line is needed
+  print(io, "  over ")
+  print(io, base_ring(P)) # the last print statement must not add a new line
```
HechtiDerLachs, line 30:
```suggestion
  println(io, "  over ")
  print(io, base_ring(P)) # the last print statement must not add a new line
```

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -37,14 +88,11 @@ function dehomogenization_map(
...
   }
-  if !isdefined(X, :dehomogenization_cache)
-    X.dehomogenization_cache = IdDict()
-  end
   i in 0:relative_ambient_dimension(X) || error("the given integer is not in the admissible range")
   S = homogeneous_coordinate_ring(X)
   C = standard_covering(X)
   U = C[i+1]
-  cache = X.dehomogenization_cache
+  cache = dehomogenization_cache(X)
```
HechtiDerLachs: Should we make this `_dehomogenization_cache`?

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
```diff
@@ -0,0 +1,82 @@
...
+
+
+@doc raw"""
+    projective_variety(R::Ring; check::Bool=true)
+
+Return the projective variety defined by the ``\mathbb{Z}`` graded ring ``R``.
+
+We require that ``R`` is a finitely generated algebra over a field ``k`` and
+moreover that the base change of ``R`` to the algebraic closure ``\bar k``
+is an integral domain.
```
HechtiDerLachs, line 40: Unfortunately, we also require that `R` is the quotient of a standard graded ring.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
```diff
@@ -0,0 +1,82 @@
...
+"""
+projective_variety(R::Ring; check::Bool=true) = ProjectiveVariety(ProjectiveScheme(R), check=check)
+
+projective_variety(R::MPolyDecRing; check::Bool=true) = ProjectiveVariety(ProjectiveScheme(R), check=check)
+
+function projective_variety(f::MPolyDecRingElem; check=true)
+  if check
+    is_irreducible(f) || error("polynomial is reducible")
+    ff = factor_absolute(f)[2]
+    # deal with weird type instability in factor_absolute
```
HechtiDerLachs, line 50: Is this well known, or can you reference an issue? Otherwise it is not clear whether some day this could/should be removed.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
```diff
@@ -0,0 +1,82 @@
...
+
+We require that ``R`` is a finitely generated algebra over a field ``k`` and
+moreover that the base change of ``R`` to the algebraic closure ``\bar k``
+is an integral domain.
+"""
+projective_variety(R::Ring; check::Bool=true) = ProjectiveVariety(ProjectiveScheme(R), check=check)
+
+projective_variety(R::MPolyDecRing; check::Bool=true) = ProjectiveVariety(ProjectiveScheme(R), check=check)
+
+function projective_variety(f::MPolyDecRingElem; check=true)
```
HechtiDerLachs, line 46:
```suggestion
function projective_variety(f::MPolyDecRingElem{<:FieldElem}; check=true)
```
I suspect that the methods used below will not work unless we are working over an admissible field of coefficients. Maybe, we can relax to `Union{<:FieldElem, <:ZZRingElem}`.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Types.jl
```diff
@@ -0,0 +1,16 @@
+@attributes mutable struct ProjectiveVariety{BaseRing<:Field, GradedRingType} <: AbsProjectiveVariety{BaseRing, GradedRingType}
+  X::AbsProjectiveScheme
+
+  function ProjectiveVariety(X::AbsProjectiveScheme{S, T}; check::Bool) where {S, T}
+    if check
+      S <:QQField || error("varieties must be geometrically integral, but we test this only over QQ at the moment, disable this check if you know the variety is geometrically integral or proceed at your own risk")
```
HechtiDerLachs:
```suggestion
      S <:QQField || error("varieties must be geometrically integral, but we test this only over QQ at the moment; disable this check if you know the variety is geometrically integral or proceed at your own risk")
```

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Types.jl
```diff
@@ -0,0 +1,16 @@
+@attributes mutable struct ProjectiveVariety{BaseRing<:Field, GradedRingType} <: AbsProjectiveVariety{BaseRing, GradedRingType}
+  X::AbsProjectiveScheme
+
+  function ProjectiveVariety(X::AbsProjectiveScheme{S, T}; check::Bool) where {S, T}
```
HechtiDerLachs:
```suggestion
  function ProjectiveVariety(X::AbsProjectiveScheme{S, T}; check::Bool=true) where {S, T}
```

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Types.jl
```diff
@@ -0,0 +1,16 @@
...
+    if check
+      S <:QQField || error("varieties must be geometrically integral, but we test this only over QQ at the moment, disable this check if you know the variety is geometrically integral or proceed at your own risk")
+      is_geometrically_integral(X) || error("varieties must be geometrically integral")
+    else
+      set_attribute!(X, :is_geometrically_integral => true)
+      set_attribute!(X, :is_integral => true)
+      set_attribute!(X, :is_geometrically_reduced => true)
+      set_attribute!(X, :is_reduced => true)
```
HechtiDerLachs, line 12: Alternatively once can override the getters for the corresponding abstract type. That saves the time for the dictionary lookup.

test/AlgebraicGeometry/Schemes/runtests.jl
```diff
@@ -3,21 +3,22 @@ using Test
...
+include("CoveredProjectiveSchemes.jl")
 include("CoveredScheme.jl")
+include("CoherentSheaves.jl")
 include("FunctionFields.jl")
-include("singular_locus.jl")
+include("Glueing.jl")
+include("K3.jl")
+include("ProjectiveSchemes.jl")
+include("ProjectiveVarieties.jl")
+include("IdealSheaves.jl")
```
HechtiDerLachs: This should move up, right?

HechtiDerLachs (commented):
Looks fine. 

One thing one could think about: If I remember correctly, there is a hierarchy `AbsProjectiveScheme:>AbsProjectiveAlgebraicSet:>AbsProjectiveVariety` with potentially increasing functionality. In order to inherit functionality from `AbsProjectiveAlgebraicSet` to `AbsProjectiveVariety`, the latter should contain a concrete instance of  `AbsProjectiveAlgebraicSet` rather than a plain `ProjectiveScheme`. Further, the `underlying_scheme` should return this instance of `AbsProjectiveAlgebraicSet` and not a `ProjectiveScheme`.

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Methods.jl
wdecker: Minimal primes may have different dimension and describe the irreducible components of an algebraic set. In contrast, embedded primes show only up on the geometric side if we consider schemes.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
simonbrandhorst: At least we can tell if it is not geometricaly integral like this.

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
simonbrandhorst: It is not implemented on the scheme side yet, but I figured the PR is large enough. Hence let us keep that for later.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
simonbrandhorst: It is intended that the base ring is printed in the same line.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
simonbrandhorst: Good point.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
simonbrandhorst: what is a standard graded ring?

simonbrandhorst: Every variable has the same degree?

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
simonbrandhorst: I can open an issue

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
wdecker: Usually, this means ZZ-graded with all variables of degree 1

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
simonbrandhorst: The radical knows that it is a radical.  (I added this) Hence the additional checking is trivial.
But we need it to be geometrically irreducible, and then radical is not good enough
unless we work over a perfect field.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
simonbrandhorst: Thanks. Where is this requirement coming from?  Are there intrinsic reasons? 
(I would imagine that a groebner basis of a homogeneous ideal of a non-standard graded ring is possibly not homogeneous? Or is it just a matter of picking a better monomial ordering?)

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
simonbrandhorst: Since varieties are over a field anyways, adding `ZZRingElemen` won't give us much.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Types.jl
simonbrandhorst: It has the drawback that we do not have
`has_attribute(:is_reduced)` to check if the scheme knows that it is reduced. 
We could introduce an interface for this of course e.g. 
`has_is_reduced` . But for now we have more important stuff to do.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Constructors.jl
wdecker: We require that the ideals in ANY graded ring are given by homogeneous generators. This means that any GB computed from this data has homogeneous elements as well.

simonbrandhorst: So why are we restricting `ProjectiveScheme` to standard gradings if the most important tool works?

wdecker: I have not written this code, so I can only guess: You probably want to define a projectice variety by entering its coordinate ring? If so, and if the variety is NOT STANDARD ZZ-graded, you need to consider this as a variety in a weighted projective space. Or, what else is meant?

simonbrandhorst: Exactly. Or more generally `Proj S` where `S` is a `Z`-graded ring. Most textbooks define this.

HechtiDerLachs: Yes, that's it. Non-standard gradings are corresponding to weighted projective spaces. The proj-construction is different there when it comes to passing to the affine charts. We simply don't have that functionality yet.

simonbrandhorst:
> Looks fine.
> 
> One thing one could think about: If I remember correctly, there is a hierarchy `AbsProjectiveScheme:>AbsProjectiveAlgebraicSet:>AbsProjectiveVariety` with potentially increasing functionality. In order to inherit functionality from `AbsProjectiveAlgebraicSet` to `AbsProjectiveVariety`, the latter should contain a concrete instance of `AbsProjectiveAlgebraicSet` rather than a plain `ProjectiveScheme`. Further, the `underlying_scheme` should return this instance of `AbsProjectiveAlgebraicSet` and not a `ProjectiveScheme`.

For now it does not make a difference since there is no extra functionality. 
In the long run one could also make the construtor lazy in the sense that it does not compute the radical outright but stores the ideal. Then some computations can run without computing the radical, but that is for the future.

simonbrandhorst:
@HechtiDerLachs  Thank you for the review. Once this is merged could you please introduce the jacobi criterion as requested by @wdecker ?

HechtiDerLachs (approved):

PR #2209, opened 2023-04-04
provide `character_table` functionality for `GrpAbFinGen`

ThomasBreuer:
For that, is must be made explicit that the `conjugacy_classes` of the character table objects on the Oscar side and on the GAP side correspond to each other, and the class fusions between `GrpAbFinGen` groups must be computed via the explicit embeddings.

related changes:

- renamed the `GAPGroup` field of character tables to `group`, since the value may now also be a `GrpAbFinGen`, and added a field `isomorphism` to a group in GAP; this way, all `.X` accesses happen now inside the functions that define this isomorphism
- renamed `induced_class_function` to `induce`
- added `restrict` for characters
- added `trivial_subgroup` and `index` for `GrpAbFinGen`
- added a workaround for an inconsistency in `GAP.Globals.AbelianGroup`
- use `@req` where possible

### DIFF
--- a/experimental/GModule/Cohomology.jl
+++ b/experimental/GModule/Cohomology.jl
@@ -2,6 +2,7 @@ module GrpCoh
 
 using Oscar
 import Oscar:action
+import Oscar:induce
 import Oscar:GAPWrap, pc_group, direct_product, direct_sum
 import AbstractAlgebra: Group, Module
 import Base: parent
--- a/experimental/GModule/Cohomology.jl
+++ b/experimental/GModule/Cohomology.jl
@@ -1,9 +1,9 @@
 module GrpCoh
 
 using Oscar
-import Oscar:action
-import Oscar:induce
-import Oscar:GAPWrap, pc_group, direct_product, direct_sum
+import Oscar: action
+import Oscar: induce
+import Oscar: GAPWrap, pc_group, direct_product, direct_sum
 import AbstractAlgebra: Group, Module
 import Base: parent
 
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -78,11 +78,12 @@ This is the type of (ordinary or Brauer) character tables that can delegate
 tasks to an underlying character table object in the GAP system
 (field `GAPTable`).
 
-An object of type `GAPGroup` can (but need not) be stored
-in the field `GAPGroup`.
-
 The value of the field `characteristic` determines whether the table
 is an ordinary one (value `0`) or a `p`-modular one (value `p`).
+
+A group can (but need not) be stored in the field `group`.
+If it is available then also the field `isomorphism` is available,
+its value is a bijective map from the `group` value to a group in GAP.
 
 Objects of type `GAPGroupCharacterTable` support [`get_attribute`](@ref),
 for example in order to store the already computed `p`-modular tables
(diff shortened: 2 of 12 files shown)


experimental/GModule/Cohomology.jl
```diff
@@ -2,6 +2,7 @@ module GrpCoh
 
 using Oscar
 import Oscar:action
+import Oscar:induce
 import Oscar:GAPWrap, pc_group, direct_product, direct_sum
```
fingolfin:
```suggestion
import Oscar: action
import Oscar: induce
import Oscar: GAPWrap, pc_group, direct_product, direct_sum
```

src/Groups/group_characters.jl
```diff
@@ -78,32 +78,111 @@ This is the type of (ordinary or Brauer) character tables that can delegate
...
+# If the character table stores an Oscar group `G` then
+# the columns correspond to conjugacy classes of `G`.
+# The vector `conjugacy_classes(G)` (stored or not stored in `G`) can be
+# independent of the vector of conjugacy classes of the corresponding
+# group on the GAP side.
+# Therefore we store a vector `conjugacy_classes(tbl)` via an attribute,
+# which is compatible with the `GAP.Globals.ConjugacyClasses` value of
+# the underlying GAP character table.
+function conjugacy_classes(tbl::GAPGroupCharacterTable)
+    return get_attribute!(tbl, :conjugacy_classes) do
```
fingolfin: Can't we use the `@attr` method here? (I don't know which return type to use from the top of my head, but the implementation here doesn't have a type stable return value either, so...)
```suggestion
@attr SOME_TYPE function conjugacy_classes(tbl::GAPGroupCharacterTable)
```

src/Groups/group_characters.jl
```diff
@@ -187,12 +270,12 @@ nothing
 ```
 """
 function character_table(id::String, p::Int = 0)
-    hasproperty(GAP.Globals, :CTblLib) || error("no character table library available")
+    @req hasproperty(GAP.Globals, :CTblLib) "no character table library available"
```
fingolfin, line 271: `@req` throws an argument error, that doesn't seem quite appropriate here?

Also, aren't we loading ctbllib during startup? (Though, I guess, we are not checking whether that worked... perhaps we should?

src/Groups/group_characters.jl
```diff
@@ -625,10 +719,10 @@ end
 
 ##############################################################################
 #
-length(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(tbl.GAPTable)::Int
-Oscar.nrows(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(tbl.GAPTable)::Int
-Oscar.ncols(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(tbl.GAPTable)::Int
-number_conjugacy_classes(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(tbl.GAPTable)::Int
+length(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(GAPTable(tbl))::Int
+Oscar.nrows(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(GAPTable(tbl))::Int
+Oscar.ncols(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(GAPTable(tbl))::Int
```
fingolfin: Can we drop the `Oscar.` here?
```suggestion
nrows(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(GAPTable(tbl))::Int
ncols(tbl::GAPGroupCharacterTable) = GAPWrap.NrConjugacyClasses(GAPTable(tbl))::Int
```

src/Groups/group_characters.jl
```diff
@@ -1142,33 +1239,136 @@ julia> maps = possible_class_fusions(s, t);  length(maps)
...
-function induced_class_function(chi::GAPGroupClassFunction, tbl::GAPGroupCharacterTable)
-  ind = GAP.Globals.InducedClassFunction(chi.values, tbl.GAPTable)::GapObj
-  return GAPGroupClassFunction(tbl, ind)
+function induce(chi::GAPGroupClassFunction, tbl::GAPGroupCharacterTable)
+  # If the class fusion is stored then use it.
+  known, fus = known_class_fusion(chi.table, tbl)
+  known && return induce(chi, tbl, fus)
+
+  subtbl = chi.table
+  if (!isdefined(tbl, :group)) || (!isdefined(subtbl, :group))
```
fingolfin:
```suggestion
  if !isdefined(tbl, :group) || !isdefined(subtbl, :group)
```

src/Groups/group_characters.jl
```diff
@@ -1142,33 +1239,136 @@ julia> maps = possible_class_fusions(s, t);  length(maps)
...
+2
+```
+"""
+function restrict(chi::GAPGroupClassFunction, subtbl::GAPGroupCharacterTable)
+  # If the class fusion is stored then use it.
+  known, fus = known_class_fusion(subtbl, chi.table)
+  known && return restrict(chi, subtbl, fus)
+
+  tbl = chi.table
+  if (!isdefined(tbl, :group)) || (!isdefined(subtbl, :group))
```
fingolfin:
```suggestion
  if !isdefined(tbl, :group) || !isdefined(subtbl, :group)
```

fingolfin (commented):
Great!

I have some minor remarks.

src/Groups/group_characters.jl
ThomasBreuer: Yes, we call `GAP.Packages.load` with `"ctbllib"` and several other names of GAP packages, without looking at the results.
(For `"browse"` we even add the optional argument `install=true`, but again without consequences if this fails.)
This means "load this GAP package if it is available, but do not complain if it is not available".
If we are sure that no Oscar users will run into problems because of missing GAP packages then we can check that the `GAP.Packages.load` calls return `true`, and then we can rely on the packages' availability.

Concerning `@req`:
Yes, I had also thought in which cases `ArgumentError` is the right reaction, but came to the conclusion that any error that occurs inside a function can be regarded as the consequence of a problem with the arguments.
(Well, there could also be a bug in the code of the function, but typically such bugs are not detected with error messages.)
In the situation in question, using a string as the first argument of `character_table` is not supported if the character table library is not available.

fingolfin: but if the character table library is missing then that isn't an error in the arguments, at least if the exact same input works on a different computer

fingolfin (approved):

PR #2213, opened 2023-04-05
ProjectiveVarieties more tests and fixes

simonbrandhorst:
fix bugs, include missing  test file and more tests


src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Constructors.jl
```diff
@@ -117,10 +117,10 @@ end
...
 ################################################################################
 
-function reduced_scheme(X::ProjectiveScheme)
+function reduced_scheme(X::AbsProjectiveScheme)
   I = defining_ideal(X)
   Irad = radical(I)
-  Xred = subscheme(ambient_space(X), I)
+  Xred = subscheme(ambient_space(X), Irad)
   set_attribute!(Xred, :is_reduced=>true)
   return Xred
```
HechtiDerLachs, line 125: We should think about having this return a map as well. But that is for another pr.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Properties.jl
```diff
@@ -38,8 +38,9 @@ function is_smooth(P::AbsProjectiveScheme; algorithm=:covered)
 end
 
 @attr Bool function is_irreducible(P::AbsProjectiveScheme)
+  is_empty(P) && return false  # the empty set is not irreducible
   I = defining_ideal(P)
-  return is_prime(I)
+  return is_prime(radical(I)) # topological property
```
HechtiDerLachs, line 43: Thanks for catching this! 

But maybe something like `is_primary` exists and is faster? I don't know! But let's ask around.

HechtiDerLachs (approved):

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Properties.jl
simonbrandhorst: The documentation of `is_primary` states that it computes a primary decomposition. 
No clue what is faster here. @wdecker  any advice?

PR #2215, opened 2023-04-05
Avoid linking a non-existing `docs/src` directory

joschmitt:
Fix a minor problem with the new experimental documentation: if a `docs/src` directory does not exist, building the documentation will create a dead link and then building it for a second time will error.
I made `symlink_target` an absolute path for this to work. I assume it was not a relative path for a reason?

### DIFF
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -83,7 +83,7 @@ function setup_experimental_package(Oscar::Module, package_name::String)
 
   # Set symlink inside docs/src/experimental
   symlink_link = joinpath(Oscar.oscardir, "docs/src/Experimental", package_name)
-  symlink_target = joinpath(Oscar.oscardir, "experimental", package_name, "docs/src")
+  symlink_target = joinpath(Oscar.oscardir, "experimental", package_name, "docs", "src")
 
   if !ispath(symlink_target)
     return []
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -110,11 +110,9 @@ end
 function doit(Oscar::Module; strict::Bool = true, local_build::Bool = false, doctest::Union{Bool,Symbol} = true)
 
   # Remove symbolic links from earlier runs
-  expdocdir = joinpath(Oscar.oscardir, "docs/src/Experimental")
-  old_links = filter(x -> islink(joinpath(expdocdir, x)) && x in Oscar.exppkgs,
-                     readdir(expdocdir))
-  for x in old_links
-    rm(joinpath(expdocdir, x))
+  expdocdir = joinpath(Oscar.oscardir, "docs", "src", "Experimental")
+  for x in readdir(expdocdir; join=true)
+    islink(x) && rm(x)
   end
 
   # include the list of pages, performing substitutions
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -94,7 +94,7 @@ function setup_experimental_package(Oscar::Module, package_name::String)
   elseif !islink(symlink_link) || readlink(symlink_link) != symlink_target
       error("$symlink_link already exists, but is not a symlink to $symlink_target
 Please investigate the contents of $symlink_link,
-optionally move them somewhere else and delete the directiory once you are done.")
+optionally move them somewhere else and delete the directory once you are done.")
   end
 
   # Read doc.main of package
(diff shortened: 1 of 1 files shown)


joschmitt:
CC @lkastner
(I'm not sure whether I'm too stupid to find the button or don't have the rights, but I can't request a review from you?)

joschmitt:
Now I'm wondering whether there should also be a proper tidying-up which removes the links after the documentation was built... I just switched branches and it errors because some directory doesn't exist anymore.

docs/make_work.jl
```diff
@@ -83,19 +83,24 @@ function setup_experimental_package(Oscar::Module, package_name::String)
 
   # Set symlink inside docs/src/experimental
   symlink_link = joinpath(Oscar.oscardir, "docs/src/Experimental", package_name)
-  symlink_target = joinpath("../../../experimental", package_name, "docs/src")
+  symlink_target = joinpath(Oscar.oscardir, "experimental", package_name, "docs/src")
```
lkastner: Weird, I was just told that symlink targets are always relative paths, but apparently in julia they aren't. @benlorenz

lkastner (approved):

docs/make_work.jl
joschmitt: I just made it absolute because then the `ispath`-check is easier. And it appears to work like this...

lkastner: Well, the testsuite runs and it works on my machine, and I also think that the code is more readable like this, so thank you. I probably just misunderstood something.

lkastner:
> Now I'm wondering whether there should also be a proper tidying-up which removes the links after the documentation was built... I just switched branches and it errors because some directory doesn't exist anymore.

Maybe we could just remove all symlinks in `docs/src/Experimental` right at the beginning of `doit` in `docs/make_work.jl`?
```
julia> expdocdir = joinpath(Oscar.oscardir, "docs/src/Experimental")
"/home/lars/repos/oscars/master/docs/src/Experimental"

julia> filter(x->islink(joinpath(expdocdir, x)), readdir(expdocdir))
4-element Vector{String}:
 "BasisLieHighestWeight"
 "FTheoryTools"
 "LieAlgebras"
 "PlaneCurve"
```
Then these would just have to be removed.

joschmitt:
> > Now I'm wondering whether there should also be a proper tidying-up which removes the links after the documentation was built... I just switched branches and it errors because some directory doesn't exist anymore.
> 
> Maybe we could just remove all symlinks in `docs/src/Experimental` right at the beginning of `doit` in `docs/make_work.jl`?
> 
> ```
> julia> expdocdir = joinpath(Oscar.oscardir, "docs/src/Experimental")
> "/home/lars/repos/oscars/master/docs/src/Experimental"
> 
> julia> filter(x->islink(joinpath(expdocdir, x)), readdir(expdocdir))
> 4-element Vector{String}:
>  "BasisLieHighestWeight"
>  "FTheoryTools"
>  "LieAlgebras"
>  "PlaneCurve"
> ```
> 
> Then these would just have to be removed.

Sounds good! I added a commit doing basically this.

docs/make_work.jl
```suggestion
  symlink_target = joinpath(Oscar.oscardir, "experimental", package_name, "docs", "src")
```

docs/make_work.jl
```diff
@@ -104,6 +109,14 @@ end
 
 function doit(Oscar::Module; strict::Bool = true, local_build::Bool = false, doctest::Union{Bool,Symbol} = true)
 
+  # Remove symbolic links from earlier runs
```
fingolfin, line 112: Good, but it won't help with switching to other branches, will it? For that I think we'd also need to clean all symlinks after building the manual?

fingolfin (commented):
Thank you!

docs/make_work.jl
joschmitt: I'm not sure I see your point? I was thinking whether one should clean up before or after the build and doing it before seemed a bit more reliable, but I can't say I have a strong opinion.

lkastner: Am I getting the order of things wrong? I thought once you run `build_doc` it would now remove the symlinks and create new ones. Or is it already stumbling on the issue with the broken symlinks before?

fingolfin: @joschmitt you wrote:
> I just switched branches and it errors because some directory doesn't exist anymore.

There are (at least) two different errors you might refer to here, and I probably guessed wrongly
1. you switch successfully from one branch to another and try to build docs, but it fails due to a bad symlink being present; your patch here will deal with that because it clears it before switching
2. switching could fail because something that's a symlink now used to be an actual file, and switching won't work because the symlink is in the way. To solve that we need to also clean symlinks *after* building the documentation. That's the kind of problem I was thinking about. One may argue that it is not so important because it will go away over time anyway (unless someone tries to `git bisect` a regression).

docs/make_work.jl
```diff
@@ -104,6 +109,14 @@ end
...
+  expdocdir = joinpath(Oscar.oscardir, "docs/src/Experimental")
+  old_links = filter(x -> islink(joinpath(expdocdir, x)) && x in Oscar.exppkgs,
+                     readdir(expdocdir))
+  for x in old_links
+    rm(joinpath(expdocdir, x))
+  end
```
```suggestion
  expdocdir = joinpath(Oscar.oscardir, "docs", "src", "Experimental")
  for x in readdir(expdocdir; join=true)
    islink(x) && rm(x)
  end
```

fingolfin (approved):

docs/make_work.jl
joschmitt: I was talking about problem 1, sorry for the misunderstanding.

lkastner:
Thank you and sorry for the mess.

PR #2216, opened 2023-04-05
Adjust `LinearQuotients` to new structure of `experimental`

joschmitt:
... and restructure the code a bit. Nothing changed fundamentally.


experimental/LinearQuotients/docs/src/cox_rings.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+# Cox rings
+
+## Cox rings of linear quotients
+
+By a theorem of Arzhantsev and Gaifullin [AG10](@cite), the Cox ring of a linear
+quotient $V/G$ is graded isomorphic to the invariant ring $K[V]^{[G,G]}$, where
+$[G,G]$ is the derived subgroup of $G$.
+This functionality is so far only available, if the group does not contain any
```
fingolfin:
```suggestion
This functionality is so far only available if the group does not contain any
```

fingolfin (approved):
Very nice, thanks!

PR #2115, opened 2023-03-24
Import code from the project basisLieHighestWeight

BenWilop:

PR #2217, opened 2023-04-05
Coxeter groups

ulthiel:
This is an empty draft pull request to indicate that I would really like to bring the package https://github.com/ulthiel/CoxeterGroups.jl for computation in Coxeter groups into OSCAR. This is work done by @joelgibson,  @CameronBraunstein, and @schto223. It currently features:

* Computation in Coxeter groups with arbitrary Coxeter matrix via a recursive algorithm by Casselman. (By "computation" I always mean writing elements in a unique normal form.)
* Computation in Coxeter groups for a generalized Cartan matrix via minimal roots algorithm. This is at a speed comparable to Magma (which uses the same algorithm basically). I think it can be made to work for general Coxeter matrices as well.
* Computation in the symmetric group as Coxeter group (generated by the transpositions). Very fast and useful for this special case.
* Structures for Coxeter matrices and Coxeter systems, especially "identifying" the type.

Not everything is OSCAR ready and perfect but it's basically there and shouldn't lie around any longer. Help integrating this is greatly welcome!

It would be optimal if the current open PRs #2207 and #2115 on Lie theory will be somewhat compatible (in either direction).


fingolfin:
Closing in favor of PR #2689

PR #2227, opened 2023-04-06
Stop importing `exponent_vectors` from AbstractAlgebra

### LINKED ISSUE #1903: Make `exponent_vectors` use the same ordering as `coefficients` etc.
#1696 introduced new functions for `coefficients`, `terms`, etc. which now use the `default_ordering` of the ring. However, there is still `exponent_vectors` from AbstractAlgebra which keeps using "some" ordering.
In consequence the output of `coefficients` and `exponent_vectors` might not match up anymore:
```
julia> R, x = GradedPolynomialRing(QQ, ["x1", "x2"], [1, 2])
(Multivariate Polynomial Ring in x1, x2 over Rational Field graded by 
  x1 -> [1]
  x2 -> [2], MPolyElem_dec{fmpq, fmpq_mpoly}[x1, x2])

julia> f = x[1] + 2*x[2]
x1 + 2*x2

julia> collect(coefficients(f))
2-element Vector{fmpq}:
 2
 1

julia> collect(exponent_vectors(f))
2-element Vector{Vector{Int64}}:
 [1, 0]
 [0, 1]

```
Could we either make `exponent_vectors` an alias for the new function `exponents` or at least not export it anymore (it appears to be imported from Hecke)? Like this it is just an accident waiting to happen, especially because the relevant documentation was deleted in c3d516a5916bc4c6a2e5768eeb408c50344795b4; see the deletions in `orderings.md`.
(In fact, that accident already happened in code of mine which was running fine until I updated...)

joschmitt:
Closes #1903 .
I might have overlooked a place where it is used, so lets see the CI run.
~~We could also make `exponent_vectors` an alias for `exponents`?~~
Also, should there be a documentation on this as discussed in #1903 ?

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
@@ -296,7 +296,7 @@ function Base.show(io::IO, ac::RationalEquivalenceClass)
       # otherwise, extract properties to represent the rational equivalence class
       r = representative(ac)
       coeffs = [c for c in AbstractAlgebra.coefficients(r)]
-      expos = [matrix(ZZ, [k for k in exponent_vectors(m)]) for m in AbstractAlgebra.monomials(r)]
+      expos = [matrix(ZZ, [k for k in AbstractAlgebra.exponent_vectors(m)]) for m in AbstractAlgebra.monomials(r)]
       indets = gens(chow_ring(toric_variety(ac)))
 
       # form string to be printed
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 
 [compat]
-AbstractAlgebra = "0.29.2"
+AbstractAlgebra = "0.29.3"
 AlgebraicSolving = "0.3.0"
 DocStringExtensions = "0.8, 0.9"
 GAP = "0.9.4"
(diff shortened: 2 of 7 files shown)


joschmitt:
I created the alias now because some tests use the conformance test scripts from AbstractAlgebra.

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
```diff
@@ -296,7 +296,7 @@ function Base.show(io::IO, ac::RationalEquivalenceClass)
       # otherwise, extract properties to represent the rational equivalence class
       r = representative(ac)
       coeffs = [c for c in AbstractAlgebra.coefficients(r)]
-      expos = [matrix(ZZ, [k for k in exponent_vectors(m)]) for m in AbstractAlgebra.monomials(r)]
+      expos = [matrix(ZZ, [k for k in AbstractAlgebra.exponent_vectors(m)]) for m in AbstractAlgebra.monomials(r)]
```
HereAround, line 299: With your changes, maybe also the following works?
```suggestion
      expos = [matrix(ZZ, [k for k in exponents(m)]) for m in AbstractAlgebra.monomials(r)]
```

joschmitt: Seeing that `m` is a monomial anyway and there is hence no monomial ordering involved, I guess so. `AbstractAlgebra.exponent_vectors` will be faster though, from what I understand.

thofma:
Hm, keeping the alias around does not feel right. What exactly goes wrong? Maybe we can find a better way.

joschmitt:
In `test/runtests.jl` the ring conformance test script from AbstractAlgebra is included which is then used to test the interface conformance. As far as I can see, the script assumes to work in the AbstractAlgebra namespace, but we use it in Oscar. At least the tests in `mpoly-localizations.jl` fail, if we don't have `exponent_vectors`.

thofma:
Yes, I opened https://github.com/Nemocas/AbstractAlgebra.jl/pull/1335 to fix the test. Then the alias won't be necessary anymore.

thofma (approved):

PR #2229, opened 2023-04-06
Resolving various issues

wdecker:
adding to PR #2225


wdecker:
I do not understand why the docu test works fine, but some other tests show diffs in the documentation.

fingolfin:
The docu tests were changed to only build the documentation, not test it.

fingolfin (approved):

wdecker:
Why? That means if I have mistakes I have to wait, say, 130 minutes before I see them, otherwise it would be 30 minutes.

PR #2230, opened 2023-04-06
Add some missing functions in PolyhedralGeometry

lkastner:
- `is_simplicial(Cone)`
- `is_smooth(Cone)`
- `Base.in(Cone, Cone)`
- `Base.in(Polyhedron, Polyhedron)`
- `contains(Polyhedron, Vector)` changed to `Vector in Polyhedron`
- `contains(Cone, Vector)` changed to `Vector in Cone`

### DIFF
--- a/src/PolyhedralGeometry/Cone/properties.jl
+++ b/src/PolyhedralGeometry/Cone/properties.jl
@@ -558,23 +558,23 @@ Base.in(C0::Cone{T}, C1::Cone{T}) where T<:scalar_types = Polymake.polytope.incl
 
 
 @doc raw"""
-    contains(C::Cone, v::AbstractVector)
+    in(v::AbstractVector, C::Cone)
 
-Check whether `C` contains `v`.
+Check whether the vector `v` is contained in the cone `C`.
 
 # Examples
 The positive orthant only contains vectors with non-negative entries:
 ```jldoctest
 julia> C = positive_hull([1 0; 0 1]);
 
-julia> contains(C, [1, 2])
+julia> [1, 2] in C
 true
 
-julia> contains(C, [1, -2])
+julia> [1, -2] in C
 false
 ```
 """
-contains(C::Cone, v::AbstractVector) = Polymake.polytope.contains(pm_object(C), v)::Bool
+Base.in(v::AbstractVector, C::Cone) = Polymake.polytope.contains(pm_object(C), v)::Bool
 
 
 @doc raw"""
--- a/src/PolyhedralGeometry/Cone/properties.jl
+++ b/src/PolyhedralGeometry/Cone/properties.jl
@@ -349,7 +349,7 @@ is_simplicial(C::Cone) = pm_object(C).SIMPLICIAL_CONE::Bool
 
 
 @doc raw"""
-    is_smooth(C::Cone)
+    is_smooth(C::Cone{QQFieldElem})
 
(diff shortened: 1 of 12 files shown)


src/PolyhedralGeometry/Cone/properties.jl
```diff
@@ -486,6 +534,29 @@ _generator_matrix(::Val{_hilbert_generator}, C::Polymake.BigObject; homogenized=
 _matrix_for_polymake(::Val{_hilbert_generator}) = _generator_matrix
 
 
+@doc raw"""
+    contains(C0::Cone, C1::Cone)                           
+    
+Check whether `C0` contains `C1`.
```
fingolfin: Contains in which sense? As a set, as a subcone, or something else?

src/PolyhedralGeometry/Cone/properties.jl
```diff
@@ -486,6 +534,29 @@ _generator_matrix(::Val{_hilbert_generator}, C::Polymake.BigObject; homogenized=
...
+
+julia> contains(C1, C0)
+false
+```
+"""
+contains(C0::Cone{T}, C1::Cone{T}) where T<:scalar_types = Polymake.polytope.included_polyhedra(pm_object(C1), pm_object(C0))::Bool
+
+
 @doc raw"""
     contains(C::Cone, v::AbstractVector)
```
fingolfin: BTW I think this `contains` really should be `in`

src/PolyhedralGeometry/Cone/properties.jl
lkastner: A subset, I clarified.

src/PolyhedralGeometry/Cone/properties.jl
lkastner: I also changed this for the vector containment.

antonydellavecchia (approved):

src/PolyhedralGeometry/Cone/properties.jl
```diff
@@ -487,23 +535,46 @@ _matrix_for_polymake(::Val{_hilbert_generator}) = _generator_matrix
 
 
 @doc raw"""
-    contains(C::Cone, v::AbstractVector)
+    in(C0::Cone, C1::Cone)                           
+    
+Check whether `C0` is contained in `C1` as a subset.
```
fingolfin, line 540: A subset is not an element of a set, though. So this does not seem like an appropriate `in` method? Rather sounds like it should be a method for `issubset`?

fingolfin: That would also remove the need to write "as a subset" in the docstring, it would be obvious from the function name

lkastner: https://github.com/oscar-system/Oscar.jl/pull/2230#discussion_r1161150807

Sorry, I think I misunderstood the above comment, I thought it referred to the first line of the code block. Anyway I'll fix it.

fingolfin: No worries. I've spent easter with fever and am slowly recovering right now so I probably am not very coherent:-/.

HereAround:
Sorry for being so late. (Vacation yesterday and afk during the Easter break.)

Thank you very much for working on this @lkastner. I like the changes!

PR #2234, opened 2023-04-07
Generic kernel

jankoboehm:
Make the kernel for module homomorphisms fully generic, reducing the number of implementations required by the generic abstraction layer by one, to essentially now three. Remove signature not necessary any more.


jankoboehm:
A test in AlgebraicGeometry/Schemes/CoherentSheaves.jl leads to Illegal instruction signal in computation of presentation.

simonbrandhorst:
> A test in AlgebraicGeometry/Schemes/CoherentSheaves.jl leads to Illegal instruction signal in computation of presentation.

Is this something that needs work on the schemes side, or the test just happens to catch some bug somewhere else?

fieker:
resolved

PR #2236, opened 2023-04-09
Graded subquotient modules

jankoboehm:
This defines the user interface for the graded subquotient modules, provides tests for the functionality, and implements a backend to make everything work. 

The functionality includes graded free modules, graded submodules, graded subquotients, homomorphisms thereof, graded tensor product for graded free modules and subquotients, as well as graded direct product, graded direct sum, graded hom modules, tensor product of modules and resolutions, tensor product of resolutions and modules, hom for module and resolution, hom for resolution and module, graded Ext, graded Tor, Groebner bases, graded free resolutions, Betti hash tables, Betti diagrams, tensor product of morphisms. 

The backend is one of several we have. It has proven by far the most elegant one, since it is generic, makes all the higher functionality work directly in the graded case, and thus avoids code duplication with the ungraded case to almost 100%. There is also a version using the dec types, which will be used for the filtered case, where we do not put degrees through degrees on the unit vectors of the embedding free module, as in the graded case, but on the generators of the module. The code still needs some polishing and additions, in particular to update to the recent concept of show functions, but in most places it implements at least compact and standard for the moment.


src/Modules/UngradedModules.jl
```diff
@@ -146,10 +171,14 @@ Here, `F` and `G` are equal iff their base rings, ranks, and names for printing
 function (==)(F::FreeMod, G::FreeMod)
   # two free modules are equal if the rank and the ring are
   # TODO it this enough or e.g. stored morphisms also be considered?
+  if is_graded(F) && is_graded(G) 
+    return F.R == G.R && F.d == G.d && F.S == G.S
```
lgoettgens, line 178: What is with the case where one compares a graded with an ungraded module? Is this supposed to forget the grading on the first?

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+function is_graded(M::FreeMod)
+  return isa(M.d, Vector{GrpAbFinGenElem})
+end
+
+###############################################################################
+# Graded Free Modules functions
+###############################################################################
+
+# for isomorphism of graded free modules, perhaps exists somewhere
+function compare_vectors_as_multisets(v1::Vector{T}, v2::Vector{T}) where {T<:Any}
```
fieker: ```MSet(v1) == MSet(v2)```
MSets exist

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+
+function betti(b::FreeResolution; reverse_direction::Bool = false)
+	return betti_table(b, project = nothing, reverse_direction = reverse_direction)
+end
+
+function dict(b::BettiTable)
+  return b.B
+end
+
+function reverse_direction!(b::BettiTable)
```
fieker, line 501: ```b.reverse_direction = !b.reverse_direction```

src/Modules/UngradedModules.jl
```diff
@@ -118,11 +109,45 @@ function AbstractAlgebra.extra_name(F::FreeMod)
 end
 
 function show(io::IO, F::FreeMod)
-  @show_name(io, F)
-  @show_special(io, F)
+  compact = get(io, :compact, false)
```
fieker, line 114: the 2 wierd show calls should stay.  show_name will print the variable name if assigned, show_special allows the override of the show function (free modules representing homogenous component stuff). show name also checks for compact

src/Modules/UngradedModules.jl
```diff
@@ -161,6 +190,9 @@ Return  `true` if `F` and `G` are isomorphic, `false` otherwise.
 Here, `F` and `G` are isomorphic iff their base rings and ranks are equal.
 """
 function is_isomorphic(F::FreeMod, G::FreeMod)
+  @assert rank(F) == rank(G)
```
fieker: shouldn't this just return false? Why is it an error?

src/Exports.jl
```diff
@@ -423,6 +428,7 @@ export describe
 export desimulate_valuation
 export det
 export diameter
+export dict
```
fingolfin: What's `dict` and why do we export it?

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
 
 
+###############################################################################
+# Graded Modules constructors
+###############################################################################
+
+@doc Markdown.doc"""
```
fingolfin:
```suggestion
@doc raw"""
```
everywhere

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+    graded_free_module(R::CRing_dec, n::Int, name::String = "e"; cached::Bool = false)
```
fingolfin:
```suggestion
    graded_free_module(R::Ring, n::Int, name::String = "e"; cached::Bool = false)
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+      v2_dict[entry] += 1
+    else
+      v2_dict[entry] = 1
+    end
+  end
+  return v1_dict == v2_dict
+end
+
+
+function swap!(A, i, j)
```
fingolfin: Type annotations would help to make clear what this function does. I am guessing:
```suggestion
function swap!(A::Vector, i::Int, j::Int)
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+  A[i], A[j] = A[j], A[i]
+end
+
+function generate(k, A)
```
fingolfin: What does this function "generate"? What is `A`? Type annotations plus a little comment would be helpful. I am guessing
```suggestion
function generate(k::Int, A::Vector)
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+function dict(b::BettiTable)
+  return b.B
+end
+
+function reverse_direction!(b::BettiTable)
+  if b.reverse_direction
+    b.reverse_direction = false
+  else
+    b.reverse_direction = true
+  end
```
fingolfin: I think @fieker meant this:
```suggestion
  b.reverse_direction = ! b.reverse_direction
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+      end
+  end
+  return BettiTable(generator_count, project = project, reverse_direction = reverse_direction)
+end
+
+function betti(b::FreeResolution; reverse_direction::Bool = false)
+	return betti_table(b, project = nothing, reverse_direction = reverse_direction)
+end
+
+function dict(b::BettiTable)
```
fingolfin: This is a meaningless function name and not even with a comment. If `b.B` is a purely internal table there is no point having this access method.

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+      new_key = (i, key-i*A[1])
+      new_B[new_key] = value
+  end
+  return new_B
+end
+
+function Base.show(io::IO, b::BettiTable)
+  T = induce_shift(b.B)
+  x = collect(keys(T))
+  isempty(x) && (println("Empty table"); return)
```
fingolfin:
```suggestion
  isempty(x) && (println(io, "Empty table"); return)
```
Stylistically, I'd really prefer to spell it out:
```suggestion
  if isempty(x)
    println(io, "Empty table")
    return
  end
```
Perhaps it should also say "Empty Betti table"?

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+  step, min, max = b.reverse_direction ? (-1, maximum([x[i][1] for i in 1:length(x)]), minimum([x[i][1] for i in 1:length(x)])) : (1, minimum([x[i][1] for i in 1:length(x)]), maximum([x[i][1] for i in 1:length(x)]))
+  s1 = ndigits(max)
+  sum0 = sum([getindex(T, x[i]) for i in 1:length(x)])
+  s3 = ndigits(sum0)
+  if b.project == nothing
+      for i in 1:ngens(parent(x[1][2]))
+          s2 = ndigits(maximum(x[j][2][i] for j in 1:length(x)))
+          spaces = maximum([s2, s1, s3])
+          ngens(parent(x[1][2])) > 1 && print(io, "Betti Table for component ", i, "\n")
```
fingolfin:
```suggestion
          ngens(parent(x[1][2])) > 1 && println(io, "Betti Table for component ", i)
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+          print(io, "-" ^ divider_width)
+          print(io, "\n", "total: ")
+          for i in min:step:max
+            sum0 = sum(getindex(T, x[j]) for j in 1:length(x) if x[j][1] == i)
+            print(io, sum0, " "^(spaces - ndigits(sum0) + 1))
+          end
+          print(io, "\n", "\n")
+      end
+  else 
+      parent(b.project)==parent(x[1][2]) || return error("projection vector has wrong type")
```
fingolfin:
```suggestion
      parent(b.project)==parent(x[1][2]) || error("projection vector has wrong type")
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+          mi = minimum(L)
+          mx = maximum(L)
+          for j in mi:mx
+              print(io, j, " "^(s2 - ndigits(j) + (5 - s2)))
+              print(io, ": ")
+              for h in min:step:max
+                sum0 = sum([getindex(T, x[k]) for k in 1:length(x) if x[k][1] == h && x[k][2][i] == j])
+                print(io, sum0 == 0 ? "-" : sum0)
+                print(io, " "^(spaces - (sum0 == 0 ? 0 : ndigits(sum0)) + (sum0 == 0 ? 0 : 1)))
+                sum0 = 0
```
fingolfin: You keep reusing `sum0`, sometimes it's an `fmpz`, sometimes an `Int` -- that totally wrecks type stability of this function. Which may not matter, of course, printing is slow anyway. I just wanted to make you aware -- it can be beneficial to use different names for temporary values. E.g. in this loops, I don't see a reason to set `sum0 = 0`, and in fact the value `sum0` does not escape the loop if you just replace `sum0` in it by a unique name, like `sum1`.

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+  return new_B
+end
+
+function Base.show(io::IO, b::BettiTable)
+  T = induce_shift(b.B)
+  x = collect(keys(T))
+  isempty(x) && (println("Empty table"); return)
+  step, min, max = b.reverse_direction ? (-1, maximum([x[i][1] for i in 1:length(x)]), minimum([x[i][1] for i in 1:length(x)])) : (1, minimum([x[i][1] for i in 1:length(x)]), maximum([x[i][1] for i in 1:length(x)]))
+  s1 = ndigits(max)
+  sum0 = sum([getindex(T, x[i]) for i in 1:length(x)])
```
fingolfin: Any loop or list comprehension doing `for i in 1:length(x)` and then *only* uses `i` in the subexpression `x[i]` is usually "better" when rewritten as `for a in x` and then replacing `x[i]` by `a`. With "better" I mean the code gets shorter and arguably easier to read. E.g.
```suggestion
  sum0 = sum([getindex(T, a) for a in x])
```
or even
```suggestion
  sum0 = sum(getindex(T, a) for a in x)
```

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+
+function graded_free_module(R::Ring, W::Vector{<:IntegerUnion})
+  @assert is_graded(R)
+  A = grading_group(R)
+  return graded_free_module(R, length(W), "e", [W[i] * A[1] for i in 1:length(W)])
+end
+
+# TODO: move
+function is_graded(R::MPolyRing)
+  return false
```
fingolfin: See also PR #2241.

fingolfin (commented):
Some quick surface remarks, I did not have a deeper look at this yet (still ill)

HereAround:
The failures in 1.6 (ubuntu-latest and macOS-latest) are - as far as I can tell - because printouts in the doctests do not match with the expected outputs. If you dare, run `Oscar.build_doc(doctest = :fix)` in `Julia 1.6.x` to try to fix them. This will attempt to fix the doctests. Be careful with this option and check the changes thoroughly! (Cf. https://docs.oscar-system.org/stable/DeveloperDocumentation/documentation/#build_doc).

src/Modules/ModulesGraded.jl
```diff
@@ -1,5 +1,696 @@
...
+  find_bijections(v_dict, w_dict, 1, bijections, Dict{Int,Int}())
+  return all_bijections ? bijections : (isempty(bijections) ? nothing : bijections[1])
+end
+
+###############################################################################
+# Graded Free Module elements functions
+###############################################################################
+
+function degree(el::FreeModElem)
+  !is_graded(parent(el)) && error("The parent module is not graded.")
```
HereAround, line 174:
```suggestion
  @req is_graded(parent(el)) "The parent module is not graded"
```
`@req` can also be used in other places.

The majority of the error messages in `OSCAR` do not end with a period. Indeed, I had the impression that there was an agreement for this at some point.

src/Modules/UngradedModules.jl
jankoboehm: Thanks for spotting it.

src/Modules/ModulesGraded.jl
jankoboehm: Use MSet now.

src/Modules/ModulesGraded.jl
jankoboehm: Done

src/Modules/UngradedModules.jl
jankoboehm: Added it back in.

src/Modules/UngradedModules.jl
jankoboehm: Yes, thanks for spotting it.

src/Exports.jl
jankoboehm: The Betti table contains internally a dictionary, which is also relevant not only for internal use, but as a mathematical object in its own right. It assigns tuples of homological degrees and degrees the corresponding ranks in the resolution (without the degree shift of the Betti table). I have renamed the function. We will of course add docu.

src/Modules/ModulesGraded.jl
jankoboehm: Thanks for spotting, as far as I see it was the only spot. Should also throw an error by now.

src/Modules/ModulesGraded.jl
jankoboehm: Added type annotations for the functions finding the bijections between graded free module degrees.

src/Modules/ModulesGraded.jl
jankoboehm: Improved the Betti table code, further improvements for the case of using a projection will come.

src/Modules/ModulesGraded.jl
jankoboehm: Yes, doc tests still need an update. There is still an issue with compact printing of polynomial rings, which should be resolved.

fingolfin:
@jankoboehm seems doctests still failed.

But of course now we switched to run doctests with Julia >= 1.7 (see email to oscar-dev), so perhaps you can give it another try using a newer Julia?

I would recommend to use a *clean* environment for this. That is, switch to the Oscar directory, then start Julia with a new project, and instruct it to fix the docs, like so
```
julia --proj=$(mktemp -d) -e 'using Pkg ; Pkg.develop(PackageSpec(path=pwd())) ; using Oscar ; Oscar.build_doc(doctest=:fix)'
```
That will run 10-30 minutes depending on your machine.


If you still can't make this work, let me know, then I can fix it in this PR.

jankoboehm:
Thanks a lot! Did run it already on 1.7. It seems that build_doc(doctest=:fix) now fixes exactly what is still wrong after the last run on 1.6. Let's see what happens. If it does not work I'll do it over in a clean environment.

wdecker (approved):

PR #2237, opened 2023-04-10
Weierstrass Preparation Theorem 

Gebi95:
This is my first pull request. It is a porting from the weierstr.lib from Singular. 

(Caution: Weierstrass Preparation Theorem and Weierstrass Division Theorem are not related to Weierstrass Form of Elliptic curves)

### DIFF
--- a/experimental/Rings/WeierstrassPreparationTheorem.jl
+++ b/experimental/Rings/WeierstrassPreparationTheorem.jl
@@ -1,0 +1,214 @@
+##############################################################
+#
+#Conversion from Singular (weierstr.lib)
+#
+#############################################################
+
+"""
+`order_of_generality(f::MPolyElem)`
+
+compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` with `x_n` the n-th variable of the polynomial ring of `f` 
+"""
+function order_of_generality(f::MPolyElem,n::Int)
+	 
+	vect=fill(parent(f)(0),nvars(parent(f))) 
+	vect[n]=gens(parent(f))[n] 
+	f=evaluate(f,vect) 
+	f != 0 || return false,-1
+	return true,degree(leading_monomial(f,ordering=neglex(gens(parent(f)))),n)
+end
+
+"""
+`jet(f::MPolyElem,de::Int)`
+
+Given a polynomial `f` and an integer degree `de` , this function returns the polynomial, which contains only the monomials of `f`, which total degree is lower than or equal to `de`
+"""
+function jet(f::MPolyElem,de::Int) 
+	while f!=0
+		if total_degree(leading_monomial(f,ordering=deglex(gens(parent(f)))))>de
+			f=f-coeff(f,leading_monomial(f,ordering=deglex(gens(parent(f)))))*leading_monomial(f,ordering=deglex(gens(parent(f))))
+		else 
+			break
+		end
+	end
+	return f
+end
+
+
+"""
+`jet(f::MPolyElem,de::Int,vecw::Vector{Int})` 
+
+Given a polynomial `f`, an integer degree `de` and an integer vector `vecw`, which has to have the same length as length(parent(f)), this function returns a polynomial, which only contains monomial, whose exponents taken as a vector multiplied (scalar product) with the given weight-vector `intw` are lower than or equal to `de`
+"""
+function jet(f::MPolyElem,de::Int,vecw::Vector{Int}) 
+
+	vect=collect(exponent_vectors(f))
+	i=length(f)
+	while i>=1
+		if transpose(vecw)*vect[i]>de
+			f=f-term(f,i)
+		end
+		i=i-1	
+	end
+	return f
+end
+
+"""
+`invert_unit(u::MPolyElem,n::Int)`
+If `u` is an unit, returning boolean value for success and the inverse of `u`. Else return false and the zero polynomial. 
+"""
+function invert_unit(u::MPolyElem,n::Int)
+	if u!=0 && total_degree(leading_monomial(u,ordering=negdegrevlex(gens(parent(u)))))==0
+		u0=jet(u,0)
+		u=jet(1-divexact(u,u0),n)
+		ui=u
+		v=1+u
+		if u!=0
+			i=div(n,total_degree(leading_monomial(u,ordering=negdegrevlex(gens(parent(u))))))
+			while i > 1
+				ui=jet(ui*u,n)
+				v=v+ui
+				i=i-1
+			end
+		end
+		v=divexact(jet(v,n),u0)
+		return true,v
+	else
+		return false, parent(u)(0)
+	end
+end
+
+"""
+`last_var_general(f::MPolyElem,n::Int)`
+
+transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable.
+"""
+function var_general(f::MPolyElem,n::Int)
+	return var_general(f,n,0)
+end
+
+"""
+`var_general(f::MPolyElem,n::Int,voice::Int)`
+
+transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable. `voice` is the counter of recursive calls (like nesting_depth). First call with `voice`=0
+"""
+function var_general(f::MPolyElem,n::Int,voice::Int) 
+	R=parent(f)
+	j=nvars(R)
+	success,d=order_of_generality(f,n) 
+	if d>0 
+		return f	
+	else 
+		m=gens(R) 
+		g=initial_form(f)
+		i=1
+		while i<=j 
+			if i!=n
+				m_hilf=m
+				m_hilf[i]=R(0)
+				if length(g) > length(evaluate(g,m_hilf))
+					m[i]=gens(R)[i]+rand(1-(voice)*10:1+(voice)*10)*gens(R)[n]
+					g=evaluate(f,m)
+					break
+				end
+			end
+			i=i+1
+		end
+		
+		if voice <= 3
+			return var_general(g,n,voice+1)
+		end
+		if voice == 4
+			i=1
+			while i<=j 
+				if i!=n
+					m[i]=gens(R)[i]+gens(R)[n]*rand(1:1000)
+				end				
+				i=i+1
+			end 
+			g=evaluate(f,m)
+			return var_general(g,n,voice+1)
+		else
+			i=1
+			while i<=j 
+				if i!=n
+					m[i]=gens(R)[i]+gens(R)[n]^rand(2:voice*d)
+				end	
+				i=i+1
+			end
+			g=evaluate(f,m)
+			return var_general(g,voice+1)
+		end
+	end		
+end
+
+
+"""
+`initial_form(f::MPolyElem)` 
+
+transform `f` into an initial form for the algorithm
+"""
+function initial_form(f::MPolyElem)
+	de=total_degree(leading_monomial(f,ordering=negdegrevlex(gens(parent(f))))) 
+	g=jet(f,de) 
+	return g
+end
+
+"""
+`weierstrass_division_theorem(g::MPolyElem,f::MPolyElem,de::Int)`
+
+Given a multivariate polynomial `f`, which is general of oder `d` in the last ring variable, a polynomial `g` and an integer degree `de`, performing the weierstrass division theorem of `f` up to degree `de`, by dividing `f` by `g`. Returning a polynomial `u` and a remainder `r` up to degree `de`, s.t. f=ug+r. Additional the number of iterations will be returned. 
+"""
+function weierstrass_division_theorem(g::MPolyElem,f::MPolyElem,de::Int,n::Int)
+	R=parent(f)
+	r=R(0)
+	vect=fill(0,nvars(R))
+	vect[n]=1 
+	success,d=order_of_generality(f,n) 
+	
+	D=de+d
+	fhat=jet(f,d-1,vect) 
+	ftilde= divexact((f-fhat),gens(R)[n]^d)
+	succ,u=invert_unit(ftilde,D) 
+	
+	j=g 
+	jhat=jet(j,d-1,vect) 
+	jtilde=divexact(j-r,gens(R)[n]^d)
+	r=jhat
+	h=jtilde
+	i=0
+	while length(j) >0
+
+		j=jet(-fhat*u*jtilde,D) 
+		jhat= jet(j,d-1,vect)
+		jtilde=divexact(j-jhat,gens(R)[n]^d)
+		r=r+jhat
+		h=h+jtilde
+		i=i+1
+	
+	end
+	return jet(u*h,de),jet(r,de),i
+
+end
+
+"""
+`weierstrass_preparation_theorem(f::MPolyElem,de::Int)`
+
+Given a multivariate polynomial `f` and an integer degree `de`, performing the weierstrass preparation theorem of `f` up to degree `de`. Returning an unit `u` and a weierstrass polynomial `fhat`, s.t. `u``fhat`=`f`, while `fhat` is weierstrass. Additional the number of iterations used will be returned. 
+"""
+function weierstrass_preparation_theorem(f::MPolyElem,de::Int,n::Int)
+	
+	if f==0
+		throw(DomainError(f, "polynomial is zero"))
+	end
+	R=parent(f)
+	success,d=order_of_generality(f,n)
+	if success!=true
+		f=var_general(f,n,0) 
+		success,d=order_of_generality(f,n)
+	end
+	u,f2,iter_count=weierstrass_division_theorem(gens(R)[n]^d,f,de,n) 
+	return u,gens(R)[n]^d-f2,iter_count
+
+end
+
--- a/experimental/Rings/WeierstrassPreparationTheorem.jl
+++ b/experimental/Rings/WeierstrassPreparationTheorem.jl
@@ -1,6 +1,6 @@
 ##############################################################
 #
-#Conversion from Singular (weierstr.lib)
+# Conversion from Singular (weierstr.lib)
 #
 #############################################################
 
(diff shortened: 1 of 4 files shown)


experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,214 @@
...
+		else
+			i=1
+			while i<=j 
+				if i!=n
+					m[i]=gens(R)[i]+gens(R)[n]^rand(2:voice*d)
+				end	
+				i=i+1
+			end
+			g=evaluate(f,m)
+			return var_general(g,voice+1)
```
lkastner: These two blocks look identical except for the line `m[i]=...` and the return statement. Any chance of refactoring this two make a small function for the inner loop?

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,214 @@
...
+transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable.
+"""
+function var_general(f::MPolyElem,n::Int)
+	return var_general(f,n,0)
+end
+
+"""
+`var_general(f::MPolyElem,n::Int,voice::Int)`
+
+transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable. `voice` is the counter of recursive calls (like nesting_depth). First call with `voice`=0
```
lkastner: If `voice` "is the counter of recursive calls (like nesting_depth)", why not call it `nesting_depth` immediately? `voice` definitely is a non-standard name and makes the code harder to read.

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,214 @@
...
+#
+#Conversion from Singular (weierstr.lib)
+#
+#############################################################
+
+"""
+`order_of_generality(f::MPolyElem)`
+
+compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` with `x_n` the n-th variable of the polynomial ring of `f` 
+"""
```
lkastner: The docs need to be brushed up a bit, please have a look at
https://docs.oscar-system.org/dev/DeveloperDocumentation/documentation/
Also please wrap the lines to max 90-100 chars, I think that is the number we have in the style guide.

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,214 @@
...
+end
+
+"""
+`last_var_general(f::MPolyElem,n::Int)`
+
+transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable.
+"""
+function var_general(f::MPolyElem,n::Int)
+	return var_general(f,n,0)
+end
```
lkastner:
```suggestion
var_general(f::MPolyElem,n::Int) = var_general(f,n,0)
```

lkastner (commented):
A few general points, probably you already have them on your agenda, just to make sure:
- There are no tests yet.
- Nothing is exported yet.
- Maybe add some spaces here and there? E.g. `f = evaluate(...)` instead of `f=evaluate(...)`. Also after `,` for separating arguments.
- Have a look at our [styleguide](https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/)
- Use longer variable names where it makes sense. E.g. `degree` instead of `de`, or `degree_f` if you are also using the function `degree`. This makes it easier to read and to grep for stuff.
- I think you can discard stuff from a function outputting a tuple by using `_` if you are not interested in that variable.

Thanks!

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,214 @@
...
+Given a multivariate polynomial `f` and an integer degree `de`, performing the weierstrass preparation theorem of `f` up to degree `de`. Returning an unit `u` and a weierstrass polynomial `fhat`, s.t. `u``fhat`=`f`, while `fhat` is weierstrass. Additional the number of iterations used will be returned. 
+"""
+function weierstrass_preparation_theorem(f::MPolyElem,de::Int,n::Int)
+	
+	if f==0
+		throw(DomainError(f, "polynomial is zero"))
+	end
+	R=parent(f)
+	success,d=order_of_generality(f,n)
+	if success!=true
```
lkastner:
```suggestion
	if !success
```
`success` is a bool, right?

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,214 @@
...
+`weierstrass_division_theorem(g::MPolyElem,f::MPolyElem,de::Int)`
+
+Given a multivariate polynomial `f`, which is general of oder `d` in the last ring variable, a polynomial `g` and an integer degree `de`, performing the weierstrass division theorem of `f` up to degree `de`, by dividing `f` by `g`. Returning a polynomial `u` and a remainder `r` up to degree `de`, s.t. f=ug+r. Additional the number of iterations will be returned. 
+"""
+function weierstrass_division_theorem(g::MPolyElem,f::MPolyElem,de::Int,n::Int)
+	R=parent(f)
+	r=R(0)
+	vect=fill(0,nvars(R))
+	vect[n]=1 
+	success,d=order_of_generality(f,n)
```
lkastner: It seems like `success` is never used, if I am not mistaken. Then could you just do
```suggestion
	_,d = order_of_generality(f,n) 
```

experimental/Rings/WeierstrassPreparationTheorem.jl
Gebi95: Thanks for your comments. 

I tried to solve this by creating the function rand_coeff, but i am not sure, whether it is a good idea to have an operator as an argument for this function?

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
+##############################################################
+#
+#Conversion from Singular (weierstr.lib)
```
fingolfin:
```suggestion
# Conversion from Singular (weierstr.lib)
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
...
+#
+#############################################################
+
+export weierstrass_preparation_theorem
+
+export weierstrass_division_theorem
+
+
+"""
+`order_of_generality(f::MPolyElem, n::Int)`
```
fingolfin: Please use exactly four spaces to indent the signature(s) in docstrings, and use no extra backticks around it
```suggestion
    order_of_generality(f::MPolyElem, n::Int)
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
...
+
+compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d`
```
fingolfin: General rules of grammar apply for docstrings, too (so e.g. sentences are capitalized and end with a period, question mark or exclamation mark),
```suggestion
Compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` 
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
...
+export weierstrass_preparation_theorem
+
+export weierstrass_division_theorem
+
+
+"""
+`order_of_generality(f::MPolyElem, n::Int)`
+
+compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` 
+with `x_n` the n-th variable of the polynomial ring of `f`
```
fingolfin:
```suggestion
with `x_n` the `n`-th variable of the polynomial ring of `f`.
```

experimental/Rings/WeierstrassPreparationTheorem.jl
fingolfin: What does `` `x_n`-general `` mean?

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
...
+"""
+`order_of_generality(f::MPolyElem, n::Int)`
+
+compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` 
+with `x_n` the n-th variable of the polynomial ring of `f` 
+"""
+function order_of_generality(f::MPolyElem, n::Int)
+	 
+	vect = fill(parent(f)(0), nvars(parent(f))) 
+	vect[n] = gens(parent(f))[n]
```
fingolfin:
```suggestion
	vect[n] = gen(parent(f), n)
```
`gen(R,i)` is more efficient than `gens(R)[i]`; so this should be also done in other places (e.g. a few lines below in this function)

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
...
+"""
+`jet(f::MPolyElem, degree_f::Int)`
+
+Given a polynomial `f` and an integer degree `degree_f` , this function returns the polynomial,
+which contains only the monomials of `f`, which total degree is lower than or equal to `degree_f`
+"""
+function jet(f::MPolyElem, degree_f::Int) 
+	while f != 0
+		if total_degree(leading_monomial(f, ordering=deglex(gens(parent(f))))) > degree_f
+			f = f-coeff(f, leading_monomial(f, ordering=deglex(gens(parent(f)))))*leading_monomial(f, ordering=deglex(gens(parent(f))))
```
fingolfin: It seems you are computing `leading_monomial(f, ordering=deglex(gens(parent(f))` three times here? Better to store it in a variable... and also the ordering you keep re-creating...

```suggestion
    ordering = deglex(gens(parent(f)))
	while !iszero(f)
	    lm = leading_monomial(f, ordering)
		if total_degree(lm) > degree_f
			f = f - coeff(f,lm)*lm
```

That said, I am pretty sure this code is very inefficient in its current form. What you really want is to iterate over all *terms* and discard those whose total degree is too big. That does not require a specific monomial ordering; in particular the "intrinsic" ordering of the polynomial implementation could be used...

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,240 @@
...
+transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable.
+`nesting_depth` is the counter of recursive calls. First call with `nesting_depth` = 0
+"""
+function var_general(f::MPolyElem, n::Int, nesting_depth::Int) 
+	R = parent(f)
+	j = nvars(R)
+	_, d = order_of_generality(f, n) 
+	if d > 0 
+		return f	
+	else
```
fingolfin: The more indented a function, the harder it gets to understand it. Here, you can save one level: just do
```suggestion
	end
```
and then de-indent the following code, and finally drop the closing `end`

fingolfin (changes requested):
Thanks for your contribution!

I've left a couple comments. Please apply them in all places where it makes sense

experimental/Rings/WeierstrassPreparationTheorem.jl
Gebi95: My idea was, to use this ordering to not iterate over all terms of f. But maybe this is inefficient. So i changed it now to 
  i = length(f)
  while i >= 1
    if total_degree(term(f,i)) > degree_f
      f = f - term(f,i)
    end
    i = i - 1
  end

I hope, this matches your idea.

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+	vect[n] = gen(parent(f),n) 
+	f = evaluate(f, vect) 
+	f != 0 || return false, -1
+	return true, degree(leading_monomial(f, ordering=neglex(gens(parent(f)))), n)
+end
+
+"""
+    jet(f::MPolyElem, degree_f::Int)
+
+Given a polynomial `f` and an integer degree `degree_f` , this function returns the polynomial,
```
fingolfin, line 30:
```suggestion
Given a polynomial `f` and an integer degree `degree_f`, this function returns the polynomial
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+which contains only the monomials of `f`, which total degree is lower than or equal to `degree_f`.
```
fingolfin, line 31:
```suggestion
which contains only the monomials of `f` whose total degree is lower than or equal to `degree_f`.
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+"""
+function jet(f::MPolyElem, degree_f::Int) 
+  i = length(f)
+  while i >= 1
+    if total_degree(term(f,i)) > degree_f
+      f = f - term(f,i)
+    end
+    i = i - 1
+  end 
+	return f
```
fingolfin, line 41:
```suggestion
  return f
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+"""
+    jet(f::MPolyElem, degree_f::Int, vecw::Vector{Int})
+
+Given a polynomial `f`, an integer degree `degree_f` and an integer vector `vecw`, which has to have 
+the same length as length(parent(f)), this function returns a polynomial, which only contains 
+monomial, whose exponents taken as a vector multiplied (scalar product) with the given 
+weight-vector `intw` are lower than or equal to `degree_f`.
+"""
+function jet(f::MPolyElem, degree_f::Int, vecw::Vector{Int}) 
+
```
fingolfin, line 54:
```suggestion
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+end
+
+
+"""
+    jet(f::MPolyElem, degree_f::Int, vecw::Vector{Int})
+
+Given a polynomial `f`, an integer degree `degree_f` and an integer vector `vecw`, which has to have 
+the same length as length(parent(f)), this function returns a polynomial, which only contains 
+monomial, whose exponents taken as a vector multiplied (scalar product) with the given 
+weight-vector `intw` are lower than or equal to `degree_f`.
```
fingolfin, line 51:
```suggestion
weight-vector `vecw` are lower than or equal to `degree_f`.
```
(Although the `vec` is somewhat redundant, so perhaps just call it `w` or `weights`?)

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+"""
+function jet(f::MPolyElem, degree_f::Int, vecw::Vector{Int}) 
+
+	vect = collect(exponent_vectors(f))
+	i = length(f)
+	while i >= 1
+		if transpose(vecw)*vect[i] > degree_f
```
fingolfin, line 58:
```suggestion
		if dot(vecw,vect[i]) > degree_f
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+			f = f-term(f, i)
+		end
+		i = i-1	
+	end
+	return f
+end
+
+"""
+    invert_unit(u::MPolyElem, n::Int)
+If `u` is an unit, returning boolean value for success and the inverse of `u`.
```
fingolfin, line 68:
```suggestion

If `u` is an unit, returning boolean value for success and the inverse of `u`. 
```

And what is `n` doing?

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+	return f
+end
+
+"""
+    invert_unit(u::MPolyElem, n::Int)
+If `u` is an unit, returning boolean value for success and the inverse of `u`. 
+Else return false and the zero polynomial. 
+"""
+function invert_unit(u::MPolyElem, n::Int)
+	if u != 0 && total_degree(leading_monomial(u, ordering=negdegrevlex(gens(parent(u))))) == 0
```
fingolfin, line 72: Can't you also use this equivalently?
```suggestion
	if u != 0 && total_degree(leading_monomial(u, ordering=:negdegrevlex)) == 0
```

fingolfin: Is this code perhaps assuming that the coefficient ring is a field?

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+    var_general(f::MPolyElem, n::Int, nesting_depth::Int)
+
+Transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable.
+`nesting_depth` is the counter of recursive calls. First call with `nesting_depth` = 0.
+"""
+function var_general(f::MPolyElem, n::Int, nesting_depth::Int) 
+	R = parent(f)
+	j = nvars(R)
+	_, d = order_of_generality(f, n) 
+	if d > 0
```
fingolfin, line 109: Please replace all tabs by spaces

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+Transform `f` into a polynomial, which is x_n general of finite order for x_n the `n`-th ring variable.
+`nesting_depth` is the counter of recursive calls. First call with `nesting_depth` = 0.
+"""
+function var_general(f::MPolyElem, n::Int, nesting_depth::Int) 
+	R = parent(f)
+	j = nvars(R)
+	_, d = order_of_generality(f, n) 
+	if d > 0 
+		return f	
+  end
```
fingolfin, line 111: Indention is wrong here (probably a tabs versus spaces issue?)

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+			if length(g) > length(evaluate(g, m_hilf))
+				m[i] = gen(R,i)+rand(1-(nesting_depth)*10 : 1+(nesting_depth)*10)*gen(R,n)
+				g = evaluate(f, m)
+				break
+			end
+		end
+		i = i+1
+		end
+		
+		if nesting_depth <= 3
```
fingolfin, line 128: Indentation is definitely way off

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+function initial_form(f::MPolyElem)
+	degree_f = total_degree(leading_monomial(f, ordering=negdegrevlex(gens(parent(f))))) 
+	g = jet(f, degree_f) 
+	return g
+end
+
+"""
+    rand_coeff(m::Vector{fmpq_mpoly}, n::Int, mini::Int, maxi::Int, operator)
+
+Given a vector of monomials `m`, an Integer `n`, integer bounds mini und maxi and an
```
fingolfin, line 155:
```suggestion
Given a vector of monomials `m`, integer `n`, `mini` and `maxi`, and an 
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+which is general for the `n`-th ring variable.
+"""
+function rand_coeff(m::Vector{fmpq_mpoly}, n::Int, mini::Int, maxi::Int, operator)
+	R=parent(m[1])
+	j=nvars(R)
+	i = 1
+	while i <= j 
+		if i != n
+			m[i] = gen(R,i)+operator(gen(R,n),rand(mini : maxi))
+		end
```
fingolfin, line 166: maybe pass the range `mini:maxi`?

experimental/Rings/WeierstrassPreparationTheorem.jl
lkastner: I think that is ok, we have operators as arguments elsewhere too. Thanks.

fingolfin (changes requested):

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+Compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` 
+with `x_n` the `n`-th variable of the polynomial ring of `f`.
+"""
+function order_of_generality(f::MPolyElem, n::Int)
+	 
+	vect = fill(parent(f)(0), nvars(parent(f))) 
+	vect[n] = gen(parent(f),n) 
+	f = evaluate(f, vect) 
+	f != 0 || return false, -1
+	return true, degree(leading_monomial(f, ordering=neglex(gens(parent(f)))), n)
```
fingolfin, line 24:
```suggestion
	R = parent(f)
	vect = fill(R(0), nvars(R)) 
	vect[n] = gen(R,n) 
	f = evaluate(f, vect) 
	f != 0 || return false, -1
	return true, degree(leading_monomial(f, ordering=neglex(gens(R))), n)
```

experimental/Rings/WeierstrassPreparationTheorem.jl
```diff
@@ -0,0 +1,239 @@
...
+export weierstrass_preparation_theorem
+
+export weierstrass_division_theorem
+
+
+"""
+    order_of_generality(f::MPolyElem, n::Int)
+
+Compute for a given polynomial `f` an integer `d` such that `f` is `x_n`-general of order `d` 
+with `x_n` the `n`-th variable of the polynomial ring of `f`.
```
fingolfin, line 16:
```suggestion
Return an integer `d` such that `f` is ``x_n``-general of order `d` 
with ``x_n`` the `n`-th variable of the polynomial ring of `f`.
```

experimental/Rings/WeierstrassPreparationTheorem.jl
afkafkafk13: Textbook definition:
A power series f \in CC{x_1,..,x_n} is called x_n-general of order d if f(0,..,0,x_n) has a zero of order d at x_n=0.

experimental/Rings/WeierstrassPreparationTheorem.jl
afkafkafk13: In the context of the Weierstrass Preparation theorem, I would assume that the coefficient ring is technically QQ and theoretically CC.

fingolfin:
What's the plan here? There are a bunch of comments that are not yet addressed.

afkafkafk13:
I did a few tests with this implementation and it is far from optimal; I doubt that it is worth adressing the comments, before its major flaws have been addressed.  We will discuss with the author whether this will be closed or refactored.

afkafkafk13:
Refactoring will not take place in this PR.

PR #2241, opened 2023-04-12
Support more rings for `is_graded` and friends

fingolfin:
Allow invoking `is_graded` and friends on any multivariate polynomial
ring, even undecorated ones -- in the undecorated case just always
return `false`.

This allows to simplify code that performs argument checking and also
seems fine in terms of correctness.


simonbrandhorst (commented):
Thanks for the cleanup. Looks good to me. Could you please add a test that everything works as expected?

simonbrandhorst (approved):

simonbrandhorst:
Well, if it works it works.

PR #2243, opened 2023-04-12
Remove Markdown as a dependency

fingolfin:
As long as these are in place, new PRs may bring in new uses of `Markdown.@doc`. With this PR here merged, those PRs will be flagged by CI


lkastner (approved):
At this point I really wonder why we needed it...

ThomasBreuer (approved):

PR #2245, opened 2023-04-13
ToricVarieties: Add dictionaries linking homogeneous variables to rays

HereAround:
In preparation of a proper version of https://github.com/oscar-system/Oscar.jl/pull/2228.


lkastner:
I do not understand why this is necessary. Both the rays and the variables are indexed. Does this info not suffice?

HereAround:
> I do not understand why this is necessary. Both the rays and the variables are indexed. Does this info not suffice?

As discussed on slack - these are convenience methods (for me). I have adjusted this PR as we discussed on slack. Thank you for the suggestions and improvements @lkastner .

lkastner (approved):

PR #2254, opened 2023-04-14
support `Oscar.iso_gap_oscar` for Alnuth's `FieldByMatrices`

ThomasBreuer:
The idea is to get a better interface between Oscar and those functions from Alnuth which used to delegate tasks to Pari/GP, see #2220 for the current situation.

Note that the proposed `Oscar.iso_gap_oscar` and the already available `Oscar.iso_gap_oscar` code that deals with algebraic extensions on the GAP side cannot be used to simply replace the `_OscarField` that has been introduced in #2220, because of the difference between `DefiningPolynomial` and `IntegerDefiningPolynomial` (see my comment in the discussion of #2220).


src/GAP/iso_gap_oscar.jl
```diff
@@ -180,6 +183,46 @@ function _iso_gap_oscar_number_field(FG::GapObj)
...
      end
+   elseif GAPWrap.IsNumberFieldByMatrices(FG)
+     # `pol` is the minimal polynomial of `M` because the GAP code
+     # caches the objects.
+     M = GAPWrap.PrimitiveElement(FG)
+     pol = GAPWrap.DefiningPolynomial(FG)
+
+     # Construct the isomorphic number field in Oscar.
+     cfs = GAPWrap.CoefficientsOfUnivariatePolynomial(pol)
+     R, x = polynomial_ring(QQ, "x")
```
thofma:
```suggestion
     R, x = polynomial_ring(QQ, "x", cached = false)
```

src/GAP/iso_gap_oscar.jl
```diff
@@ -180,6 +183,46 @@ function _iso_gap_oscar_number_field(FG::GapObj)
...
+     polFO = R(Vector{QQFieldElem}(cfs))
+     FO, _ = number_field(polFO, "z")
```
thofma:
```suggestion
     FO, _ = number_field(polFO, "z", cached = false)
```

src/GAP/iso_gap_oscar.jl
ThomasBreuer: Why shall we force the creation of a new polynomial ring whenever a new number field in Oscar gets created?

thofma: I am happy if you want to keep it, but why shall we force a useless dictionary lookup whenever a new number field is created?

fingolfin: Do we have a written down policy when to enable the caching and when not? @fieker?

fingolfin (approved):
Looks good to me. Unless @thofma has objections, maybe we can merge it later today?

thofma (approved):

PR #2255, opened 2023-04-14
Update TagBot integration

fingolfin:


thofma:
Will this fix the missing release that we have now? (after retriggering)

fingolfin:
Dunno, but this is what the TagBot repo recommends now and it was added in order to fix some issues with generating tags... Plus it makes it possible to use the `workflow_dispatch` with a configurable "lookback" so that we can re-run it for older commits, such as the 0.12.0 release commit...

Seee https://github.com/JuliaRegistries/TagBot/pull/236 which fixed https://github.com/JuliaRegistries/TagBot/issues/222 and https://github.com/JuliaRegistries/TagBot/issues/229

fingolfin:
So perhaps we just merge this now (running the CI is pointless) and then try and see we'll see if it helps by triggering tagbot manually?

thofma:
https://github.com/oscar-system/Oscar.jl/actions/runs/4700190747

thofma:
Does not seem to work at all (https://github.com/oscar-system/Oscar.jl/actions/runs/4700190747/jobs/8334742285). The error is also different to the reported issues I think. Maybe it does not like retriggering.

fingolfin:
The error there is actually helpful:
```
github.GithubException.RateLimitExceededException: 403 {"message": "API rate limit exceeded for installation ID 6710914.", "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```
That doesn't sound like retrying will help. Maybe we just have to wait a bit?

PR #2257, opened 2023-04-14
Experimental template

HechtiDerLachs:
As discussed in today's meeting. 

@lkastner @JohnAAbbott

### DIFF
--- a/experimental/ExperimentalTemplate/docs/doc.main
+++ b/experimental/ExperimentalTemplate/docs/doc.main
@@ -1,0 +1,8 @@
+# An example documentation
+
+This is a sample text to outline the structure of the packages in the `Experimental` folder.
+You can show docstrings like this:
+```@docs
+    content(S::ExampleStruct)
+```
+    
--- a/experimental/ExperimentalTemplate/docs/doc.main
+++ b/experimental/ExperimentalTemplate/docs/doc.main
@@ -3,6 +3,6 @@
 This is a sample text to outline the structure of the packages in the `Experimental` folder.
 You can show docstrings like this:
 ```@docs
-    content(S::ExampleStruct)
+    my_access_func(S::ExampleStruct)
 ```
     
--- a/experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
+++ b/experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
@@ -1,0 +1,15 @@
+# Add your new types, functions, and methods here.
+
+mutable struct ExampleStruct
+  i::Int
+end
+
+@doc raw"""
+    content(S::ExampleStruct)
+
+This is a dummy sample function to teach the use of docstrings.
+"""
+function content(S::ExampleStruct)
+  return S.i
+end
+
--- a/experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
+++ b/experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
@@ -5,11 +5,11 @@ mutable struct ExampleStruct
 end
 
 @doc raw"""
-    content(S::ExampleStruct)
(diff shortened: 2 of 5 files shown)


experimental/ExperimentalTemplate/test/runtests.jl
```diff
@@ -0,0 +1,4 @@
+@testset "Experimental template tests" begin
+  S = oscar.ExampleStruct(5)
+  @test 5 == oscar.content(S)
```
fingolfin:
```suggestion
  S = Oscar.ExampleStruct(5)
  @test 5 == Oscar.content(S)
```

experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
```diff
@@ -0,0 +1,15 @@
...
+mutable struct ExampleStruct
+  i::Int
+end
+
+@doc raw"""
+    content(S::ExampleStruct)
+
+This is a dummy sample function to teach the use of docstrings.
+"""
+function content(S::ExampleStruct)
```
fingolfin: There is a genuine `Oscar.content` function and now this docstring will show up in its docstring. Not good.

I would suggest (a) a different name for this dummy function, and (b) perhaps put this entire file into a `module` to avoid further clashes

fingolfin (commented):
Maybe also add a `README.md`?

experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
lkastner: Could we maybe just ignore the template when loading Oscar by putting it into the `oldexppkgs` in `src/Oscar.jl`? I don't really see any benefit for maintenance when we load this. Only very minor one.

fingolfin: That seems sensible to me.

I'd still recommend against defining a `content` function. We can call it `dummy`?

experimental/ExperimentalTemplate/docs/doc.main
```diff
@@ -0,0 +1,8 @@
+# An example documentation
+
+This is a sample text to outline the structure of the packages in the `Experimental` folder.
+You can show docstrings like this:
+```@docs
+    content(S::ExampleStruct)
```
fingolfin:
```suggestion
example_function(S::ExampleStruct)
```

experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
```diff
@@ -0,0 +1,15 @@
+# Add your new types, functions, and methods here.
+
+mutable struct ExampleStruct
+  i::Int
+end
+
+@doc raw"""
+    content(S::ExampleStruct)
```
fingolfin:
```suggestion
    example_function(S::ExampleStruct)
```

experimental/ExperimentalTemplate/src/ExperimentalTemplate.jl
fingolfin:
```suggestion
function example_function(S::ExampleStruct)
```

lkastner: I thought a bit and I think it makes sense to have both: A template that people with existing code can use for adjusting their stuff, as well as a function that copies the template and renames appropriately if someone wants to start from scratch. I'll do the second part once this PR is merged.

experimental/ExperimentalTemplate/docs/doc.main
fingolfin:
```suggestion
    my_access_func(S::ExampleStruct)
```

fingolfin (approved):

PR #2265, opened 2023-04-16
Enhance `Oscar.test_module`

### LINKED ISSUE #2204: `Oscar.test_module` error message and namespaces
This has been originally raised by @fingolfin in https://github.com/oscar-system/Oscar.jl/pull/2188#discussion_r1156911058.

The error message https://github.com/oscar-system/Oscar.jl/blob/915aeb9de5c3ffe97ffcbeca20b6fbca8e61d22c/src/Oscar.jl#L432 can be widely misleading.

Instead of guessing that perhaps using Test is missing, we can do much better: We could first check whether it is available, even before the include, and refuse to even do the include if it isn't.

My second concern is: we are includeing the test file into the Oscar namespace. But I don't think that's how tests are run when one uses Pkg.test and thus also not how CI runs it; and also not how this function does it when new is specified. Rather, the include is always done from the REPL.

So perhaps instead of include(test_file) we should be doing Base.include(Base.Main, test_file) ?

lgoettgens:
Resolves #2204.

cc @fingolfin

### DIFF
--- a/src/Oscar.jl
+++ b/src/Oscar.jl
@@ -425,16 +425,9 @@ function test_module(file::AbstractString; new::Bool=true)
     @info("spawning ", `$julia_exe -e \"$cmd\"`)
     run(`$julia_exe -e $cmd`)
   else
+    @req isdefined(Base.Main, :Test) "You need to do \"using Test\""
     @info("Running tests for $rel_test_file in same session")
-    try
-      include(test_file)
-    catch e
-      if isa(e, LoadError)
-        println("You need to do \"using Test\"")
-      else
-        rethrow(e)
-      end
-    end
+    Base.include(Base.Main, test_file)
   end
 end
 


lkastner (approved):

lkastner:
The docs error looks like it has nothing to do with your changes but I could be mistaken.

lgoettgens:
The docs error should be due to me working in a fork. See https://github.com/oscar-system/Oscar.jl/actions/runs/4713120756/jobs/8360576339?pr=2265#step:6:425
It occurs in all of my PRs.

benlorenz:
An error should not happen, even from a fork. Documenter usually skips the deploy step when the PR is from a fork. But for some reason the autodetection does not work:
```
┌ Warning: Unable to verify if PR comes from destination repository -- assuming it does.
└ @ Documenter ~/.julia/packages/Documenter/H5y27/src/deployconfig.jl:529
┌ Info: Deployment criteria for deploying preview build from GitHub Actions:
│ - ✔ ENV["GITHUB_REPOSITORY"]="oscar-system/Oscar.jl" occurs in repo="github.com/oscar-system/Oscar.jl.git"
│ - ✔ ENV["GITHUB_REF"] corresponds to a PR number
│ - ✔ PR originates from the same repository
│ - ✔ `push_preview` keyword argument to deploydocs is `true`
│ - ✔ ENV["GITHUB_ACTOR"] exists and is non-empty
│ - ✔ ENV["GITHUB_TOKEN"] exists and is non-empty
└ Deploying: ✔
```
(Maybe this is related to the tagbot problem?)

fingolfin (approved):

fingolfin:
Thank you!

PR #2268, opened 2023-04-17
LieAlgebras: Fix combinatorics tests

lgoettgens:
Adds tests for edge cases of wrappers around Hecke functions, that have been fixed in thofma/Hecke.jl#1040.
Furthermore, does no longer do `using submodule` in testing and thus messing up the namespaces.


experimental/LieAlgebras/test/Combinatorics-test.jl
```diff
@@ -1,5 +1,7 @@
 @testset "LieAlgebras.Combinatorics" begin
   @testset "combinations" begin
+    combinations = Oscar.LieAlgebras.combinations
```
fingolfin, line 3: Shouldn't this be 
```suggestion
    import Oscar.LieAlgebras: combinations
```
Possibly at the top of this file? Same for the other similar ones

lgoettgens: In my understanding, `import` would pull it into the current namespace. What I was trying to achieve here, is only to be able to use the short form in this one test set without polluting the global namespace.
I see the possibility for the success of the test ci being dependent on the order of the test files run, if every file pulls something more into the global namespace, and possibly using something non-exported that has only been imported by a previous test.

If this, in your opinion, is no problem, or if there is a cleaner way to do it, I am happy to change it.

thofma: Yes, this is the right way to do it. Alternatively use a `let` around everything.

fingolfin: or put the tests into a module of their own?

thofma: I learned to use `let` from the julia tests (if it is not already inside a `@testset` block). There it is used all over the place to avoid polluting the namespace. Not sure if it makes the use of `let` better.

thofma: P.S.: I thought `@testset` introduces automatically a local scope block.

lgoettgens: I am a bit confused. What would be the/your preferred way to change this into?

fingolfin: Either use `import` or `let` or put everything into a module -- but don't do `combinations = Oscar.LieAlgebras.combinations`.

@thofma suggests using `let`, which seems to fit with the goal you want to achieve; I'd be happy with that as well

thofma: I think `@testset` introduces a local scope, so the `combinations = ...` should be fine? @fingolfin can you elaborate what the problem is?

lgoettgens: Yes, indeed there is a local scope. The following fails with an UndefVarError in line 6.
```julia
@testset "Some tests" begin
  x = 3
  @test x == 3
end

@test x == 3
```

fingolfin (approved):

PR #2270, opened 2023-04-17
Better printing for weyl_algebra & exterior_algebra

JohnAAbbott:
More compact printing for weyl_algebra and exterior_algebra (see discussion 2238).
In both cases output is still via `expressify` -- **no separate handling** for `detailed`, `one_line` and `supercompact`!
Implementation is "clunky": weyl_algebra handled via attributes (`:is_weyl_algebra`), exterior_algebra is handled via the pre-existing function `have_special_impl`  (an attempt to use attributes did not succeed: trouble with constructor signatures)


test/Experimental/ExteriorAlgebra-test.jl
```diff
@@ -4,28 +4,31 @@
 
 ######## CONSTRUCTOR TESTs
 @test_throws  ArgumentError  exterior_algebra(QQ, 0);
-@test_broken  exterior_algebra(QQ, 1);  # --> special case (is commutative)
```
fingolfin, line 7: I don't know what the plan with this PR is (whether it will be merged soon or still needs work or whatever). But if the merge is not imminent, could you please put this fix into a separate PR, so that we get rid of the additional CI failure?

Thanks!

fingolfin: Submitted via PR #2294

fingolfin (approved):
Fine by me, thanks!

PR #2272, opened 2023-04-17
Refactoring styleguide

### LINKED ISSUE #2117: Developer Docs: rule (or rather, "recommendation") for file names
It is a bit irritating that we now have `src/Exports.jl` and `src/imports.jl` (not the different case in the filename). This should be unified one way or another. Indeed, we have:

- `src/Aliases.jl`
- `src/Deprecations.jl`
- `src/Exports.jl`
- `src/fallbacks.jl`
- `src/imports.jl`
- `src/printing.jl`

So it is 50-50 (I am deliberately omitting `src/Oscar.jl` because its name is forced upon us by Julia, there is no choice there, unless we also rename the package, which... nope). 

There are many more examples in the subdirectories. There we also start to see the use of `-` and `_` as separator. This results in e.g. these:
- `src/Rings/MPolyQuo.jl`
- `src/Rings/mpolyquo-localizations.jl`
- `src/Rings/mpoly_types.jl`

Now, I am a bit reluctant about adding more and more strict rules, as they can start to choke us and hold up things; file name changes are easy to do from a technical point of view, but at the same time can be very disruptive for ongoing work on code. Yet I do think it would be good to at least make a recommendation for people how they should name *new* files. And then we can very carefully and selectively change existing files over time, and in coordination with people working on them (so in particular, I don't want a mass renaming PR).

So what could the recommendation look like? I see three patterns above:
1. `snake_case.jl`
2. `CamelCase.jl`
3. `kebab-case.jl`

I have ordered them in order of my personal preference, but overall I can live with any of them (and I don't mind if `src/Oscar.jl` sticks out, as I said, it's special anyway). I just would prefer if we settled on one default. (Someone might suggest that different rules for the file names could be used depending on their content, as is seemingly the case for `Nemo`. I am very reluctant to do that, as I don't see a benefit, but it increases the mental burden again: now I have to figure out if my file with that rule or not).

Some observations on what other projects use:
- Julia seems to mostly use `snake_case`
- AbstractAlgebra uses `CamelCase`
- Nemo uses a mixture (perhaps: files with FLINT interface code uses `snake_case`, pure Julia uses `CamelCase`??)
- Hecke uses a mixture
- the only place I see `-` in file names in our other packages is inside `test/` and only in the form `BLAH-test.jl` (I don't understand what the point of this suffix is; but I don't mind it; but perhaps here, too, we could decided to either always require it, or never, or what?)

lkastner:
I am refactoring the styleguide, since it got large.
- Move some things about printing to a separate file.
- Introduce a general section
- Mention recommendation as per #2117 
- Introduce approved abbreviations closing #1981.


fingolfin (approved):
Fine by me

PR #2273, opened 2023-04-17
Fix Alnuth override when Oscar is not in Main

fingolfin:
If OSCAR is not loaded by the user in the REPL, but indirectly, then nf_elem won't be available in the Main module, and hence evaluating 'Vector{nf_elem}' in there will fail.

So instead eval this inside the Oscar module.

### DIFF
--- a/gap/OscarInterface/gap/alnuth.gi
+++ b/gap/OscarInterface/gap/alnuth.gi
@@ -127,14 +127,14 @@ BindGlobal("NormCosetsDescriptionOscar", function(F, norm)
   return rec(units := units, creps := creps);
 end);
 
-
+BindGlobal("_Vector_nf_elem", Oscar.eval(Julia.Meta.parse(GAPToJulia("Vector{nf_elem}"))));
 
 BindGlobal("PolynomialFactorsDescriptionOscar", function(F, coeffs)
   local K, cf, poly, facs, result, f, g, i;
 
   K := _OscarField(F);
 
-  cf := JuliaEvalString( "Vector{nf_elem}")(Reversed(List(coeffs, x -> K(GAPToJulia(x)))));
+  cf := _Vector_nf_elem(Reversed(List(coeffs, x -> K(GAPToJulia(x)))));
   poly := Oscar.polynomial(K, cf);
   facs := Oscar.factor(poly);
   Assert(0, Oscar.is_one(facs.unit));


ThomasBreuer (approved):
It is funny that we did not notice this problem with `JuliaEvalString` earlier.
I am wondering what should be the recommended solution in general.

- Computing the Julia type in question only once and caching it in a GAP variable is reasonable, independent of the way how it gets computed.
- An alternative to the proposed code would be to call `JuliaEvalString("Vector{Oscar.nf_elem}")`.
- Another alternative would be to provide `JuliaEvalString(module, string)` which then calls `module.eval`.
- In any case, the documentation of `JuliaEvalString` should be extended such that it mentions that one may need to specify Julia modules in the string.
- Should we perhaps replace the C-code of `JuliaEvalString`, which is based on `jl_eval_string`, by Julia code of the form which is proposed in this pull request?

fingolfin:
We already talked about this on gather this morning, but to record this for the future, here are my replies in written form:

> It is funny that we did not notice this problem with `JuliaEvalString` earlier. I am wondering what should be the recommended solution in general.
> 
> * Computing the Julia type in question only once and caching it in a GAP variable is reasonable, independent of the way how it gets computed.
> * An alternative to the proposed code would be to call `JuliaEvalString("Vector{Oscar.nf_elem}")`.

Unfortunately this doesn't work if `Oscar` is not defined in `Main`; e.g. if the user did `using FooBar`, where `FooBar` is a package that depends on `Oscar`, but does not expert the symbol `Oscar`; then to access the `Oscar` module from the REPL, one would need to enter `FooBar.Oscar`.


> * Another alternative would be to provide `JuliaEvalString(module, string)` which then calls `module.eval`.

I think that would be useful, agreed!

> * In any case, the documentation of `JuliaEvalString` should be extended such that it mentions that one may need to specify Julia modules in the string.

Indeed. It could explain that evaluation takes place (by default) in the `Main` module, i.e. essentially the context of the REPL, and hence care must be taken that everything the evaluated code is supposed to access is available there.

> * Should we perhaps replace the C-code of `JuliaEvalString`, which is based on `jl_eval_string`, by Julia code of the form which is proposed in this pull request?

Fine by me in principle, but then we should add some more tests that deliberately involve errors to verify behavior does not change (resp. only changes in a way we are OK with). That includes:
- basic syntax errors, e.g. evaluating `1++1`
- runtime errors, e.g. evaluating `sin(nothing)`

ThomasBreuer:
@fingolfin Thanks for the comments.

> Unfortunately this doesn't work if Oscar is not defined in Main; e.g. if the user did using FooBar, where FooBar is a package that depends on Oscar, but does not expert the symbol Oscar; then to access the Oscar module from the REPL, one would need to enter FooBar.Oscar.

Now I understand what happened in my tests.
I had called `JuliaImportPackage( "Oscar" )` on the GAP side, which is natural if one wants to use some Oscar variables in GAP. After this call, Oscar *is* defined in Main.

Due to the `GAP.Globals.BindGlobal(GapObj("Oscar"), Oscar)` call in Oscar's `__init__`, one does not need `JuliaImportPackage( "Oscar" )` anymore in order to access Oscar variables, and in fact this variant avoids overhead on the GAP side.

PR #2275, opened 2023-04-17
Add dummy docs for LieAlgebras experiment

### LINKED ISSUE #2274: Documentation updates are not deployed anymore
It seems the "pages build and deployment" GH Action stopped working ~2 weeks ago, (see https://github.com/oscar-system/Oscar.jl/actions/workflows/pages/pages-build-deployment?page=4).

Consequently https://docs.oscar-system.org/dev/ is outdated.

The error message looks like this:

> github-pages 228 | Error:  No such file or directory @ rb_check_realpath_internal - /github/workspace/dev/Experimental/LieAlgebras

So it doesn't like something about the LieAlgebras stuff (CC @lgoettgens) -- perhaps that it doesn't contain any docs?

fingolfin:
Maybe fixes #2274 ? 

CC @lgoettgens

### DIFF
--- a/experimental/LieAlgebras/docs/src/introduction.md
+++ b/experimental/LieAlgebras/docs/src/introduction.md
@@ -1,0 +1,7 @@
+```@meta
+CurrentModule = Oscar
+```
+
+# Introduction
+
+TODO


benlorenz:
I don't think this change is necessary, this issue might have been fixed by https://github.com/oscar-system/Oscar.jl/pull/2215.
The recent runs of that job have failed due to `previews/PR2202/Experimental/LieAlgebras` which is indeed a broken symlink in the gh-pages branch. Another one exists in the PR2102 folder, but there are no more such links from more recent runs.

I just deleted these to symlinks from the gh-pages branch, let's see what happens.

fingolfin:
seems to have worked!

PR #2277, opened 2023-04-18
Revert "Update TagBot.yml"

lgoettgens:
This reverts commit 73bfbd84771e8a32677c40cf5ea45d185ff31acc.

Possibly fixes parts of #2274.

According to the [TagBot docs](https://documenter.juliadocs.org/stable/man/hosting/#GitHub-Actions) (in the yellow box `TagBot & tagged versions`) and the [TagBot Readme.md](https://github.com/JuliaRegistries/TagBot#setup), the ssh key should be contained. The rate-limit problem of tagbot hasn't been fixed by 73bfbd84771e8a32677c40cf5ea45d185ff31acc, so I see no point in keeping it longer.

Edit: Added `workflow_dispatch` as a trigger for easier re-runs in the future.

cc @benlorenz @fingolfin

PR #2278, opened 2023-04-18
Add `v0.12.0` docs

lgoettgens:
Contains the second workaround explained in https://github.com/oscar-system/Oscar.jl/issues/2274#issuecomment-1512552053 for adding the `v0.12.0` docs.
Since 73bfbd84771e8a32677c40cf5ea45d185ff31acc removed the ssh key from tagbot, its commits are not allowed to create CI runs themselves. Thus, for the `v0.12.0` tag, there has been no documentation run. For future releases, this is proposed to be reverted to work again in #2277.

cc @fingolfin @benlorenz


v0.12.0/siteinfo.js
```diff
@@ -0,0 +1 @@
+var DOCUMENTER_CURRENT_VERSION = "dev";
```
benlorenz:
```suggestion
var DOCUMENTER_CURRENT_VERSION = "v0.12.0";
```
I think this should be the released version instead of `dev`.

lgoettgens: Yes, indeed. This was somehow not staged.

fingolfin (approved):
Thank you

Issue #1914, opened 2023-02-09
Oscar silently changes orderings of ideal gens


YueRen:
Edit: @thofma: There is something strange going on for the ideal generator things, since Oscar silently changes the orderings of `IdealGen` objects:

```julia
julia> R, (x, y) = polynomial_ring(QQ, ["x", "y"])
(Multivariate Polynomial Ring in x, y over Rational Field, fmpq_mpoly[x, y])

julia> I = ideal(R, x*y);

julia> G = groebner_basis(I);

julia> G.ord
degrevlex([x, y])

julia> II = ideal(G);

julia> groebner_basis(II, ordering = degrevlex([y, x]));

julia> G.ord
degrevlex([y, x])
```

<details><summary>Original example from @YueRen </summary>
<p>

In the example below, the final `groebner_basis(I,ordering=degrevlex(R))` returns the correct Groebner basis with respect to `degrevlex(R)` (see also `G`), but it claims that the Groebner basis is with respect to `adjacentOrdering` which it is not (see also `Gprime`)

```
using Oscar
R,(p01,p02,p03,p04,p12,p13,p14,p23,p24,p34) = PolynomialRing(QQ,["p01","p02","p03","p04","p12","p13","p14","p23","p24","p34"])

I = ideal(R,[p23*p14-p13*p24+p12*p34,
             p23*p04-p03*p24+p02*p34,
             p13*p04-p03*p14+p01*p34,
             p12*p04-p02*p14+p01*p24,
             p12*p03-p02*p13+p01*p23])

G = groebner_basis(I)

u = Int[-1, -1, 0, 0, 0, 0, -1, 0, -2, -1]
v = Int[0, 0, 0, 0, 1, 0, -1, -1, 0, 1]
n = ngens(R)
adjacentOrdering = weight_ordering(ones(Int,n),
                                   weight_ordering(u,
                                                   weight_ordering(v,lex(R))))
Gprime = groebner_basis(ideal(G),ordering=adjacentOrdering,complete_reduction=true)

groebner_basis(I,ordering=degrevlex(R))
```

**System:**
```
ulia> Oscar.versioninfo(full=true)
OSCAR version 0.11.3-DEV - #yr/groebnerFan, e7373cf7e -- 2023-02-07 15:50:54 +0000
  combining:
    AbstractAlgebra.jl   v0.27.8
    GAP.jl               v0.9.2
    Hecke.jl             v0.17.2
    Nemo.jl              v0.32.7
[...]

PR #2109, opened 2023-03-22
Fix for silent changes of monomial ordering

jankoboehm:
This aims to fix the use of a wrong function signature of singular_assure for IdealGens and a monomial ordering which results in a change of the input object (silently changing the monomial ordering) when doing a computation. This problem was first observed in the standard_basis command and this hence also aims to fix #1914, but the issue is occures at various places (normal_form and friends, reduce and friends, is_cohen_macaulay, hilbert driven buchberger, invariant theory). The wrong signature of singular_assure is still in, but is eliminated at those places found. It should be deleted eventually. This changes code of a couple of people, so I hope it does not create new issues.

PR #2280, opened 2023-04-18
Revise some `AffAlgHom` functionality (and related things)

joschmitt:
I wanted to clean up something in the invariant theory code and somehow this escalated...

- Revise functionality for `AffAlgHom` to reduce duplication of code etc. 
- Avoid passing around a monomial ordering via a symbol in the internal function `_groebner_data(::AffAlgHom)`. This function does in fact not do any Gröbner basis computations any more, but only builds and caches the ideal which then caches the bases on itself.
- Remove (outdated?) uses of `<` for the comparison of monomials which does not know or ask about the used monomial ordering.
- Please check the small change in `src/Rings/mpoly.jl`; as far as I can see without this the ideal on the Singular side gets rebuild every time.
- There are two test files with essentially the same content; I removed one.


thofma:
Documentation build is failing

joschmitt:
> Documentation build is failing

~~I don't understand it. I copied the function signature from the doc string and it still doesn't find it.~~

Sorry, my bad.

src/Rings/mpoly-affine-algebras.jl
```diff
@@ -887,32 +846,18 @@ julia> V = [x[1]^3*x[2]^3-x[1]^3*x[3]^3, x[1]^3*x[2]^3+x[1]^3*x[3]^3]
...
-      psi = hom(T, TT, vcat(A, B))
-      return (true, psi(D[1]))
-   else
-      return (false, TT(0))
-   end
+function subalgebra_membership(f::T, v::Vector{T}) where T <: Union{MPolyRingElem, MPolyQuoRingElem}
+  R = parent(f)
+  @req coefficient_ring(R) isa Field "The coefficient ring must be a field"
+  @assert !isempty(v)
+  @assert all(x -> parent(x) === R, v)
```
thofma: Since we changed some of the `@assert` to `@req`, we might as well do it here.

src/Rings/mpoly.jl
```diff
@@ -661,15 +661,15 @@ function singular_assure(I::IdealGens, ordering::MonomialOrdering)
   else
       #= singular ideal exists, but the singular ring has the wrong ordering
        = attached, thus we have to create a new singular ring and map the ideal. =#
-      if !isdefined(I, :ord) || I.ord != ordering.o
+      if !isdefined(I, :ord) || I.ord != ordering
```
thofma, line 664: Hm? Was this never called? Or was this never true?

thofma:
Looks good to me. The only user-visible change is the introduction of the `induce` for orderings?

src/Rings/mpoly.jl
ederc: You can intermediately fix it in this way, still we need to fix #1914 (hopefully via #2109).

ederc (approved):

src/Rings/mpoly.jl
joschmitt: As far as I understand it, the condition `I.ord != ordering.o` would always evaluate to `true` because it doesn't compare the right things. So it builds a new ring and a new ideal every time.

joschmitt:
> Looks good to me. The only user-visible change is the introduction of the `induce` for orderings?

Yes, I would say so.

src/Rings/mpoly-affine-algebras.jl
joschmitt: True, I pushed a commit.

thofma:
@wdecker do you also want to have a look or is this fine for your?

wdecker:
Fine for me as @ederc checked it.

Von meinem iPhone gesendet

PR #2281, opened 2023-04-18
ToricVarieties: Fix bug in toric blow_up

HereAround:
cc @apturner

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
@@ -582,8 +582,8 @@ function blow_up(v::AbstractNormalToricVariety, I::MPolyIdeal; coordinate_name::
     indices = [findfirst(y -> y == x, gens(cox_ring(v))) for x in gens(I)]
     @req length(indices) == ngens(I) "All generators must be indeterminates of the cox ring of the toric variety"
     cone_list = cones(v)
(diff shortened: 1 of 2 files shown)


src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -583,7 +583,7 @@ function blow_up(v::AbstractNormalToricVariety, I::MPolyIdeal; coordinate_name::
     @req length(indices) == ngens(I) "All generators must be indeterminates of the cox ring of the toric variety"
     cone_list = cones(v)
     cone_indices = [filter(l -> cone_list[k,l], 1:ncols(cone_list)) for k in 1:n_cones(v)]
-    cone_index = findfirst(x -> x == indices, cone_indices)
+    cone_index = findfirst(x -> Set(x) == Set(indices), cone_indices)
```
lkastner: Does the following maybe work?
```
indexset = Set(indices)
cone_index = findfirst(i -> Polymake.row(cone_list, i) == indexset, 1:nrows(cone_list))
```
The `cone_indices` could then be left out entirely. Maybe also `Set{Int}` is better than just `Set`.

lkastner (approved):

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
HereAround: Thank you. Yes - this indeed works. If have also changed `Set` to `Set{Int}`.

PR #2287, opened 2023-04-18
Enhance star_subdivision

### LINKED ISSUE #2286: Maintain order of rays/variables after toric blow_up
Currently, the toric `blow_up` function sometimes changes the order of the rays/variables in the output relative to the input toric variety. This may result in user error when performing a sequence of blowups. It would be good to maintain the order of the original variables and then append the blowup variable to the end of the list.

@HereAround

lkastner:
Our implementation produces a non-redundant fan description, so we can use the special constructor for fans.


HereAround (approved):
Thank you for the quick fix to https://github.com/oscar-system/Oscar.jl/issues/2286. Looks great!

PR #2291, opened 2023-04-19
Rework of `minimal_generating_set` for graded polynomial ideals and graded quotient ideals

RafaelDavidMohr:
This PR reworks both methods for the `minimal_generating_set` function to use `Singular.mstd` which computes and returns a Gröbner basis and a minimal generating set for the ideal in question. 

In the new methods we 

a) Make sure to pass any known Gröbner basis to `Singular.mstd` in the hopes that it doesn't recompute one.
b) Make sure to cashe the Gröbner basis computed by `Singular.mstd`.


src/Rings/MPolyQuo.jl
```diff
@@ -1550,50 +1550,38 @@ return an array containing a minimal set of generators of `a`.
...
-  IS = I.gens.S
-  GC.@preserve IS QS begin
-    ptr = Singular.libSingular.idMinBase(IS.ptr, QS.ptr)
-    gensS = gens(typeof(IS)(QS, ptr))
-  end
-
-  i = 1
-  while i <= length(gensS)
-    if iszero(gensS[i])
-      deleteat!(gensS, i)
```
joschmitt, line 1584: Without this, the following happens:
```
julia> R, (x, y) = grade(polynomial_ring(QQ, [ "x", "y"])[1], [ 1, 2 ])
(Multivariate Polynomial Ring in x, y over Rational Field graded by 
  x -> [1]
  y -> [2], MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}[x, y])

julia> I = ideal(R, [ x*y ])
ideal(x*y)

julia> Q, _ = quo(R, I)
(Quotient of Multivariate Polynomial Ring in x, y over Rational Field graded by 
  x -> [1]
  y -> [2] by ideal(x*y), Map from
R to Q defined by a julia-function with inverse)

julia> J = ideal(Q, [ x*y ])
ideal(x*y)

julia> minimal_generating_set(J)
1-element Vector{MPolyQuoRingElem{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}}:
 0
```
So an ideal that is trivial in the quotient ring gets minimal generating set `[0]`. This is why the tests in the `LinearQuotient`-code fail. Is this behaviour of `minimal_generating_set` intended?

RafaelDavidMohr: Thanks for the hint! I just pushed a change to remove zero from the list of generators.

src/Rings/MPolyQuo.jl
```diff
@@ -1574,13 +1574,13 @@ function minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})
...
     _, sing_min = Singular.mstd(I.gb.gens.S)
-    return (Q).(gens(sing_min))
+    return filter(!iszero, (Q).(gens(sing_min)))
   else
     singular_assure(I)
     sing_gb, sing_min = Singular.mstd(I.gens.gens.S)
     I.gb = IdealGens(I.gens.Ox, sing_gb, true)
     I.gb.gens.S.isGB = I.gb.isGB = true
-    return (Q).(gens(sing_min))
+    return filer(!iszero, (Q).(gens(sing_min)))
```
joschmitt: `filer` -> `filter`

src/Rings/MPolyQuo.jl
joschmitt: Thanks!

wdecker: So what's the return value for the zero ideal now?

RafaelDavidMohr: An empty list, this should be the same behaviour as the old `minimal_generating_set` methods.

wdecker: Both Macaulay2 and Singular return 0. There must be reasons for this: @hannes14

hannes14: That is a convention: Singular generates the zero ideal by a single 0.

joschmitt: I can also make a small fix to make my code (and the tests) work with `[0]`, if that is the convention we follow.

wdecker: I talked to @fieker and we agree in that we think there is no "best" solution for this. So let us pick one solution: I suggest to go along with Macaulay2 and Singular: We return the vector consistsing of the zero element of the given ring.

thofma: Is it allowed to ask why? This is not a minimal generating set and contradicts the docstring. This is confusing from a user point of view.

fingolfin: I agree with @thofma that this is problematic in that it contradicts the documentation. Also from a "pure" point of view, it is not right.

However, I also know from experience in GAP that people tend to write code with an implicit assumptions that generator lists are never empty... In GAP a common reason is that often only generator lists are passed around, *without* a matching parent object. And then if you just have an empty list, you cannot recover the parent. I suspect a similar reason might be behind the decision in Singular and Macaulay: code is lazy and/or *needs* an element to access a parent?

Whatever it is, I'd really prefer the empty list.

Whatever we do, I think it'd be worthwhile to point this out in the manual, as either behavior will throw off some people: returning `[0]` is not minimal, so must be mentioned; returning an empty list is just something people forget can happen, so it's also helpful to mention.

(Oh, and of course it should not be `return []`, rather the code should be type stable and so `return SomeType[]`)

fingolfin: (A doctest with a zero ideal might be useful, too)

src/Rings/mpoly-ideals.jl
```diff
@@ -1334,24 +1334,22 @@ function minimal_generating_set(I::MPolyIdeal{<:MPolyDecRingElem}; ordering::Mon
...
+    singular_assure(G, G.ord)
+    G.gens.S.isGB = true
+    _, sing_min = Singular.mstd(G.gens.S)
+    return filter(!iszero, (R).(gens(sing_min)))
+  else
+    singular_assure(I)
+    sing_gb, sing_min = Singular.mstd(I.gens.gens.S)
+    ring = I.gens.Ox
+    computed_gb = IdealGens(ring, sing_gb, true)
+    I.gb[eval(ordering(ring))(ring)] = computed_gb
```
fingolfin: Wait, `eval`? That sounds super fragile; also can't orderings be more than symbols these days?

Also, what is the difference between `ordering(ring)` and `default_ordering(ring`) ?

test/Rings/mpoly-graded-test.jl
```diff
@@ -208,6 +208,7 @@ end
   R, (x, y) = grade(polynomial_ring(QQ, [ "x", "y"])[1], [ 1, 2 ])
   I = ideal(R, [ x^2, y, x^2 + y ])
   @test minimal_generating_set(I) == [ y, x^2 ]
+  @test haskey(I.gb, eval(ordering(R))(R))
```
fingolfin: Another use of `eval`, which I find problematic

src/Rings/MPolyQuo.jl
wdecker: O.k:, if we will guarantee type stabilty as suggested my @fingolfin, it is fine with me to return the empty list.

fingolfin (changes requested):

src/Rings/MPolyQuo.jl
joschmitt: Just for entertainment: there is a test (not a doctest) testing this:
```
R, (x, y) = grade(polynomial_ring(QQ, [ "x", "y"])[1], [ 1, 2 ])
@test minimal_generating_set(ideal(R, [ R() ])) == elem_type(R)[]
```
(see `test/Rings/mpoly-graded-test.jl` line 212). That test works also with the earlier version of this pull request... Apparently Singular does not always return a non-empty list?

src/Rings/mpoly-ideals.jl
RafaelDavidMohr: I removed the call to `eval` just know, yesterday doing it like this felt good but I can't remember why (and safe because `ordering` always returns a symbol).

src/Rings/MPolyQuo.jl
RafaelDavidMohr: >Whatever we do, I think it'd be worthwhile to point this out in the manual, as either behavior will throw off some people: returning [0] is not minimal, so must be mentioned; returning an empty list is just something people forget can happen, so it's also helpful to mention.

See https://github.com/oscar-system/Oscar.jl/pull/2291/commits/2f89770702d0c7e2f13f66565300dce9c131ba0f for a corresponding change to the docs, let me know if I should also add it explicitly as a doc test.

src/Rings/MPolyQuo.jl
```diff
@@ -1544,56 +1544,45 @@ end
...
 julia> minimal_generating_set(a)
 2-element Vector{MPolyQuoRingElem{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}}:
- x
+ y
  z^2
 ```
 """
-function minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem}; ordering::MonomialOrdering = default_ordering(base_ring(base_ring(I))))
+function minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})
   # This only works / makes sense for homogeneous ideals. So far ideals in an
```
afkafkafk13, line 1567: Stupid question (not to the code of this PR, but to the minimal generating set): 
The notion also makes sense in a local ring as emulated in Singular by a choice of a local ordering for the localization at the origin. Do we envision to also support the local case in the future?

wdecker:
See Issue #2295

fingolfin (approved):
Thank you. This seems sensible to me.

I cannot judge the mathematical details, though, hence I'd appreciate if some of the people working on algebraic geometry / commutative algebra would also sign off on this

src/Rings/MPolyQuo.jl
wdecker: @RafaelDavidMohr Thank you for changing the docs, this looks fine to me. Please add it also to the doc tests in both cases. @fingolfin Otherwise, from my point of view, this is ready to me merged.

ederc (approved):

fingolfin:
Test fail, but for good reasons: a test that is marked as broken now passes

RafaelDavidMohr:
As @ederc found out, the reason is that the new `minimal_generating_set` function (based on `Singular.mstd`) may produce a different  minimal generating set than the old function (based on `Singular.libSingular.idMinBase`). 

The test above is based on an example produced by `minimal_generating_set`. The test doesn't break anymore because a different minimal generating set than before is produced. I've hardcoded the old minimal generating set into the test now, the tests should break "as desired".

PR #2296, opened 2023-04-20
Use kwarg 'algorithm' instead of 'alg', 'algo'

### LINKED ISSUE #2258: Unify kwarg names `alg` vs. `algo` vs. `algorithm` vs other
A bunch of functions allow selecting an algorithm to use for the computation via a keyword argument. I think we should standardize the name for the kwarg used for this. Right now we have code using either of these:
1. `alg`
2. `algo`
3. `algorithm`
4. ... perhaps also others?

For consistency we should settle for a single one. Our rules would suggest that should be `algorithm`. But since this is a rather technical knob, I could also live with `algo`. I really dislike `alg`, though, as it can be confused with "algebra"

fingolfin:
Resolves #2258

I used `algorithm`, which @thofma and me prefer, but I know @fieker prefers `algo`. I can certainly live with that, too, but in order to make some progress, I decided to just make this PR -- if necessary it can still be changed to use `algorithm`

### DIFF
--- a/docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/homological_algebra.md
+++ b/docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/homological_algebra.md
@@ -40,7 +40,7 @@ homology(C::ComplexOfMorphisms{<:ModuleFP}, i::Int)
 ## Hom and Ext
 
 ```@docs
-hom(M::ModuleFP, N::ModuleFP, alg::Symbol=:maps)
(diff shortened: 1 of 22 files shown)


thofma (approved):

simonbrandhorst (approved):

fingolfin:
I've rebased this now. Unless @fieker objects, I suggest we merge this later today.

fingolfin:
Just talked to @fieker about this and he is willing to accept this compromise (or at least he didn't care enough to put up a fight ;-) ).

PR #2297, opened 2023-04-20
Yet more uses of @req (plus some other random cleanup)

fingolfin:
Also got rid of various "i.e. the same OSCAR variable" occurrences in
error messages -- IMHO these are more confusing then helpful. What even
is an "OSCAR variable"?

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
@@ -217,17 +217,13 @@ end
 ####################################################
 
 function Base.:+(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
-    if toric_variety(ac1) !== toric_variety(ac2)
-        throw(ArgumentError("The rational equivalence classes must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(ac1) === toric_variety(ac2) "The rational equivalence classes must be defined on the same toric variety"
     return RationalEquivalenceClass(toric_variety(ac1), polynomial(ac1) + polynomial(ac2))
 end
 
 
 function Base.:-(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
-    if toric_variety(ac1) !== toric_variety(ac2)
-        throw(ArgumentError("The rational equivalence classes must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(ac1) === toric_variety(ac2) "The rational equivalence classes must be defined on the same toric variety"
     return RationalEquivalenceClass(toric_variety(ac1), polynomial(ac1) - polynomial(ac2))
 end
 
@@ -242,9 +238,7 @@ Base.:*(c::T, ac::RationalEquivalenceClass) where {T <: IntegerUnion} = Rational
 ####################################################
 
 function Base.:*(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
-    if toric_variety(ac1) !== toric_variety(ac2)
-        throw(ArgumentError("The rational equivalence classes must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(ac1) === toric_variety(ac2) "The rational equivalence classes must be defined on the same toric variety"
     return RationalEquivalenceClass(toric_variety(ac1), polynomial(ac1) * polynomial(ac2))
 end
 
@@ -253,25 +247,19 @@ Base.:^(ac::RationalEquivalenceClass, p::T) where {T <: IntegerUnion} = Rational
 
 
 function Base.:*(ac::RationalEquivalenceClass, sv::ClosedSubvarietyOfToricVariety)
-    if toric_variety(ac) !== toric_variety(sv)
-        throw(ArgumentError("The rational equivalence class and the closed subvariety must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(ac) === toric_variety(sv) "The rational equivalence class and the closed subvariety must be defined on the same toric variety"
     return ac * rational_equivalence_class(sv)
 end
 
 
 function Base.:*(sv::ClosedSubvarietyOfToricVariety, ac::RationalEquivalenceClass)
-    if toric_variety(ac) !== toric_variety(sv)
-        throw(ArgumentError("The rational equivalence class and the closed subvariety must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(ac) === toric_variety(sv) "The rational equivalence class and the closed subvariety must be defined on the same toric variety"
     return ac * rational_equivalence_class(sv)
 end
 
 
 function Base.:*(sv1::ClosedSubvarietyOfToricVariety, sv2::ClosedSubvarietyOfToricVariety)
-    if toric_variety(sv1) !== toric_variety(sv2)
-        throw(ArgumentError("The closed subvarieties must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(sv1) === toric_variety(sv2) "The closed subvarieties must be defined on the same toric variety"
     return rational_equivalence_class(sv1) * rational_equivalence_class(sv2)
 end
 
--- a/src/AlgebraicGeometry/ToricVarieties/Subvarieties/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/Subvarieties/constructors.jl
@@ -6,12 +6,8 @@
     toric_variety::AbstractNormalToricVariety
     defining_ideal::MPolyIdeal
     function ClosedSubvarietyOfToricVariety(toric_variety::AbstractNormalToricVariety, defining_ideal::MPolyIdeal)
-      if !is_simplicial(toric_variety)
-        throw(ArgumentError("Currently, closed subvarieties are only supported for simplicial toric varieties"))
-      end
-      if parent(gens(defining_ideal)[1]) != cox_ring(toric_variety)
-        throw(ArgumentError("The defining ideal must be contained in the Cox ring of the toric supervariety"))
-      end
+      @req is_simplicial(toric_variety) "Currently, closed subvarieties are only supported for simplicial toric varieties"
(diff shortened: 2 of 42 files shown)


src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
```diff
@@ -217,17 +217,13 @@ end
 ####################################################
 
 function Base.:+(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
-    if toric_variety(ac1) !== toric_variety(ac2)
-        throw(ArgumentError("The rational equivalence classes must be defined on the same toric variety, i.e. the same OSCAR variable"))
-    end
+    @req toric_variety(ac1) === toric_variety(ac2) "The rational equivalence classes must be defined on the same toric variety"
     return RationalEquivalenceClass(toric_variety(ac1), polynomial(ac1) + polynomial(ac2))
 end
```
HereAround, line 222: @fingolfin Regarding the `OSCAR variable`: The key question here is what `same` means. In theory, we could say it is sufficient if `toric_variety(ac1)` and `toric_variety(ac2)` are isomorphic, for then we could apply this isomorphism say to `ac1` and then compare the result with `ac2`.

In practice, we do not do this. (Among others because it can be challenging to find such an isomorphism.) For an illustrative example, consider the following:
```julia
P2a = projective_space(NormalToricVariety, 2)
P2b = projective_space(NormalToricVariety, 2)
```
Then we find that `P2a == P2b` (as well as `P2a === P2b`) both return `false`. And so, besides `P2a` and `P2b` being isomorphic via the identity, we (at least as of now) cannot compare rational equivalence classes on `P2a` with those on `P2b`.

Rather, for addition, subtraction, comparison etc. of two rational equivalence classes, they would both have to be defined on `P2a` (or both on `P2b`). We tried to describe this fact by saying that the rational equivalence classes would have to be defined on the "same OSCAR variable" `P2a` (respectively `P2b`). Is there a better formulation that comes to your mind? We can of course also add a description, similar to this message. What do you think?

(This also applies mutually in a couple of other places, e.g. for toric divisors.)

HereAround (approved):
Thank you for these changes @fingolfin . I like them a lot!

(Regarding the `OSCAR variable`, I have elaborated more on this below.)

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
thofma: Is the wording "blabla must be equal" ambiguous? Would people equate this with being isomorphic?

fingolfin: I understand why you used "same OSCAR variable" now, but I still think it's confusing and doesn't make sense (no offense intended).

I don't think anyone would expect "isomorphic" to be meant by "same". But yes, someone might think that two rings that "look the same" are "the same"... Though personally "same" for me carries a strong connotation of "identical" (i.e., `===`), which is more than "equal" (i.e., `==`).

So to me, the message as it is now is perfectly clear, but of course some people will read "same" as "equal" and then might be confused... Perhaps using the word "identical" would help them; perhaps not. Using "the same OSCAR variable" does not help, though, because it (a) still uses "same" (so if you think it means `==` or even "are isomorphic" the first time around, why wouldn't you think the same the second time around?), and (b) "OSCAR variable" is a non-existing concept, and even "Julia variable" is wrong. Perhaps "must be identical objects"  would be helpful?


All that said: In the end, no matter what you do, you can't stop people from bashing their head against this. Even if you write "must be identical as in '===' must return true for both objects" some people will either (a) not read the error and just complain that they don't understand the error, or (b) complain that this is an unjust restricting and the comparison should be changed to "==". This is a problem that is not specific to math either, it affects most people who do non-trivial programming at some point.

lkastner: We are not writing software for us, rather for mathematicians. So yes, they will not understand this. They will probably already be confused by `!(P2a == P2b)`. I thought that the sentence "i.e. the same OSCAR/Julia variable" at least gives a hint that this is not mathematical equality. At least with that sentence they will be confused for the right reason.

src/AlgebraicGeometry/ToricVarieties/Subvarieties/constructors.jl
```diff
@@ -6,12 +6,8 @@
...
     defining_ideal::MPolyIdeal
     function ClosedSubvarietyOfToricVariety(toric_variety::AbstractNormalToricVariety, defining_ideal::MPolyIdeal)
-      if !is_simplicial(toric_variety)
-        throw(ArgumentError("Currently, closed subvarieties are only supported for simplicial toric varieties"))
-      end
-      if parent(gens(defining_ideal)[1]) != cox_ring(toric_variety)
-        throw(ArgumentError("The defining ideal must be contained in the Cox ring of the toric supervariety"))
-      end
+      @req is_simplicial(toric_variety) "Currently, closed subvarieties are only supported for simplicial toric varieties"
+      @req parent(gens(defining_ideal)[1]) == cox_ring(toric_variety) "The defining ideal must be contained in the Cox ring of the toric supervariety"
```
fingolfin, line 10: BTW I should point out that I did not try to address the question whether to use `==` or `===` to compare parents, base rings etc. -- but while making this PR I noticed that we do both frequently, and there is no clear pattern.

Perhaps that's OK; or perhaps we should agree to e.g. always use `===` and then add this to the developer guide, and announce it to all devs (e.g. via oscar-dev).

Meta: perhaps that's something we should indeed do more often (announcing changes on how we do things on oscar-dev), as mostly nobody is able to follow all PRs and Slack communication etc., so it's easy to miss out on some change. E.g. that we don't use `Markdown.@doc` anymore for docstrings anymore.

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
HereAround: I agree: With this explanation in mind, the error message is now perfectly clear. So let us change it as you proposed @fingolfin . Thank you!

src/Rings/MPolyQuo.jl
```diff
@@ -106,28 +106,22 @@ end
...
-     end
+   if length(B) >= 1 && is_graded(R)
+     @req all(is_homogeneous, B.gens.O) "The generators of the ideal must be homogeneous"
    end
    return r
   end
 
   function MPolyQuoIdeal(Ox::MPolyQuoRing{T}, I::MPolyIdeal{T}) where T <: MPolyRingElem
-    base_ring(Ox) === base_ring(I) || error("base rings must match")
+    @req base_ring(Ox) === base_ring(I) "base rings must match"
```
fingolfin, line 124: e.g. here we compare base rings with `===` ...

src/Rings/MPolyQuo.jl
```diff
@@ -298,7 +292,7 @@ ideal(x + y, x^2 + y^2)
 ```
 """
 function Base.:+(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}) where T
-  base_ring(a) == base_ring(b) || error("base rings must match")
+  @req base_ring(a) == base_ring(b) "base rings must match"
```
fingolfin, line 295: .. and here we compare base rings with `==` (in the same file)

src/AlgebraicGeometry/ToricVarieties/Subvarieties/constructors.jl
HereAround: I would indeed appreciate announcements on such general changes.

Regarding `==` vs. `===`, that is yet another good point...

src/Rings/MPolyQuo.jl
```diff
@@ -530,7 +524,7 @@ end
 
 
 function check_parent(a::MPolyQuoRingElem, b::MPolyQuoRingElem)
-  a.P == b.P || error("wrong parents")
+  @req parent(a) == parent(b) "parents must match"
```
fingolfin, line 527: Hmmm actually perhaps this `@req` is not appropriate? But then again I am not sure because I don't know what the semantics of `check_parent` are supposed to be ... we have dozens and dozens of methods for it, but I couldn't find any documentation for it (@thofma @fieker ?)

src/Rings/MPolyQuo.jl
thofma: It is the same as `==`.

src/Rings/MPolyQuo.jl
thofma: It is a helper introduced for matrices to not create parent objects when comparing parents.

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
thofma: I like "must be identical blabla" (this is also what the docstring for `===` uses).

Alternatively and with more CS lingo: "must be egal".

src/Rings/MPolyQuo.jl
thofma: Why should the `@req` not be appropriate?

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
fingolfin: @lkastner but there is no variable involved. I don't see how that could be illuminating to anyone? If the only thing it does is to give a hint that something "non-mathematical" happens (and for the record, I strongly disagree with this!), then we should rather spell it out ("here identical refers to the programming concept, this is stricter than mathematical equality"). But that seems cumbersome.

Anyway: I am happy to have this use `identical` consistently. But that affects many places, not just those touched by this PR, and possibly in other repos, too, so I'd prefer to not do it in this PR.

src/Rings/MPolyQuo.jl
fingolfin: OK, I just discovered that `check_parent` is described in the *Nemo* documentation, but not the AA documentation where I was searching (since AA uses `check_parent` extensively, I assumed it'd be explained there or nowhere).

(Funnily, that documentation suddenly uses the name `parent_check` instead, presumably that was an older name for it)

As to why `@req` might not be appropriate: well I wasn't sure what it is meant to be used for. But since it seems only input validation, I guess it'd be OK. Though, I should point that this `check_parent` methods used `error` before. And all other `check_parent` methods I checked (only a handful I randomly picked) also use `error`. So we are getting inconsistent. But OK, one step after the other.

PR #2303, opened 2023-04-21
`cyclic_generator` for groups, `is_faithful` for characters; make `findfirst(fun, t::GAPGroupCharacterTable)` work

ThomasBreuer:
(suggested by @fieker )

- `cyclic_generator` for groups
- `is_faithful` for characters
- make `findfirst(fun, t::GAPGroupCharacterTable)` work


fieker (approved):

src/Groups/group_constructors.jl
```diff
@@ -127,6 +127,27 @@ i.e., if `G` can be generated by one element.
...
+Group([ (1,2,3), (4,5) ])
+
+julia> cyclic_generator(g)
+(1,2,3)(4,5)
+```
+"""
+function cyclic_generator(G::GAPGroup)
+  @req is_cyclic(G) "the group is not cyclic"
+  length(gens(G)) == 1 && return gen(G,1)
+  return group_element(G, GAPWrap.MinimalGeneratingSet(G.X)[1])
```
fingolfin, line 149: Can we also get a `minimal_generating_set` function (not in this PR, but in a follow-up perhaps)?

fingolfin (approved):

src/Groups/group_constructors.jl
ThomasBreuer: Is GAP's `MinimalGeneraingSet` the right thing to ask for? Often it is sufficient to have a generating set of minimal length.

thofma:
Would another option be `generator(G)` instead of `cyclic_generator(G)`? I never heard of a cyclic generator before.

fieker:
On Fri, Apr 21, 2023 at 05:11:56AM -0700, Tommy Hofmann wrote:
> Would another option be `generator(G)` instead of `cyclic_generator(G)`? I never heard of a cyclic generator before.
So far, generator usually refers to the natural one that is used: 
C2 x C3 would have so far 2 gens of order 2 and 3. The new function will
produce some element of order 6

fieker:
On Fri, Apr 21, 2023 at 05:11:56AM -0700, Tommy Hofmann wrote:
> Would another option be `generator(G)` instead of `cyclic_generator(G)`? I never heard of a cyclic generator before.
> 
is_cyclic_with_data...

thofma:
Yes, `generator(G)` means give me one generator if it exists. It is not related to `generators(G)`.

fingolfin:
Do you mean `gen(G)` versus `gens(G)` (versus `gen(G, i)`)? Because right now there are no functions `generators` / `generator` that work in general (well, there is `generators` for  `AbstractLat`... but that's all?)

fingolfin:
What I like about `cyclic_generator` is that it is kind of obvious what it does. While `gen(G)` can be a typo for `gens(G)`. I mean, it logically "makes sense" and seems cute, but perhaps it is *too* cute...

ThomasBreuer:
Would `generator` then be something that is interesting in other contexts as well?
For field extensions, `generator` would return a primitive element.
For univariate polynomial rings, `generator` would return the indeterminate.

fingolfin:
That said, I don't mind much if `cyclic_generator` is renamed to `gen` *shrug*

thofma:
I my mind the functions `gen` and `gens` are already changed to `generator` and `generators` (the name they should have).

OK, maybe `cyclic_generator` is the best we can do.

fingolfin:
Actually, will `cyclic_generator` work for trivial groups?

thofma:
> Actually, will `cyclic_generator` work for trivial groups?

Hm, I don't see a reason why not. Can you elaborate?

ThomasBreuer:
> Actually, will cyclic_generator work for trivial groups?

Ah, I will update the pull request.
Thanks for the hint.
(And we have found a bug in GAP's grpconst package.)

PR #2306, opened 2023-04-23
TropicalGeometry: groebner_fan.jl, new implementation for Groebner fans

YueRen:
Main new command:
- groebner_fan

Secondary new commands:
- maximal_groebner_cone
- homogeneity_space
- homogeneity_vector

@martapanizzut @ederc Would you be so kind and review this pull request please? I'm happy to give you a quick overview of the code over Zoom if you want.


YueRen:
Hm, any introductory tips for what I must do to make codecov bot happy?

benlorenz:
The coverage went down because the doctests failed and most of the failures I found are because the code uses `contains(Cone, Vector)` which was deprecated recently:
```
ERROR: `contains(C::Cone, v::AbstractVector)` is deprecated, use `Base.in(v, C)` instead.
│    Stacktrace:
│     [1] depwarn(msg::String, funcsym::Symbol; force::Bool)
│       @ Base ./deprecated.jl:82
│     [2] depwarn(msg::String, funcsym::Symbol)
│       @ Base ./deprecated.jl:80
│     [3] contains(C::Cone{QQFieldElem}, v::Vector{ZZRingElem})
│       @ Oscar ./deprecated.jl:71
│     [4] homogeneity_vector(G::Vector{QQMPolyRingElem})
│       @ Oscar ~/work/Oscar.jl/Oscar.jl/src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl:123
│     [5] top-level scope
│       @ none:1
```
Just change these to `v in C`.

YueRen:
Ah, got it.  I fixed it and changed the test whether `homogeneityVector` is inside the positive orthant from  `homogeneityVector>zeros(fmq,length(homogeneityVector))` to `isnothing(findfirst(vi -> (vi<=0), homogeneityVector))` which should be minimally faster. I also rebased it to the latest master just to make sure that all my examples still work.

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    end
+    return cone_from_inequalities(inequalities,matrix(QQ,equations))
+end
+
+
+
+@doc raw"""
+    homogeneity_vector(G::Vector{<:MPolyElem})
+
+If it exists, a positive `Vector{fmpz}` under which every of `G` is weighted homogeneous.  Otherwise, `nothing`.
```
ederc:
```suggestion
If it exists, a positive `Vector{fmpz}` under which every element of `G` is weighted homogeneous.  Otherwise, `nothing`.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+
+# Note
+Suppose `G` is a reduced Groebner basis of an ideal `I` with respect to a global ordering. If a `Vector{fmpz}` is returned, then `I` will be weighted homogeneous with respect to it.  If `nothing` is returned, then `I` is not weighted homogeneous with respect to any positive weight vector.
```
ederc:
```suggestion
Suppose `G` is the reduced Groebner basis of an ideal `I` with respect to a global ordering. If a `Vector{fmpz}` is returned, then `I` is weighted homogeneous with respect to it.  If `nothing` is returned, then `I` is not weighted homogeneous with respect to any positive weight vector.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    return nothing
+end
+
+
+
+@doc raw"""
+    maximal_groebner_cone(G::Vector{<:MPolyElem}, ord::MonomialOrdering, homogeneityWeight::Union{Nothing,Vector{fmpz}}=nothing)
+
+# Input
+- `G`: A reduced Groebner basis of an ideal `I` with respect to `ord`,
```
ederc:
```suggestion
- `G`: The reduced Groebner basis of an ideal `I` with respect to `ord`,
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+- `ord`: A global monomial ordering,
+- `homogeneityWeight`: Either `Nothing` or a positive weight vector under which `I` is weighted homogeneous.
```
ederc:
```suggestion
- `homogeneityWeight`: Either `nothing` or a positive weight vector under which `I` is weighted homogeneous.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
ederc: Wouldn't it make more sense to return a zero vector instead of `nothing` if no such homgeneity vector exists?

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    # means that C needs to be restricted to the positive orthant
+    C = maximal_groebner_cone_extended(G,ord)
+    positiveOrthant = positive_hull(identity_matrix(ZZ,ambient_dim(C)))
+    return intersect(C,positiveOrthant)
+end
+function maximal_groebner_cone(G::Vector{<:MPolyElem}, ord::MonomialOrdering, homogeneityWeight::Vector{fmpz})
+    # calling with a positive homogeneityWeight
+    # means that C does not be restricted to the positive orthant
+    return maximal_groebner_cone_extended(G,ord)
+end
```
ederc: Here I am again have to ask: Instead of `nothing` wouldn't it be easier to have an all zeroes vector?

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    # otherwise compute the sum of rays, cast it to a rational vector
+    # and return a primitive integer vector
+    pt = Vector{fmpq}(sum(R))
+    return numerator.(pt .* lcm(denominator.(pt)))
+
+end
+
+
+
+# Returns an interior points on each facet.
```
ederc:
```suggestion
# Returns an interior point on each facet, respectively.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    return normals
+end
+
+
+
+# Returns the initial form of polynomial g with respect to weight vector u.
+# Requires a monomial ordering that is compatible with respect to u, i.e., the leading monomial is of highest weighted degree.
+function initial(g::MPolyElem, ordering::MonomialOrdering, u::Vector{fmpz})
+
+    lt,tail = Iterators.peel(terms(g,ordering=ordering));
```
ederc, line 272: There is also `leading_term()` and `tail()` available in Oscar.

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    return G
+
+end
+
+
+
+# Assumes:
+# - Hnew is a Groebner basis of the initial ideal w.r.t. ordH
+# - G is a Groebner basis of the original ideal w.r.t. ordG
+# - ordH and ordG are adjacent orderings
```
ederc, line 307: Is the adjacency of `ordG` and `ordH` tested in `groebner_lift`?

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+end
+
+
+
+# Assumes:
+# - G is a Groebner basis with respect to ordG
+# - u is an interior facet point of the maximal Groebner cone around ordG
+# - v is an outer normal vector of the facet containing u
+# Returns:
+# - a reduced Groebner basis with respect to the adjacent ordering
```
ederc:
```suggestion
# - the reduced Groebner basis with respect to the adjacent ordering
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
YueRen: That is very true.  I guess I just wanted to play around with type dispatch and use it to distinguish between the (quasi-)homogeneous and the non-homogeneous case.

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
YueRen: No, it is not.  I left it out because it requires some non-trivial computations and it is not necessary.

To test the adjacency under the assumptions that `Hnew` and `G` are Groebner bases w.r.t. `ordH` and `ordG`, respectively, one would need to verify
1. that there is a weight vector `w` with `initial_w(G)=Hnew` (as sets of polynomials),
2. `lead(G,ordG)` and `lead(G,ordH)` are not the same (as sets of polynomials).

Point 1 implies that `ordG` and `ordH` are either adjacent or the same, Point 2 implies that they are not the same. (Note that "adjacent" here means that the Groebner cones share a common face containing `w`, but that face need not be one-codimensional)

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+#   coefficient case. The convention is always max, as in OSCAR leading monomials
+#   under weighted orderings are the ones with maximal weighted degree.
+#
+###
+
+# Computes the space of weight vectors w.r.t. which G is weighted homogeneous
+@doc raw"""
+    homogeneity_space(G::Vector{<:MPolyElem})
+
+A `Cone` that is the space of all weight vectors under which all elements of `G` are weighted homogeneous.  The cone will be a linear subspace without any rays.
```
fingolfin:
```suggestion
Return a `Cone` that is the space of all weight vectors under which all elements of `G` are weighted homogeneous.  The cone will be a linear subspace without any rays.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+
+# Note
+If `G` is a reduced Groebner basis of an ideal `I` with respect a global ordering, then the returned cone is the space of all weight vectors with respect to which `I` is weighted homogeneous.
```
fingolfin: Please avoid lines which are too long (say < 100 chars, though as a rule of thumb I use between 78 and 92 lines). This helps with reviewing changes to these lines in GitHub and via `git` diff. It's just reality that if later we need to fix a typo in a line with for example 300 characters, it is a hassle to review this. 

Separately: maybe use an [admonition](https://docs.julialang.org/en/v1/stdlib/Markdown/#Admonitions) for notes.

```suggestion
!!! note
    If `G` is a reduced Groebner basis of an ideal `I` with respect a global
    ordering, then the returned cone is the space of all weight vectors with
    respect to which `I` is weighted homogeneous.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    end
+    return cone_from_inequalities(inequalities,matrix(QQ,equations))
+end
+
+
+
+@doc raw"""
+    homogeneity_vector(G::Vector{<:MPolyElem})
+
+If it exists, a positive `Vector{fmpz}` under which every element of `G` is weighted homogeneous.  Otherwise, `nothing`.
```
fingolfin:
```suggestion
Return a positive `Vector{fmpz}` under which every element of `G` is weighted
homogeneous. If no such vector exists, return `nothing`.
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+ 4
+
+julia> G3 = [x1^2+x2,x2^2+x3,x3+1]
+3-element Vector{QQMPolyRingElem}:
+ x1^2 + x2
+ x2^2 + x3
+ x3 + 1
+
+julia> homogeneity_vector(G3)
+
```
fingolfin: Should there be a return value here?

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
fingolfin: In which sense would it be easier?

(Don't get me wrong, I am not opposed, I'd just like to understand the argument).

I see pros and cons to both. But at least `nothing` makes it less likely that one accidentally uses a zero vector as if it was a weight vector?

The downside of `nothing` I can think of is that one has to be careful to preserve type stability. But Julia does have provisions for that: if one branches based on the return type then Julia is able to deduce that correctly. E.g. consider this example:
```
julia> f(x) = x > 0 ? x : nothing
f (generic function with 1 method)

julia> function g(x)
         y = f(x)
         y === nothing && return 0
         return y + 1
       end
g (generic function with 1 method)

julia> @code_warntype g(7)
MethodInstance for g(::Int64)
  from g(x) in Main at REPL[13]:1
Arguments
  #self#::Core.Const(g)
  x::Int64
Locals
  y::Union{Nothing, Int64}
Body::Int64
1 ─      (y = Main.f(x))
│   %2 = (y === Main.nothing)::Bool
└──      goto #3 if not %2
2 ─      return 0
3 ─ %5 = (y::Int64 + 1)::Int64
└──      return %5
```
Here initially the type for `y` is inferred as `Union{Nothing, Int64}`. But after the `y === nothing && return 0` Julia is clever enough to deduce that now `y` really is an `Int`. This also works when using `isnothing(y)` or `y isa Nothing`.

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+    groebner_fan(I::MPolyIdeal; return_groebner_bases::Bool=false, return_orderings::Bool=false, verbose_level::Int=0)
+
+# Input
+- `I`: A multivariate polynomial ideal over a coefficient field
+- `return_groebner_bases`: If `true`, return all Groebner bases computed during the traversal
+- `return_orderings`: If `true`, return all monomial orderings used during the traversal
+- `verbose_level`: If positive, print out how many cones have been computed during the traversal
+
+# Output
+A `PolyhedralFan` that is the Groebner fan of `I`.
```
fingolfin: This Input/Output format is discouraged by Julia, and I agree with that.  It should ideally only be used if it just can't be reasonably avoided. But here I think it can easily be avoided? E.g. like this:
```suggestion
Return a `PolyhedralFan` representing the Groebner fan of `I`, where is a multivariate polynomial ideal over a field.

If `verbose_level` is positive, also print how many cones have been computed during the traversal.
```
The description of the `return_*` kwargs seems to be redundant with what follows?

BTW also `verbose_level` should perhaps be removed and instead replaced with use of `@vprint` and friends -- but that could be done in a future PR, of course

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,544 @@
...
+
+If `I` is not weighted homogeneous with respect to a positive weight vector, the Groebner fan will be restricted to the positive orthant.  Otherwise, the Groebner fan will span the entire space and have a non-trivial lineality space.
+
+If `return_groebner_bases==true`, returns a dictionary whose keys are interior points of the maximal cones and whose values are the Groebner bases for those cones.  Their union will be a universal Groebner basis.
```
fingolfin: My understanding is that the dictionary is returned *in addition* to the fan, is that right? I'd suggest to make this clear, one way or the other. E.g.:
```suggestion
If `return_groebner_bases==true`, also return a dictionary whose keys are interior points of the maximal cones and whose values are the Groebner bases for those cones.  Their union will be a universal Groebner basis.
```

fingolfin (commented):
I can't comment on the math in here, but I have some nitpicky stylistic comments. They are not meant to annoy you, @YueRen (I apologies if they do), but rather aim at improving consistency on the long term -- in that spirit I'd rather "complain" early to give you chance to react and adapt. But of course I understand this code is under development, so that perhaps you meant to change some of it anyway. Also it's also OK to do some or all of this later.

Finally, don't hesitate to question any of my suggestions :-)

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
YueRen: Quick question: will the line breaks that I enter here manually cause line breaks in the documentation?

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
YueRen: No, the return is `nothing`, but I can see how that can be confusing, I will add a `== nothing` (which will then be `true`) to make it clearer.

martapanizzut (approved):

thofma:
The interface is for `(::Vector{<: MPolyElem}, ::MonomialOrdering)`, but the docstrings speak of Gröbner bases. But in Oscar land these are not considered to be Gröbner bases. Is there a reason why the input is not the type of a Gröbner basis?

thofma:
I would suggest we add some tests? This would also make codecov happy.

YueRen:
@thofma 

> The interface is for (::Vector{<: MPolyElem}, ::MonomialOrdering), but the docstrings speak of Gröbner bases. But in Oscar land these are not considered to be Gröbner bases. Is there a reason why the input is not the type of a Gröbner basis?

I've changed the "official" input type of `Oscar.IdealGens` now, which is what `groebner_basis` returns (though internally I'm still using the old version).

> I would suggest we add some tests? This would also make codecov happy.

I added two more examples for groebner_fan as tests (which take 4 and 6 seconds, respectively), the remaining functions are already well-covered by their examples in the documentation.



The pull-request is almost good to go. Two things are missing:
1. I needed and hence added the function `ordering(G::Oscar.IdealGens) = G.ord`.  But `groebner_basis.jl` is definitely not the right place for it.  Where should I put it?
2. Squashing all commits before the merge.

thofma:
Thanks. I think this needs a rebase before CI runs again.

YueRen:
It's rebased onto the current master and I've squashed all commits.  As soon as all tests are passed, it's ready to be merged.

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,541 @@
...
+```
+"""
+function maximal_groebner_cone(G::Oscar.IdealGens{<:MPolyElem})
+    @assert is_groebner_basis(G)
+    ord = ordering(G)
+    G = collect(G)
+    homogeneityWeight = homogeneity_vector(G)
+    return maximal_groebner_cone(G,ord,homogeneityWeight)
+end
+function maximal_groebner_cone(G::Oscar.IdealGens{<:MPolyElem}; homogeneityWeight::Vector{fmpz}=fmpz[])
```
fingolfin: This is a minor nitpick, and not a merge blocker, but: please put an empty line between the end of one function and the start of the next
```suggestion

function maximal_groebner_cone(G::Oscar.IdealGens{<:MPolyElem}; homogeneityWeight::Vector{fmpz}=fmpz[])
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,541 @@
...
+    # sort G by its leading monomials (result is smallest first)
+    # then reduce G[i] by G[1:i-1] for i=length(G),...,2
+    sort!(G,
+          by=g->leading_monomial(g,ordering=ord),
+          lt=(f,g)->(cmp(ord,f,g)<0))
+    for i in 2:length(G)
+        G[i] = reduce(G[i],G[1:i-1],ordering=ord)
+    end
+    return G
+
```
fingolfin: Another minor nitpick, not a blocker: it seems unusual to leave this empty line between `return` and `end`

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,541 @@
...
+    return Gnew,adjacentOrdering
+
+end
+# helper functions for groebner_flip
+function groebner_flip_adjacent_ordering(R::MPolyRing,
+                                         homogeneityVector::Vector{fmpz},
+                                         interior_facet_point::Vector{fmpz},
+                                         outer_normal_vector::Vector{fmpz})
+
+    return weight_ordering(Int.(homogeneityVector),
```
fingolfin, line 349: Wait, so does that mean all those `Vector{fmpz}` can be safely converted to `Vector{Int}`? So why don't we store them as `Vector{Int}` in the first place?

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,541 @@
...
+    # Post processing
+    ###
+    # construct polyhedral fan and return it if nothing else was required
+    Sigma = polyhedral_fan_from_cones([entry[3] for entry in finishedList])
+    if !return_groebner_bases && !return_orderings
+        return Sigma
+    end
+
+    # otherwise return what was required
+    output = Any[]
```
fingolfin: Use of `Any` here is redundant.
```suggestion
    output = []
```

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,541 @@
...
+    finishedList = typeof(workingList)()     # list of Groebner cones whose neighbours are known
+    finishedFacets = Vector{Vector{fmpz}}()  # list of interior facet points whose facet has been traversed
+
+
+    ###
+    # Fan traversal
+    ###
+    while !isempty(workingList)
+        if verbose_level>0
+            println("#workingList / #finishedList: ",length(workingList)," / ",length(finishedList))
```
fingolfin, line 458: You may want to look into using `@vprint` here instead (in a future revision of this code, not in this PR)

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
```diff
@@ -0,0 +1,541 @@
...
+
+function groebner_fan(I::MPolyIdeal; return_groebner_bases::Bool=false, return_orderings::Bool=false, verbose_level::Int=0)
+
+    ###
+    # Preparation:
+    #   Test whether the ideal is weighted homogeneous with respect to a positive weight vector
+    #   Preference is given to the ones vector for the sake of optimisation
+    ###
+    G = groebner_basis(I,complete_reduction=true)
+    ord = G.ord
```
fingolfin: Presumably here you want to use your new `ordering` method?
```suggestion
    ord = ordering(G)
```

fingolfin (approved):

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
YueRen: A lot of objects in the tropical code lives in the intersection between polymake and singular.  The former uses `fmpz` and the latter uses `int`.   Since changing polymake to `int` makes less sense to me than teaching Singular to use `fmpz`, I decided to keep everything as `fmpz` and convert to `int` only when necessary.  It's easier to remember than deciding on a case-by-case basis when to use `fmpz` and when to use `int`.

YueRen:
Okay, the pull request is ready to be merged:
- I commit the changes suggested by Max on Github
- I changed the empty new lines to be more consistent with the rest of Oscar
- I rebased to the latest version of the master branch
- I squashed all commits into one

I will be merging it oncethe tests complete.

src/AlgebraicGeometry/TropicalGeometry/groebner_fan.jl
YueRen: Fixed in the latest commit.

thofma:
I restarted so that the coverage (should) work.

YueRen:
Tests are passing now, if somebody could merge the pull request that would be much appreciated!

YueRen:
Tests are passing now, if somebody could merge the pull request that would be much appreciated!

PR #2307, opened 2023-04-23
Varieties docu and avoid computing radicals

simonbrandhorst:
The documentation now explanes the relation to schemes as suggested by @wdecker. 

At creaton of a variety from an ideal, no radical computation is involved. 
To achieved this, I added a method called `overlying_scheme(X)` [opposed to the `underlying_scheme(X)`]
which returns any scheme whose reduced subscheme is the `underlying_scheme(X)`. 
Therefore methods can be implemented which do not require knowing the radical/reduced structure. 
Calling `underlying_scheme(X)` will trigger the computation of the radical.


src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -3,28 +3,25 @@
 ########################################################
 
 @doc raw"""
-    affine_algebraic_set(X::Spec; check::Bool=true) -> AffineAlgebraicSet
+    affine_algebraic_set(X::Spec; is_reduced=false, check=true) -> AffineAlgebraicSet
 
-Convert `X` to an `AffineAlgebraicSet` by taking the underlying reduced scheme.
+Convert the `X` to an `AffineAlgebraicSet` by considering its reduced structure.
```
fingolfin: The inserted "the" seems odd to me. Perhaps you meant "Convert the NOUN `X` to" with a missing NOUN to be inserted? Otherwise, remove "the" again?

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -3,28 +3,25 @@
...
-    Xred = X
-  end
-  AffineAlgebraicSet(Xred, check=check)
+function affine_algebraic_set(X::Spec; is_reduced::Bool=false, check::Bool=true)
+  return AffineAlgebraicSet(X, is_reduced=is_reduced, check=check)
 end
 
 @doc raw"""
-    vanishing_locus(I::MPolyIdeal; check::Bool=true)
+    vanishing_locus(I::MPolyIdeal; is_radical=false, check::Bool=true)
```
fingolfin:
```suggestion
    vanishing_locus(I::MPolyIdeal; is_radical::Bool=false, check::Bool=true)
```

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -36,25 +33,21 @@ Vanishing locus
...
 """
-function vanishing_locus(I::MPolyIdeal{<:MPolyElem}; check::Bool=true)
-  if check
-    Irad = radical(I)
-  else
-    Irad = I
-  end
-  X = Spec(base_ring(Irad), Irad)
-  return AffineAlgebraicSet(X, check=check)
+function vanishing_locus(I::MPolyIdeal{<:MPolyElem}; is_radical=false, check::Bool=true)
```
fingolfin:
```suggestion
function vanishing_locus(I::MPolyIdeal{<:MPolyElem}; is_radical::Bool=false, check::Bool=true)
```

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -36,25 +33,21 @@ Vanishing locus
...
+  X = Spec(base_ring(I), I)
+  return AffineAlgebraicSet(X, is_reduced=is_radical, check=check)
 end
 
-affine_algebraic_set(I::MPolyIdeal{<:MPolyElem}; check::Bool=true) = vanishing_locus(I, check=check)
+affine_algebraic_set(I::MPolyIdeal{<:MPolyElem}; kwargs...) = vanishing_locus(I; kwargs...)
 
 @doc raw"""
-    vanishing_locus(p::MPolyRingElem, check::Bool=true)
+    vanishing_locus(p::MPolyRingElem; is_radical=false, check::Bool=true)
```
fingolfin:
```suggestion
    vanishing_locus(p::MPolyRingElem; is_radical::Bool=false, check::Bool=true)
```

However, the actual function `vanishing_locus` below does not seem to accept the kwarg `is_radical`.

Also, `check` remains a positional arg, when this change indicates it is now supposed to be a kwarg.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -82,17 +75,59 @@ affine_algebraic_set(p::MPolyElem, check::Bool) = vanishing_locus(p, check=check
...
-Return the set theoretic intersection of `X` and `Y` as an AlgebraicSet.
+julia> Zred = set_theoretic_intersection(X, Y)
+Vanishing locus
+  in Affine 2-space over Rational Field
+  of ideal(-x^2 + y, y)
 
-This is the reduced subscheme of the scheme theoretic intersection.
+
+```
+Note that the set theoretic intersection forgets the intersection multiplicities
```
fingolfin, line 103: "Forgets" and "remembers" reads oddly to me, but perhaps this is the standard phrasing used in this area of math?

It seems oddly anthropomorphic. Plus something I forgot I might be able to recall later, which is not the case here.

I might instead use "ignores" and "honors"; or "does (not) take into account"; or something like that.

But again, I don't know what's usual in this domain, so feel free to ignore me :-)

simonbrandhorst:
There are still further things to do, for instance unions. But it should be okay for a first version.

docs/oscar_references.bib
```diff
@@ -1425,6 +1425,13 @@ @Article{Shi18
   url           = {https://doi.org/10.1307/mmj/1528941621}
 }
 
+
+@MISC{stacks-project,
+    AUTHOR = "Authors, The Stacks Project",
+    TITLE = "Stacks Project",
+    URL = "https://stacks.math.columbia.edu/"
+}
+
```
afkafkafk13: Something needs to be done to keep the bibtex-test from complaining. But the entry by itself is correct -- except for the fact that there are no authors to be named.

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
```diff
@@ -7,37 +7,97 @@ using Oscar
...
+and $I = (f_1, \dots f_r) \subseteq k[x_1,\dots x_n]$ the ideal they generate,
+we denote by $X = V(I)$ the (affine) algebraic set defined by the ideal $I$
+and call $k$ its base field.
+
+If $k \subseteq K$ is any field extension, we denote the set of $K$-points of $X$ by
+
+$$\begin{aligned}X(K) &= \{ P \in \mathbb{A}^n(K) \mid f_1(P)=\dots = f_n(P)=0\}\\&=\{P \in \mathbb{A}^n(K) \mid \forall f\in I : f(P)=0\}.\end{aligned}$$
+
+Most properties of the algebraic set $X$ refer to $X(K)$ where $K$ is an
+algebraically closed field. For instance `is_empty` returns if $X(K) = \emptyset$.
```
afkafkafk13: ... returns *whether* ...

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
```diff
@@ -7,37 +7,97 @@ using Oscar
...
+## Rational points
+To study the $k$-points, also called $k$-rational points, of the algebraic set $X$
+one first considers the solutions $X(K)$ over an algebraically closed field
+extension $K$ of $k$. Then in a second step one studies $X(k)$ as a subset of $X(K)$.
+
+The first step involves calculations with ideals.
+For instance Hilbert's Nullstellensatz implies that $X(K)$ is empty
+if and only if the ideal $I=(1)$. This is decided by computing a
+Gröbner basis of $I$ and is carried out in $k[x_1,\dots x_n]$
+without taking any field extensions.
```
afkafkafk13: 'This is decided by an ideal membership test relying on a Gröbner basis computation and can be carried out...'

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
```diff
@@ -7,37 +7,97 @@ using Oscar
...
+The second step involves methods from number theory (if $k$ is a number field)
+or from real algebraic geometry (if $k = \mathbb{R}$).
+
+Algebraic sets in Oscar are designed for the first step.
+Most of their properfties should be interpreted as properties
+of the set $X(K)$ of their $K$-points over an algebraic closure $K$.
+
+## Relation to Schemes
+One may a view an (affine) algebraic set as a geometrically reduced (affine)
+scheme over a field $k$.
```
afkafkafk13: remove the third word of line 52 ('a') between 'may' and 'view'.

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
```diff
@@ -7,37 +7,97 @@ using Oscar
...
+!!! note
+    If a construction returns a scheme $Z$, but you want to ignore the scheme
+    structure, call the function `algebraic_set(Z)` to convert the scheme
+    $Z$ to an affine algebraic set.
+
+For example `algebraic_set(intersection(X, Y))`
+is equivalent to `set_theoretic_intersection(X, Y)`.
+
+Internally an `AffineAlgebraicSet` is constructed from a possibly
+non-reduced affine scheme which we call the `fat_scheme` of `X`.
```
afkafkafk13: 'to which we refer as fat scheme for lack of a better word to distinguish it from the reduced scheme 'X_red', which is referred to as underlying_scheme.'

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
```diff
@@ -7,37 +7,97 @@ using Oscar
...
+### More general affine algebraic sets
 By abuse of terminology we say that a scheme is an affine algebraic set
 if it is isomorphic to one. For example a hypersurface complement is an
 affine algebraic set.
-In particular, affine algebraic sets are not necessarily Zariski closed in their
-ambient affine space.
-
-More formally they are defined as follows:
+In particular, we allow affine algebraic sets which are not necessarily
+Zariski closed in their ambient affine space.
```
afkafkafk13, line 86: 'in the ambient affine space in which they have been specified'

docs/src/AlgebraicGeometry/AlgebraicSets/ProjectiveAlgebraicSet.md
```diff
@@ -7,33 +7,43 @@ using Oscar
...
-In Oscar we work with projective algebraic sets over non-closed fields,
-by viewing them as reduced schemes. See [Projective schemes](@ref).
-
-More formally they are defined as follows:
+For finitely many homogeneous polynomials $f_1,\dots f_r \in k[x_0,\dots x_n]$,
+and $I=(f_1,\dots , f_n) \leq k[x_0,\dots x_n]$ the homogeneous ideal they generate, we denote by $X = V(I) \subseteq \mathbb{P}^n$ the
+projective algebraic set defined by $I$ and call $k$ its base field.
+
+Let $\mathbb{P}^n(k)=(k^n\setminus\{0\})/k^*$ be the set of $k$-points of projective space of dimension $n$.
+If $k \subseteq K$ is any field extension, we denote the set of $K$-points of $X$ by
```
afkafkafk13: There is a mismatch of exponents here: you need k^{n+1} \setminus \{0\}

docs/src/AlgebraicGeometry/AlgebraicSets/ProjectiveAlgebraicSet.md
```diff
@@ -7,33 +7,43 @@ using Oscar
...
+
+$$\begin{aligned}X(K) &= \{ P \in \mathbb{P}^n(K) \mid f_1(P)=\dots = f_n(P)=0\}\\
+&=\{P \in \mathbb{P}^n(K) \mid \forall f\in I : f(P)=0\}.\end{aligned}$$
+
+Most properties of the projective variety $X$ refer to $X(K)$ where $K$ is an
+algebraically closed field.
+Just like for affine schemes there are a few exceptions to this rule,
+for instance whether $X$ is irreducible or not depends on its base field.
```
afkafkafk13: suggestion: 'for instance irreducibility of $X$ depends on its base field.'

docs/src/AlgebraicGeometry/AlgebraicSets/ProjectiveAlgebraicSet.md
```diff
@@ -7,33 +7,43 @@ using Oscar
...
+More formally we define a projective algebraic set as follows:
 ```@docs
 AbsProjectiveAlgebraicSet
 ```
 
 ## Constructors
-The recommended way to create an algebraic set is as a
-vanishing locus of a homogeneous ideal or polynomial.
+Projective algebraic set can be created from homogeneous polynomials and
+homogeneous ideals in standard graded rings.
```
afkafkafk13: Did you want to write 'A projective algebraic set' or 'Projective algebraic sets'? Both are ok.

docs/src/AlgebraicGeometry/AlgebraicVarieties/AffineVariety.md
```diff
@@ -7,22 +7,19 @@ using Oscar
...
-An affine variety over an algebraically closed field
-is an irreducible affine algebraic set. See [Affine Algebraic Sets](@ref).
+An affine variety is an algebraic set such that $X(K)$ is irreducible for $k \subseteq K$ an algebraic closure.
+See [Affine Algebraic Sets](@ref).
 
-In practice we work over non-closed fields. To be called a variety
-an algebraic set $V$ must stay irreducible when viewed over the algebraic closure.
-
-In Oscar varieties are [Affine schemes](@ref) and more formally defined as follows.
+In Oscar varieties are schemes as well [Affine schemes](@ref) and more formally defined as follows.
```
afkafkafk13: This phrase could be misleading, because it does not provide enough context for the (non-Oscar-expert) reader.
Maybe this might be the point to state a clear implementational remark, saying that:
- varieties are implemented as special instances of schemes
- functionality, which is not (yet) provided by a variety-specific implementation, falls back to appropriate functionality of schemes in a suitable way.

After that you could then go on with the constructors.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Attributes.jl
```diff
@@ -4,21 +4,91 @@
...
+
+function underlying_scheme(X::AffineAlgebraicSet{<:Field,<:MPolyQuoRing})
+  if isdefined(X, :Xred)
+    return X.Xred
+  end
+  # avoid constructing a morphism
+  I = ambient_closure_ideal(fat_scheme(X))
+  Irad = radical(I)
+  X.Xred = Spec(base_ring(Irad), Irad)
+  return X.Xred
```
afkafkafk13: Please, make sure to respect the is_reduced flag of the X -- X might already know that it is reduced, while the calling function doesn't have this information.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Attributes.jl
```diff
@@ -4,21 +4,91 @@
...
+#
+# (3) Further attributes
+#
+########################################################################
 
 @doc raw"""
     vanishing_ideal(X::AbsAffineAlgebraicSet) -> Ideal
 
-Return the ideal of all polynomials vanishing in ``X``.
+Return the radical ideal of all polynomials vanishing in ``X``.
```
afkafkafk13: Maybe state:
Return the ideal of all polynomials vanishing on ``X``, which is radical by Hilfert's Nullstellensatz.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Methods.jl
```diff
@@ -15,15 +17,24 @@ function Base.show(io::IO, X::AffineAlgebraicSet{<:Field,<:MPolyQuoRing})
...
   end
 end
 
 # special case for Zariski opens
 function Base.show(io::IO, ::MIME"text/plain", X::AffineAlgebraicSet)
-  println(io, "Affine algebraic set")  # at least one new line is needed
-  print(io, "  with coordinate ring $(OO(X))")
+  io = pretty(io)
+  println(io, "Reduced subscheme")  # at least one new line is needed
+  print(io, Indent(),"of ", Lowercase(), overlying_scheme(X))
```
afkafkafk13: fat_scheme instead of overlying_scheme?

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Methods.jl
```diff
@@ -15,15 +17,24 @@ function Base.show(io::IO, X::AffineAlgebraicSet{<:Field,<:MPolyQuoRing})
...
+  print(io, Dedent())
   # the last print statement must not add a new line
 end
 
+function Base.show(io::IO, X::AffineAlgebraicSet)
+  io = pretty(io)
+  println(io, "Reduced subscheme of ", Lowercase(), "overlying_scheme(X)")  # at least one new line is needed
```
afkafkafk13: fat_scheme instead of overlying_scheme?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -493,14 +493,18 @@ with coordinates
   I = modulus(OO(X))
   J = radical(pre_saturated_ideal(I))
   inc = ClosedEmbedding(X, ideal(OO(X), OO(X).(gens(J))))
-  return domain(inc), inc
+  Xred, inc = domain(inc), inc
+  set_attribute!(Xred, :is_reduced=>true)
+  return Xred, inc
   return Spec(base_ring(J), J, inverted_set(OO(X)))
 end
```
afkafkafk13: Please check here, whether is_reduced is set for X and bypass the radical in this case.
Do we need a ``check`` here?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -493,14 +493,18 @@ with coordinates
...
 
 @attr function reduced_scheme(X::AbsSpec{<:Field, <:MPolyQuoRing})
   J = radical(modulus(OO(X)))
   inc = ClosedEmbedding(X, ideal(OO(X), OO(X).(gens(J))))
-  return domain(inc), inc
+  Xred, inc = domain(inc), inc
+  set_attribute!(Xred, :is_reduced=>true)
+  return Xred, inc
   return Spec(base_ring(J), J)
 end
```
afkafkafk13: Again respect is_reduced. 
Again: Do wie need a ``check``?

src/AlgebraicGeometry/Schemes/AffineVariety/Objects/Properties.jl
```diff
@@ -1,3 +1,4 @@
 ################################################################################
-# Everything is inherited from AbsSpec
+# Avoid computing the reduced structure by using the overlying_scheme
```
afkafkafk13: shouldn't this be fat_scheme?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
```diff
@@ -3,84 +3,62 @@
...
-    Xred = X
-  end
-  ProjectiveAlgebraicSet(Xred, check=check)
+function algebraic_set(X::AbsProjectiveScheme; is_reduced::Bool=false, check::Bool=true)
+  return ProjectiveAlgebraicSet(X, is_reduced=is_reduced, check=check)
 end
 
 @doc raw"""
-    vanishing_locus(I::MPolyIdeal{MPolyDecRingElem}; check::Bool=true)
+    (I::MPolyIdeal{MPolyDecRingElem})
```
afkafkafk13: This line looks incomplete ?!?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Types.jl
```diff
@@ -1,15 +1,19 @@
...
+                                  check::Bool=true) where {S, T}
+    Y = new{S, T}()
+    Y.X = X
+    if is_reduced
+      @check is_geometrically_reduced(X) || error("algebraic sets must be geometrically reduced")
+      Y.Xred = X
     else
       # unlock the scheme methods for geometrically reduced schemes.
       set_attribute!(X, :is_geometrically_reduced, true)
       set_attribute!(X, :is_reduced, true)
```
afkafkafk13: These two lines are altering properties of X, which was the input !!! Is this a typo or is there something I am missing here?

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Constructors.jl
```diff
@@ -129,3 +129,7 @@ function reduced_scheme(X::AbsProjectiveScheme)
   set_attribute!(Xred, :is_reduced=>true)
   return Xred
 end
+
+function reduced_scheme(X::AbsProjectiveScheme{S,T}) where {S, T<:MPolyDecRing}
+  return X
+end
```
afkafkafk13, line 135: I do not see, where the radical computation is hidden here? E.g. the ideal < x^2 > is a perfectly fine ideal in a standard graded polynomial ring in 2 variables, but it is not radical.

By the way: respect is_reduced (also a few lines above in the other function reduced_scheme)

afkafkafk13 (commented):
Thank you for this large PR contributing algebraic sets and varieties.
I have added plenty of comments of varying severity (but nothing really bad in it -- ranging from nitpicks to minor changes).

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
afkafkafk13: This use of the term 'to forget' is in accordance with its use in things like the 'forgetful functor'. So from my point of view this is fine, as we are moving from the schemetheoretic world to the set theoretic world.

docs/oscar_references.bib
simonbrandhorst: This is how the Stacks-project wants to be cited. I took the bibtex entry from their homepage.

afkafkafk13: Then we have to change the bibtex test to allow this exception (@HereAround ?). Otherwise, we might as well remove it, because it will always be red.

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
simonbrandhorst: This sounds as if we do not like our own notation.

afkafkafk13: Then maybe
'to which we refer as fat scheme to distinguish it from the reduced scheme 'X_red', which is referred to as underlying_scheme.'

By the way, we do not like our own notation, but we do not have a better one.

docs/src/AlgebraicGeometry/AlgebraicSets/AffineAlgebraicSet.md
simonbrandhorst: The ambient space is not necessarily specified by the user. 
It is just implicitly derived from the polynomial ring where the variables live in.

docs/src/AlgebraicGeometry/AlgebraicSets/ProjectiveAlgebraicSet.md
simonbrandhorst: That is your style of writing. I prefer mine :-).

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
simonbrandhorst: where and for what?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Types.jl
simonbrandhorst: You are right, I should add these properties to `Y`.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Constructors.jl
simonbrandhorst: `X` is projective space by type dispatch. Nothing to compute here.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
afkafkafk13: We are entering reduced_scheme with a scheme X. 
If is_reduced happens to be set for X, we can avoid the whole computation of the radical. In this case, ``check`` could control whether we trust this and simply set ``Xred, inc = X, identity`` or whether we compute the radical anyway. 
(I am not talking about @check but about check)

afkafkafk13 (approved):
Thanks for the work and the quick changes.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
simonbrandhorst: Such an argument would be called `recompute`. But it would result in some trouble w.r.t. caching because it would create a new `reduced_scheme` which is `==` to the original one but not `===`.

fieker:
On Mon, Jun 05, 2023 at 01:17:32AM -0700, simonbrandhorst wrote:
> 
> 
> 
> >    I = modulus(OO(X))
>    J = radical(pre_saturated_ideal(I))
>    inc = ClosedEmbedding(X, ideal(OO(X), OO(X).(gens(J))))
> -  return domain(inc), inc
> +  Xred, inc = domain(inc), inc
> +  set_attribute!(Xred, :is_reduced=>true)
> +  return Xred, inc
>    return Spec(base_ring(J), J, inverted_set(OO(X)))
>  end
> 
> Such an argument would be called `recompute`. But it would result in some trouble w.r.t. caching.

To override caching and force a recpompute (very important for
developement and testing), we're using redo = true in a few places
already (mostly Hecke I guess)

docs/oscar_references.bib
simonbrandhorst: @HereAround @fingolfin  I ran bibtool, but it does not seem to like my citation-key.  How to proceed?

simonbrandhorst:
@wdecker this is the next iteration. Do you want to give it a look?

simonbrandhorst:
Unless there are further requests/comments, I will merge this on friday.

docs/oscar_references.bib
```diff
@@ -33,6 +33,13 @@ @Book{AL94
   url           = {https://doi.org/10.1090/gsm/003}
 }
 
+@Misc{Aut00,
+  author        = {{A}uthors, {T}he {S}tacks {P}roject},
+  title         = {{S}tacks {P}roject},
+  year          = {0000},
+  url           = {https://stacks.math.columbia.edu/}
```
fingolfin, line 40: My view on this is: this is just not a sensible bib entry. Can't you just put a URL link into the text as we do for e.g. the Singular homepage

fingolfin: (but there is no reason to delay merging this PR because of this)

docs/src/AlgebraicGeometry/intro.md
```diff
@@ -4,13 +4,24 @@ CurrentModule = Oscar
...
-We refer to the individual sections for more details.
+Computations in affine and projective algebraic geometry rely on [Commutative Algebra](@ref).
+
+
+Similarly, most algorithms for toric varieties and schemes are are based on
+[Polyhedral Geometry](@ref).
+
+General textbooks offering details on the theory of varieties and schemes include:
+- [Har77](@cite)
+- [Aut00](@cite)
```
fingolfin: To add to my comment above: I'd just change this to
```suggestion
- [The Stacks Project](https://stacks.math.columbia.edu)
```

simonbrandhorst: Thanks!

PR #2310, opened 2023-04-24
Prefer 'Return' over 'Returns' in docstrings

fingolfin:


experimental/GModule/GModule.jl
```diff
@@ -789,7 +789,7 @@ function Oscar.composition_factors_with_multiplicity(C::GModule{<:Any, <:Generic
 end
 
 """
-A decomposition of the module into indecomposable summands. Returns a list
+Return a decomposition of the module `C` into indecomposable summands as a list
```
YueRen, line 796: I don't think that mentioning `C` is necessary here, as the docstring does not specify names for the input.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -572,7 +572,7 @@ end
 @doc raw"""
     maximal_associated_points(I::IdealSheaf)
 
-Returns a `Vector` of `IdealSheaf`s corresponding to the non-embedded associated points of ``I`` on ``scheme(I)``.
+Return a `Vector` of `IdealSheaf`s corresponding to the non-embedded associated points of ``I`` on ``scheme(I)``.
```
YueRen, line 575: What is the difference between single backtick and double backtick? 

Or: shouldn't ``scheme(I)`` be in single backticks?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
```diff
@@ -707,7 +701,7 @@ end
 @doc raw"""
     is_geometrically_integral(X::AbsSpec)
 
-Return if ``X/k`` is geometrically integral.
+Test if ``X/k`` is geometrically integral.
```
YueRen, line 704: Maybe use "Check" instead of "Test" for consistency.

YueRen (approved):
Looks good!

lkastner:
We probably need to do this once in a while anyway.

lkastner (approved):

experimental/Schemes/IdealSheaves.jl
fingolfin: Odd number of backticks: code mode. Even number of backticks: math mode (unlike `$`, this then also works in non-raw strings). See https://docs.julialang.org/en/v1/stdlib/Markdown/#Literals

experimental/GModule/GModule.jl
fingolfin: Here's how I view it: This docstring is simply incomplete, it should start with the signature and then `C` would make sense. Without the `C`, the phrase "the module" doesn't make sense. One could rephrase as "the module given as an argument" -- but I think "the module `C`" is both shorter and clearer -- once the signature is added.

So I'll add that missing signature now

fingolfin: Of course most docstrings in that file are just rough sketches to start with -- the plan still is to clean them all up and migrate them to the "regular" codebase... (CC @ThomasBreuer)

experimental/Schemes/IdealSheaves.jl
fingolfin: So yeah maybe it should be `scheme(I)` with a single backtick, but I deliberately tried not to mix in such changes, too, because I'd be opening a bottom less pit 😂 . We'd first have to decide what the rules when to use what should be...

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Properties.jl
fingolfin: Perhaps... But note that we aren't consistent here to begin with. I also looked at a bunch of Julia docstrings for `is*` function, and they also used a mixture between "Check" and "Test". If someone would like to make a decision which to use, fine by me, but I feel like that would be work for a separate PR perhaps?

In the end, I think both are equally clear, so I wouldn't bother -- IMHO both are better than "Return `true` if XYZ holds, `false` otherwise" which we currently use in a lot of places and which feels rather stilted to me (full disclaimer: I've used it myself in the past)

fingolfin:
> We probably need to do this once in a while anyway.

Yeah, it's one of these things I try to remember to point out in PR reviews, to make people aware, but it's easy to overlook (and admittedly not the most important problem)

lkastner:
> > We probably need to do this once in a while anyway.
> 
> Yeah, it's one of these things I try to remember to point out in PR reviews, to make people aware, but it's easy to overlook (and admittedly not the most important problem)

I really think it depends. When it is a newbie I think it is perfectly fine only to point out the most important things and leave something like this for later.

PR #2314, opened 2023-04-24
Use codecovs `after_n_builds`

lgoettgens:
Currently, we have 3 ci jobs uploading coverage data to codecov (test-ubuntu-1.8, doctest-ubuntu-1.8, doctest-macos-1.8).
After the first one of these finishing, codecov adds the comment, and thus everyone participating in e.g. a PR will get an email. In the following minutes, after the other two jobs finish, the comment will get edited automatically (the email of course will not). This can lead to big drops in coverage being reported after only one of three jobs is finished (e.g. doctests will be fast, but have a lot less coverage).

This change would delay the codecov comment and notification next to the running ci jobs until `3` jobs have uploaded coverage data. This number has to be kept in sync with the ci.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,7 +1,8 @@
-coverage:
+codecov:
   notify:
     # Keep this in sync with the number of CI jobs uploading coverage.
     after_n_builds: 3
+coverage:
   status:
     project:
       default:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,7 @@
 coverage:
+  notify:
+    # Keep this in sync with the number of CI jobs uploading coverage.
+    after_n_builds: 3
   status:
     project:
       default:


benlorenz (approved):
I agree that this would be good but at least for this PR it did not work as expected.
The comment (+email) was after the first of the three jobs completed, maybe this only works once it is applied on master?

fingolfin:
Let's find out.

In the past there were other ways to achieve this: namely one could set up a CI job that runs after certain other jobs, and let that job notify Codecov that coverage collection is complete. I don't know if that is still possible, but if somebody would like to check...)

benlorenz:
Hm, there seems to be something wrong, the codecov page for the commit has:
```
Commit YAML is invalid

Coverage data is unable to be displayed, as the commit YAML appears to be invalid.
```
Here: https://app.codecov.io/github/oscar-system/Oscar.jl/commit/b25a64d9ca3e0e44c73792dd90e2099577845965

Edit: More precisely:
```
$ curl --data-binary @.codecov.yml https://codecov.io/validate
Error at ['coverage', 'notify', 'after_n_builds']: 
unknown field
```

lgoettgens:
Stupid me unable to copypaste the codecov docu :see_no_evil: 
Fix will come in a minute.

.codecov.yml
```diff
@@ -1,4 +1,7 @@
 coverage:
+  notify:
+    # Keep this in sync with the number of CI jobs uploading coverage.
+    after_n_builds: 3
```
benlorenz:
```suggestion
codecov:
  notify:
    # Keep this in sync with the number of CI jobs uploading coverage.
    after_n_builds: 3
coverage:
```
This seems to be valid and I think matches their docs.

lgoettgens: yep, thanks for fixing! Let's see if it succeeds now.

benlorenz:
No error here: https://app.codecov.io/github/oscar-system/Oscar.jl/commit/829c614d3b7f33bf737de5c0db75f23e35ca7469

And the first upload did not seem to trigger an edit for the comment, so that looks promising.

benlorenz:
Now it is stuck at processing because somehow the upload for one of the three required jobs failed:
```
[2023-04-24T19:38:05.026Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```
https://github.com/oscar-system/Oscar.jl/actions/runs/4789665118/jobs/8517806309#step:8:32

_Edit:_ I have restarted that job now ...

lgoettgens:
Now, just after the restarted job has finished, the comment got edited.

PR #2316, opened 2023-04-24
Prefer gen(A,i) over gens(A)[i]

fingolfin:
This is usually more efficient.

Also replace `parent(gen(I, 1))` by `base_ring(I)`,
and replace `length(gens(R))` by `ngens(R)`.


test/Rings/mpoly-graded-test.jl
```diff
@@ -107,7 +107,7 @@ end
         @test (RR in graded_rings) == Oscar.is_graded(RR)
         @test (RR in filtered_rings) == Oscar.is_filtered(RR)
         polys = _random_poly(RR, 4) # create 4 random polynomials
-        @test ngens(RR) == length(gens(RR))
+        @test ngens(RR) == ngens(RR)
```
afkafkafk13: For this one test: No, you do not want to replace it here, as it tests ngens agains length(gens).

fingolfin: Indeed, thanks!

afkafkafk13:
This replacement does not make sense in src/Rings, as we have the following setting there (example taken from  MPolyQuo.jl, starting line 7841, docstrings removed):
```
function gens(a::MPolyQuoIdeal)
  oscar_assure(a)
  return map(a.gens.Ox, a.gens.O)
end

gen(a::MPolyQuoIdeal, i::Int) = gens(a)[i]

function ngens(a::MPolyQuoIdeal)
  return length(gens(a))
end
```
So as long as gen(a,i) falls back to gens(a)[i] and ngens falls back to length(gens(a)) anyway, the change does not provide any advantage.

fieker:
On Mon, Apr 24, 2023 at 11:23:20PM -0700, Anne Frühbis-Krüger wrote:
> This replacement does not make sense in src/Rings, as we have the following setting there (example taken from  MPolyQuo.jl, starting line 7841):
> ```
> function gens(a::MPolyQuoIdeal)
>   oscar_assure(a)
>   return map(a.gens.Ox, a.gens.O)
> end
> 
> gen(a::MPolyQuoIdeal, i::Int) = gens(a)[i]
> 
> function ngens(a::MPolyQuoIdeal)
>   return length(gens(a))
> end
> ```
> So as long as gen(a,i) falls back to gens(a)[i] and ngens falls back to length(gens(a)) anyway, the change does not provide any advantage. 
We might write a sane ngens as well as a sane gen(a)[i]...

fingolfin:
@afkafkafk13 besides allowing for future improved versions of `gen` as @fieker hinted, if you look at this PR, you'll see that the methods you point out are actually replaced by better ones in this PR

afkafkafk13:
> @afkafkafk13 besides allowing for future improved versions of `gen` as @fieker hinted, if you look at this PR, you'll see that the methods you point out are actually replaced by better ones in this PR

ok. Yes, I missed that line in the diffs. I now understand the overall picture and agree that it makes sense.

fingolfin:
Still working on dealing with the fallout and adding missing methods... Also created https://github.com/oscar-system/Singular.jl/pull/636 (which adds missing methods uncovered by this PR here, though I am avoiding the problems with those missing methods here.

fingolfin:
This is ready for review now. I've added a bunch of missing `gen` methods, in a separate commit from the one which changes `gens(A)[i]` to `gen(A,i)`, so that it is a bit easier to review.

fieker (approved):

PR #2318, opened 2023-04-25
FTheoryTools: Rename Y4 to cy_hypersurface

HereAround:


apturner (approved):
Looks good! Thanks for making this change

HereAround:
The idea of this PR is to make the (internal) code self-explanatory. For this, I replaced the internal variable `Y4` by `cy_hypersurface`. As @lkastner suggested, we could also go all the way and rename to `calabi_yau_hypersurface`. Indeed, I have just renamed it like this, so that every ambiguity should now be removed - at least with regard to "what is the internal variable `Y4`?".

(For reasons of consistency, I have also adjusted the exported method that makes this internal variable available to the user. Previously, this method was called `cy_hypersurface`. This PR changes the name of this method to `calabi_yau_hyperface`. (I did not deprecate this method, as it clearly resides in the experimental section of OSCAR.))

lkastner (approved):

PR #2320, opened 2023-04-26
Use VarName in yet more places

fingolfin:


src/Rings/mpoly-graded.jl
```diff
@@ -2260,29 +2260,31 @@ julia> homogenization(I, "w", ordering = deglex(gens(base_ring(I))))
 ideal(x*z - y^2, -w*z + x*y, -w*y + x^2, -w*z^2 + y^3)
 ```
 """
-function homogenization(f::MPolyRingElem, var::String, pos::Int = 1)
+function homogenization(f::MPolyRingElem, var::VarName, pos::Int = 1)
   R = parent(f)
-  A = String.(symbols(R))
+  A = copy(symbols(R))
```
simonbrandhorst, line 2265: why the copy?

fingolfin: Because for many (most? all?) ring, `symbols` (and sometimes `gens`) returns the `Vector` stored inside the ring. And this code next does `push!(A, ...)`. Without making a copy, we'd end up modifying the symbols list of the ring.

This is of course a big gotcha. We'd really want to "write protect" these lists, or perhaps use a kind of `ImmutableVector`. Unfortunately we don't have these things :/.

simonbrandhorst: Thank you for the explanation.

simonbrandhorst (approved):

PR #2321, opened 2023-04-26
Literature models - first steps

HereAround:
This PR will introduce literature models for both Tate and Weierstrass F-theory models.

A couple of other (minor) cleanups and improvements will also be included.

~~(For my convenience, this currently includes the changes in https://github.com/oscar-system/Oscar.jl/pull/2323.)~~


lkastner (changes requested):
You are parsing some JSON manually. Are you aware that there are ongoing serialization efforts in OSCAR? Maybe you could talk to @antonydellavecchia . In any case I'd hope that it would be possible to implement it in some serialization form as described in https://docs.oscar-system.org/dev/DeveloperDocumentation/serialization/

Furthermore there is a `database.json`. Does that imply that there will be more of these files? If that is the case maybe we should think about separating database and source code from each other in some form.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,91 @@
...
+  e -> [-1 0]
+```
+"""
+function literature_tate_model(arxiv_id::String, equ_nr::String)
+  # Read all literature models
+  values = JSON.parsefile(joinpath(@__DIR__, "database.json"))
+  #return values
+  
+  # Find the entry that we are looking for
+  fitting_entries = filter(v -> (v["arxiv_id"] == arxiv_id && v["equ_nr"] == equ_nr), values)
```
lkastner: This is not how we should deal with data. Every time you access the database, you
- Read the entire database
- Search the entire database

It would already be much more sensible if there wasn't one big `database.json`, but instead many small `arxiv_id.json` files. What size will this database have?

experimental/FTheoryTools/src/LiteratureTateModels/database.json
```diff
@@ -0,0 +1,23 @@
...
+                    [["x", "y"], ["y", "s", "w"], ["s", "e4"], ["s", "e3"], ["s", "e1"], ["s", "w", "e3", "e1", "e2"]]]
+
+},
+{
+    "arxiv_id": "xxxxxxxx",
+    "equ_nr": "xxxxxx",
+    "description": "some other model",
+    "dim": 3
+}
+]
```
lkastner: This lacks almost all metadata necessary. Using the OSCAR serialization will already add a substantial amount.

HereAround:
> You are parsing some JSON manually. Are you aware that there are ongoing serialization efforts in OSCAR? Maybe you could talk to @antonydellavecchia . In any case I'd hope that it would be possible to implement it in some serialization form as described in https://docs.oscar-system.org/dev/DeveloperDocumentation/serialization/
> 
> Furthermore there is a `database.json`. Does that imply that there will be more of these files? If that is the case maybe we should think about separating database and source code from each other in some form.

Thank you @lkastner. I am aware of these efforts and looked a bit into it earlier today. I was hoping to discuss with you and @antonydellavecchia at some point (I must have mentioned this to you in slack a couple of days back? Or maybe I wanted to mention it and forgot about it?).

In any case, this PR is an (almost) "working version" and so hopefully a good starting point for discussion. So let me elaborate about the intention of this PR.

There are many FTheory constructions (in the form of singular elliptic fibrations) in the literature. It is very often the case, that in this field of research we eventually go back to such a construction and analyze it again - either with newly developed tools or a different focus. For some constructions this happened so very often, that they became famous (at least in the FTheory world). The intention of this PR is (a first step) to make such famous constructions easily available within OSCAR.

For each model/construction - besides technical details of the construction - I would envision the arxiv-id, equation number and possibly a description (some models have special names). This is necessary because some papers introduce several models. And one might want to look for the model based on its description/name rather than the arxiv-id.

As of a couple of minutes back, I was thinking of one big json-file that would have details of all such famous FTheory constructions. However, I think your suggestion @lkastner of having multiple json-files makes more sense. One advantage that I immediately see is that one could look for the correct file (or all that match certain criteria) and just read that file, instead of the big database. That might give a speed advantage.

As for the number of constructions: To get this effort of literature models started, I would estimate maybe a dozen. As we make more progress and cover more and more ground in the literature, this might eventually extend to a hundred or maybe even a few hundred models. (Let me ping @apturner in case he wants to correct my estimate.)

apturner:
Yes, I’d say 10s of models, with an upper bound in the low hundreds, is a safe estimate.

I do have one question about having a separate file for each model/paper. I definitely see the efficiency advantage in not reading a huge database for each query. However, the intention in also including a “description” or “name” field was to allow users to search for models by name, as Martin mentioned. One could also imagine filtering the list of models based on other features, such as their gauge algebra, and indeed this was a feature set I was imagining we would include at some point. But this seems to become impossible when splitting each model into a separate file, or at least you lose all of the efficiency bonus by having to query all the files. Is there a way to keep the efficiency improvement but also allow for such queries?

fieker:
On Thu, Apr 27, 2023 at 02:47:23PM -0700, Andrew P Turner wrote:
> Yes, I’d say 10s of models, with an upper bound in the low hundreds, is a safe estimate.
> 
> I do have one question about having a separate file for each model/paper. I definitely see the efficiency advantage in not reading a huge database for each query. However, the intention in also including a “description” or “name” field was to allow users to search for models by name, as Martin mentioned. One could also imagine filtering the list of models based on other features, such as their gauge algebra, and indeed this was a feature set I was imagining we would include at some point. But this seems to become impossible when splitting each model into a separate file, or at least you lose all of the efficiency bonus by having to query all the files. Is there a way to keep the efficiency improvement but also allow for such queries?
> 
An index: put all names ALSO into a seperate file and have a function
that can re-generate the index file...

experimental/FTheoryTools/src/LiteratureTateModels/database.json
antonydellavecchia: parts of the serialization will have to be extended, but anything mathematical should be using the serialization tools, and if they aren't there we should have a discussion on what is needed. Handling the meta data is also something we would like to implement

HereAround:
> On Thu, Apr 27, 2023 at 02:47:23PM -0700, Andrew P Turner wrote: Yes, I’d say 10s of models, with an upper bound in the low hundreds, is a safe estimate. I do have one question about having a separate file for each model/paper. I definitely see the efficiency advantage in not reading a huge database for each query. However, the intention in also including a “description” or “name” field was to allow users to search for models by name, as Martin mentioned. One could also imagine filtering the list of models based on other features, such as their gauge algebra, and indeed this was a feature set I was imagining we would include at some point. But this seems to become impossible when splitting each model into a separate file, or at least you lose all of the efficiency bonus by having to query all the files. Is there a way to keep the efficiency improvement but also allow for such queries?
> An index: put all names ALSO into a seperate file and have a function that can re-generate the index file...
> […](#)
> -- Reply to this email directly or view it on GitHub: [#2321 (comment)](https://github.com/oscar-system/Oscar.jl/pull/2321#issuecomment-1526551520) You are receiving this because you are subscribed to this thread. Message ID: ***@***.***>

I like that idea!

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
+@doc raw"""
+    literature_tate_model(; arxiv_id::String="", equ_nr::String="", description::String="")
+
+Many Tate models have been created in the F-theory literature.
+A significant number of them have even been given specific
+names, for instance the "U(1)-restricted SU(5)-GUT model".
+This method has access to a JSON-database, from which it can
```
fingolfin: "JSON database" is an implementation detail, I would not mention it in a docstring

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
...
+look up such literature models. Provided that the model is in
```
fingolfin: But how do the arguments to this function work? What should their content be, how can they be combined?

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
...
+  end
+  
+  # check if we found exactly one file, i.e. we were able to identify the model uniquely
+  @req length(candidate_files) == 1 "We could not uniquely identify the model"
+  
+  # we were able to find a unique model, so read that model
+  model = JSON.parsefile(joinpath(@__DIR__, "Models/" * candidate_files[1]))
+  
+  # Construct the model in question
+  auxiliary_base_ring, vars = PolynomialRing(QQ, string.(model["variables"]))
```
fingolfin: Why is the `string.` necessary, what does `model["variables"]` contain if not a list of strings?

Should this be using `cached=false`, @thofma @fieker ?

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
...
+  a1 = zero(auxiliary_base_ring)
+  if length(model["a1"]) > 0
+    a1 = prod([vars[p[1]]^p[2] for p in model["a1"]])
```
fingolfin, line 95: That looks like an expensive way to produce a polynomial?

As @lkastner already said, it seems you are reinventing polynomial serialization here instead of using what we already have. Is this intentional? If so it would be good to know what is missing in our existing serialization code so that it can ideally be improved to match your needs

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
...
+  end
+  t = global_tate_model([a1, a2, a3, a4, a6], auxiliary_base_ring, Int.(model["dim"]))
+  
+  # Read in the resolutions and process them to the correct format
+  resolutions = model["resolutions"];
+  resolutions = [[string.(x) for x in r] for r in resolutions]
+  set_attribute!(t, :resolutions => resolutions)
+  
+  # Remember saved attributes of this model, in particular the resolution sequences known
+  set_attribute!(t, :arxiv_id => model["arxiv_id"])
```
fingolfin: so it seems you only support data stored in the arXiv? That seems shortsighted (no offense intended)

experimental/FTheoryTools/src/LiteratureTateModels/index.json
```diff
@@ -0,0 +1,14 @@
+[
+{
+    "file": "model1.json",
+    "arxiv_id": "1109.3454",
```
fingolfin, line 5: this arxiv paper has two versions. I think the version must be stored here, too

experimental/FTheoryTools/src/TateModels/attributes.jl
```diff
@@ -279,3 +286,90 @@ julia> singular_loci(t)[2]
...
+
+#####################################################
+# 8: Attributes for literature Tate models
+#####################################################
+
+@doc raw"""
+    description(t::GlobalTateModel)
+
+Return the description of a global Tate model,
+should such a description have been provided.
```
fingolfin: also explain what happens uf this is not the case? It seems an exception is thrown, or perhaps an empty string returned? But then perhaps you also need a `has_description` method? Or else return `nothing` to indicate the absence of a description?

experimental/FTheoryTools/src/TateModels/attributes.jl
```diff
@@ -279,3 +286,90 @@ julia> singular_loci(t)[2]
...
+function description(t::GlobalTateModel)
+  @req has_attribute(t, :description) "No description known for this model"
+  return get_attribute(t, :description)
+end
+
+@doc raw"""
+    arxiv_id(t::GlobalTateModel)
+
+Return the arxiv identifier of a global Tate model,
+should such an identifier be known.
```
fingolfin: ditto

experimental/FTheoryTools/src/TateModels/methods.jl
```diff
@@ -58,4 +58,125 @@ function analyze_fibers(model::GlobalTateModel, centers::Vector{<:Vector{<:Integ
...
+    set_description(t::GlobalTateModel, description::String)
+
+Set a description for a global Tate model.
+
+```jldoctest
+julia> t = literature_tate_model(arxiv_id = "1109.3454", equ_nr = "3.5")
+Global Tate model over a not fully specified base -- SU(5)xU(1) restricted Tate model based on arxiv paper 1109.3454 (equ. 3.5)
+
+julia> set_description(t, "An SU(5)xU(1) GUT-model")
+
```
fingolfin, line 89:
```suggestion
```

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
...
+    
+  # read index file
+  file_index = JSON.parsefile(joinpath(@__DIR__, "index.json"))
+  
+  # create list of possible candidate files
+  candidate_files = Vector{String}()
+  for k in 1:length(file_index)
+    test = true
+    if arxiv_id != "" && file_index[k]["arxiv_id"] != arxiv_id
+      test = false
```
fingolfin: Can't you do away with the `test` variable and instead just `continue` here?
```suggestion
      continue
```

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
```diff
@@ -0,0 +1,111 @@
...
+    end
+    if equ_nr != "" && file_index[k]["equ_nr"] != equ_nr
+      test = false
+    end
+    if description != "" && file_index[k]["description"] != description
+      test = false
+    end
+    if test
+      push!(candidate_files, string(file_index[k]["file"]))
```
fingolfin: Most if not all the `string` calls in this PR seem redundant?

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
thofma: Yes, `cached = false` is recommended.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: I just removed that.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: I have just tried to elaborate more on this in the docstring. Currently, the expectation is that all information is provided a string and any combination of them is ok. If we cannot find a unique model with this data (i.e. in case there is none or multiple models that match the given data/criteria), then we raise an error.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: Ok. Thank you. Changed.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: I am pretty sure this can be improved. (When setting this up, I was pragmatic and "just" wanted it to work.) On Friday, I discussed with @lkastner and @antonydellavecchia on ways to improve. @lkastner offered to have a look into this and provide a serialization directly for objects of type `GlobalTateModels`.

One additional hurdle is that we (that is Andrew and I) will likely wish to add more information about these models in future. For instance, a lot is known about the topological data of many such models. I currently fail to provide a complete list for this additional information that could be added in the future. So we agreed to leave a bit of flexibility by adding a dictionary to which the new data can be added.

Once the `LiteratureModels` have been used for a while, we plan to revisit the matter in order to gauge if more improvements are possible.

Does that seem to make sense?

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: I believe we could in addition provide the `DOI`. Among others, we will then also be able to provide support for books. (Still, my expectation is that the majority of our literature models are rooted in publications that have been uploaded to the `arxiv`.)

What other (publication) sources did you think of @fingolfin ?

experimental/FTheoryTools/src/LiteratureTateModels/index.json
HereAround: Done.

experimental/FTheoryTools/src/TateModels/attributes.jl
HereAround: I have explained that an error is thrown if the information is not available. I have also added `has_*` methods.

experimental/FTheoryTools/src/TateModels/attributes.jl
HereAround: Fixed as mentioned above.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: I have indeed just added the `doi` as additional (potential) identifier of a publication.

experimental/FTheoryTools/src/TateModels/methods.jl
HereAround: Thank you. Done.

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: Yes, that is a good idea. Thank you!

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: As mentioned, I discussed with @lkastner and @antonydellavecchia how to properly execute the serialization for the objects of interest here. So this might eventually be redundant.

In any case: When I read in the data from the JSON file, then for instance the resolutions were of type `Vector{Vector{Vector{Any}}}`. So a cast from `Any` to `String` seemed reasonable. (But yes, I also noticed that the code would run without this cast, or at least without most of them.)

What do you think? Should I remove the casts?

experimental/FTheoryTools/src/LiteratureTateModels/constructors.jl
HereAround: Yes that is a great idea. In the beginning, we will have probably about 10 models. As time goes, Andrew and I anticipate an upper bound in the lower hundreds. So this seems to make perfect sense. A first (and with an eye towards serialization humble) attempt of this sorts has been added.

experimental/FTheoryTools/src/LiteratureTateModels/database.json
HereAround: Just to summarize my discussion last Friday with @lkastner and @antonydellavecchia - I fully agree.

This here, should (including with quite a few technical improvements as pointed out by @fingolfin) be a working version that we can improve.

@lkastner : As agreed last Friday, I will not make changes to this branch until I hear from you. From this point on, it is all yours. Please feel free to propose/push changes that marry this humble attempt with the existing serialization techniques in OSCAR.

HereAround:
As a consequence of the discussions that I had with @apturner over the last couple of days, we are currently extending the scope of `FTheoryTools` a lot. Therefore, the interface is subject to big changes. In particular, it is not (yet) absolutely clear what data we would ultimately need/want to serialize. Consequently, @lkastner agreed that we could postpone the proper serialization step and pass this PR (-> into experimental of course) as is. The issue https://github.com/oscar-system/Oscar.jl/issues/2362 should remind us that we need to come back to the serialization at a later stage.

I rebased this PR to the latest master. It is ready for review.

(The failures in nightly seem unrelated to this PR. Unless I am mistaken and understood correctly what @benlorenz mentioned on slack, then first step to the resolution of these failures happens in https://github.com/JuliaPackaging/Yggdrasil/pull/6724.)

experimental/FTheoryTools/src/TateModels/auxiliary.jl
lkastner, line 1: This is something we definitely need to address.

lkastner (approved):
As discussed with @HereAround , I will accept this without proper serialization for now. Two reasons:
- There are some features that @HereAround and @apturner need on master
- This is still in experimental, so we do not have to enforce everything at once.

I will assist in addressing the serialization later and also the `_eval_poly`. The whole `FTheoryTools` section will need another thorough review before moving to `src`. What is the schedule for that?

benlorenz:
> (The failures in nightly seem unrelated to this PR. Unless I am mistaken and understood correctly what @benlorenz mentioned on slack, then first step to the resolution of these failures happens in [JuliaPackaging/Yggdrasil#6724](https://github.com/JuliaPackaging/Yggdrasil/pull/6724).)

That is correct and all rebuilds (+ version bumps) are now complete. I have restarted the failed jobs.

HereAround:
> As discussed with @HereAround , I will accept this without proper serialization for now. Two reasons:
> 
>     * There are some features that @HereAround and @apturner need on master
> 
>     * This is still in experimental, so we do not have to enforce everything at once.
> 
> 
> I will assist in addressing the serialization later and also the `_eval_poly`. The whole `FTheoryTools` section will need another thorough review before moving to `src`. What is the schedule for that?

The first step is to get the toric schemes into `src`. This should hopefully happen in the upcoming weeks.

For `FTheoryTools` itself, I would anticipate that the final migration to source could happen towards the fall. It will simply take some time to achieve a stable interface (hopefully based on feedback of possible/future users). I hope this seems like a reasonable time scale.

~~I activate the automerge now, so that we can move forward. Thank you!~~

Actually, I am curious to see if the nightly tests succeed. So will wait until I see the result for that.

HereAround:
Cool. All tests did indeed pass. Nice!

PR #2322, opened 2023-04-26
Add coefficient_ring(I::MPolyIdeal)

### LINKED ISSUE #911: allow coefficient_ring for ideals
I suggest to allow `coefficient_ring` for ideals, unless there is a reason that I'm currently not seeing:

```
Kt, t = RationalFunctionField(QQ,"t")
Ktx, (x1,x2,x3,x4,x5) = PolynomialRing(Kt,5)
I = ideal([x1+x2,x2+x3,x3+x4,x4+x5])
coefficient_ring(I[1]) # works
coefficient_ring(I) # does not work
```

As far as I can see, `coefficient_ring` for ideals can be problematic for two reasons, both of which currently cannot occur in Oscar:

- the ideal has no generators
- the generators of the ideal belong to different polynomial rings

fingolfin:
Resolves #911 by @YueRen

### DIFF
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -1178,7 +1178,7 @@ julia> leading_coefficient(G[8])
 ```
 """
 function fglm(I::MPolyIdeal; start_ordering::MonomialOrdering = default_ordering(base_ring(I)), destination_ordering::MonomialOrdering)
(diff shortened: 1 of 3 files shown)


YueRen (approved):
Exactly what I asked for!

fingolfin:
Great!

(of course these days I hope you'd just open a PR yourself :-)

PR #2323, opened 2023-04-26
FTheoryTools: Minor changes

HereAround:
* Moved some code for Tate models to `methods.jl`.
* Adjusted white spaces to match `OSCAR` convention.

cc @apturner


apturner (approved):
This all looks good, @HereAround, thank you!

experimental/FTheoryTools/src/TateModels/methods.jl
```diff
@@ -0,0 +1,61 @@
+#####################################################
+# 1: Fiber analysis
+#####################################################
```
HechtiDerLachs, line 3: This has been moved from another file, right?

It seems that the code which has been moved is not tested. Any specific reason why this is the case? I.e. are the tests, for example, disabled because they run too long?

HechtiDerLachs (commented):
I'm afraid that the change of indentations in the docstrings might break the docs. We talked about it and @HereAround is checking it.

experimental/FTheoryTools/src/TateModels/methods.jl
HereAround: Yes, it has been moved from another file - correct.

Originally (i.e. before `FTheoryTools` became part of `OSCAR`), this has been tested. However, the test is really slow. So I disabled it a couple of weeks ago. Of course, this should be changed on the long run. Ideally, we (@apturner and myself) would find a better/quicker setup to test this function.

Thank you for reminding us. I would suggest to fix this in another PR.

HereAround:
> I'm afraid that the change of indentations in the docstrings might break the docs. We talked about it and @HereAround is checking it.

Thank you for the pointer. This indeed seems to have some undesired consequences. Let me fix this...

HereAround:
Regarding the unwanted changes in the documentation: @benlorenz, my vague recollection is that you enabled a preview on github that shows the documentation as changed in a PR? Correct? If yes, could you remind me how I can access this preview? Thank you!

HereAround:
> > I'm afraid that the change of indentations in the docstrings might break the docs. We talked about it and @HereAround is checking it.
> 
> Thank you for the pointer. This indeed seems to have some undesired consequences. Let me fix this...

I just pushed changes that should revert those unwanted indentations for all docstrings in `FTheoryTools`.

thofma:
> Regarding the unwanted changes in the documentation: @benlorenz, my vague recollection is that you enabled a preview on github that shows the documentation as changed in a PR? Correct? If yes, could you remind me how I can access this preview? Thank you!

This only works if the PR comes from a branch in the `oscar-system/Oscar.jl` repository (see https://github.com/oscar-system/Oscar.jl/actions/runs/4818992814/jobs/8581659128?pr=2323#step:6:428).

HechtiDerLachs (approved):
Once you have checked that the docs are intact, this should be good to go. 

It mostly moves code and adjusts indentations; no critical new functionality is added, so no deeper discussion seems necessary.

HereAround:
Yes, the documentation works as anticipated.

PR #2325, opened 2023-04-26
fixed `is_subgroup(H::T, G::T) where T <: GAPGroup`

ThomasBreuer:
such that the `domain` of the returned embedding is identical with `H`


fieker (approved):

ThomasBreuer:
The test failure in "doctest (1.8, macOS-latest)" looks strange. A segmentation fault happens in a situation that checks the proposed change, thus it is interesting. On the other hand, the same tests in "test (1.8, ubuntu-latest)" etc. pass without problem.

fingolfin (approved):

PR #2326, opened 2023-04-28
Make gens consistent for PrincipalOpenSubset.

### LINKED ISSUE #2324: Semantic conflict between `gens(U::PrincipalOpenSubset)` and `getindex(U::PrincipalOpenSubset, i::Int)`
Consider these methods:
```julia
gens(U::PrincipalOpenSubset) = [lifted_numerator(complement_equation(U))]
ngens(U::PrincipalOpenSubset) = 1
getindex(U::PrincipalOpenSubset, i::Int) = (i == 1 ? U : error("index out of range"))
```

Normally for most types, we'd expect `gens(U)[i] == U[i]`.

In fact we'd expect both to be equal to `gen(U, i)` which however not exist here. I wanted to add one, but the above contradiction doesn't allow it, esp. in light of https://github.com/Nemocas/AbstractAlgebra.jl/pull/1332 where we want to make the equality `gen(X,i) = X[i]` universal for all parent objects.

HechtiDerLachs:
I hope this fixes #2324. @fingolfin : Does it?


simonbrandhorst:
Tests fail.

src/AlgebraicGeometry/Schemes/PrincipalOpenSubset/Objects/Attributes.jl
```diff
@@ -18,7 +18,7 @@ complement_equation(U::PrincipalOpenSubset) = U.f::elem_type(OO(ambient_scheme(U
 ### assure compatibility with SpecOpen 
 gens(U::PrincipalOpenSubset) = [lifted_numerator(complement_equation(U))]
 ngens(U::PrincipalOpenSubset) = 1
-getindex(U::PrincipalOpenSubset, i::Int) = (i == 1 ? U : error("index out of range"))
+getindex(U::PrincipalOpenSubset, i::Int) = (i == 1 ? lifted_numerator(complement_equation(U)) : error("index out of range"))
```
fingolfin: This method is actually called in at least one test (that's how I discovered the discrepancy), so at the very least that test would have to be updated as well.

HechtiDerLachs:
Can this be merged?

afkafkafk13 (approved):
Yes, changing the inconsistently chosen name of gen for complement_equation is the right way to go. The PR makes sense to me and resolves this inconsistency.
@fingolfin : Since you brought up the point, please have a last look and merge, if you agree with my approval.

simonbrandhorst (approved):

src/AlgebraicGeometry/Schemes/SpecOpen/Objects/Attributes.jl
```diff
@@ -105,7 +104,10 @@ of ``U``. This function can also be called using the
 `getindex` method or simply via `U[i]`.
 """
 affine_patch(U::SpecOpen, i::Int) = affine_patches(U)[i]
+gens(U::SpecOpen) = affine_patches(U)
+gen(U::SpecOpen, i::Int) = affine_patches(U)[i]
 getindex(U::SpecOpen, i::Int) = affine_patches(U)[i]
+ngens(U::SpecOpen) = length(U.gens)
```
fingolfin: So is `length(U.gens)` always equal to `length(affine_patches(U))` ?

Looking above, I see that `npatches` is defined to also return `length(U.gens)`, so I'll assume this answers my question in the affirmative. But if so, I think it would be cleaner to do this:
```suggestion
ngens(U::SpecOpen) = npatches(U)
```

src/AlgebraicGeometry/Schemes/PrincipalOpenSubset/Objects/Attributes.jl
```diff
@@ -16,8 +16,11 @@ ambient_scheme(U::PrincipalOpenSubset) = U.X
 complement_equation(U::PrincipalOpenSubset) = U.f::elem_type(OO(ambient_scheme(U)))
 
 ### assure compatibility with SpecOpen 
-gens(U::PrincipalOpenSubset) = [lifted_numerator(complement_equation(U))]
+complement_equations(U::PrincipalOpenSubset) = [lifted_numerator(complement_equation(U))]
+number_of_complement_equations(U::PrincipalOpenSubset) = 1
 ngens(U::PrincipalOpenSubset) = 1
+gens(U::PrincipalOpenSubset) = [U]
+gen(U::PrincipalOpenSubset, i::Int) = (i == 1 ? U : error("index out of range"))
```
fingolfin, line 23: This is not an objection or change request, just a genuine question: why even define `gen`, `gens` and `ngens` here? Do you expect to write "generic" code which will accept `PrincipalOpenSubset` but also other types that don't define `complement_equations`, and so that generic code will somehow use `gens` ?

fingolfin (commented):
Fine by me to merge; I still left two comments about things that are unclear to me, but my original concern clearly is addressed

src/AlgebraicGeometry/Schemes/PrincipalOpenSubset/Objects/Attributes.jl
simonbrandhorst: As far as I remember we basically consider `PrincipalOpenSubset` and `SpecOpen` as covered schemes implicitly. 
One way to access the affine charts is via some index.
Aside: I wonder if we want to rename `SpecOpen` into `ZariskiOpen` after all it is not (even isomorphic to) the spectrum of a ring.

afkafkafk13: Shouldn't this be again an instance of complement_equation instead of gens? It seems to have changed in specOpen, but not in principalOpenSubset

HechtiDerLachs: The reason is that `SpecOpen` and `PrincipalOpenSubset` are treated in parallel in glueings. In some places I wanted to write generic code which wouldn't distinguish between those two cases. And that's why I thought this was a good idea: It makes the `PrincipalOpenSubset`s implement the `SpecOpen` interface.

HechtiDerLachs: The reason is that `SpecOpen` and `PrincipalOpenSubset` are treated in parallel in glueings. In some places I wanted to write generic code which wouldn't distinguish between those two cases. And that's why I thought this was a good idea: It makes the `PrincipalOpenSubset`s implement the `SpecOpen` interface.

Issue #2295, opened 2023-04-20
Make minimal_generating_set consistent


wdecker:
@HechtiDerLachs Please make sure that the function computing minimal generating sets for ideals in local rings fits to what comes out of PR #2291 (this applies in particular to the name of the function).

HechtiDerLachs:
The code in `src/Rings/mpoly-local.jl` is deprecated. Instead of adjusting it, we should discuss about finally removing that file so that not even more work is put into these 'old' local rings and their ideals. We can then eventually add the method with the new correct name for the respective new localizations.

wdecker:
I agree.

fingolfin:
@HechtiDerLachs what is the status of this?

afkafkafk13:
fixed with PR #2545

PR #2327, opened 2023-04-28
Implement minimal_generating_set for localized rings.

HechtiDerLachs:
This addresses #2295. 

I found it quite hard to understand how to handle the 'singular side' with all the `IdealGens` and nested `BiPolyArray`s floating around. I tried from reading the code in `MPolyIdeals.jl` and the like, but I can often only guess what there is to be done. 

One thing that confused me is that I don't seem to be able to call `Singular.mstd` as was done in #2291 . The only command I found was `Singular.minimal_generating_set`. Are the corresponding lines in #2291 tested? 

In either case it would be good if the people more involved with the internals of Groebner basis computations could have a look whether I did things right. My code seems to work, but I had no means to verify that what I do is correctly following the design ideas of `IdealGens`, `BiPolyArray`, for instance when it comes to caching of gbs and  preserving orderings and the like.

### DIFF
--- a/src/Rings/mpoly-localizations.jl
+++ b/src/Rings/mpoly-localizations.jl
@@ -2711,3 +2711,36 @@ end
   return dim(saturated_ideal(I))
 end
 
+@attr Vector{<:MPolyLocRingElem} function minimal_generating_set(
+    I::MPolyLocalizedIdeal{<:MPolyLocRing{<:Field, <:FieldElem, 
+                                          <:MPolyRing, <:MPolyElem, 
+                                          <:MPolyComplementOfKPointIdeal}
+                          }
+  )
+  L = base_ring(I)::MPolyLocRing # Type annotation for the reader's convenience.
+  R = base_ring(L)::MPolyRing
+
+  shift, back_shift = base_ring_shifts(L) # The shifting maps of R moving the localization point to 
+                                          # origin and back. 
+  I_shift = shifted_ideal(I)::MPolyIdeal # The pre_saturated_ideal of I in R after the shift 
+                                         # moving the localization point to the origin
+  o = negdegrevlex(gens(R)) # the default local ordering
+
+  # Three lines as suggested by @afkafkafk13:
+  singular_assure(I_shift, o) # Fills the singular side with a ring with the correct ordering.
+  I_shift_gb, I_shift_min = Singular.mstd(I_shift.gens.gens.S) # The duplication of gens seems awkward, but it is probably correct.
+                                                               # This returns a pair of singular ideals
+  I_min = L.(back_shift.(R.(gens(I_shift_min)))) # We convert the generators back to the oscar side, 
+                                                 # shift them, and consider them as elements in the 
+                                                 # localization.
+  return I_min
+# std_I_shift = standard_basis(I_shift, ordering=o)::IdealGens # Refers to a pre-computed standard basis if there is any.
+# bpa = std_I_shift.gens::BiPolyArray # Only the BiPolyArray has a "singular side".
+# l_sing = Singular.minimal_generating_set(bpa.S) # This should have the correct ordering set due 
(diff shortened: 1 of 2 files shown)


src/Rings/mpoly-localizations/methods.jl
```diff
@@ -0,0 +1,28 @@
+
```
HechtiDerLachs: Btw: I introduced this extra file since the recent regression of Revise makes it impossible to have type definitions in the same file you're working on. Eventually, `src/Rings/mpoly-localizations.jl` should be split into several files in this folder, but I didn't want to do that now as it would completely clutter this specific PR.

src/Rings/mpoly-localizations/methods.jl
```diff
@@ -0,0 +1,28 @@
...
+  shift, back_shift = base_ring_shifts(L) # The shifting maps of R moving the localization point to 
+                                          # origin and back. 
+  I_shift = shifted_ideal(I)::MPolyIdeal # The pre_saturated_ideal of I in R after the shift 
+                                         # moving the localization point to the origin
+  o = negdegrevlex(gens(R)) # the default local ordering
+
+  std_I_shift = standard_basis(I_shift, ordering=o)::IdealGens # Refers to a pre-computed standard basis if there is any.
+  bpa = std_I_shift.gens::BiPolyArray # Only the BiPolyArray has a "singular side".
+  l_sing = Singular.minimal_generating_set(bpa.S) # This should have the correct ordering set due 
+                                                  # to the construction history. TODO: Please confirm!
```
HechtiDerLachs: @jankoboehm @ederc

src/Rings/mpoly-localizations/methods.jl
fingolfin: Just pin Revise to 3.4.0 and all should be good?

thofma: I agree with @fingolfin that we should not change file structures because some other unrelated package is not working.

HechtiDerLachs: I see the point. But the original file is already much too big, anyway. So in principle this is a good occasion to start splitting it up. I just didn't want to do it here because one could then not see what this PR is really about. Should I change it back anyway? 

I'm still waiting for the feedback on my actual questions below, though.

src/Rings/mpoly-localizations/methods.jl
afkafkafk13: I am neither @jankoboehm nor @ederc , but from my limited understanding, we do not want to see an explicit BiPolyArray at this level of code. Paralleling the implementation of minimal_generating_set for graded rings, I would rather expect lines 17-22 to be something like
"""
  singular_assure(I_shift,o)
  I_shift_min = Singular.minimal_generating_set(I_shift.gens.gens.S)
  I_min = L.(back_shift.(I_shift_min))
"""
(of course fixing types whereever necessary....)

Please note that the side-effect of mstd computing a standard basis, which is exploited in the graded case, does not exist for minbase in the local case in Singular. So there might be no point in spending computation time on the standard basis here.

HechtiDerLachs:
I converted to draft for the time being since I was informed by @afkafkafk13 that some more functionality is in the making for `Singular.jl` that we wish to use here. We will update this PR once these are ready.

HechtiDerLachs:
It seems that the code suggested above by @afkafkafk13 does not run, yet. You can try with the new implementation. For me it gives a warning `// ** no minimal generating set computed` and then a list containing only one zero polynomial is returned as the second return value of `Singular.mstd`. 

@afkafkafk13 : I don't think you can avoid seeing a BiPolyArray here. Also in the code suggested by you it is visible, even though a bit hidden in a double `.gens`-statement. Every code that uses the singular side directly will have this. 

Please, ping me again when things are ready to be used.

afkafkafk13:
obsoleted by PR #2545

PR #2333, opened 2023-05-01
Replace GAP variable `_Vector_nf_elem` in Julia

fingolfin:
It turns out that using `Oscar.eval` in our GAP code like this causes problems when trying to use `Oscar` from within other Julia packages:

    [ Info: Precompiling test [e4ec5996-89d2-46bd-a9a8-6b9987a9712d]
    ERROR: LoadError: InitError: Error thrown by GAP: Error, Evaluation into
    the closed module `Oscar` breaks incremental compilation because the side
    effects will not be permanent. This is likely due to some other module
    mutating `Oscar` with `eval` during precompilation - don't do this.

### DIFF
--- a/gap/OscarInterface/gap/alnuth.gi
+++ b/gap/OscarInterface/gap/alnuth.gi
@@ -127,14 +127,12 @@ BindGlobal("NormCosetsDescriptionOscar", function(F, norm)
   return rec(units := units, creps := creps);
 end);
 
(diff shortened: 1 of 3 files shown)


ThomasBreuer (approved):
Yes,
perhaps it is the best solution to avoid `eval`.
(Should we then recommend in general to introduce a new Julia module in order to give GAP access to some Julia objects, instead of the attempt from oscar-system/GAP.jl/issues/862 ?)

PR #2302, opened 2023-04-21
Notions of transform (cleanup)

afkafkafk13:
missing tests/documentation for prematurely merged 'notions of transform' PR

PR #2340, opened 2023-05-04
Irreducible Decomposition for Cartier divisors

afkafkafk13:
follow-up to PR #2302 Notions of transform (cleanup)
- specialized decomposition algorithm for Cartier divisors
- remove method primary_decomposition for IdealSheaf 
  (incorrect terminology and old code, correct terms: assoicated_points;
   methods for associated_points available by #2302, specialized method for Cartier divisors in the current PR)
-  remove test of primary_decomposition for IdealSheaf


afkafkafk13:
@HechtiDerLachs: Is my use of trivializing_covering correct or should this read default_covering or even something completely different?

experimental/Schemes/CartierDivisor.jl
```diff
@@ -174,34 +174,60 @@ function cartier_divisor(IP::AbsProjectiveScheme, f::Union{MPolyDecRingElem, MPo
...
+### Decompostion of an effective Cartier Divisor into irreducible components
+### (specialized variant of associated_points, using pure codimension 1
+###  and taking multiplicities into account)
+@doc raw"""
+    irreducible_decomposition(C::EffectiveCartierDivisor)
+
+Return a `Vector` of pairs ``(I,k)`` corresponding to the irreducible components of ``C``. More precisely,  each ``I`` is a prime  `IdealSheaf` corresponding to an irreducible component of ``C`` and ``k``is the multiplicity of this component in ``C``.
+"""
+function irreducible_decomposition(C::EffectiveCartierDivisor)
+  cov = trivializing_covering(C)
```
HechtiDerLachs: The point of using `trivializing_covering` is that you get principal ideals on every `affine_patch` of such a covering. If you need that, this is the way to go. Otherwise, you should probably use `default_covering` since that has a lesser amount of charts.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -174,34 +174,60 @@ function cartier_divisor(IP::AbsProjectiveScheme, f::Union{MPolyDecRingElem, MPo
...
+
+  # run through all charts and collect further irreducible components
+  while length(charts_todo) > 0
+    U = pop!(charts_todo)
+    !is_one(I(U)) || continue                                ## supp(C) might not meet all charts
+    I_temp=I(U)
+
+    for (J,_) in associated_primes_temp
+      !is_one(J(U)) || continue
+      I_temp=saturation(I_temp,J(U))                         ## kick out known components
```
HechtiDerLachs, line 202: It seems that you are not using that any of these ideals is principal, are you? Then I suggest to use `default_covering` instead.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -174,34 +174,60 @@ function cartier_divisor(IP::AbsProjectiveScheme, f::Union{MPolyDecRingElem, MPo
...
+      !is_one(I_temp) || break                               ## break if nothing left
+    end
+
+    !is_one(I_temp) || break                                 ## break if nothing left
+    components_here = minimal_primes(I(U))
```
HechtiDerLachs: Wouldn't it suffice to take the minimal primes of `I_temp`? The other components are already taken care of, aren't they?

experimental/Schemes/CartierDivisor.jl
```diff
@@ -174,34 +174,60 @@ function cartier_divisor(IP::AbsProjectiveScheme, f::Union{MPolyDecRingElem, MPo
...
+    for comp in components_here
+      I_temp, saturation_index = saturation_with_index(I_temp, comp)
+      temp_dict=IdDict{AbsSpec,Ideal}()
+      temp_dict[U] = comp
+      I_sheaf_temp = IdealSheaf(X, extend!(cov, temp_dict), check=false)
```
HechtiDerLachs, line 212: Is this line tested? Frankly, I do not understand what's going on. It seems you're extending a given covering by a dictionary. I don't recall having implemented such functionality and I can't really think of what this should mean. Could you explain what you intend to do here?

test/AlgebraicGeometry/Schemes/IdealSheaves.jl
```diff
@@ -111,29 +111,3 @@ end
   @test pbII(U[2]) isa Ideal
   @test haskey(pbII.I.obj_cache, U[2])
 end
-
-@testset "primary decomposition of ideal sheaves" begin
```
HechtiDerLachs, line 115: Can we rewrite these tests to use the new methods?

HechtiDerLachs (commented):
It seems that you're using the fact that for prime ideals an `IdealSheaf` can be constructed by specifying it on one dense affine patch `U` on which it is non-trivial. Then one can ask for the unique extension of that ideal sheaf to the other charts. Is it correct that you're using this functionality? 

Either way, I'm not sure that the call to the constructor of the ideal sheaves is correct; see my comments below. It seems we need some test code for all this. 

Regarding the question about `trivializing_covering` vs. `default_covering`: I remember from some oral discussions that you wanted to exploit the ideal for the cartier divisor being principal. But I don't find this anywhere in the code. Am I missing something, or did you postpone implementing these more specialized features?

test/AlgebraicGeometry/Schemes/IdealSheaves.jl
afkafkafk13: The new methods are tested in the Cartier to Weil divisor tests, as I replaced the primary decomposition in weil_divisor.

experimental/Schemes/CartierDivisor.jl
afkafkafk13: Yes, this is a tradeoff, which I am not sure about: saturation w.r.t. principal ideals is cheaper and i tried to give this the best chances to stay principal. But the price is a higher number of charts.

experimental/Schemes/CartierDivisor.jl
afkafkafk13: Only for the cheaper price of the saturation.

experimental/Schemes/CartierDivisor.jl
afkafkafk13: Indeed. This is a simple typo (no changed back after some testing)

experimental/Schemes/CartierDivisor.jl
afkafkafk13: We are extending the dictionary to the covering, not the covering by a dictionary. (we are dealing with irreducible codimension 1 subschemes, so there is no guesswork going on here).
This is functionality which you provide in IdealSheaves.jl and which is documented and tested (not by me..)

test/AlgebraicGeometry/Schemes/IdealSheaves.jl
afkafkafk13: These tests do not make sense for the new method, as they decompose IdealSheaves which are not Cartier divisors. However, the preceding PR #2302 already contained such tests using assoicated points and maximal_associated_points in another testfile (transforms.jl), as these methods were parts of that PR.

experimental/Schemes/CartierDivisor.jl
afkafkafk13: But you are right: 
I could have written 
            IdealSheaf(X,U,gens(comp)) 
but would not have had the option to swith off the tests there (which are true by construction here). I have borrowed this line from your code there.

experimental/Schemes/CartierDivisor.jl
HechtiDerLachs: Wasn't it that having a principal ideal only pays of in the second argument? Then this seems to almost never be the case here because `J(U)` will usually be constructed by passing from another chart via a saturation process. I'm afraid that the use of the `trivializing_covering` then does not really give any speedup.

experimental/Schemes/CartierDivisor.jl
HechtiDerLachs: Oh, je... It was me then? At least it seems to have been documented...

test/AlgebraicGeometry/Schemes/IdealSheaves.jl
HechtiDerLachs: If similar tests take place elsewhere, then I think removing these is fine.

experimental/Schemes/CartierDivisor.jl
afkafkafk13: That was the heart of my question: does trivializing_covering make sense. I agree with your argument for no. So I will change this.

HechtiDerLachs: I would even take `default_covering` of the scheme. I'm not sure, but `default_covering` of the divisor `C` might also refer to a refinement on which `C` was first defined.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -174,34 +174,60 @@ function cartier_divisor(IP::AbsProjectiveScheme, f::Union{MPolyDecRingElem, MPo
...
+Return a `Vector` of pairs ``(I,k)`` corresponding to the irreducible components of ``C``. More precisely,  each ``I`` is a prime  `IdealSheaf` corresponding to an irreducible component of ``C`` and ``k``is the multiplicity of this component in ``C``.
+"""
+function irreducible_decomposition(C::EffectiveCartierDivisor)
+  X = scheme(C)
+  cov = default_covering(X)
+  OOX = OO(X)
+
+  charts_todo = copy(patches(cov))
+  I = ideal_sheaf(C)
+  associated_primes_temp = Vector{Tuple{IdealSheaf, Int}}()  ## already identified components
```
simonbrandhorst: `IdealSheaf` is still an abstract type. If you want a fast vector you need  
```suggestion
  associated_primes_temp = Vector{Tuple{typeof(I), Int}}()  ## already identified components
```

simonbrandhorst (approved):

experimental/Schemes/CartierDivisor.jl
afkafkafk13: done

PR #2347, opened 2023-05-05
More on F-TheoryTools and ToricSchemes

HereAround:
This is a first stab towards providing support for much more general F-theory constructions. Sadly, much more work is needed (and I am working on this right now). However, I hope that this is an acceptable first step.

One major change is to introduce 5 types for various F-theory constructions.
* `GlobalWeierstrassModel` and `GlobalTateModel` remain as before.
* `GlobalHypersurfaceModel`: In the making, but requires a separate PR - otherwise too messy. Also, I am still working on this.
* `CompleteIntersectionModel` and `ClosedSubschemeModel` are "place-holders" for something that we may or may not introduce at a later stage. For `ClosedSubschemeModel`, it is even in theory not entirely clear to me what data needs to be provided.

Another major change was to support `ToricSchemes` as base spaces for `GlobalWeierstrassModel` and `GlobalTateModel`. In anticipation of future extensions, I also allow an (arbitrary) covered scheme. But no constructors are provided - among others this would require the ability to compute sections of the (anti)canonical bundle in such a setting, which I believe is beyond our current abilities.

Other than that - some more or less random clean-up of things that had to be addressed earlier or later anyways. This includes changes in `ToricSchemes` for `snake_case` constructor as well as providing support for (weighted) projective space, del Pezzo surface and Hirzebruch surface. The latter also required some minor changes in the toric varieties. Similar to the agreement for `projective_space`, the constructors for del Pezzo surfaces and Hirzebruch surfaces now accept a type as first argument. The toric methods were deprecated accordingly.

@apturner @lkastner @HechtiDerLachs : Unless you see major obstacles/drawbacks/problems with this PR, I would truly appreciate if we could merge this PR more or less as is, so that I (and @apturner) can proceed to code something meaningful for the `hypersurface_models` tomorrow and, once complete, are able to return to the literature models in https://github.com/oscar-system/Oscar.jl/pull/2321. Thank you!

(In retrospect, I wonder if it would have been better to merge the literature models first as is, despite the fact that - of course - we are not really using the true serialization techniques in `OSCAR`. At the time, I did not even think of a larger class of models to be supported. This came up during discussions with @apturner this week. The longer I look at this and let these discussions sink in, the more doubtful I am that `FTheoryTools` will the sufficiently stable by the end of the upcoming week for a full-fledged serialization approach to work. So it might have been worth/good to merge the `LiteratureModel` PR as is. Either way - by now one PR needs a massive rebase anyways...)


experimental/FTheoryTools/src/types.jl
```diff
@@ -0,0 +1,107 @@
...
+  end
+end
+
+
+@attributes mutable struct GlobalTateModel
+  tate_a1::MPolyRingElem{QQFieldElem}
+  tate_a2::MPolyRingElem{QQFieldElem}
+  tate_a3::MPolyRingElem{QQFieldElem}
+  tate_a4::MPolyRingElem{QQFieldElem}
+  tate_a6::MPolyRingElem{QQFieldElem}
```
lkastner, line 69: Why not something like
```suggestion
  tate_ai::Vector{MPolyRingElem{QQFieldElem}}
```
?

Then the accessor could also be something like
```
tate_section_ai(bla, i::Int)
```
So you would only need one instead of 5.

lkastner: You could probably also use a `Dict{Symbol, MPolyRingElem{QQFieldElem}}` and then assign these to the symbols `:ai`.

experimental/FTheoryTools/src/types.jl
```diff
@@ -0,0 +1,107 @@
...
+  tate_polynomial::MPolyRingElem{QQFieldElem}
+  base_space::ToricOrNonToricCoveredScheme
+  ambient_space::ToricOrNonToricCoveredScheme
+  fiber_ambient_space::ToricOrNonToricCoveredScheme
```
lkastner: These three appear in both `GlobalTateModel` and `GlobalWeierstrassModel`. Could this be hinting at another object that lives beneath?

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -169,27 +179,27 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
     locus = ring_map(gens(id)[1])
     
     if f_ord == 0 && g_ord == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(9 * poly_g, [locus], [0]), evaluate(2 * poly_f, [locus], [0]))
+      monodromy_poly = _psi^2 + divexact(evaluate(9 * weierstrass_g, [locus], [0]), evaluate(2 * weierstrass_f, [locus], [0]))
       if _count_factors(monodromy_poly) == 2
         kod_type = "Split I_$d_ord"
       else
         kod_type = "Non-split I_$d_ord"
       end
```
lkastner: 1

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -169,27 +179,27 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
     elseif d_ord == 4 && g_ord == 2 && f_ord >= 2
-      monodromy_poly = _psi^2 - evaluate(divexact(poly_g, locus^2), [locus], [0])
+      monodromy_poly = _psi^2 - evaluate(divexact(weierstrass_g, locus^2), [locus], [0])
       if _count_factors(monodromy_poly) == 2
         kod_type = "Split IV"
       else
         kod_type = "Non-split IV"
       end
```
lkastner: 2

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -199,21 +209,21 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
         kod_type = "Non-split I^*_0"
       end
     elseif f_ord == 2 && g_ord == 3 && d_ord >= 7 && d_ord % 2 == 1
-      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * poly_f, locus^2)^3, [locus], [0]), 4 * evaluate(divexact(9 * poly_g, locus^3), [locus], [0])^3)
+      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * weierstrass_f, locus^2)^3, [locus], [0]), 4 * evaluate(divexact(9 * weierstrass_g, locus^3), [locus], [0])^3)
       if _count_factors(monodromy_poly) == 2
         kod_type = "Split I^*_$(d_ord - 6)"
       else
         kod_type = "Non-split I^*_$(d_ord - 6)"
       end
```
lkastner: 3

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -199,21 +209,21 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
     elseif f_ord == 2 && g_ord == 3 && d_ord >= 8 && d_ord % 2 == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * poly_f, locus^2)^2, [locus], [0]), evaluate(divexact(9 * poly_g, locus^3), [locus], [0])^2)
+      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * weierstrass_f, locus^2)^2, [locus], [0]), evaluate(divexact(9 * weierstrass_g, locus^3), [locus], [0])^2)
       if _count_factors(monodromy_poly) == 2
         kod_type = "Split I^*_$(d_ord - 6)"
       else
         kod_type = "Non-split I^*_$(d_ord - 6)"
       end
```
lkastner: 4 appearances of almost identical code. This multiplication should be eliminated by adding a small helper function. Maybe it can even swallow the other less similar blocks. If there is less code, it is easier to maintain.

lkastner (changes requested):
I'd like to have a look at this later again.

I don't like the usage of
```
ToricOrNonToricCoveredScheme = Union{ToricCoveredScheme{BRT}, CoveredScheme{BRT}} where {BRT<:Ring}
```
(I guess these come from somewhere else, but they don't show up on master). This seems to be done to avoid duplication, but on the other hand current usage obscures the underlying schemes.

Questions:
- Are these always of the same type in e.g. `GlobalWeierstrassModel`?
- What is the reason against using something like
    ```
    struct GlobalWeierstrassModel{B,A,F} where {B<:TONT, A<:TONT, F<:TONT}
    ```
  This would allow for easier specialization for the different cases. If these three were always of the same type, then everything becomes even easier, since you only need one parameter. Additionally you could even have the model depend on the ring.

experimental/FTheoryTools/docs/src/tate.md
```diff
@@ -58,109 +60,171 @@ following table---often referred to as the Tate table and taken from
...
+To circumvent this very demanding computation, our toric constructors operate in the
+opposite direction. That is, they begin by extracting the rays and maximal cones of the chosen
+toric base space. Subsequently, those rays and cones are extended to form one of the many
+toric ambient spaces. This proves hugely superior in performance than going through the
+triangulation task of enumerating all possible toric ambient spaces. One downside of this strategy
+is that the so-constructed ambient space need not be smooth.
+
+### A toric variety as base space
+
+We require that the provided toric base space is complete. Oftentimes, this is an expensive check.
```
HechtiDerLachs: But at least in the "universal" or "unspecified" case this does not hold, does it? Geometrically, we only had affine space in 5-6 variables there, if I remember correctly. Either way, providing the possibility to disable this check seems useful. I wouldn't know how to do this for an `AbsCoveredScheme` which does not come from a concrete projective variety.

experimental/FTheoryTools/docs/src/tate.md
```diff
@@ -58,109 +60,171 @@ following table---often referred to as the Tate table and taken from
...
 * ``a_1 = a_{10} w^0``,
 * ``a_2 = a_{21} w^1``,
 * ``a_3 = a_{32} w^2``,
 * ``a_4 = a_{43} w^3``,
 * ``a_6 = a_{65} w^5``,
-then the polynomial ring in question is the ring with indeterminates
+In this case, it is useful to consider the polynomial ring with indeterminates
 ``a_{10}``, ``a_{21}``, ``a_{32}``, ``a_{43}``, ``a_{65}`` and ``w``.
+In theory, one can consider these indeterminates as local coordinate of an
+auxiliary base space. Indeed, for our computer implementation the polynomial
```
HechtiDerLachs, line 142: Doesn't this rather make it a "local tate model" in some sense? 

Nothing to block this from being merged, but something I would like to discuss at some point.

experimental/FTheoryTools/docs/src/weierstrass.md
```diff
@@ -46,33 +46,61 @@ This can be read-off from the Weierstrass table, which we have reproduced from
...
+spaces. Namely, in order to construct a Weierstrass model as a hypersurface in an ambient
+space, we first wish to construct the ambient space in question. For a toric base, one way to
+achieve this is to first focus on the Cox ring of the toric ambient space. This ring must be
+graded such that the Weierstrass polynomial is homogeneous and cuts out a Calabi-Yau
+hypersurface. Given this grading, one can perform a triangulation task. Typically, this
+combinatorial task is very demanding, consumes a lot of computational power and takes a
+long time to complete. Even more, it will yield a large, often huge, number of candidate
+ambient spaces of which the typical user will only pick one. For instance, a common and
+often appropriate choice is a toric ambient space which contains the toric base space in a
+manifest way.
```
HechtiDerLachs, line 69: I think this is at least partially a duplicate of the text above, isn't it?

experimental/FTheoryTools/docs/src/weierstrass.md
```diff
@@ -82,69 +110,84 @@ Weierstrass model in question is well-defined.
...
+used in the desired factorization of the Weierstrass sections ``f`` and ``g``.
+In theory, one can consider the indeterminates of this polynomial ring as local
+coordinate of an auxiliary base space. Indeed, for our computer implementation
+the polynomial ring with these indeterminates serve as the coordinate ring of
+an auxiliary *toric* base space. Despite this auxiliary base space being toric,
+the predictions from such an analysis are not limited to the world of toric varieties.
+
+For constructions along these lines, we support the following constructor:
+```@docs
+global_weierstrass_model(weierstrass_f::MPolyRingElem{QQFieldElem}, weierstrass_g::MPolyRingElem{QQFieldElem}, auxiliary_base_ring::MPolyRing, d::Int)
```
HechtiDerLachs, line 128: Did you test this? From what I remember, there is a special type for polynomials over `QQ` and it should rather be something like `QQMPolyElem` rather than `MPolyRingElem{QQFieldElem}`. But I'm not 100% sure. Maybe the concrete type is also a subtype of this and it is fine again...

experimental/FTheoryTools/src/TateModels/attributes.jl
```diff
@@ -93,56 +100,79 @@ Global Tate model over a not fully specified base
...
+Global Tate model over a not fully specified base
 
-julia> toric_base_space(t)
-Normal toric variety without torusfactor
+julia> base_space(t)
+Scheme of a toric variety with fan spanned by RayVector{QQFieldElem}[[1, 0, 0, 0, 0, 0], [0, 0, 0, 0, 1, 0], [0, 0, 0, 0, 0, 1], [0, 1, 0, 0, 0, 0], [0, 0, 1, 0, 0, 0], [0, 0, 0, 1, 0, 0]]
 ```
 """
-@attr AbstractNormalToricVariety function toric_base_space(t::GlobalTateModel)
+@attr ToricOrNonToricCoveredScheme function base_space(t::GlobalTateModel)
```
HechtiDerLachs: Maybe it would make sense to even split this up? Can we identify a type parameter in `GlobalTateModel` which allows us to distinguish the two cases?

experimental/FTheoryTools/src/TateModels/attributes.jl
```diff
@@ -213,10 +244,12 @@ julia> t = su5_tate_model_over_arbitrary_3d_base()
 Global Tate model over a not fully specified base
 
 julia> discriminant(t);
-
 ```
 """
-@attr MPolyRingElem{QQFieldElem} discriminant(t::GlobalTateModel) = discriminant(global_weierstrass_model(t))
+@attr MPolyRingElem{QQFieldElem} function discriminant(t::GlobalTateModel)
+  @req typeof(base_space(t)) === ToricCoveredScheme{QQField} "Discriminant of global Tate model is currently only supported for toric varieties/schemes as base space"
```
HechtiDerLachs, line 250: Also here it would probably make sense to distinguish those cases in which a discriminant polynomial makes sense already from the signature. If possible.

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -14,6 +14,8 @@ end
 # 2: Construct ambient space from given base
 ################################################################
 
+_ambient_space_from_base(base::ToricCoveredScheme) = _ambient_space_from_base(underlying_toric_variety(base))
+
 function _ambient_space_from_base(base::AbstractNormalToricVariety)
```
HechtiDerLachs, line 19: Does the return value of this function also provide information about the embedding of the original variety? If not, can we get that information through other channels?

experimental/FTheoryTools/src/types.jl
```diff
@@ -0,0 +1,107 @@
+################################################
+# 1: The Julia types for FTheoryTools
+################################################
+
+abstract type AbstractFTheoryModel end
+
+ToricOrNonToricCoveredScheme = Union{ToricCoveredScheme{BRT}, CoveredScheme{BRT}} where {BRT<:Ring}
+
+@attributes mutable struct ClosedSubschemeModel <: AbstractFTheoryModel
```
HechtiDerLachs, line 7: Here, and probably already for the abstract supertypes, I would introduce type parameters which already allow for the identification of the type of, say, the base space, etc. But, since time is pressing this week, we can also do that later.

experimental/FTheoryTools/src/types.jl
```diff
@@ -0,0 +1,107 @@
...
+    return new(tate_a1, tate_a2, tate_a3, tate_a4, tate_a6, tate_polynomial, base_space, ambient_space, weighted_projective_space(ToricCoveredScheme, [2,3,1]))
+  end
+end
+
+
+################################################
+# 2: conceptual parents
+################################################
+
+@attr ClosedSubschemeModel function conceptual_parent(cim::CompleteIntersectionModel)
```
HechtiDerLachs, line 91: This looks interesting. But I do not understand it straight away; I would like to have these functions explained to me later.

src/Exports.jl
```diff
@@ -168,13 +168,11 @@ export SubQuoHom
 export SubdivisionOfPoints
 export SubquoModule
 export SubquoModuleElem
-export ToricCoveredScheme
```
HechtiDerLachs, line 171: Are these exports taken elsewhere?

HechtiDerLachs (approved):
This is a first step towards a very much appreciated widening of the F-Theorytools framework to also accomodate Weierstrass models beyond the toric world. The details are probably not completely spelled out and there will be more changes necessary as we progress. But this is also something that we need to learn and discuss as we progress and this looks like a good step in the right direction.

experimental/FTheoryTools/src/types.jl
apturner: Thanks for this suggestion, we will go with the vector option and create an accessor as you suggested.

HereAround: Actually, we do not have `a5` but would prefer the `vector`. So we need a vector that has no 5-th entry sort of...

Is most likely not ultra-critical. Please allow us to ignore this for now.

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Just pushed a commit with this. Thank you for the suggestion @lkastner.

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Done (see below).

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Done (see below).

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Done (see below).

HereAround:
> 

@apturner, @HechtiDerLachs and I just discussed this and decided that it is most beneficial to remove the union type `ToricorNonToricCoveredScheme` and replaced it by the most general type `AbsCoveredScheme`. In the future, this might even allow for schemes with group action as base (or ambient) space. This change has been done in https://github.com/oscar-system/Oscar.jl/pull/2347/commits/2772f643468c1cb3e8c83b24dabdd98e3e3a165f.

Does that address your requested changes @lkastner ?

experimental/FTheoryTools/docs/src/tate.md
HereAround: Thank you. I need to specify that this is - as of now - a technical limitation. In the toric setting, we cannot find a finite set of sections if there are infinitely many sections (as in affine cases).

And yes, for the auxiliary/not specified base cases, there is not such check. In fact, most of these auxiliary bases will the affine.

experimental/FTheoryTools/docs/src/tate.md
HereAround: As discussed, there are local and global aspects. We leave it as is for now.

experimental/FTheoryTools/docs/src/weierstrass.md
HereAround: That is true, the documentation for Tate and Weierstrass models is very similar (and in parts, identical).

experimental/FTheoryTools/docs/src/weierstrass.md
HereAround: I will try to look into this later. Not super critical (I hope).

src/Exports.jl
HereAround: They are taken (back) into experimental.

experimental/FTheoryTools/src/types.jl
HereAround: Just explained and discussed in person. More to come... maybe.

experimental/FTheoryTools/src/types.jl
HereAround: Yes, sounds good. We will look into this in more detail. For now, we have just removed the `ToricOrNonToricScheme`, but I am pretty sure more functionality and then advantages for parametrized types will be seen in the future.

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Implicitly, you can read-off the embedding from comparing the `cox_ring` of `base` and `ambient_space`, but this is not (at least not yet) very explicit. We can consider a (toric) morphism for this at some later stage.

experimental/FTheoryTools/src/TateModels/attributes.jl
HereAround: As of now, we cannot compute the discriminant for technical reasons once we are outside of the toric regime.

experimental/FTheoryTools/src/TateModels/attributes.jl
HereAround: Oh... this should not be the `@attr`. This needs fixing. Thank you. Will do.

HereAround:
> > 
> 
> @apturner, @HechtiDerLachs and I just discussed this and decided that it is most beneficial to remove the union type `ToricorNonToricCoveredScheme` and replaced it by the most general type `AbsCoveredScheme`. In the future, this might even allow for schemes with group action as base (or ambient) space. This change has been done in [2772f64](https://github.com/oscar-system/Oscar.jl/commit/2772f643468c1cb3e8c83b24dabdd98e3e3a165f).
> 
> Does that address your requested changes @lkastner ?

For more changes in this regard, I have raised an issue: https://github.com/oscar-system/Oscar.jl/issues/2356. Hopefully, we can address this in larger detail in the future.

lkastner (approved):

PR #2351, opened 2023-05-06
Add missing `deepcopy_internal` methods for module elems

### LINKED ISSUE #2152: `deepcopy` for module elems
`deepcopy` behaves differently for different OSCAR types. For polynomial rings, it creates an identical polynomial in the same parent. For `FreeMod` (and I think other module types, but I am not sure about that), it copies the parent as well, resulting in a mathematically identical element in a mathematically identical parent.

```julia
julia> R, (x, y, z) = PolynomialRing(QQ, ["x", "y", "z"])
(Multivariate Polynomial Ring in x, y, z over Rational Field, QQMPolyRingElem[x, y, z])

julia> x == deepcopy(x)
true

julia> F = free_module(R, 2)
Free module of rank 2 over Multivariate Polynomial Ring in x, y, z over Rational Field

julia> t = gen(F, 1)
e[1]

julia> t == deepcopy(t)
false
```

I would love to have consistent behavior on this for most/all of our types.

lgoettgens:
Resolves #2152.

Furthermore fixed some whitespace in the testset that has been updated.

### DIFF
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -1098,7 +1098,7 @@ end
 ###############################################################################
 
 @doc raw"""
-    FreeMod_dec(R::CRing_dec, n::Int,VarName = :e; cached::Bool = false) 
+    FreeMod_dec(R::CRing_dec, n::Int, name::VarName = :e; cached::Bool = false) 
 
 Construct a decorated (graded or filtered) free module over the ring `R` with rank `n`
 with the standard degrees, that is the standard unit vectors have degree 0.
@@ -1106,12 +1106,12 @@ Additionally one can provide names for the generators. If one does
 not provide names for the generators, the standard names e_i are used for 
 the standard unit vectors.
 """
-function FreeMod_dec(R::CRing_dec, n::Int,VarName = :e; cached::Bool = false) 
+function FreeMod_dec(R::CRing_dec, n::Int, name::VarName = :e; cached::Bool = false) 
   return FreeMod_dec{elem_type(R)}(R, [Symbol("$name[$i]") for i=1:n], [decoration(R)[0] for i=1:n])
 end
 
 @doc raw"""
-    free_module_dec(R::CRing_dec, n::Int,VarName = :e; cached::Bool = false)
+    free_module_dec(R::CRing_dec, n::Int, name::VarName = :e; cached::Bool = false)
 
 Create the decorated free module $R^n$ equipped with its basis of standard unit vectors
 and standard degrees, that is the standard unit vectors have degree 0.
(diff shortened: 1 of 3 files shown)


src/Modules/UngradedModules.jl
```diff
@@ -491,6 +491,10 @@ function hash(a::AbstractFreeModElem, h::UInt)
   return hash(tuple(parent(a), coordinates(a)), h)
 end
 
+function Base.deepcopy_internal(a::AbstractFreeModElem, dict::IdDict)
```
thofma, line 494: Is this the concrete type or a supertype?

lgoettgens: This is the abstract supertype of FreeModElem and FreeModElem_dec. But since arithmetics are implemented similar for the abstract type, I see no problem in that.

thofma: It seems like the function will always return a `FreeModElem` and never ` FreeModElem_dec`? Or am I missing something.

lgoettgens: There is the method https://github.com/oscar-system/Oscar.jl/blob/8f222b014ad54decbba1d581716d9ca99becf5d7/src/Modules/ModulesGraded.jl#L1325 detecting that the parent is decorated and then creating a decorated element.
This seems to be some kind of ugly hack, but all arithmetics, etc., for decorated free modules work exactly the same.

thofma: Thanks for the explanation. I did not know we had something this unconventional.

thofma (approved):

src/Modules/ModulesGraded.jl
```diff
@@ -1098,20 +1098,20 @@ end
...
 
 Construct a decorated (graded or filtered) free module over the ring `R` with rank `n`
 with the standard degrees, that is the standard unit vectors have degree 0.
 Additionally one can provide names for the generators. If one does 
 not provide names for the generators, the standard names e_i are used for 
 the standard unit vectors.
 """
-function FreeMod_dec(R::CRing_dec, n::Int,VarName = :e; cached::Bool = false) 
+function FreeMod_dec(R::CRing_dec, n::Int, name::VarName = :e; cached::Bool = false) 
   return FreeMod_dec{elem_type(R)}(R, [Symbol("$name[$i]") for i=1:n], [decoration(R)[0] for i=1:n])
```
fingolfin, line 1110: Huh, wait, how did this code even work before, when `name` was not defined? Is it simply not exercised? Or does it perhaps accidentally reference the function `name` here when it does `$name`? Really weird...

lgoettgens: I thought the same, somehow this was syntactic correct julia code 😅
The point I spotted it was some weird printing when fiddling around with deepcopy.

PR #2357, opened 2023-05-08
(Even) more changes in FTheoryTools

HereAround:
* Remove restrictions to rationals whenever possible,
* Raise warning whenever variables names are duplicated among base and fiber. (One can still proceed in this case if really desired.)


HechtiDerLachs (approved):
Just like we discussed. Looks good.

apturner (commented):
Looks good, thank you @HereAround!

PR #2358, opened 2023-05-08
small generating set

afkafkafk13:
- make mstd-functionality available for multivariate polynomial rings and quotients thereof
  (naming: small generating set, because we do not have minimality here!)
- make small_generating_set available for localizations at powers of elements
- make small_generating_set available on charts for IdealSheaves 
- use small_generating_set in example K3.jl


afkafkafk13:
@HechtiDerLachs , @simonbrandhorst : This should do, what you wanted to achieve by calling radical in K3.jl (even for ideals known to already be radical). Please have a look.

simonbrandhorst:
Since this terminology in principle makes sense also for ideals which are not multivariate, I cc. @fieker

afkafkafk13:
> Since this terminology in principle makes sense also for ideals which are not multivariate, I cc. @fieker

From that perspective, I would suggest:
- small_generating_set also exists for graded rings as an alias for minimal_generating_set (to avoid confusion for users -- each of the two is available whenever it makes sense)  -- this will happen in this PR, if you agree
- dito for the local case, as soon as it is finished/merged -- on my plate, but not here, waiting for PR#2327
- I see univariate polynomial rings in a different context compared to multivariate ones (more precisely, they are euclidean rings), so it is up to @fieker to decide whether he also provides the two function names as aliases there -- independent of the current PR

simonbrandhorst:
What I mean to say is that `small_generating_set` makes sense for almost any algebraic structure that has generators.
Groups, ideals, modules, algebras,  algebras, non-commutative algebras, vector spaces etc.
So I want to make sure we are reasonably consistent here.
The common definition of a `small_generating_set` would be: 1. It is a generating set. 2. It is "fast" to compute 3. It is "small", but in general we do not give any guarantees (except maybe that it should be at least as small as the input?) 
The second question: 
`minimal_generating_set` minimal with respect to inclusion or cardinality? I think for finite groups it is cardinality.

afkafkafk13:
> What I mean to say is that `small_generating_set` makes sense for almost any algebraic structure that has generators. Groups, ideals, modules, algebras, algebras, non-commutative algebras, vector spaces etc. So I want to make sure we are reasonably consistent here. The common definition of a `small_generating_set` would be: 1. It is a generating set. 2. It is "fast" to compute 3. It is "small", but in general we do not give any guarantees (except maybe that it should be at least as small as the input?) 

Indeed, this is precisely the idea -- including your statement about at least as small as the input!

> The second question: `minimal_generating_set` minimal with respect to inclusion or cardinality? I think for finite groups it is cardinality.

Minimal is meant as in minimal free resolution in the local and graded case, where we have uniqueness of Betti numbers. 
So it has a standard mathematical definition here, which coincides with cardinality (1st Betti number of R/I).

afkafkafk13:
Aside: tests are in doctests, not in the test-directory. So they might not show up correctly in the codecov (in particular for the quotient ring case).

experimental/Schemes/IdealSheaves.jl
```diff
@@ -793,6 +793,17 @@ end
...
+function small_generating_set(II::IdealSheaf)
+    X = scheme(II)
+  # If there is a simplified covering, do the calculations there.
+  covering = (has_attribute(X, :simplified_covering) ? simplified_covering(X) : default_covering(X))
+  ID = IdDict{AbsSpec, Ideal}()
+  for U in patches(covering)
+    ID[U] = ideal(base_ring(II(U)),Oscar.small_generating_set(saturated_ideal(II(U))))
+  end
+  return(IdealSheaf(X, ID, check = false))
+end
```
simonbrandhorst: Please fix the indentation. Also the `Oscar.` should not be necessary.

src/Rings/MPolyQuo.jl
```diff
@@ -1586,6 +1586,63 @@ function minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})
...
+
+```
+"""
+function small_generating_set(I::MPolyQuoIdeal)
+  # For non-homogeneous ideals, we do not have a notion of minimal generating
+  # set, but Singular.mstd still provides a good heuristic to find a small
+  # generating set.
+  Q = base_ring(I)
+
+  @req coefficient_ring(Q) isa AbstractAlgebra.Field "The coefficient ring must be a field"
```
simonbrandhorst:
```suggestion
  @req coefficient_ring(Q) isa Field "The coefficient ring must be a field"
```

src/Rings/MPolyQuo.jl
```diff
@@ -1586,6 +1586,63 @@ function minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})
...
+  return_value = filter(!iszero, (Q).(gens(sing_min)))
+  if length(return_value) <= ngens(I)
+    return return_value
+  else
+    return gens(I)
+  end
+end
+
+# in the graded case, reusing a cached gb makes sense, so use minimal_generating set there
+small_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})=minimal_generating_set(I)
```
simonbrandhorst:
```suggestion
small_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem}) = minimal_generating_set(I)
```

simonbrandhorst (approved):

experimental/Schemes/IdealSheaves.jl
afkafkafk13: done.

src/Rings/mpoly-ideals.jl
```diff
@@ -1383,6 +1383,81 @@ function minimal_generating_set(I::MPolyIdeal{<:MPolyDecRingElem})
...
+ x*y^2 - z
+```
+"""
+function small_generating_set(I::MPolyIdeal)
+  # For non-homogeneous ideals, we do not have a notion of minimal generating
+  # set, but Singular.mstd still provides a good heuristic to find a small
+  # generating set.
+
+  R = base_ring(I)
+  @req coefficient_ring(R) isa AbstractAlgebra.Field "The coefficient ring must be a field"
```
simonbrandhorst:
```suggestion
  @req coefficient_ring(R) isa Field "The coefficient ring must be a field"
```

Issue #1964, opened 2023-02-22
Add CI tests to our projects looking for "bad" exports


fingolfin:
We sometimes accidentally export non-existing things. E.g. from current Oscar.jl:
```
julia> filter(n->!isdefined(Oscar, n), names(Oscar))
32-element Vector{Symbol}:
 :AbsPrePreSheaf
 :CoherentSheaf
 :DirectProductOfElem
 :LineBundle
 :_non_degeneration_cover
 :affine_patch_type
 :ambient
 :as_smooth_lci_of_cod
 :as_smooth_local_complete_intersection
 :base_scheme_type
 :controlled_transform
 :coset_decomposition
 :covered_patches
 :covering
 :dualsubdivision
 :generator
 :has_complement_class_reps
 :ideal_dict
 :ideal_sheaf_type
 :images_of_variables
 :intersect_stably
 :load_simplicialcomplex
 :normalize
 :p
 :prepare_smooth_center
 :save_simplicialcomplex
 :set_complement_class_reps
 :set_is_finite
 :subdivision_of_vertices
 :total_transform
 :weak_transform
 :weight
```

The `:p` one is due to a typo, @HereAround is already fixing it.

I suggest the following:
- [ ] resolve these (either by adding missing functions or removing the exports)
- [ ] add something like this to our tests:
  ```julia
  @test all(n->isdefined(Oscar, n), names(Oscar))
  ```
- add similar tests to our other packages (AA, Nemo, Hecke, GAP, Singular, Polymake, ...)

HereAround:
For completeness, the fix for `export p` is included in the PR https://github.com/oscar-system/Oscar.jl/pull/1958.

thofma:
We could just use https://github.com/JuliaTesting/Aqua.jl (I used it successfully locally in the past).

afkafkafk13:
I am working on 
 :total_transform
 :weak_transform
 :controlled_transform
right now.

lkastner:
I like @thofma s suggestion very much.

lgoettgens:
> We could just use https://github.com/JuliaTesting/Aqua.jl (I used it successfully locally in the past).

I added a first try to use it in #2370 now that the wrong exports have been removed (thanks, @lkastner!).

PR #2359, opened 2023-05-09
Misc refactorings

lkastner:
- Move includes out of `Oscar.jl` into subfolders into a more package like structure whenever possible
- Remove `-test` from filenames, since these are already in the test folder
- Extract some parts of `Oscar.jl` into `src/utils.jl` (docs, versioninfo)
- Remove unused exports, related to #1964 . I have decided to deal with this issue in an independent PR, as not to block the stuff here from arriving at master.


src/Oscar.jl
```diff
@@ -470,82 +470,19 @@ include("exports.jl")
...
-include("GAP/customize.jl")
-include("GAP/gap_to_oscar.jl")
-include("GAP/oscar_to_gap.jl")
-include("GAP/iso_gap_oscar.jl")
-include("GAP/iso_oscar_gap.jl")
+
+include("Groups/Groups.jl")
+
+include("Rings/Rings.jl")
+include("GAP/GAP.jl")
```
lgoettgens, line 253: Many files in `Groups/` depend on the `GAPWrap` module in `GAP/wrappers.jl`. Could you try to reorder the includes accordingly?

test/runtests.jl
```diff
@@ -16,7 +16,7 @@ end
 Oscar.set_seed!(seed)
 
 import Oscar.Nemo.AbstractAlgebra
-include(joinpath(pathof(AbstractAlgebra), "..", "..", "test", "Rings-conformance-tests.jl"))
+include(joinpath(pathof(AbstractAlgebra), "..", "..", "test", "Rings-conformances.jl"))
```
lgoettgens, line 19: This is a file in `AbstractAlgebra`.

src/Oscar.jl
lkastner: It seems fine though?

src/utils.jl
```diff
@@ -0,0 +1,181 @@
+# When a specific branch is loaded via `]add Package#branch` julia will only
```
lgoettgens, line 8: Maybe add even more utility stuff from `Oscar.jl` like `test_module`, etc.? What do you think?

lkastner: Done.

src/Oscar.jl
lgoettgens: I reread the julia docs. As long as you only have functions, the order of inclusion is irrelevant. Only for types it matters. Since `GAP/wrappers.jl` defines no types, this is fine as it is. Sorry for the confusion.

benlorenz (approved):
Thank you very much for doing this!

src/Oscar.jl
```diff
@@ -470,82 +246,19 @@ include("exports.jl")
...
-
-include("GAP/customize.jl")
-include("GAP/gap_to_oscar.jl")
-include("GAP/oscar_to_gap.jl")
-include("GAP/iso_gap_oscar.jl")
-include("GAP/iso_oscar_gap.jl")
+
+include("Groups/Groups.jl")
+
+include("Rings/Rings.jl")
```
fingolfin, line 252: Could we include this before `Groups.jl`, and then move this:
```
include("Groups/group_characters.jl")  # needs some Rings functionality
include("Groups/action.jl")  # needs some PolynomialRings functionality
```
also into `Groups/Groups.jl`?

Or is there a dependency in the `Rings.jl` code on code in `Groups.jl` that prevents it (if so, perhaps a comment could be added explaining that)

src/utils.jl
```diff
@@ -0,0 +1,248 @@
...
+##  versioninfo
+##
+###############################################################################
+###############################################################################
+# When a specific branch is loaded via `]add Package#branch` julia will only
+# create a checkout and keep a bare git repo in a separate directory.
+# In a bare repo HEAD will not point to the correct commit so we use the git
+# tree-hash that Pkg.jl provides and manually map this to a corresponding
+# commit.
+function _lookup_commit_from_cache(url::AbstractString, tree::AbstractString)
```
fingolfin, line 13: I think also the const globals `jll_deps` and `cornerstones` should be moved to `utils.jl`.

(And perhaps it would make sense to split this into three files, say `utils/versioninfo.jl`, `utils/docs.jl`, `utils/tests.jl` or so -- but I am fine either way, just an idea)

src/Oscar.jl
```diff
@@ -70,94 +70,7 @@ end
 # force trigger recompile when folder changes
 include_dependency("../experimental")
```
fingolfin, line 71: How about moving the code from line 31 to here, which deals with `experimental`, over to `../experimental/Experimental.jl`?

src/utils.jl
fingolfin: Another possible addition to the hypothetical `utils` dir might be `utils/rand.jl` with the `get_seeded_rng` function and friends, also from `Oscar.jl`.

fingolfin:
Hmm, seems I was too late. Ah well.

fingolfin:
Also
```julia
const aadir = Base.pkgdir(AbstractAlgebra)
const nemodir = Base.pkgdir(Nemo)
const heckedir = Base.pkgdir(Hecke)
```
should go to `utils.jl` (or `utils/docs.jl` or whatever)

PR #2366, opened 2023-05-10
Basic constructors for "standard finite fields" as described by Lübeck.

mjrodgers:
This code constructs finite fields in a standardized way that avoids the usage of Conway polynomials.

The ideas used are described by Lübeck in [[1]](#1) and some of the code is based on Lübeck's StandardFF GAP package.

<a id="1">[1]</a> Lübeck, F. Standard generators of finite fields and their cyclic subgroups, Journal of Symbolic Computation (2023)([arXiv](https://arxiv.org/pdf/2107.02257.pdf)).


experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+end
+# just return degrees
+function standard_monomial_degrees(n::IntegerUnion)
+    if n == 1
+        return [1]
+    end
+    # need the largest prime factor a of n
+    nfactorization = factor(ZZ(n))
+    nfactors = sort([r for (r,e) in nfactorization])
+    a = Int(nfactors[end])
```
fingolfin: You are allocating a vector, sorting it, just to compute the maximum. Better is to just ask for the maximum. I haven't tested it but perhaps something like this:
```suggestion
    a = maximum(x -> Int(x[1]), nfactorization)
```
Or alternatively even this:
```suggestion
    a = maximum(nfactorization)
```
at which point `a` is a pair, with `a[1]` resp. `a.first` taking the place of the old `a`, and `a[2]` resp. `a.second` replacing `nfactorization[a]`.

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+    return L
+
+end
+
+
+# TODO: this should work also if p is an integer
+function standard_finite_field(p::T, n::IntegerUnion) where T<:IntegerUnion
+    if !isprime(p)
+        error()
+    end
```
fingolfin: How about
```suggestion
    @req isprime(p) "first argument must be a prime"
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+    L, X = FiniteField(polynomial(F, poly), vname)
+    set_standard_finite_field!(L)
+    set_primitive_powers_in_tower_basis!(L, pmat)
+
+    return L
+
+end
+
+
+# TODO: this should work also if p is an integer
```
fingolfin: I don't understand this comment, `p` is always an integer below, isn't it? Or do you perhaps mean a specific type perhaps?

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+function standard_finite_field(p::T, n::IntegerUnion) where T<:IntegerUnion
```
fingolfin:
```suggestion
function standard_finite_field(p::IntegerUnion, n::IntegerUnion)
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+    if !isprime(p)
+        error()
+    end
+    F = GF(p)
+    set_standard_prime_field!(F)
+    get_standard_extension!(F, n) do
+      nfactorization = factor(ZZ(n));
+      nfactors = sort([r for (r,e) in nfactorization]);
+      lastfactor = nfactors[end]
```
fingolfin:
```suggestion
      lastfactor = maximum(x -> x[1], nfactorization)
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+
+# TODO: this should work also if p is an integer
+function standard_finite_field(p::T, n::IntegerUnion) where T<:IntegerUnion
+    if !isprime(p)
+        error()
+    end
+    F = GF(p)
+    set_standard_prime_field!(F)
+    get_standard_extension!(F, n) do
+      nfactorization = factor(ZZ(n));
```
fingolfin:
```suggestion
      nfactorization = factor(ZZ(n))
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+      nfactors = sort([r for (r,e) in nfactorization]);
+      lastfactor = nfactors[end]
+      nK = div(n,lastfactor)
+      K = standard_finite_field(p, nK)
+
+      stn = steinitz_number_for_prime_degree(p, Int(lastfactor), nfactorization[lastfactor])
+      n1 = big(lastfactor)^(nfactorization[lastfactor]-1)
+      # BUG this overflows if p is an Int64
```
fingolfin: Is this a BUG (if so, in how far?), or rather a *warning*?

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+# TODO : Should be fixed in Hecke for prime fields
+function coords(x::FinFieldElem)
+    return absolute_coordinates(x)
+end
+function coords(x::PrimeFieldElem)
+    return [x]
+end
+
+# TODO : this should be pushed to Nemo.jl/src/flint/gfp_fmpz_mat.jl
+# BUG the fact that this is missing is a major performance problem
```
fingolfin: Then ideally please open a PR there ASAP so that we'll have it soon :-)

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+  z = similar(a)
+  r = ccall((:fmpz_mod_mat_inv, libflint), Int,
+          (Ref{FpMatrix}, Ref{FpMatrix}), z, a)
+  !Bool(r) && error("Matrix not invertible")
+  return z
+end
+
+# TODO : this just makes sense for writing general code
+import Base: big
+big(a::ZZRingElem) = a
```
fingolfin: I'd rather we didn't do that. We don't want to sometimes use `BigInt` and sometimes `ZZRingElem`, do we? So then better to write `ZZ(a)` instead of `big(a)`, I'd say? Or what am I missing?

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+
+
+# _attributes = [
+    # :is_standard_finite_field      - bool
+    # :is_standard_prime_field       - bool
+    # :primitive_powers_in_tower_basis - Matrix{FinFieldElem}
+    # :tower_basis                  - Matrix{FFieldElem} = inv(:primitive_powers_in_tower_basis)
+    # NOTE I guess we used ZZ ring elements for these instead of Int; does it matter?
```
fingolfin: Operations with `Int` are of course usually MUCH faster than with `ZZRingElem` (or `BigInt`). But they can overflow. So yeah, it'll matter,  for performance and/or for correctness... Am I misunderstanding the question?

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+end
+function set_steinitz_prime_degree!(F::FinField, r::IntegerUnion, k::IntegerUnion, nr::IntegerUnion)
+    spd = get_attribute!(F, :steinitz_prime_degree, Dict{ZZRingElem, Dict{ZZRingElem, ZZRingElem}}())
+    spdr = get!(spd, r, Dict{ZZRingElem, ZZRingElem})
+    spdr[k] = nr
+end
+function set_steinitz_prime_degree!(f::Function, F::FinField, r::IntegerUnion, k::IntegerUnion)
+    spd = get_attribute!(F, :steinitz_prime_degree, Dict{ZZRingElem, Dict{ZZRingElem, ZZRingElem}}())
+    spdr = get!(spd, r, Dict{ZZRingElem, ZZRingElem})
+    spd[r][k] = f()
```
fingolfin:
```suggestion
    spdr[k] = f()
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+function set_primitive_powers_in_tower_basis!(F::FinField, M::PrimeFieldMatrix)
+    set_attribute!(F, :primitive_powers_in_tower_basis, M)
+    set_attribute!(F, :tower_basis, inv(M))
+end
+function set_steinitz_prime_degree!(F::FinField, r::IntegerUnion, k::IntegerUnion, nr::IntegerUnion)
+    spd = get_attribute!(F, :steinitz_prime_degree, Dict{ZZRingElem, Dict{ZZRingElem, ZZRingElem}}())
+    spdr = get!(spd, r, Dict{ZZRingElem, ZZRingElem})
+    spdr[k] = nr
+end
+function set_steinitz_prime_degree!(f::Function, F::FinField, r::IntegerUnion, k::IntegerUnion)
```
fingolfin: Hmm, it seems neither of the two `set_steinitz_prime_degree!` functions are used?

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
+using Oscar
+
+IntegerUnion   = Union{Integer, ZZRingElem}
+PrimeField     = Union{Nemo.fpField, Nemo.FpField}
+PrimeFieldElem = Union{fpFieldElem, FpFieldElem}
+PrimeFieldMatrix = Union{FpMatrix, fpMatrix}
```
fingolfin:
```suggestion
const IntegerUnion     = Oscar.IntegerUnion
const PrimeField       = Union{Nemo.fpField, Nemo.FpField}
const PrimeFieldElem   = Union{fpFieldElem, FpFieldElem}
const PrimeFieldMatrix = Union{FpMatrix, fpMatrix}
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+function get_steinitz_prime_degree!(f::Function, F::FinField, r::IntegerUnion, k::IntegerUnion)
+    spd = get_attribute!(F, :steinitz_prime_degree, Dict{ZZRingElem, Dict{ZZRingElem, ZZRingElem}}() )
+    spdr = get!(spd, r, Dict{ZZRingElem, ZZRingElem}())
+    get!(spdr, k, f())
+end
+function get_standard_extensions(F::PrimeField)
+  get_attribute(F, :standard_extensions, nothing)
+end
+function get_standard_extensions!(F::PrimeField)
+  get_attribute(F, :standard_extensions, Dict{ZZRingElem, FinField}())
```
fingolfin: Note that this always allocate a `Dict{ZZRingElem, FinField}`, even if you don't need it because the `:standard_extensions` already exists. Hence it is better to do this:
```suggestion
  get_attribute(F, :standard_extensions) do
    Dict{ZZRingElem, FinField}()
  end
```
or even this
```suggestion
  get_attribute(Dict{ZZRingElem, FinField}, F, :standard_extensions)
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+
+# TODO: Lübeck speeds this up by caching triples [q,m,a] resulting from this
+function standard_affine_shift(q::IntegerUnion, i::IntegerUnion)
+    m = div(4*q, 5)
+    while gcd(m,q) != 1
+        m -= 1
+    end
+    a = div(2*q, 3)
+    return mod((m*i + a), q)
+end
```
fingolfin: So to see if I understand your caching remark above right: the values `m` and `a` only depend on `q`, so we could cache these values in say a dictionary mapping `q => (m,a)`.

Given that `standard_affine_shift` seems to be called in loops, each times with fixed `q`, this sounds sensible. 

In both places calling this, the value `q` is the order of some field `F`. So a natural place to cache `(m,a)` would be in an attribute of `F`. So you could do something like this
```suggestion
function standard_affine_shift(F::Field, i::IntegerUnion)
    m, a = get_attribute!(F, :standard_affine_shift_data) do
        m = div(4*q, 5)
        while gcd(m,q) != 1
            m -= 1
        end
        a = div(2*q, 3)
        return m, a
    end::Tuple{Int,Int}
    return mod((m*i + a), q)
end
```
But that then still requires a hash table lookup each time it is called. It seems thus better to do something like this:
```suggestion
@attr Tuple{Int,Int} function standard_affine_shift_data(F::Field, i::IntegerUnion)
    m, a = get_attribute!(F, :standard_affine_shift_data) do
    m = div(4*q, 5)
    while gcd(m,q) != 1
        m -= 1
    end
    a = div(2*q, 3)
    return m, a
end
```
and then call this at the start of the two functions that invoke `standard_affine_shift`. Then each call to `standard_affine_shift(q,i)` can be replaced by `mod((m*i + a), q)`.

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+# we REQUIRE that F is a standard finite field TODO: using @assert?
+function element_from_steinitz_number(F::PrimeField, n::IntegerUnion)
+    return F(n)
+end
+function element_from_steinitz_number(F::FinField, n::IntegerUnion)
+    p = characteristic(F)
+    q = order(F)
+    if n < 0 || n > q
+        error("We need to have 0 <= n <= q")
+    end
```
fingolfin:
```suggestion
    @req 0 <= n <= q "We need to have 0 <= n <= q"
```

experimental/StandardFiniteFields/StandardFiniteFields.jl
```diff
@@ -0,0 +1,462 @@
...
+    if n == 0
+        return zero(F)
+    else
```
fingolfin: More idiomatic would be
```suggestion
    n == 0 && return zero(F)
```
and then unindent the following code

fingolfin (commented):
Just a few quick comments

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: Ah yes, a specific type. The first version required p::ZZRingElem, it now allows p::Union{ZZRingElem, Integer} so I should remove this comment as resolved.

experimental/StandardFiniteFields/StandardFiniteFields.jl
fingolfin: Oh you already did, see https://github.com/Nemocas/Nemo.jl/pull/1464 -- great!!!

experimental/StandardFiniteFields/StandardFiniteFields.jl
thofma: Yes, I agree that it should just be `ZZ(a)`. It is also shorter.

mjrodgers: Hm, I was hoping to just allow generic code so that if I am working with an `Int`, it gets promoted to `BigInt` to avoid some overflow problems (while still being an integer), but if I am working with a `ZZRingElem`, it just gets left alone. But maybe it makes sense for me to promote all my big numbers to `ZZ`. I'll try this out.

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: I guess I'm not doing any operations with these values here, it's just a difference of what gets used as a dictionary key. Probably an `Int` is sufficient in all cases here (and my NOTE was really just noticing the discrepancy between my original comments and my actual implementation).

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: I guess you are right. I felt like they should be there, but maybe not. It makes the most sense to always set these using `get_steinitz_prime_degree!` instead, since we always want to check if it is already computed before computing it again.

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: I've made this change, but the more I think about it, this is used in two recursive functions where I factor n, divide out the largest prime factor to obtain n', and then pass this to a new function call that.... factor it again. So I will work on passing the factorization through instead of repeating the work.

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: Fixed in commit 72efa8d7359ea482b970d67180a4a8a631cfe97c

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: fixed in commit 72efa8d7359ea482b970d67180a4a8a631cfe97c

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: fixed in commit 72efa8d7359ea482b970d67180a4a8a631cfe97c

experimental/StandardFiniteFields/StandardFiniteFields.jl
mjrodgers: fixed in commit 72efa8d7359ea482b970d67180a4a8a631cfe97c

fieker:
On Mon, May 15, 2023 at 02:22:32AM -0700, Morgan Rodgers wrote:
> @mjrodgers commented on this pull request.
> 
> 
> 
> > +    nfactors = sort([r for (r,e) in nfactorization])
> +    a = Int(nfactors[end])
> 
> I've made this change, but the more I think about it, this is used in two recursive functions where I factor n, divide out the largest prime factor to obtain n', and then pass this to a new function call that.... factor it again. So I will work on passing the factorization through instead of repeating the work.
> 
There is s.th. called FacElem...
julia> FacElem(ZZ, Dict{ZZRingElem, Int}(2=>3, 3=>16))
Factored element with data
Dict{ZZRingElem, ZZRingElem}(2 => 3, 3 => 16)

julia> evaluate(ans)
344373768

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -0,0 +1,506 @@
...
+
+
+@doc raw"""
+    standard_finite_field(p::Union{ZZRingElem, Integer}, n::Union{ZZRingElem, Integer}) -> FinField
+
+Returns a finite field of order $p^n$.
+
+# Examples
+```jldoctest
+julia> standard_finite_field(3, 24)
```
fingolfin, line 458: This can't work because `standard_finite_field` is not exported by `Oscar`

mjrodgers: ah, okay. So should I comment out the example and then the doctest will pass? Or can it be tested as `Oscar.standard_finite_field`?

experimental/StandardFiniteFields/docs/doc.main
```diff
@@ -0,0 +1,9 @@
+# An example documentation
+```@meta
+CurrentModule = Oscar
```
fingolfin:
```suggestion
CurrentModule = Oscar
DocTestSetup = quote
  using Oscar
end
```

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -0,0 +1,506 @@
...
+    return _extension_with_tower_basis(K, m, c, b)
+  end
+
+  return get_standard_extension!(F, n) do
+    N = factor(ZZ(n))
+    return _sff(N)
+  end
+end
+
+end
```
fingolfin, line 498:
```suggestion
end

using .StandardFiniteFields
export standard_finite_field
```

experimental/StandardFiniteFields/docs/doc.main
mjrodgers: resolveed by ab03fee

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
+using Test
+using Oscar
```
lgoettgens:
```suggestion
```
You do not need these two lines, as the test procedure or `Oscar.test_experimental_module` handle that.

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
...
+GAP.Packages.load("StandardFF")
+
```
lgoettgens, line 2:
```suggestion

@testset "standard_finite_field unit tests" begin
```
I would recommend starting your testset already here. This has the upside of not cluttering the global namespace with your functions `MyPoly` etc., sinse `@testset` introduces a new scope.

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
...
+function MyPoly(p,n)
+    return defining_polynomial(standard_finite_field(p,n))
```
lgoettgens:
```suggestion
    return defining_polynomial(standard_finite_field(p, n))
```

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
...
+end
+
+function GAPPoly(p,n)
+    poly = GAP.evalstr( "StandardFiniteField(" * string(p) * ", " * string(n) * ")!.DefiningPolynomial!.CoefficientsOfUnivariatePolynomial")
+    poly = polynomial(GF(p), map(x -> GF(p)(x), GAP.gap_to_julia(Vector{GAP.FFE}, poly)))
```
lgoettgens:
```suggestion
    poly = polynomial(GF(p), map(GF(p), Vector{GAP.FFE}(poly)))
```
The use of `gap_to_julia` is not type-safe.

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
+using Test
+using Oscar
+GAP.Packages.load("StandardFF")
+
+function MyPoly(p,n)
+    return defining_polynomial(standard_finite_field(p,n))
+end
+
+function GAPPoly(p,n)
+    poly = GAP.evalstr( "StandardFiniteField(" * string(p) * ", " * string(n) * ")!.DefiningPolynomial!.CoefficientsOfUnivariatePolynomial")
```
lgoettgens:
```suggestion
    poly = GAP.evalstr("StandardFiniteField($(p), $(n))!.DefiningPolynomial!.CoefficientsOfUnivariatePolynomial")
```
easier to read

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -0,0 +1,509 @@
...
+  return l
+end
+
+# returns the Steinitz number corresponding to the polynomial g(X),
+# where f = X^r + g(X) is the standard irreducible polynomial over FF(p, r^(k-1))
+function steinitz_number_for_prime_degree(
+  p::IntegerUnion,
+  r::IntegerUnion,
+  k::IntegerUnion,
+)::ZZRingElem
```
lgoettgens: Return type annotation should, in general, not be used. See https://docs.julialang.org/en/v1/manual/functions/#Return-type (last paragraph)

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -0,0 +1,509 @@
...
+
+# Embed an element x of Fp^n into Fp^m by Steinitz numbers
+# where nr = steinitz_number(Fp^n, x)
+# I hate hate hate these variable names copied (mostly) from Lübeck
+function embed_steinitz(p::IntegerUnion, n::IntegerUnion, m::IntegerUnion, nr::IntegerUnion)
+  if n == m || iszero(nr)
+    return nr
+  end
+  l = digits(nr, base = Int(p))
+  m = @view standard_monomial_map(n, m)[1:length(l)]
```
lgoettgens: this function is not typesafe, as `m` first is an `IntegerUnion`, and now some kind of container. Maybe use two distinct names to distinguish?

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -0,0 +1,509 @@
...
+  K::T,
+  deg::IntegerUnion,
+  lcoeffs::Vector,
+  b::FinFieldElem,
+) where {T<:FinField}
+  @assert parent(b) === K
+
+  dK = absolute_degree(K)
+  if dK == 1
+    then
```
lgoettgens:
```suggestion
```

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
lgoettgens: This should be fine now due to the export outside of your module in line 509.

lgoettgens (commented):
A few remarks I found while scrolling through your code

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
lgoettgens: This should be fine now due to the export outside of your module in line 509.

experimental/StandardFiniteFields/test/runtests.jl
fingolfin: Better yet, don't use `evalstr` (it is almost never necessary and should be avoided). Also don't access internal fields of objects if possible, use the proper accessors instead. So:
```suggestion
    poly = GAP.Globals.StandardFiniteField(p, n)
    poly = GAP.Globals.DefiningPolynomial(poly)
    poly = GAP.Globals.CoefficientsOfUnivariatePolynomial(poly)
```

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
...
+    poly = polynomial(GF(p), map(x -> GF(p)(x), GAP.gap_to_julia(Vector{GAP.FFE}, poly)))
+end
+
+small_test_primes = ZZ.([3, 5, 7])
+medium_test_primes = ZZ.([3433, 4073])
```
fingolfin: These two don't seem to be used. However the tests later on reference `small_primes` which in turn doesn't seem to be defined anywhere?

experimental/StandardFiniteFields/test/runtests.jl
mjrodgers: Thanks Lars, Max, this cleans it up quite nicely

experimental/StandardFiniteFields/test/runtests.jl
mjrodgers: thanks, this was an oversight about `small_primes` vs `small_test_primes`. I decided that testing with small primes should be sufficient for now, so I commented out the `medium_test_primes`.

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
mjrodgers: Hm, I've tried to avoid it in general. But in this case, I have several possible return points, and I want to make sure that no matter what the result is converted and returned as a `ZZRingElem` and not an `Integer`.... I actually think I can get around this by making sure that `steinitz_number` will always return a `ZZRingElem`.

lgoettgens: Great! Explicit conversion in each of the return cases is the way to go. You can check if you got all cases with `@code_warntype`

experimental/StandardFiniteFields/test/runtests.jl
```diff
@@ -0,0 +1,44 @@
+GAP.Packages.load("StandardFF")
+
+@testset "standard_finite_field unit tests" begin
+  function MyPoly(p, n)
+    return defining_polynomial(standard_finite_field(p, n))
+  end
+
+  function GAPPoly(p, n)
+    poly = GAP.Globals.StandardFiniteField(Integer(p), n)
+    poly = GAP.Globalse.DefiningPolynomial(poly)
```
lgoettgens:
```suggestion
    poly = GAP.Globals.DefiningPolynomial(poly)
```

mjrodgers:
I think this is ready to merge into experimental, it now passes all tests. I think there is still room for performance enhancements, and some more functionality, but the basic field constructor is in place.

thofma:
I bumped the Nemo version and adjusted something here. If there are no objections, I will merge this in a few hours/days. @fieker @fingolfin

mjrodgers:
I got the `.toml` issue fixed, and the failing tests seem to be unrelated to anything in the pull request. I think this can be merged. @thofma @fingolfin

PR #2370, opened 2023-05-10
Add Aqua.jl for some static testing

### LINKED ISSUE #1964: Add CI tests to our projects looking for "bad" exports

lgoettgens:
Resolves #1964.

Three of the performed tests (`ambiguities`, `unbound_args`, and `piracy`) currently fail and are thus disabled (with a comment to check if/with how much work they can be fixed).

`deps_compat` requires all dependencies to have a `[compat]` entry, so I added the missing ones. The JuliaRegistry currently does not require those for JLLs (cf. the last paragraph of https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/#Upper-bounded-[compat]-entries), but adding them should not hurt.

`stale_deps` fails on my local setup with some weird error containing the UUID of `GAP_lib_jll`. This can be investigated in a separate PR.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,6 +26,7 @@ cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 [compat]
 AbstractAlgebra = "0.29.3"
 AlgebraicSolving = "0.3.0"
+Aqua = "0.6"
 DocStringExtensions = "0.8, 0.9"
 GAP = "0.9.4"
 Hecke = "0.18.9"
@@ -36,13 +37,16 @@ Preferences = "1"
 RandomExtensions = "0.4.3"
 RecipesBase = "1.2.1"
 Singular = "0.18.2"
+TOPCOM_jll = "0.17.8"
+cohomCalg_jll = "0.32.0"
 julia = "1.6"
 
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Documenter", "PrettyTables", "Printf"]
+test = ["Test", "Aqua", "Documenter", "PrettyTables", "Printf"]
--- a/test/Aqua.jl
+++ b/test/Aqua.jl
@@ -1,0 +1,13 @@
+@testset "Aqua.jl" begin
+  Aqua.test_all(
+    Oscar;
+    ambiguities=false,      # TODO: fix ambiguities
+    unbound_args=false,     # TODO: fix unbound type parameters
+    undefined_exports=true,
+    project_extras=true,
+    stale_deps=false,       # some weird error with GAP_lib_jll
+    deps_compat=true,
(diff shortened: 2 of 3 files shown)


Project.toml
```diff
@@ -26,6 +26,7 @@ cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 [compat]
 AbstractAlgebra = "0.29.3"
 AlgebraicSolving = "0.3.0"
+Aqua = "0.6"
```
fingolfin, line 29: Do we really need this as a dependency when it is only used by the tests?

lgoettgens: It is not a dependency but only listed in the `[extra]`. With the compact entry, we can make sure that an `Aqua.jl` release (that probably will change some of the performed tests) does not break our CI.

lgoettgens: At least in theory. I could unfortunately not find a statement in the docs if `[compact]` is used at all for `[extra]` dependencies.

fingolfin: yeah, I misphrased that, sorry

Anyway, I think we should move away from the old-style usage of `[extras]` etc in the main `Project.toml` and instead move to a dedicated `test/Project.toml`, that would neatly resolve this issue.

lgoettgens: I strongly agree. But I think this should be done in a separate PR. I created #2374 to remember.

thofma: What exactly is the issue? I don't see a problem that is solved by this.

lgoettgens: > What exactly is the issue? I don't see a problem that is solved by this.

`Aqua.jl` may release a new version `v0.7` in the future that includes some additional test, which fails for OSCAR.
Without this, the second `v0.7` is released, our CI breaks (on all branches). With this change, we can change the used `Aqua.jl` version ourselves, in a controlled environment/PR.

Note:
> At least in theory. I could unfortunately not find a statement in the docs if [compact] is used at all for [extra] dependencies.

thofma: Yes, the `[compat]` entry is also valid for the `[extra]` entries, which you did here. So everything is fine.

fingolfin: @lgoettgens agreed (about the separate PR)

fingolfin (commented):
Fine by me

thofma (approved):

Issue #2222, opened 2023-04-05
Missing `hash` methods


fingolfin:
There are a bunch of our types which implement `==` but not `hash`, which leads to confusing behavior and bugs. E.g. for a vector consisting of two such objects which are equal but not identical, the function `unique` will return a vector with both objects, instead of just one. @YueRen just re-discovered this for `Polyhedron`, but there are more affected types.

We have a small script (taken from https://github.com/oscar-system/Oscar.jl/issues/2222#issuecomment-2578162578 by @lgoettgens) that finds affected types (not perfect, but finds many cases):
```julia
typelist = [];

for package in [Oscar, Hecke, Nemo, AbstractAlgebra]
  Oscar.walkmodules(package) do mod
    append!(typelist, filter(!isnothing, map(methods(==, mod)) do meth
      meth.nargs == 3 || return nothing
      sig = Base.unwrap_unionall(meth.sig)
      arg1_type = sig.types[2]
      arg2_type = sig.types[3]
      arg1_type == arg2_type || return nothing                  # arg1 and arg2 must have the same type to be considered atm
      T = arg1_type
      meth.module == Base && return nothing                     # remove everything without custom ==
      parentmodule(hash, (T, UInt)) == Base || return nothing   # remove everything with custom hash
      T
    end))
  end;
end;

list = map(typelist) do T
  loc = functionloc(==, (T, T))
  if occursin("Oscar.jl/", loc[1])
    loc_cleaned = loc[1][first(findlast("Oscar.jl/", loc[1])):end]
  elseif occursin("packages/", loc[1])
    loc_cleaned = loc[1][(first(findlast("packages/", loc[1])) + length("packages/")):end]
  elseif occursin("julia/base/", loc[1])
    loc_cleaned = loc[1][first(findlast("julia/base/", loc[1])):end]
  else
    loc_cleaned = loc[1]
  end
  T, (T isa TypeVar ? parentmodule(T.ub) : parentmodule(T)), (loc_cleaned, loc[2])
end;

show(IOContext(stdout, :limit => false), "text/plain", list)
```

current list (as of 50eb9d8ef43c5b57ba9eff54686dad6e6d615cb2):
```julia
11-element Vector{Tuple{Type, Module, Tuple{String, Int32}}}:
 (Oscar.PartialCharacter{T<:FieldElem}, Oscar, ("Oscar.jl/src/Rings/binomial_ideals.jl", 408))
[...]

PR #2373, opened 2023-05-10
Add `hash` methods for many types with `==` method

### LINKED ISSUE #1264: ToricVarieties: Add Base.hash
Should we add `Base.hash` for our custom types `(AffineNormalToricVarieties`, `AbstractNormalToricVarieties`, `ToricDivisors`, `ToricDivsiorClasses`, `ToricLineBundles`, (soon `ToricVectorBundles` and `ToricCoherentSheaves`) and many more...)?

C.f.: https://docs.julialang.org/en/v1/base/base/#Base.hash

lgoettgens:
This addresses parts of #2222 and resolves #1264.

Julia requires that equal elements (w.r.t. `==`) hash to the same value to make `Set`s etc. work.
For many OSCAR types (cf. #2222), there currently is a dedicated `==`, but no `hash`; thus contradicting the requirement.

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
@@ -267,11 +267,17 @@ end
 
 
 ####################################################
-# 6: Equality
+# 6: Equality and hash
 ####################################################
 
 function Base.:(==)(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
     return toric_variety(ac1) === toric_variety(ac2) && iszero(polynomial(ac1-ac2))
+end
+
+function Base.hash(ac::RationalEquivalenceClass, h::UInt) 
+  b = 0xb5d4ac6b9084eb6e  % UInt
+  h = hash(toric_variety(ac), h)
+  return xor(h, b)
 end
 
 
--- a/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
@@ -271,13 +271,14 @@ end
 ####################################################
 
 function Base.:(==)(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
-    return toric_variety(ac1) === toric_variety(ac2) && iszero(polynomial(ac1-ac2))
+    return toric_variety(ac1) === toric_variety(ac2) && polynomial(ac1) == polynomial(ac2)
 end
 
 function Base.hash(ac::RationalEquivalenceClass, h::UInt) 
-  b = 0xb5d4ac6b9084eb6e  % UInt
-  h = hash(toric_variety(ac), h)
-  return xor(h, b)
+    b = 0xb5d4ac6b9084eb6e  % UInt
+    h = hash(toric_variety(ac), h)
+    h = hash(polynomial(ac), h)
+    return xor(h, b)
 end
 
 
--- a/src/Groups/homomorphisms.jl
+++ b/src/Groups/homomorphisms.jl
@@ -6,7 +6,12 @@ function Base.show(io::IO, x::GAPGroupHomomorphism)
 end
 
 function ==(f::GAPGroupHomomorphism{S,T}, g::GAPGroupHomomorphism{S,T}) where S where T
-   return f.map == g.map
+  return f.map == g.map
+end
+
+function Base.hash(f::GAPGroupHomomorphism{S,T}, h::UInt) where S where T
+  b = 0xdc777737af4c0c7b % UInt
+  return xor(hash(f.map, hash((S, T), h)), b)
 end
 
 Base.:*(f::GAPGroupHomomorphism{S, T}, g::GAPGroupHomomorphism{T, U}) where S where T where U = compose(f, g)
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -3523,37 +3523,50 @@ function +(a::SubquoModuleElem, b::SubquoModuleElem)
   check_parent(a,b)
   return SubquoModuleElem(coordinates(a)+coordinates(b), a.parent)
 end
+
 function -(a::SubquoModuleElem, b::SubquoModuleElem) 
   check_parent(a,b)
   return SubquoModuleElem(coordinates(a)-coordinates(b), a.parent)
 end
+
 -(a::SubquoModuleElem) = SubquoModuleElem(-coordinates(a), a.parent)
+
 function *(a::MPolyDecRingElem, b::SubquoModuleElem) 
   if parent(a) !== base_ring(parent(b))
     return base_ring(parent(b))(a)*b # this will throw if conversion is not possible
   end
   return SubquoModuleElem(a*coordinates(b), b.parent)
 end
+
 function *(a::MPolyRingElem, b::SubquoModuleElem) 
   if parent(a) !== base_ring(parent(b))
     return base_ring(parent(b))(a)*b # this will throw if conversion is not possible
   end
   return SubquoModuleElem(a*coordinates(b), b.parent)
 end
+
 function *(a::RingElem, b::SubquoModuleElem) 
   if parent(a) !== base_ring(parent(b))
     return base_ring(parent(b))(a)*b # this will throw if conversion is not possible
   end
   return SubquoModuleElem(a*coordinates(b), b.parent)
 end
+
 *(a::Int, b::SubquoModuleElem) = SubquoModuleElem(a*coordinates(b), b.parent)
 *(a::Integer, b::SubquoModuleElem) = SubquoModuleElem(a*coordinates(b), b.parent)
 *(a::QQFieldElem, b::SubquoModuleElem) = SubquoModuleElem(a*coordinates(b), b.parent)
+
 function (==)(a::SubquoModuleElem, b::SubquoModuleElem) 
   if parent(a) !== parent(b)
     return false
   end
   return iszero(a-b)
+end
+
+function Base.hash(a::SubquoModuleElem, h::UInt)
+  b = 0x0361a2f6467e4200 % UInt
+  h = hash(parent(a), h)
+  return xor(h, b)
 end
 
 function Base.deepcopy_internal(a::SubquoModuleElem, dict::IdDict)
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -3564,9 +3564,7 @@ function (==)(a::SubquoModuleElem, b::SubquoModuleElem)
 end
 
 function Base.hash(a::SubquoModuleElem, h::UInt)
-  b = 0x0361a2f6467e4200 % UInt
-  h = hash(parent(a), h)
-  return xor(h, b)
+  error("not implemented")
 end
 
 function Base.deepcopy_internal(a::SubquoModuleElem, dict::IdDict)
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -4820,6 +4820,15 @@ function (==)(f::ModuleFPHom, g::ModuleFPHom)
   end
   return true
 end
+
+function Base.hash(f::ModuleFPHom, h::UInt)
+  b = 0x535bbdbb2bc54b46 % UInt
+  h = hash(typeof(f), h)
+  h = hash(domain(f), h)
+  h = hash(codomain(f), h)
+  return xor(h, b)
+end
+
 ###################################################################
 
 @doc raw"""
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -181,8 +181,12 @@ function (==)(F::FreeMod, G::FreeMod)
 end
 
 function hash(F::FreeMod, h::UInt)
-  is_graded(F) && return hash((base_ring(F), rank(F), F.S, F.d), h)
-  return hash((base_ring(F), rank(F), F.S), h)
+  b = is_graded(F) ? (0x2d55d561d3f7e215 % UInt) : (0x62ca4181ff3a12f4 % UInt)
+  h = hash(base_ring(F), h)
+  h = hash(rank(F), h)
+  h = hash(F.S, h)
+  is_graded(F) && (h = hash(F.d, h))
+  return xor(h, b)
 end
 
 @doc raw"""
@@ -490,7 +494,11 @@ function (==)(a::AbstractFreeModElem, b::AbstractFreeModElem)
 end
 
 function hash(a::AbstractFreeModElem, h::UInt)
-  return xor(hash(tuple(parent(a), coordinates(a)), h), hash(typeof(a)))
+  b = 0xaa2ba4a32dd0b431 % UInt
+  h = hash(typeof(a), h)
+  h = hash(parent(a), h)
+  h = hash(coordinates(a), h)
+  return xor(h, b)
 end
 
 function Base.deepcopy_internal(a::AbstractFreeModElem, dict::IdDict)
--- a/src/Rings/mpoly-localizations.jl
+++ b/src/Rings/mpoly-localizations.jl
@@ -674,11 +674,8 @@ function ==(T::AbsMPolyMultSet, U::AbsMPolyMultSet)
   return (issubset(T, U) && issubset(U, T))
 end
 
-function Base.hash(T::AbsMPolyMultSet, h::UInt)
-  b = 0x7ce51a28c47ec5e1 % UInt
-  h = hash(typeof(T), h)
-  h = hash(ambient_ring(T), h)
-  return xor(h, b)
+function Base.hash(T::AbsMPolyMultSet{BRT, BRET, RT, RET}, h::UInt) where {BRT,BRET,RT,RET}
+  error("not implemented")
 end
 
 function issubset(
--- a/src/Rings/mpoly-localizations.jl
+++ b/src/Rings/mpoly-localizations.jl
@@ -674,6 +674,13 @@ function ==(T::AbsMPolyMultSet, U::AbsMPolyMultSet)
   return (issubset(T, U) && issubset(U, T))
 end
 
+function Base.hash(T::AbsMPolyMultSet, h::UInt)
+  b = 0x7ce51a28c47ec5e1 % UInt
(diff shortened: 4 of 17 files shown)


lgoettgens:
Most implementations are purely based on the things that are compared in the `==` implementation right above. If `x == y` does something like `issubset(x,y) || issubset(y,x)` or `iszero(reduce(x.arg - y.arg))` this is not used, and the hash will be the same for all elements with the same parent (or similar).

fieker (commented):
The hash functions apprea to be correct, but are very weak in many cases. Can someone from teh geometry crowed maybe have a look?
Fundamentally, I'm not sure if a weak hash function is better than throwing an error. Several hash functinos introduced here shuold never be used.

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
```diff
@@ -265,13 +265,19 @@ end
...
 ####################################################
 
 function Base.:(==)(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
     return toric_variety(ac1) === toric_variety(ac2) && iszero(polynomial(ac1-ac2))
 end
 
+function Base.hash(ac::RationalEquivalenceClass, h::UInt) 
+  b = 0xb5d4ac6b9084eb6e  % UInt
+  h = hash(toric_variety(ac), h)
+  return xor(h, b)
```
HereAround: Thank you for working on this @lgoettgens !

As you probably know, the toric variety alone does not specify a rational equivalence class. So suppose `ac1` and `ac2` are two distinct rational equivalence classes on the *same* toric variety, then your proposed hash function would return the same value for `ac1` and `ac2`, despite `ac1` and `ac2` being distinct.

From discussion with @fieker , I learned that the default hash (that julia must be using right now) is the memory address, which of course identifies `ac1` and `ac2` uniquely. In comparison, your proposed hash function is significantly weaker. My feeling (and also that of @fieker) is that this is not desired. (Please correct me if I am wrong.)

I see two ways out.
1. Stick to the current `julia` default. But maybe there are good reasons to not do this?
2. Extend your hash functions to identify `ac1` and `ac2` uniquely. To this end, we can look at the type definition:
```julia
@attributes mutable struct RationalEquivalenceClass
    v::AbstractNormalToricVariety
    p::MPolyQuoRingElem
    RationalEquivalenceClass(v::AbstractNormalToricVariety, p::MPolyQuoRingElem) = new(v, p)
end
```
So a possibility would be to use e.g. the following:
```julia
function Base.hash(ac::RationalEquivalenceClass, h::UInt) 
  b = 0xb5d4ac6b9084eb6e  % UInt
  h = hash(toric_variety(ac), h) + hash(ac.p,h)
  return xor(h, b)
```
And similarly for the other toric objects.

cc @lkastner

HereAround (changes requested):

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
lgoettgens: Thanks for your remark. Unfortunately, your proposal does not work in the intended way (if I understand it correctly; sorry if not). 

Julia requires for its internal functionality (e.g. `Set`) to work properly, that the equality of hashes is not finer than the usual equality `==`.

For your 1. Idea, this is clearly violated since hashing via the julia default results in the most finest equivalence relation, but there is a coarses one defined via `==`.

For your 2. idea, I am not familiar enough with the implementation to make a reasonable statement here. As I understood it, there can be equal RationalEquivalenceClass objects with a different `p` (at least this is what I thought based on the implementation of `==`). If this is true, I cannot see a way of sensibly defining a hash function without first choosing a canonical representative of the class (if that is even possible; not my field of knowledge). If my assumption is wrong, I would suggest to change the implementation of `==` to check for equality of the `toric_variety` and of `p` and change `hash` according to your proposal.

HereAround: > Julia requires for its internal functionality (e.g. `Set`) to work properly, that the equality of hashes is not finer than the usual equality `==`.
 
Ok. What I am proposing is to make the hash equally fine as `==`.

> For your 1. Idea, this is clearly violated since hashing via the julia default results in the most finest equivalence relation, but there is a coarses one defined via `==`.

I see.

> For your 2. idea, I am not familiar enough with the implementation to make a reasonable statement here. As I understood it, there can be equal RationalEquivalenceClass objects with a different `p` (at least this is what I thought based on the implementation of `==`). If this is true, I cannot see a way of sensibly defining a hash function without first choosing a canonical representative of the class (if that is even possible; not my field of knowledge). If my assumption is wrong, I would suggest to change the implementation of `==` to check for equality of the `toric_variety` and of `p` and change `hash` according to your proposal.

The toric variety together with p specifies a `RationalEquivalenceClass` uniquely.

The equality function is:
```julia
function Base.:(==)(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
    return toric_variety(ac1) === toric_variety(ac2) && iszero(polynomial(ac1-ac2))
end
```
Note that `polynomial(ac1) = ac1.p`. In other words, this boils down to asking whether or not `ac1.p - ac2.p` is zero.

lgoettgens: > ```julia
> function Base.:(==)(ac1::RationalEquivalenceClass, ac2::RationalEquivalenceClass)
>     return toric_variety(ac1) === toric_variety(ac2) && iszero(polynomial(ac1-ac2))
> end
> ```
> 
> Note that `polynomial(ac1) = ac1.p`. In other words, this boils down to asking whether or not `ac1.p - ac2.p` is zero.

If this is indeed equivalent to `toric_variety(ac1) === toric_variety(ac2) && polynomial(ac1) == polynomial(ac2)`, I would be happy to adapt both `==` and `hash`. (Other possibility would be some implentation details of `MPolyQuoElem` having elements which are not `==`-equal, but whose difference is zero. Maybe you know, I did not check for that yet.) This would then satisfy julia's requirement, given that this is already satisfied for `MPolyQuoElem` (which I didn't check).

src/Rings/mpoly-localizations.jl
```diff
@@ -533,6 +533,13 @@ function ==(T::AbsMPolyMultSet, U::AbsMPolyMultSet)
   return (issubset(T, U) && issubset(U, T))
 end
 
+function Base.hash(T::AbsMPolyMultSet, h::UInt)
+  b = 0x7ce51a28c47ec5e1 % UInt
+  h = hash(typeof(T), h)
+  h = hash(ambient_ring(T), h)
+  return xor(h, b)
+end
```
HechtiDerLachs: Thanks for the effort! However, this kind of hashing will not be sufficient. Instead, I propose to throw an error here which requires the user to overwrite the hash function for their concrete type. 

Can I do that and push to this PR?

src/Rings/mpoly-localizations.jl
```diff
@@ -2584,6 +2591,13 @@ function ==(f::MPolyLocalizedRingHom, g::MPolyLocalizedRingHom)
   return true
 end
 
+function Base.hash(f::MPolyLocalizedRingHom, h::UInt)
+  b = 0xe29cdc1ecb68b7a0 % UInt
+  h = hash(domain(f), h)
+  h = hash(codomain(f), h)
```
HechtiDerLachs, line 2900: I would add a hash of the images of the generators. Or alternatively the `restricted_map`.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1055,6 +1055,13 @@ function ==(f::MPolyQuoLocalizedRingHom, g::MPolyQuoLocalizedRingHom)
   return true
 end
 
+function Base.hash(f::MPolyQuoLocalizedRingHom, h::UInt)
+  b = 0xd6d389598ad28724  % UInt
+  h = hash(domain(f), h)
+  h = hash(codomain(f), h)
```
HechtiDerLachs, line 1085: Same here.

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
lgoettgens: I discussed with @HereAround on slack how to procede for the toric types. We will update the implementations of `==` and then change `hash` accordingly.

src/Rings/mpoly-localizations.jl
lgoettgens: I should have re-read my code before pushing :sweat_smile: 
Done in 98c94979a6d4e4ae204aae9f6d5f695278e70fd6

src/Rings/mpolyquo-localizations.jl
lgoettgens: Done in 98c94979a6d4e4ae204aae9f6d5f695278e70fd6

src/Rings/mpoly-localizations.jl
lgoettgens: Do you mean something similar to this?
 https://github.com/oscar-system/Oscar.jl/blob/96c2eed7c8b4ed2289290ffc0254b7ddf5122d12/src/Rings/mpoly-localizations.jl#L529

Yeah, that would be great. Please text me on slack to coordinate, so that we delete nothing by accidentally force pushing anything.

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
HereAround: Thank you @lgoettgens !

lgoettgens: The toric types have been adapted in 7b1796cf2d34c7e7a258168aa71db9ac9a63bf85. Please check if this it what you imagined @HereAround.

HereAround: Thank you @lgoettgens . The toric part looks good to me! In this sense I approved.

(However, I understand that @HechtiDerLachs and @fieker definitely wanted to take a closer look at the changes outside of the toric varieties.)

HereAround (approved):

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
HereAround: I am working on a bugfix for the toric varieties (https://github.com/oscar-system/Oscar.jl/pull/2396). There is a good chance that there will be a merge conflict for toric line bundles. The new and "good" methods use `picard_class` to compare two line bundles:
```julia
function Base.:(==)(l1::ToricLineBundle, l2::ToricLineBundle)
    return toric_variety(l1) === toric_variety(l2) && picard_class(l1) == picard_class(l2)
end
```
In other words, `divisor_class(...)` was/is to be replaced by `picard_class(...)`.

fingolfin:
In principle this is fine by me, but there is (a) a conflict and (b) this is marked as a draft, so I won't look closer for now.

src/Rings/mpoly-localizations.jl
lgoettgens: I added a warning in 0dfcbe8a7148f564351b7d6c979e85b385e3b859. Is this fine with you?

lgoettgens: After looking a bit deeper into the concrete subtypes, I think there cannot be special implementations of `hash` for each type. There are many different implementations for `issubset` for different pairs of types, and equality is just defined generically using `issubset`. Since AbsMPolyMultSets of different concrete types can be equal, their hashes should coincide as well. And I do not see how to do that with specialized implementations.

lgoettgens:
Let's see if the CI runs without any warnings.

fieker (dismissed):
function Base.hash(a::SubquoModuleElem, h::UInt)

the current hash does not depend on the element, only on the parent. This is julia-correct, but is this what you want? I'd rather throw an error... @HechtiDerLachs ?

Similar: GAPGroupHomomorphism @fingolfin  is there a decent hash for f.map or should this be an error?

fingolfin:
I don't see a way to define an efficient hash for group morphisms, so I think that should throw an error

lgoettgens:
I would refrain from throwing errors in these cases to not render these types useless for anyone who needs to use them in some container/setting with hashing (even without knowning). Instead, I would like to have some kind of warning. If possible, this warning should be marking the CI as failed.
Is something like this possible?

thofma:
If the warning is supposed to make CI fail, why not error in the first place?

Or why not define the hash to be `0 % UInt` for the types where there is no efficient hashing? Then one can still use them for containers.

lgoettgens:
> If the warning is supposed to make CI fail, why not error in the first place?
>
> Or why not define the hash to be `0 % UInt` for the types where there is no efficient hashing? Then one can still use them for containers.

The latter is my proposal in this PR. I use every little bit of information to make at least some hashes different to at least circumvent hash conflicts for objects of completely different types.

fingolfin:
> Or why not define the hash to be `0 % UInt` for the types where there is no efficient hashing? Then one can still use them for containers.

That's one way to put it, as an advantage. But there's also the counterpoint: with a 0-hash, one can use these objects as keys for a container, without noticing at first that this is a terrible idea with worst case performance. Other than for toy examples it won't be useful...

That's why I'd rather have an error in those cases...

src/AlgebraicGeometry/ToricVarieties/AlgebraicCycles/constructors.jl
lgoettgens: Fixed in the recent merge.

src/Modules/UngradedModules.jl
```diff
@@ -3515,39 +3523,52 @@ function +(a::SubquoModuleElem, b::SubquoModuleElem)
...
   if parent(a) !== parent(b)
     return false
   end
   return iszero(a-b)
 end
 
+function Base.hash(a::SubquoModuleElem, h::UInt)
+  b = 0x0361a2f6467e4200 % UInt
+  h = hash(parent(a), h)
+  return xor(h, b)
```
fingolfin: So this is (one of?) the contentious parts: we (@fieker and me) don't think this is really better (in practical applications) than a function returning a constant, and that of course is mostly useless. So we'd favor an error indicating "not implemented".

Actually, to make it stronger: we require this to merge the PR. However, of course if we run into actual problems caused by this, we can revisit the situation.

src/Groups/homomorphisms.jl
```diff
@@ -6,7 +6,12 @@ function Base.show(io::IO, x::GAPGroupHomomorphism)
 end
 
 function ==(f::GAPGroupHomomorphism{S,T}, g::GAPGroupHomomorphism{S,T}) where S where T
-   return f.map == g.map
+  return f.map == g.map
+end
+
+function Base.hash(f::GAPGroupHomomorphism{S,T}, h::UInt) where S where T
+  b = 0xdc777737af4c0c7b % UInt
+  return xor(hash(f.map, hash((S, T), h)), b)
```
fingolfin, line 14: This is also problematic, because there is no hash for `f.map` only the generic one.

Perhaps I should in fact add a method `hash(::GapObj)` which throws an error to prevent such cases.

lgoettgens: This already exists in GAP.jl in https://github.com/oscar-system/GAP.jl/blob/master/src/adapter.jl#L367

src/Modules/UngradedModules.jl
lgoettgens: Changed in 722eccf78db48580e5131a6a178f5555a9f80764.

I still think this is a bad idea and could potentially render Oscar useless for some users that rely on some julia Base functionality that uses hashing. I would prefer to return 0 or some constant and give a warning (that in the best case should make CI fail) instead. Then, some user can still use Oscar types for their use-case, although it may be very inefficient in those cases.

But if this is what you want, let's go with it.

lgoettgens:
Due to letting some `hash` implementations throw errors, I guess this is now technically a breaking change.

src/Groups/homomorphisms.jl
fingolfin: Yeah that's a trivial (hence useless) hash. See https://github.com/oscar-system/GAP.jl/pull/891 for a PR which turns that into an error.

lgoettgens: I suggest keeping the version proposed here. The error will then be thrown by GAP.jl while trying to hash `f.map`.

fingolfin:
According to @HereAround this should actually completely resolve #1264 -- is that right? If so, let's add `resolves #1264` to the PR description

lgoettgens:
> According to @HereAround this should actually completely resolve #1264 -- is that right? If so, let's add `resolves #1264` to the PR description

Some types from the toric varieties do not get a hash function in this PR, e.g. `CohomologyClass`, `RationalEquivalenceClass`,...

Anyway, I think #1264 can be closed as a duplicate of #2222.

HereAround:
> > According to @HereAround this should actually completely resolve #1264 -- is that right? If so, let's add `resolves #1264` to the PR description
> 
> Some types from the toric varieties do not get a hash function in this PR, e.g. `CohomologyClass`, `RationalEquivalenceClass`,...
> 
> Anyway, I think #1264 can be closed as a duplicate of #2222.

Could you maybe add hashes for these types too? Or is there a reason why this should not be done?

lgoettgens:
> > > According to @HereAround this should actually completely resolve #1264 -- is that right? If so, let's add `resolves #1264` to the PR description
> > 
> > 
> > Some types from the toric varieties do not get a hash function in this PR, e.g. `CohomologyClass`, `RationalEquivalenceClass`,...
> > Anyway, I think #1264 can be closed as a duplicate of #2222.
> 
> Could you maybe add hashes for these types too? Or is there a reason why this should not be done?

I just didn't check this PR's changes thoroughly enough. Everything mentioned in https://github.com/oscar-system/Oscar.jl/issues/2222#issuecomment-1498672088 and being in the `AlgebraicGeometry/ToricVarieties` folder is handled by this PR. So this actually resolves #1264. I change the initial message of this PR accordingly.

lgoettgens:
The force-push is just a rebase onto current master.

lgoettgens:
nightly failure is: https://github.com/oscar-system/GAP.jl/issues/901 / https://github.com/oscar-system/Oscar.jl/issues/2476

fingolfin (approved):
Let's merge this if tests pass now

PR #2375, opened 2023-05-11
Migrate the toric schemes from experimental to src.

HechtiDerLachs:
This is to meet the requirements of F-TheoryTools. 

We would like to use the types for toric schemes already in there and wish to have no dependencies on the experimental section arising from this.


fingolfin:
There are conflicts here and this is marked as draft so I won't look further at it for now.

PR #2376, opened 2023-05-11
Toric divisors on schemes

HechtiDerLachs:
This is a draft for how we can/should proceed to make all functionality for objects living on an `AbstractNormalToricVariety` available on `ToricCoveredScheme`s.


experimental/Schemes/ToricDivisorOnScheme.jl
```diff
@@ -0,0 +1,181 @@
+########################################################################
+# Toric divisors on (toric) schemes
+#
+# This is a wrapper for the `ToricDivisor` functionality to port them 
+# to the realm of `ToricCoveredScheme`s. All the functionality is lazy 
+# and should by default only trigger computations on the toric side. 
+# However, despite inheriting all those methods, the actual type fits 
+# within the scheme's framework and the truely geometric objects are 
+# derived and cached via the respective calls. In particular, these 
+# divisors can seamlessly be used with usual divisors.
```
HechtiDerLachs, line 10: @HereAround , @lkastner : 
This is a draft for how to hopefully facilitate seamless wrappers for `ToricDivisor`s. You can try it out by modifying another file (I will write it below). 

Let me know what you think.

src/AlgebraicGeometry/Schemes/ToricSchemes/ToricDivisors.jl
```diff
@@ -0,0 +1,87 @@
...
+  return divisor_of_character(underlying_toric_variety(X), character)
+end
+
+# Later, we shall return an instance of a new wrapper-type which has the 
+# full functionality of a `ToricDivisor`:
+
+#function divisor_of_character(X::ToricCoveredScheme, character::Vector{T}) where {T<:Integer}
+#  td = divisor_of_character(underlying_toric_variety(X), character)
+#  return ToricDivisorOnScheme(X, td)
+#end
```
HechtiDerLachs, line 25: Just uncomment this to replace lines 15 -- 17. Then you're using the `ToricDivisorOnScheme` stuff. It's a bit messy with the order of loading, since this is in experimental, but if you manage that, then the tests in `toric_divisors.jl` still go through! At least for me.

HechtiDerLachs:
The `ToricDivisorOnScheme` heavily relies on stuff in `experimental`. I'm afraid that when we want to use it, we either have a lot to do migrating stuff, or else we could consider pulling the `ToricDivisors.jl` file from `src` back to experimental. 

I should warn here, though: Migration might probably entail a lot of discussions about design that had been postponed.

HereAround:
@HechtiDerLachs Does it make sense to close this PR and start fresh based on what we now have for the toric schemes?

Also, maybe we should first play around with the varieties/schemes and see where we need/should/must synchronize? For instance `cox_ring` vs. `coordinate_ring` comes to my mind.

HechtiDerLachs:
> Does it make sense to close this PR and start fresh based on what we now have for the toric schemes?

Yes, it does. We should remember the branch, though. I think a lot of work here can be recycled.

PR #2377, opened 2023-05-11
Adjust to recent printing changes

thofma:
This makes the recent "pretty printing" changes available in Oscar, which were implemented here: https://github.com/Nemocas/AbstractAlgebra.jl/pull/1320.

I did not change the printing of any Oscar types. The diff comes only from the changes upstream.

CC: @simonbrandhorst 

Edit: I copied the docs over so that one sees the `Indent()` etc in action.


thofma:
Hm, I found a bug in the printing. I will have to fix this first.

thofma:
Should be good to

docs/src/CommutativeAlgebra/rings.md
```diff
@@ -119,67 +119,71 @@ Gröbner and standard bases are implemented for multivariate polynomial rings ov
 
 ```jldoctest
 julia> QQ
-Rational Field
+Fraction field
+  of integer Ring
```
simonbrandhorst: Is this intended?

thofma: Good catch! Fixed

simonbrandhorst (approved):

thofma:
Any objections to merging this quickly @fingolfin @fieker?

fieker:
On Fri, May 12, 2023 at 02:03:39PM -0700, Tommy Hofmann wrote:
> Any objections to merging this quickly @fingolfin @fieker?
Be my guest and do it

PR #2378, opened 2023-05-12
update_version: Fix disappearing , in README.md

lkastner:


etc/update_version.sh
```diff
@@ -28,7 +28,7 @@ echo "Setting version to $relvers, released $reldate_iso"
 
 # update version in several files
 perl -pi -e 's;version = "[^"]+";version = "'$relvers'";' Project.toml
-perl -pi -e "s;Version [^ }]+;Version $relvers;" README.md
+perl -pi -e "s;Version [^ },]+;Version $relvers;" README.md
```
thofma, line 31:
```suggestion
perl -pi -e "s;Version [^ },]+;Version $relvers,;" README.md
```
? Not sure. Let @fingolfin review this here.

benlorenz (approved):

etc/update_version.sh
benlorenz: It will capture (+replace) only the version, not including the subsequent comma (or `}` or ` ` for different places in the readme), so no need to add another comma after the new version number.

PR #2379, opened 2023-05-12
[FTheoryTools] Add literature models

apturner:
Add nine literature models, including many additional fields beyond those in the previous models.

Models 1b and 2b will ultimately replace 1 and 2, but the latter are kept for now so that tests will pass while we are still updating the code to catch up to the new format.

@HereAround, @lkastner


lkastner (approved):
I'll work on the serialization.

HereAround (approved):
Thank you @apturner

PR #2380, opened 2023-05-12
Small fixes in `docs/make_work.jl`

lgoettgens:
Changes included here:

- Adding `add_prefix_to_experimental_docs(Oscar::Module, docs::Pair{String,String}, prefix::String)`, which was missing.
- Use multi-line string to fix syntax highlighting.
- `normpath` already includes `joinpath` -> remove the `joinpath`
- Make the formatting consistent with the style guide. At some places, even inside the same function, the size of an indentation changed.

cc @lkastner (since I touched your code for docs of experimental modules)

PR #2381, opened 2023-05-12
Restrict imported documentation to exactly the listed files

### LINKED ISSUE #643: docs: the Oscar manual imports more from the AA/Nemo/Hecke manuals than is visible
So I search for `lift`, results on https://oscar-system.github.io/Oscar.jl/dev/search/?q=lift

Second hit is [lift (method, Univariate polynomials)](https://oscar-system.github.io/Oscar.jl/dev/Nemo/polynomial/#lift-Tuple{FmpzPolyRing,%20gfp_poly})

This page is *not* linked to from the TOC. It also states at the top:
> Nemo allow the creation of dense

So we probably don't want this `.html` file in there.

lgoettgens:
Resolves #643.

With this change, only the files occurring in `doc.main` or `doc_additional.main` will be imported from the Hecke/Nemo/AbstractAlgebra docs before building the Oscar docs.
`doc.main` stays as it is currently, representing the index structure.
`doc_additional.main` (name subject to change) contains a list of additional files that need to be copied, i.e. images or doc pages that are referenced.

Overall, this should lead to better search results in the Oscar docs, as we can remove some weird Hecke/Nemo/AbstractAlgebra pages from appearing (cf. #643).

For my convenience, this includes the changes from #2380. Once that PR is merged, I will rebase this one onto master.

### DIFF
--- a/docs/doc_additional.main
+++ b/docs/doc_additional.main
@@ -1,0 +1,4 @@
+[
+    # Here, one can add documentation files that should not be part of the
+    # index, but included in the documentation, e.g., images.
+]
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -106,6 +106,12 @@ function setup_experimental_package(Oscar::Module, package_name::String)
   return result
 end
 
+# Lists all files in `doc`, recursively.
+list_doc_files(doc::String) = doc
+list_doc_files(doc::Pair{String, String}) = list_doc_files(doc.second)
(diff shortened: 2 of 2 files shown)


lgoettgens:
The doc fails are more or less expected since many existing references to other AA/Hecke/Nemo pages do not exist.

How do we handle those?

lgoettgens:
This does not seem to lead to a valid solution. Thus closing

PR #2382, opened 2023-05-12
[FTheoryTools] Hypersurface model

HereAround:
This implements the next class of models for FTheoryTools, namely hypersurface models. For now, this works only over a concrete base space. Generalization to arbitary base spaces hopefully soon.

cc @apturner 

(As discussed on slack: The grading of the auxiliary base space currently does not know what divisor classes of the actual base space are reflected. For instance, if we have a Z^2 grading, then the first Z could refer to Kbar and the second to a gauge divisor class W. But it could also be the other way around. Or could even be 2 Kbar - 3 W and 3 Kbar + 4 W. Some places in the code assume that it is always of the form (Kbar, W1, W2, ...) and will form divisor classes [2,0,...] and [3,0,...] for the construction of the auxiliary ambient space. We should think about this/add consistency checks/another field to the types and possibly also reflect this in the literature models...)

### DIFF
--- a/docs/oscar_references.bib
+++ b/docs/oscar_references.bib
@@ -999,6 +999,21 @@
   url           = {https://doi.org/10.1007/978-3-030-52200-1_37}
 }
 
+@Article{KM-POPR15,
+  author        = {Klevers, Denis and Mayorga Pena, Damian Kaloni and Oehlmann, Paul-Konstantin and Piragua, Hernan and
+                  Reuter, Jonas},
+  title         = {F-Theory on all Toric Hypersurface Fibrations and its Higgs Branches},
+  journal       = {JHEP},
+  volume        = {01},
+  pages         = {142},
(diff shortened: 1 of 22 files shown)


HereAround:
I just split off the part of this PR that is likely not working yet (and will take me at least several weeks to fix, as I currently see it). Those additional changes are here: https://github.com/oscar-system/Oscar.jl/pull/2470

With these changes being separate, I hope that this PR here can be moved forward/merged shortly.

HereAround:
@apturner What do you think of this PR?

apturner (approved):
This looks great @HereAround; thanks for splitting this into pieces so that we can move ahead with the bits that are already working!

HereAround:
> This looks great @HereAround; thanks for splitting this into pieces so that we can move ahead with the bits that are already working!

Thank you @apturner.

PR #2383, opened 2023-05-12
Fix the order in `doc_init`

thofma:
The current order is wrong in view of the package version resolver.

@fieker


thofma:
Maybe sometimes one wants the other order. I will close this for now. Good luck to whoever runs into weird version problems with `doc_init` in the future.

PR #2386, opened 2023-05-12
Move test dependencies from `Project.toml`'s `[extra]` to `test/Project.toml`

### LINKED ISSUE #2374: Move test dependencies from `Project.toml`'s `[extra]` to `test/Project.toml`
Brought up by @fingolfin in https://github.com/oscar-system/Oscar.jl/pull/2370#discussion_r1190032346.

This issue is there to remind me to work on that.

Objections, comments, etc., are warmly welcome.

lgoettgens:
As originally requested by @fingolfin in https://github.com/oscar-system/Oscar.jl/pull/2370#discussion_r1190032346.

For the reasoning behind, please refer to [the Pkg docs](https://pkgdocs.julialang.org/v1.9/creating-packages/#test/Project.toml-file-test-specific-dependencies).
tl;dr: The current way using `[extra]` is pre-julia-1.2 compatible. The proposed way has been introduced in julia-1.2.

Closes #2374.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,6 @@ cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 [compat]
 AbstractAlgebra = "0.29.3"
 AlgebraicSolving = "0.3.0"
-Aqua = "0.6"
 DocStringExtensions = "0.8, 0.9"
 GAP = "0.9.4"
 Hecke = "0.18.9"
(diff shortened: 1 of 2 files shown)


fieker:
I expect this is going to run tests locally, as in `Oscar.test_nodule("bla", false)` harder, which I'd regard as a major problem....

lgoettgens:
> I expect this is going to run tests locally, as in `Oscar.test_nodule("bla", false)` harder, which I'd regard as a major problem....

For building the docs, there already is `docs/Project.toml` which works similarly. And this does not seem to bring up issues with `Oscar.build_docs`, right?

fingolfin:
I already explained to @fieker in a meeting a bit earlier today that I don't think `Oscar.test_module` will be affected by this change at all (or, for that matter, anything else that one does -- except that things get *better* in the sense that now one can do `]activate test` to be able to instantiate and access the packages needed for the test suite)

fingolfin (approved):
Looks good to me

PR #2387, opened 2023-05-12
Reduction mod p for schemes

HechtiDerLachs:
@simonbrandhorst : Have a look at the test file, please. Do you think this goes in the right direction? If yes, I will extend to `CoveredScheme`s.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
@@ -138,6 +138,9 @@ and a ring homomorphism ``φ : 𝕜 → 𝕂`` this returns a triple
     ↑ b₁          ↑ b₂
   X×ₖSpec(𝕂) → Y×ₖSpec(𝕂)
              F
+
+The optional arguments `domain_map` and `codomain_map` can be used 
+to specify the morphisms `b₁` and `b₂`, respectively. 
 ```
 """
 function base_change(phi::Any, f::AbsSpecMor; 
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
@@ -124,9 +124,9 @@ end
 ########################################################################
 
 @doc raw"""
-    change_base(phi::Any, f::AbsSpecMor)
-        domain_map::AbsSpecMor=change_base(phi, domain(f))[2],
-        codomain_map::AbsSpecMor=change_base(phi, codomain(f))[2]
+    base_change(phi::Any, f::AbsSpecMor)
+        domain_map::AbsSpecMor=base_change(phi, domain(f))[2],
+        codomain_map::AbsSpecMor=base_change(phi, codomain(f))[2]
       )
 
 For a morphism ``f : X → Y`` between two schemes over a `base_ring` ``𝕜`` 
@@ -140,9 +140,9 @@ and a ring homomorphism ``φ : 𝕜 → 𝕂`` this returns a triple
              F
 ```
 """
-function change_base(phi::Any, f::AbsSpecMor; 
-    domain_map::AbsSpecMor=change_base(phi, domain(f))[2],
-    codomain_map::AbsSpecMor=change_base(phi, codomain(f))[2]
+function base_change(phi::Any, f::AbsSpecMor; 
+    domain_map::AbsSpecMor=base_change(phi, domain(f))[2],
+    codomain_map::AbsSpecMor=base_change(phi, codomain(f))[2]
   )
   X = domain(f)
   Y = codomain(f)
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
@@ -119,3 +119,50 @@ function Base.show(io::IO, f::AbsSpecMor)
   print(io, "$(pullback(f)(last(x)))")
 end
 
+########################################################################
+# (6) Base change
+########################################################################
+
+@doc raw"""
+    change_base(phi::Any, f::AbsSpecMor)
+        domain_map::AbsSpecMor=change_base(phi, domain(f))[2],
+        codomain_map::AbsSpecMor=change_base(phi, codomain(f))[2]
+      )
+
+For a morphism ``f : X → Y`` between two schemes over a `base_ring` ``𝕜`` 
+and a ring homomorphism ``φ : 𝕜 → 𝕂`` this returns a triple 
+`(b₁, F, b₂)` consisting of the maps in the commutative diagram 
+```
+             f
+    X        →    Y
+    ↑ b₁          ↑ b₂
+  X×ₖSpec(𝕂) → Y×ₖSpec(𝕂)
+             F
+```
+"""
+function change_base(phi::Any, f::AbsSpecMor; 
+    domain_map::AbsSpecMor=change_base(phi, domain(f))[2],
+    codomain_map::AbsSpecMor=change_base(phi, codomain(f))[2]
+  )
+  X = domain(f)
+  Y = codomain(f)
+  XX = domain(domain_map)
+  YY = domain(codomain_map)
+  pbf = pullback(f)
+  pb1 = pullback(domain_map)
+  pb2 = pullback(codomain_map)
+
+  R = OO(Y)
+  S = OO(X)
+  RR = OO(YY)
+  SS = OO(XX)
+
+  img_gens = [pb1(pbf(x)) for x in gens(R)]
+  # For the pullback of F no explicit coeff_map is necessary anymore 
+  # since both rings in domain and codomain have the same (extended/reduced)
+  # coefficient ring by now.
+  pbF = hom(RR, SS, img_gens, check=true) # TODO: Set to false after testing
+
+  return domain_map, SpecMor(XX, YY, pbF, check=true), codomain_map # TODO: Set to false after testing
+end
+
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -157,8 +157,12 @@ function ambient_space(X::AbsSpec{BRT, RT}) where {BRT, RT<:MPolyRing}
   return X
 end
 
-@attr function ambient_space(X::Spec{BRT,RT}) where {BRT, RT <: Union{MPolyQuoRing,MPolyLocRing,MPolyQuoLocRing}}
+@attr function ambient_space(X::Spec{BRT,RT}) where {BRT<:Field, RT <: Union{MPolyQuoRing,MPolyLocRing,MPolyQuoLocRing}}
   return affine_variety(Spec(ambient_coordinate_ring(X)), check=false)
+end
+
+@attr function ambient_space(X::Spec{BRT,RT}) where {BRT, RT <: Union{MPolyQuoRing,MPolyLocRing,MPolyQuoLocRing}}
+  return Spec(ambient_coordinate_ring(X))
 end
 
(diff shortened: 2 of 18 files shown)


src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
```diff
@@ -199,6 +199,67 @@ function components(X::AbsSpec)
...
 end
 
+########################################################################
+# Base change and reduction modulo p
+########################################################################
+
+@doc raw"""
+    change_base(phi::Any, X::AbsSpec)
+
+For an affine scheme `X` over a `base_ring` ``𝕜`` and a morphism
```
simonbrandhorst, line 209:
```suggestion
For an affine scheme `X` over a `base_ring` ``𝕜`` and a morphism 
```
The notation `k` and `K` suggests a field. Is `B` or `R` an option?

simonbrandhorst: We have the name `change_base_ring` already and apparently also the function `base_ring` so that name seems appropriate to me? Less names to remember. And it would be consistent with our current use.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
```diff
@@ -199,6 +199,67 @@ function components(X::AbsSpec)
...
+########################################################################
+
+@doc raw"""
+    change_base(phi::Any, X::AbsSpec)
+
+For an affine scheme `X` over a `base_ring` ``𝕜`` and a morphism 
+``φ : 𝕜 → 𝕂`` this computes ``Y = X × Spec(𝕂)`` and returns a pair 
+`(Y, psi)` where `psi` is the canonical map ``Y → X``.
+"""
+function change_base(phi::Any, X::AbsSpec)
```
simonbrandhorst: No need to annotate `::any`  .

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
```diff
@@ -199,6 +199,67 @@ function components(X::AbsSpec)
...
+  R = OO(X)
+  R_red, Phi = _change_base_ring(phi, R)
+  Y = Spec(R_red)
+  return Y, SpecMor(Y, X, Phi)
+end
+
+### Some helper functions
+function _change_base_ring(phi::Any, R::MPolyRing)
+  K = coefficient_ring(R)
+  kk = parent(phi(zero(K)))
```
simonbrandhorst, line 225: This looks curious. Why not do `parent(codomain(phi))` ? 
You do not expect phi to have a codomain?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
simonbrandhorst: A less rigorous but common situation is when we have a scheme over `QQ` and want to reduce that mod `p`  and get an error if we divide by zero.  This depends on the chosen generators for the ideal involved of course.
Looks to me like this code can be used for this situation as well?

simonbrandhorst (commented):
I think this goes in the right direction.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
HechtiDerLachs: There is a discussion on this on Slack. I'd be open to make these methods of the already existing function `change_base_ring`, but it's not yet completely clear whether we want that. 

@thofma : Any suggestions?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
HechtiDerLachs: I know. But it makes it clear that the programmer really meant this and was not just too lazy to spell out the annotation and then later forgot about it.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
HechtiDerLachs: I took this from what we did for maps from polynomial rings: If you, for example, also wish to use coercions or anonymous functions, you need this flexibility and can not expect `phi` to have a `domain` and/or `codomain.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
HechtiDerLachs: > A less rigorous but common situation is when we have a scheme over `QQ` and want to reduce that mod `p` and get an error if we divide by zero. This depends on the chosen generators for the ideal involved of course.
> Looks to me like this code can be used for this situation as well?

That is a good example for a case where you really can not have `phi` as a valid `Hecke.Map` (with domain and codomain as discussed below). Yes, this code should be able to accomodate that. You just have to pass some function that does the job for you to try to reduce mod p and throw an error if something goes wrong.

simonbrandhorst: O.k. It seems that `base_change` exists already for elliptic and hyperelliptic curve. It is also used in the literature. So we can stick with that, especially if `change_base_ring` is a little controversial. 
So please go ahead with this ticket.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
simonbrandhorst: Then you could document, that it also takes any kind of function on the coefficients?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
thofma: No suggestions.

I think `base_change` is also more appropriate if we have relative schemes and want to change the base for those?

HechtiDerLachs: But should it then be `base_change` or `change_base`?

simonbrandhorst: Looks like `saturation` versus `saturate`. We are not globally consistent here. Is there any guideline?
I would go for `base_change`, also because that already exists. But I do not have a strong opinion.

HechtiDerLachs:
This should be ready to go as a first version from my side. Let's see if the tests go through. Then I'll take care of switching off some internal tests and eventually rename everything to `base_change`. But you can already go through the code to complain about other stuff.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -119,3 +119,50 @@ function Base.show(io::IO, f::AbsSpecMor)
...
+
+For a morphism ``f : X → Y`` between two schemes over a `base_ring` ``𝕜`` 
+and a ring homomorphism ``φ : 𝕜 → 𝕂`` this returns a triple 
+`(b₁, F, b₂)` consisting of the maps in the commutative diagram 
+```
+             f
+    X        →    Y
+    ↑ b₁          ↑ b₂
+  X×ₖSpec(𝕂) → Y×ₖSpec(𝕂)
+             F
```
simonbrandhorst, line 140: Does this render o.k. in the documentation?

simonbrandhorst: Please add a sentence about optional arguments.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -157,10 +157,14 @@ function ambient_space(X::AbsSpec{BRT, RT}) where {BRT, RT<:MPolyRing}
   return X
 end
 
-@attr function ambient_space(X::Spec{BRT,RT}) where {BRT, RT <: Union{MPolyQuoRing,MPolyLocRing,MPolyQuoLocRing}}
+@attr function ambient_space(X::Spec{BRT,RT}) where {BRT<:Field, RT <: Union{MPolyQuoRing,MPolyLocRing,MPolyQuoLocRing}}
```
simonbrandhorst, line 160: Thanks for catching this.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
```diff
@@ -65,3 +65,33 @@ function maps_with_given_codomain(f::AbsCoveredSchemeMorphism, V::AbsSpec)
   return result
 end
 
+########################################################################
+# Comparison
+########################################################################
+function ==(f::AbsCoveredSchemeMorphism, g::AbsCoveredSchemeMorphism)
```
simonbrandhorst, line 71: If we have `==`, I think we need a hash function?

src/AlgebraicGeometry/Schemes/Glueing/Methods.jl
```diff
@@ -124,3 +124,48 @@ function ==(G::AbsGlueing, H::AbsGlueing)
   return true
 end
 
+########################################################################
+# Base change
+########################################################################
+struct BaseChangeGlueingData{T1, T2}
```
simonbrandhorst, line 130:
```suggestion
struct BaseChangeGlueingData{T1, T2}  # used for LazyGlueing
```

test/AlgebraicGeometry/Schemes/Glueing.jl
```diff
@@ -122,3 +122,13 @@ end
   @test glueing_domains(G1) == glueing_domains(DG)
   @test inverse(DG) == DG
 end
+
+@testset "base change" begin
+  kk, pr = quo(ZZ, 5)
+  IP1 = covered_scheme(projective_space(ZZ, 1))
+  C = default_covering(IP1)
+  G = first(values(glueings(C)))
+  GG = oscar.base_change(pr, G)
```
simonbrandhorst:
```suggestion
  GG = oscar.base_change(pr, G)
```
is the `oscar.` still needed?

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
HechtiDerLachs: I am not aware of this. Can you elaborate or give a link describing what has to be done here?

simonbrandhorst: https://github.com/oscar-system/Oscar.jl/issues/2222

simonbrandhorst (approved):

simonbrandhorst:
Could you please also add a test when one works over a non-trivial base and restricts to a fiber, i.e.
some base like $K[x,y]$ and then some base change to  $K[x]$ via $y \mapsto 15, x \mapsto x$ or similar.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
HechtiDerLachs: We spoke about this a bit today with @fieker . I'm not yet 100% sure how I want to address these issues and would like to leave this for a later cleanup. But thanks for notifying me on this.

HechtiDerLachs:
> Could you please also add a test when one works over a non-trivial base and restricts to a fiber, i.e.
> some base like

I probably could. However, our schemes are not really equipped with a backend to handle these kind of relative schemes. I don't think that translating something like `K[x,y]` to a coefficient ring in Singular works. And if it does, it's probably not effective. 

Edit: I could try to do something with my `RingFlattening`s, though... Just kidding. I suggest to leave this as is.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
simonbrandhorst: o.k.

PR #2388, opened 2023-05-15
Parallel Testing 

aaruni96:
Attempting to parallelize testing to save some time.


thofma:
I think @benlorenz mentioned to me once, that the CI github actions runner do not have enough memory to make this an option. (Which I can confirm. The Hecke tests are run in parallel only on OSX, because there is more memory available).

Edit: I think it would be more robust to just copy the setup of https://github.com/JuliaLang/julia/blob/master/test/runtests.jl  For example, `ctrl+c` will work properly (this is what we used to do in other projects).

aaruni96:
The motivation for this work was to speed up local testing (parts of my work require me to run `Pkg.test("Oscar")` very often).

I've tried to implement it based on an environment variable. If `NUMPROCS` >= 2, julia will spawn `NUMPROCS` many workers and distribute testing between them. If the environment variable is not set, or is set to 1, old behavior is preserved.

(I also expect currently queued CI actions to fail. Sorry, I forgot to add `[skip ci]` to the PR message.)

thofma:
OK, makes sense. I would suggest to also make `Pkg.test("Oscar", test_args = ["-j4"])` work. This is quite convenient in the REPL.

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.12.0"
 [deps]
 AbstractAlgebra = "c3fe647b-3220-5bb0-a1ea-a7954cac585d"
 AlgebraicSolving = "66b61cbe-0446-4d5d-9090-1ff510639f9d"
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
```
fingolfin: This should be moved into the (new!) `test/Project.toml` I guess?

fingolfin (commented):
This PR needs to be rebased on latest `master` to resolve a conflict.

Project.toml
```diff
@@ -6,6 +6,7 @@ version = "0.12.1-DEV"
...
```
lgoettgens:
```suggestion

```
This should not be needed here, only in `test/Project.toml`.

lgoettgens:
Should parallel testing be used in CI, or is it only meant for local execution of tests?

fingolfin:
Only for local execution, because the CI runners can't cope with the extra memory usage.

Project.toml
fingolfin: Indeed. Also, this unfortunately now needs to be rebased again

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,22 @@
     # to avoid unnecessary failures we skip the timing tests
     haskey(ENV, "JULIA_PKGEVAL") && return
 
-    using Oscar
+    # If running multiple workers, and many OMP threads, CPU oversubscription may occur
+    # which fails the timing tests. In this situation, skip instead.
+    try
```
fingolfin: Why this `try`, there is nothing in it that could throw an exception, is there?

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,22 @@
...
+        if (length(workers())>1)
```
fingolfin:
```suggestion
        if length(workers())>1
```

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,22 @@
...
+            if haskey(ENV, "OMP_NUM_THREADS")
+                if (ENV["OMP_NUM_THREADS"] != "1")
+                    return
+                end
+            else
+                return
+            end
```
fingolfin: could also do this:
```suggestion
            if get(ENV, "OMP_NUM_THREADS", "") != "1"
                return
            end
```

test/runtests.jl
```diff
@@ -1,33 +1,53 @@
...
+if !isempty(ARGS)
+  jargs = [arg for arg in ARGS if startswith(arg, "-j")]
+  if !isempty(jargs)
+    numprocs = parse(Int,split(jargs[end], "-j")[end])
+  end
+elseif haskey(ENV, "NUMPROCS")
+  numprocs = parse(Int,ENV["NUMPROCS"])
+else
+  numprocs = 1
+end
```
fingolfin: So if `ARGS` is non-empty but contains no `-j` flag, then `numprocs` is left uninitialized...

Here is an alternative that avoids this.
```suggestion
numprocs_str = get(ENV, "NUMPROCS", "1")
if !isempty(ARGS)
  jargs = [arg for arg in ARGS if startswith(arg, "-j")]
  if !isempty(jargs)
    numprocs_str = split(jargs[end], "-j")[end]
  end
end

const numprocs = parse(Int, numprocs_str)
```

test/runtests.jl
```diff
@@ -1,33 +1,53 @@
...
+
+if (numprocs >= 2)
```
fingolfin:
```suggestion
if numprocs >= 2
```

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,22 @@
...
+        end
```
lgoettgens:
```suggestion
        if (length(workers()) > 1)
            haskey(ENV, "OMP_NUM_THREADS") || return
            ENV["OMP_NUM_THREADS"] == "1" || return
        end
```
This would be easier to read and more idiomatic.

test/StraightLinePrograms/runtests.jl
```diff
@@ -2,6 +2,7 @@
 module SLPTest
 using ..Test
 using ..Oscar
+const SLP = Oscar.StraightLinePrograms
 using ..Oscar: SLP
```
lgoettgens:
```suggestion

```
This is basically a duplicate of the line above it.

test/runtests.jl
```diff
@@ -82,62 +102,73 @@ end
...
-include("AlgebraicGeometry/Surfaces/K3Auto.jl")
+push!(testlist, "Serialization/runtests.jl")
 
-include("Serialization/runtests.jl")
+push!(testlist, "StraightLinePrograms/runtests.jl")
 
-include("StraightLinePrograms/runtests.jl")
+# if many workers, distribute tasks across them
+# otherwise, is essentially a serial loop
+  @time pmap(x -> include(x), testlist)
```
lgoettgens:
```suggestion
@time pmap(x -> include(x), testlist)
```

test/runtests.jl
```diff
@@ -82,62 +102,73 @@ end
...
 
 @static if compiletimes
   Base.cumulative_compile_timing(false);
 end
 
-if haskey(ENV, "GITHUB_STEP_SUMMARY") && compiletimes
-  open(ENV["GITHUB_STEP_SUMMARY"], "a") do io
-    print_stats(io, fmt=PrettyTables.tf_markdown)
+# Doctests
```
lgoettgens:
```suggestion

```
There are no doctests to be run in this file.

test/PolyhedralGeometry/timing.jl
thofma: This should not be a `try catch`. This will just hide real problems.

aaruni96: If `using Distributed()` hasn't been run already (for example, in the case of  `Oscar.test_module("PolyhedralGeometry/timing")`), then `workers()` throws an `UndefVarError`, and halt execution. This try catch avoids that, and shouldn't catch anything else.

Maybe to be more explicit with what we catch, we can try

```
...
catch e
    if e != UndefVarError(:workers)
        throw(e)
    end
end
```

thofma: You can check that by invoking `isdefined`.

aaruni96:
Hello, I think I have addressed all comments thus far.

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,12 @@
     # to avoid unnecessary failures we skip the timing tests
     haskey(ENV, "JULIA_PKGEVAL") && return
 
-    using Oscar
+    # If running multiple workers, and many OMP threads, CPU oversubscription may occur
+    # which fails the timing tests. In this situation, skip instead.
+    if (isdefined(Main, :Distributed) && (length(workers()) > 1))
+            haskey(ENV, "OMP_NUM_THREADS") || return
+            ENV["OMP_NUM_THREADS"] == "1" || return
```
fingolfin:
```suggestion
        haskey(ENV, "OMP_NUM_THREADS") || return
        ENV["OMP_NUM_THREADS"] == "1" || return
```

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,12 @@
     # to avoid unnecessary failures we skip the timing tests
     haskey(ENV, "JULIA_PKGEVAL") && return
 
-    using Oscar
+    # If running multiple workers, and many OMP threads, CPU oversubscription may occur
+    # which fails the timing tests. In this situation, skip instead.
+    if (isdefined(Main, :Distributed) && (length(workers()) > 1))
```
fingolfin:
```suggestion
    if isdefined(Main, :Distributed && length(workers()) > 1)
```

test/runtests.jl
```diff
@@ -80,63 +101,72 @@ end
...
   Base.cumulative_compile_timing(false);
 end
 
-if haskey(ENV, "GITHUB_STEP_SUMMARY") && compiletimes
-  open(ENV["GITHUB_STEP_SUMMARY"], "a") do io
-    print_stats(io, fmt=PrettyTables.tf_markdown)
+#currently, print_stats will fail when running tests with external workers
+#TODO: potentially rewrite include as well as print_stats 
+#      to comply with parallel decisions
+if (numprocs == 1)
```
fingolfin:
```suggestion
if numprocs == 1
```

test/runtests.jl
```diff
@@ -80,63 +101,72 @@ end
...
-include("TropicalGeometry/runtests.jl")
+push!(testlist, "Serialization/runtests.jl")
 
-include("Serialization/runtests.jl")
+push!(testlist, "StraightLinePrograms/runtests.jl")
 
-include("StraightLinePrograms/runtests.jl")
+# if many workers, distribute tasks across them
+# otherwise, is essentially a serial loop
+@time pmap(x -> include(x), testlist)
```
fingolfin:
```suggestion
pmap(include, testlist)
```
Or did you want to use `@test` in here on purpose?

test/runtests.jl
```diff
@@ -80,63 +101,72 @@ end
   Base.cumulative_compile_timing(true)
 end
 
-# Used in both Rings/slpolys.jl and StraightLinePrograms/runtests.jl
-const SLP = Oscar.StraightLinePrograms
+println("Making test list")
 
-include("Aqua.jl")
+testlist = []
```
fingolfin:
```suggestion
const testlist = []
```

fingolfin: Actually, on second thought, perhaps instead of using lots of `push!`, we should just do this:
```
const testlist = [
  "Aqua.jl",

  "printing.jl",

  "PolyhedralGeometry/runtests.jl"
  "Combinatorics/runtests.jl",
  ...
]
```

test/runtests.jl
```diff
@@ -80,63 +101,72 @@ end
...
 
-include("Rings/runtests.jl")
+push!(testlist, "Rings/runtests.jl")
 
-include("NumberTheory/nmbthy.jl")
+push!(testlist, "NumberTheory/nmbthy.jl")
 
 if Oscar.is_dev
-  include("Experimental/GITFans.jl")
+  push!(testlist, "Experimental/GITFans.jl")
```
fingolfin: This conditional `push!` would be an obstacle to declaring `const testlist = [ ... ]`, but luckily it will be removed in PR #2416.

fingolfin (commented):
Sorry, I started a review yesterday and then forgot to submit it, so I am doing that now (with a few more additions)

test/runtests.jl
```diff
@@ -1,31 +1,63 @@
...
+if !isempty(ARGS)
+  jargs = [arg for arg in ARGS if startswith(arg, "-j")]
+  if !isempty(jargs)
+    numprocs = parse(Int,split(jargs[end], "-j")[end])
+  end
+elseif haskey(ENV, "NUMPROCS")
+  numprocs = parse(Int,ENV["NUMPROCS"])
+else
+  numprocs = 1
+end
```
fingolfin: This is redundant now and must be removed (actually I don't understand how this even passes CI, as we have `const numprocs = ..` a few lines above?!?)
```suggestion
```

test/PolyhedralGeometry/timing.jl
```diff
@@ -4,7 +4,11 @@
     # to avoid unnecessary failures we skip the timing tests
     haskey(ENV, "JULIA_PKGEVAL") && return
 
-    using Oscar
+    # If running multiple workers, and many OMP threads, CPU oversubscription may occur
+    # which fails the timing tests. In this situation, skip instead.
+    if isdefined(Main, :Distributed) && (length(workers()) > 1)
```
fingolfin:
```suggestion
    if isdefined(Main, :Distributed) && nworkers() > 1
```

test/runtests.jl
aaruni96: I just tried this in repl, and apparently redefining a constant is only a warning, not an error..

`WARNING: redefinition of constant numprocs. This may fail, cause incorrect answers, or produce other errors.`

aaruni96: Some further reading reveals that right now in Julia, `const` only fixes the type of the variable and errors if you assign it a value which would change its type. If you assign it a different value of the same type, its undefined behavior, which results in a warning and continues.

```
julia> const a = 5
5

julia> a = 7
WARNING: redefinition of constant a. This may fail, cause incorrect answers, or produce other errors.
7

julia> a = 7.0
ERROR: invalid redefinition of constant a
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1

```

fingolfin (approved):

PR #2396, opened 2023-05-19
[ToricVarieties] Bugfixes in toric line bundles

HereAround:
While working on `FTheoryTools`, I wanted to work with the anticanonical bundle of a particular toric space, so I tried:

```julia
v = underlying_toric_variety(ambient_space(global_weierstrass_model_over_projective_space(3)))
candidate1 = prod(toric_line_bundle(v, d) for d in torusinvariant_prime_divisors(v))
```

To my surprise, this method did not succeed because one torusinvariant prime divisor (actually, there are two in this case) are not Cartier. Indeed, here is the current implementation is as follows:
```julia
@attr ToricLineBundle anticanonical_bundle(v::AbstractNormalToricVariety) = prod(toric_line_bundle(v, d) for d in torusinvariant_prime_divisors(v))
```
So this only works, if every torusinvariant prime divisor is Cartier. But this is not a given.

Rather, by CLS11 it is guaranteed that the sum of all torusinvariant prime divisors is Cartier. And so, this implementation should be changed as follows:

```julia
@attr ToricLineBundle anticanonical_bundle(v::AbstractNormalToricVariety) = toric_line_bundle(v, anticanonical_divisor_class(v))
```
Sadly, this did not work because no method was implemented that would allow to define a toric variety by a divisor class (in the class group).

The third approach that I tried was then the following:
```julia
candidate3 = toric_line_bundle(v, anticanonical_divisor(v))
```
This finally managed to construct the bundle. It is worth noting that the anticanonical divisor in this case has degrees [0,6] and so the global sections are all monomials with this multidegree in the cox ring:
```julia
julia> cox_ring(v)
Multivariate polynomial ring in 7 variables over QQ graded by 
  x1 -> [1 0]
  x2 -> [1 0]
  x3 -> [1 0]
  x4 -> [1 0]
  x -> [0 2]
  y -> [0 3]
  z -> [-4 1]
```
Clearly, there are many such monomials (x^3, y^2, ...). But still...
```julia
julia> all_cohomologies(candidate3)
  6-element Vector{ZZRingElem}:
   0
   0
   0
   35101
   0
   0
```
The reason for this last failure is that the following distinct notions were mixed up in parts of the code:
1. The divisor class in the class group of the Cartier toric divisor that defines the line bundle in question.
2. The element in the Picard group that corresponds to the line bundle.

Specifically, in the case at hand the Picard group is one to one to the toric divisor classes with multidegrees `[a, 6b]`. In particular, the anticanonical bundle is associated to the anticanonical divisor `[0,6]`. Indeed, the image of the mapping of Cartier toric divisors (isomorphic to Z^7 here) to the Picard group (here isomorphic to Z^2) has an image embedding given by the matrix
```julia
[1 0]
[0 6]
```
We use the method `restrict_codomain` to restrict this mapping to its image. This "chooses" to represent the image of the anticanonical bundle of `v` to be the element `[0, 1]` in the Picard group.

In summary, I stumbled upon a combination of the following three issues:
1. Wrong implementation of the anticanonical bundle.
2. Missing implementation to create a toric line bundle from a divisor class (in the class group).
4. Mixing up of elements in the class group and in the Picard group.

This PR aims to fix all of this.

### DIFF
--- a/docs/src/AlgebraicGeometry/ToricVarieties/ToricLineBundles.md
+++ b/docs/src/AlgebraicGeometry/ToricVarieties/ToricLineBundles.md
@@ -11,10 +11,12 @@ CurrentModule = Oscar
 ### Generic constructors
 
 ```@docs
-toric_line_bundle(v::AbstractNormalToricVariety, class::GrpAbFinGenElem)
-toric_line_bundle(v::AbstractNormalToricVariety, c::Vector{T}) where {T <: IntegerUnion}
+toric_line_bundle(v::AbstractNormalToricVariety, picard_class::GrpAbFinGenElem)
+toric_line_bundle(v::AbstractNormalToricVariety, picard_class::Vector{T}) where {T <: IntegerUnion}
 toric_line_bundle(v::AbstractNormalToricVariety, d::ToricDivisor)
 toric_line_bundle(d::ToricDivisor)
+toric_line_bundle(v::AbstractNormalToricVariety, dc::ToricDivisorClass)
+toric_line_bundle(dc::ToricDivisorClass)
 ```
 
 ### Tensor products
@@ -49,8 +51,9 @@ is_very_ample(l::ToricLineBundle)
 
 ```@docs
 degree(l::ToricLineBundle)
-divisor_class(l::ToricLineBundle)
+picard_class(l::ToricLineBundle)
 toric_divisor(l::ToricLineBundle)
+toric_divisor_class(l::ToricLineBundle)
 toric_variety(l::ToricLineBundle)
 ```
 
--- a/src/AlgebraicGeometry/ToricVarieties/ToricLineBundles/attributes.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/ToricLineBundles/attributes.jl
@@ -3,9 +3,9 @@
 #####################
 
 @doc raw"""
-    divisor_class(l::ToricLineBundle)
+    picard_class(l::ToricLineBundle)
 
-Return the divisor class which defines the toric line bundle `l`.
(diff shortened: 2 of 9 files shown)


fingolfin:
```
ERROR: LoadError: LoadError: LoadError: `map_from_cartier_divisor_group_to_torusinvariant_divisor_group(v::AbstractNormalToricVariety)` is deprecated, use `map_from_torusinvariant_cartier_divisor_group_to_torusinvariant_weil_divisor_group(v)` instead.
```

HereAround:
> ```
> ERROR: LoadError: LoadError: LoadError: `map_from_cartier_divisor_group_to_torusinvariant_divisor_group(v::AbstractNormalToricVariety)` is deprecated, use `map_from_torusinvariant_cartier_divisor_group_to_torusinvariant_weil_divisor_group(v)` instead.
> ```

Thank you. Just fixed.

HereAround:
I dont understand why one test fails. @lkastner @fingolfin any idea? Is there a time-out for an individual test file (here roughly 55s)? In the log it just says that it was terminated with signal 15 - but what does that mean?

HereAround:
> I dont understand why one test fails. @lkastner @fingolfin any idea? Is there a time-out for an individual test file (here roughly 55s)? In the log it just says that it was terminated with signal 15 - but what does that mean?

Seems to have indeed been a timeout problem. I have slightly modified the tests, so that they pass quicker and... all tests did pass.

fingolfin (approved):
I cannot really comment on the content of this other than from a surface level, where it all seems fine.

Dunno if you would like to get a review from @lkastner first, but from my POV you could also just merge it?

lkastner (approved):

PR #2397, opened 2023-05-19
Cleanup some using statements

fingolfin:


test/runtests.jl
```diff
@@ -1,7 +1,5 @@
 using Oscar
 using Test
-using Aqua
-using Documenter
```
lgoettgens, line 4: Could you additionally remove `Documenter` from `test/Project.toml` to make it consistent?

fingolfin: Done. Thanks!

lgoettgens (approved):
looks good to me

PR #2398, opened 2023-05-22
generalize `matrix_action_on_cones`

ThomasBreuer:
changed `matrix_action_on_cones` such that it works also for cones with lineality


fingolfin (approved):
Looks good to me on the surface level -- I have no idea how the cones work so I can't really comment on the code. But since it seems to work I am happy to trust @ThomasBreuer that this is either correct, or if not, will be fixed ;-)

PR #2401, opened 2023-05-22
Disable superfluous internal checks systematically. 

HechtiDerLachs:
As noted by @simonbrandhorst some code for resolutions of singularities of elliptic fibrations was horribly slow. Today, I have been going through the code and systematically disabled all superfluous internal checks that were triggered by our test file. 

To this end, I introduced a flag called `CHECK_ERROR`. I will point out the description in the source code below. This is probably not in its optimal form, but it was the best I could do for the moment. And it has worked quite well for me so far. 

@thofma @fingolfin : Could you have a look at this design and let me know what you think? Maybe one should solve this differently, but after what I saw today, I think we definitely need this functionality in one way or the other. 

@thofma : I also changed some parts in your code on the maps from polynomial rings. Have a look please, whether you approve.


src/AlgebraicGeometry/Schemes/main.jl
```diff
@@ -147,3 +147,30 @@ include("ProjectiveVariety/Objects/Constructors.jl")
...
+#
+# Every function using the optional `check::Bool=true` argument should have the line 
+#
+#   check && CHECK_ERROR && error("internal checks were enabled")
+#
+# Then the developer can set the flag below to `true` and run their test code. This will 
+# throw the error whenever an unexpected internal check would have been triggered together 
+# with a subsequent stacktrace. Then the internal checks can be switched off. 
+
+CHECK_ERROR = false
```
HechtiDerLachs: This is the flag I was talking about.

simonbrandhorst: Surely there is a better place to put this?
My impression is that this makes sense for prohibitively expensive checks in the schemes world. 
But not as a general pattern in Oscar.

src/AlgebraicGeometry/Schemes/main.jl
```diff
@@ -147,3 +147,30 @@ include("ProjectiveVariety/Objects/Constructors.jl")
...
+# well defined, one has to check that the lift F : P -> R of this map to the polynomial 
+# ring P with A = P/I takes all elements in I to zero. In the default case, the constructor 
+# for f will check that the given map is well defined, but one does not want this check 
+# to be run when calling the constructor internally on verified sane input. 
+#
+# To this end, one should do the function calls with `check=false` internally. However, 
+# it is very likely that one forgets to include this statement in all calls when programming. 
+# To this end, we introduce a global flag here which does the following. 
+#
+# Every function using the optional `check::Bool=true` argument should have the line
```
simonbrandhorst:
```suggestion
# A function using the optional `check::Bool=true` argument can have the line 
```

simonbrandhorst:
Disabling the checks is fine.  Testbots fail.

src/AlgebraicGeometry/Schemes/main.jl
HechtiDerLachs: I thought so, too. But as I progressed, I saw this spreading more and more into the algebra section. So I do think this should become a more general thing, unfortunately. I just didn't know where to put it and wanted to get some opinions first.

thofma:
I understand what this is trying to do, but I must say, that all these `CHECK_ERROR && error("check was enabled")` look awful.

It should be `const CHECK_ERROR`.

HechtiDerLachs:
I know, it looks awful. I just needed something quick and dirty to work with today. We should change it, but I wanted to hear some opinions on how.

experimental/Schemes/Types.jl
```diff
@@ -147,13 +147,14 @@ end
     codomain(g) === X || error("map is not compatible")
 
     if check
+      CHECK_ERROR && error("check was enabled")
       is_identity_map(compose(f, g)) && is_identity_map(compose(g, f)) || error("maps are not inverse to each other")
```
afkafkafk13: I do not understand, why line 150 was introduced. Is this a leftover from searching all unnecessary tests?

simonbrandhorst: It is to facilitate catching unnecessary tests in the future.

afkafkafk13: ok. got it -- when I started reading the changes (and was interrupted many times), the discussion here had not yet started.

src/AlgebraicGeometry/Schemes/main.jl
simonbrandhorst:
```suggestion
const CHECK_ERROR = false
```

HechtiDerLachs: I think we can change this once people have agreed on how they want to address the "awfulness" of the `check && CHECK_ERROR && error(...)` line in general, whether all of this should be moved to another place, or whether we can leave it as is for now. No need to re-run all the tests in the meantime, unless they do something useful like heating a building which actually needs heating.

fieker:
I don't think this is a good idea at all. It is going to bite you terribly if seriously implemented "everywhere": you're going to set this to false globally and then goging to run, via illegal input to say number fields, into interesting errors everywhere.
We've tried that before...

fieker:
Talking to Mathias: I misunderstood. This idea is only to notify when input validation is happening with the aim to not use input validation internally

HechtiDerLachs:
I have added a new macro in `src/Assertions.jl` similar to what I found in Hecke. You can also use it without any arguments which might seem a bit odd at first. Then it only makes sure the error message is thrown when snooping for unnecessary internal checks. Let me know what you think.

src/Oscar.jl
```diff
@@ -151,7 +151,7 @@ function build()
   system("Build.jl")
 end
 
-
+include("Assertions.jl")
```
HechtiDerLachs: I was not sure where to put this. Let me know if there is a more appropriate place.

fingolfin: place is fine, but the filename should be lowercase

src/AlgebraicGeometry/Schemes/SpecOpen/Rings/Methods.jl
```diff
@@ -18,6 +18,7 @@ function restrict(
     V::AbsSpec{<:Ring, <:MPolyQuoLocRing};
     check::Bool=true
   )
+  @check
   check && isempty(V) && return zero(OO(V))
```
fingolfin: Huh, so that's weird: why does `check` control whether we test for empty `V` here, and lead to a different (??) return value? That looks as if it might be either an optimization (but then why is it restricted to `check == true`), or else a semantical change (but that would seem very undesirable). What am I missing?

experimental/Schemes/CoveredScheme.jl
```diff
@@ -424,7 +424,8 @@ image_ideal(phi::CoveredClosedEmbedding) = phi.I
 ### user facing constructors
 function CoveredClosedEmbedding(X::AbsCoveredScheme, I::IdealSheaf; 
         covering::Covering=default_covering(X), check::Bool=true)
-  space(I) == X || error("ideal sheaf is not defined on the correct scheme")
+  @check
```
fingolfin: This function just passes `check` on to other functions, so I don't think it needs `@check`, provided those functions use it?

experimental/Schemes/WeilDivisor.jl
```diff
@@ -33,6 +33,7 @@ stored as a formal linear combination over some ring ``R`` of
       coefficients::IdDict{<:IdealSheaf, CoefficientRingElemType};
       check::Bool=true
     ) where {CoefficientRingType, CoefficientRingElemType}
+    @check
     if check
```
fingolfin: I rather dislike these stand-alone `@check` invocations. I find them very hard to model right mentally. I would find it much more natural if this code read as
```suggestion
    @check do
```

experimental/Schemes/WeilDivisor.jl
```diff
@@ -100,7 +102,8 @@ weil_divisor(X::AbsCoveredScheme, R::Ring) = WeilDivisor(X, R)
 Return the `WeilDivisor` ``D = 1 ⋅ V(I)`` with coefficients 
 in ``R`` for a sheaf of prime ideals ``I``.
 """
-function WeilDivisor(I::IdealSheaf, R::Ring)
+function WeilDivisor(I::IdealSheaf, R::Ring; check::Bool=true)
+  @check
```
fingolfin: the `check` argument is completely ignored by this method, why does it have it then? (also, if you add it here, shouldn't it also be added to the docstring above?

experimental/Schemes/WeilDivisor.jl
```diff
@@ -114,7 +117,8 @@ weil_divisor(I::IdealSheaf, R::Ring) = WeilDivisor(I, R)
 Return the `WeilDivisor` ``D = 1 ⋅ V(I)`` with coefficients
 in ``ℤ`` for a sheaf of prime ideals ``I``.
 """
-function WeilDivisor(I::IdealSheaf)
+function WeilDivisor(I::IdealSheaf; check::Bool=true)
+  @check
```
fingolfin: again: `check` unused and not mentioned in docstring

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -10,6 +10,7 @@ Convert `X` to an `AffineAlgebraicSet` by taking the underlying reduced scheme.
 If `check=false`, assumes that `X` is already reduced.
 """
 function affine_algebraic_set(X::Spec; check::Bool=true)
+  @check
   if check
     Xred,_ = reduced_scheme(X)
```
fingolfin: It seems to me this is an incorrect use of `check`, or at least different from other uses of `check`: instead of checking that the input is valid, it extends the set of valid inputs by reducing the input first. This could be a source of bad bugs. So either `check` should check for whether `X` is reduced, and if not, raise an error; or else it should be renamed to something else.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -37,6 +38,7 @@ Vanishing locus
 ```
 """
 function vanishing_locus(I::MPolyIdeal{<:MPolyElem}; check::Bool=true)
+  @check
   if check
     Irad = radical(I)
```
fingolfin: Here again `check` does not validate the input but rather modifies behavior: not good

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Types.jl
```diff
@@ -1,6 +1,7 @@
 @attributes mutable struct AffineAlgebraicSet{BaseRing<:Field, RingType<:MPolyAnyRing} <: AbsAffineAlgebraicSet{BaseRing, RingType}
   X::Spec
   function AffineAlgebraicSet(X::Spec; check::Bool=true)
+    @check
     if check
       is_geometrically_reduced(X) || error("algebraic sets must be geometrically reduced")
     else
```
fingolfin: Another opportunity to avoid the "stand-alone" `@check`:
```suggestion
    @check is_geometrically_reduced(X) "algebraic sets must be geometrically reduced"
```
(and then adjust the following code, too)

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
```diff
@@ -151,6 +151,7 @@ function preimage(
                                                <:MPolyPowersOfElement}};
     check::Bool=true
   )
+  @check true
   X = domain(phi)
   Y = codomain(phi)
   check && (issubset(Z, Y) || (Z = intersect(Y, Z)))
```
fingolfin: Again: if `check` is true then arguments should be validated, not "made to fit"

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
```diff
@@ -164,6 +165,7 @@ function preimage(
 end
 
 function preimage(f::AbsSpecMor, Z::AbsSpec{<:Ring, <:MPolyRing}; check::Bool=true)
+  @check true
```
fingolfin: Is `@check true` any different from `@check`?

This is one of several methods which have unused `check` arguments. Why is that so? Perhaps this is meant to enable potential future extensions of this methods which use `check` after all?

src/Assertions.jl
```diff
@@ -0,0 +1,45 @@
...
+      THROW_ERROR_FOR_INTERNAL_CHECKS && error("internal checks are enabled")
+      @assert $(esc(cond))
+    end
+  end
+end
+
+macro check(cond, msg)
+  quote
+    if $(esc(:(check)))
+      THROW_ERROR_FOR_INTERNAL_CHECKS && error("internal checks are enabled")
```
lgoettgens: Is the compiler smart enough to optimize lines like this away, if the flag is `false`?
If not, would it be beneficial for performance to move the check outside, i.e. define different macros depending on the flag? Even if not necessary, I could imagine this being more readable.

src/AlgebraicGeometry/Schemes/SpecOpen/Rings/Methods.jl
afkafkafk13: You are missing that the emptyness test and the appropriate return value are part of the mathematics and the algorithm itself.
Testing isempty for non-empty V is not cheap (Groebner basis!), unless the value has already been cached. But depending on the setting, in which one calls the function, it is often inevitable to do this test. There are however numerous settings in which it is known beforehand from the construction that V is non-empty -- without having called isempty on V before. So check=false is an optimization here, but not in the direction that you were expecting.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
afkafkafk13: You are right, mathematically the wording for 'check' is misleading here. It should be something like 'bypass_tayloring' (corresponding to the value of !check). Again this opens the possibility to bypass a Groebner basis computation, if the input is already known to satisfy the condition.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
afkafkafk13: You are right, mathematically the wording for 'check' is misleading. It should be something like 'bypass_tayloring' (corresponding to the value of !check). This opens the possibility to bypass the rather expensive radical computation, if the input is already known to satisfy the condition.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
afkafkafk13: You are right, mathematically the wording for 'check' is misleading here. It should be something like 'bypass_tayloring' (corresponding to the value of !check).

thofma:
When I suggested the name `@check`, I thought we would use it as
```
@check blabla [optional error message]
```
I agree with @fingolfin that the single `@check` at the top of the function looks odd.

src/Oscar.jl
HechtiDerLachs: It was uppercase in Hecke. But I can change it.

src/AlgebraicGeometry/Schemes/SpecOpen/Rings/Methods.jl
HechtiDerLachs: For the final outcome does not make a difference: `isempty` being `true` implies `iszero(one(OO(V)) == true`. So it can not return anything but zero, regardless of the check being enabled or not. But yes: In general, we want to avoid the Groebner basis computation.

experimental/Schemes/CoveredScheme.jl
HechtiDerLachs: It doesn't need it, correct. However, this is a way to catch a faulty passing on of `check` at an earlier stage, so I included it.

experimental/Schemes/WeilDivisor.jl
HechtiDerLachs: I agree that the standalone check is rather awkward, especially in conjunction with another `if check` - block. I was not aware of the possibility to do a `do` block. It looks way more reasonable.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
HechtiDerLachs: This is rather recent code of @simonbrandhorst . We should consult with him.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
HechtiDerLachs: > Is `@check true` any different from `@check`?

No. It is just a deprecated way in which I was writing it at first. 

I can't say anymore why there are so many methods which apparently don't use `check`. But sometimes it's also just easier to add that optional argument to all methods of a given function, even though only some use it. For instance: The `hom(A, B, img_gens)` methods to create maps of rings has a `check` argument. We want to write generic code which accepts all combinations of types `MPolyRing`, `MPolyQuoRing`, `MPolyLocRing`, `MPolyQuoLocRing` for `A` and `B`. Whenever there is a modulus involved in the domain `A`, there is some well-definedness-check which is eventually to be skipped. So we write the code as `hom(A, B, img_gens, check=false)` for internal calls with sane input. But then also the method with signature `A::MPolyRing` needs to have a dummy optional argument `check::Bool`, even though there's nothing to do there. 

Does that make sense?

src/Assertions.jl
HechtiDerLachs: I'm not sure, I understand. Maybe we can talk on Slack/Gather? I'd be interested to hear more about this.

HechtiDerLachs:
The signature you describe exists and is used. I will try to replace the standalone cases.

src/Assertions.jl
thofma: It does not really matter for performance, so I would not bother trying to change it. It will also be more tedious to make it a runtime option if the the check is moved outside.

HechtiDerLachs:
I removed the standalone statements of `@check` up to the controversial files introduced by @simonbrandhorst .

src/AlgebraicGeometry/Schemes/SpecOpen/Rings/Methods.jl
fingolfin: OK, I was indeed missing this! Which to me suggest that a (brief!) comment explaining this would be helpful :-)

experimental/Schemes/CoveredScheme.jl
fingolfin: OK... Perhaps then we can come up with a better name for the `@check` macro, at least the zero argument version? As it is, the name to me suggests that it checks something, and for the other variants of the `@check` macro, that's true. Moreover, the difference to `@assert` or `@req` etc. is that it also takes the value of the `check` variable into account.

And then we tack on this extra check which makes it complain about `check = true` -- as I understand it, this is something purely for developers, meant to help them catch places that "forgot" to disable `check`. 

The zero-argument version of `@check` then does nothing in "normal" mode. Only in the "developer looking for forgotten `check=false` mode does it do something. So perhaps `@debug_check` would make more sense for the standalone version?

All that said, I am not going to block this PR over this nitpicking. I am just worried that code using `@check` gets harder to reason about.

So what I *really* want then is at least a good docstring for the macro that explain all this.

experimental/Schemes/WeilDivisor.jl
fingolfin: Correction: a `do` block won't work, what I should have suggested is a `begin`-block. I've already communicate this to @HechtiDerLachs via Slack, but I wanted to clarify it here, too, in case I am confusing any other readers. Sorry for that :-(

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
fingolfin: Yeah, that makes sense. Thanks

src/Oscar.jl
fingolfin: yeah Hecke, AA, Nemo (still) do many things a bit different, and also Oscar is not yet consistent, but we kinda agree that filenames should use `snake_case` and not `kebab-case` or `CamelCase`. But of course we have dozens and dozens of files with existing names; there is no strong reason to rename them all; and if inside a directory *all* files already use `CamelCase`, I am fine with new files also using that. But for the files directly inside `src`, we are using `snake_case`, so this file should as well (of course `src/Oscar.jl` is the sole exception, but we don't have a choice there).

src/Rings/MPolyMap/MPolyQuo.jl
```diff
@@ -93,7 +94,7 @@ function hom(R::MPolyQuoRing, S::NCRing, images::Vector; check::Bool = true)
   @req n == length(images) "Number of images must be $n"
   # Now coerce into S or throw an error if not possible
   imgs = _coerce(S, images)
-  if check
+  @check begin
     _check_imgs_quo(R, S, imgs)
     _check_homo(S, imgs)
```
fingolfin, line 98:
```suggestion
    _check_homo(S, imgs)
    true
```
But actually, I would suggest a different approach: drop these `true` again (readers may wonder why they are they, plus they are very easy to forget, as this example shows), and instead improve the `check` macro to better deal with these `begin` blocks.

Two options come to mind:
1. we could simply accept the return value `nothing` just like we accept `true`; the easiest way would be to replace `$(esc(cond)) || error($(esc(msg)))` by `$(esc(cond)) != false || error($(esc(msg)))`. This way, only something that really returns `false` constitutes a failed check.
2. we could literally check for `begin` expressions and treat those differently than other expressions... not sure that's a good idea, though...

src/Assertions.jl
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+end
+
+macro check(cond)
+  quote
+    if $(esc(:(check)))
+      THROW_ERROR_FOR_INTERNAL_CHECKS && error("internal checks are enabled")
+      @assert $(esc(cond))
```
fingolfin: But `@assert` can be turned off; yet in the next macro, you use
```
$(esc(cond)) || error($(esc(msg)))
```
which cannot be turned off like other assertions. I don't think users of the `@check` macro would expect that adding a message turns it from an assertion into a non-assertion...

So either both should use `@assert` (the second macro then in the two argument form `@assert cond msg`, or both should not use `@assert` (and thus *always* evaluate `cond`. I think the latter is more appropriate here, isn't it?

fingolfin:
BTW, lest all my complaining and whining makes it sound as if I am opposed, let me emphasise that I think this is overall a good idea and I am grateful @HechtiDerLachs works on it. Thank you!

src/Oscar.jl
thofma: Even if the filename is the name of a type? I somehow missed that discussion.

HechtiDerLachs:
All files by @simonbrandhorst have been reverted. He will take care of them later. From my side this should be ready to go.

experimental/Schemes/Types.jl
```diff
@@ -364,9 +362,9 @@ identifications given by the glueings in the `default_covering`.
         G = default_covering(X)(ambient_scheme(V), ambient_scheme(W))
         f, g = glueing_morphisms(G)
         VG = intersect(V, domain(f))
-        preV = preimage(g, VG)
+        preV = preimage(g, VG, check=false)
```
afkafkafk13, line 365: This line is unreached, as the whole function is deactivated in line 352. Is this intentional and do you intentionally do changes in this unreached part?

src/AlgebraicGeometry/Schemes/Covering/Morphisms/Types.jl
```diff
@@ -37,6 +37,7 @@ mutable struct CoveringMorphism{DomainType<:Covering,
     # TODO: check domain/codomain compatibility
     # TODO: if check is true, check that all morphisms glue and that the domain patches
     # cover the basic patches of `dom`.
+    @check
```
afkafkafk13, line 40: Is this orphaned '@check' on purpose or was it forgotten in the last pass through the file?

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Constructors.jl
```diff
@@ -25,6 +25,7 @@ Vanishing locus
 ```
 """
 function projective_algebraic_set(X::AbsProjectiveScheme; check::Bool=true)
+  @check
```
afkafkafk13, line 28: again an orphaned '@check'... and one more further down in the file.

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Types.jl
```diff
@@ -1,6 +1,7 @@
 @attributes mutable struct ProjectiveAlgebraicSet{BaseRing<:Field, GradedRingType<:Ring} <: AbsProjectiveAlgebraicSet{BaseRing, GradedRingType}
   X::ProjectiveScheme{BaseRing,GradedRingType}
   function ProjectiveAlgebraicSet(X::ProjectiveScheme; check::Bool=true)
+    @check
```
afkafkafk13, line 4: orphaned '@check'...

src/AlgebraicGeometry/Schemes/SpecOpen/Rings/Types.jl
```diff
@@ -21,9 +21,7 @@ mutable struct SpecOpenRing{SpecType, OpenType} <: Ring
       U::OpenType;
       check::Bool=true
     ) where {SpecType<:AbsSpec, OpenType<:SpecOpen}
-    if check
-      issubset(U, X) || error("open set does not lay in the scheme")
-    end
+    @check issubset(U, X) "open set does not lay in the scheme"
```
afkafkafk13, line 24: typo: lie instead of lay

afkafkafk13 (commented):
Looks good to me. Mind the minor comments..

Thanks for this huge amount of work!

simonbrandhorst (approved):

src/AlgebraicGeometry/Schemes/Covering/Morphisms/Types.jl
simonbrandhorst: the check parameter is ignored too, but it is there for consistency.

simonbrandhorst:
Since all the comments have been addressed, I will merge this in 24 hours unless someone speaks up.

PR #1128, opened 2022-02-28
Add sparse multivariate polynomial rings to experimental

lgoettgens:
When computing stuff on low degree polynomials in polynomial rings with a lot of variables, the usual implementations (e.g. `AbstractAlgebra.Generic.MPolyRing`) are not efficient to use at all.
A sparse polynomial does not save the whole exponent vector for each monomial, but only pairs of (variable id, exponent) for non-zero exponents.

The increase in speed gets visible in the following example:
```julia
R, x = ZZ["y"]

S, varlist = PolynomialRingSparse(R, 10000)
@time sum(gens(S));   # 1.086290 seconds

T, varlist = PolynomialRing(R, 10000)
@time sum(gens(T));   # 124.527147 seconds
```

A lot of tests could be used from `AbstractAlgebra.Generic.MPolyRing`.

PR #2405, opened 2023-05-23
Remove `MPolyRingSparse` from experimental

lgoettgens:
Alternative to and thus closes #1990.

Instead of renaming the type, this PR fully removes it from `experimental`.

`MPolyRingSparse` is kind of useless in its current state, since all `MPolyRing`s are sparse, and most are more optimized and thus wayyy faster.
Furthermore, the use-case I added `MPolyRingSparse` for in #1128 (https://github.com/PBWDeformations/PBWDeformations.jl), has reverted to using `QQMPolyRing` from Nemo a long time ago, since that is actually faster.

Thus, I would propose to remove `MPolyRingSparse` from Oscar, let it be a "discarded experiment" (cf. https://docs.oscar-system.org/dev/Experimental/intro/#Procedure-for-adding-a-new-feature), and not waste any additional time with it.


fingolfin (approved):
Sounds reasonable

thofma:
Tests are failing

lgoettgens:
> Tests are failing

Should be fixed now (local tests succeed). Some other parts of OSCAR were relying on some imports into global scope by `MPolySparse.jl`.

PR #2410, opened 2023-05-24
first attempt to serialize some groups

ThomasBreuer:
This is just for discussing what we want for groups.

The current code tries to save and restore only data living on the Oscar side, and ignores additional knowledge stored in the objects on the GAP side.
(Still some objects have to be manipulated on the GAP side because they have no counterpart in Oscar; in the current code, this holds for the "full f. p. group" of a subgroup of an f. p. group.)

An alternative approach would be to serialize mainly the objects on the GAP side, together with their known attributes, and to rebuild the Oscar objects from the deserialized GAP objects.
(For serializing/deserializing GAP objects in GAP, there is currently essentially the "pickling/unpickling" approach provided by the IO package.)


ThomasBreuer:
The serialization interface promises that one can save some Oscar objects x, y, z, ... to (several) files, such that the following holds.

1. Loading the contents of the files into the *same* Oscar session will yield objects that are equal to the given objects x, y, z, ... --this is what `test_save_load_roundtrip` tests.
2. If one loads the contents of the files into a *different* Oscar session then the relations between the loaded objects x', y', z', ... corresponding to x, y, z, ... must be compatible with the relations between the original objects x, y, z, ...; that is, if x and y are elements in the same group then also x' and y' must be elements in the same group.

For an object that is based on a GAP object, requirement 1. makes it necessary to record the information about the `FamilyObj` value of the underlying GAP object, because otherwise we cannot create an object in the correct family.
The idea of the current code is to keep a dictionary of families of objects that have been serialized in the current Oscar session, and to make the keys for the families part of the serialization information.
Concerning requirement 2., dictionaries of families of deserialized objects from other Oscar sessions are kept, in order to create compatible objects. (*Testing* that deserializing objects from earlier Oscar sessions works seems to require that some serialization output becomes part of the test input; this part of the input may then need updates when the file format changes.)

It is likely that using Julia's `objectid` for computing unique keys of GAP family objects is not safe; perhaps it will be necessary to provide a function on the GAP side that yields a unique key.

(Note that GAP's "pickling/unpickling" approach does not deal (yet) with objects which rely on the identity of families.)

ThomasBreuer:
Currently two tests are failing:
`Run tests / test (1.6, ubuntu-latest) (pull_request)` was canceled, and `Run tests / test (1.9, ubuntu-latest)` found an error at `test/AlgebraicGeometry/Schemes/duValSing.jl:16`, which is not likely to be related to the changes from this pull request.

(O.k., restarted the tests, now they passed.)

antonydellavecchia:
you can take a look at my branch to see how one would go about reformatting for the refactor
https://github.com/oscar-system/Oscar.jl/tree/adv/start-reformat-serialize-groups

src/Serialization/GAP.jl
```diff
@@ -0,0 +1,249 @@
...
+  push!(_GAP_serializations, filtsymbol => meth)
+  return
+end
+
+function install_GAP_deserialization(filtsymbol::Symbol, meth::Function)
+  push!(_GAP_deserializations, filtsymbol => meth)
+  return
+end
+
+function _describe_GAP_flags(flags::GapObj)
```
fingolfin: This function isn't used anywhere, is it?

src/Serialization/GAP.jl
```diff
@@ -0,0 +1,249 @@
...
+  return [string(GAPWrap.NameFunction(x))
+          for x in GAPWrap.ELMS_LIST(GAP.Globals.FILTERS,
+                     GAP.Globals.TRUES_FLAGS(flags))]
+#TODO: omit implied filters
+end
+
+function _GAP_flags(filtnames::Vector)
```
fingolfin: This function also isn't used, is it?

src/Serialization/GAP.jl
```diff
@@ -0,0 +1,249 @@
...
+          end
+        end
+      end
+    end
+  end)
+
+install_GAP_deserialization(:IsFreeGroup,
+  function(filt::GapObj, s::DeserializerState, T, d::Dict)
+    # We have to create a new free group.
+    wfilt = GAP.evalstr(d[:wfilt])
```
fingolfin: Maybe this instead?
```suggestion
    wfilt = getproperty(GAP.Globals, Symbol(d[:wfilt]))
```

src/Serialization/GAP.jl
```diff
@@ -0,0 +1,249 @@
...
+           GAPWrap.IsWholeFamily(X))
+        # store generators
+        save_data_array(s, :gens) do
+          for x in GAPWrap.GeneratorsOfGroup(X)
+            v = Vector{Int}(GAPWrap.ExtRepOfObj(x))
+            save_data_array(s) do
+              for i in v
+                save_object(s, i)
+              end
+            end
```
fingolfin: Can't we just do
```suggestion
            save_object(s, v)
```

src/Serialization/GAP.jl
```diff
@@ -0,0 +1,249 @@
...
+      # relators
+      relators = GAP.getbangproperty(elfam, :relators)::GapObj
+      save_data_array(s, :relators) do
+        for x in relators
+          v = Vector{Int}(GAPWrap.ExtRepOfObj(x))
+          save_data_array(s) do
+            for i in v
+              save_object(s, i)
+            end
+          end
```
fingolfin: As above, can't we do this (and also in more places):
```suggestion
          save_object(s, v)
```

src/Serialization/GAP.jl
```diff
@@ -0,0 +1,249 @@
...
+  return [string(GAPWrap.NameFunction(x))
+          for x in GAPWrap.ELMS_LIST(GAP.Globals.FILTERS,
+                     GAP.Globals.TRUES_FLAGS(flags))]
+#TODO: omit implied filters
+end
+
+function _GAP_flags(filtnames::Vector)
+  # We need `GAP.evalstr` because of function names involving brackets.
+  l = [GAP.Globals.FLAGS_FILTER(GAP.evalstr(x))::GapObj
+       for x in filtnames]
```
fingolfin: I don't understand this code, and I could not find any callers to `_GAP_flags` so I am not sure how to test it. If `_GAP_flags` is removed this is of course perfectly fine.

src/Serialization/GAP.jl
ThomasBreuer: O.k.

src/Serialization/GAP.jl
ThomasBreuer: O.k.

src/Serialization/GAP.jl
ThomasBreuer: Yes.
(I had copied this code from somewhere. Apparently the simplified code yields the same result.)

src/Serialization/GAP.jl
ThomasBreuer: This belonged to experiments about setting certain filters programmatically, that is, to collect in the serialization  what GAP knows and to set this in the deserialization. I think this was a dangerous idea. Now I have removed this code.

src/Serialization/GAP.jl
ThomasBreuer: Yes. (See below.)

src/Serialization/GAP.jl
ThomasBreuer: Yes. (See below.)

antonydellavecchia:
After a quick look, seems good for a first attempt

thofma:
I restarted CI and merge once it is green. I would really like to use this.

thofma:
Thanks @ThomasBreuer!

Issue #2414, opened 2023-05-25
`Oscar.test_module()` subprocess uses a different project environment


aaruni96:
**Describe the bug**

`Oscar.test_module("path"; new=true)` spawns a julia subprocess, which does not inherit all parent properties, in particular, the `--project`. This can lead to strange behaviour when the dev version of Oscar is different from the installed version of Oscar on the system.

**To Reproduce**

```
$ julia --project=/path/to/project
julia> using Oscar
julia> Oscar.test_module(""; new=true)
```

**Expected behavior**
New process is spawned with the right path for `--project`

**System:**

```
julia> Oscar.versioninfo(full=true)
OSCAR version 0.12.0
  combining:
    AbstractAlgebra.jl   v0.29.4
    GAP.jl               v0.9.6
    Hecke.jl             v0.18.10
    Nemo.jl              v0.33.8
    Polymake.jl          v0.9.0
    Singular.jl          v0.18.3
  building on:
    Antic_jll               v0.201.500+0
    Arb_jll                 v200.2300.0+0
    Calcium_jll             v0.401.100+0
    FLINT_jll               v200.900.6+0
    GAP_jll                 v400.1200.200+1
    Singular_jll            v403.201.1+0
    libpolymake_julia_jll   v0.9.0+1
    libsingular_julia_jll   v0.30.0+1
    polymake_jll            v400.900.0+0
See `]st -m` for a full list of dependencies.

Julia Version 1.9.0
Commit 8e630552924 (2023-05-07 11:25 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, skylake)
  Threads: 1 on 8 virtual cores
Official https://julialang.org/ release
```

PR #2415, opened 2023-05-25
test_module: set julia project dir to Oscar of the calling process

aaruni96:
Solves #2414

### DIFF
--- a/src/utils/tests.jl
+++ b/src/utils/tests.jl
@@ -11,13 +11,14 @@ For experimental modules, use [`test_experimental_module`](@ref) instead.
 """
 function test_module(file::AbstractString; new::Bool=true)
   julia_exe = Base.julia_cmd()
+  project_path = split(pathof(Oscar), "/src")[1]
   rel_test_file = normpath("test", "$file.jl")
   test_file = joinpath(oscardir, rel_test_file)
 
   if new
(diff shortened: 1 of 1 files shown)


src/utils/tests.jl
```diff
@@ -11,13 +11,14 @@ For experimental modules, use [`test_experimental_module`](@ref) instead.
 """
 function test_module(file::AbstractString; new::Bool=true)
   julia_exe = Base.julia_cmd()
+  project_path = split(pathof(Oscar), "/src")[1]
```
lgoettgens:
```suggestion
  project_path = dirname(dirname(pathof(Oscar)))
```
This should be more robust, e.g. in cases where the path is `/home/myuser/src-code/Oscar.jl/src/Oscar.jl`.

thofma: Should use `joinpath` and not `/`.

thofma: How about `Oscar.oscardir`?

benlorenz: Maybe even `dirname(Pkg.project().path)`?
The project may have Oscar developed from a different directory.

fingolfin: I have used this with good success in various places:
```julia
run(`$(Base.julia_cmd()) --project=$(Base.active_project()) $(ARGS)`)
```

fingolfin (approved):

PR #2417, opened 2023-05-25
Introduce shifted_module for localizations at rational points.

HechtiDerLachs:
As requested by @afkafkafk13 . We will need to hang in the functionality with local orderings later.


afkafkafk13 (approved):
Looks good -- analogous to the ideal case.

PR #2419, opened 2023-05-26
added `multiplicities_eigenvalues` for characters

ThomasBreuer:


fingolfin (approved):

fieker:
too late now, but why that complicated name? We already have e.g. spectrum (and a bad version of eigenvalues)

ThomasBreuer:
The point is that the eigenvalues themselves are not computed/returned, just the multiplicities.
For `spectrum`, one would have to decide where the eigenvalues live: in a cyclotomic field (which one?), in the abelian closure of `QQ`?

PR #2420, opened 2023-05-26
[FTheoryTools] More updates to literature models

apturner:
Add additional data to literature models and add `create_literature_model_index` to automatically create index file from existing model .json files.

Propagates the changes in format for literature model data into the rest of the code, so that all existing functions should now work with the new format.


experimental/FTheoryTools/src/LiteratureTateModels/create_index.jl
```diff
@@ -0,0 +1,48 @@
+function create_literature_model_index()
```
HereAround: Maybe this should be an internal function?
```suggestion
_function create_literature_model_index()
```

HereAround: It might be good to add a brief comment on this function too. After all, this might be super relevant for the serialization/once we add new models. So a comment (in the code) will probably be very useful for the (future) developers?

In the (hopefully not too distant future), I would hope that we could have a section in the documentation that says "how you can contribute/submit your model". Maybe we can anticipate this change an even write a docstring for this function? Or does that seem to be too technical as of now?

HereAround (approved):
Thank you for working on this @apturner . I like the fact that you made all of the fields/information of the models available within julia (as attributes of the global Tate models). That is cool!

I have left a few questions/comments regarding `function create_literature_model_index()`. This should be super useful to update the literature models once we finally get this started. So on the one hand, this seems to call for a entry in the documentation. On the other hand, this seems a rather technical function (from a user perspective), which would call for a leading underscore.

Still, this appears to be a nitpick (in comparison to the challenges with the hypersurface models). Let me know if you want to adjust this. Other than that, from my perspective this is good to be merged.

experimental/FTheoryTools/src/LiteratureTateModels/create_index.jl
apturner: Thank you for pointing this out! I've converted the function to an internal function, and added a comment in the code to describe its functionality.

I agree that ultimately we should have documentation on how users can add models, but that doesn't seem to go with the index-creating function. We should probably make a separate function to let users do this at some point; this will be easiest after the real serialization is done. Then the user can simply create a model in Oscar, adorn it with any resolutions and paper data they have, and then run said function to serialize it and send the result to us in some way for inclusion.

apturner:
Thanks for the review @HereAround. As mentioned above, I converted `_create_literature_model_index` to an internal function, as I don't think we ever want the users to be regenerating the index themselves. If you're happy with this as it is, I think it's ready to merge! Thanks.

HereAround:
> Thanks for the review @HereAround. As mentioned above, I converted `_create_literature_model_index` to an internal function, as I don't think we ever want the users to be regenerating the index themselves. If you're happy with this as it is, I think it's ready to merge! Thanks.

Thank you. Yes, looks good now.

PR #2421, opened 2023-05-26
Do not mess with documentation of upstream packages

thofma:
The function `Oscar.build_doc(;doctest = :fix)` runs the doctests also for all packages listed under `makedocs(..., modules = [...])`. But the documentation/docstrings of those packages does not use the magical printing, so those doctests are doomed to fail.

Edit: We do not see this during CI, because doctests are run with `doctest(Oscar)`.


fingolfin (approved):
Fine by me. I guess on the long run we should have `doctestsetup` methods in each of these packages and call them here.

Issue #2422, opened 2023-05-27
Do not run all the doctests in `build_doc`


thofma:
At the moment , `Oscar.build_doc` calls `makedocs(...)` with all our upstream packages. I think this is necessary, so that all docstrings are picked up properly. But this has the big disadvantage, that doctests for all upstream packages are run. This is quite a waste of time and also leads to errors, if run with `:fix` and something unrelated is broken in a dependency. And it is also not what we do test with CI. I propose to replace 
https://github.com/oscar-system/Oscar.jl/blob/b350dd12616a6771fd2d6c9ef089065d7c5c93f2/docs/make_work.jl#L184
by a separate manual `Documenter.doctest(Oscar; doctest = doctest)` call.

Any objections @fingolfin @benlorenz? Or something that I am missing?

fingolfin:
Makes sense to me

lgoettgens:
This has been fixed by https://github.com/oscar-system/Oscar.jl/pull/2423.

fingolfin:
We now have `doctest_fix`. I am not sure if there is anything left to be done, @thofma ?

thofma:
It was already fixed in #2423.

Issue #2413, opened 2023-05-25
OscarCI-doctest failures in upstream projects


lgoettgens:
The `matching: [Oscar#master]` CI job fails on some doctests, e.g. https://github.com/oscar-system/Singular.jl/actions/runs/5078789737/jobs/9123769090?pr=652.

The failure seems to be due to some changes in https://github.com/oscar-system/Oscar.jl/pull/2345, in particular `set_current_module(@__MODULE__)` not being available in `OscarCI`.

After a Oscar release, this could probably be fixed by adding the above to https://github.com/oscar-system/OscarDevTools.jl/blob/master/src/OscarCI.jl#L340

@benlorenz proposed on slack:
> maybe it would make sense to add a `Oscar.doctestsetup()`  function which returns that expression? And in the oscar ci we try to use that or fall back to `using Module`, that way also other modules could add code to the `doctestsetup`

cc @benlorenz @thofma @fingolfin

benlorenz:
fixed with #2423 and the [latest OscarDevTools.jl](https://github.com/oscar-system/OscarDevTools.jl/releases/tag/v0.2.12)

PR #2423, opened 2023-05-27
Tweaks for the documentation build

thofma:
- Introduce `doctestsetup()`.
- Do not run upstream doctests in `build_doc`.

### DIFF
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -170,10 +170,14 @@ function doit(
   end
 
   cd(joinpath(Oscar.oscardir, "docs")) do
-    DocMeta.setdocmeta!(Oscar, :DocTestSetup, :(using Oscar; Oscar.AbstractAlgebra.set_current_module(@__MODULE__)); recursive=true)
+    DocMeta.setdocmeta!(Oscar, :DocTestSetup, Oscar.doctestsetup(); recursive=true)
     DocMeta.setdocmeta!(Oscar.Hecke, :DocTestSetup, :(using Hecke); recursive=true)
     DocMeta.setdocmeta!(Oscar.AbstractAlgebra, :DocTestSetup, :(using AbstractAlgebra); recursive=true)
     DocMeta.setdocmeta!(Oscar.Nemo, :DocTestSetup, :(using Nemo); recursive=true)
+
+    if doctest !== false
+      Documenter.doctest(Oscar, fix = doctest === :fix)
+    end
 
     makedocs(
       bib;
@@ -181,7 +185,7 @@ function doit(
       sitename="Oscar.jl",
       modules=[Oscar, Oscar.Hecke, Oscar.Nemo, Oscar.AbstractAlgebra, Oscar.Singular],
       clean=true,
-      doctest=doctest,
+      doctest=false,
       strict=strict,
       checkdocs=:none,
       pages=doc,
--- a/src/utils/docs.jl
+++ b/src/utils/docs.jl
@@ -6,6 +6,16 @@
 ###############################################################################
 ###############################################################################
 
+################################################################################
+#
+#  DocTestSetup
+#
+################################################################################
+
+# Oscar needs some complicated setup to get the printing right. This provides a
+# helper function to set this up consistently.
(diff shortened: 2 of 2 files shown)


thofma:
Fixes #2422 and implements some of the changes required to fix #2413.

docs/make_work.jl
```diff
@@ -6,6 +6,16 @@ module BuildDoc
...
 
+################################################################################
+#
+#  DocTestSetup
+#
+################################################################################
+
+# Oscar needs some complicated setup to get the printing right. This provides a
+# helper function to set this up consistently.
+doctestsetup() = :(using Oscar; Oscar.AbstractAlgebra.set_current_module(@__MODULE__))
```
benlorenz: Can we put this into `src/utils/docs.jl` instead? Note that there are already two more such lines in this file. Or will this cause problems because the `@__MODULE__` will probably be resolved differently?

That would make it easier to use in OscarCI where make_work.jl is not used. I wanted the code there to stay as generic as possible, maybe like this:
```
push!(testcmd, """using $pkg; DocMeta.setdocmeta!($pkg, :DocTestSetup, isdefined($pkg, :doctestsetup) ? $pkg.doctestsetup() : :(using $pkg); recursive = true); doctest($pkg);""")
```

thofma: Yes, good idea. I will adjust it.

docs/make_work.jl
```diff
@@ -170,18 +180,22 @@ function doit(
   end
 
   cd(joinpath(Oscar.oscardir, "docs")) do
-    DocMeta.setdocmeta!(Oscar, :DocTestSetup, :(using Oscar; Oscar.AbstractAlgebra.set_current_module(@__MODULE__)); recursive=true)
+    DocMeta.setdocmeta!(Oscar, :DocTestSetup, doctestsetup(); recursive=true)
     DocMeta.setdocmeta!(Oscar.Hecke, :DocTestSetup, :(using Hecke; import AbstractAlgebra; AbstractAlgebra.set_current_module(@__MODULE__)); recursive=true)
     DocMeta.setdocmeta!(Oscar.AbstractAlgebra, :DocTestSetup, :(using AbstractAlgebra; AbstractAlgebra.set_current_module(@__MODULE__)); recursive=true)
     DocMeta.setdocmeta!(Oscar.Nemo, :DocTestSetup, :(using Nemo; import AbstractAlgebra; AbstractAlgebra.set_current_module(@__MODULE__)); recursive=true)
 
+    if !(doctest isa Bool && doctest == false)
```
fingolfin: Wouldn't this also work?
```suggestion
    if doctest !== false
```

docs/make_work.jl
```diff
@@ -170,18 +180,22 @@ function doit(
...
+      Documenter.doctest(Oscar, fix = doctest isa Symbol && doctest == :fix)
```
```suggestion
      Documenter.doctest(Oscar, fix = doctest === :fix)
```

fingolfin (approved):
Thank you. Fine by me, left some minor suggestions, and of course @benlorenz had a good one, too

docs/make_work.jl
thofma: Ah, much better, yes

docs/make_work.jl
```diff
@@ -170,18 +180,22 @@ function doit(
   end
 
   cd(joinpath(Oscar.oscardir, "docs")) do
-    DocMeta.setdocmeta!(Oscar, :DocTestSetup, :(using Oscar; Oscar.AbstractAlgebra.set_current_module(@__MODULE__)); recursive=true)
+    DocMeta.setdocmeta!(Oscar, :DocTestSetup, doctestsetup(); recursive=true)
```
fingolfin: One more suggestion: let's modify (in a future PR, not this one) all `.md` files which currently use
````
```@setup oscar
using Oscar
```
````
to instead use
````
```@setup oscar
using Oscar
@eval Oscar.doctestsetup()
```
````
Or something like that; to ensure we get a uniform setup everywhere

fingolfin: (After the modification suggested by @benlorenz )

thofma:
Changed following to your comments.

PR #2424, opened 2023-05-27
Make algebraic sets and varieties internal

simonbrandhorst:
There were still some controversial points about varieties. So they are still unstable and should not 
be exported when @thofma  does the next oscar release.


simonbrandhorst:
cc @afkafkafk13

thofma (approved):

Issue #1505, opened 2022-08-11
`coefficient_ring` for modules


thofma:
If I remember correctly, the fanbase for `base_ring` is rather small (mainly because we used it in a different way than it was designed for) and we started introducing mathematically well-defined terms like `coefficient_ring` for polynomial rings.

I think we should do the same for modules and introduce `coefficient_ring`. So whenever we have a ring `R` and `M` is an `R`-module (does not matter which type it has or how it is constructed), I propose to have `coefficient_ring(M) == R`.

Any concerns? @wdecker @fingolfin @fieker?

ederc:
I'm not sure I understand correctly but wouldn't it be more natural to have

`base_ring(M) == R`
and
`coefficient_ring(M) == coefficient_ring(R)`
?

wdecker:
If R is a polynomial ring with coefficients in C, and M is an R-module, then M is an R-module but also a C-algebra. But still we are concerned with the R-module structure of M when we talk about modules. For example, if we have elements of M, we talk about R-linear combinations of these elements, or about linear combinations of these elements with coefficients in R. So `coefficient_ring(M) == R` seems to be fine. If we would consider instead an OSCAR type for C-algebras, what we don't, then ...

ederc:
I see, alright.

wdecker:
By the way, what I wrote fits also to the coefficient function for module elements:

julia> R, (x, y, z) = PolynomialRing(QQ, ["x", "y", "z"])
(Multivariate Polynomial Ring in x, y, z over Rational Field, fmpq_mpoly[x, y, z])

julia> A = R[x; y]
[x]
[y]

julia> B = R[x^2; y^3; z^4]
[x^2]
[y^3]
[z^4]

julia> M = SubQuo(A, B)
Subquotient of Submodule with 2 generators
1 -> x*e[1]
2 -> y*e[1]
by Submodule with 3 generators
1 -> x^2*e[1]
2 -> y^3*e[1]
3 -> z^4*e[1]

julia> coefficients(x^2*M[1])
Sparse row with positions [1] and values fmpq_mpoly[x^2]

fingolfin:
We just discussed this here. There weren't many concerns with the suggestion `coefficient_ring(M) == R`, but still some:
1. it does sound somewhat weird (but we didn't really have a better suggestion either...)
[...]

Issue #938, opened 2022-01-03
The use of base_ring


wdecker:
I understand that base_ring should mean "one up" in the ring hierarchy. Like here:
```
julia> R, (x, y, z) = PolynomialRing(QQ, ["x", "y", "z"])
(Multivariate Polynomial Ring in x, y, z over Rational Field, fmpq_mpoly[x, y, z])

julia> base_ring(R)
Rational Field

julia> coefficient_ring(R)
Rational Field
```
Now how about this example:
```
julia> R, (x, y, z) = GradedPolynomialRing(QQ, ["x", "y", "z"])
(Multivariate Polynomial Ring in x, y, z over Rational Field graded by 
  x -> [1]
  y -> [1]
  z -> [1], MPolyElem_dec{fmpq, fmpq_mpoly}[x, y, z])

julia> base_ring(R)
Rational Field

julia> coefficient_ring(R)
Rational Field
```
Shouldn't base_ring here refer to the underlying polynomial ring of the graded ring? Or  do we need another identifier for this?

tthsqe12:
I think coefficient_ring was added precisely so that base_ring is free to return whatever it wants on graded polynomial rings. So, the current behavior is a bit strange to me.

wdecker:
So the suggestion would be to replace

   base_ring(W::MPolyRing_dec) = base_ring(W.R)

by

   base_ring(W::MPolyRing_dec) = W.R

in mpoly-graded.jl, and then replace the old base_ring by coefficient_ring wherever this is needed. 

Are there objections?

thofma:
@tthsqe12 Can you confirm that this won't work, since the code for generic (arbitrary) multivariate polynomials always uses `base_ring`, see https://github.com/Nemocas/AbstractAlgebra.jl/blob/master/src/MPoly.jl? I think the graded/decorated polynomials are falling back to these methods a lot.

wdecker:
So coefficient_ring is defined via base_ring? I am confused.

thofma:
`base_ring` was there before and we just added the alias `coefficient_ring(x) = base_ring(x)`.

fingolfin:
Part of the reason we added `coefficient_ring` was because I claimed (and still claim) that `base_ring` is not a well-defined concept; and I believe that using it results in brittle code, because it is not clear that different implementations of the same concept will handle it equally. I think @thofma's comment illustrates that.

[...]

PR #2425, opened 2023-05-27
LieAlgebras: Add docs

lgoettgens:
This PR adds documentation to the LieAlgebras folder in experimental.

Weirdly, some docstrings will not show up in the HTML, e.g. `(::LieAlgebra{C})(::Vector{C}) where {C<:RingElement}`.
Can someone help me with that?

Edit: Additionally some minor changes.

### DIFF
--- a/experimental/LieAlgebras/docs/src/lie_algebras.md
+++ b/experimental/LieAlgebras/docs/src/lie_algebras.md
@@ -4,9 +4,12 @@ CurrentModule = Oscar
 
 # Lie algebras
 
-Lie algebras in OSCAR are always finite dimensional, and represented by two different types, 
-namely `LinearLieAlgebra` and `AbstractLieAlgebra`, depending on whether a matrix representation is available or not.
-Both types are subtypes of `LieAlgebra`. Similar to other types in OSCAR, each Lie algebra type has a corresponding element type.
+Lie algebras in OSCAR are always finite dimensional, and represented by two different types,
+namely `LinearLieAlgebra{C}` and `AbstractLieAlgebra{C}`, depending on whether a matrix
+representation is available or not.
+Both types are subtypes of `LieAlgebra{C}`. Similar to other types in OSCAR, each Lie algebra
+type has a corresponding element type.
+The type parameter `C` is the element type of the coefficient ring. 
 
 ```@docs
 zero(::LieAlgebra{C}) where {C<:RingElement}
@@ -17,7 +20,7 @@ basis(::LieAlgebra{C}, ::Int) where {C<:RingElement}
 coefficients(::LieAlgebraElem{C}) where {C<:RingElement}
 coeff(::LieAlgebraElem{C}, ::Int) where {C<:RingElement}
 getindex(::LieAlgebraElem{C}, ::Int) where {C<:RingElement}
-symbols(_::LieAlgebra{C}) where {C<:RingElement}
+symbols(::LieAlgebra{C}) where {C<:RingElement}
 ```
 
 ## Special functions for `LinearLieAlgebra`s
@@ -26,15 +29,12 @@ symbols(_::LieAlgebra{C}) where {C<:RingElement}
 matrix_repr_basis(::LinearLieAlgebra{C}) where {C<:RingElement}
 matrix_repr_basis(::LinearLieAlgebra{C}, ::Int) where {C<:RingElement}
 matrix_repr(::LinearLieAlgebraElem{C}) where {C<:RingElement}
-```
-```
 (::LinearLieAlgebra{C})(::MatElem{C}) where {C<:RingElement}
-
 ```
 
 ## Element constructors
 
-```
+```@docs
 (::LieAlgebra{C})() where {C<:RingElement}
 (::LieAlgebra{C})(::Vector{C}) where {C<:RingElement}
 (::LieAlgebra{C})(::Vector{Int}) where {C<:RingElement}
@@ -44,7 +44,7 @@ matrix_repr(::LinearLieAlgebraElem{C}) where {C<:RingElement}
 ```
 
 ## Arithmetics
-The usual arithmetics, e.g. `+`, `-`, and `*` are defined for `LieAlgebraElem`s.
+The usual arithmetics, e.g. `+`, `-`, and `*`, are defined for `LieAlgebraElem`s.
 
 !!! warning
     Please note that `*` refers to the Lie bracket and is thus not associative.
--- a/experimental/LieAlgebras/docs/src/lie_algebras.md
+++ b/experimental/LieAlgebras/docs/src/lie_algebras.md
@@ -32,19 +32,26 @@ symbols(::LieAlgebra{C}) where {C<:RingElement}
 matrix_repr_basis(::LinearLieAlgebra{C}) where {C<:RingElement}
 matrix_repr_basis(::LinearLieAlgebra{C}, ::Int) where {C<:RingElement}
 matrix_repr(::LinearLieAlgebraElem{C}) where {C<:RingElement}
-(::LinearLieAlgebra{C})(::MatElem{C}) where {C<:RingElement}
 ```
 
 ## Element constructors
 
-```@docs
-(::LieAlgebra{C})() where {C<:RingElement}
-(::LieAlgebra{C})(::Vector{C}) where {C<:RingElement}
-(::LieAlgebra{C})(::Vector{Int}) where {C<:RingElement}
-(::LieAlgebra{C})(::MatElem{C}) where {C<:RingElement}
-(::LieAlgebra{C})(::SRow{C}) where {C<:RingElement}
-(::LieAlgebra{C})(::LieAlgebraElem{C}) where {C<:RingElement}
-```
+`(L::LieAlgebra{C})()` returns the zero element of the Lie algebra `L`.
+
+`(L::LieAlgebra{C})(x::LieAlgebraElem{C})` returns `x` if `x` is an element of `L`,
+and fails otherwise.
+
+`(L::LieAlgebra{C})(v)` constructs the element of `L` with coefficient vector `v`.
+`v` can be of type `Vector{C}`, `Vector{Int}`, `SRow{C}`,
+or `MatElem{C}` (of size $1 \times \dim(L)$).
+
+If `L` is a `LinearLieAlgebra` of `dim(L) > 1`, the call
+`(L::LinearLieAlgebra{C})(m::MatElem{C})` returns the Lie algebra element whose
+matrix representation corresponds to `m`.
+This requires `m` to be a square matrix of size `n > 1` (the dimension of `L`), and
+to lie in the Lie algebra `L` (i.e. to be in the span of `basis(L)`).
+The case of `m` being a $1 \times \dim(L)$ matrix still works as explained above.
+
 
 ## Arithmetics
 The usual arithmetics, e.g. `+`, `-`, and `*`, are defined for `LieAlgebraElem`s.
--- a/experimental/LieAlgebras/docs/src/lie_algebras.md
+++ b/experimental/LieAlgebras/docs/src/lie_algebras.md
@@ -1,0 +1,62 @@
+```@meta
+CurrentModule = Oscar
+```
+
+# Lie algebras
+
+Lie algebras in OSCAR are always finite dimensional, and represented by two different types, 
+namely `LinearLieAlgebra` and `AbstractLieAlgebra`, depending on whether a matrix representation is available or not.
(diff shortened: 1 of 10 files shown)


experimental/LieAlgebras/docs/src/lie_algebras.md
```diff
@@ -0,0 +1,69 @@
+```@meta
+CurrentModule = Oscar
+```
+
+# Lie algebras
+
+Lie algebras in OSCAR are always finite dimensional, and represented by two different types,
```
fingolfin:
```suggestion
Lie algebras in OSCAR are currently always finite dimensional, and represented by two different types,
```

I hope this restricting to finite dimensional will be eventually lifted, and e.g. free Lie algebras (via GAP's `FreeLieAlgebra`) be supported, and more generally finitely presented Lie algebras (e.g. via the GAP package `fpsla`), Kac-Moody algebras and more.

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -114,6 +114,20 @@ end
...
 ###############################################################################
 
+@doc raw"""
+    lie_algebra(R::Ring, struct_consts::Matrix{SRow{C}}, s::Vector{<:VarName}; cached::Bool, check::Bool) -> AbstractLieAlgebra{elem_type(R)}
+
+Construct the Lie algebra over the ring `R` with structure constants given by
+`struct_consts` and with basis element names `s`.
+
+* `struct_consts`: The entry with indices `[i,j][k]` is the scalar $a_{i,j}^k$
+  such that $[x_i, x_j] = \sum_k a_{i,j}^k x_k$.
```
fingolfin: Should say what `x_i` is

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -114,6 +114,20 @@ end
 #
 ###############################################################################
 
+@doc raw"""
+    lie_algebra(R::Ring, struct_consts::Matrix{SRow{C}}, s::Vector{<:VarName}; cached::Bool, check::Bool) -> AbstractLieAlgebra{elem_type(R)}
+
+Construct the Lie algebra over the ring `R` with structure constants given by
+`struct_consts` and with basis element names `s`.
+
+* `struct_consts`: The entry with indices `[i,j][k]` is the scalar $a_{i,j}^k$
```
fingolfin: Personally I'd rename `struct_consts` to something short, e.g. `SC`, and then write more like the following (but feel free to disagree and ignore this). Also, while I know that `a_{i,j}^k` is semi-standard, it is still easy to get confused and think the `^x` refers to exponentiation. So here is what I'd write:
```suggestion
The Lie bracket on the newly constructed Lie algebra `L` is determined by the structure
constants in `SC` as follows: let `x_i` denote the `i`-th standard basis vector
of `L`. Then the entry `SC[i,j][k]` is a scalar $a_{i,j,k}$
```

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -143,8 +206,16 @@ function lie_algebra(
   return AbstractLieAlgebra{elem_type(R)}(R, struct_consts2, Symbol.(s); cached, check)
 end
 
+@doc raw"""
+    lie_algebra(R::Ring, dynkin::Tuple{Char,Int}; cached::Bool) -> AbstractLieAlgebra{elem_type(R)}
+
+Construct the Lie algebra over the ring `R` with Dynkin type given by `dynkin`.
```
fingolfin:
```suggestion
Construct the simple Lie algebra over the ring `R` with Dynkin type given by `dynkin`.
```

Ideally this would also describe what `dynkin` is exactly.

On the long term, I guess we want dedicated machinery to deal with (extended) Dynkin diagrams, (generalized) Cartan matrices, etc. etc., and then use these everywere. But baby steps... :-)

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -102,33 +147,65 @@ end
...
+"""
 function (L::LieAlgebra{C})() where {C<:RingElement}
   return zero(L)
 end
 
+@doc raw"""
+    (L::LieAlgebra{C})(v::Vector{Int}) -> LieAlgebraElem{C}
+
+Return the element of `L` with coefficent vector `v`.
+Fails, if `Int` cannot be coerced into the base ring of `L`.
```
fingolfin:
```suggestion
Fail, if `Int` cannot be coerced into the base ring of `L`.
```

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -46,6 +46,11 @@ base_ring(V::LieAlgebraModule{C}) where {C<:RingElement} = base_ring(base_lie_al
 
 base_ring(v::LieAlgebraModuleElem{C}) where {C<:RingElement} = base_ring(parent(v))
```
fingolfin, line 47: Not about this PR but about existing code: I find this surprising, I would have expected `base_ring` to return the Lie algebra. But instead it seems to return the base ring of the Lie algebra, i.e. its "coefficient ring" or "left-acting domain" (in GAP parlance). Of course in part this is because `base_ring` is notoriously ill-defined, so guessing what it does is always problematic.

And right now, this definition is inconsistent with other existing module types, where the full acting ring is returned.

So this is a bit unfortunate. One solution then would be to change `base_ring(V::LieAlgebraModule)` to return the same as `base_lie_algebra(V)`, and then add a separate function to access the "coefficients domain". 

E.g. under the name `coefficient_ring`; while it is not super common to refer to the `R` in `R-algebra` by the name `coefficient ring`, I think it is fine as long as we specify it explicitly: 
> The coefficient ring of a module is the coefficient ring of its base ring.
> The coefficient ring of a ring which is *constructed* as an R-algebra is R.

The bit about "constructed as" is meant to clarify that while e.g. the polynomial ring `GF(4)[:x]` is both an $F_4$ and an $F_2$ algebra, it was constructed as an $F_4$ algebra.

After typing this I remembered we have #1505 (and also #938) ... so there :-)

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -54,21 +59,46 @@ gens(L::LieAlgebraModule{C}) where {C<:RingElement} = basis(L)
 
 gen(L::LieAlgebraModule{C}, i::Int) where {C<:RingElement} = basis(L, i)
 
+@doc raw"""
+    dim(V::LieAlgebraModule{C}) -> Int
+
+Return the dimension of the Lie algebra module `V`.
+"""
 dim(V::LieAlgebraModule{C}) where {C<:RingElement} = V.dim
```
fingolfin, line 67: Again off-topic, but: you can do away with `C` in many cases, making the code maybe more readable?
```suggestion
    dim(V::LieAlgebraModule) -> Int

Return the dimension of the Lie algebra module `V`.
"""
dim(V::LieAlgebraModule) = V.dim
```

fingolfin (approved):
Thank you, great improvement!

I have some suggestions but mostly optional and/or could be done later. Also the point on base_ring / coefficient_ring of course need not be addressed in this PR, but we should still address them (so either do so, or if you don't have time right now, perhaps you can open an issue for them?)

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens: I understand your point.

However, I found it quite confusing to refer to the lie algebra as `base_ring`, although it is not a ring. My idea now would be to abolish `base_ring`, as it is misleading and instead use `base_lie_algebra` and `coefficient_ring`.
But that will happen in a future PR.

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens: Didn't know that. I will look into it for a future PR.

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
lgoettgens: I personally think that `struct_consts` is better following the principle of self-commenting code. Some of your wording has been adapted in e19efeaca58f9bdf1ffa03cff8ed36f8c5cde75d.

lgoettgens:
Thanks for your comments @fingolfin!
I hope I addressed all of them.

Unfortunately, the docs are not building. Do you have any idea why?

experimental/LieAlgebras/src/LieAlgebraModule.jl
fingolfin: Well, every Lie algebra is also a Lie ring, and many people refer to "non-associative rings". But yeah, a more generic name than `base_ring` would be helpful...

But in general my view remains that I'd wish we could do completely away with `base_ring` ;-). And if we have `coefficient_ring` and `base_lie_algebra` (or, for that matter, just `lie_algebra(M)` ???), it's not so important what `base_ring` does exactly (or we could even consider omitting it entirely?).

experimental/LieAlgebras/docs/src/lie_algebras.md
```diff
@@ -0,0 +1,69 @@
+```@meta
+CurrentModule = Oscar
```
fingolfin, line 2:
```suggestion
CurrentModule = Oscar
DocTestSetup = quote
  using Oscar
end
```
I think you need this for the doctests

fingolfin:
There still are errors in your docs:
```
┌ Error: unable to get the binding for '(::LinearLieAlgebra{C})(::MatElem{C}) where {C<:RingElement}' in `@docs` block in src/Experimental/LieAlgebras/lie_algebras.md:31-36 from expression ':((::LinearLieAlgebra{C})(::MatElem{C}) where C <: RingElement)' in module Oscar
│ ```@docs
│ matrix_repr_basis(::LinearLieAlgebra{C}) where {C<:RingElement}
│ matrix_repr_basis(::LinearLieAlgebra{C}, ::Int) where {C<:RingElement}
│ matrix_repr(::LinearLieAlgebraElem{C}) where {C<:RingElement}
│ (::LinearLieAlgebra{C})(::MatElem{C}) where {C<:RingElement}
│ ```
│   exception = `binding` cannot understand expression `::LinearLieAlgebra{C}`.
└ @ Documenter.Expanders ~/.julia/packages/Documenter/H5y27/src/Utilities/Utilities.jl:32
```

I am afraid documenting "call syntax" is simply not supported at this point, see https://github.com/JuliaDocs/Documenter.jl/issues/558

lgoettgens:
That's unfortunate. I guess I then just copy the docstrings into the corresponding `.md` file for now.

experimental/LieAlgebras/src/LieAlgebraModule.jl
thofma: I think for the name of genericity (is this a word), the name should not contain "lie". For the modules over finite-dimensional associative algebras, I called this just `algebra(M)` and `coefficient_ring(M) == coefficient_ring(algebra(M))` (there is no `base_ring(M)`).
```
julia> A = matrix_algebra(QQ, 2);

julia> V = Hecke.Amodule(A, [matrix(a) for a in basis(A)]);

julia> algebra(V)
Matrix algebra of dimension 4 over Rational field

julia> coefficient_ring(V)
Rational field
```
Not sure about the `coefficient_ring` anymore, but I like `algebra(M)`. I think it would be great if we could try to be somehow consistent. (Sorry about the missing documentation.) I don't insist that the names I have chosen are the best ones and I am happy to adjust them.

lgoettgens: With `algebra`, I have kind of the same problem as with `base_ring` for denoting the Lie algebra, as it is not an algebra in the usual sense.

lgoettgens:
Everything builds now fine. From my point, this is good to go (once the CI finishes).

The two remaining conversations (on `base_ring` and removing `{C}`) will be postponed to a future PR.

experimental/LieAlgebras/src/LieAlgebraModule.jl
thofma: We just relax what we mean with an "algebra", so it could be fine. I don't think this would be a big problem. But we will discuss this later.

lgoettgens:
Is there something that needs to be changed about this @fingolfin? Otherwise, I would appreciate a merge and I will work on the things brought up here in a follow-up PR.

fingolfin:
Oops, sorry, I forgot about this :-(

PR #2432, opened 2023-05-29
Load `hulpke/extraperfect` as lazy artifact

### LINKED ISSUE #1245: perfect groups: add additional data as lazy artifacts
We currently have this:
```
julia> perfect_group(1966080,1)
ERROR: Error thrown by GAP: Error,

For reasons of size, the perfect groups library for orders >10^6 is
not distributed fully by default. To access the group requested, get
the file perf33.grp from
https://github.com/hulpke/extraperfect
and put it in the `grp` subdirectory of your GAP installation. Then type

return;

to continue in this GAP session (which will read in the file).

[ ... stacktrace omitted ... ]
```
Needless to say this is not helpful for an Oscar user.

I think we should catch this, and take care of it, by making these two files available as "Lazy artifacts", loaded on demand.

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/1245.

~Hook into `GAP`'s `PerfGrpLoad` to create a symlink to a lazy artifact instead of throwing the error shown in #1245.
`lazy = true` for an artifact needs `LazyArtifacts` to be imported, and only downloads the artifact the first time the artifact string (the `artifact"gap_hulpke_extraperfect"`) is encountered.~

I used a fixed commit of [hulpke/extraperfect](https://github.com/hulpke/extraperfect) instead of `master` to avoid hash mismatches once there is a new commit on `hulpke/extraperfect#master`. This means that the used version of `hulpke/extraperfect` has to be actively updated in `Artifacts.toml`.


gap/OscarInterface/gap/perf.grp.gi
```diff
@@ -0,0 +1,40 @@
+BindGlobal( "PerfGrpLoadOscar", function(sz)
```
fingolfin: Replacing `PerfGrpLoad` like that is a bit brittle :-/.

Here's an alternative idea: perhaps we could always provide `perf27.grp` and `perf33.grp` files, but instead of them being symlinks, provide actual files that essentially do something like this:
```gap
Read(JuliaToGAP(IsString, JuliaEvalString("""artifact"gap_hulpke_extraperfect/extraperfect-80c0cf27bf98ba3eabd287db5ab4fd0f249e0559/perf27.grp" """)));
```

lgoettgens: Yeah, I thought similar. Your idea seems a bit less hacky. I tried to implement it and after some fiddling around got it to work in daa3f0a.

test/Groups/libraries.jl
```diff
@@ -108,6 +108,9 @@ end
    @test number_perfect_groups(ZZRingElem(60)^3) == 1
    @test_throws ArgumentError number_perfect_groups(0) # invalid argument
    @test_throws ArgumentError number_perfect_groups(ZZRingElem(60)^10)  # result not known
+
+   # lazy artifact loading
+   @test perfect_group(1376256, 1) isa PermGroup
```
fingolfin: Hmm, that adds (another?) network access to the test suite, as such the tests may fail if one is offline. And of course it increases disk usage for the testing machine (which is fine for me locally, it's CI jobs I am very slightly worried about).

Dunno if this is a problem with e.g. NanoSoldier.

Anyway, one way to try to deal with this might be to modify this to catch whatever exception is thrown when a download fails, and just accept that?

fingolfin (approved):
Great, this is looking pretty nice now! I have one minor concern about the test which includes network access (CC @ThomasBreuer @thofma @benlorenz )

test/Groups/libraries.jl
ThomasBreuer: Would it make sense to introduce a global `offline` flag for the tests, meaning that if the flag is set to `true` then all those tests that (may) require network access are not executed?

Then we would get an overview of those situations that are expected to require network access, and we can exclude such tests from the CI runs.

lgoettgens: I could imagine 3 ways how to handle this

1. Add some global bool `is_offline` to the tests and replace the test by `is_offline || @test ...`
2. Add a macro `@test_if_online` that decides based on either a global bool or some heuristic (e.g. test network access by pinging something) if the test gets executed or not.
3. Add a macro `@test_online` that tries to run the tests and catches whatever exceptions may occur while using the network, and does not fail in the case of such an exception.

Names are of course open to discussion. What would be your preferred way?

fingolfin: I am open to all. Let's ask @thofma @fieker @benlorenz if they have thoughts on this, too?

thofma: I lean towards 3. (with a `@warning` or `@info` when the exceptions occur), but have no strong opinion.

Although things will get messy once people start writing (and testing) functions, which might require indirectly network access. It is not obvious to a caller which function might have this problem.

benlorenz: Whatever way this is implemented, I would prefer not to run this by default for users testing this (even with a network check, I might be using a limited mobile hotspot since my office doesn't have WiFi...). For github actions CI we may run this but I think this should not fail the CI for a pull request.

But this leads to the question when would this really run and how would it be noticed if it fails? Another separate job which will only run the network dependent jobs and is allowed to fail?

There is the possibility to set a warning in github actions, this would add a small icon to the job: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message

I think the NanoSoldier environment does allow network access, but again I would prefer if it does not fail due to network issues.

src/GAP/utils.jl
```diff
@@ -0,0 +1,11 @@
+function _write_gap_file(relpath::String, content::String)
+  gaproot = GAP.Setup.gaproot()
+  path = joinpath(gaproot, relpath)
+  mkpath(dirname(path))
+  # write to a temporary file and move later to avoid race conditions
+  tmpfile = tempname(; cleanup=false)
+  open(tmpfile, "w") do file
+    write(file, content)
+  end
+  Base.Filesystem.mv(tmpfile, path; force=true)
```
benlorenz:
```suggestion
  tmpfile = tempname(dirname(path); cleanup=false)
  open(tmpfile, "w") do file
    write(file, content)
  end
  Base.Filesystem.rename(tmpfile, path)
```
To make this race-condition free you need to put the temporary file in the same directory as the target file (the same filesystem would suffice but directory is easier) and then use the undocumented `rename` function.

Artifacts.toml
```diff
@@ -0,0 +1,7 @@
+[gap_hulpke_extraperfect]
+git-tree-sha1 = "9ef6636e5ca29fea6e6513aec2c977ab5c1b89cd"
+lazy = true
+
+    [[gap_hulpke_extraperfect.download]]
+    sha256 = "2190f2d128f55194b7afc905fef7d77e940a196c102fb3ecdd456813a4f62670"
+    url = "https://github.com/hulpke/extraperfect/archive/80c0cf27bf98ba3eabd287db5ab4fd0f249e0559.tar.gz"
```
benlorenz: Please do not use an archive URL with a hash, it is not stable since the compression settings might change in the future: https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#stability-of-source-code-archives
Only the tree-hash is guaranteed to be stable, not the hash of the archive.

The recommended way is to create a release and upload a tarball as a release asset.

lgoettgens: Ok. So I would change this here using a fixed release asset name for the v0.12.2 release, provide a slightly modified tarball (I would deflate it one level) to someone with write access and then we do a v0.12.2 shortly after merging this?
Or would it be easier to add the tarball to the v0.12.1 release so that we do not break master between merging this and the release?.

benlorenz: On your repository (https://github.com/hulpke/extraperfect) go to the releases page (https://github.com/hulpke/extraperfect/releases) and create a new release. You can choose whatever name and description you want, and in this form upload any (compressed) tarball (e.g. you could download the current tarball and re-upload it there).
Once the release is published you can copy the url of the release asset, it should look like:
`https://github.com/hulpke/extraperfect/releases/download/<releasename>/<tarballname.tar.gz>`

(No need for anything related to oscar releases, just make sure it is not a github-auto-generated tarball with this `.../archive/...` url, these are created on demand and may change)

benlorenz: Sorry, I totally missed that this is not your repository. We could indeed put it in some Oscar release asset but this seems somewhat strange. Maybe we could put it somewhere on the oscar website @fingolfin ?

lgoettgens: > Sorry, I totally missed that this is not your repository. We could indeed put it in some Oscar release asset but this seems somewhat strange.

This is how Hecke does it, see e.g. https://github.com/thofma/Hecke.jl/releases/tag/v0.13.4
Thus I proposed to do it similarly.

benlorenz: I see, ok, fine with me then, just give it a reasonable name.

lgoettgens: Ok, so in Hecke, the assets just got added to the newest release as of the time of merge. So in our case, that would be Oscar.jl v0.12.1.
I will prepare the tarball, send it to @fingolfin to put in the v0.12.1 release, and then update this PR accordingly.

lgoettgens: > I see, ok, fine with me then, just give it a reasonable name.

After thinking about it for a bit: The https://github.com/hulpke/extraperfect repo does not have any license. By providing a tarball as a release asset in the Oscar repo, we would redistribute some unlicensed code. I am not that firm with copyright law, but that still seems a bit whacky to me.
Giving some artifact URL on hulpke's namespace seems a lot less weird and risky to me. Maybe I/me/someone/... should reach out to hulpke (there is a mail address on his github profile) and ask him if he could make a release, or put a license in the repo, or ....
(no @-mentioning him by on purpose)

benlorenz: Mirroring should be fine, both files do have a license header:
```
##  This file is part of GAP, a system for computational discrete algebra.
##  It contains the perfect groups of order 1376256
##  This data was computed by Alexander Hulpke
##  It is distributed under the artistic license 2.0
##  https://opensource.org/licenses/Artistic-2.0
```
and
```
##  This file is part of GAP, a system for computational discrete algebra.
##  It contains the perfect groups of order 1966080
##  This data was computed by Alexander Hulpke
##  It is distributed under the artistic license 2.0
##  https://opensource.org/licenses/Artistic-2.0
```

_Edit_: It would still be nice if this is was mentioned on the repository as well.

test/Groups/libraries.jl
fingolfin: In order to not let this patch bit-rot, I suggest we comment this test out for now, file an issue to remind us about the possibility for optional online tests.

Artifacts.toml
fingolfin: Ah, I missed this bit about mirroring. OK, I'll look into that as well.

fingolfin: OK, I uploaded https://github.com/oscar-system/GAP.jl/releases/download/archive-tag/extraperfect.tar.gz

I created it like this, for maximum reproducibility:
```
#!/bin/sh
export SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)
rm -rf extraperfect
mkdir extraperfect
gzip -9 -n -c perf27.grp > extraperfect/perf27.grp.gz
gzip -9 -n -c perf33.grp > extraperfect/perf33.grp.gz
gtar --mtime="$(git log -1 --pretty=%cI)" --owner=0 --group=0 --numeric-owner -cvf extraperfect.tar extraperfect
gzip -n -c extraperfect.tar > extraperfect.tar.gz
```
Note that I compressed the two `.grp` files on purpose -- this saves a little bit of storage for the user, and is handled transparently by GAP.

lgoettgens:
I updated the artifact location, included the generation script as a comment in the `Artifacts.toml`, and commented out the test in favor of https://github.com/oscar-system/Oscar.jl/issues/2480.

test/Groups/libraries.jl
lgoettgens: See https://github.com/oscar-system/Oscar.jl/issues/2480.

benlorenz (approved):

benlorenz:
nightly failure is: https://github.com/oscar-system/GAP.jl/issues/901 / https://github.com/oscar-system/Oscar.jl/issues/2476
doctest-nightly failure is: https://github.com/Nemocas/Nemo.jl/issues/1497

fingolfin (approved):

PR #2433, opened 2023-05-29
Change argument order in `factor` and `roots`

lgoettgens:
Implements the changes necessary for https://github.com/thofma/Hecke.jl/pull/1094 and https://github.com/Nemocas/AbstractAlgebra.jl/pull/1374.


fingolfin:
So we get this error, which I believe is because we first need a new Nemo release (and possibly AA release) with the changes there; is that right?
```
Experimental.galois_group: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/galois.jl:1
  MethodError: no method matching factor(::Nemo.fpField, ::QQPolyRingElem)
```

PR #2436, opened 2023-05-31
cache automorphism groups of number fields

fieker:
and related:
 - aut grp local fields
 - IdelClassGmodule
 - allow for use of lll-order by changeing ideal parents
 - galois group via cohomolgy: make sure the rcf is normal

### DIFF
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -920,6 +920,9 @@ function hom_base(C::_T, D::_T) where _T <: GModule{<:Any, <:Generic.FreeModule{
       #bad prime...
       continue
     end
+    if length(t[1]) == 0
+      return []
+    end
     pv = [findfirst(!iszero, x) for x = t[1]]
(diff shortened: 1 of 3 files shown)


fingolfin (approved):
Fine by me, though I note that none of the new code is exercised by any tests -- because [almost none of the existing code in this file has code coverage](https://app.codecov.io/gh/oscar-system/Oscar.jl/blob/master/experimental/GModule/GaloisCohomology.jl)

Since this is in experimental, that's OK, though I note that a few simple tests would help prevent us from accidentally breaking this code

thofma:
Hm, why `(PermGroup,)` and not just `PermGroup` as the key?

fieker:
On Fri, Jun 02, 2023 at 04:06:47AM -0700, Tommy Hofmann wrote:
> Hm, why `(PermGroup,)` and not just `PermGroup` as the key?
So it's a tuple
either
  PermGroup,    of length 1
or
  PermGroup, field of length 2 if one needs a relative aut grup

PR #2437, opened 2023-06-01
Make abelian closure respect unicode setting

thofma:


src/Rings/AbelianClosure.jl
```diff
@@ -150,22 +150,33 @@ parent(::QQAbElem{NfAbsNSElem}) = _QQAb_sparse
 #
 ################################################################################
 
-_variable(K::QQAbField) = K.s
+function _variable(K::QQAbField)
+  if isdefined(K, :s)
+    return K.s
+  else
+    if Oscar.is_unicode_allowed()
+      return "ζ"
```
ThomasBreuer, line 158: If we want to avoid the explicit `"ζ"`  in the code then we can also write `transcode(String, [UInt16(950)])`.
(The `950` can be computed as `Int(codepoint('ζ'))`.)

ThomasBreuer (approved):
Looks good, thanks.
Thus the "fancy" output shall be avoided not only in the code but also in docstrings.

src/Rings/AbelianClosure.jl
fingolfin: I think this should be fine as-is. At least we have similar things in other places.

PR #2439, opened 2023-06-01
Adv/meta data

antonydellavecchia:
A start on storing meta data for serialized objects

### DIFF
--- a/src/Serialization/main.jl
+++ b/src/Serialization/main.jl
@@ -25,10 +25,15 @@ end
 
 const backref_sym = Symbol("#backref")
 
-struct MetaData
-    author::Union{String, Vector{String}}
-    name::String
-    email::String
+@Base.kwdef struct MetaData
+    author::Union{String, Vector{String}, Nothing} = nothing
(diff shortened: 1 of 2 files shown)


src/Serialization/main.jl
```diff
@@ -26,6 +25,17 @@ end
 
 const backref_sym = Symbol("#backref")
 
+@Base.kwdef struct MetaData
+    author::Union{String, Vector{String}, Nothing} = nothing
+    name::Union{String, Nothing} = nothing
+    email::Union{String, Nothing} = nothing
+    affiliation::Union{String, Nothing} = nothing
```
lkastner: I propose to condense this into
```suggestion
    author_orcid::Union{String, Nothing} = nothing
    name::Union{String, Nothing} = nothing
```
The reason is that email and affiliation often change for researchers, while the orcid stays constant. Most researchers have an orcid nowadays and otherwise it's not hard to set up. If we kept email and affiliation we will very fast have data where the author cannot be contacted.

lkastner (approved):

PR #2442, opened 2023-06-02
fix docstrings

fieker:


fingolfin:
As far as I can tell, doctests pass fine on master right now, and this PR breaks them, so I am unsure what this is about?

thofma:
Requires a (breaking) Hecke release.

PR #2448, opened 2023-06-02
First stab at more functionality for Laurent polynomial rings

thofma:
To get started with some ideal and map functionality for Laurent polynomial rings (multivariate).

The main application so far is to do the following:
```
julia> QQxy, (x, y) = LaurentPolynomialRing(QQ, [:x, :y]);

julia> I = ideal(QQxy, [1 + x^-1, y + 1]);

julia> R, (xx, yy) = polynomial_ring(QQ, ["xx", "yy"]);

julia> f = hom(R, QQxy, gens(QQxy));

julia> preimage(f, I)
ideal(yy + 1, xx + 1)
```

This at the moment a bit too slow for the application, but I probably did something not too clever. What we need to do: Given an ideal $I$ of $K[x_1^{\pm},\dotsc,x_n^{\pm}]$ determine $I \cap K[x_1,\dotsc,x_n]$. We can assume that $I = \langle f_1,\dotsc,f_n \rangle$ with $f_i \in K[x_1,\dotsc,x_n]$. At the moment we do
1. Let $\tilde I = \langle f_1,\dotsc, f_n \rangle \subseteq K[x_1,\dotsc,x_n]$ the ideal generated in the polynomial ring.
2. Saturate $\tilde I$ at $x_1,\dotsc,x_n$.

At the moment I am doing step 2. iteratively, but there is problem a better way? Happy to hear what the commutative algebra wizards have to say (@ederc @wdecker).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,7 +23,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 
 [compat]
-AbstractAlgebra = "0.30.7"
+AbstractAlgebra = "0.30.9"
 AlgebraicSolving = "0.3.0"
 DocStringExtensions = "0.8, 0.9"
 GAP = "0.9.4"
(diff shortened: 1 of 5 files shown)


thofma:
I think this is a good start and can be merged. We will see how much it will benefit from the improvements to the saturation.

fingolfin (approved):
Looks good to me but of course @wdecker should ideally also have a look...? Or we just merge and let him complain later ;-)

PR #2450, opened 2023-06-05
Address TODO for finite algebras as vector spaces.

HechtiDerLachs:
@afkafkafk13 @thofma

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -1066,7 +1066,7 @@ function vector_space(K::AbstractAlgebra.Field, Q::MPolyQuoRing)
   # The inverse function. We use the fact that for a chosen monomial ordering 
   # the monomials which are not in the leading ideal, form a basis for the 
   # quotient; see Greuel/Pfister "A singular introduction to Commutative Algebra".
-  function prim(a::MPolyQuoElem)
+  function prim(a::MPolyQuoRingElem)
     @assert parent(a) === Q
     b = lift(a)::MPolyElem
     o = default_ordering(R)
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -1041,13 +1041,11 @@ function _kbase(Q::MPolyQuoRing)
   G = singular_origin_groebner_basis(Q)
   s = Singular.kbase(G)
   if iszero(s)
-    error("ideal was no zero-dimensional")
+    error("ideal is not zero-dimensional")
   end
   return [base_ring(Q)(x) for x = gens(s)]
 end
 
-#TODO: the reverse map...
-# problem: the "canonical" reps are not the monomials.
 function vector_space(K::AbstractAlgebra.Field, Q::MPolyQuoRing)
   R = base_ring(Q)
   @assert K == base_ring(R)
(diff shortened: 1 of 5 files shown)


src/Rings/mpolyquo-localizations.jl
```diff
@@ -1966,3 +1966,97 @@ function (f::Oscar.MPolyAnyMap{<:MPolyRing, <:MPolyQuoLocRing, <:MPolyQuoLocaliz
...
+  # the monomials which are not in the leading ideal, form a basis for the 
+  # quotient; see Greuel/Pfister "A singular introduction to Commutative Algebra".
+  function prim(f::MPolyQuoLocRingElem)
+    @assert parent(f) === W
+    isone(denominator(f)) || error("handling of non-trivial denominators not implemented")
+    b = lifted_numerator(f)
+    @show b
+    b = shift(b)
+    @show b
+    # TODO: The normal form command seems to do something unexpected here.
```
HechtiDerLachs, line 2042: @afkafkafk13 : I tried to implement the use of local orderings here. However, things do not go as I expected here. I'm not sure whether this is due to a bug in the normal form, or a wrong understanding of the matter on my side. Could you take the code from the (broken) tests and have a look here?

src/Rings/MPolyQuo.jl
```diff
@@ -1056,15 +1054,36 @@ function vector_space(K::AbstractAlgebra.Field, Q::MPolyQuoRing)
...
+  # The inverse function. We use the fact that for a chosen monomial ordering 
+  # the monomials which are not in the leading ideal, form a basis for the 
+  # quotient; see Greuel/Pfister "A singular introduction to Commutative Algebra".
+  function prim(a::MPolyQuoRingElem)
+    @assert parent(a) === Q
+    b = lift(a)::MPolyElem
+    o = default_ordering(R)
+    # TODO: Make sure the ordering is the same as the one used for the _kbase above
+    @assert is_global(o) "ordering must be global"
+    b = normal_form(b, modulus(Q), ordering=o)
```
afkafkafk13, line 1075: It seems to me that you need to ensure complete reduction here, but I do not see it.
Does normal_form ensure complete reduction?

HechtiDerLachs: Indeed. The normal form command probably has a corresponding keyword. But the more important thing is the case with local orderings.

HechtiDerLachs: There is no such keyword as `with_complete_reduction` for the `normal_form` command. It seems that we need some more functionality there; maybe even something like `weak_normal_form`? 

Edit: Or just work directly at the singular side again. But I don't know how to do that anymore since all the undocumented changes with the `IdealGens`, etc.

afkafkafk13: You might want to use reduce here. According to its docstring (in src/Rings/groebner.jl), it should ensure complete reduction in the global case.

src/Rings/MPolyQuo.jl
```diff
@@ -1056,15 +1054,36 @@ function vector_space(K::AbstractAlgebra.Field, Q::MPolyQuoRing)
...
   end
-  return V, MapFromFunc(im, V, Q)
+
+  # The inverse function. We use the fact that for a chosen monomial ordering 
+  # the monomials which are not in the leading ideal, form a basis for the 
+  # quotient; see Greuel/Pfister "A singular introduction to Commutative Algebra".
+  function prim(a::MPolyQuoRingElem)
+    @assert parent(a) === Q
+    b = lift(a)::MPolyElem
+    o = default_ordering(R)
```
afkafkafk13, line 1072: Suggestion: Please set ordering somewhere around line1053 and already do a kbase w.r.t. this ordering. If this is not available yet, compute a leading_ideal w.r.t. the ordering and a kbase thereof (for a monomial ideal, the ordering is obviously irrelevant). Then use the same ordering in 1072.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1966,3 +1966,97 @@ function (f::Oscar.MPolyAnyMap{<:MPolyRing, <:MPolyQuoLocRing, <:MPolyQuoLocaliz
   g = get_attribute(f, :lifted_map)
   return codomain(f)(g(a), check=false)
 end
+
+function vector_space(kk::Field, W::MPolyQuoLocRing;
+    ordering::MonomialOrdering=degrevlex(gens(base_ring(W)))
```
afkafkafk13, line 1971: If you are in MPolyQuoLocRing, then this should not be the default. Maybe negdegrevlex ?

HechtiDerLachs: No. This is correct.

src/Rings/MPolyQuo.jl
HechtiDerLachs: The choice of ordering is hidden somewhere in the internals of code that I didn't program. It would be good if whoever did that could just quickly confirm that what I'm doing here is correct.

HechtiDerLachs:
From my side this is good to go. Please merge.

afkafkafk13 (approved):
Looks good now -- with the obvious caveat that normal_form is broken for local orderings. The workaround used here is not pretty, as it works properly, but slowly.

PR #2454, opened 2023-06-06
Trivial cleanup in hilbert_series code

fingolfin:


src/Rings/mpoly-graded.jl
```diff
@@ -1399,27 +1399,27 @@ mutable struct HilbertData
   end
 end
 
-function hilbert_series(H::HilbertData, i::Int= 1)
+function hilbert_series(H::HilbertData)
   Zt, t = ZZ["t"]
```
fieker, line 1403: General comment: in "library code" one should never create global rings (cached = true), they should always be created with `cache = false`, ie. `Zt, t = polynomial_ring(ZZ, :t, cached = false)` 
Same is true for the Qt in the other function.
Try to do
```
Hecke.print_cache()
```
to get an idea how many of this rings/ fields are stored...
If you really want a global univariate polyring over ZZ or QQ, try `Hecke.Global.Zx` or `Hecke.Globals.Qx`

src/Rings/mpoly-graded.jl
```diff
@@ -1399,27 +1399,27 @@ mutable struct HilbertData
...
-    return den(0)*h, den(0)*den
-  end
-  error("2nd parameter must be 1 or 2")
+  h = Zt(map(ZZRingElem, H.data[1:end-1]))
+  return h, den
+end
+
+function hilbert_series_reduced(H::HilbertData)
+  h, den = hilbert_series(H)
+  c = gcd(h, den)
```
fieker: There is a cool function `gcd_with_cofactors` that is, for some rings, much more efficient...

fieker: How does this relate to the recent idea of storing the num/ den in factored form?

src/Rings/mpoly-graded.jl
```diff
@@ -1399,27 +1399,27 @@ mutable struct HilbertData
...
+  h = divexact(h, c)
+  den = divexact(den, c)
+  return den(0)*h, den(0)*den
```
fingolfin, line 1412: @wdecker Is there a specific reason to multiply by `den(0)` here, i.e., the constant term of the denominator, i.e., either 1 or -1 ?

src/Rings/mpoly-graded.jl
fingolfin: Indeed. Actually, shouldn't the polynomial ring be an (optional) argument?

src/Rings/mpoly-graded.jl
fingolfin: Yes, `gcd_with_cofactors` is better here.

This is unrelated to the recent discussions.

src/Rings/mpoly-graded.jl
wdecker: yes! The constant term of the denominator should be one.

src/Rings/mpoly-graded.jl
JohnAAbbott: An idea we considered for CoCoA was to store a suitable "Hilbert series ring" as a data-member of each graded polynomial ring.  This avoids asking the caller to specify the ring for the Hilbert series, and also ensures that Hilbert series of quotients of the same graded ring belong to the same Hilbert series ring.  This idea has **not yet** been implemented in CoCoA... so maybe there are unforeseen hitches.

wdecker:
Already the theoretical definition of such a ring requires some thoughts. See the
books by Miller-Sturmfels and Kreuzer-Robbiano for suggestions. At the moment
I do not see the necessity to think about an implementation. We should provide
the functionality as discussed asap and then move on to new adventures.

PR #2456, opened 2023-06-07
Cleanup of `Groups/directproducts.jl`

### LINKED ISSUE #1718: Some cleanup for the groups code
instead of comparing `typeof` it is better to use `isa`, and to check that something holds for all elements of a list, use the `all` function (which short-circuits). 
Consider this:
```
function as_perm_group(G::DirectProductGroup)
    if [typeof(H)==PermGroup for H in G.L]==[true for i in 1:length(G.L)]
```
This could be replaced by something like this (untested)
```suggestion
function as_perm_group(G::DirectProductGroup)
   if all(H -> H isa PermGroup, G.L)
```

_Originally posted by @fingolfin in https://github.com/oscar-system/Oscar.jl/pull/1661#discussion_r1018852955_

lgoettgens:
This PR includes the following changes to the file `Groups/directproducts.jl`:

- Refactor the code mentioned in https://github.com/oscar-system/Oscar.jl/issues/1718 using `@req` and the proposal there (thus closes https://github.com/oscar-system/Oscar.jl/issues/1718).
- Some more uses of `@req`
- Add some checks for empty input lists
- Make functions with vararg params unambiguous by explicitly expecting at least one object
- Run formatter

### DIFF
--- a/src/Groups/directproducts.jl
+++ b/src/Groups/directproducts.jl
@@ -44,6 +44,7 @@ julia> elements(G)
 ```
 """
 function direct_product(L::AbstractVector{<:GAPGroup}; morphisms=false)
+  @req length(L) > 0 "the collection of groups must be non-empty"
   X = GAP.Globals.DirectProduct(GapObj([G.X for G in L]))
   DP = DirectProductGroup(X, L, X, true)
   if morphisms
@@ -57,8 +58,8 @@ function direct_product(L::AbstractVector{<:GAPGroup}; morphisms=false)
   end
 end
 
-function direct_product(L::GAPGroup...; morphisms=false)
-  return direct_product([x for x in L]; morphisms=morphisms)
+function direct_product(L::GAPGroup, Ls::GAPGroup...; morphisms=false)
+  return direct_product([L, Ls...]; morphisms=morphisms)
 end
 
 """
@@ -76,6 +77,7 @@ vectors of the embeddings (resp. projections) of the direct product `G`.
 function inner_direct_product(
   L::AbstractVector{T}; morphisms=false
 ) where {T<:Union{PcGroup,FPGroup}}
+  @req length(L) > 0 "the collection of groups must be non-empty"
   P = GAP.Globals.DirectProduct(GapObj([G.X for G in L]))
   DP = T(P)
   if morphisms
@@ -90,6 +92,7 @@ function inner_direct_product(
 end
 
 function inner_direct_product(L::AbstractVector{PermGroup}; morphisms=false)
+  @req length(L) > 0 "the collection of groups must be non-empty"
   P = GAP.Globals.DirectProductOfPermGroupsWithMovedPoints(
     GapObj([G.X for G in L]), GAP.Obj([collect(1:degree(G)) for G in L]; recursive=true)
   )
@@ -106,9 +109,9 @@ function inner_direct_product(L::AbstractVector{PermGroup}; morphisms=false)
 end
 
 function inner_direct_product(
-  L::T...; morphisms=false
+  L::T, Ls::T...; morphisms=false
 ) where {T<:Union{PcGroup,PermGroup,FPGroup}}
-  return inner_direct_product([x for x in L]; morphisms=morphisms)
+  return inner_direct_product([L, Ls...]; morphisms=morphisms)
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/Groups/directproducts.jl
```diff
@@ -71,33 +74,44 @@ The parameter `morphisms` is `false` by default. If it is set `true`, then
...
-   DP = T(P)
-   if morphisms
-      emb = [GAPGroupHomomorphism(L[i],DP,GAP.Globals.Embedding(P,i)) for i in 1:length(L)]
-      proj = [GAPGroupHomomorphism(DP,L[i],GAP.Globals.Projection(P,i)) for i in 1:length(L)]
-      return DP, emb, proj
-   else
-      return DP
-   end
+function inner_direct_product(
+  L::AbstractVector{T}; morphisms=false
```
fingolfin:
```suggestion
  L::AbstractVector{T}; morphisms::Bool=false
```
Maybe also adjust the docstring to refer to `morphisms` as "keyword argument" instead of "parameter" ?

src/Groups/directproducts.jl
```diff
@@ -71,33 +74,44 @@ The parameter `morphisms` is `false` by default. If it is set `true`, then
...
-   P = GAP.Globals.DirectProductOfPermGroupsWithMovedPoints(GapObj([G.X for G in L]), GAP.Obj([collect(1:degree(G)) for G in L], recursive = true))
-   DP = PermGroup(P, sum([degree(G) for G in L], init = 0))
-   if morphisms
-      emb = [GAPGroupHomomorphism(L[i],DP,GAP.Globals.Embedding(P,i)) for i in 1:length(L)]
-      proj = [GAPGroupHomomorphism(DP,L[i],GAP.Globals.Projection(P,i)) for i in 1:length(L)]
-      return DP, emb, proj
-   else
-      return DP
-   end
+function inner_direct_product(L::AbstractVector{PermGroup}; morphisms=false)
```
fingolfin:
```suggestion
function inner_direct_product(L::AbstractVector{PermGroup}; morphisms::Bool=false)
```

src/Groups/directproducts.jl
```diff
@@ -71,33 +74,44 @@ The parameter `morphisms` is `false` by default. If it is set `true`, then
...
+  else
+    return DP
+  end
 end
 
-
-function inner_direct_product(L::T... ; morphisms=false) where T<:Union{PcGroup,PermGroup,FPGroup}
-   return inner_direct_product([x for x in L]; morphisms=morphisms)
+function inner_direct_product(
+  L::T, Ls::T...; morphisms=false
```
fingolfin:
```suggestion
  L::T, Ls::T...; morphisms::Bool=false
```

src/Groups/directproducts.jl
```diff
@@ -44,19 +44,22 @@ julia> elements(G)
...
+    return DP, emb, proj
+  else
+    return DP
+  end
 end
 
-function direct_product(L::GAPGroup...; morphisms=false)
-   return direct_product([x for x in L]; morphisms=morphisms)
+function direct_product(L::GAPGroup, Ls::GAPGroup...; morphisms=false)
+  return direct_product([L, Ls...]; morphisms=morphisms)
```
ThomasBreuer, line 62: Just for curiosity:
Why is the additional `GAPGroup` argument needed?
When the `direct_product(L::GAPGroup...)` method gets called then `L` is automatically nonempty.
(I could think of the idea that one wants to admit `direct_product()` via a dedicated method.)

ThomasBreuer (approved):
Looks good. Thanks.

src/Groups/directproducts.jl
lgoettgens: When you currently try to call `direct_product()` you get an ambiguity error, since all of the dispatches of the form `direct_product(::SomeType...)` are equally good.

This can only be resolved by changing all of these to take at least one argument of their preferred type.

src/Groups/directproducts.jl
lgoettgens: Done in a65db37ee2740413118425c901feb2c665d30035

src/Groups/directproducts.jl
lgoettgens: Done in a65db37ee2740413118425c901feb2c665d30035

src/Groups/directproducts.jl
lgoettgens: Done in a65db37ee2740413118425c901feb2c665d30035

fingolfin (approved):

fingolfin:
Thank you!

PR #2458, opened 2023-06-09
Add absolute_automorphism_group(::Type{PermGroup}, ...)

thofma:


fieker (commented):
what happened to the caching introduced in #2436?

fieker (approved):
we'll have to sort out the caching, but I'd suggest to do this independently

fieker:
There is (?) a missing elem_type for sets of morphisms?

thofma:
Yes, I am on it.

PR #2461, opened 2023-06-12
Two fixes

joschmitt:
Fix two bugs and add corresponding tests.


joschmitt:
I don't think the failing test is my fault and I can't reproduce it locally. Looks scary though.

fingolfin (approved):

PR #2463, opened 2023-06-12
[TropicalGeometry] Fix weird looking docs

lkastner:
f6da3dff5ed made these two "formulas" centered.
Could it be that `$bla$` at the beginning of a line is always centered? @fingolfin is there a preferred way to prevent this?


fingolfin (approved):
I have no idea what's going on there. If this fixes a problem, by all means let's merge it

benlorenz:
Current dev (master):
dev

With this PR:
pr

PR #2466, opened 2023-06-13
Remove duplicate import of `MPolyRingElem`

joschmitt:
... another small fix (sorry!).
I don't really know about the history, but I assume this originated in the big renaming. (In case you wonder: `MPolyRingElem` is imported in line 93 in `src/imports.jl`.)


fingolfin (approved):

fingolfin:
Thank you!

PR #2467, opened 2023-06-13
Add lowercase constructor for `Partition`

### LINKED ISSUE #2298: `Partition` needs a `partition` method
We are missing the lowercase version. Maybe an alias? I forgot how we do this for types.

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/2298.

### DIFF
--- a/experimental/IntersectionTheory/src/Misc.jl
+++ b/experimental/IntersectionTheory/src/Misc.jl
@@ -63,10 +63,10 @@ function partitions(n::Int, k::Int=n, m::Int=-1)
   ans = Partition[]
   (n > 0 && k == 0) && return ans
   if m < 0 m = n end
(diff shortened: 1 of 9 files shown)


fingolfin (approved):
Fine by me

ulthiel (commented):
Basically ok, I'm just wondering whether all types are now supposed to be lower case. Here, Partition is really changed everywhere to partition, so there's no Partition anymore!?

lgoettgens:
> Basically ok, I'm just wondering whether all types are now supposed to be lower case. Here, Partition is really changed everywhere to partition, so there's no Partition anymore!?

As far as I understood the big renaming, there should be a distinction between types (Uppercase) and constructors (lowercase), since the latters should be handled as "normal" functions

ulthiel (approved):

thofma:
Thanks Lars!

Issue #2155, opened 2023-03-28
standard_basis for modules accepts monomial ordering, but silently ignores it


afkafkafk13:
standard_basis accepts a keyword argument ordering containing a module ordering (even based on a local or mixed ring ordering), but ignores it

fragment of Oscar session:
```
julia> r
Multivariate Polynomial Ring in y, z over Rational Field

julia> o1
negdegrevlex([y, z])*lex([gen(1), gen(2), gen(3)])

julia> Jm3
Submodule with 2 generators
1 -> 2*e[1] - 3*y^2*e[2] - 6*z^2*e[3]
2 -> 2*e[1] + 3*y^2*e[2] - 3*z^2*e[3]

julia> standard_basis(Jm3;ordering=o1)
Gröbner basis with elements
-4*e[1] + 9*z^2*e[3]
2*e[1] + 3*y^2*e[2] - 3*z^2*e[3]
 with respect to the ordering
negdegrevlex([y, z])*lex([gen(1), gen(2), gen(3)])
```

Explanation:
As the first comparison is on some negative weighted degree ordering, the leading monomial of each of the two generators is 2*e[1] and the first spoly should give rise to somthing with leading monomial y^2*e[2]. The above output shows clearly that this is not the case. Obviously some global ordering had been applied.

**System (please complete the following information):**
```
julia> Oscar.versioninfo(full=true)
OSCAR version 0.12.0-DEV - #notions_of_transforms, d77776aa -- 2023-03-27 09:27:14 +0200
  combining:
    AbstractAlgebra.jl   v0.28.2
    GAP.jl               v0.9.4
    Hecke.jl             v0.18.3
    Nemo.jl              v0.33.1
    Polymake.jl          v0.9.0
    Singular.jl          v0.17.0
  building on:
    Antic_jll               v0.201.500+0
    Arb_jll                 v200.2300.0+0
    Calcium_jll             v0.401.100+0
    FLINT_jll               v200.900.4+0
    GAP_jll                 v400.1200.200+0
    Singular_jll            v403.101.500+0
    libpolymake_julia_jll   v0.9.0+0
    libsingular_julia_jll   v0.29.0+0
    polymake_jll            v400.900.0+0
See `]st -m` for a full list of dependencies.

Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
[...]

PR #2468, opened 2023-06-13
Du val sing detector

afkafkafk13:
Testing for duVal singularities

includes several smaller fixes necessary on the way:
- vdim for local setting (and removal of vdim-hack)
- Ideal * FreeModule and Ideal*SubModuleOfFreeModule
- fix for issue #2155
- better support for localized modules at maximal ideals


afkafkafk13:
Tests and documentation still to come.
Sorry for the untidy history -- it will go away after with the squash and merge at the end.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1678,6 +1678,13 @@ end
 
 function saturated_ideal(I::MPolyQuoLocalizedIdeal{LRT}) where {LRT<:MPolyQuoLocRing{<:Any, <:Any, <:Any, <:Any, <:MPolyPowersOfElement}}
   return saturated_ideal(pre_image_ideal(I),strategy=:iterative_saturation,with_generator_transition=false)
+end 
+
+function vdim(R::MPolyQuoLocRing)
```
HechtiDerLachs: Signature should be restricted to the case of localization at a K-point.

src/Modules/mpoly-localizations.jl
```diff
@@ -176,4 +176,19 @@ the shift map ``Φ : R → R`` which is moving the point of ``𝔪`` to the orig
...
+@attr function shifted_module(
+    M::SubModuleOfFreeModule{T}
+  ) where {T<:MPolyLocRingElem{<:Field, <:FieldElem, <:MPolyRing, <:MPolyRingElem,
+                               <:MPolyComplementOfKPointIdeal}}
 
+  R = base_ring(M)::MPolyLocRing
+  P = base_ring(R)::MPolyRing
+  FL = ambient_free_module(M)
+  F = base_ring_module(FL)
+  (A,D) = clear_denominators(M.matrix)
```
HechtiDerLachs, line 199: One should not access the internals of a module. Could you check whether something like `matrix(M)` exists and does the right job?

src/Modules/UngradedModules.jl
```diff
@@ -1539,8 +1543,17 @@ function is_canonically_isomorphic(M::SubModuleOfFreeModule, N::SubModuleOfFreeM
   return SubModuleOfFreeModule(G, [f(v) for v in gens(M)]) == N
 end
 
-#+(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule) = sum(M, N)
+Base.:+(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule) = sum(M, N)
+
+function Base.:*(I::MPolyIdeal{T}, M::SubModuleOfFreeModule{T}) where {T<: MPolyRingElem}
+  F = ambient_free_module(M)
+  newgens = filter( x -> !is_zero(x),[a*b for a in gens(I) for b in gens(M)])
+  return oscar.SubModuleOfFreeModule(F, newgens)
```
HechtiDerLachs: The `oscar.` should not be necessary.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,225 @@
+function has_du_val_singularities(F::MPolyDecRingElem)::Bool
```
HechtiDerLachs: It seems possible that one could also return a full list with all singularities found. Wouldn't that be nice?

HechtiDerLachs (commented):
As was already mentioned, tests are still missing. I think @simonbrandhorst had something to be checked for du Val singularities.

src/Modules/UngradedModules.jl
```diff
@@ -1539,8 +1543,17 @@ function is_canonically_isomorphic(M::SubModuleOfFreeModule, N::SubModuleOfFreeM
   return SubModuleOfFreeModule(G, [f(v) for v in gens(M)]) == N
 end
 
-#+(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule) = sum(M, N)
+Base.:+(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule) = sum(M, N)
+
+function Base.:*(I::MPolyIdeal{T}, M::SubModuleOfFreeModule{T}) where {T<: MPolyRingElem}
```
HechtiDerLachs: I thought these methods existed already. The failure of the tests suggests that the this overrides an existing generic method, but with incompatible output format. Return a map also?

src/Rings/mpolyquo-localizations.jl
afkafkafk13: done

src/Modules/mpoly-localizations.jl
afkafkafk13: If it were available for SubModuleOfFreeModule, I would gladly use it. Helas, it only exists for free modules and subquos.

fieker:
On Wed, Jun 14, 2023 at 04:38:04AM -0700, Anne Frühbis-Krüger wrote:
> @afkafkafk13 commented on this pull request.
> 
> 
> 
> >  
> +  R = base_ring(M)::MPolyLocRing
> +  P = base_ring(R)::MPolyRing
> +  FL = ambient_free_module(M)
> +  F = base_ring_module(FL)
> +  (A,D) = clear_denominators(M.matrix)
The name, elsewhere, would be integral_split
Generically taking an object and a ring, returning an obejct over the
ring and a denominator in the ring....

HechtiDerLachs:
> The name, elsewhere, would be integral_split Generically taking an object and a ring, returning an obejct over the ring and a denominator in the rini

good to know. I might change that at some point. Since this is an internal method, I would suggest to leave this as is for the moment.

src/Modules/mpoly-localizations.jl
HechtiDerLachs: Right. I never worked with `SubModuleOfFreeModule`. Should we maybe introduce a getter here? It might not be safe to access the field as it might not even be set (could be used for caching only). 

@jankoboehm : How should we proceed here?

src/Modules/UngradedModules.jl
afkafkafk13: I renamed my method to solve the problem .

experimental/Schemes/duValSing.jl
afkafkafk13: Easily!
But before implementing that, we would have to agree (among all classifiers, not just this basic one) on an output format for singularity types so that we can output tuples consisting of the point and the singularity.

simonbrandhorst: For DuVal singularities I suggest `((0,0), (:A, 4))`
The format `(:A, n)`  `(:D ,n)`, `(:E ,n)` is what we use for the root lattices.

simonbrandhorst:
The name `vdim` seems to be an abbreviation coming from singular.  Could you please spell it out?
E.g. `vector_space_dimension` ?

experimental/Schemes/duValSing.jl
afkafkafk13: Great. Thanks. I will do this for the hypersurface case -- for the higher codimension case, this will come as soon as a 'vdim' (by whatever name) is available for modules.

afkafkafk13:
> The name `vdim` seems to be an abbreviation coming from singular. Could you please spell it out? E.g. `vector_space_dimension` ?

On one hand, you are perfectly right. On the other hand, @wdecker  has already introduced vdim for the non-localized case. Do you want me to change this as well -- otherwise, we are inconsistent. Maybe we discuss this at the meeting later on today?

experimental/Schemes/duValSing.jl
jankoboehm: The more fancy classifiers return normal forms with all kinds of additional data associated to the singularity. That would be overkill here. So the format is fine.

afkafkafk13:
> > The name `vdim` seems to be an abbreviation coming from singular. Could you please spell it out? E.g. `vector_space_dimension` ?
> 
> On one hand, you are perfectly right. On the other hand, @wdecker has already introduced vdim for the non-localized case. Do you want me to change this as well -- otherwise, we are inconsistent. Maybe we discuss this at the meeting later on today?

As decided in today's meeting: all vdim have been changed to vector_space_dimension.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,227 @@
...
+
+  return true
+end
+
+function _check_duval_at_point(I_r::MPolyIdeal,I_sl_r::MPolyIdeal)::Bool
+    ## initialize
+    r = base_ring(I_r)
+
+    ## localize at point
+    a = Oscar.rational_point_coordinates(I_r)
```
simonbrandhorst:
```suggestion
    a = rational_point_coordinates(I_r)
```

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,227 @@
...
+      J_changed = ideal(r_changed, [change_coefficient_ring(kk,a) for a=gens(Jdehom_dict[i])])
+      _check_duval_at_point(I2,JM_changed,J_changed)
+    end
+  end
+
+  return true
+end
+
+## CAUTION!!!
+## The following needs to be cleaned up properly after vdim for modules is available in Oscar
```
simonbrandhorst: what keeps us from having vdim for modules?

src/Modules/mpoly-localizations.jl
```diff
@@ -176,4 +176,19 @@ the shift map ``Φ : R → R`` which is moving the point of ``𝔪`` to the orig
   return result, a, b
 end
 
+@attr function shifted_module(
```
simonbrandhorst, line 190: please document

src/Modules/UngradedModules.jl
```diff
@@ -1539,8 +1543,18 @@ function is_canonically_isomorphic(M::SubModuleOfFreeModule, N::SubModuleOfFreeM
...
+  F = ambient_free_module(M)
+  newgens = filter( x -> !is_zero(x),[a*b for a in gens(I) for b in gens(M)])
+  return oscar.SubModuleOfFreeModule(F, newgens)
+end
 
+function times(I::MPolyIdeal, M::FreeMod)
```
simonbrandhorst: I would overload the  operator here. @fieker 
```suggestion
`Base.(*)(I::MPolyIdeal, M::FreeMod)`
```

simonbrandhorst:
We should think through the interface carefully here, since it is the first case of a singularity classifier and surely there are more to come?
It seems that at the moment `has_duval_singularities` is available only for some ideals in homogeneous space.
For me this is less a property of an ideal, but of the corresponding scheme it defines. Therefore the interface should be at the scheme/variety side. 
I would suggest to document
`has_duval_singularities(X::Scheme{<:Field})`
and to implement the methods
`has_duval_singularities(X::AbsSpec{<:Field})`
`has_duval_singularities(X::AbsCoveredScheme{<:Field})`
`has_duval_singularities(X::AbsProjectiveScheme{<:Field})`

Further Something like
`is_duval_singularity(X::AbsSpec, rational_point::Vector)`
could check it locally at a point. 
I wonder, is there is a way around taking field extensions?

src/Modules/UngradedModules.jl
afkafkafk13: This breaks something in the ReesAlgebra-Tests and I have not found out yet, how to repair it. So for the moment I used another name for it to get the true content of the PR running first and investigate later on (but before merge!).

afkafkafk13:
> We should think through the interface carefully here, since it is the first case of a singularity classifier and surely there are more to come? It seems that at the moment `has_duval_singularities` is available only for some ideals in homogeneous space. For me this is less a property of an ideal, but of the corresponding scheme it defines. Therefore the interface should be at the scheme/variety side. 

Indeed, I have not finished the interface yet and your suggestions coincide mostly with my ideas -- but I would leave 'has_duval_singularities' also available for graded polynomials and ideals to accomodate for the older generation -- but not prominently documented.

> I would suggest to document `has_duval_singularities(X::Scheme{<:Field})` and to implement the methods `has_duval_singularities(X::AbsSpec{<:Field})` `has_duval_singularities(X::AbsCoveredScheme{<:Field})` `has_duval_singularities(X::AbsProjectiveScheme{<:Field})`

Yes, I agree. Additonally, we should have it for SpaceGerms and as you suggested below for Spec at point.

> Further Something like `is_duval_singularity(X::AbsSpec, rational_point::Vector)` could check it locally at a point. I wonder, is there is a way around taking field extensions?
yes there is, but a lot more care needs to be applied and for this application I prefer the more conceptual way of taking fieldextension -- more over it has become very easy to handle in Oscar.

experimental/Schemes/duValSing.jl
afkafkafk13: Yes, these will go away before moving from draft to review required...

experimental/Schemes/duValSing.jl
afkafkafk13: As I mentioned in our meeting on thursday, I need to discuss a bit with @jankoboehm to fully embrace the philosophy behind the choices made in the modules implementation and to find out how much work it is to provide vdim-functionality -- at least for ungraded modules w.r.t. global ordering, all the rest (except graded slices) can easily be reduced to this situation by passing to a leading module w.r.t. the appropriate ordering. 

So the summary answer to your question is: What keeps me from doing it, is lack of knowledge.

src/Modules/UngradedModules.jl
simonbrandhorst: There is the following already 
`*(I::Ideal, M::ModuleFP) in Oscar at .julia/dev/Oscar/src/Modules/UngradedModules.jl:6500`
Are your modules graded?

simonbrandhorst:
> Indeed, I have not finished the interface yet and your suggestions coincide mostly with my ideas -- but I would leave 'has_duval_singularities' also available for graded polynomials and ideals to accomodate for the older generation -- but not prominently documented.

I have some concerns here.  
- Either it is documented and part of the interface and people use it - or not. 
We should not have anything in between, like not so well documented and hard to find, but part of the interface and some experts use it
- This follows the singular style: speaking algebraic geometry but doing and inputing/outputing commutative algebra. In Oscar we try to follow the (textbook) mathematics more closely. 
- Doing a mixture of both could lead to inconsistencies. The user will have a hard time to decide if they should input ideals or schemes? And what to expect as output? Ideals or Schemes? Of course we could support everything. But that is a lot of work.

In general I would say, if the input is commutative algebra, then the name and the output should be commutative algebra as well. And similar for the schemes. I can see that this division cannot always be strict, but that is what we should strive for.

src/Modules/UngradedModules.jl
afkafkafk13: The problem is not finding this function (which happens to return a subquo), but my lack of knowledge concerning the structural decisions in the design of the (ungraded) modules. I naively wrote the duVal-Tester using SubModuleOfFreeModule -- because this type tab-completed and seemed to encode precisely what I needed to prepare a submodule of the form N + J*M in a free module M. This may have been a wrong decision on my side -- its origin is the intrinsic analogy of the duVal-singularity tester between the hypersurface&ideal case and the non-hypersurface&module case. I will probably have to rewrite everything in terms of sub and quo.

afkafkafk13:
@simonbrandhorst : I took your remarks seriously and rewrote the whole thing. Could you give a rough feedback whether this is the interface you wanted -- before I start documenting and moving stuff to the correct files.

simonbrandhorst:
The new interface looks great! Thanks a lot. 
For now we give a K-rational point by an ideal. That works well, but it is not optimal. 
This indicates that we should think soon about how we want to work with k-rational-points.  We can leave this for now as long as it is in experimental.

afkafkafk13:
> The new interface looks great! Thanks a lot. For now we give a K-rational point by an ideal. That works well, but it is not optimal. This indicates that we should think soon about how we want to work with k-rational-points. We can leave this for now as long as it is in experimental.

For completeness: We talked on the phone and decided to continue to allow ideals as input for points, as the functions can also handle finite sets of points and points over the algebraic closure. For users' convenience a variant taking k-rational point coordinates as input will be added.

src/Modules/mpoly-localizations.jl
afkafkafk13: done

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,513 @@
...
+  cubiccount < 3 || return  (false, typof(smooth))                      ## at least X_9
+  cubiccount > 1 || return  (true, (:D,tau))                            ## D_series
+
+  # it is definitely in the E/J series
+  tau < 9 || return(false, typeof(smooth))                              ## at least J_10
+  return (true, (:E,tau))                                               ## E_6, E_7, E_8
+end
+
+@doc raw"""
+    vector_space_dimension(M::SubquoModule, d::Int)
```
simonbrandhorst: Please move this to
src/Modules/Ungraded_Modules

Notifying @wdecker @fieker @jankoboehm , that we have some new module functionality here.

afkafkafk13:
@jankoboehm : This PR contains vdim- and kbase-functionality for modules under the names vector_space_dimension and vector_space_basis. Now that everything is nearly ready, I put these in src/Modules/UngradedModules.jl.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+  I = modulus(R)
+  
+  J = image_ideal(singular_locus(X)[2])
+  J = ideal(base_ring(I), lift.(gens(J))) + I
+  dim(J) == 0 || return false                            ## non-isolated
+  return is_du_val_singularity(X,J)
+end
+
+function has_du_val_singularities(X::AbsCoveredScheme{<:Field})
+  C = (has_attribute(X, :simplified_covering) ? simplified_covering(X) : default_covering(X))
```
simonbrandhorst, line 41:
```suggestion
  C = simplified_covering(X)
```
Works.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+function is_du_val_singularity(X::AbsSpec{<:Field,<:Any},I::Ideal)
+  OOX = OO(X)
+  dim(X) == 2 || error("Scheme not of dimension 2")
+  J = modulus(OOX)
+  !isone(J) || error("Scheme is empty")
+  !iszero(J) || return true                  ## X smooth
+
+  R = base_ring(I)
+  kk = base_ring(R)
+  characteristic(kk) == 0 || error("only available in characteristic zero")
```
simonbrandhorst: Can we do characteristic $p\geq 5$ ,say?

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+
+  dim(I) == 0 || error("second argument does not describe a 0-dimensional scheme")
+  if get_attribute(I,:is_absolutely_prime, false)
+    return _check_duval_at_point(J,I)[1]
+  end
+
+  decomp = absolute_primary_decomposition(I)
+
+  for (_,_,I2,mult) in decomp
+    set_attribute!(I2,:is_absolutely_prime,true)
```
simonbrandhorst, line 108:
```suggestion
    set_attribute!(I2,:is_absolutely_prime,true)
```
This should be done in `absolute_primary_decomposition` and not here. Please add it to the cache there.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+
+@doc raw"""
+    decide_du_val_singularity(X::AbsSpec, I::Ideal)
+
+Return a vector of tuples T with the following data:
+- T[1]::Bool answers 'X' has at most du Val (surface) singularities at the geometric points
+specified by the ideal 'I'.
+- T[2]::Ideal I_P associated prime of I (possibly over a suitable field extension)
+  describing some geometrically irreducible point
+- T[3]::Tuple type of the singularity at P
```
simonbrandhorst:
```suggestion
- T[3]::Tuple type of the singularity at ``P``  e.g. `(:A, 3)`
```

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+  end
+  
+  return true
+end
+
+@doc raw"""
+    decide_du_val_singularity(X::AbsSpec, I::Ideal)
+
+Return a vector of tuples T with the following data:
+- T[1]::Bool answers 'X' has at most du Val (surface) singularities at the geometric points
```
simonbrandhorst:
```suggestion
- `T[1]::Bool` answers 'X' has at most du Val (surface) singularities at the geometric points
```
and alike

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+      x = (x[1],x[2],x[3],mult)
+      push!(result_vector,x)
+    end
+  end
+
+  return result_vector
+end
+
+@doc raw"""
+    _check_du_val_at_point(IX:Ideal,Ipt::Ideal))
```
simonbrandhorst:
```suggestion
    _check_du_val_at_point(IX:Ideal,Ipt::Ideal)
```

src/Modules/UngradedModules.jl
```diff
@@ -8712,3 +8716,258 @@ function dual(f::ModuleFPHom{<:ModuleFP, <:ModuleFP, Nothing}; # Third parameter
...
+    vector_space_dimension(M::SubquoModule, d::Int)
+
+Let R be a MPolyAnyRing over a field kk and let M be a subquotient module over R.
+Then the command returns the dimension of the kk-vectorspace corresponding to the
+degree d slice of M, where the degree of each variable of R is counted as one and
+the one of each generator of the ambient free module of M as zero.
+
+    vector_space_dimension(M::SubquoModule)
+
+If M happens to be finite-dimensional as a kk-vectorspace, this returns its dimension.
```
```suggestion
Let ``R`` be a `MPolyAnyRing` over a field ``k`` and let ``M`` be a subquotient module over ``R``.
Then the command returns the dimension of the ``k``-vectorspace corresponding to the
degree ``d`` slice of ``M``, where the degree of each variable of ``R`` is counted as one and
the one of each generator of the ambient free module of ``M`` as zero.

    vector_space_dimension(M::SubquoModule)

If ``M`` happens to be finite-dimensional as a ``k``-vectorspace, this returns its dimension.
```
and what if it is infinite dimensional?

src/Modules/UngradedModules.jl
```diff
@@ -8712,3 +8716,258 @@ function dual(f::ModuleFPHom{<:ModuleFP, <:ModuleFP, Nothing}; # Third parameter
...
+  R = base_ring(M)
+  F = ambient_free_module(M)
+  Mq,_ = sub(F,rels(M))
+
+  ambient_representatives_generators(M) == gens(F) || error("not implemented for M/N with non-trivial M")
+
+  o = default_ordering(M)
+  LM = leading_module(Mq,o)
+
+  return length([x*e for x in Oscar.all_monomials(R, d) for e in gens(F) if !(x*e in LM)])
```
```suggestion
  return length([x*e for x in all_monomials(R, d) for e in gens(F) if !(x*e in LM)])
```

simonbrandhorst (commented):
Thank you for this pull request! I have some minor nitpicks. 
Similar comments apply also to other methods.

experimental/Schemes/duValSing.jl
afkafkafk13: No, I deliberately do not want to compute a simplified covering. Its cost outweighs the benefit, but if it is there anyway, I will use it for the speedup.

experimental/Schemes/duValSing.jl
afkafkafk13: According to the classification of Greuel and Kröning, the classification differs for p \in \{2,3,5\}. Therefore I put in >5 and this should do the job.
Sideremark: duVal-Singularities are only called duVal-Singularities in characteristic zero -- but I understand your need to pass to large positive characteristic to avoid intermediate coefficient swell.

experimental/Schemes/duValSing.jl
afkafkafk13: I definitely plan to do so, but that will be another PR, which sets suitable attributes to cache such properties and also computed decompositions. This looks like a larger cleanup in this area (ranging from CommAlg to Schemes and passing on information as apropriate).

src/Modules/UngradedModules.jl
afkafkafk13: done

experimental/Schemes/SpaceGerms.jl
```diff
@@ -124,37 +124,10 @@ function rational_point_coordinates(I::MPolyIdeal)
...
-        push!(result,JN[i])
-        push!(Jtemp, JN[i])
-      end
-    end
-    J = ideal(R,Jtemp)
-    J = ideal(R,normal_form(gens(J*M),I))
-  end
-  return length(result),result
+  nf_vec = [normal_form(v,I) for v in gens(R)]
+  return [ a == zero(parent(a)) ? zero(coefficient_ring(a)) : leading_coefficient(a) for a in nf_vec] # TODO does the ordering matter?
```
fingolfin:
```suggestion
  return [ iszero(a) ? zero(coefficient_ring(a)) : leading_coefficient(a) for a in nf_vec] # TODO does the ordering matter?
```
Please don't construct objects (here: a zero) just to check for equality to it. Likewise, don't do `list == []`, do `isempty(list)` etc.

afkafkafk13: Thank you for the hint. I am still learning how to do things Oscar-style.

experimental/Schemes/duValSing.jl
simonbrandhorst: I understand.

experimental/Schemes/duValSing.jl
simonbrandhorst: o.k. we can keep this for another pr.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+  describing some geometrically irreducible point
+- T[3]::Tuple type of the singularity at ``P``  e.g. `(:A, 3)`
+- T[4]::Int number of conjugate points
+
+If X has a least one singularity which is not du Val, the returned vector contains a
+single tuple T, with the following values:
+- T[1] = false
+- T[2] = point at which some non-du-Val singularity is present
+- T[3] = empty tuple
+- T[4] = 1
```
simonbrandhorst:
```suggestion
- `T[1]::Bool` answers ``X`` has at most du Val (surface) singularities at the geometric points
specified by the ideal ``I``.
- `T[2]::Ideal` is ``I_P`` the associated prime of `I` (possibly over a suitable field extension)
  describing some geometrically irreducible point
- `T[3]::Tuple` the type of the singularity at ``P``  e.g. `(:A, 3)`
- `T[4]::Int` number of conjugate points

If X has a least one singularity which is not du Val, the returned vector contains a
single tuple T, with the following values:
- `T[1]`  is `false`
- `T[2]` represents a point at which some non-du-Val singularity is present
- `T[3]` is the empty tuple
- `T[4] = 1`
```

src/Modules/UngradedModules.jl
```diff
@@ -8712,3 +8716,258 @@ function dual(f::ModuleFPHom{<:ModuleFP, <:ModuleFP, Nothing}; # Third parameter
...
+  R = base_ring(M)
+  F = ambient_free_module(M)
+  Mq,_ = sub(F,rels(M))
+
+  ambient_representatives_generators(M) == gens(F) || error("not implemented for M/N with non-trivial M")
+
+  o = default_ordering(M)
+  LM = leading_module(Mq,o)
+
+  return length([x*e for x in all_monomials(R, d) for e in gens(F) if !(x*e in LM)])
```
simonbrandhorst: The list comprehension is not necessary
```suggestion
  return count(t->!(t[1]*t[2] in LM), Iterators.product(all_monomials(R, d), gens(F)))
```

src/Modules/UngradedModules.jl
```diff
@@ -8712,3 +8716,258 @@ function dual(f::ModuleFPHom{<:ModuleFP, <:ModuleFP, Nothing}; # Third parameter
...
+
+function vector_space_dimension(M::SubquoModule{T},d::Int64
+  ) where {T<:MPolyLocRingElem}
+  error("only available in global case and for localization at a point")
+end
+
+@doc raw"""
+    vector_space_basis(M::SubquoModule, d::Int)
+
+Let ``R`` be a `MPolyAnyRing` over a field ``kk`` and let ``M`` be a subquotient module over ``R``.
```
```suggestion
Let ``R`` be a `MPolyAnyRing` over a field ``k`` and let ``M`` be a subquotient module over ``R``.
```
Should be ``k`` or $\mathbb{k}$ as we would write in an article. I would say that `kk` is for code only.

simonbrandhorst (commented):
Can't say much on the mathematics. The overall design looks good. 
Just needs a few trivial syntax changes to the docu and this can be merged.

afkafkafk13:
Suggested changes incorporated.

simonbrandhorst:
There are other places where your docstrings need some more care. Please go through your code again and fix them.

experimental/Schemes/duValSing.jl
```diff
@@ -0,0 +1,262 @@
...
+  smooth = (:A,0)
+
+  ## localize at point
+  a = rational_point_coordinates(Ipt)
+  U = complement_of_point_ideal(R,a)
+  RL, loc_map = localization(R,U)
+  IX_loc = loc_map(IX)
+  JM_loc =  map(x ->loc_map(x), JM[:,:])
+
+  if !all(iszero(a))
```
fingolfin, line 225: I would suggest either
```suggestion
  if !all(iszero,a)
```
or
```suggestion
  if !iszero(a)
```

src/Modules/UngradedModules.jl
```diff
@@ -1539,7 +1543,7 @@ function is_canonically_isomorphic(M::SubModuleOfFreeModule, N::SubModuleOfFreeM
   return SubModuleOfFreeModule(G, [f(v) for v in gens(M)]) == N
 end
 
-#+(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule) = sum(M, N)
+Base.:+(M::SubModuleOfFreeModule, N::SubModuleOfFreeModule) = sum(M, N)
```
fingolfin, line 1546: I wonder why this was commented out (@jankoboehm do you know)? According to Codecov, nothing in our test suite seems to call it right now, making me wary if there is something is subtly wrong here?

simonbrandhorst (approved):

src/Modules/UngradedModules.jl
```diff
@@ -8712,3 +8716,258 @@ function dual(f::ModuleFPHom{<:ModuleFP, <:ModuleFP, Nothing}; # Third parameter
...
+  ambient_rank = ngens(ambient_free_module(M))
+  genlist = ambient_representatives_generators(M)
+  explist = Tuple{Vector{Int64}, Int64, Int}[]
+  for x in genlist
+    tempexp = leading_exponent(x)
+    tempdeg = sum(tempexp[1])
+    push!(explist,(tempexp[1],tempexp[2],tempdeg))
+  end
+  all(findfirst(x -> (x[1][i] == x[3] && x[2]==j),explist) != nothing for i in 1:ngens(R) for j in 1:ambient_rank) || return false
+  return true
```
fingolfin: Could be rewritten as
```suggestion
  return all(findfirst(x -> (x[1][i] == x[3] && x[2]==j),explist) != nothing for i in 1:ngens(R) for j in 1:ambient_rank)
```
That said, I personally find it difficult to read these long constructs, and personally think that the explicit imperative version is easier to understand and debug -- but I'll not complain either way, use whatever you prefer.
```suggestion
  for i in 1:ngens(R), j in 1:ambient_rank
    if findfirst(x -> (x[1][i] == x[3] && x[2]==j), explist) === nothing
      return false
    end
  end
  return true
```
or for that matter
```suggestion
  for i in 1:ngens(R), j in 1:ambient_rank
    if !any(x -> (x[1][i] == x[3] && x[2]==j), explist)
      return false
    end
  end
  return true
```

fingolfin (commented):
Some purely stylistic comments, feel free to ignore :-)

simonbrandhorst:
I think with my approval you can merge yourself, when you are ready?

experimental/Schemes/duValSing.jl
afkafkafk13: Thanks for the hint.

PR #2470, opened 2023-06-14
[FTheoryTools] Overhaul base-independent models

HereAround:
I split the part of the hypersurface model PR (https://github.com/oscar-system/Oscar.jl/pull/2382) off that will likely not run yet (and which I will likely not have the time to fix in the upcoming days). In this way, the original PR can now  be merged shortly. This then allows @apturner to continue working on the F-Theory Tools.

### DIFF
--- a/experimental/FTheoryTools/docs/src/hypersurface.md
+++ b/experimental/FTheoryTools/docs/src/hypersurface.md
@@ -133,3 +133,34 @@ auxiliary ambient space or auxiliary hypersurface have been computed.
 To this end, one invokes `set_verbosity_level(:HypersurfaceModel, 1)`.
 More background information is available
 [here](http://www.thofma.com/Hecke.jl/dev/features/macros/).
+
+### Attributes in toric settings
+
+If the base space of the hypersurface model is a toric space, then we
+also provide a special type for the Calabi-Yau hypersurface:
+```@docs
+calabi_yau_hypersurface(h::HypersurfaceModel)
+```
+
+### Attributes based on the corresponding global Tate and Weierstrass models
+
+Currently, we do not provide functionality to convert a hypersurface model
+into a Weierstrass or global Tate model. Still, for some constructions this might
+be known or detailed in the literature. If the user wishes, one can then associate
+a corresponding Weierstrass or global Tate model as follows:
+```@docs
+set_weierstrass_model(h::HypersurfaceModel, w::WeierstrassModel)
+set_global_tate_model(h::HypersurfaceModel, w::GlobalTateModel)
+```
+These models can then be accessed with the following functions:
+```@docs
+weierstrass_model(h::HypersurfaceModel)
+global_tate_model(h::HypersurfaceModel)
+```
+Provided that the corresponding Weierstrass model is known for a hypersurface
+model, the following functionality is available. It returns the attribute in question
+of the corresponding Weierstrass model.
+```@docs
+discriminant(h::HypersurfaceModel)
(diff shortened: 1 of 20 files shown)


HereAround:
In https://github.com/oscar-system/Oscar.jl/pull/2496, I am proposing an implementation of the most general star subdivision (following the toric geometry book by Cox Little Schenk, chapter 11). This was needed to get this PR running. I am now confident, that the tests for this PR shall pass or will do so shortly.

For my convenience (to make the necessary rebase as painless as possible), this PR contains the changes in https://github.com/oscar-system/Oscar.jl/pull/2496. Consequently, this other PR should be discussed first... 

[I understand that the current failures in the nightly tests are not related to the changes in this PR.]

HereAround:
As long as features are concerned, I have just completed the overhaul of the base independent models. This includes:
* Other auxiliary base spaces (by means of the most general star subdivision for toric spaces).
* Have Kbar as homogeneous coordinate of the auxiliary base space.
* Consistent order of the input variables of models over auxiliary base spaces. Information about the base space comes in first (base variables/ring, grading, dimension), fiber information second (special sections, fiber ambient space...).

Note that this also allowed me to switch on a test, that I initially had to switch off for this PR to work. In this particular test, initially no star triangulation could be found. By adding Kbar as base variable and using the most general star subdivisions for fans (as defined in CLS11), this is now working.

I suggest to wait for the tests to go green. Then this should be ready for review/merge.

@apturner What do you think?

HereAround:
I noticed that the `doctests` time out. The reason is the following:
```julia
julia> m = literature_model(arxiv_id = "1507.05954", equation = "A.1")
Weierstrass model over a not fully specified base -- U(1)xU(1) Weierstrass model based on arXiv paper 1507.05954 Eq. (A.1)
```
This model has an auxiliary base grading:
```julia
"auxiliary_base_grading": [
            [6, 4, 2, 3, 1, 0, 0, 0, 0, -1, -1, -1],
            [-3, -2, -1, -1, 0, 1, 0, 0, 0, 1, 1, 1],
            [-2, -1, 0, -1, 0, 0, 1, 0, 0, 1, 0, 0],
            [-2, -1, 0, -1, 0, 0, 0, 1, 0, 0, 1, 0],
            [-2, -1, 0, -1, 0, 0, 0, 0, 1, 0, 0, 1]
        ],
```
This means, the attempt to construct an auxiliary base space result in the triangulation of a 5 dimensional polytope with 12 vertices. That is excessive and, based on the above observation, leads to time outs of the doctests after 150 minutes. So this is no good for a doctest. Even more, this is likely not something that a user could hope to use...

I would like this PR to complete and be merged at some point (I have been working on these points since Andrew and I met in Kaiserslautern in May...). Hence, I have taken the liberty to just removed the calls and the tests based on this model.

Let us discuss how we can make this model usable again in another PR. One possibility would be to issue the triangulation on a cluster and save the result in the json file, so that one merely has to construct the toric space. Or possibly one can directly save the toric space. The question of course remains if such a triangulation will be at all successful when performed (on a cluster). That I cannot tell...

HereAround:
Generally speaking, it might be worth to save a triangulated base space in the json files. We should expect that this is a very time consuming operation. So users of the data base should not be punished by this...

apturner (approved):
This looks great, thank you for all the effort @HereAround!

HereAround:
Thank you @apturner I will open issues to keep track of the remaining points that could not (yet) be addressed.

Issue #2427, opened 2023-05-28
is_prime buggy for multivariate polynomial rings


simonbrandhorst:
By definition the (1) ideal is never prime. 
Yet:
```
julia> R,x = polynomial_ring(QQ,2);

julia> I = ideal(R,[1])
ideal(1)

julia> is_prime(I)
true
```
The reason is that 
```
julia> minimal_primes(I)
1-element Vector{MPolyIdeal{QQMPolyRingElem}}:
 ideal(1)
```
Refuses to return an empty list but instead returns a non-prime ideal. 
(Matching its documentation, so it is not a bug.)
In any case I would suggest to rather return only prime ideals,
even if it means that you cannot do 
`intersect(minimal_primes(I))`  (if the return value is the empty array). One can still do something like
`intersect(minimal_primes(I), init=R)` instead. 
We have the same situation when we sum the elements of an array.
@wdecker

fingolfin:
We discussed this today and agreed that `minimal_primes` should return an empty vector in this case. Moreover, there are other functions that may suffer from similar problems, so we should check those as well. E.g. this is also questionable:
```julia-repl
julia> primary_decomposition(I)
1-element Vector{Tuple{MPolyIdeal{QQMPolyRingElem}, MPolyIdeal{QQMPolyRingElem}}}:
 (ideal(1), ideal(1))
```

There are further considerations (e.g. maybe `primary_decomposition` should return a special type; or should always include a "unit factor" etc.) but that goes beyond the point here.

fingolfin:
@wdecker agreed to work on this. Meaning: just changing `minimal_primes` as discussed, verifying `is_prime` now works, perhaps adding a test. And depending on his time also look for other problematic commands that need similar fixes.

wdecker:
Solved with PR https://github.com/oscar-system/Oscar.jl/pull/2473

simonbrandhorst:
Thank you!

PR #2473, opened 2023-06-15
addresseing Issue #2427

wdecker:


fingolfin (approved):
This PR looks good to me, with the caveat that I have no idea what the change in the schemes code from `isprime(D)` to `isone(D) || isprime(D)` does and if it is the right thing to do.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -430,7 +430,7 @@ end
 #
```
HechtiDerLachs, line 434:
```suggestion
function isone(I::IdealSheaf)
  return all(x->isone(I(x)), affine_charts(scheme(X)))
end
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -430,7 +430,7 @@ end
...
 
 function is_prime(I::IdealSheaf) 
-  return all(U->is_prime(I(U)), basic_patches(default_covering(space(I))))
+  return all(U->(is_one(I(U)) || is_prime(I(U))), basic_patches(default_covering(space(I))))
```
HechtiDerLachs, line 437:
```suggestion
  all(x->isone(I(x)), affine_charts(scheme(I))) && return false
  return all(U->(is_one(I(U)) || is_prime(I(U))), basic_patches(default_covering(space(I))))
```

experimental/Schemes/WeilDivisor.jl
```diff
@@ -35,7 +35,7 @@ stored as a formal linear combination over some ring ``R`` of
     ) where {CoefficientRingType, CoefficientRingElemType}
     @check begin
       for D in keys(coefficients)
-        isprime(D) || error("components of a divisor must be sheaves of prime ideals")
+        isone(D) || isprime(D) || error("components of a divisor must be sheaves of prime ideals")
```
HechtiDerLachs:
```suggestion
        isprime(D) || error("components of a divisor must be sheaves of prime ideals")
```

experimental/Schemes/WeilDivisor.jl
```diff
@@ -46,7 +46,7 @@ stored as a formal linear combination over some ring ``R`` of
     X = scheme(C)
     @check begin
       for D in keys(coefficient_dict(C))
-        isprime(D) || error("components of a divisor must be sheaves of prime ideals")
+        isone(D) || isprime(D) || error("components of a divisor must be sheaves of prime ideals")
```
HechtiDerLachs:
```suggestion
        isprime(D) || error("components of a divisor must be sheaves of prime ideals")
```

HechtiDerLachs (commented):
Let's see whether this works.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -428,9 +428,12 @@ end
 #  return W, spec_dict
 #end
 #
-
+function isone(I::IdealSheaf)
+  return all(x->isone(I(x)), affine_charts(scheme(X)))
```
afkafkafk13: Did you want to write
  return all(x->isone(I(x)), affine_charts(scheme(I)))
(There is no X in this context and scheme(I) would make sense.)

experimental/Schemes/IdealSheaves.jl
```diff
@@ -428,9 +428,12 @@ end
...
+end
 function is_prime(I::IdealSheaf) 
-  return all(U->is_prime(I(U)), basic_patches(default_covering(space(I))))
+  all(x->isone(I(x)), affine_charts(scheme(I))) && return false
```
afkafkafk13: It might be more readable, if you write
  !isone(I) || return false

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: Yes. This should have been it.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: That would also be fine, yes. should also increase test coverage.

experimental/Schemes/IdealSheaves.jl
afkafkafk13: done

experimental/Schemes/IdealSheaves.jl
afkafkafk13: done

simonbrandhorst (approved):

simonbrandhorst:
Looks good to me.  Somewhat aside to this PR:

I think that `is_prime(::IdealSheaf)`  does not do what it is supposed to.  We only check that it is locally prime at some open cover. But I would say that is not a local property. Because we can take a disconnected scheme of two points {P1,P2} and take the open cover U1 = {P1} and U2={P2}. Now consider the zero sheaf of ideals. 
It will be locally prime but not globally.  
So what our function checks is just that a sheaf of prime ideals is locally prime. But we do not check that it is prime for every affine open (e.g. the affine open containing both {P1,P2}). And even if we could, I am not sure if we still get something irreducible in the non-separated case.  

In any case this behavior is unrelated to the PR and we should discuss it in an issue.

afkafkafk13:
> Looks good to me. Somewhat aside to this PR:
> 
> I think that `is_prime(::IdealSheaf)` does not do what it is supposed to. We only check that it is locally prime at some open cover. But I would say that is not a local property. Because we can take a disconnected scheme of two points {P1,P2} and take the open cover U1 = {P1} and U2={P2}. Now consider the zero sheaf of ideals. It will be locally prime but not globally. So what our function checks is just that a sheaf of prime ideals is locally prime. But we do not check that it is prime for every affine open (e.g. the affine open containing both {P1,P2}). And even if we could, I am not sure if we still get something irreducible in the non-separated case.
> 
> In any case this behavior is unrelated to the PR and we should discuss it in an issue.

I agree and would like to add that is_connected is not yet implemented for schemes, whence the above issue is not one to be fixed within days.

PR #2474, opened 2023-06-15
Finite dimensional quotients as vector spaces.

HechtiDerLachs:
@afkafkafk13 `vector_space` for modules over polynomial rings.

### DIFF
--- a/src/Modules/missing_functionality.jl
+++ b/src/Modules/missing_functionality.jl
@@ -81,10 +81,9 @@ subquo_type(R::RingType) where {RingType<:Ring} = subquo_type(typeof(R))
 function _kbase(I::SubModuleOfFreeModule{<:MPolyRingElem{<:FieldElem}};
     ordering::ModuleOrdering=default_ordering(ambient_free_module(I))
   )
-  @show gens(I)
-  @show groebner_basis(I)
   lead_I = leading_module(I, ordering)
-  @show gens(lead_I)
+  @show ordering
+  @show lead_I
   F = ambient_free_module(I)
   R = base_ring(F)
   # Iterate through all components until we have everything
@@ -103,7 +102,7 @@ function _kbase(I::SubModuleOfFreeModule{<:MPolyRingElem{<:FieldElem}};
 end
 
 function vector_space(
-    K::AbstractAlgebra.Field, M::SubQuo{<:MPolyRingElem{<:FieldElem}};
+    K::AbstractAlgebra.Field, M::SubquoModule{<:MPolyRingElem{<:FieldElem}};
     ordering::ModuleOrdering=default_ordering(ambient_free_module(M))
   )
   R = base_ring(M)
@@ -114,9 +113,8 @@ function vector_space(
   F1 = C[1]
   # TODO: Is there a way to access this directly? If yes, do so.
   I = SubModuleOfFreeModule(F0, map(C, 1).(gens(F1)))
-  gb_I = groebner_basis(I, ordering)
+  gb_I = standard_basis(I, ordering=ordering)
   mons = _kbase(I, ordering=ordering)
-  @show mons
   l = length(mons)
   V = free_module(K, l)
   function im(a::Generic.FreeModuleElem)
@@ -137,7 +135,6 @@ function vector_space(
   function prim(a::SubquoModuleElem)
     @assert parent(a) === M
     b = preimage(map(C, 0), a)
-    @show b
     b = normal_form(b, gb_I) #TODO: Is this normal_form reduced?
     result = zero(V)
     while !iszero(b)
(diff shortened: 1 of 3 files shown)


src/Modules/missing_functionality.jl
```diff
@@ -77,3 +77,114 @@ end=#
...
+  R = base_ring(F)
+  # Iterate through all components until we have everything
+  result = elem_type(F)[]
+  for i in 1:ngens(F)
+    m = ideal(R, gens(R))
+    pom = ideal(R, one(R))
+    while true
+      new_mons = [x for x in gens(pom) if !(x*F[i] in lead_I)]
+      iszero(length(new_mons)) && break
+      result = vcat(result, [x*F[i] for x in new_mons])
```
fingolfin, line 97: This is an extremely minor thing, but I just wanted to point out that depending on the ring, `F[i]` might allocate a new object each time it is called, so perhaps pull it out before the loop? Also, `x*F[i]` is computed twice for all remaining in `new_mons`. 
```suggestion
    Fi = F[i]
    m = ideal(R, gens(R))
    pom = ideal(R, one(R))
    while true
      new_mons = [x*Fi for x in gens(pom)]
      filter!(x -> !(x in lead_I), new_mons)
      ismpty(new_mons) && break
      result = vcat(result, new_mons)
```

src/Modules/missing_functionality.jl
```diff
@@ -77,3 +77,114 @@ end=#
     
 subquo_type(::Type{RingType}) where {RingType<:Ring} = SubquoModule{elem_type(RingType)}
 subquo_type(R::RingType) where {RingType<:Ring} = subquo_type(typeof(R))
+
+function _kbase(I::SubModuleOfFreeModule{<:MPolyRingElem{<:FieldElem}};
+    ordering::ModuleOrdering=default_ordering(ambient_free_module(I))
+  )
+  lead_I = leading_module(I, ordering)
+  @show ordering
+  @show lead_I
```
fingolfin, line 86: I guess this means this code is not yet ready? Then perhaps mark it as draft?

src/Modules/missing_functionality.jl
fingolfin, line 97: This is an extremely minor thing, but I just wanted to point out that depending on the ring, `F[i]` might allocate a new object each time it is called, so perhaps pull it out before the loop? Also, `x*F[i]` is computed twice for all remaining in `new_mons`. 
```suggestion
    Fi = F[i]
    m = ideal(R, gens(R))
    pom = ideal(R, one(R))
    while true
      new_mons = [x*Fi for x in gens(pom)]
      filter!(x -> !(x in lead_I), new_mons)
      ismpty(new_mons) && break
      result = vcat(result, new_mons)
```

fingolfin:
Does the geometry team have a plan who will review this PR?

afkafkafk13:
> Does the geometry team have a plan who will review this PR?

I am planning to do this and I was already in contact with @HechtiDerLachs about certain details.

fingolfin:
Excellent. I've assigned the PR to you then to reflect this in the webinterface :-)

afkafkafk13:
@HechtiDerLachs :  Could you check whether some of the problems you needed to avoid are fixed by now? There is a vector_space_basis for quo's by now in the ideal and module setting.

test/Modules/module-localizations.jl
```diff
@@ -131,3 +131,20 @@ end
...
 end
+
+@testset "as vector spaces" begin
+
+R, (x, y) = QQ["x", "y"]
+R, _ = localization(R, complement_of_point_ideal(R, [0, 1]))
+F2 = free_module(R, 2)
+I, inc_I = sub(F2, [x^3*F2[1], (y-1)*y*F2[1], x^2*F2[1]-(y-1)*F2[2], x*F2[2], (y-1)^2*F2[2]])
+M, pr_M = quo(F2, I)
+# tests below should run one day.
```
afkafkafk13, line 142: can you outline, where the problems are?

HechtiDerLachs: I don't remember in full detail. I think the map `V -> M` works. But `M -> V` needs to bring elements in `M` to normal form and then collect coefficients. Even in the case of trivial denominators, there was something missing from the module side (was it messing with the local orderings...?). Another step would then be to also allow non-trivial denominators.

afkafkafk13: ah, yes.
Reduced normal forms for localized rings do not make sense in the general setting (will not terminate), but in the present situation (with a weighted local ordering and in the presence of a highest corner) they exist and can be computed. This is, what you will need here.
But given the fact, that even for the reduced normal form of an element (not ideal!) in the global case a hack was necessary, I see things that are more urgent. 
(The denominators are then nothing to worry about, as soon as the rest works)

simonbrandhorst:
What is the status of this PR?

fingolfin:
Note that we have a function `monomial_basis`, and also a function `vector_space_basis`. And now this here. It would be good to make sure we are not reinventing the wheel everywhere, and that these things fit together...

wdecker:
@fingolfin As I can see from the docstrings in UngradedModules.jl, the `vector_space_basis` function also
returns a monomial basis in each case so that this function could be renamed to `monomial_basis`.
Similarly for `_kbase`. We should discuss, however, whether we should add an additional argument for the monomial ordering to be used whenever this is appropriate. @jankoboehm @HechtiDerLachs @afkafkafk13 ???

src/Modules/missing_functionality.jl
```diff
@@ -77,3 +77,114 @@ end=#
...
+      new_mons = [x for x in gens(pom) if !(x*F[i] in lead_I)]
+      iszero(length(new_mons)) && break
+      result = vcat(result, [x*F[i] for x in new_mons])
+      pom = pom * m
+    end
+  end
+  return result
+end
+
+function vector_space(
```
fingolfin, line 104: A comment (or even better, docstring) that explains what this function does (what does it take, what does it return) would be helpful

fingolfin:
So what exactly needs to be done here? Maybe next week some people can discuss this in person in Saarbrücken?

fingolfin:
@afkafkafk13 mentions that one blocked is the lack of "strong normal forms in the local case" which apparently are needed to write down the isomorphism. But apparently it is not that hard to implement, so "someone" just has to do it... But again, maybe it can be discussed next week (and details added here)

simonbrandhorst:
@HechtiDerLachs is this still relevant? Can this be closed?

HechtiDerLachs:
I forgot what all this was really about and why it didn't proceed. But somewhere local orderings and modules pop up and I don't have the appetite to get into any of these things right now. The name of the PR suggests that we wanted this functionality for our School in Lille last year. Since it has now already taken place and we seem to have managed without this stuff, I would be OK with throwing it away. It can still be digged out if anyone ever wanted to come back to this.

PR #2481, opened 2023-06-20
Fix `standard_basis` documentation

RafaelDavidMohr:
This fixes the function signature in the documentation string for `standard_basis`.


wdecker (approved):
Thanks!

PR #2490, opened 2023-06-23
Import and extend conjugate from upstream

thofma:


lgoettgens:
I think it would be nicer to export `conjugate` in Hecke instead and then make use of the everything-import-export. This would then make the `conjugate` exports inside and outside the JuLie submodule obsolete.

thofma:
Yes, but I want it now. We will export it from Hecke in the next release.

fieker (approved):

Issue #2482, opened 2023-06-21
Improve documentation for `character_table`


fingolfin:
Consider this:
```
julia> g = cyclic_group(6)
<pc group of size 6 with 2 generators>

julia> character_table(g)
<pc group of size 6 with 2 generators>

  2  1        1        1  1        1        1
  3  1        1        1  1        1        1

    1a       6a       3a 2a       3b       6b

X_1  1        1        1  1        1        1
X_2  1       -1        1 -1        1       -1
X_3  1 -z_3 - 1      z_3  1 -z_3 - 1      z_3
X_4  1  z_3 + 1      z_3 -1 -z_3 - 1     -z_3
X_5  1      z_3 -z_3 - 1  1      z_3 -z_3 - 1
X_6  1     -z_3 -z_3 - 1 -1      z_3  z_3 + 1
```
There is no hint what `z_3` is, nor what `X_1`, `X_2`, refers to, now are the header rows explained.

Contrast this with GAP's output:
```
gap> Display(CharacterTable(CyclicGroup(6)));
CT2

     2  1   1  1  1  1   1
     3  1   1  1  1  1   1

       1a  6a 3a 2a 3b  6b

X.1     1   1  1  1  1   1
X.2     1  -1  1 -1  1  -1
X.3     1   A /A  1  A  /A
X.4     1  -A /A -1  A -/A
X.5     1  /A  A  1 /A   A
X.6     1 -/A  A -1 /A  -A

A = E(3)^2
  = (-1-Sqrt(-3))/2 = -1-b3
```
which at least tell us what `A` (taking the role of `z_3`) is.

However it would be nice if a little bit was said about the various headers. Either in the `character_table` docstring directly; or somewhere else in the manual (and then the `character_table` docstring should say something to the effect of "for more information about the printed data, look at REF")

PR #2491, opened 2023-06-23
Rename default non-unicode variable for QQAbElem from 'z' to 'zeta'

### LINKED ISSUE #2484: Problems with the `binomial_primary_decomposition` docstring resp. with QQAb printing
Two problems with the docstring for `binomial_primary_decomposition`:

1. It references ζ (zeta), but the code example was changed to use `z`
2. In the example `z` actually turns up with two different meanings: once it is a variable in a polynomial ring; and once for a root of unity `z(3)`. I believe that `z` comes from the `QQAb` code.

Since `z` is used a lot in polynomial rings, it seems best if in the printing of roots of unity it is changed to something else, e.g. `zeta` (spelled out) -- in unicode printing mode it can of course stay as ζ

Secondly, the `binomial_primary_decomposition` docstring should be adjusted to use that new symbol used for `QQAb` printing (it could also point to `Oscar.with_unicode` I guess)

fingolfin:
Resolves #2484

### DIFF
--- a/src/Rings/AbelianClosure.jl
+++ b/src/Rings/AbelianClosure.jl
@@ -89,12 +89,12 @@ or slower.
 julia> K, z = abelian_closure(QQ);
 
 julia> z(36)
-z(36)
+zeta(36)
 
 julia> K, z = abelian_closure(QQ, sparse = true);
 
 julia> z(36)
--z(36, 9)*z(36, 4)^4 - z(36, 9)*z(36, 4)
+-zeta(36, 9)*zeta(36, 4)^4 - zeta(36, 9)*zeta(36, 4)
 
 ```
 """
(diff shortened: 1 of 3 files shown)


thofma:
I think this is also relevant for character tables. I forgot what @ThomasBreuer preferred. Was it `z` or `zeta`?

ThomasBreuer:
@thofma Thanks for the hint.
The `show` method for character tables use the `show` method for elements in cyclotomic fields (which creates terms of the form `z_N`), not the one for `QQAbElem`. This is good because otherwise the column widths would be unnecessarily large.
The `show` method for characters currently delegates to the `show` method for `Vector{QQAbElem}`. With the same argument, one can argue that `z_N` should be preferred to `zeta_N`; for that, I can change the `show` method for characters.
(I am anyhow working on docstrings that involve irrational values in characters in character tables because of #2482.)

PR #2493, opened 2023-06-26
Implement degree(I::MPolyIdeal)

taboege:
Supply a `degree` method for ideals. It homogenizes the ideal if it has to and then computes the degree of its quotient algebra.

This feature is largely for convenience. Especially in `Macaulay2`, it is very easy to compute the pair `(dimension, degree)` for any (inhomogeneous) ideal to get very rough data about the geometry of an affine variety. With this commit, that becomes equally easy in Oscar.

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -1255,12 +1255,16 @@ codim(I::MPolyIdeal) = nvars(base_ring(I)) - dim(I)
 @doc raw"""
     degree(I::MPolyIdeal)
 
-Return the degree of `I` if it is homogeneous. Otherwise return the degree
-of its `homogenization`.
+If `base_ring(I)` is standard-graded, in which case `I` is necessarily
+homogeneous with respect to this grading, return the degree of `I` (that
+is, the degree of the quotient of `base_ring(I)` modulo `I`). Otherwise,
+return the degree of the homogenization of `I` with respect to the
+standard-grading.
 
-The degree of a homogeneous ideal `I` is the number of intersection points
-of its projective variety with a generic linear subspace of complementary
-dimension (counted with multiplicities). See also [MS21](@cite).
+!!! note
+Geometrically, the degree of a homogeneous ideal as above is the number of
+intersection points of its projective variety with a generic linear subspace
+of complementary dimension (counted with multiplicities). See also [MS21](@cite).
 
 # Examples
 ```jldoctest
@@ -1275,9 +1279,8 @@ julia> degree(I)
 ```
 """
 @attr Int function degree(I::MPolyIdeal)
-  J = homogenization(I, "_h")
-  A, _ = quo(base_ring(J), J)
-  return Int(degree(A))
+  is_standard_graded(base_ring(I)) || (I = homogenization(I, "_h"))
+  return Int(degree(HilbertData(I)))
 end
 
 ################################################################################
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -1250,6 +1250,32 @@ julia> codim(I)
 """
 codim(I::MPolyIdeal) = nvars(base_ring(I)) - dim(I)
 
+#######################################################
+#######################################################
+@doc raw"""
+    degree(I::MPolyIdeal)
+
+Return the degree of `I` if it is homogeneous. Otherwise return the degree
+of its `homogenization`.
+
+# Examples
+```jldoctest
+julia> R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
+(Multivariate polynomial ring in 3 variables over QQ, QQMPolyRingElem[x, y, z])
+
+julia> I = ideal(R, [y-x^2, x-z^3])
+ideal(-x^2 + y, x - z^3)
+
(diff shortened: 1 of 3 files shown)


src/Rings/mpoly-ideals.jl
```diff
@@ -1250,6 +1250,32 @@ julia> codim(I)
 """
 codim(I::MPolyIdeal) = nvars(base_ring(I)) - dim(I)
 
+#######################################################
+#######################################################
+@doc raw"""
+    degree(I::MPolyIdeal)
+
+Return the degree of `I` if it is homogeneous. Otherwise return the degree
```
fingolfin: As a non-algebraic geometer: what is the definition of the degree of a (homogeneous) ideal? Do you have a text book reference?

wdecker:
If A is a projective variety with coordinate ring R/I, the topological dimension of A in its Zariski topology agrees with the Krull dimension of the coordinate ring R/I on the algebraic side. The Krull dimension of the ideal I itself is by definition the Krull dimension of R/I.

The degree of A is defined to be the number of intersection points with a generic hyperplane of complementary dimension. On the algebraic side, this number can be recovered as the degree of R/I which is defined to be (dim R/I)! times the leading coefficient of the Hilbert polynomial of A. As with the notion of dimension above you might now also speak of the degree of I rather than the degree of R/I. This is, however, not commonly used. In a CAS, I guess, the reason for implementing such a function is just to save some typing. 

A further comment will follow.

src/Rings/mpoly-ideals.jl
taboege: One reference I know is  [Michałek and Sturmfels: *Invitation to non-linear algebra*, Section 1.3](https://www.math.uni-konstanz.de/~michalek/NonLinearAlgebra.pdf#page=31). In the homogeneous case, the definition (and the reason for adding this method) are just like [Wolfram says](https://github.com/oscar-system/Oscar.jl/pull/2493#issuecomment-1609401317).

If the ideal is inhomogeneous, Michałek and Sturmfels define the degree to be that of the (homogeneous!) initial ideal with respect to any *degree-compatible* monomial order. Corollary 2.19 in their book then implies that their definition coincides with the degree of the homogenization of the ideal.

fingolfin: OK, it's good there is a reference. Still, this is not something that seems to be very general knowledge, so I feel this docstring should at least give a rough description of what "the degree" means in this context. Could you please add something?

fingolfin: To be clear: I won't be the one to approve/reject that, I'll leave this to our geometers like @wdecker @afkafkafk13 @simonbrandhorst -- so perhaps we should first hear whether they are willing to merge this, before you waster your valuable time for nothing.

In any case: thank you for your contribution!

wdecker: @taboege Thank you for your contribution. @fingolfin As I already said, a further remark from my side will follow. The code can be made more efficient. I will make a corresponding suggestion, but first have to see a larger picture.

fingolfin: @taboege can you add the explanation to the docstring?

taboege: I added a small paragraph explaining what the degree is, including a reference to the textbook above.

src/Rings/mpoly-ideals.jl
```diff
@@ -1250,6 +1250,36 @@ julia> codim(I)
...
+ideal(-x^2 + y, x - z^3)
+
+julia> degree(I)
+6
+```
+"""
+@attr Int function degree(I::MPolyIdeal)
+  J = homogenization(I, "_h")
+  A, _ = quo(base_ring(J), J)
+  return Int(degree(A))
```
wdecker: Suggested change:  

  is_standard_graded(base_ring(I)) || (I = homogenization(I, "_h")) 
  return Int(degree(HilbertData(I)))

In this way, in the homogeneous case, only one GB calculation is needed.
I

wdecker:
This is a comment on the provided help text:

####################

  @doc raw"""
      degree(I::MPolyIdeal)

  Return the degree of `I` if it is homogeneous. Otherwise return the degree of its `homogenization`.

  The degree of a homogeneous ideal `I` is the number of intersection points of its projective variety with a generic linear subspace of complementary dimension (counted with multiplicities). See also [MS21](@cite).

####################

In the cited book, only standard $\mathbb Z$-gradings are considered. In OSCAR, however, also other gradings are allowed. To avoid confusion, this should be reflected by the help text. Here is one possibility:

####################

If `base_ring(I)` is standard $\mathbb Z$-graded, in which case `I` is neccessarily  homogeneous with respect to this grading, return the degree of `I` (that is, the degree of the quotient of `base_ring(I)` modulo `I`). Otherwise, return the degree of the homogenization of `I` with respect to the standard $\mathbb Z$-grading.

!!! note
    Geometrically, the degree of a homogeneous ideal as above is the number of intersection points of its projective variety with a generic linear subspace of complementary dimension (counted with multiplicities). See also [MS21](@cite).

####################

taboege:
Sorry for the delay, I was away at a conference. Suggestions are implemented in the last commit.

wdecker (approved):

PR #2495, opened 2023-06-27
Transform cones for toric varieties

lkastner:
Some maps are not the same in polymake vs Oscar.


HereAround (approved):
Thank you @lkastner. Looks perfect to me.

PR #2497, opened 2023-06-27
Add some more optimizations.

HechtiDerLachs:
Another patch from work in progress by @simonbrandhorst and myself.

### DIFF
--- a/experimental/Schemes/Auxiliary.jl
+++ b/experimental/Schemes/Auxiliary.jl
@@ -302,19 +302,19 @@ function _compute_inherited_glueing(gd::InheritGlueingData)
 
   UX = intersect(U, codomain(iso_X))
   UX isa PrincipalOpenSubset && ambient_scheme(UX) === A || error("incorrect intermediate result")
-  h_Y = pullback(f)(complement_equation(codomain(iso_Y)))
+  h_Y = pullback(f)(complement_equation(codomain(iso_Y)), check=false)
   UXY = PrincipalOpenSubset(codomain(iso_X), 
-                            OO(codomain(iso_X))(lifted_numerator(h_Y)*complement_equation(U)))
+                            OO(codomain(iso_X))(lifted_numerator(h_Y)*complement_equation(U), check=false))
   UXY isa PrincipalOpenSubset && ambient_scheme(UXY) === codomain(iso_X) || error("incorrect intermediate output")
   UY = PrincipalOpenSubset(U, h_Y)
   XY = PrincipalOpenSubset(X, pullback(iso_X)(complement_equation(UXY)))
 
   VY = intersect(V, codomain(iso_Y))
   VY isa PrincipalOpenSubset && ambient_scheme(VY) === B || error("incorrect intermediate result")
-  h_X = pullback(g)(complement_equation(codomain(iso_X)))
+  h_X = pullback(g)(complement_equation(codomain(iso_X)), check=false)
   VX = PrincipalOpenSubset(V, h_X)
   VYX = PrincipalOpenSubset(codomain(iso_Y), 
-                            OO(codomain(iso_Y))(lifted_numerator(h_X)*complement_equation(V)))
+                            OO(codomain(iso_Y))(lifted_numerator(h_X)*complement_equation(V), check=false))
   VYX isa PrincipalOpenSubset && ambient_scheme(VYX) === codomain(iso_Y) || error("incorrect intermediate output")
   YX = PrincipalOpenSubset(Y, pullback(iso_Y)(complement_equation(VYX)))
 
@@ -323,7 +323,7 @@ function _compute_inherited_glueing(gd::InheritGlueingData)
 
   x_img = gens(OO(X))
   x_img = pullback(inverse(iso_X)).(x_img)
-  x_img = OO(UXY).(x_img)
+  x_img = [OO(UXY)(x, check=false) for x in x_img]
   x_img = pullback(gres).(x_img)
   phi = restrict(iso_Y, YX, VYX, check=false)
   x_img = pullback(phi).(x_img)
@@ -331,7 +331,7 @@ function _compute_inherited_glueing(gd::InheritGlueingData)
 
   y_img = gens(OO(Y))
   y_img = pullback(inverse(iso_Y)).(y_img)
-  y_img = OO(VYX).(y_img)
+  y_img = [OO(VYX)(y, check=false) for y in y_img]
   y_img = pullback(fres).(y_img)
   psi = restrict(iso_X, XY, UXY, check=false)
   y_img = pullback(psi).(y_img)
--- a/experimental/Schemes/CoherentSheaves.jl
+++ b/experimental/Schemes/CoherentSheaves.jl
@@ -689,8 +689,8 @@ function twisting_sheaf(IP::AbsProjectiveScheme{<:Field}, d::Int)
     (UU, VV) = glueing_domains(G)
     h_U = complement_equation(UU)
     h_V = complement_equation(VV)
-    MG[(U, V)] = (d>= 0 ? OO(VV)(h_V^d) : (inv(OO(VV)(h_V))^(-d)))*identity_matrix(OO(VV), 1)
-    MG[(V, U)] = (d>= 0 ? OO(UU)(h_U^d) : (inv(OO(UU)(h_U))^(-d)))*identity_matrix(OO(UU), 1)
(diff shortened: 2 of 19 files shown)


src/Rings/mpoly-graded.jl
```diff
@@ -588,7 +588,7 @@ function (W::MPolyDecRing{T})(c::Vector{T}, e::Vector{Vector{Int}}) where T
   return W(forget_decoration(W)(c, e))
 end
 
-(W::MPolyDecRing)(g::MPolyDecRingElem) = MPolyDecRingElem(forget_decoration(W)(forget_decoration(g)), W)
+(W::MPolyDecRing)(g::MPolyDecRingElem) = MPolyDecRingElem(forget_decoration(g), W)
```
HechtiDerLachs:
```suggestion
(W::MPolyDecRing)(g::MPolyDecRingElem) = MPolyDecRingElem(forget_decoration(W)(forget_decoration(g)), W)
```

experimental/Schemes/Auxiliary.jl
```diff
@@ -323,15 +323,15 @@ function _compute_inherited_glueing(gd::InheritGlueingData)
...
+  x_img = [OO(UXY)(x, check=false) for x in x_img]
   x_img = pullback(gres).(x_img)
   phi = restrict(iso_Y, YX, VYX, check=false)
   x_img = pullback(phi).(x_img)
   gg = SpecMor(YX, XY, hom(OO(XY), OO(YX), x_img, check=false), check=false)
 
   y_img = gens(OO(Y))
   y_img = pullback(inverse(iso_Y)).(y_img)
-  y_img = OO(VYX).(y_img)
+  y_img = [OO(VYX)(y, check=false) for y in y_img]
```
simonbrandhorst, line 334:
```suggestion
  y_img = OO(VYX).(y_img; check=false)
```
You can broadcast with keyword arguments too as far as I remember.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -689,8 +689,8 @@ function twisting_sheaf(IP::AbsProjectiveScheme{<:Field}, d::Int)
     (UU, VV) = glueing_domains(G)
     h_U = complement_equation(UU)
     h_V = complement_equation(VV)
-    MG[(U, V)] = (d>= 0 ? OO(VV)(h_V^d) : (inv(OO(VV)(h_V))^(-d)))*identity_matrix(OO(VV), 1)
-    MG[(V, U)] = (d>= 0 ? OO(UU)(h_U^d) : (inv(OO(UU)(h_U))^(-d)))*identity_matrix(OO(UU), 1)
+    MG[(U, V)] = (d>= 0 ? (x->OO(VV)(x, check=false))(h_V^d) : (inv((x->OO(VV)(x, check=false))(h_V))^(-d)))*identity_matrix(OO(VV), 1)
```
simonbrandhorst: you could just create a matrix directly instead of multiplying by the identity?

src/Rings/mpoly-localizations.jl
```diff
@@ -143,13 +143,24 @@ function simplify!(S::MPolyPowersOfElement)
...
   S.a = new_denom
   return S
 end
 
+function simplify(S::MPolyPowersOfElement)
+  R = ambient_ring(S)
+  new_denom = Vector{elem_type(R)}()
+  for d in denominators(S)
+    for a in factor(d)
+      !(a in new_denom) && push!(new_denom, a[1])
```
simonbrandhorst:
```suggestion
      !(a[1] in new_denom) && push!(new_denom, a[1])
```
?

simonbrandhorst (approved):
Looks good to me once the tests pass

experimental/Schemes/Auxiliary.jl
HechtiDerLachs: I didn't know. Thanks.

experimental/Schemes/CoherentSheaves.jl
HechtiDerLachs: I do not really see what you are doing here, sorry.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: Oops... Yes, this should probably be fixed. Thanks.

experimental/Schemes/CoherentSheaves.jl
fingolfin: Neither do I, in either the old or new version of that code. So perhaps that can be improved, by adding a comment, and/or splitting this overly clever computation into multiple lines? E.g. instead of
```
MG[(U, V)] = (d>= 0 ? (x->OO(VV)(x, check=false))(h_V^d) : (inv((x->OO(VV)(x, check=false))(h_V))^(-d)))*identity_matrix(OO(VV), 1)
```
how about
```
if d >= 0
  tmp = (x->OO(VV)(x, check=false))(h_V^d)
else
  tmp = inv((x->OO(VV)(x, check=false))(h_V))^(-d)
end
MG[(U, V)] = tmp*identity_matrix(OO(VV), 1)
```

This is already *much* easier to read and understand, and to me immediately shows further potential for improvement, resp. makes me ask questions; e.g.
- that last line should probably be using `diagonal_matrix` (please read its docstring);
- the auxiliary function `x->OO(VV)(x, check=false)` seems completly redundant? I.e. why can't `(x->OO(VV)(x, check=false))(h_V^d)` be replaced by `OO(VV)(h_V^d, check=false)` ?
- in one place first power then apply  `OO(VV)`, in the other it's the other way around. Intentional? Does it make a difference?

src/Rings/MPolyQuo.jl
```diff
@@ -907,26 +907,26 @@ lift(a::MPolyQuoRingElem) = a.f
 
 (Q::MPolyQuoRing)() = MPolyQuoRingElem(base_ring(Q)(), Q)
 
-function (Q::MPolyQuoRing)(a::MPolyQuoRingElem)
+function (Q::MPolyQuoRing)(a::MPolyQuoRingElem; check::Bool=true)
   if parent(a) === Q
     return a
   else
     return Q(base_ring(Q)(a))
```
fingolfin, line 914: Shouldn't the `check=false` be propagated here?

(For that matter, it is not clear to me why this code does not lead to an infinite recursion, but that's a different question ;-) )

HechtiDerLachs: No, `true` should be correct here. You might have different moduli in different quotient rings and mapping the representative might not be well defined.

experimental/Schemes/CoherentSheaves.jl
```diff
@@ -689,8 +689,8 @@ function twisting_sheaf(IP::AbsProjectiveScheme{<:Field}, d::Int)
     (UU, VV) = glueing_domains(G)
     h_U = complement_equation(UU)
     h_V = complement_equation(VV)
-    MG[(U, V)] = (d>= 0 ? OO(VV)(h_V^d) : (inv(OO(VV)(h_V))^(-d)))*identity_matrix(OO(VV), 1)
-    MG[(V, U)] = (d>= 0 ? OO(UU)(h_U^d) : (inv(OO(UU)(h_U))^(-d)))*identity_matrix(OO(UU), 1)
+    MG[(U, V)] = diagonal_matrix((d>= 0 ? (x->OO(VV)(x, check=false))(h_V^d) : (inv((x->OO(VV)(x, check=false))(h_V))^(-d))), 1)
+    MG[(V, U)] = diagonal_matrix((d>= 0 ? (x->OO(UU)(x, check=false))(h_U^d) : (inv((x->OO(UU)(x, check=false))(h_U))^(-d))), 1)
```
fingolfin, line 693: This is better but ignores the other points I made...

src/Rings/MPolyQuo.jl
simonbrandhorst: maybe @fingolfin  means
```suggestion
    return Q(base_ring(Q)(a), check=check)
```

HechtiDerLachs: I would like this. But since quotient rings are not really my field to work in, I wanted to keep interference as low as possible.

experimental/Schemes/CoherentSheaves.jl
HechtiDerLachs: Yes, sorry. I think you were right. But I don't think it's worth it to run all the tests again just for this now.

src/Rings/MPolyQuo.jl
simonbrandhorst: If someone complains, you can send them to me.

simonbrandhorst:
I agree that there are some improvements possible, but they should not delay merging. I will merge after lunch if no one disagrees.

Issue #2498, opened 2023-06-29
Oscar seems to spend a lot of time comparing monomial orderings for lookups in dictionaries


HechtiDerLachs:
I had already mentioned this several times, but could never really reproduce the problem. Let me make another attempt here with the following test file:
```julia
using ProfileView
X = projective_space(QQ, 3)
E = direct_sum([twisting_sheaf(X, 0), twisting_sheaf(X, -2)])
Y = covered_scheme(projectivization(E))
C = oscar.simplified_covering(Y)
U = patches(C)
GG = glueings(C)
# trigger the computation once to reduce compilation overhead
underlying_glueing(GG[(U[1],U[end])])
for i in 1:length(U)
  for j in 1:length(U)
    @show i, j
    @profview underlying_glueing(GG[(U[i],U[j])])
  end
end
```
This produces some (actually a lot) of flame graphs. Most of them should exhibit the following problem: For every ideal membership test in polynomial rings, a cached groebner basis is looked up in a dictionary. The keys for this dictionary are `MonomialOrdering`s and these are compared using the `==` operator. To this end, it seems that a `canonical_matrix` is computed and then the equality check is performed on these matrices. 

For some of my samples **this uses up more than 50% of the actual computing time**. Since I don't know how to send the interactive flame graphs, I have made a series of screenshots to illustrate what I mean. I will send this as a .zip on the oscar slack.

HechtiDerLachs:
I tried a dirty hack by modifying the `ideal_membership` routine in `src/Rings/mpoly-ideals.jl` to the following:
```julia
function ideal_membership(f::T, I::MPolyIdeal{T}; ordering::MonomialOrdering = default_ordering(base_ring(I))) where T
  if !isempty(keys(I.gb))
    GI = first(values(I.gb))
    singular_assure(GI)
    Sx = base_ring(GI.S)
    return Singular.iszero(Singular.reduce(Sx(f), GI.S))
  end
  GI = standard_basis(I, ordering=ordering, complete_reduction=false)
  singular_assure(GI)
  Sx = base_ring(GI.S)
  return Singular.iszero(Singular.reduce(Sx(f), GI.S))
end
```
With the modified script from above
```julia

X = projective_space(QQ, 3)
E = direct_sum([twisting_sheaf(X, 0), twisting_sheaf(X, -2)])
Y = covered_scheme(projectivization(E))
C = oscar.simplified_covering(Y)
U = patches(C)
GG = glueings(C)
[...]

PR #2499, opened 2023-06-29
Cache `canonical_matrix(::MonomialOrdering)`

lgoettgens:
Implements the proposal of https://github.com/oscar-system/Oscar.jl/issues/2498#issuecomment-1612689919.

The speed change is shown in https://github.com/oscar-system/Oscar.jl/issues/2498#issuecomment-1612720461.

Feel free to close this and propose an alternative to this @HechtiDerLachs.

### DIFF
--- a/src/Rings/orderings.jl
+++ b/src/Rings/orderings.jl
@@ -169,11 +169,16 @@ end
 """
 Orderings actually applied to polynomial rings (as opposed to variable indices)
 """
-mutable struct MonomialOrdering{S}
+@attributes mutable struct MonomialOrdering{S}
   R::S
   o::AbsGenOrdering
   is_total::Bool
   is_total_is_known::Bool
+
+  # default constructor for @attributes
+  function MonomialOrdering{S}(R::S, o::AbsGenOrdering, is_total::Bool, is_total_is_known::Bool) where S
+    return new{S}(R, o, is_total, is_total_is_known)
+  end
 end
 
 function MonomialOrdering(R::S, o::AbsGenOrdering) where S
@@ -1312,7 +1317,7 @@ julia> canonical_matrix(o2)
 [0   -1   0]
 ```
 """
-function canonical_matrix(M::MonomialOrdering)
+@attr ZZMatrix function canonical_matrix(M::MonomialOrdering)
   return canonical_matrix(nvars(base_ring(M)), M.o)
 end
 
--- a/src/Rings/orderings.jl
+++ b/src/Rings/orderings.jl
@@ -169,16 +169,12 @@ end
 """
 Orderings actually applied to polynomial rings (as opposed to variable indices)
 """
-@attributes mutable struct MonomialOrdering{S}
+mutable struct MonomialOrdering{S}
   R::S
   o::AbsGenOrdering
   is_total::Bool
   is_total_is_known::Bool
-
-  # default constructor for @attributes
-  function MonomialOrdering{S}(R::S, o::AbsGenOrdering, is_total::Bool, is_total_is_known::Bool) where S
-    return new{S}(R, o, is_total, is_total_is_known)
-  end
+  canonical_matrix::ZZMatrix
 end
 
 function MonomialOrdering(R::S, o::AbsGenOrdering) where S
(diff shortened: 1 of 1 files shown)


src/Rings/orderings.jl
```diff
@@ -169,11 +169,16 @@ end
 """
 Orderings actually applied to polynomial rings (as opposed to variable indices)
 """
-mutable struct MonomialOrdering{S}
+@attributes mutable struct MonomialOrdering{S}
   R::S
   o::AbsGenOrdering
   is_total::Bool
   is_total_is_known::Bool
```
HechtiDerLachs, line 176:
```suggestion
  is_total_is_known::Bool
  canonical_matrix::ZZMatrix
```

src/Rings/orderings.jl
```diff
@@ -1312,7 +1317,7 @@ julia> canonical_matrix(o2)
 [0   -1   0]
 ```
 """
-function canonical_matrix(M::MonomialOrdering)
+@attr ZZMatrix function canonical_matrix(M::MonomialOrdering)
   return canonical_matrix(nvars(base_ring(M)), M.o)
 end
```
HechtiDerLachs:
```suggestion
function canonical_matrix(M::MonomialOrdering)
  if !isdefined(M, :canonical_matrix)
    M.canonical_matrix = canonical_matrix(nvars(base_ring(M)), M.o)
  end
  return M.canonical_matrix
end
```

src/Rings/orderings.jl
```diff
@@ -169,11 +169,16 @@ end
 """
 Orderings actually applied to polynomial rings (as opposed to variable indices)
 """
-mutable struct MonomialOrdering{S}
+@attributes mutable struct MonomialOrdering{S}
```
HechtiDerLachs:
```suggestion
mutable struct MonomialOrdering{S}
```

HechtiDerLachs (commented):
I'd suggest not to go via attributes. This just entails another dictionary lookup and this is data structure is very low-level, i.e. it is called with high frequency.

HechtiDerLachs:
And as @fieker said: Avoiding the lookup of a groebner basis in the first place is also possible and should be quicker.

But thanks for the very quick and constructive response!

src/Rings/orderings.jl
fieker: In general, if s.th. should be stored in almost all instances, I'd not put it in attributes, but into the structure. This allows for discovery...

fingolfin:
This just hides over the actual problem, which is that `canonical_matrix` is being called far too often. See the discussion in #2099

fingolfin:
Lest I be misunderstood: I still think this PR is useful and I am fine with merging it, but I removed the `Resolves #2498` because I think it would be a bad idea to keep fixing these "leave issues" instead of attacking the root.

src/Rings/orderings.jl
```diff
@@ -174,6 +174,7 @@ mutable struct MonomialOrdering{S}
   o::AbsGenOrdering
   is_total::Bool
   is_total_is_known::Bool
+  canonical_matrix::ZZMatrix
 end
 
 function MonomialOrdering(R::S, o::AbsGenOrdering) where S
```
fingolfin: I think you need to adjust this constructor to fix the errors: either make it an non-parametric inner constructor (so use `new{S}` in the next line); or else add a new inner parametric constructor that allows omitting the final field.

src/Rings/orderings.jl
```diff
@@ -174,6 +174,7 @@ mutable struct MonomialOrdering{S}
   o::AbsGenOrdering
   is_total::Bool
   is_total_is_known::Bool
+  canonical_matrix::ZZMatrix
```
fingolfin, line 177:
```suggestion
  canonical_matrix::ZZMatrix
  function MonomialOrdering(R::S, o::AbsGenOrdering) where S
     return new{S}(R, o, false, false)
  end
```
or for more flexibility:
```suggestion
  canonical_matrix::ZZMatrix
  function MonomialOrdering(R::S, o::AbsGenOrdering) where S
     return new{S}(R, o, false, false)
  end
  function MonomialOrdering(R::S, o::AbsGenOrdering, is_total::Bool) where S
     return new{S}(R, o, is_total, true)
  end
```
and then adjust the outer constructor below

simonbrandhorst:
I am glad for anything that helps my computations to actually terminate.

simonbrandhorst (approved):

PR #2500, opened 2023-06-29
Add `mres` and `nres` Singular Calls for Free resolutions

RafaelDavidMohr:
Adds corresponding keywords to `free_resolution` and adjusts docu accordingly.

### DIFF
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -6465,7 +6465,7 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     res = Singular.mres(gbpres, length)
   elseif algorithm == :nres
     gbpres = singular_kernel_entry
-    res = Singular.mres(gbpres, length)
+    res = Singular.nres(gbpres, length)
   else
     error("Unsupported algorithm $algorithm")
   end
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -6371,7 +6371,8 @@ end
 Return a free resolution of `M`.
 
 If `length != 0`, the free resolution is only computed up to the `length`-th free module.
-At the moment, `algorithm` only allows the option `:fres`.
+At the moment, `algorithm` options `:fres`, `:mres` and `:nres`. With `:mres` or `:nres`,
+minimal free resolutions are returned.
 
 # Examples
 ```jldoctest
@@ -6460,6 +6461,9 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     error("LaScala's method is not yet available in Oscar.")
     gbpres = singular_kernel_entry # or as appropriate, taking into account base changes
   elseif algorithm == :mres
+    gbpres = singular_kernel_entry
+    res = Singular.mres(gbpres, length)
+  elseif algorithm == :nres
     gbpres = singular_kernel_entry
     res = Singular.mres(gbpres, length)
   else
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -6436,7 +6436,7 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
 
   #= Start with presentation =#
   pm = presentation(M)
-  maps = [map(pm, j) for j in Hecke.map_range(pm)]
+  maps = [pm.maps[j] for j in 2:3]
 
   br = base_ring(M)
   kernel_entry          = image(pm.maps[1])[1]
@@ -6459,6 +6459,9 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
   elseif algorithm == :lres
     error("LaScala's method is not yet available in Oscar.")
     gbpres = singular_kernel_entry # or as appropriate, taking into account base changes
+  elseif algorithm == :mres
+    gbpres = singular_kernel_entry
+    res = Singular.mres(gbpres, length)
   else
     error("Unsupported algorithm $algorithm")
   end
@@ -6467,23 +6470,6 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     cc_complete = true
   end
 
-  codom = codomain(maps[1])
-
-  if is_graded(codom)
-    rk    = Singular.ngens(gbpres)
-    SM    = SubModuleOfFreeModule(codom, gbpres)
-    generator_matrix(SM)
-    ff = graded_map(codom, SM.matrix)
-    dom = domain(ff)
-  else
-    dom   = free_module(br, Singular.ngens(gbpres))
-    SM    = SubModuleOfFreeModule(codom, gbpres)
-    generator_matrix(SM)
-    ff = hom(dom, codom, SM.matrix)
-  end
-
-  maps[1] = ff
-
   br_name = AbstractAlgebra.find_name(base_ring(M))
   if br_name === nothing
     br_name = "R"
@@ -6491,10 +6477,10 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
 
   #= Add maps from free resolution computation, start with second entry
    = due to inclusion of presentation(M) at the beginning. =#
-  j   = 2
+  j   = 1
   while j <= Singular.length(res)
-    if is_graded(dom)
-      codom = dom
+    if is_graded(M)
+      codom = domain(maps[1])
       rk    = Singular.ngens(res[j])
       SM    = SubModuleOfFreeModule(codom, res[j])
       generator_matrix(SM)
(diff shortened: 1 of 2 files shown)


src/Modules/UngradedModules.jl
```diff
@@ -6361,6 +6362,10 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     res = Singular.fres(singular_kernel_entry, length, "complete")
   elseif algorithm == :sres
     res = Singular.fres(singular_kernel_entry, length)
+  elseif algorithm == :mres
```
fingolfin: Don't these also require a recent enough Singular.jl? If so then Project.toml should be adjusted, too

RafaelDavidMohr: Sure! I am not sure I am doing this correctly in https://github.com/oscar-system/Oscar.jl/pull/2500/commits/e72cb47fb35fe8d53dbb9deee7dec1d77582bf52, do I just by hand bump the compat entry in Project.toml? Never had to fiddle around with this before.

src/Modules/UngradedModules.jl
```diff
@@ -6361,6 +6362,10 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     res = Singular.fres(singular_kernel_entry, length, "complete")
   elseif algorithm == :sres
     res = Singular.fres(singular_kernel_entry, length)
```
wdecker: There seems to be an old bug: you want sres but ask for fres!

wdecker:
Once this is merged, I will extend the docu and explain what the difference between the different methods is.

src/Modules/UngradedModules.jl
RafaelDavidMohr: Fixed in https://github.com/oscar-system/Oscar.jl/pull/2500/commits/d6abd9fff03290d7a8cf394abccbfe60a73ae88d

Project.toml
```diff
@@ -35,7 +35,7 @@ Polymake = "0.10.0"
 Preferences = "1"
 RandomExtensions = "0.4.3"
 RecipesBase = "1.2.1"
-Singular = "0.18.2"
+Singular = "0.18.6"
```
fingolfin: This looks right (I didn't check the versions, but I trust you did).

Thanks!

fingolfin:
This PR has no test for the new code, so naturally it will pass, even if actually calling the code was broken.

So either a test should be added in this PR (could be a doctest or could be a regular test); or else I'd hope that @wdecker would add one when he extends the documentation for this?

RafaelDavidMohr:
A word on the now failing tests: I could be wrong, but to me it seems that the issue is the following:

Oscar starts computing a free resolution by computing the input module as a cokernel and then passes that cokernel to Singular to let it compute a free resolution for that. `Singular.sres` returns this free resolution but with the generators of this cokernel potentially permuted. Now the presentation of the input module and the Singular resolution don't "patch together" correctly and thus give a wrong free resolution in Oscar. 

My question is: Should we try to handle this bug in this PR or a seperate one?

wdecker (commented):
There are several problems here: To begin with, fres and sres require as input a gröbner basis. This is missing so far. There is another problem on which @jankoboehm is already working. 

For the time being, I suggest to replace sres by fres in the test. If this makes the test pass, let us merge the PR. More tests and docu will be added as soon as all problems are fixed. @RafaelDavidMohr please open an issue which reminds us to do the above.

RafaelDavidMohr:
Unfortunately the same problem seems to be happening with `mres`: Singular changes the cokernel representation of the module and then the free resolution Singular gives doesn't glue correctly with what is computed on the Oscar side.

So `mres` would require similar changes to the free resolution code that `sres` needs.

wdecker:
The situation for `mres` is different and more subtle as `mres` computes a new presentation. In particular, this requires that we return an isomorphism between the given module and the new module. That means that we have to introduce some book-keeping in Singular and add the result of this to the output of `mres`. So let me suggest to make this PR a draft until all these prblems have been addressed.

RafaelDavidMohr:
> The situation for `mres` is different and more subtle as `mres` computes a new presentation. In particular, this requires that we return an isomorphism between the given module and the new module. That means that we have to introduce some book-keeping in Singular and add the result of this to the output of `mres`. So let me suggest to make this PR a draft until all these prblems have been addressed.

Ok! In a sence, `sres` also computes a new representation (but just by shuffling around generators). 

Let me know if I can help with fixing this.

fingolfin:
We think there are some changes in the latest Singular.jl that may help, but couldn't find them just now; but @hannes14 will check

RafaelDavidMohr:
I added `mres`, the bug that I originally described (with Singular changing the presentation of some module) looks to be actually easily fixed by some global changes to the `free_resolution` function, we just have to also take the first map that Singular gives us.

Let me know if I missed something.

ederc:
Looks like a valid solution to me.

test/Modules/UngradedModules.jl
```diff
@@ -171,16 +171,20 @@ end
 	B = R[x^2; x*y; y^2; z^4]
 	M = SubquoModule(A, B)
 	free_res = free_resolution(M, length=1)
-    @test is_complete(free_res) == false
-    free_res[2]
-    @test length(free_res.C.maps) == 4
+  @test is_complete(free_res) == false
+  free_res[2]
+  @test length(free_res.C.maps) == 4
```
lgoettgens, line 176: This part seems to be indented weirdly (when viewing it on github). Maybe there is a mixup of tabs and spaces here?

RafaelDavidMohr: There were some 4 space indentations in this file which I removed, everything looks good in my local files.

lgoettgens: Let's just leave this as is and let https://github.com/oscar-system/Oscar.jl/pull/2785 handle the tab issue.

fingolfin:
If the geometry would like to merge this, fine by me, but someone needs to explicitly approve it

wdecker:
We are still discussing it. Will be merged tomorrow.

RafaelDavidMohr:
Here is a bit more explanation on why things should work now. Originally I diagnosed the problem wrong.

Previously, for a given module `M`, `free_resolution(M)` would perform the following computations:

We start by computing a presentation of `M`
```
0 <-- M <-- F <-- G
```
then we let `C` be the cokernel of this representation (i.e. the image of `G` in `F`) and let Singular compute a free resolution of this cokernel:
```
0 <-- F/C <-- F0 <-- F1 <-- F2 <-- ...
```
and now the two chain complexes were patched together as follows:
```
0 <-- M <-- F <-- G <-- F2 <-- ...
```
`F` and `F0` are always the same, but the problematic part is `F <-- G <-- F2` because the image of  `G` in `F` (`==F0`) might have its generators sorted differently or altogether changed  compared to the image of `F1` in `F0`. Then the homology is nonzero at `G`. This is now fixed by instead returning
```
0 <-- M <-- F <-- F1 <-- F2 <-- ...
```
This is now always a correct free resolution (because we basically just take the one Singular gives us, `F = F0` is always guaranteed). I hope this helps.

wdecker:
@RafaelDavidMohr Thx for doing this.

I have a small remark: The methods

   `free_resolution(I::MPolyIdeal)`

   `free_resolution(Q::MPolyQuoRing)`

and also the "dummy" method

   `free_resolution(F::FreeMod)`

should be adjusted to the syntax we have now for `free_resolution`.

jankoboehm:
As I understand it, we switch the presentation matrix originally computed by OSCAR to the (in case of mres, not fres) possibly modified one returned by Singular. Are we sure that this is the same module and not just an isomorphic one (then we have to keep track of the isomorphism), what if we resolve a submodule. We cannot change the generators, since the representation of the elements is referencing them. It also looks to me as if the grading is removed? Best have a quick talk in person once possible.

RafaelDavidMohr:
> As I understand it, we switch the presentation matrix originally computed by OSCAR to the (in case of mres, not fres) possibly modified one returned by Singular. Are we sure that this is the same module and not just an isomorphic one (then we have to keep track of the isomorphism), what if we resolve a submodule. We cannot change the generators, since the representation of the elements is referencing them. It also looks to me as if the grading is removed? Best have a quick talk in person once possible.

Yes, in person (Saarbrücken for instance) is probably easier. I will however have a look at the grading issue and at the things @wdecker mentioned.

Edit: Actually, @jankoboehm, I am not sure what you mean by "the grading is removed". I tried a random example and I'm getting a graded free resolution. Could you specify?

RafaelDavidMohr:
> @RafaelDavidMohr Thx for doing this.
> 
> I have a small remark: The methods
> 
> `free_resolution(I::MPolyIdeal)`
> 
> `free_resolution(Q::MPolyQuoRing)`
> 
> and also the "dummy" method
> 
> `free_resolution(F::FreeMod)`
> 
> should be adjusted to the syntax we have now for `free_resolution`.

See https://github.com/oscar-system/Oscar.jl/pull/2500/commits/347349cfef508bef64bc977876118e02bf0efccc. I adjusted the docu accordingly, the docu for `free_resolution(F::FreeMod)` is not ideal, let me know if I should change it.

wdecker:
This is good for now. However, for the `mres` part, we also need to minimize the presentation we work with. This requires some work in Singular first. We will fix this asap.

wdecker (approved):

PR #2502, opened 2023-06-30
Introduce lazy computation of glueing domains.

HechtiDerLachs:
As discussed with @simonbrandhorst .

### DIFF
--- a/experimental/Schemes/LazyGlueing.jl
+++ b/experimental/Schemes/LazyGlueing.jl
@@ -14,6 +14,8 @@
   GD::GlueingDataType
   compute_function::Function
   compute_glueing_domains::Function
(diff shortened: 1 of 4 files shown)


simonbrandhorst (approved):

HechtiDerLachs:
It seems that something with Polymake is breaking the CI here. @benlorenz : Do you have an idea of what's going on here? 

There will be more errors due to my code, but this might be something different, so I wanted to ask.

benlorenz:
The breakage is "just" julia nightly and everything that depends on `libjulia` (GAP, CxxWrap, libsingular, libpolymake) fails to precompile due to https://github.com/JuliaLang/julia/pull/49012.

Working on it: https://github.com/JuliaPackaging/Yggdrasil/pull/6962

PR #2504, opened 2023-06-30
Decomposition info

HechtiDerLachs:
We introduce a new pattern to keep track of a decomposition of a covered scheme as a disjoint union. The extended version of this should allow for the computation of intersection numbers without computing any glueings.

Let me know, what you think @simonbrandhorst @afkafkafk13 .

### DIFF
--- a/experimental/Schemes/WeilDivisor.jl
+++ b/experimental/Schemes/WeilDivisor.jl
@@ -210,34 +210,49 @@ function intersect(D::WeilDivisor, E::WeilDivisor)
   return result
 end
 
-function colength(I::IdealSheaf)
+function colength(I::IdealSheaf; covering::Covering=default_covering(scheme(I)))
   X = scheme(I)
-  C = default_covering(X)
   patches_todo = copy(affine_charts(X))
   patches_done = AbsSpec[]
   result = 0
   while length(patches_todo) != 0
     U = pop!(patches_todo)
     J = I(U)
-    # To avoid overcounting, throw away all components that 
-    # were already visible in other charts.
-    for V in patches_done
-      if !haskey(glueings(C), (U, V))
-        continue
+    if has_decomposition_info(covering)
+      h = decomposition_info(covering)[U]
+      # The elements in h indicate where components must 
+      # be located so that they can not be spotted in other charts.
+      # We iteratively single out these components by adding a sufficiently high 
+      # power of the equation to the ideal.
+      for f in h
+        k = 0
+        while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
+          k = k + 1
+        end
+        J = J + ideal(OO(U), f^k)
+        isone(J) && break
       end
-      G = C[U, V]
-      (UV, VU) = glueing_domains(G)
-      UV isa PrincipalOpenSubset || error("method is only implemented for simple glueings")
-      f = complement_equation(UV)
-      # Find a sufficiently high power of f such that it throws
-      # away all components away from the horizon, but does not affect
-      # those on the horizon itself.
-      k = 0
-      while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
-        k = k + 1
+    else
+      # To avoid overcounting, throw away all components that 
+      # were already visible in other charts.
+      for V in patches_done
+        if !haskey(glueings(covering), (U, V))
+          continue
+        end
+        G = covering[U, V]
+        (UV, VU) = glueing_domains(G)
+        UV isa PrincipalOpenSubset || error("method is only implemented for simple glueings")
+        f = complement_equation(UV)
+        # Find a sufficiently high power of f such that it throws
+        # away all components away from the horizon, but does not affect
+        # those on the horizon itself.
+        k = 0
+        while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
+          k = k + 1
+        end
+        J = J + ideal(OO(U), f^k)
+        isone(J) && break
       end
-      J = J + ideal(OO(U), f^k)
-      isone(J) && break
     end
     if !isone(J)
       JJ = leading_ideal(saturated_ideal(J))
--- a/experimental/Schemes/WeilDivisor.jl
+++ b/experimental/Schemes/WeilDivisor.jl
@@ -212,7 +212,7 @@ end
 
 function colength(I::IdealSheaf; covering::Covering=default_covering(scheme(I)))
   X = scheme(I)
-  patches_todo = copy(affine_charts(X))
+  patches_todo = copy(patches(covering))
   patches_done = AbsSpec[]
   result = 0
   while length(patches_todo) != 0
--- a/experimental/Schemes/WeilDivisor.jl
+++ b/experimental/Schemes/WeilDivisor.jl
@@ -225,11 +225,11 @@ function colength(I::IdealSheaf; covering::Covering=default_covering(scheme(I)))
       # We iteratively single out these components by adding a sufficiently high 
       # power of the equation to the ideal.
       for f in h
-        k = 0
-        while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
-          k = k + 1
+        g = f
+        while !(g in ideal(OO(U), g*f) + J)
+          g = g * g
         end
-        J = J + ideal(OO(U), f^k)
+        J = J + ideal(OO(U), g)
         isone(J) && break
       end
     else
@@ -246,11 +246,11 @@ function colength(I::IdealSheaf; covering::Covering=default_covering(scheme(I)))
         # Find a sufficiently high power of f such that it throws
         # away all components away from the horizon, but does not affect
         # those on the horizon itself.
-        k = 0
-        while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
-          k = k + 1
+        g = f
+        while !(g in ideal(OO(U), g*f) + J)
+          g = g * g
         end
-        J = J + ideal(OO(U), f^k)
+        J = J + ideal(OO(U), g)
         isone(J) && break
       end
     end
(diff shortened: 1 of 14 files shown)


test/AlgebraicGeometry/Schemes/IdealSheaves.jl
```diff
@@ -111,3 +111,16 @@ end
...
+  P3 = projective_space(QQ, 3)
+  X = covered_scheme(P3)
+  S = homogeneous_coordinate_ring(P3)
+  (x,y, z, w) = gens(S)
+  I = ideal(S, [x^3+y^3+z^3+w^3, x+y+z+w, 37*x^2-x*z+4*w^2-5*w*z])
+  II = ideal_sheaf(P3, I)
+  @test oscar.colength(II) == 6
+  J = ideal(S, [x, y, z^4])
+  JJ = ideal_sheaf(P3, J)
+  @test oscar.colength(JJ) == 4
```
HechtiDerLachs: These are the relevant tests for this PR.

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
```diff
@@ -46,3 +46,30 @@ function glueing_graph(C::Covering)
   return C.glueing_graph
 end
 
+@doc raw"""
+    decomposition_info(C::Covering)
+
+Return an `IdDict` `D` with the `patches` of `C` as keys and values `D[U]` a list 
+of elements ``f₁,…,fᵣ ∈ 𝒪(U)``. These elements are chosen so that for 
+``Xᵤ = V(f₁,…,fᵣ) ⊂ U`` the scheme ``X`` covered by ``C`` decomposes as a 
+disjoint union ``X = ∪ Xᵤ``.
```
HechtiDerLachs: This should explain the idea.

experimental/Schemes/WeilDivisor.jl
```diff
@@ -210,34 +210,49 @@ function intersect(D::WeilDivisor, E::WeilDivisor)
...
-    # were already visible in other charts.
-    for V in patches_done
-      if !haskey(glueings(C), (U, V))
-        continue
+    if has_decomposition_info(covering)
+      h = decomposition_info(covering)[U]
+      # The elements in h indicate where components must 
+      # be located so that they can not be spotted in other charts.
+      # We iteratively single out these components by adding a sufficiently high 
+      # power of the equation to the ideal.
```
HechtiDerLachs, line 226: ...and this here how this can be brought to work.

The next step would be to automatically keep track of this information for blowups. Then integration of cycles should go through in little time once we allow weil divisor components to be non-prime.

HechtiDerLachs:
If you agree with this concept, I could move forward with it.

afkafkafk13:
May I sum up the idea, how I understood it from the first glance (i5 minutes after receiving the first mail on this PR): 
In addition to the charts, you also note the 'focus' (not my terminology, introduced by Bodnar and Schicho 2002) such that intersection loci of the charts are only considered once. The disjoint union mentioned above is then the decomposition of the scheme into these constructible sets.

experimental/Schemes/WeilDivisor.jl
simonbrandhorst: We also need this information to down to subschemes.

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
simonbrandhorst: I don't get it.   What covers what and how? Please make a clean statement and use 
LaTeX. Here Unicode is not readable enough for me.

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
```diff
@@ -46,3 +46,30 @@ function glueing_graph(C::Covering)
...
+
+!!! note This attribute might not be defined!
```
simonbrandhorst: So one should not use this function?

simonbrandhorst (commented):
I think that overall this is a great Idea! But it needs a bit more documentation.

experimental/Schemes/WeilDivisor.jl
```diff
@@ -210,34 +210,49 @@ function intersect(D::WeilDivisor, E::WeilDivisor)
...
   result = 0
   while length(patches_todo) != 0
     U = pop!(patches_todo)
     J = I(U)
-    # To avoid overcounting, throw away all components that 
-    # were already visible in other charts.
-    for V in patches_done
-      if !haskey(glueings(C), (U, V))
-        continue
+    if has_decomposition_info(covering)
```
afkafkafk13, line 221: Before doing any further computations (i.e. above this line), you might want to check that the colength is finite -- to avoid being stuck in a loop forever.

experimental/Schemes/WeilDivisor.jl
```diff
@@ -210,34 +210,49 @@ function intersect(D::WeilDivisor, E::WeilDivisor)
...
+      # The elements in h indicate where components must 
+      # be located so that they can not be spotted in other charts.
+      # We iteratively single out these components by adding a sufficiently high 
+      # power of the equation to the ideal.
+      for f in h
+        k = 0
+        while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
+          k = k + 1
+        end
+        J = J + ideal(OO(U), f^k)
```
afkafkafk13: Do you need the exakt value of k? I do not think so. Hence you might want to keep squaring the polynomial instead of multiplying f to it in each pass of the loop. This avoids a lot of standard basis computations hidden in the ideal membership test.

experimental/Schemes/WeilDivisor.jl
```diff
@@ -210,34 +210,49 @@ function intersect(D::WeilDivisor, E::WeilDivisor)
...
+        (UV, VU) = glueing_domains(G)
+        UV isa PrincipalOpenSubset || error("method is only implemented for simple glueings")
+        f = complement_equation(UV)
+        # Find a sufficiently high power of f such that it throws
+        # away all components away from the horizon, but does not affect
+        # those on the horizon itself.
+        k = 0
+        while !(f^(k) in ideal(OO(U), f^(k+1)) + J)
+          k = k + 1
+        end
```
afkafkafk13: same comment as above.

afkafkafk13:
One general remark: Make sure to correctly carry along this information through a blow up and hence through any sequence of blow-ups.

HechtiDerLachs:
From my side this is good to go.

experimental/Schemes/WeilDivisor.jl
HechtiDerLachs: But this is another groebner basis computation and therefore should be hidden behind an `@check`.

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
```diff
@@ -46,3 +46,42 @@ function glueing_graph(C::Covering)
...
 end
 
+@doc raw"""
+    decomposition_info(C::Covering)
+
+Return an `IdDict` `D` with the `patches` ``Uᵢ`` of `C` as keys and values a list 
+of elements ``fᵢ₁,…,fᵢᵣ ∈ 𝒪(Uᵢ)``. These elements are chosen so that for every 
+affine patch `Uᵢ` of ``X`` in the covering `C` the closed subvarieties ``Zᵢ ⊂ Uᵢ`` 
+defined by the ``fᵢⱼ`` give rise to a decomposition of ``X`` as a **disjoint** union 
+``X = \bigcup_{i} Z_i`` of locally closed subvarieties.
```
simonbrandhorst, line 56: So $f_{i1}=0$?

simonbrandhorst: I still don't understand this.

simonbrandhorst: What is $Z_i$ exactly? I guess an example would help.

simonbrandhorst:
I set to auto merge. Please improve the docstring with the next pr.

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
HechtiDerLachs: $Z_i \subset U_i$ is the subvariety defined by the elements $f_{i,1}, \dots, f_{i,r}$ returned by the call `decomposition_info(X)[U_i]`.

simonbrandhorst: O.k. please add this to the docstring with the next pr.

simonbrandhorst: But still how do you get 

$X = \bigcup_{i} Z_i$?

Shouldn't $\bigcup_{i} Z_i\subseteq U_i \subset X$?

HechtiDerLachs: This must hold by construction of the `decomposition_info`. You do not get a decomposition with inclusion maps, etc. You get a list of ring elements for every patch and you may continue working with them while assuming that the decomposition described above is a disjoint union.

simonbrandhorst: Still don't get it. Because I do not get your definition of a decomposition_info. 
Please give clean mathematical definition.

experimental/Schemes/WeilDivisor.jl
simonbrandhorst: Please add this input check, so that we can merge this.

simonbrandhorst:
@HechtiDerLachs  promised to add the check later. So I merge this now.

simonbrandhorst (approved):

PR #2509, opened 2023-07-03
More lazy algebraic cycles

HechtiDerLachs:
As discussed: We allow equidimensional rather than prime ideals in the construction of algebraic cycles. Using `irreducible_decomposition` one may request a cycle in the old format.


HechtiDerLachs:
Intersection of Weil divisors on smooth projective surfaces now works without computing the glueings in the covering.

HechtiDerLachs:
@benlorenz  Just a heads up: Something's failing on nightly again. Probably you already know about it? To me it doesn't look like it's my fault, really.

test/AlgebraicGeometry/Schemes/WeilDivisor.jl
```diff
@@ -170,3 +171,35 @@ end
...
+  II = ideal_sheaf(P3, I)
+  P = covered_scheme(P3)
+  inc = oscar.CoveredClosedEmbedding(covered_scheme(P3), II)
+  X = domain(inc)
+  C1 = EffectiveCartierDivisor(ideal_sheaf(P3, [x+y+z+w]))
+  C2 = EffectiveCartierDivisor(ideal_sheaf(P3, [x^2*y + y^2*z + z^2*w + w^2*x]))
+  C1 = pullback(inc)(C1)
+  C2 = pullback(inc)(C2)
+  d = intersect(weil_divisor(C1), weil_divisor(C2))
+  pts = oscar.irreducible_decomposition(intersect(C1, C2))
```
HechtiDerLachs, line 209: The first line computes without using the glueings. The second one proceeds via the general method and first writes `C1` as an `AlgebraicCycle` with irreducible `components`. In particular, this uses `maximal_associated_points` and thus needs glueings.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -220,6 +220,9 @@ end
 function weil_divisor(C::EffectiveCartierDivisor;
     is_prime::Bool=false # Indicate whether this divisor is already prime
   )
+  return WeilDivisor(ideal_sheaf(C), ZZ, check=false)
+
+  # TODO: See what we can recycle from the code below.
```
HechtiDerLachs, line 225: @afkafkafk13 : You once implemented more specialized routines to extract the multiplicities. It seems that this was still a bottleneck, though. For the moment I switched back to the proof-of-concept implementation using composition series, but this is also very slow. Depending on your discussions with @simonbrandhorst this afternoon, we should overhaul your code and find a suitable binding for it again. For instance, one could specify a task here to directly produce a `WeilDivisor` with irreducible `components`.

experimental/Schemes/AlgebraicCycles.jl
```diff
@@ -275,6 +279,42 @@ end
...
 +(D::AbsAlgebraicCycle, I::IdealSheaf) = D + AbsAlgebraicCycle(I)
 
+@doc raw"""
+    irreducible_decomposition(D::AbsAlgebraicCycle)
+
+Returns a divisor ``E`` equal to ``D`` but as a formal sum ``E = ∑ₖ aₖ ⋅ Iₖ``
+where the `components` ``Iₖ`` of ``E`` are all sheaves of prime ideals.
+"""
+function irreducible_decomposition(D::AbsAlgebraicCycle)
+  all(I->is_prime(I), keys(coefficient_dict(D))) && return D
```
simonbrandhorst, line 311: At the moment `is_prime`  still calls `maximal_associated_points`. @afkafkafk13  has some ideas to make this faster.

experimental/Schemes/CartierDivisor.jl
```diff
@@ -248,10 +251,11 @@ function weil_divisor(C::CartierDivisor)
   return result
 end
 
-function intersect(W::WeilDivisor, C::EffectiveCartierDivisor)
+function intersect(W::WeilDivisor, C::EffectiveCartierDivisor; check::Bool=true)
```
simonbrandhorst, line 254: please document `intersect`

experimental/Schemes/IdealSheaves.jl
```diff
@@ -459,6 +459,32 @@ function is_prime(I::IdealSheaf)
...
+  is_equidimensional(saturated_ideal(I))
+end
+
+function is_equidimensional(I::MPolyLocalizedIdeal)
+  return is_equidimensional(saturated_ideal(I))
+end
+
+function is_equidimensional(I::MPolyQuoLocalizedIdeal)
+  return is_equidimensional(pre_image_ideal(I))
+end
```
simonbrandhorst, line 566: could this be unified by using `is_equidimentional(ambient_closure_ideal(I))`?

experimental/Schemes/IdealSheaves.jl
```diff
@@ -744,14 +770,28 @@ function match_on_intersections(
...
-        check || break
+      if UV isa SpecOpen && VU isa SpecOpen
+        I_res = [OOX(U, UV[i])(I) for i in 1:ngens(UV)]
+        IV_res = [OOX(V, UV[i])(IV) for i in 1:ngens(UV)]
+        if all(i->(I_res[i] == IV_res[i]), 1:ngens(UV))
+          match_found = !all(I->is_one(I), I_res)                               ## count only non-trivial matches
+          check || break
+        else
+          match_contradicted = true
+          check || break
```
simonbrandhorst, line 881: would it make sense to extend the functionality of SpecOpen a little more instead of making a special case here? 
In the long run extending `SpecOpen` a bit, will be more sustainable I hope.

experimental/Schemes/WeilDivisor.jl
```diff
@@ -133,6 +134,11 @@ function copy(D::WeilDivisor)
   return WeilDivisor(scheme(D), coefficient_ring(D), new_dict, check=false)
 end
 
+function irreducible_decomposition(D::WeilDivisor)
```
simonbrandhorst, line 128: cache this?

benlorenz:
> @benlorenz Just a heads up: Something's failing on nightly again. Probably you already know about it? To me it doesn't look like it's my fault, really.

I was more or less expecting it. We need another round of yggdrasil builds (some of them with updated compat entries) since julia 1.10 was branched and master is now 1.11-DEV.
cc: @fingolfin

fingolfin:
😭 I hope we'll get back to a state were we don't need to update JLLs every week to deal with Julia nightly changes

experimental/Schemes/AlgebraicCycles.jl
```diff
@@ -275,6 +279,42 @@ end
 
 +(D::AbsAlgebraicCycle, I::IdealSheaf) = D + AbsAlgebraicCycle(I)
 
+@doc raw"""
+    irreducible_decomposition(D::AbsAlgebraicCycle)
+
+Returns a divisor ``E`` equal to ``D`` but as a formal sum ``E = ∑ₖ aₖ ⋅ Iₖ``
```
fingolfin, line 307: Not a change request right now:  I don't think mixing math mode with unicode is a good idea. Pick one?

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: The problem with SpecOpen is that ideals over `SpecOpenRing`s do not really behave like we want them to. We don't know whether they are finitely generated and even if they were, there wouldn't be a method to compute lifts. All this accounted for not introducing ideals for these rings in the first place. But then we have to do the manual checking here. I don't know what's best.

simonbrandhorst: Mh yea, they are kind of a special case of global sections. 
But we can check for equality which is what we do here. 
What do you mean by lift? [This discussion should not hold us from merging this]

simonbrandhorst:
Merge conflicts.

experimental/Schemes/WeilDivisor.jl
afkafkafk13: definitely.

experimental/Schemes/AlgebraicCycles.jl
HechtiDerLachs: I think we should also make use of the `decomposition_info` when decomposing stuff. That should also allow us to evaluate only rather few glueings.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -609,22 +635,36 @@ Background:
...
+  result = IdealSheaf[]
+  # run through all charts and try to match the components
   while length(charts_todo) > 0
     U = pop!(charts_todo)
+    J = I(U)
+    # Do a quick check whether we even need to worry about this chart
+    if has_decomposition_info(covering)
+      J = J + ideal(OO(U), decomposition_info(covering)[U])
+      isone(J) && continue
+    end
```
HechtiDerLachs: Here starts what I had in mind to simplify decomposition. We basically avoid the matching process. The constructor for `IdealSheaf` which is finally called still needs to compute *some* path of glueings to go in every chart once, but this should only be `n` glueings, not `n^2/2` with `n` the number of charts.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: I mean `coordinates`, i.e. writing an element in the ideal as a linear combination of its generators. I doubt that it's worth introducing an extra type for (finitely generated) ideals over these rings only to implement the `==` method.

experimental/Schemes/WeilDivisor.jl
HechtiDerLachs: I wouldn't do this, really. An IdealSheaf is a lot of information. It can be one Groebner basis for every affine chart! If we keep caching these everywhere, then the garbage collector never has a chance to collect these things. And this is an instance where I would leave it to the user's responsibility to keep the information; for instance by just overwriting `D` with the output.

simonbrandhorst: I see the point. We are quite heavy on memory already.

experimental/Schemes/IdealSheaves.jl
simonbrandhorst: `O(n)` sound much better than `O(n^2)`

simonbrandhorst:
As a general comment: Would it make sense to implement a `decomposition_info`  as its own type? 
Along with some methods such as a pullback? That could make it more transparent.

HechtiDerLachs:
I have hope that with the changes from the last commit, we will also be able to decompose ideal sheaves more efficiently. When extending a (prime) ideal sheaf from one patch to others preference is given to those patches which can be reached via a glueing which has already been computed. It has not been properly tested though (this would require manual construction of rather exotic schemes, I guess), but I nevertheless wanted to expose it here so that we can start playing with it.

simonbrandhorst (approved):
lgtm

PR #2515, opened 2023-07-04
Use Rabinowitsch's trick for conversion.

HechtiDerLachs:
This is to speed up the computation of glueings in @simonbrandhorst 's code on elliptic surfaces. 

@wdecker : I have some questions about quotient rings which I will state below.

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -1046,11 +1046,11 @@ function divides(a::MPolyQuoRingElem, b::MPolyQuoRingElem)
   Q = parent(a)
   J = oscar_groebner_basis(Q)
 
-  BS = IdealGens([a.f], keep_ordering = false)
+  BS = IdealGens([a.f], keep_ordering = true)
   singular_assure(BS)
 
   J = vcat(J, [b.f])
-  BJ = IdealGens(J, keep_ordering = false)
+  BJ = IdealGens(J, keep_ordering = true)
   singular_assure(BJ)
 
   s, rest = Singular.lift(BJ.S, BS.S)
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -1046,11 +1046,11 @@ function divides(a::MPolyQuoRingElem, b::MPolyQuoRingElem)
   Q = parent(a)
   J = oscar_groebner_basis(Q)
 
-  BS = IdealGens([a.f], keep_ordering = true)
+  BS = IdealGens([a.f], keep_ordering = false)
   singular_assure(BS)
 
   J = vcat(J, [b.f])
-  BJ = IdealGens(J, keep_ordering = true)
+  BJ = IdealGens(J, keep_ordering = false)
   singular_assure(BJ)
 
   s, rest = Singular.lift(BJ.S, BS.S)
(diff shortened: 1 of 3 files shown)


src/Rings/MPolyQuo.jl
```diff
@@ -1046,11 +1046,11 @@ function divides(a::MPolyQuoRingElem, b::MPolyQuoRingElem)
   Q = parent(a)
   J = oscar_groebner_basis(Q)
 
-  BS = IdealGens([a.f], keep_ordering = false)
```
HechtiDerLachs: The constructor for `MPolyQuo` has a monomial ordering as a third argument. Here I came to appreciate it: In my application I use a block ordering with first block the "Rabinowitsch-variables" and then the original ones. 
It proved to be essential to use this ordering in the computations; for `degrevlex` everything died. 

However, it seems that not all internal methods use the ordering provided in the constructor. Some examples came through after changing these arguments here to `true`, so I assume it had the anticipated effect that my block ordering was used instead. Could you confirm that this is what it is supposed to do and that my changes don't do any harm here?

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2088,5 +2100,41 @@ function (W::MPolyDecRing)(f::MPolyLocRingElem)
...
+  end
+  Q = MPolyQuoRing(A, I, ordering)
+  function my_fun(g)
+    a = Q(phi(lifted_numerator(g)))
+    isone(lifted_denominator(b)) && return a
+    b = Q(phi(lifted_denominator(g)))
+    success, c = divides(a, b)
+    success || error("element can not be mapped")
+    return c
+  end
```
HechtiDerLachs: I had to use this complicated workaround because the usual mapping routine for homomorphisms `L -> Q` (`L` an `MPolyQuoLocRing` and `Q` an `MPolyQuo`) is using `inv` on the quotient ring `Q`. This refers to `is_invertible_with_inverse` and, I'm afraid, this is not using the block ordering provided by me. In that case, however, I did not find how to change this. Could someone look into this with me?

HechtiDerLachs:
Update: It seems that at the moment nobody really has an overview over the current design and the full implementation for quotient rings. We plan to have a coding sprint on this, eventually at some point in the beginning of August. 

CC: @afkafkafk13 @jankoboehm 

For the time being, I suppose that my changes here are OK.

HechtiDerLachs:
I'm getting the message "operation was cancelled" in the tests. Am I doing something wrong?

afkafkafk13:
> I'm getting the message "operation was cancelled" in the tests. Am I doing something wrong?

Just a guess: Run the tests on your local machine and check, how much memory you are using during the tests. One possible reason for such behaviour can be exhaustion of alotted memory.

src/Rings/MPolyQuo.jl
afkafkafk13: I quote from the type definition of MPolyQuoRing in MPolyQuo.jl (lines 13-14):
``` 
  #= the current design decision is to fix the ordering to be default_ordering(R) =#
  #= that is, the user is not allowed to change this with the outer constructor quo =#
```
So your impression is correct. There are (or were) design decisions denying the user influence on the ordering.

HechtiDerLachs:
Indeed it seems that my changes to `keep_ordering=true` are causing the trouble: It hangs at a Groebner-basis computation for `divides` in quotient rings for the `SpecOpen` tests. Switching back to `keep_ordering=false` allows the tests to pass within seconds. But for that setting the changes in this PR become useless.

afkafkafk13:
Let us keep this in mind for the coding sprint on quotient rings -- unless you want to avoid the quotient rings completely in this approach and choose to simulate the quotient ring computations in the polynomial ring, which gives you full control of the orderings, but also creates a lot more logistics to keep track of.

HechtiDerLachs:
I'd prefer to have a quick fix for this now. The changes in this PR made the difference of coming through with a computation of a glueing vs. getting stuck while using several Terabyte of RAM. We need this for the Oscar book chapter and the coding sprint has not even been scheduled, yet.

simonbrandhorst (approved):

simonbrandhorst:
Since this does not seems to hurt anyone else and speeds up my computaitons, I merge it.

PR #2520, opened 2023-07-05
updates of the test code concerning new Oscar functionality

ThomasBreuer:
- construction and visualization of graphs
- functions applicable to the fan object

### DIFF
--- a/experimental/GITFans/test/runtests.jl
+++ b/experimental/GITFans/test/runtests.jl
@@ -20,7 +20,7 @@
     Qt, T = polynomial_ring(QQ, :T => 1:n)
     D = free_abelian_group(ncols(Q))
     w = [D(Q[i, :]) for i = 1:n]
-    R = grade(Qt, w)
+    R, T = grade(Qt, w)
     a = ideal([
         T[5]*T[10] - T[6]*T[9] + T[7]*T[8],
         T[1]*T[9]  - T[2]*T[7] + T[4]*T[5],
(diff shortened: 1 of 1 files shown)


experimental/GITFans/test/runtests.jl
```diff
@@ -52,7 +52,10 @@
            BitSet([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
            BitSet([1])]
 
-    intergraph = Polymake.graph.graph_from_edges(collect(edges));
+    orbit_graph = Graph{Undirected}(length(hash_list));
+    [add_edge!(orbit_graph, e...) for e in edges];
```
fingolfin, line 56: Lots of trailing semicolons here and elsewhere - remove?

ThomasBreuer: I am using parts of these tests for copy & paste in a Julia session, and the output of intermediate steps is uninteresting. In particular the commands mentioned here produce a full screen of useless output.
What is the problem with suppressing output?
I would argue that output should not be shown except if there is a reason to show it.

jankoboehm (approved):

experimental/GITFans/test/runtests.jl
thofma: We do not add `;` to every line of all files, just in case someone wants to copy them to the REPL.

ThomasBreuer: I am not thinking of source code.
Test files have the flavour of Julia sessions, and some of them are interesting enough to be used in a Julia session, similar to Jupyter notebooks.

thofma: Hm, I don't agree on this point. I can see that doctests have some REPL flavour, but not test files. Some of them require  some setup (which might not even be part of that file), and often the calls are "non-local" (e.g. require variables defined a hundered lines before).

Anyway, this should not hold up this PR. We are inconsistent anyway.

ThomasBreuer: What is the difference in processing a test file with or without a semicolon at the end of a command? Is perhaps the `show` method of the return value called in the case that there is no semicolon, and thus one tests also whether this method does not run into an error? That would be a reason to force that there is no semicolon.

fingolfin (approved):

PR #2521, opened 2023-07-06
Literature model improvements

### LINKED ISSUE #2506: F-theory tools - ArxivDOI vs. Journal DOI
We should address the following points:

1. The look-up of literature models currently only checks if a provided DOI matches the journal DOI. We should also check if it matches an arxiv DOI.
2. The model files, include model data such as equation, section, etc. This information may depend on whether or not one looks at the journal or the arxiv version. So this information should be doubled and integrated into the arxiv and journal information section of the json file.

cc @apturner

apturner:
1. Adds Weierstrass model functionality to literature models
2. Renames the LiteratureTateModels directory to LiteratureModels
3. Fixes the model location data to be specific to the arXiv and published version of the paper (addressing #2506)
4. Pulls all functions that are directly relevant to literature models in general out of the Tate model files and into corresponding files for literature models
5. Renames model files by arXiv identifier for ease of identification
6. Adds many attributes/properties to allow for decorating models with all data from the database


HereAround (approved):
Thank you very much indeed @Apturner. This looks great!

PR #2523, opened 2023-07-06
Fix some fallout of fixing piracy in Hecke

lgoettgens:
Fixes the errors now occurring in Oscar after https://github.com/Nemocas/Nemo.jl/pull/1501 and https://github.com/thofma/Hecke.jl/pull/1143.


fingolfin (approved):

fingolfin:
So this is now merging into a target branch `bump`, is that intentional?

thofma:
Yes, we collect all necessary commits in that branch.

PR #2529, opened 2023-07-06
Refinements for separation of ideal sheaves with disjoint supports.

HechtiDerLachs:
As requested by @simonbrandhorst .

This is still WIP for the time being, but ready to be tried.

### DIFF
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -938,3 +938,49 @@ function _cofactors(comp::Vector{<:Ideal})
   result = [prod(pairwise_cof[1:n, i]) for i in 1:n]
   return result
 end
+
+function _one_patch_per_component(covering::Covering, comp::Vector{<:IdealSheaf})
+  new_patches2 = Vector{AbsSpec}()
+  patches_todo = copy(patches(covering))
+  for P in comp
+    # Find one patch in which this component is supported
+    i = findfirst(U->!isone(P(U)), patches_todo)
+    U = patches_todo[i]
+    # Take this patch out of the list
+    deleteat!(patches_todo, i)
+    # Add it to the list of patches for the new covering
+    push!(new_patches2, U)
+    # For every other patch V in which P appears we do the following:
+    # Replace V by the complement of the support of P.
+    # This will not be affine in general, but can be covered by hypersurface 
+    # complements. Even though this may lead to many charts, they will be harmless
+    # in the mext blowup.
+    done = Int[]
+    for (j, V) in enumerate(patches_todo)
+      # Check whether P is visible in this patch; if not leave it
+      isone(P(V)) && continue
+      # Remember this patch to be done 
+      push!(done, j)
+      sg = small_generating_set(P(V))
+      new_patches2 = append!(new_patches2, [PrincipalOpenSubset(V,a) for a in sg])
+      # TODO: Cache that P = 1 on all these new patches?
+    end
+    deleteat!(patches_todo, done)
+  end
+  new_patches2 = append!(new_patches2, patches_todo)
+  new_cov = Covering(new_patches2)
+  inherit_glueings!(new_cov, covering)
+  return new_cov
+end
+
+@attr Vector{<:MPolyQuoLocRingElem} function small_generating_set(I::MPolyQuoLocalizedIdeal)
+  L = base_ring(I)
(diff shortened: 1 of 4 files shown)


simonbrandhorst:
This is not quite what I need.

simonbrandhorst:
Here come the precise specification:

Given a covering $C$ and a list of 0-dimensional prime ideal sheaves $L$ 
Return a covering $D$ refining $C$ such that 
For each $U \in D$ there is at most one $P \in L$ such that $P \in U$.
[edit: I messed it up sorry.  What I need is: for each $P \in L$ there is exactly one $U \in D$ with $P\in U$.

Example 
$\mathbb{P}^2$ with the standard covering $C = [U, V, W]$ and $L = [P]$ where $P$ is the ideal sheaf describing $(1:1:0)$. Then $P \in U$ and $P \in V$, $P \notin W$.
A possible return value is $D = [U , V \setminus {P}, W]$ 
One might want to work with affine schemes and so prefer to take principal opens. Then
[U, V\setminus\{x_0/x_1=1\} , W] should do the job.

simonbrandhorst:
```
julia> P = projective_space(QQ,1)
Projective space of dimension 1
  with homogeneous coordinates s0 s1 
  over Rational field

julia> X = covered_scheme(P)
covered scheme with 2 affine patches in its default covering

julia> t = coordinates(X[1][1])[1]
t

julia> I = ideal_sheaf(X,X[1][1],[t-1])
Ideal Sheaf on Covered Scheme with 2 Charts


julia> C = Oscar._separate_disjoint_components([I])
Covering with 2 patches

julia> [I(U) for U in C]
2-element Vector{MPolyIdeal{QQMPolyRingElem}}:
 ideal(t - 1)
 ideal(s - 1)
```

What I want is that all but one of these ideals is $(1)$

experimental/Schemes/IdealSheaves.jl
```diff
@@ -891,3 +891,50 @@ function show_details(io::IO, I::IdealSheaf)
     println(io," ")
   end
 end
+
+function _separate_disjoint_components(comp::Vector{<:IdealSheaf}; covering::Covering=default_covering(scheme(first(comp))))
+  isempty(comp) && error("list of components must not be empty")
+  X = scheme(first(comp))
+  all(x->scheme(x) === X, comp) || error("components must be defined over the same scheme")
+  isone(length(comp)) && return covering
```
simonbrandhorst, line 899: this line is offending

simonbrandhorst: What you seem to do is to assure that each patch contains at most one singular point.
This is good. But I want the extra condition that each singular point is visible in EXACTLY one patch. 
So we need to add a second refinement step.

afkafkafk13: May I ask again about the precise task:
You have an IdealSheaf of a 0-dimensional subscheme of some covered scheme X and want to find a covering of X such that each point in the support of your 0-dim. subscheme appears in precisely one chart. Do you want to still cover all of X, possibly allowing charts with no points? Are we still in the surface case or do you need it in more general context?

simonbrandhorst: should still cover the scheme. Need it to resolve a surface.
But if the same point appears in two charts, then each blowup is performed twice.
This leads to unnecessarily many charts.

simonbrandhorst:
adding the following should work (not tested thoroughly)
```
  new_patches2 = Vector{AbsSpec}()
  patches_todo = copy(patches(covering))
  for P in comp
    i = findfirst(U->!isone(P(U)), patches_todo)
    U = patches_todo[i]
    deleteat!(patches_todo, i)
    push!(new_patches2, U)
    # remove P from all other patches
    done = Int[]
    for (j,V) in enumerate(patches_todo)
      if isone(P(V))
        continue
      end
      push!(done, j)
      sg = small_generating_set(P(V))
      new_patches2 = append!(new_patches2, [PrincipalOpenSubset(V,a) for a in sg])
    end
    deleteat!(patches_todo, done)
  end
  new_patches2 = append!(new_patches2, patches_todo)
  new_cov = Covering(new_patches2)
  inherit_glueings!(new_cov, covering)
  return new_cov
```

simonbrandhorst:
Thanks. Should do the job for now. We can still improve later.

simonbrandhorst (approved):

PR #2533, opened 2023-07-07
Update julia versions in CI

lgoettgens:
Now that julia `master` is `1.11.0-DEV` and the first `1.10` alpha is released, we should reflect that in our CI.

Open question:
Do we still test 1.8, or only 1.6, 1.9, 1.10, and nightly (i.e. just "add" 1.10 or keep the total number the same)?
If we want to adapt the default from 1.8 to 1.9, this has to be changed in multiple places in the workflow files as well.

Once this PR is approved, I can make similar PRs for the other repos.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,7 +25,8 @@ jobs:
         julia-version:
           - '1.6'
           - '1.8'
-          - '~1.9.0-0'
+          - '1.9'
+          - '~1.10.0-0'
           - 'nightly'
         os:
           - ubuntu-latest
@@ -75,7 +76,8 @@ jobs:
       matrix:
         julia-version:
           - '1.8'
-          - '~1.9.0-0'
+          - '1.9'
+          - '~1.10.0-0'
           - 'nightly'
         os:
           - ubuntu-latest
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,7 +24,6 @@ jobs:
       matrix:
         julia-version:
           - '1.6'
-          - '1.8'
           - '1.9'
           - '~1.10.0-0'
           - 'nightly'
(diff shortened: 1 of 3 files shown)


lgoettgens:
And once this is merged, somebody with admin rights should update the required CI jobs to pass for a merge.

thofma:
To keep CI times reasonable, I would suggest we drop something, maybe 1.8? Or we drop 1.9 as soon as 1.10 is out? I have no preference, but I can imagine dropping 1.8 is more work, since we changed a few things recently to use 1.8.

fingolfin:
Normally I would say that we should test the oldest supported Julia (version 1.6; I expect we might switch it to the next LTS as soon as one is selected), the latest stable release (version 1, right now 1.9.x), nightly. Ideally we also start testing pre-releases of the next stable as soon as they become available (which this PR is about). 

So that would suggest dropping 1.8 -- except that would also have repercussions on the OscarCI tests, so we might want to keep it for those for now (CC @benlorenz)

benlorenz:
We can easily change the default for OscarCI to 1.9 as well, for the Oscar tests + doctests this should be fine.

I don't really know whether the doctests for all other packages work with 1.9 as well. So far for Singular, Hecke, AbstractAlgebra and Nemo the doctests are only run with 1.8.


I also think dropping 1.8 and testing 1.6 + 1.9 + 1.10-dev + nightly would be reasonable.

lgoettgens:
Alright, I will change this PR tomorrow to drop 1.8, and then provide similar PRs to OscarCI and the other packages.

lgoettgens:
> I also think dropping 1.8 and testing 1.6 + 1.9 + 1.10-dev + nightly would be reasonable.

Now included in this PR.

> We can easily change the default for OscarCI to 1.9 as well, for the Oscar tests + doctests this should be fine.

See https://github.com/oscar-system/OscarDevTools.jl/pull/31.


> I don't really know whether the doctests for all other packages work with 1.9 as well. So far for Singular, Hecke, AbstractAlgebra and Nemo the doctests are only run with 1.8.

I tested it locally and the doctests succeeded in julia 1.9.2.

lgoettgens:
All packages (apart from Polymake which already tests using 1.10-DEV) now have a PR changing the default julia version to use for everything from 1.8 to 1.9, and the additional test version ~1.9.0-0 to ~1.10.0-0.

lgoettgens:
The failure in 1.10 is due to a timing test. This will be tracked in https://github.com/oscar-system/Oscar.jl/issues/2539.

fingolfin (approved):

PR #2535, opened 2023-07-08
Some improvement on default ordering

jankoboehm:
This is a further step to improve the lookup of Gröbner bases via a MonomialOrdering key in issue #2498 and #2099. The default ordering is now cached in the ring. Using the modified test file from issue #2498 we get now from

```julia
 1.573602 seconds (8.53 M allocations: 349.677 MiB, 20.44% gc time)
```

to

```julia
 0.900222 seconds (5.71 M allocations: 262.352 MiB, 18.01% gc time)
```

Caching in addition the hash and is_global in the ordering did not give any further improvement, but rather made it slightly slower and increased allocations and memory usage.

### DIFF
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -161,7 +161,16 @@ using .Orderings
 #type for orderings, use this...
 #in general: all algos here needs revision: do they benefit from gb or not?
 
-default_ordering(R::MPolyRing) = degrevlex(R)
+function default_ordering(R::MPolyRing)
+  if has_attribute(R, :default_ordering)
+    return get_attribute(R, :default_ordering)
+  else
+    default_ordering = degrevlex(R)
+    set_attribute!(R, :default_ordering => default_ordering)
+    return default_ordering
+  end
+end
+
 
 mutable struct BiPolyArray{S}
   Ox::NCRing #Oscar Poly Ring or Algebra
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -162,12 +162,8 @@ using .Orderings
 #in general: all algos here needs revision: do they benefit from gb or not?
 
 function default_ordering(R::MPolyRing)
-  if has_attribute(R, :default_ordering)
-    return get_attribute(R, :default_ordering)
-  else
-    default_ordering = degrevlex(R)
-    set_attribute!(R, :default_ordering => default_ordering)
-    return default_ordering
+  return get_attribute!(R, :default_ordering) do
+    degrevlex(R)
   end
 end
 
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -745,11 +745,10 @@ function has_weighted_ordering(R::MPolyDecRing)
 end
 
 function default_ordering(R::MPolyDecRing)
-  fl, w_ord = has_weighted_ordering(R)
-  if fl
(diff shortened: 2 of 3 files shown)


afkafkafk13 (approved):
Indeed, this change makes a lot of sense.

afkafkafk13:
Are there any methods implicitly relying on default_ordering being degrevlex? Would this kind of problem be detected by the tests?

jankoboehm:
The idea is that any global ordering will do for a polynomial ring wrt mathematical correctness and is chosen according to best performance. If the default ordering would be changed (which we don't want) some tests would probably fail (remainders). In the graded case (which should also be addressed wrt caching), when considering non-standard Z-grading, a weight ordering is chosen.

afkafkafk13:
I am aware that you do not want to change the default_ordering, but only avoid the look-ups. And this is what I approved to.

The following remark is not blocking a merge, just a pointer:
If a sophisticated user decides to set the default ordering themself, a warning in the documentation that the default_ordering is assumed to be global,... (and other choices may raise havoc) may be necessary or even a secure method to set the default ordering which performs some sanity checks once and for all.

jankoboehm:
The docu has a remark on the default ordering, which presents the default ordering as something fixed which cannot be changed. So there is probably little risk that someone changes it, especially since it is hidden in an attribute. I think it is a good idea to provide a function to set the default ordering including docu and safeguards. Especially in the graded case, the user might sometimes want to use a more sophisticated ordering in case the ring is Z^n graded (matrix ordering or cleverly chosen weight ordering). Currently one would get dp, which is reasonable, but not always the best.

src/Rings/mpoly.jl
```diff
@@ -161,7 +161,16 @@ using .Orderings
...
+function default_ordering(R::MPolyRing)
+  if has_attribute(R, :default_ordering)
+    return get_attribute(R, :default_ordering)
+  else
+    default_ordering = degrevlex(R)
+    set_attribute!(R, :default_ordering => default_ordering)
+    return default_ordering
+  end
+end
+
```
lgoettgens:
```suggestion
function default_ordering(R::MPolyRing)
  return get_attribute!(R, :default_ordering) do
    degrevlex(R)
  end
end
```

I would suggest to simplify it as above, as it is more readable and should be a hinge faster, as it avoids some attribute lookup. See https://nemocas.github.io/AbstractAlgebra.jl/dev/misc/#AbstractAlgebra.get_attribute!

afkafkafk13:
still draft?

jankoboehm:
I think new functionality to manage the default ordering should go into a separate effort, so we can go ahead.

src/Rings/mpoly.jl
```diff
@@ -161,7 +161,12 @@ using .Orderings
 #type for orderings, use this...
 #in general: all algos here needs revision: do they benefit from gb or not?
 
-default_ordering(R::MPolyRing) = degrevlex(R)
+function default_ordering(R::MPolyRing)
+  return get_attribute!(R, :default_ordering) do
+    degrevlex(R)
+  end
+end
```
fingolfin, line 168: Just a note, this would have been a typical usecase for the `@attr` macro -- or rather `@attr RETURN_TYPE` to make it type stable (this code now is not type stable for this reason)

jankoboehm: Could you give an example how we should do that alterative?

PR #2544, opened 2023-07-11
Introduce rational maps of covered schemes.

HechtiDerLachs:
Still work in progress and poor in documentation, but put here as a preview. CC: @simonbrandhorst

### DIFF
--- a/test/AlgebraicGeometry/Schemes/RationalMap.jl
+++ b/test/AlgebraicGeometry/Schemes/RationalMap.jl
@@ -22,3 +22,26 @@
   oscar.realize_on_patch(Phi, X[1][3])
   oscar.realize(Phi)
 end
+
+@testset "The standard Cremona transformation" begin
+  IP2_proj = projective_space(QQ, [:x, :y, :z])
+  IP2 = covered_scheme(IP2_proj)
+  S = homogeneous_coordinate_ring(IP2_proj)
+  (x, y, z) = gens(S)
+  I1 = ideal(S, [x, y])
+  I2 = ideal(S, [x, z])
+  I3 = ideal(S, [y, z])
+  I = I1*I2*I3
+  II = simplify!(IdealSheaf(IP2_proj, I))
+  pr = blow_up(II)
+  X = domain(projection(pr))
+  set_attribute!(X, :is_irreducible, true)
+  V = first(affine_charts(IP2))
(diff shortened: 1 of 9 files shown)


test/AlgebraicGeometry/Schemes/RationalMap.jl
```diff
@@ -0,0 +1,24 @@
...
+  I = ideal(S, x*w - y*z)
+  X_proj = subscheme(IP3, I)
+  X = covered_scheme(X_proj)
+
+  P1 = projective_space(QQ, 1)
+  IP1 = covered_scheme(P1)
+  U = X[1][4]
+  V = IP1[1][2]
+  (x, y, z) = gens(ambient_coordinate_ring(U))
+  Phi = oscar.RationalMap(X, IP1, U, V, [x//y])
```
HechtiDerLachs, line 14: You create a rational map by specifying it on two `affine_charts` on the domain and codomain in terms of fractions.

test/AlgebraicGeometry/Schemes/RationalMap.jl
```diff
@@ -0,0 +1,24 @@
...
+
+  @test domain(Phi) === X 
+  @test codomain(Phi) === IP1
+
+  oscar.realize_on_patch(Phi, U)
```
HechtiDerLachs, line 19: Then you can ask to realize it in a refinement on each patch. This will give you a list of actual morphisms on `PrincipalOpenSubset`s of `U`. Note that these may have different codomains!

test/AlgebraicGeometry/Schemes/RationalMap.jl
```diff
@@ -0,0 +1,24 @@
...
+  oscar.realize_on_patch(Phi, X[1][1])
+  oscar.realize_on_patch(Phi, X[1][2])
+  oscar.realize_on_patch(Phi, X[1][3])
+  oscar.realize(Phi)
```
HechtiDerLachs, line 23: The type for rational maps is lazy. You may, however, as for a full realization of the morphism as a morphism of covered schemes. In this sense, the name "RationalMap" is a bit misleading, because it is assumed to extend to a full morphism here. We should hence find a better name.

simonbrandhorst:
Thanks. I will play with it.

simonbrandhorst:
Is this one still alive?

PR #2545, opened 2023-07-12
Minimal_generating_set/small_generating_set -- local cases

afkafkafk13:
PR contains:
- minimal_generating_set for localizations at k-points
- small_generating_set for localizations at k-points and localizations at powers of elements

Standard Basis for shifted_ideal w.r.t. local ordering is stored in dictionary as in the global case.

Obsoletes PR #2327
Resolves Issue #2295


afkafkafk13:
@fingolfin : What am I doing wrong causing some tests to not have started?

benlorenz:
Close and reopen will restart CI (with a fresh merge-commit with master).

The required status checks have been updated for the changes from #2533.

src/Rings/mpoly-localizations.jl
```diff
@@ -3040,3 +3040,128 @@ end
...
+# Minimal generating set
+#
+################################################################################
+
+@doc raw"""
+    minimal_generating_set(I::MPolyLocalizedIdeal)
+
+Given an ideal `I` in the localization of a multivariate polynomial ring
+over a field at a point, return an array containing a minimal set of
+generators of `I`. If `I` is the zero ideal an empty list is returned.
```
simonbrandhorst, line 3054: Minimal with respect to inclusion or cardinality?

src/Rings/mpoly-localizations.jl
```diff
@@ -3040,3 +3040,128 @@ end
...
+  L = base_ring(I)
+  R = base_ring(L)
+
+  @req coefficient_ring(R) isa AbstractAlgebra.Field "The coefficient ring must be a field"
+
+  shift, back_shift = base_ring_shifts(L)
+  I_shift = shifted_ideal(I)
+  o = negdegrevlex(R) # the default local ordering
+
+  if !isempty(I_shift.gb)
```
simonbrandhorst, line 3093: Do we have an interface for this? If we have, would you please use it?
Or we could introduce one? (in a different pr) like `has_groebner_basis`?

src/Rings/mpoly-localizations.jl
```diff
@@ -3040,3 +3040,128 @@ end
...
+
+    # use available gb for shifted ideal is available
+    G = first(values(I_shift.gb))
```
simonbrandhorst, line 3096: Similar here.

src/Rings/mpoly-localizations.jl
```diff
@@ -3040,3 +3040,128 @@ end
...
+    is_local(G.ord) || error("inconsistent data: local ring, but global ordering for I_shift")
+    singular_assure(G, G.ord)
```
simonbrandhorst, line 3098: and here

src/Rings/mpoly-localizations.jl
```diff
@@ -3040,3 +3040,128 @@ end
...
+    G.gens.S.isGB = true
+    _, I_shift_min = Singular.mstd(G.gens.S)
```
simonbrandhorst, line 3100: and here

simonbrandhorst (approved):
The interface looks good to me. Can't say much about the math involved. But I assume you know what you do.

src/Rings/mpoly-localizations.jl
afkafkafk13: This is a fixed technical term. It is minimal as in minimal presentation of R/I (exactly like in the graded case), which in turn implies minimal w.r.t. cardinality.

src/Rings/mpoly-localizations.jl
afkafkafk13: No, we do not have an interface for this in the local case. And the function in question is located at a very low level in the hierarchy of code so that I see it as no problem here.

At some point in the future, we need something to properly interact with the standard basis of the shifted ideal in the localized case. But this reaches into the localized ring code and is not intended as content of this PR (parts of which could however serve as blueprint for this).

afkafkafk13: P.S: This is a comment also to the whole sequence of commands to which you commented below.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: I would probably add a `note` saying that minimality is strict only in the case of localization at a rational point.

src/Rings/mpoly-localizations.jl
```diff
@@ -3040,3 +3040,128 @@ end
...
+@attr Vector{<:MPolyLocRingElem} function minimal_generating_set(
+    I::MPolyLocalizedIdeal{<:MPolyLocRing{<:Field, <:FieldElem,
+                                          <:MPolyRing, <:MPolyElem,
+                                          <:MPolyComplementOfKPointIdeal}
+                          }
+  )
+  L = base_ring(I)
+  R = base_ring(L)
+
+  @req coefficient_ring(R) isa AbstractAlgebra.Field "The coefficient ring must be a field"
```
HechtiDerLachs, line 3087: This is already asserted by the signature of the method.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: I agree that such an interface would be nice to have. But there is none. This code is just poking in the internals and might not even be safe. But for the moment, I'm afraid, we can't do better. It's a bigger task that we have to address here in the beginning of August.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,60 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+function minimal_generating_set(
+    I::MPolyQuoLocalizedIdeal{<:MPolyQuoLocRing{<:Field, <:FieldElem,
+                                          <:MPolyRing, <:MPolyElem,
+                                          <:MPolyComplementOfKPointIdeal},
+                              <:Any,<:Any}
+  )
+  Q = base_ring(I)
+  L = localized_ring(Q)
+
+  J = ideal(L, lift.(gens(I))) + modulus(Q)
```
HechtiDerLachs: Shouldn't there also be a shift here? This method does not seem to do anything useful for now. Is this code still due to something I did?

test/Rings/mpolyquo-localizations.jl
```diff
@@ -296,3 +296,20 @@ end
...
+  Q1 = MPolyQuoLocRing(R,IQ,U1)
+  Q2 = MPolyQuoLocRing(R,IQ,U2)
+  Q3 = MPolyQuoLocRing(R,IQ,U3)
+  J1 = ideal(Q1,[x^2-y^2,y^2-z^2,x^2-z^2])
+  @test length(minimal_generating_set(J1)) == 1
+  @test length(small_generating_set(J1)) == 1
+  J2 = ideal(Q2,[x^2-y^2,y^2-z^2,x^2-z^2])
+  @test length(small_generating_set(J2)) == 1
+  J3 = ideal(Q3,[x^2-z^2,y*(y-1)])
+  @test length(small_generating_set(J3)) == 1
```
HechtiDerLachs: I'm afraid, these tests are still too trivial. Since there is no shifting visible in the tested methods, I expect these tests to fail for ideals with more generators.

HechtiDerLachs (commented):
Maybe I'm missing some things, but this doesn't look fully functional to me for now. 

For the `MPolyQuoLocalizedIdeal`s I propose to use the `oscar.pre_image_ideal` (sorry about the name of this method) to link this to the `MPolyLocalizedIdeal`s which have already been covered.

src/Rings/mpoly-localizations.jl
afkafkafk13: So we can be sure that no other fields than AbstractAlgebra.Fields are floating around somewhere? I was not sure about that.

src/Rings/mpolyquo-localizations.jl
afkafkafk13: This is pulling the ideal in the localized quotient ring back to the localized ring, where we fall back to the method there. The shift is going on in the ring L not in Q -- to keep the ugly code at one specific place and not pour it over many files.

src/Rings/mpoly-localizations.jl
afkafkafk13: It is also strict for localizations at prime ideal (seen als localized ring over a field extension) -- but we do not have functionality for this kind of ring in the backend. So I put a note there stating that it is only available for rational points.

afkafkafk13:
> Maybe I'm missing some things, but this doesn't look fully functional to me for now.
> 
> For the `MPolyQuoLocalizedIdeal`s I propose to use the `oscar.pre_image_ideal` (sorry about the name of this method) to link this to the `MPolyLocalizedIdeal`s which have already been covered.

Thanks for the hint. I will change it accordingly.

Why is this command neither documented nor exported? I tried to use as few non-exported methods as possible.

test/Rings/mpolyquo-localizations.jl
afkafkafk13: You are right. More is in the next PR.

src/Rings/mpoly-localizations.jl
afkafkafk13: I put a guard in to make sure that a non-local ordering will not be used -- this could however throw an error. 
But I did not want to go through the whole list I_shift.gb to find a local one. I_shift is cached internal data for an ideal in a localized ring after all.

src/Rings/mpolyquo-localizations.jl
afkafkafk13: I replaced it by pre_image_ideal. Thanks for the pointer.

afkafkafk13:
> Maybe I'm missing some things, but this doesn't look fully functional to me for now.

I tested it (beyond the tests in test/Rings/ directory) and it is fully functional for the localizations at a rational point and parallels the graded code. However, I use knowledge on the internal workings of mstd and only provide an interface for that.

HechtiDerLachs:
> Why is this command neither documented nor exported? I tried to use as few non-exported methods as possible.

It is supposed to be an internal method. I don't think the average user is interested in this. But it should be considered to be legitimate to use it internally for such purposes as here. At least that's what I wrote it for. 

But: For quotient rings there is some magic happening with `singular_assure` so that you can get a quotient ring on the singular side directly. This affects the outcome of `minimal_generating_set`: A minimal generating set of the `pre_image_ideal` might not be minimal modulo the `modulus` of the ring. So I'm afraid we have to be more careful here and do go via shiftings etc. again. The method for `MPolyQuoRing`s should provide us with samples how to bind to the singular backend correctly, then.

afkafkafk13:
> 

I see your point and will take care about it.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,62 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+function small_generating_set(
+      I::MPolyQuoLocalizedIdeal{<:MPolyLocRing{<:Field, <:FieldElem,
+                                          <:MPolyRing, <:MPolyElem,
+                                          <:MPolyComplementOfKPointIdeal},
+                              <:Any,<:Any}
+  )
+  Q = base_ring(I)
+  L = localized_ring(Q)
+
+  J = ideal(L, lift.(gens(I))) + modulus(Q)
```
HechtiDerLachs:
```suggestion
  J = pre_image_ideal(I)
```

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,62 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+function small_generating_set(
+    I::MPolyQuoLocalizedIdeal{<:MPolyLocRing{<:Field, <:FieldElem,
+                                          <:MPolyRing, <:MPolyElem,
+                                          <:MPolyPowersOfElement}
+                          }
+  )
+  Q = base_ring(I)
+  L = localized_ring(Q)
+
+  J = ideal(L, lift.(gens(I))) + modulus(Q)
```
HechtiDerLachs:
```suggestion
  J = pre_image_ideal(I)
```

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,62 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+                              <:Any,<:Any}
+  )
+  Q = base_ring(I)
+  L = localized_ring(Q)
+
+  J = ideal(L, lift.(gens(I))) + modulus(Q)
+  return filter(!iszero, Q.(small_generating_set(J)))
+end
+
+function small_generating_set(
```
HechtiDerLachs, line 2235:
```suggestion
@attr function small_generating_set(
```
Maybe we should turn all these into attributes?

HechtiDerLachs (approved):
Sorry, I think, I misread something earlier. This should all be rather fine.

HechtiDerLachs:
Sorry, but I have to revoke my approval. It seemed to be working, but looking at the code, I couldn't see why, unless this was a coincidence. And indeed I found the following counterexample:
```
R, (x, y) = QQ[:x, :y]
I = ideal(R, x^6-y)
U = complement_of_point_ideal(R, [1, 1])
L = MPolyQuoLocRing(R, I, U)
J = ideal(L, [x-1, y-1])^2
oscar.minimal_generating_set(J)
```
The modulus suggests that the ring is effectively a polynomial ring in one variable `x`. Then the minimal generating set of the ideal `J` should consist of one generator, but my output is
```
julia> oscar.minimal_generating_set(J)
2-element Vector{MPolyQuoLocRingElem{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem, MPolyComplementOfKPointIdeal{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem}}}:
 y^2 - 2*y + 1
 6*x - y - 5
```

afkafkafk13:
As I already wrote before your approval and the taking back thereof: 
I saw your point and I am working on a solution. 

Unfortunately, it turned out that the difficulties lie in functionality which is available for MPolyLocRing, but not for MPolyLocQuoRing (for good reasons). I know how to amend for that, but  it will take at least until monday.

src/Rings/mpolyquo-localizations.jl
lgoettgens: Please add the type of the cached object after `@attr` to ensure type safety.

afkafkafk13:
Thanks @HechtiDerLachs for pointing out the problem. I now do the locquo-case 'by hand' as there is not enough underlying functionality available. Please have look.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,92 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+Given an ideal `I` in the localization of a quotient of a  multivariate
+polynomial ring over a field at a point, return an array containing a
+minimal set of generators of `I`. If `I` is the zero ideal an empty list
+is returned.
+
+Note: This is only available for localizations at rational points. 
+"""
+@attr Vector{<:MPolyQuoLocRingElem}function minimal_generating_set(
+    I::MPolyQuoLocalizedIdeal{<:MPolyQuoLocRing{<:Field, <:Any,
+                                          <:Any, <:Any,
```
HechtiDerLachs, line 2170: Out of curiosity? Did the type assertions here cause any trouble? Otherwise: Why not keep them?

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,92 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+    minimal_generating_set(I::MPolyLocalizedIdeal)
+
+Given an ideal `I` in the localization of a quotient of a  multivariate
+polynomial ring over a field at a point, return an array containing a
+minimal set of generators of `I`. If `I` is the zero ideal an empty list
+is returned.
+
+Note: This is only available for localizations at rational points. 
+"""
+@attr Vector{<:MPolyQuoLocRingElem}function minimal_generating_set(
```
HechtiDerLachs:
```suggestion
@attr Vector{<:MPolyQuoLocRingElem} function minimal_generating_set(
```

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2155,4 +2155,92 @@ function inverse(phi::MPolyQuoLocalizedRingHom)
...
+    if !(gen(F,i) in leading_module(M,oF))
+      append!(res_vec,[Q.(gen(I,i))])
+    end
+  end
+
+  return res_vec
+end
+
+@doc raw"""
+    small_generating_set(I::MPolyLocalizedIdeal)
```
HechtiDerLachs: This docstring is misplaced. Also, the function below should probably just be a synonym for `minimal_generating_set`.

HechtiDerLachs (approved):
This looks good. It's hard to really understand just from reading what the singular code is doing, but if the tests work out, it should be fine.

src/Rings/mpolyquo-localizations.jl
afkafkafk13: done.

src/Rings/mpolyquo-localizations.jl
afkafkafk13: done

src/Rings/mpolyquo-localizations.jl
afkafkafk13: Yes and no, it did cause trouble during the my many tests and debugging runs because of my incompetence to always feed suitable objects to it. The two imposed conditions suffice for us, so I left out the others.

src/Rings/mpolyquo-localizations.jl
afkafkafk13: done

src/Rings/mpolyquo-localizations.jl
afkafkafk13: No, the doctring has a Copy&Paste Error. The next commit amends this.
And no, this should not be an alias, because the other variant (which does it on foot)  is more clumsy and this one often suffices, if you do not require minimality, but only something reasonably small.

PR #2547, opened 2023-07-12
Refactor parts of toric geometry

lkastner:
- `fan` renamed to `polyhedral_fan`
- Eliminate hack in `polyhedral_fan` that was used at some point since casting did not work
- The `polyhedral_fan` is also not stored as an attribute anymore, as this only duplicates information and the toric variety itself already provides the full fan functionality.
- Disable printing of `rays` of a fan in toric schemes part. We will never want that kind of detail in the printout except if someone calls rays directly. Also in case of a fan this information does not even describe the fan completely.
- Enable `--depwarn=error` for the doctests
- Fix doctests

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -117,7 +117,7 @@ jobs:
             Pkg.instantiate()'
       - name: "Run doctests"
         run: |
-          julia --project=docs --color=yes -e '
+          julia --project=docs --depwarn=error --color=yes -e'
             using Documenter
             using Oscar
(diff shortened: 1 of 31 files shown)


HereAround (approved):
Thank you @lkastner  for those changes. In retrospect, I agree that we should not print the rays for the toric schemes. Sadly, this is the exact part that requires rebase/adjusting after I merged https://github.com/oscar-system/Oscar.jl/pull/2470. Sorry about that.

Other than that - thank you very much. Looks great!

HereAround:
I am actually surprised by `julia --project=docs --depwarn=error --color=yes -e'`. I recall making that change a while back. But maybe I only switched it on for the "real" tests and not the doctests? It is definitely good to have that.

benlorenz (approved):

PR #2551, opened 2023-07-13
Rename invalidations CI test

fingolfin:
In some parts of the GitHub UI, only 'evaluate' is showing, which
is not helpful. A more meaningful name alleviates this problem.


lgoettgens:
Should this be changed in the other repos as well?

fingolfin:
sure

Issue #2555, opened 2023-07-13
Two different skew-symmetry checks


lgoettgens:
**Describe the bug**
We have two different implementations for skew symmetry checks of a `MatElem`: 
```julia
julia> methods(is_skew_symmetric)
# 1 method for generic function "is_skew_symmetric" from AbstractAlgebra:
 [1] is_skew_symmetric(M::MatElem)
     @ ~/path/AbstractAlgebra.jl/src/Matrix.jl:2403

julia> methods(is_skewsymmetric_matrix)
# 1 method for generic function "is_skewsymmetric_matrix" from Oscar:
 [1] is_skewsymmetric_matrix(B::MatElem{T}) where T<:RingElem
     @ ~/path/Oscar.jl/src/Groups/matrices/matrix_manipulation.jl:171
```

As far as I can tell, the implementations differ in the behavior in characteristic 2, and the latter imposes a restriction on the element type of the matrices.

It would be great to have only one function doing that and not two. Similar functions would be named `is_skew_symmetric` (without `matrix`), so I would like that name here.

Can somebody with knowledge about the use-cases say a few words about the characteristic 2 stuff, so that we can decide, which version to keep? In all cases, this should probably go into AbstractAlgebra.

fingolfin:
We are comparing 

https://github.com/oscar-system/Oscar.jl/blob/3b64d46d9ab90b1cd6e9b503fa1b7adf6667dbfa/src/Groups/matrices/matrix_manipulation.jl#L171-L183
 https://github.com/Nemocas/AbstractAlgebra.jl/blob/bfa185a7c074b20527009e039ad45be8ab1bdd7a/src/Matrix.jl#L2403-L2410

I would say the latter (in AbstractAlgebra) is what we want. The former (in Oscar) actually tests for an *alternating* form (resp. a representation matrix for such a form) -- we may also want such a test, but it should be under its own name. And indeed the restriction on the typ seems unnecessary.

Issue #2543, opened 2023-07-11
nightly + 1.10-alpha1: GC marking segfault during GAP tests


benlorenz:
Another rare crash, this time during the GAP OscarInterface tests, log:
```
-> Testing GAP/oscar_to_gap.jl took: runtime 0.082 seconds + compilation 2.391 seconds + recompilation 0.000 seconds, 89.49 MB
Architecture: x86_64-pc-linux-gnu-julia1.11-64-kv8

testing: /home/runner/work/Oscar.jl/Oscar.jl/src/../gap/OscarInterface//tst/al\
nuth/ALNUTH.tst
   20958 ms (1274 ms GC) and 5.58MB allocated for alnuth/ALNUTH.tst
testing: /home/runner/work/Oscar.jl/Oscar.jl/src/../gap/OscarInterface//tst/al\
nuth/examples.tst
    1444 ms (915 ms GC) and 15.2MB allocated for alnuth/examples.tst
testing: /home/runner/work/Oscar.jl/Oscar.jl/src/../gap/OscarInterface//tst/al\
nuth/manual.tst
    1319 ms (1096 ms GC) and 2.90MB allocated for alnuth/manual.tst
testing: /home/runner/work/Oscar.jl/Oscar.jl/src/../gap/OscarInterface//tst/al\
nuth/polynome.tst

[3037] signal (11.1): Segmentation fault
in expression starting at /home/runner/work/Oscar.jl/Oscar.jl/test/GAP/runtests.jl:11
gc_mark_outrefs at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:2499 [inlined]
gc_mark_loop_serial_ at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:2669
gc_mark_loop_serial at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:2692
gc_mark_loop at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:2805 [inlined]
_jl_gc_collect at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:3124
ijl_gc_collect at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:3446
maybe_collect at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:916 [inlined]
ijl_gc_counted_malloc at /cache/build/default-amdci4-1/julialang/julia-master/src/gc.c:3610
[...]

PR #2556, opened 2023-07-13
Rename `is_hermitian_matrix` to `is_hermitian`

lgoettgens:
`is_hermitian` already exists for other types and is more streamlined with similar function names, e.g. `is_symmetric`.

### DIFF
--- a/test/Groups/operations.jl
+++ b/test/Groups/operations.jl
@@ -86,7 +86,7 @@ end
    x=matrix(F,4,4,[1,z,0,0,0,1,z^2,z,z,0,0,1,0,0,z+1,0])
    y=x+transpose(x)
    @test is_symmetric(y)
-   @test is_hermitian_matrix(x+conjugate_transpose(x))
+   @test is_hermitian(x+conjugate_transpose(x))
    @test is_skewsymmetric_matrix(y)
    y[1,1]=1
    @test !is_skewsymmetric_matrix(y)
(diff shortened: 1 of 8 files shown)


lgoettgens:
The nightly failure seems to be https://github.com/oscar-system/Oscar.jl/issues/2543.

thofma (approved):

fingolfin (approved):

test/Groups/operations.jl
```diff
@@ -86,7 +86,7 @@ end
    x=matrix(F,4,4,[1,z,0,0,0,1,z^2,z,z,0,0,1,0,0,z+1,0])
    y=x+transpose(x)
    @test is_symmetric(y)
-   @test is_hermitian_matrix(x+conjugate_transpose(x))
+   @test is_hermitian(x+conjugate_transpose(x))
    @test is_skewsymmetric_matrix(y)
```
fingolfin, line 90: I am fine with this change, but then perhaps we should also do the same for `is_skewsymmetric` ?

lgoettgens: Please see https://github.com/oscar-system/Oscar.jl/issues/2555 for this.
There are two semantically different implementations that would be merged by renaming `is_skewsymmetric_matrix`.

lgoettgens:
The errors on nightly and 1.10-alpha1 are not due to changes here.

PR #2559, opened 2023-07-13
add is_isomorphic_with_permutation for graphs

simonbrandhorst:
This originates in a question that I asked a while ago on slack.
One could also argue for the name
`is_isomorphic_with_permutation` 
instead of
`is_isomorphic_with_map`

### DIFF
--- a/src/Combinatorics/Graphs/functions.jl
+++ b/src/Combinatorics/Graphs/functions.jl
@@ -536,8 +536,7 @@ julia> add_edge!(g,1,2); add_edge!(g,2,3); add_edge!(g,3,4); add_edge!(g,4,5); a
 julia> signed_incidence_matrix(g)
 5×5 Matrix{Int64}:
  -1   0   0   0   1
-  1  -1   0   0   0Tuple{Vector{WeilDivisor{CoveredScheme{Nemo.fpField}, ZZRing, ZZRingElem}}, ZZMatrix}
-
+  1  -1   0   0   0
   0   1  -1   0   0
   0   0   1  -1   0
   0   0   0   1  -1
--- a/src/Combinatorics/Graphs/functions.jl
+++ b/src/Combinatorics/Graphs/functions.jl
@@ -40,8 +40,10 @@ end
 
 Return the graph with adjacency matrix `G`.
 
-This means that the nodes ``i > j`` are connected by an edge
+This means that the nodes ``i, j`` are connected by an edge
 if and only if ``G_{i,j}`` is one.
+In the undirected case, it is assumed that ``i > j`` i.e. the upper triangular
+part of ``G`` is ignored.
 
 # Examples
 ```jldoctest
@@ -878,7 +880,8 @@ Return a permutation `I` such that `A1[I,I] == A2` and whether it exists.
 
 The method assumes that both matrices are symmetric, their diagonal entries
 are all equal (and so irrelevant) and the off-diagonal entries are either ``0``
-or ``1``.
+or ``1``. It is assumed that `A1` and `A2` are symmetric and
+their upper triangular part is ignored.
 """
 function _is_equal_up_to_permutation_with_permutation(A1::MatElem, A2::MatElem)
   g1 = graph_from_adjacency_matrix(Undirected, A1)
--- a/src/Combinatorics/Graphs/functions.jl
+++ b/src/Combinatorics/Graphs/functions.jl
@@ -536,7 +536,8 @@ julia> add_edge!(g,1,2); add_edge!(g,2,3); add_edge!(g,3,4); add_edge!(g,4,5); a
 julia> signed_incidence_matrix(g)
 5×5 Matrix{Int64}:
  -1   0   0   0   1
-  1  -1   0   0   0
+  1  -1   0   0   0Tuple{Vector{WeilDivisor{CoveredScheme{Nemo.fpField}, ZZRing, ZZRingElem}}, ZZMatrix}
+
   0   1  -1   0   0
   0   0   1  -1   0
   0   0   0   1  -1
@@ -788,8 +789,13 @@ false
 """
 is_isomorphic(g1::Graph{T}, g2::Graph{T}) where {T <: Union{Directed, Undirected}} = Polymake.graph.isomorphic(pm_object(g1), pm_object(g2))::Bool
 
+@doc raw"""
+    is_isomorphic_with_permutation(G1::Graph, G2::Graph) -> Bool, Vector{Int}
 
-function is_isomorphic_with_map(G1::Graph, G2::Graph)
+Return whether `G1` is isomorphic to `G2` as well as a permutation
+of the nodes of `G1` such that both graphs agree.
+"""
+function is_isomorphic_with_data(G1::Graph, G2::Graph)
   f12 = Polymake.graph.find_node_permutation(G1.pm_graph, G2.pm_graph)
   if isnothing(f12)
     return false, Vector{Int}()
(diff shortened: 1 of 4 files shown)


afkafkafk13:
Could you add tests and documentation? Then you are good to go!

fieker:
On Thu, Jul 13, 2023 at 12:30:51PM -0700, simonbrandhorst wrote:
> This originates in a question that I asked a while ago on slack.
> One could also argue for the name
> `is_isomorphic_with_permutation` 
> instead of
> `is_isomorphic_with_map`
Does it return a map or a permutation?
Since it is not returning a map (of type <: Map) and neither a permutation
(of type PermGroupElem)... maybe even is_isomorphic_with_data?

Do we have maps (of type <: Map) for Graphs? And, maybe @ThomasBreuer
do permutations (GroupPermElem) operate on graphs? Should they?

simonbrandhorst:
For me it is convenient to return a list of integers  `I`. Then `M[I,I]` for some square matrix `M` of the right size works.

src/Combinatorics/Graphs/functions.jl
```diff
@@ -788,6 +788,35 @@ false
...
+      if isone(G[i,j])
+        add_edge!(g,i,j)
+      end
+    end
+  end
+  return g
+end
+
+function is_isomorphic_with_permutation(A1::MatElem, A2::MatElem)
+  b, T = is_isomorphic_with_map(graph(A1),graph(A2))
```
fingolfin: This function name makes no sense to me; those matrices aren't "isomorphic"; the graphs they define are.

src/Combinatorics/Graphs/functions.jl
```diff
@@ -788,6 +788,35 @@ false
...
+end
+
+function graph(G::MatElem)
+  n = nrows(G)
+  g = Graph{Undirected}(n)
+  for i in 1:n
+    for j in 1:i-1
+      if isone(G[i,j])
+        add_edge!(g,i,j)
+      end
```
fingolfin: Maybe also reject invalid inputs?
```suggestion
      else !iszero(G[i,j])
        error("invalid incidence matrix")
      end
```

src/Combinatorics/Graphs/functions.jl
```diff
@@ -788,6 +788,35 @@ false
 """
 is_isomorphic(g1::Graph{T}, g2::Graph{T}) where {T <: Union{Directed, Undirected}} = Polymake.graph.isomorphic(pm_object(g1), pm_object(g2))::Bool
 
+
+function is_isomorphic_with_map(G1::Graph, G2::Graph)
```
fingolfin: The second return value is not a map, so this name is seems misleading to me

src/Combinatorics/Graphs/functions.jl
```diff
@@ -788,6 +788,35 @@ false
...
+  f12 = Polymake.graph.find_node_permutation(G1.pm_graph, G2.pm_graph)
+  if isnothing(f12)
+    return false, Vector{Int}()
+  end
+  return true, Polymake.to_one_based_indexing(f12)
+end
+
+function graph(G::MatElem)
```
fingolfin: definitely should have a comment (or better: docstring) that says what this does

src/Combinatorics/Graphs/functions.jl
afkafkafk13: I support @fieker 's suggestions of is_isomorphic_with_data for the comparison of the graphs.
This suggests to the user that one should look at the documentation to find out, what the data is.

src/Combinatorics/Graphs/functions.jl
simonbrandhorst: will do :-)

simonbrandhorst:
I agree that `map` makes no sense right now. 
The return value is a permutation of the nodes. And the nodes are labeled by `1,...,n`.  
I prefer  `permutation` over `data` because `data` is less informative.
Note that we do not see the permutation as a group element here. 
Therefore I do not see much of a benefit in encapsulating the return value in a gap permutation group element.

In any case I do not have a strong opinion on any of this.

lkastner (commented):
As already remarked, this looks good, but
- the functions are not exported
- there are no tests
- there is no documentation.

Furthermore, if the matrices are incidence matrices, why not use Oscar's type `IncidenceMatrix`?
```
julia> IM = IncidenceMatrix(3,3)
3×3 IncidenceMatrix
[]
[]
[]


julia> IM[1,2] = 1
1

julia> IM
3×3 IncidenceMatrix
[2]
[]
[]


julia> Polymake.Meta.translate_type_to_pm_string(::Type{Directed})="Directed"

julia> Polymake.Meta.translate_type_to_pm_string(::Type{Undirected})="Undirected"

julia> GD = Polymake.graph.Graph{Directed}(ADJACENCY=IM)
type: Graph<Directed>

ADJACENCY
	{1}
	{}
	{}


julia> Graph{Directed}(GD.ADJACENCY)
Graph{Directed}(pm::graph::Graph<pm::graph::Directed>
{1}
{}
{}
)

julia> GUD = Polymake.graph.Graph{Undirected}(ADJACENCY=IM)
type: Graph<Undirected>

ADJACENCY
	{1}
	{0}
	{}


julia> Graph{Undirected}(GUD.ADJACENCY)
Graph{Undirected}(pm::graph::Graph<pm::graph::Undirected>
{1}
{0}
{}
)
```
There are some hacks here, to make everything work, but we will move these elsewhere then later. The output of graphs is currently the raw polymake output, so it is 0 based. I'll have to modify this.

src/Combinatorics/Graphs/functions.jl
simonbrandhorst: I renamed the function ,documented it and now it has an `_` because it is quite specialized (I need it for elliptic surfaces).

simonbrandhorst:
It is  documented, tested and the relevant methods are exported. 
Note that the names & interfaces have changed quite a lot.
May I leave the hacks to the polymake experts and another pull request?

src/Combinatorics/Graphs/functions.jl
```diff
@@ -35,6 +35,68 @@ function Graph{T}(nverts::Int64) where {T <: Union{Directed, Undirected}}
...
+Graph{Directed}(pm::graph::Graph<pm::graph::Directed>
+{}
+{0}
+)
+
+julia> graph_from_adjacency_matrix(Undirected, G)
+Graph{Undirected}(pm::graph::Graph<pm::graph::Undirected>
+{1}
+{0}
+)
```
afkafkafk13, line 64: Tiny nitpick on the doctest:
This output is not very helpful. It is also suppressed in the documentation of the graphs. So we might do the same thing here and print information on vertices and edges instead.

src/Combinatorics/Graphs/functions.jl
```diff
@@ -35,6 +35,68 @@ function Graph{T}(nverts::Int64) where {T <: Union{Directed, Undirected}}
...
+
+```
+"""
+graph_from_adjacency_matrix(::Type, G::Union{MatElem, Matrix})
+
```
afkafkafk13, line 69: Is this line a leftover or does it have some important meaning that I have missed?

src/Combinatorics/Graphs/functions.jl
```diff
@@ -788,6 +850,47 @@ false
...
+end
+
+@doc raw"""
+    _is_equal_up_to_permutation_with_permutation(A1::MatElem, A2::MatElem) -> Bool, Vector{Int}
+
+Return a permutation `I` such that `A1[I,I] == A2` and whether it exists.
+
+The method assumes that both matrices are symmetric, their diagonal entries
+are all equal (and so irrelevant) and the off-diagonal entries are either ``0``
+or ``1``.
```
afkafkafk13: Just a nitpick:
You require matrices to be symmeric here, but further up in graph_from_adjacency_matrix, you only looked at the lower triangular part. This makes perfect sense to me, but it might be useful to state it more clearly (entries above the diagonal are ignored!).

lkastner (approved):

lkastner:
Since there are open comments from @afkafkafk13 , I'll not merge it, but feel free to go ahead. I will then have another look later on whether there is something else to improve.

src/Combinatorics/Graphs/functions.jl
simonbrandhorst: It serves as an anchor for the documentation. It allows me to document both methods at the same time.

src/Combinatorics/Graphs/functions.jl
simonbrandhorst: Yes. Once the show functions of the graphs improve, this doctests will be more helpful. 
For now it is just intended to show how to use the constructor and so that the user has something to copy paste.

simonbrandhorst:
I have followed @afkafkafk13  suggestions and added the methods to our documentation. 
Thank for insisting on doing this properly. (I was somewhat a bit too lazy).

afkafkafk13 (approved):

PR #2561, opened 2023-07-14
[FTheoryTools] Improve construction of models over arbitrary base spaces

HereAround:


src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
```diff
@@ -491,6 +491,49 @@ end
 
 
 
+@doc raw"""
+    normal_toric_varieties_from_star_triangulations(P::Polyhedron; set_attributes::Bool = true)
```
apturner: The function name at the start of the docstring here doesn't seem to match that actual function name (pluralization)

apturner (approved):
Looks good, save for the misnamed function in the docstring. Thanks @HereAround!

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
HereAround: Thx for spotting this. Just pushed an update which (should) fix this.

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -2,36 +2,44 @@
...
-  fan_max_cone_matrix = matrix(ZZ, cones(f))
-  new_max_cones = Vector{Int}[]
-  for k in 1:nrows(fan_max_cone_matrix)
-    indices = findall(x -> x == 1, vec(fan_max_cone_matrix[k,:]))
-    if length(indices) == d
-      push!(new_max_cones, indices)
+  auxiliary_base_space = variety
+  if dim(auxiliary_base_space) != d
+    integral_rays = matrix(ZZ, rays(variety))
+    integral_rays = [vec([Int(a) for a in integral_rays[k,:]]) for k in 1:nrows(integral_rays)]
```
lkastner: Casting to `Int` can result in a loss of precision. Why is this necessary?

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -2,36 +2,44 @@
...
+    fan_max_cone_matrix = matrix(ZZ, cones(variety))
+    new_max_cones = Vector{Int}[]
+    for k in 1:nrows(fan_max_cone_matrix)
+      cone = positive_hull(integral_rays[all_cones[k,:]])
+      if dim(cone) == d
+        indices = findall(x -> x == 1, vec(fan_max_cone_matrix[k,:]))
+        push!(new_max_cones, indices)
+      end
     end
+    auxiliary_base_space = normal_toric_variety(integral_rays, new_max_cones; non_redundant = true)
```
lkastner: Please do not use  `cones(*)` here, instead use `cones(*,Int)`. Everytime you call `positive_hull` a new cone is assembled and you restart computations that are already in the fan.

Something like the following should work:
```suggestion
    new_max_cones = IncidenceMatrix(cones(variety, d))
    auxiliary_base_space = normal_toric_variety(integral_rays, new_max_cones; non_redundant = true)
```
In case this does not work immediately, here is an example:
```
julia> p2 = projective_space(NormalToricVariety, 2)
Normal, non-affine, smooth, projective, gorenstein, fano, 2-dimensional toric variety without torusfactor

julia> skel = normal_toric_variety(matrix(ZZ,rays(p2)), IncidenceMatrix(cones(p2, 1)))
Normal toric variety

julia> dim(skel)
1

julia> skel = normal_toric_variety(rays(p2), IncidenceMatrix(cones(p2, 1)))
Normal toric variety

julia> dim(skel)
1
```

lkastner (changes requested):

experimental/FTheoryTools/src/auxiliary.jl
HereAround: It has been a while. My recollection is that one of the functions that I am using required a matrix with `Int` entries. Maybe `convex_hull`. But then, your suggestion below (which looks great) avoids that altogether. So there is then probably no need for this cast anymore. Let me try out your suggestion.

HereAround: Perfect. Works!

lkastner: This is not resolved, since there was no suggestion here. Maybe something like:
```suggestion
```

lkastner (changes requested):

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Sorry. My bad. I simply forgot to delete this line (as I see it, it is no longer needed once I applied your suggestion below). I just pushed the corresponding change.

lkastner (approved):

PR #2562, opened 2023-07-14
Wrap saturation2 and use it for Laurent polynomials

thofma:
@wdecker: I found out that the new saturation is not wrapped yet on the Oscar side. We need this for the Laurent polynomial ideal business, so I would like to add it in this preliminary form. Someone can improve the integration later.

@micjoswig: This will make the example from Laura run within a few seconds (without this change I could not make it finish).

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -273,9 +273,19 @@ ideal(z, x*y)
 function saturation(I::MPolyIdeal{T}, J::MPolyIdeal{T}) where T
   singular_assure(I)
   singular_assure(J)
-  K, _ = Singular.saturation(I.gens.S, J.gens.S)
+  K, _ = Singular.saturation2(I.gens.S, J.gens.S)
   return MPolyIdeal(base_ring(I), K)
 end
+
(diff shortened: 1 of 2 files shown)


src/Rings/mpoly-ideals.jl
```diff
@@ -273,9 +273,19 @@ ideal(z, x*y)
 function saturation(I::MPolyIdeal{T}, J::MPolyIdeal{T}) where T
   singular_assure(I)
   singular_assure(J)
-  K, _ = Singular.saturation(I.gens.S, J.gens.S)
+  K, _ = Singular.saturation2(I.gens.S, J.gens.S)
```
```suggestion
  K, _ = Singular.saturation(I.gens.S, J.gens.S)
```
Due to this change `saturation` and `_saturation2` are exactly the same. I think this change here was not meant to be made.

thofma:
```suggestion
  K, _ = Singular.saturation(I.gens.S, J.gens.S)
```

thofma: Oops, thanks!

wdecker:
@thofma: Go ahead!

Von meinem iPhone gesendet

PR #2566, opened 2023-07-17
Change `is_skewsymmetric_matrix` to `is_alternating`

### LINKED ISSUE #2555: Two different skew-symmetry checks

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/2555.

I renamed `is_skewsymmetric_matrix` to `is_alternating`, and updated all uses accordingly. In one test, there was a natural place to test that this indeed differs from `is_skew_symmetric` in characteristic 2.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -394,9 +394,6 @@ function SubdivisionOfPoints{T}(obj::Polymake.BigObject) where T<:scalar_types
   return SubdivisionOfPoints{T}(obj, _detect_default_field(T, obj))
 end
 
-@deprecate is_skewsymmetric_matrix(B::MatElem{T}) where T <: RingElem is_alternating(B)
-@alias isskewsymmetric_matrix is_skewsymmetric_matrix
-
 @deprecate is_hermitian_matrix(B::MatElem{T}) where T <: FinFieldElem is_hermitian(B)
 @alias ishermitian_matrix is_hermitian_matrix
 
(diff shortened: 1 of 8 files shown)


src/deprecations.jl
```diff
@@ -394,5 +394,8 @@ function SubdivisionOfPoints{T}(obj::Polymake.BigObject) where T<:scalar_types
   return SubdivisionOfPoints{T}(obj, _detect_default_field(T, obj))
 end
 
+@deprecate is_skewsymmetric_matrix(B::MatElem{T}) where T <: RingElem is_alternating(B)
+@alias isskewsymmetric_matrix is_skewsymmetric_matrix
+
```
lgoettgens: Do we really want this deprecation? I find it very misleading from its name. As the next release will be breaking anyway, we could skip adding it.

fingolfin: I am happy with omitting it

fingolfin (approved):

PR #2567, opened 2023-07-18
[FTheoryTools] Add Morrison–Park model and KBar parameter

apturner:
Adds the Morrison–Park model as a simpler example of a literature hypersurface model, and adds a parameter "KBar" to all literature models that don't already have a parameter in the class [KBar].


HereAround:
Thank you very much indeed @apturner.

HereAround (approved):

PR #2568, opened 2023-07-18
Prepare for type renaming

thofma:
This will make https://github.com/thofma/Hecke.jl/pull/1170 non-breaking.


benlorenz (approved):
That should work, thanks!

PR #2569, opened 2023-07-18
[FTheoryTools] Support for literature model parameters

apturner:
Adds support for literature model parameters, as well as associated attributes, properties, and documentation.


HereAround:
Thank you @apturner for adding this functionality! This was the exact thing that I found missing a couple of days back. Nice!

However, unless I am mistaken, we do not have an example/test which constructs a literature model with model parameters. Please correct if I am mistaken.

If indeed correct, then I would suggest adding one, so that we can rest assured that the model parameters code works as intended.

apturner:
We do actually have such examples; all 7 of the Lawrie/Schäfer-Nameki models have a parameter, and the `related_literature_models` and `model_parameters` functions that are added in this PR have doctests that construct these models, so this PR already tests that it works.

experimental/FTheoryTools/src/LiteratureModels/attributes.jl
```diff
@@ -420,6 +420,28 @@ function model_description(m::AbstractFTheoryModel)
...
+```jldoctest
+julia> m = literature_model(arxiv_id = "1212.2949", equation = "3.2", model_parameters = Dict("k" => 5))
+Assuming that the first row of the given grading is the grading under Kbar
+
+Global Tate model over a not fully specified base -- SU(2k+1) Tate model with parameter values (k = 5) based on arXiv paper 1212.2949 Eq. (3.2)
+
+julia> model_parameters(m)
+Dict{String, Int64} with 1 entry:
+  "k" => 5
+```
```
HereAround, line 438: Very good. This here is exactly what I was looking for earlier, but somehow must have overlooked this test. Nice!

HereAround (approved):
Thank you @apturner. Looks good to me.

Issue #2541, opened 2023-07-10
BrauerGroup: assertion error in nightly CI on master


benlorenz:
While looking into the toric varieties timing error I noticed this failed assertion during the nightly tests on master, https://github.com/oscar-system/Oscar.jl/actions/runs/5509507938/jobs/10042321956.

Log:
```
Test Summary:        | Pass  Total   Time
Experimental.gmodule |   10     10  53.6s
Test Summary:        | Pass  Total     Time
Experimental LocalH2 |   75     75  1m03.3s
Test Summary:   | Pass  Total   Time
GlobalFundClass |    4      4  47.1s
BrauerGroup: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Experimental/gmodule.jl:85
  Got exception outside of a @test
  AssertionError: norm(v) == u
  Stacktrace:
    [1] local_fundamental_class_serre(mKL::Hecke.LocalFieldMor{Hecke.LocalField{qadic, Hecke.EisensteinLocalField}, Hecke.LocalField{qadic, Hecke.EisensteinLocalField}, Hecke.MapDataFromLocalField{Hecke.LocalFieldElem{qadic, Hecke.EisensteinLocalField}, Hecke.MapDataFromQadicField{Hecke.LocalFieldElem{qadic, Hecke.EisensteinLocalField}}}, Hecke.MapDataFromLocalField{Hecke.LocalFieldElem{qadic, Hecke.EisensteinLocalField}, Hecke.MapDataFromQadicField{Hecke.LocalFieldElem{qadic, Hecke.EisensteinLocalField}}}, Hecke.LocalFieldElem{qadic, Hecke.EisensteinLocalField}})
      @ Hecke ~/.julia/packages/Hecke/PJiPq/src/LocalField/neq.jl:765
    [2] local_index(CC::Vector{Oscar.GrpCoh.CoChain{2, PermGroupElem, Oscar.GrpCoh.MultGrpElem{nf_elem}}}, P::NfOrdIdl, mG::AbstractAlgebra.Generic.CompositeMap{PermGroup, Hecke.NfMorSet{AnticNumberField}}; B::Oscar.GaloisCohomology_Mod.RelativeBrauerGroup, index_only::Bool)
      @ Oscar.GaloisCohomology_Mod ~/work/Oscar.jl/Oscar.jl/experimental/GModule/GaloisCohomology.jl:1225
[...]

PR #2574, opened 2023-07-19
Start web server after building the documentation

joschmitt:
As discussed in Slack, the `open_browser` option of `build_doc` does not work (at least) on Ubuntu's sandboxed snap-firefox. I added a comment regarding this to the documentation (first commit).
@benlorenz suggested to start a web server using `LiveServer.jl` in the build directory, so I also tried this.
~I'm not really familiar with the setup for building the documentation and eventually added `LiveServer` as a dependency at every place I could find because this seemed to be the only way to make it work. Is there a better way and do we want to have this dependency at all?~

As of now this will of course simply start the server, but not update anything, when the user edits the documentation (like the `LiveServer.servedocs` function claims to do).

### DIFF
--- a/src/utils/docs.jl
+++ b/src/utils/docs.jl
@@ -120,7 +120,7 @@ function start_doc_preview_server(;open_browser::Bool = true, port::Int = 8000)
         using LiveServer;
         LiveServer.serve(dir = "$build_dir", launch_browser = $open_browser, port = $port);
         """
-  live_server_process = run(`julia -e $cmd`, wait = false)
+  live_server_process = run(`$(Base.julia_cmd()) -e $cmd`, wait = false)
   atexit(_ -> kill(live_server_process))
   @info "Starting server with PID $(getpid(live_server_process)) listening on 127.0.0.1:$port"
   return nothing
--- a/src/utils/docs.jl
+++ b/src/utils/docs.jl
@@ -155,7 +155,7 @@ terminated, so the proposed usage for this option is to run
 `build_doc(open_browser = false)` for following builds and only refresh the
 browser tab.
 
-When working on the manual the `Revise` package can significantly sped
+When working on the manual the `Revise` package can significantly speed
 up running `build_doc`. First, install `Revise` in the following way:
 ```
 using Pkg ; Pkg.add("Revise")
@@ -164,7 +164,7 @@ Second, restart Julia and load `Revise` before Oscar:
 ```
 using Revise, Oscar;
 ```
-The first run of `build_doc` will take the usual few minutes, subsequently runs
+The first run of `build_doc` will take the usual few minutes, subsequent runs
 will be significantly faster.
 """
 function build_doc(; doctest=false, strict=false, open_browser=true, start_server = false)
--- a/src/utils/docs.jl
+++ b/src/utils/docs.jl
@@ -111,9 +111,19 @@ function open_doc()
     end
 end
 
+function start_doc_preview_server(;open_browser::Bool = true, port::Int = 8000)
+  build_dir = normpath(Oscar.oscardir, "docs", "build")
+  cmd = "using Pkg; Pkg.activate(temp = true);
+         Pkg.add(\"LiveServer\"); using LiveServer;
+         LiveServer.serve(dir = \"$build_dir\", launch_browser = $open_browser, port = $port);"
+  live_server_process = run(`julia -e $cmd`, wait = false)
+  atexit(_ -> kill(live_server_process))
+  @info "Starting server with PID $(getpid(live_server_process)) listening on 127.0.0.1:$port"
+  return nothing
+end
 
 @doc raw"""
-    build_doc(; doctest=false, strict=false, open_browser=true)
+    build_doc(; doctest=false, strict=false, open_browser=true, start_server=false)
 
 Build the manual of `Oscar.jl` locally and open the front page in a
 browser.
@@ -136,6 +146,15 @@ doctesting error will always make makedocs throw an error in this mode".
 To prevent the opening of the browser at the end, set the optional parameter
 `open_browser` to `false`.
 
+Alternatively, one can use the optional parameter `start_server` to start a web
+server in the build directory which is accessible via `127.0.0.1:8000`. If both
+`start_server` and `open_browser` are `true`, the browser will show this page.
+The server keeps running in the background until the `julia` session is
+terminated, so the proposed usage for this option is to run
+`build_doc(start_server = true)` for the first build and
+`build_doc(open_browser = false)` for following builds and only refresh the
+browser tab.
+
 When working on the manual the `Revise` package can significantly sped
 up running `build_doc`. First, install `Revise` in the following way:
 ```
@@ -148,7 +167,7 @@ using Revise, Oscar;
 The first run of `build_doc` will take the usual few minutes, subsequently runs
 will be significantly faster.
 """
-function build_doc(; doctest=false, strict=false, open_browser=true)
+function build_doc(; doctest=false, strict=false, open_browser=true, start_server = false)
   versioncheck = (VERSION.major == 1) && (VERSION.minor >= 7)
   versionwarn = 
 "The Julia reference version for the doctests is 1.7 or later, but you are using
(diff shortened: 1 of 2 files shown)


benlorenz:
I don't really like adding it as a dependency.
Maybe we can just run it via a separate julia process and keep the liveserver process in the background, storing the process object in a variable to make sure we start it only once. (Stopping it with a `Base.atexit` hook).

The command line might look like this:
```
julia -e 'using Pkg; Pkg.activate(temp=true); Pkg.add("LiveServer"); using LiveServer; LiveServer.serve(dir="$(docspath), open_browser=$(open_browser)");'
```

I don't know if we want to try to integrate the `servedocs` variant, it does not work for adjusted docstrings anyway (only .md files). Maybe the best approach is to have users run the normal docs build again (with Revise) and keep the server + browser running, just refreshing the page after the build is done again.

joschmitt:
Here is a version that starts `LiveServer` as a background process. For me, this is fine, but there are some nuisances:
* If the background process fails, the user will not get any feedback. I tried to redirect `stderr` (but not `stdout`) for this, but somehow one either gets `stderr` AND `stdout` or nothing?
* If there is already something running at port 8000, `LiveServer` chooses another port and I have no idea how to find out which (other than reading the output from `stdout`...).
* If one kills the main julia process via SIGKILL, the `atexit` does not get executed and the background process stays alive.

joschmitt:
~Looks like another occurrence of #2541 (in this case not on nightly).~
Well, not any more.

joschmitt:
Small update on my comments above: one can in fact redirect only `stderr`, it's just that `Pkg.activate` prints to `stderr` by default, but one can also redirect that. Then again, I'm not too convinced any more that one should print the `stderr` of the background process to the calling julia session as it will just result in out-of-context error messages.

benlorenz:
I am still in favor of doing this, but I would love some other opinions on whether we should add this.

Maybe we should put `live_server_process` in a global variable to be able to check whether it is still running when calling the function a second time?
And I think hiding the error output should be fine.

joschmitt:
There was a short discussion on this in the Wednesday's meeting in Kaiserslautern:
* Potentially @aaruni96 is interested in this
* We could consider the (third?) option of providing the functionality in principle, but telling the user to install `LiveServer` by themselves, if they want to use it.

src/utils/docs.jl
```diff
@@ -145,10 +164,10 @@ Second, restart Julia and load `Revise` before Oscar:
 ```
 using Revise, Oscar;
 ```
-The first run of `build_doc` will take the usual few minutes, subsequently runs
+The first run of `build_doc` will take the usual few minutes, subsequent runs
 will be significantly faster.
 """
-function build_doc(; doctest=false, strict=false, open_browser=true)
+function build_doc(; doctest=false, strict=false, open_browser=true, start_server = false)
```
fingolfin:
```suggestion
function build_doc(; doctest::Union{Symbol,Bool}=false, strict::Bool=false, open_browser::Bool=true, start_server::Bool=false)
```

src/utils/docs.jl
```diff
@@ -162,9 +181,12 @@ $(VERSION). Running the doctests will produce errors that you do not expect."
   Pkg.activate(docsproject) do
     Base.invokelatest(Main.BuildDoc.doit, Oscar; strict=strict, local_build=true, doctest=doctest)
   end
-  if open_browser
+  if open_browser && !start_server
     open_doc()
   end
+  if start_server
+    start_doc_preview_server(open_browser = open_browser)
+  end
```
fingolfin: IMHO this would be more natural, but it's a matter of taste, we can also leave it *shrug*
```suggestion
  if start_server
    start_doc_preview_server(open_browser = open_browser)
  elseif open_browser
    open_doc()
  end
```

fingolfin (approved):
I haven't tested this, but it looks great, thanks!

joschmitt:
As far as I know, @aaruni96 wanted to work on an alternative solution... For the record, the following approaches have been mentioned/tried out:
1. Add LiveServer.jl as a dependency. We don't want that.
2. Start a background process that installs LiveServer.jl in a temporary environment and starts the server (the current state of this PR). Comes with some drawbacks, most notably that we start a background process. We could improve the interface a bit by remembering in a global variable (or maybe we can even check?) whether a server is already running.
3. Only provide a function that starts the server and tell the user that they have to install LiveServer.jl themselves. Would also be fine for me; the `test_module` function also tells you that you need to do `using Test` after all.
4. Aaruni's multithreading solution.

(Un)fortunately, I'm on holidays the next two weeks, so if I am further involved in this it has to wait until after that.

aaruni96:
I apologize, I was refreshing some julia multiprocessing while thinking about this PR and forgot to update here.

The solution I propose such: 

We can spawn a julia worker process (`addprocs()` from `Distributed`) , and ask it to load LiveServer in a temporary environment. We can then check if the server is still running by asking the status of the future, or kill it by sending an interrupt to the worker.

This would require adding `Distributed` to Oscar dependencies (currently, its only a dependency for the tests).

The solution *should* work on single core CPUs as well (the OS can schedule multiple jobs), but its something I am not fully sure about.

fingolfin:
@aaruni96 I think adding a dependency on the stdlib `Distributed` is fine, so feel free to go ahead with providing an alternate PR.

fingolfin:
@aaruni96 look into an alternative implementation, but it turns out that the approach here also allows some control over the subprocess running the webserver (should we want to), which we were unaware of before. So the main reason for an alternate approach really does not hold.

So I would just merge this PR -- we can still improve it later, if we want to (as with any feature).

@joschmitt you marked it as draft again -- I assume because of our discussion, but I want to double check just in case you had other reasons?

src/utils/docs.jl
```diff
@@ -111,9 +111,19 @@ function open_doc()
     end
 end
 
+function start_doc_preview_server(;open_browser::Bool = true, port::Int = 8000)
+  build_dir = normpath(Oscar.oscardir, "docs", "build")
+  cmd = "using Pkg; Pkg.activate(temp = true);
+         Pkg.add(\"LiveServer\"); using LiveServer;
+         LiveServer.serve(dir = \"$build_dir\", launch_browser = $open_browser, port = $port);"
```
fingolfin: How about using triple quotes to avoid quoting?
```suggestion
  cmd = """
         using Pkg
         Pkg.activate(temp = true)
         Pkg.add("LiveServer")
         using LiveServer
         LiveServer.serve(dir = "$build_dir", launch_browser = $open_browser, port = $port)
         """
```

src/utils/docs.jl
```diff
@@ -111,9 +111,19 @@ function open_doc()
...
+  live_server_process = run(`julia -e $cmd`, wait = false)
+  atexit(_ -> kill(live_server_process))
```
fingolfin, line 124: Did you verify this works? Because when I just tried this by copy&pasting this code, `kill` was just ignored. I suspect this was because I am using `juliaup`, which actually forks `julia` into a subprocess...

joschmitt:
> @joschmitt you marked it as draft again -- I assume because of our discussion, but I want to double check just in case you had other reasons?

Yes, I wanted to avoid that somebody merges this "accidentally". Let me add your suggestions and then I will mark it as ready for review.

src/utils/docs.jl
joschmitt: Okay... I'm very sure it worked for me at some point. I will try again.

src/utils/docs.jl
joschmitt: Didn't know about this!

src/utils/docs.jl
joschmitt: So yes, for me it works.

src/utils/docs.jl
```diff
@@ -111,9 +111,23 @@ function open_doc()
...
+function start_doc_preview_server(;open_browser::Bool = true, port::Int = 8000)
+  build_dir = normpath(Oscar.oscardir, "docs", "build")
+  cmd = """
+        using Pkg;
+        Pkg.activate(temp = true);
+        Pkg.add("LiveServer");
+        using LiveServer;
+        LiveServer.serve(dir = "$build_dir", launch_browser = $open_browser, port = $port);
+        """
+  live_server_process = run(`julia -e $cmd`, wait = false)
```
benlorenz:
```suggestion
  live_server_process = run(`$(Base.julia_cmd()) -e $cmd`, wait = false)
```
There might not be a julia binary in the path? (this might also avoid the forking by juliaup)

src/utils/docs.jl
joschmitt: Could you try again? Maybe it is fixed by https://github.com/oscar-system/Oscar.jl/pull/2574/commits/7af4c60c74ff770d47f81569206a08a13e23723e as @benlorenz suggested.

joschmitt: I tried it now with `juliaup` and it works (with https://github.com/oscar-system/Oscar.jl/commit/7af4c60c74ff770d47f81569206a08a13e23723e). Without that commit it doesn't, so I assume that this is fixed.

Issue #2093, opened 2023-03-20
Better handling of matrix groups over various rings


ThomasBreuer:
Currently Oscar deals as follows with a matrix group `G` over the ring `R`.

- The generators of `G` are Oscar matrices,  `G` itself stores a map `G.ring_iso` that maps elements of `R` to a corresponding ring in GAP, and on demand an isomorphic GAP matrix group `G.X` gets created whose elements are GAP matrices, such that the image of an element `M` of `G` is computed as `map_entries(G.ring_iso, M.elm)`. Conversely, one can map matrices on the GAP side back to Oscar matrices.

  - This works well for matrices over small finite fields, for which GAP provides a compressed internal representation and efficient arithmetic operations.

  - This still works for a finite matrix group `G` over a cyclotomic or quadratic field. Here the image of `G.ring_iso` is a GAP field of "cyclotomics" (in a special internal representation), for which GAP provides more functionality than for general algebraic
  extension fields of the field of rationals.

  - This is not a really good solution for finite matrix groups over other number fields, or for infinite matrix groups.

- For a finite matrix group over a general number field, the function `_isomorphic_group_over_finite_field` computes
  a reduction to an isomorphic matrix group over a finite field.

  This approach is currently used automatically only in a method for computing the order of a given matrix group.
  One way to use it also for other questions would be to change the generic treatment:
  Instead of using always `G.ring_iso` for creating an isomorphic GAP group and then leaving it to GAP how to delegate questions, we could decide in Oscar to which group (in GAP or Oscar) questions are delegated; we could introduce `G.nice_monomorphism` for that, analogous to the siuation in GAP.

- The code in `experimental/Matrix` wraps the Oscar matrices in GAP objects and deals with finite GAP matrix groups of them, via `NiceMonomorphism`s that are computed on the Oscar side (using the `_isomorphic_group_over_finite_field` approach).
  This is currently limited by technical difficulties (GAP wants to compare the matrices with `<`) and algorithmic problems (how to compute the Oscar matrix that is the preimage of a given GAP matrix -- the generic GAP code for that makes sense only if the group is small).

Currently we do not use matrix group functionality from several GAP packages.

- [Polenta](https://gap-packages.github.io/polenta/):
  Compute a polycyclic presentation of the matrix group and then delegate questions to the polycyclic group;
[...]

PR #2575, opened 2023-07-19
Avoid some explicit distinction of characteristic in the invariant theory code

joschmitt:
... as these are by now automatically (and more reliably) handled by the GAP interface.

### DIFF
--- a/src/InvariantTheory/invariant_rings.jl
+++ b/src/InvariantTheory/invariant_rings.jl
@@ -483,17 +483,12 @@ basis(IR::InvRing, d::Int, chi::GAPGroupClassFunction) = collect(iterate_basis(I
 function _molien_series_char0(S::PolyRing, I::InvRing)
   G = group(I)
   n = degree(G)
-  if G isa MatrixGroup{T, T1} where T1<:MatElem{T} where T<:Union{QQFieldElem, ZZRingElem, nf_elem}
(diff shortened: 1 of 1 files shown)


thofma:
Is `isomorphic_group_over_finite_field` still used for something?

joschmitt:
> Is `isomorphic_group_over_finite_field` still used for something?

Yes, to compute the order. I think, this is in the scope of #2093 .

ThomasBreuer (approved):
I think that this proposal "is in the scope of #2093" (which was mentioned in the discussion), since it replaces the decision to use `isomorphic_group_over_finite_field` by delegating the decision to the given group `G` how to handle tasks such as `conjugacy_classes(G)`.

The proposed changes are in principle correct, in the sense that not each function for a matrix group `G` in characteristic zero should have to decide how to proceed with `G`.

(The statement that `isomorphic_group_over_finite_field` is used to compute the group order is not obvious. If the group `G` does not know its order before the call of `_molien_series_char0` then calling `conjugacy_classes(G)` will delegate to GAP, which computes the group order and more, such that the later call `order(ZZRingElem, G)` will decide that it is better to call `ZZRingElem(GAPWrap.Size(G.X))` than `order(isomorphic_group_over_finite_field(G)[1])`.)

PR #2577, opened 2023-07-19
Elliptic surfaces and rational maps

HechtiDerLachs:
This introduces rational maps and some functionality for them.


HechtiDerLachs:
This is a state-of-the-art version of our `RationalMap`s. Unfortunately, I could not push it to the point that pullback of divisors works as smoothly as desired. But I would nevertheless advocate for a merge in order to not have this as a loose end. The functionality is tested for the Cremona transformation and none of the things is exported. So we should have enough freedom to still change things later on.

simonbrandhorst (approved):
Thank you for preparing this @HechtiDerLachs 
I agree that this code needs more comments and documentation.
Lest it rots in the next weeks, I merge anyways to experimental.

PR #2583, opened 2023-07-23
Fix formatting error in docstring note

taboege:
There is a formatting error in the docstring `!!! note` from #2493 causing the note to be rendered outside of the note box, see https://docs.oscar-system.org/dev/CommutativeAlgebra/ideals/#Degree


thofma (approved):

benlorenz (approved):
Thanks!

PR #2585, opened 2023-07-24
Fix printing for `EllipticSurface`

### LINKED ISSUE #2584: Check for unnecessary uses of underlying_scheme
**Describe the bug**
The show method for EllipticSurface can be extremely slow:
https://github.com/oscar-system/Oscar.jl/blob/b7b63dce3070ed7c196600dbba86e5aa5a4f9b8d/experimental/Schemes/elliptic_surface.jl#L238-L269

**To Reproduce**
Run the code from the doctest in `elliptic_surface.jl:241-253`:
```julia
julia> Qt, t = polynomial_ring(QQ, :t);

julia> Qtf = fraction_field(Qt);

julia> E = elliptic_curve(Qtf, [0,0,0,0,t^5*(t-1)^2]);

julia> X3 = elliptic_surface(E, 2)
Elliptic surface
  over rational field
with generic fiber
  -x^3 + y^2 - t^7 + 2*t^6 - t^5
and relatively minimal model
  scheme over QQ covered with 45 patches
```
The last step takes about 200 seconds including compilation time. After recreating the object it takes "only" about 60 seconds (i.e. excluding compilation time)

More precisely it happens due to `show` calling `base_ring`:
```julia
julia> Qt, t = polynomial_ring(QQ, :t);

julia> Qtf = fraction_field(Qt);

julia> E = elliptic_curve(Qtf, [0,0,0,0,t^5*(t-1)^2]);

julia> X3 = elliptic_surface(E, 2);

julia> @time base_ring(X3)
213.849025 seconds (486.34 M allocations: 23.844 GiB, 6.70% gc time, 62.53% compilation time)
Rational field

julia> X3 = elliptic_surface(E, 2);

julia> @time base_ring(X3)
 77.345964 seconds (331.51 M allocations: 13.487 GiB, 14.47% gc time, 0.03% compilation time)
Rational field
```

**Expected behavior**
If this method can take that long it should not be used in the show method and maybe not in the doctests as well. Maybe there is an easier / faster way to determine the base_ring ?

Also I don't think it is a good idea to attach an example to the docstring of `Base.show`, I can't find it in our documentation.

This was added in #2580, cc: @StevellM

StevellM:
A mistake of mine, I have not checked that `base_ring` was calling `underlying_scheme` which is in general expensive to compute for elliptic surfaces. Same with the access to the relatively minimal model: the way I wrote it, I was not accessing the relatively minimal model already computed, but I was asking to compute it again.

This should resolve #2584, @benlorenz  I also moved the doctest.

```
julia> Qt, t = QQ["t"];

julia> Qtf = fraction_field(Qt);

julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);

julia> X3 = elliptic_surface(E,2);

julia> @time base_ring(X3)
  0.015081 seconds (2.63 k allocations: 157.180 KiB, 98.94% compilation time)
Rational field

julia> @time base_ring(X3)
  0.000013 seconds
Rational field

julia> @time Base.show(stdout, MIME"text/plain"(), X3)
Elliptic surface
  over rational field
with generic fiber
  -x^3 + y^2 - t^7 + 2*t^6 - t^5  0.240701 seconds (296.77 k allocations: 15.725 MiB, 99.66% compilation time)
0

julia> @time Base.show(stdout, MIME"text/plain"(), X3)
Elliptic surface
  over rational field
with generic fiber
  -x^3 + y^2 - t^7 + 2*t^6 - t^5  0.000507 seconds (667 allocations: 32.042 KiB)
0

```

### DIFF
--- a/experimental/Schemes/elliptic_surface.jl
+++ b/experimental/Schemes/elliptic_surface.jl
@@ -47,10 +47,32 @@ For now functionality is restricted to $C = \mathbb{P}^1$.
 
 end
 
+base_ring(X::EllipticSurface) = coefficient_ring(base_ring(base_field(generic_fiber(X))))
+
 @doc raw"""
     elliptic_surface(generic_fiber::EllCrv, euler_characteristic::Int, mwl_basis::Vector{<:EllCrvPt}=EllCrvPt[]) -> EllipticSurface
 
 Return the relatively minimal elliptic surface with generic fiber ``E/k(t)``.
+
+# Examples
+```
+julia> Qt, t = polynomial_ring(QQ, :t);
+
+julia> Qtf = fraction_field(Qt);
+
+julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);
+
+julia> X3 = elliptic_surface(E, 2)
+Elliptic surface
+  over rational field
+with generic fiber
+  -x^3 + y^2 - t^7 + 2*t^6 - t^5
+and relatively minimal model
+  scheme over QQ covered with 45 patches
+
+julia> Base.show(stdout, X3)
+Elliptic surface with generic fiber -x^3 + y^2 - t^7 + 2*t^6 - t^5
+```
 """
 function elliptic_surface(generic_fiber::EllCrv{BaseField}, euler_characteristic::Int,
                           mwl_basis::Vector{<:EllCrvPt}=EllCrvPt[]) where {
@@ -208,23 +230,6 @@ Return the torsion part of the Mordell-Weil group of the generic fiber of ``S``.
   return tors
 end
 
-@doc raw"""
-
-#Examples
-```jldoctest
-julia> Qt, t = polynomial_ring(QQ, :t);
-
-julia> Qtf = fraction_field(Qt);
-
-julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);
-
-julia> X3 = elliptic_surface(E, 2);
-
-julia> Base.show(stdout, X3)
-Elliptic surface with generic fiber -x^3 + y^2 - t^7 + 2*t^6 - t^5
-
-```
-"""
 function Base.show(io::IO, S::EllipticSurface)
   io = pretty(io)
   if get(io, :supercompact, false)
@@ -235,25 +240,6 @@ function Base.show(io::IO, S::EllipticSurface)
   end
 end
 
-@doc raw"""
-#Examples
-```jldoctest
-julia> Qt, t = polynomial_ring(QQ, :t);
-
-julia> Qtf = fraction_field(Qt);
-
-julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);
-
-julia> X3 = elliptic_surface(E, 2)
-Elliptic surface
-  over rational field
-with generic fiber
-  -x^3 + y^2 - t^7 + 2*t^6 - t^5
-and relatively minimal model
-  scheme over QQ covered with 45 patches
-
-```
-"""
 function Base.show(io::IO, ::MIME"text/plain", S::EllipticSurface)
   io = pretty(io)
   println(io, "Elliptic surface")
@@ -263,7 +249,7 @@ function Base.show(io::IO, ::MIME"text/plain", S::EllipticSurface)
   if isdefined(S, :Y)
     println(io)
     println(io, "and relatively minimal model")
-    print(io, Indent(), Lowercase(), relatively_minimal_model(S)[1], Dedent())
+    print(io, Indent(), Lowercase(), S.Y, Dedent())
   end
   print(io, Dedent())
 end
--- a/experimental/Schemes/elliptic_surface.jl
+++ b/experimental/Schemes/elliptic_surface.jl
@@ -55,7 +55,7 @@ base_ring(X::EllipticSurface) = coefficient_ring(base_ring(base_field(generic_fi
 Return the relatively minimal elliptic surface with generic fiber ``E/k(t)``.
 
 # Examples
-```
+```jldoctest
 julia> Qt, t = polynomial_ring(QQ, :t);
 
 julia> Qtf = fraction_field(Qt);
--- a/experimental/Schemes/elliptic_surface.jl
+++ b/experimental/Schemes/elliptic_surface.jl
@@ -67,8 +67,6 @@ Elliptic surface
   over rational field
 with generic fiber
   -x^3 + y^2 - t^7 + 2*t^6 - t^5
-and relatively minimal model
-  scheme over QQ covered with 45 patches
 
 julia> Base.show(stdout, X3)
 Elliptic surface with generic fiber -x^3 + y^2 - t^7 + 2*t^6 - t^5


experimental/Schemes/elliptic_surface.jl
```diff
@@ -47,10 +47,30 @@ For now functionality is restricted to $C = \mathbb{P}^1$.
...
 
+base_ring(X::EllipticSurface) = coefficient_ring(base_ring(base_field(generic_fiber(X))))
+
 @doc raw"""
     elliptic_surface(generic_fiber::EllCrv, euler_characteristic::Int, mwl_basis::Vector{<:EllCrvPt}=EllCrvPt[]) -> EllipticSurface
 
 Return the relatively minimal elliptic surface with generic fiber ``E/k(t)``.
+
+# Examples
+```
```
benlorenz:
```suggestion
```jldoctest
```

experimental/Schemes/elliptic_surface.jl
```diff
@@ -47,10 +47,30 @@ For now functionality is restricted to $C = \mathbb{P}^1$.
...
+julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);
+
+julia> X3 = elliptic_surface(E, 2)
+Elliptic surface
+  over rational field
+with generic fiber
+  -x^3 + y^2 - t^7 + 2*t^6 - t^5
+
+julia> Base.show(stdout, X3)
+Elliptic surface with generic fiber -x^3 + y^2 - t^7 + 2*t^6 - t^5
```
simonbrandhorst, line 72: I disagree with @benlorenz  here and would move these two lines to the doctest of the show method.

benlorenz: I don't understand why this should be in the docstring of the show method.
- Is it supposed to be visible to the user somewhere? Then the Base.show method is not the right place as no-one will find it there as it is not included in our documentation.
- Is it supposed to be hidden from the user then put it in a normal test file, print to some buffer and compare the output. (Please check `test/Groups/group_characters.jl` for some advanced printing tests)

It does show if one wants to know how `Base.show` works in the REPL:
```julia
help?> Base.show
  show([io::IO = stdout], x)

  Write a text representation of a value x to the output stream io. New types T should overload
  show(io::IO, x::T). The representation used by show generally includes Julia-specific formatting and
  type information, and should be parseable Julia code when possible.

  repr returns the output of show as a string.

  For a more verbose human-readable text output for objects of type T, define show(io::IO,
  ::MIME"text/plain", ::T) in addition. Checking the :compact IOContext key (often checked as get(io,
  :compact, false)::Bool) of io in such methods is recommended, since some containers show their
  elements by calling this method with :compact => true.

  See also print, which writes un-decorated representations.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> show("Hello World!")
  "Hello World!"
  julia> print("Hello World!")
  Hello World!

  ─────────────────────────────────────────────────────────────────────────────────────────────────────

  show(io::IO, mime, x)

  The display functions ultimately call show in order to write an object x as a given mime type to a
  given I/O stream io (usually a memory buffer), if possible. In order to provide a rich multimedia
  representation of a user-defined type T, it is only necessary to define a new show method for T, via:
  show(io, ::MIME"mime", x::T) = ..., where mime is a MIME-type string and the function body calls
  write (or similar) to write that representation of x to io. (Note that the MIME"" notation only
  supports literal strings; to construct MIME types in a more flexible manner use MIME{Symbol("")}.)

  For example, if you define a MyImage type and know how to write it to a PNG file, you could define a
  function show(io, ::MIME"image/png", x::MyImage) = ... to allow your images to be displayed on any
  PNG-capable AbstractDisplay (such as IJulia). As usual, be sure to import Base.show in order to add
  new methods to the built-in Julia function show.

  Technically, the MIME"mime" macro defines a singleton type for the given mime string, which allows us
  to exploit Julia's dispatch mechanisms in determining how to display objects of any given type.

  The default MIME type is MIME"text/plain". There is a fallback definition for text/plain output that
  calls show with 2 arguments, so it is not always necessary to add a method for that case. If a type
  benefits from custom human-readable output though, show(::IO, ::MIME"text/plain", ::T) should be
  defined. For example, the Day type uses 1 day as the output for the text/plain MIME type, and Day(1)
  as the output of 2-argument show.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> struct Day
             n::Int
         end
  
  julia> Base.show(io::IO, ::MIME"text/plain", d::Day) = print(io, d.n, " day")
  
  julia> Day(1)
  1 day

  Container types generally implement 3-argument show by calling show(io, MIME"text/plain"(), x) for
  elements x, with :compact => true set in an IOContext passed as the first argument.

  ─────────────────────────────────────────────────────────────────────────────────────────────────────

  julia> R, (x, y) = polynomial_ring(QQ, ["x", "y"]);
  
  julia> S = complement_of_point_ideal(R,[1,2]);
  
  julia> SinvR = localization(R,S)[1]
  Localization
    of multivariate polynomial ring in 2 variables over QQ
    at complement of maximal ideal of point (1, 2)
  
  julia> I = ideal(SinvR, gens(SinvR))
  Ideal
    of localized ring
  with 2 generators
    x
    y
  
  julia> (I,)
  (ideal(x, y),)
  

  ─────────────────────────────────────────────────────────────────────────────────────────────────────

  julia> R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"]);
  
  julia> I = ideal(R, [y-x^2, z-x^3]);
  
  julia> RQ, p = quo(R, I);
  
  julia> UR = complement_of_point_ideal(R, [0, 0, 0]);
  
  julia> RQL, _ = localization(RQ, UR);
  
  julia> T, (t,) =  polynomial_ring(QQ, ["t"]);
  
  julia> UT = complement_of_point_ideal(T, [0]);
  
  julia> TL, _ =  localization(T, UT);
  
  julia> PSI = hom(TL, RQL, RQL.([x]))
  Ring homomorphism
    from localization of multivariate polynomial ring in 1 variable over QQ at complement of maximal ideal
    to   localization of quotient of multivariate polynomial ring at complement of maximal ideal
  defined by
    t -> x
  
  julia> (PSI, )
  (hom: Localized ring -> Localized quotient of multivariate polynomial ring,)
  

  ─────────────────────────────────────────────────────────────────────────────────────────────────────

  #Examples

  julia> Qt, t = polynomial_ring(QQ, :t);
  
  julia> Qtf = fraction_field(Qt);
  
  julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);
  
  julia> X3 = elliptic_surface(E, 2);
  
  julia> Base.show(stdout, X3)
  Elliptic surface with generic fiber -x^3 + y^2 - t^7 + 2*t^6 - t^5
  

  ─────────────────────────────────────────────────────────────────────────────────────────────────────

  #Examples

  julia> Qt, t = polynomial_ring(QQ, :t);
  
  julia> Qtf = fraction_field(Qt);
  
  julia> E = EllipticCurve(Qtf, [0,0,0,0,t^5*(t-1)^2]);
  
  julia> X3 = elliptic_surface(E, 2)
  Elliptic surface
    over rational field
  with generic fiber
    -x^3 + y^2 - t^7 + 2*t^6 - t^5
  and relatively minimal model
    scheme over QQ covered with 45 patches
```
And here it is mostly cluttering the output and without any explanation why it is there. Together with some other examples that also should not be there.

simonbrandhorst:
It is supposed to be hidden from the user. Thanks for the pointer to the groups.

simonbrandhorst:
The advantage of putting it in a doctest is that we can do fix-doctest if the printing changes somewhere.
If it is hardcoded in the testfile, it is much more work to change the printing. 
Imagine we have hundreds of tests like this. It would be almost impossible to change anything. 
We could also add it to some dummy function, if we do not want to clutter the documentation of the show method?

thofma:
The doctests can also go directly in the markdown files in `Oscar/docs/src`. (They need not be attached to a function/method.)

simonbrandhorst:
would they show up in the documentation in this case?

thofma:
Yes, see for example https://github.com/oscar-system/Oscar.jl/blob/master/docs/src/Combinatorics/partitions.md

benlorenz:
> The advantage of putting it in a doctest is that we can do fix-doctest if the printing changes somewhere.

But it is also very easy to miss problems if there are many doctests changed by such an operation.

> If it is hardcoded in the testfile, it is much more work to change the printing.

Preferrably the test should just check that it prints the correct contents, but for example checking how the polynomial prints does not need to be in the test. This way the test does not need to be changed when the polynomial prints differently.
For the compact printing we could check:
```
@test contains(str, "Elliptic surface with generic fiber")
@test countlines(IOBuffer(str)) == 1
```
This would make it more robust to unrelated changes.

Also the doctests are only executed on some julia versions, so we don't even know if printing works on julia 1.6 right now.


> Imagine we have hundreds of tests like this. It would be almost impossible to change anything.

I really hope we will never need hundreds of tests like this.

simonbrandhorst:
I would rather say that such tests are desirable because they make sure that printing works as intended?

simonbrandhorst (approved):
I think this PR is certainly an improvement and so I am o.k. with merging it as is.

benlorenz:
> I would rather say that such tests are desirable because they make sure that printing works as intended?

To some extent yes, but the more we print nested objects in some kind of nested form we need to make sure that changing the output for the inner object does not fail too many other tests.
Doctest with `:fix` should in my opinion be used only very rarely.

For the test of `EllipticSurface` we should be able to assume that a polynomial will print properly in a compact form (i.e. not more than one line, however that polynomial looks like should not matter for this test). There should be a separate test for the polynomial somewhere else.

benlorenz (approved):

PR #2586, opened 2023-07-24
1507 revise serialization code to avoid constructing a full internal tree representation

### LINKED ISSUE #1507: Revise serialization code to avoid constructing a full internal tree representation

antonydellavecchia:
Refactoring of save_internals and save_type_dispatch

### DIFF
--- a/src/Serialization/Fields.jl
+++ b/src/Serialization/Fields.jl
@@ -41,14 +41,12 @@ end
 
 ################################################################################
 # non-ZZRingElem variant
-@registerSerializationType(Nemo.fpFieldElem)
+@registerSerializationType(fpFieldElem)
 @registerSerializationType(Nemo.fpField)
 has_elem_basic_encoding(obj::Nemo.fpField) = true
 
 function save_internal(s::SerializerState, F::Nemo.fpField)
-    return Dict(
-        :characteristic => UInt64(characteristic(F))
-    )
+    save_type_dispatch(s, UInt64(characteristic(F)), :characteristic)
 end
 
 function load_internal(s::DeserializerState, ::Type{Nemo.fpField}, dict::Dict)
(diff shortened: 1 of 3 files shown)


antonydellavecchia:
see #2687

Issue #2531, opened 2023-07-07
decide_du_val_singularity: Ideal does not describe a single K-point


benlorenz:
**Describe the bug**
Infrequent bug during the tests after the merge of #2468 by @afkafkafk13, observed on [julia 1.8 (ubuntu)](https://github.com/oscar-system/Oscar.jl/actions/runs/5481044986/jobs/9984898081?pr=2528#step:6:1648) and [julia nightly (ubuntu)](https://github.com/oscar-system/Oscar.jl/actions/runs/5478636835/jobs/9979386953#step:6:1611)

**To Reproduce**
Maybe run the tests repeatedly?

Full error message:
```
du Val Tester: Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/AlgebraicGeometry/Schemes/duValSing.jl:13
  Test threw exception
  Expression: ((decide_du_val_singularity(X, J2))[1])[3] == (:A, 3)
  Ideal does not describe a single K-point
  Stacktrace:
   [1] error(s::String)
     @ Base ./error.jl:35
   [2] rational_point_coordinates(I::MPolyIdeal{AbstractAlgebra.Generic.MPoly{nf_elem}})
     @ Oscar ~/work/Oscar.jl/Oscar.jl/experimental/Schemes/SpaceGerms.jl:128
   [3] _check_duval_at_point(IX::MPolyIdeal{AbstractAlgebra.Generic.MPoly{nf_elem}}, Ipt::MPolyIdeal{AbstractAlgebra.Generic.MPoly{nf_elem}})
     @ Oscar ~/work/Oscar.jl/Oscar.jl/experimental/Schemes/duValSing.jl:219
   [4] decide_du_val_singularity(X::Spec{AnticNumberField, MPolyQuoRing{AbstractAlgebra.Generic.MPoly{nf_elem}}}, I::MPolyIdeal{AbstractAlgebra.Generic.MPoly{nf_elem}})
     @ Oscar ~/work/Oscar.jl/Oscar.jl/experimental/Schemes/duValSing.jl:173
   [5] decide_du_val_singularity(X::Spec{QQField, MPolyQuoRing{QQMPolyRingElem}}, I::MPolyIdeal{QQMPolyRingElem})
     @ Oscar ~/work/Oscar.jl/Oscar.jl/experimental/Schemes/duValSing.jl:185
   [6] macro expansion
[...]

Issue #1380, opened 2022-06-09
Consistency: add `snake_case` aliases for all (? most?)  `CamelCaseConstructors`


fingolfin:
For end users, it is confusing. When to write `MatrixGroup`, and when `matrix_group`? When `CyclotomicField` and when `cyclotomic_field` ? Same for ideal, matrix, and many others.

Of course "experts" know that `CamelCase` indicates a typename and hence the name of a constructor. But even for them it can be challenge to know when to use which, esp. in the many cases were already both names exist, with a few "shared" methods but each having separate methods as well....

I discussed this with @fieker and we think that we should shield end users from having to learn complicated rules for these things as much as possible. Of course there are things where you just have to deal with types, and that's fine. But where it is possible to avoid the ambiguity, we should do so by providing the `snake_case` name, too.

Moreover, this principle should be explicitly documented in the styleguide section of the Oscar manual.

Here is an incomplete list of type names where this should be addressed, to be extended
- [x] `CycloctomicField`
- [x] `Ideal`
- [x] `MatrixGroup`
- [x] `PermGroup`
- [x] `PolynomialRing`

thofma:
I agree and just want to note that for some of these cases we can do aliases, but I would not want to do it for the types. I don't think we want `matrix_group === MatrixGroup`. (While we are at it, it might be a good idea to untangle the type constructors and the user facing constructors).

ulthiel:
As someone who's trying to write a package using Oscar which implements the creation of an Oscar matrix group from data of a different package (in this case Gapjm): is it my duty to implement both MatrixGroup(X::Gapjm.PRG) and matrix_group(X::Gapjm.PRG) as well? Of course it's my package so I decide what to do but I'd follow Oscar guidelines and then I feel it's annoying implementing both all the time. Also, this feels like a general Julia problem. Isn't there a guideline/solution already?

thofma:
The way the aliases are setup, you can do both. But I think we should suggest overloading not the alias version. I hope that we are moving to a `snake_case` world (except for type names), so that in the future the rule is to always use `snake_case` version.

P.S.: The aliases are identical as methods (`===`). If you try to define both, you get a `Method overwritten` warning (not in the REPL, but in a module).

lgoettgens:
`FiniteField` has no snake_case equivalent like `finite_field` as well.

fingolfin:
[...]

PR #2588, opened 2023-07-25
Resolve two small issues regarding direct products of groups

### LINKED ISSUE #389: Fix `projection(G::DirectProductGroup, j::Base.Integer)` to correctly restrict the homomorphisms
... instead of using a `GroupHomomorphismByFunction`, which looses important information e.g. for computing preimages.

### LINKED ISSUE #1717: Replace `as_perm_group` and `as_polycyclic_group` by methods for `PermGroup` and `PcGroup`
Another thing you could work on (maybe not in this PR but a follow-up): replace `as_perm_group` and `as_polycyclic_group` by methods for `PermGroup` and `PcGroup`. That would fit better with how we do this elsewhere! Also, a similar handling but for `FpGroup` seems to be missing.

_Originally posted by @fingolfin in https://github.com/oscar-system/Oscar.jl/pull/1661#discussion_r1018851683_

joschmitt:
Resolves #389 and resolves #1717 .


lgoettgens:
#1717 talks about adding `PermGroup` methods, is your use of `perm_group` motivated by https://github.com/oscar-system/Oscar.jl/issues/1380? I am asking since I cannot find any other dispatch of `perm_group`, and I would find `PermGroup` more logical to use as it is intuitively a cast.

> Also, a similar handling but for FpGroup seems to be missing.
From #1717

It would be great if you could add this as well (if it is not too involved). Or alternatively keep #1717 open and add a comment that this is still missing.

joschmitt:
> #1717 talks about adding `PermGroup` methods, is your use of `perm_group` motivated by #1380? I am asking since I cannot find any other dispatch of `perm_group`, and I would find `PermGroup` more logical to use as it is intuitively a cast.

I implemented the methods `permutation_group` (!) and `pc_group` because these already exist and one shouldn't use camel case except for types as far as I understand it. But I don't insist on the names; if people want different ones, so be it.

> > Also, a similar handling but for FpGroup seems to be missing.
> > From #1717
> 
> It would be great if you could add this as well (if it is not too involved). Or alternatively keep #1717 open and add a comment that this is still missing.

I missed that line in the issue. I added the method for  `FPGroup`s now.

lgoettgens:
> > #1717 talks about adding `PermGroup` methods, is your use of `perm_group` motivated by #1380? I am asking since I cannot find any other dispatch of `perm_group`, and I would find `PermGroup` more logical to use as it is intuitively a cast.
> 
> I implemented the methods `permutation_group` (!) and `pc_group` because these already exist and one shouldn't use camel case except for types as far as I understand it. But I don't insist on the names; if people want different ones, so be it.
 
Oh, I am sorry for the confusion. It's just too late for me, I should review stuff when I am more awake. *sigh*

Everything looks fine as is.

And thanks for adding the pc group thing.

joschmitt:
The test failure on 1.9 is #2531 ...

ThomasBreuer (approved):
The renamings are definitely an improvement.

The changes in the `projection` methods mean that GAP will use different strategies to compute images and preimages under the maps in question.
Concerning the computation of preimages, the changes are clearly an improvement, because GAP creates an auxiliary "by images" map in the case of the old `GroupHomomorphismByFunction`, and the proposed `RestrictedMapping` approach does this already.
Concerning the computation of images, the situation is not so clear; the `GroupHomomorphismByFunction` just uses the stored projection map for the full group (plus a function call), whereas the map constructed with `RestrictedMapping` has to work harder because is does not know that it comes from a projection.
I cannot say which approach is really better.

src/Groups/directproducts.jl
```diff
@@ -171,25 +171,36 @@ DirectProduct of
  Sym( [ 1 .. 3 ] )
  Sym( [ 1 .. 2 ] )
 
-julia> as_perm_group(G)
+julia> permutation_group(G)
 Group([ (1,2,3), (1,2), (4,5) ])
 ```
 """
-function as_perm_group(G::DirectProductGroup)
+function permutation_group(G::DirectProductGroup)
```
fingolfin: So far we used `PermGroup(g)` to convert a given group `g` into a `PermGroup`, quite intentionally using the group name.

I'd also be fine with switching that to e.g. `permutation_group`, but then it should be done uniformly. Right now, with this PR we'd be in a situation were converting a PcGroup to a PermGroup works via `PermGroup`, while converting a direct product group to a `PermGroup` would require `permutation_group`. That's very confusing.

Note that the existing `permutation_group` command does something quite different (it takes a degree and a bunch of permutations). Oh, and then there are `permutation_group` methods in `experimental` which are even different... let's ignore those..

(The same of course also applies to `pc_group` versus `PcGroup`.)



*independent* of that, it actually seems these conversions already work in `master` thanks to this generic method:
```
"""
    FPGroup(G::T) where T <: Union{GAPGroup, GrpAbFinGen}
    GrpAbFinGen(G::T) where T <: GAPGroup
    PcGroup(G::T) where T <: Union{GAPGroup, GrpAbFinGen}
    PermGroup(G::T) where T <: Union{GAPGroup, GrpAbFinGen}

Return a group of type `T` that is isomorphic with `G`.
If one needs the isomorphism then
[isomorphism(::Type{T}, G::GAPGroup) where T <: Union{FPGroup, PcGroup, PermGroup}](@ref)
can be used instead.
"""
function (::Type{S})(G::T) where {S <: Union{GrpAbFinGen, GAPGroup}, T <: GAPGroup}
   return codomain(isomorphism(S, G))
end
```
So indeed:
```
julia> H = symmetric_group(3)
Sym( [ 1 .. 3 ] )

julia> K = symmetric_group(2)
Sym( [ 1 .. 2 ] )

julia> G = direct_product(H,K)
DirectProduct of
 Sym( [ 1 .. 3 ] )
 Sym( [ 1 .. 2 ] )

julia> PermGroup(G)
Group([ (1,2,3), (1,2), (4,5) ])

julia> PcGroup(G)
Group([ f1, f2, f3 ])

julia> isomorphism(PermGroup, G)
Group homomorphism from
DirectProduct of
 Sym( [ 1 .. 3 ] )
 Sym( [ 1 .. 2 ] )
to
Group([ (1,2,3), (1,2), (4,5) ])
```


In light of this, perhaps we should just remove `as_perm_group` and `as_polycyclic_group`. And then possibly open an issue about adding `permutation_group` / `pc_group` aliases?

joschmitt:
(Sorry for rewriting git history.)

Regarding the restriction of the projection maps nothing changed.
As suggested by @fingolfin, I completely removed the `as_perm_group` and `as_pc_group` methods in favour of the available generic methods.
I also made snake case versions of the camel case constructors available; this required some work-arounds in the `GModule`s, I hope this is fine.

experimental/GModule/Cohomology.jl
```diff
@@ -402,20 +402,14 @@ Compute an fp-presentation of the group generated by 'g'
 and returns both the group and the map from the new group to the
 parent of the generators.
 """
-function fp_group(g::Vector{<:Oscar.GAPGroupElem})
+function fp_group_with_isomorphism(g::Vector{<:Oscar.GAPGroupElem})
```
fingolfin, line 487: This is not an issue with this PR, but rather with the existing code (I pointed it out to @fieker before): this function does not actually use `g`, it just uses it to get `parent(g[1])`, checks that all elements of `g` have the same parent, and then ignores `g`. It also seems to always be called with `gens(G)` as argument; so I really wonder why this function doesn't just take a `GAPGroup` as argument to begin with.

Assuming such a change, a user may then wonder what difference between `fp_group_with_isomorphism(G)` and `isomorphism(FPGroup, G)` is... apparently just the use of `IsomorphismFpGroupByGenerators` versus `IsomorphismFpGroup`. This is an important difference of course, but perhaps we can and should do better when it comes to aligning this? E.g. there could be a kwarg `preserve_generators::Bool` which indicates whether we prefer an isomorphism that "imitates" the generating sequence of the source in the domain (`true`); or would prefer an isomorphism that's as efficient as possible (`false`).

joschmitt: For completeness: another difference between `fp_group_with_isomorphism` and `isomorphism(FPGroup, ...)` is the direction of the map.

joschmitt:
I rebased now and fixed some conflicts. Could we try to get this merged before there are new conflicts?

NOTE: the approving review is unfortunately quite outdated.

joschmitt:
> I rebased now and fixed some conflicts. Could we try to get this merged before there are new conflicts?
> 
> NOTE: the approving review is unfortunately quite outdated.

I rebased again and fixed a conflict...

thofma:
@fingolfin are you happy with this?

fingolfin:
No, I habe concerns about this. I'll comment later

joschmitt:
I rebased again because of a minor conflict.
Is there any feedback? If this is not as desired, we can also just merge the commit regarding #389 or nothing or ...

joschmitt:
I took back most of the `FPGroup` -> `fp_group` etc changes now because @fingolfin explained to me that one can see `FPGroup(::GAPGroup)` as casting the group in a different type.

fingolfin (approved):

PR #2592, opened 2023-07-26
Added Laurent Mpoly elem and ideals to serialization

antonydellavecchia:
Added Laurent Mpoly elem and ideals to serialization

### DIFF
--- a/src/Serialization/Rings.jl
+++ b/src/Serialization/Rings.jl
@@ -74,8 +74,9 @@ end
 @registerSerializationType(PolyRing, true)
 @registerSerializationType(MPolyRing, true)
 @registerSerializationType(UniversalPolyRing, true)
+@registerSerializationType(AbstractAlgebra.Generic.LaurentMPolyWrapRing, true)
 
-function save_internal(s::SerializerState, R::Union{UniversalPolyRing, MPolyRing, PolyRing})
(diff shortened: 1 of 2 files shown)


thofma:
A test would be nice

micjoswig (approved):
Code looks good to me, with one exception (which is not actually about the serialization).

@thofma The name of the type AbstractAlgebra.Generic.LaurentMPolyWrapRing does not look very OSCAR-ish. Can't that be exported as LaurentMPolyRing or such?

lgoettgens:
> Code looks good to me, with one exception (which is not actually about the serialization).
> 
> @thofma The name of the type AbstractAlgebra.Generic.LaurentMPolyWrapRing does not look very OSCAR-ish. Can't that be exported as LaurentMPolyRing or such?

This has been discussed recently on slack. See https://oscar-system.slack.com/archives/C5B2H11BQ/p1690446806159369?thread_ts=1690446806.159369&cid=C5B2H11BQ
@joschmitt asked for the same, but the solution seems to be a lot more involved than expected (rename most of AbstractAlgebras structs and update everything downstream).

thofma:
Just to elaborate a bit more on what @lgoettgens said. There is already a `LaurentMPolyRing`, so we cannot just rename it.

PR #2594, opened 2023-07-27
Katsura documentation

### LINKED ISSUE #1694: Documentation: katsura
The documentation in
https://oscar-system.github.io/Oscar.jl/dev/CommutativeAlgebra/ideals/#katsura-Tuple{Int64}
could use some love. The index seems to run from `n` to `n` in the first sum and maybe an explanation what the `u_i` are and what ring the ideal returned will live in would be nice.

ederc:
This should fix #1694.


simonbrandhorst (approved):
Thank you. Looks good to me.

PR #2597, opened 2023-07-27
Saturate via cascade instead of product

JohnAAbbott:
Changed `binomial-ideals.jl` to use cascade of saturations instead of saturation by product.


src/Rings/binomial_ideals.jl
```diff
@@ -846,20 +853,22 @@ function cellular_primary_decomposition(I::MPolyIdeal{QQMPolyRingElem}, RQQAb::M
   #compute product of all non cellular variables and the product of all cell variables
   R = base_ring(I)
   Variables = gens(RQQAb)
-  prodDeltaC = elem_type(RQQAb)[Variables[i] for i = 1:nvars(R) if !(i in cell[2])]
-  prodDelta = elem_type(RQQAb)[Variables[i] for i in cell[2]]
+  varDeltaC = elem_type(RQQAb)[Variables[i]  for i = 1:nvars(R) if !(i in cell[2])]
+  varDelta = elem_type(RQQAb)[Variables[i]  for i in cell[2]]
```
lgoettgens:
```suggestion
  varDeltaC = elem_type(RQQAb)[Variables[i] for i = 1:nvars(R) if !(i in cell[2])]
  varDelta = elem_type(RQQAb)[Variables[i] for i in cell[2]]
```
too much whitespace

wdecker (approved):

src/Rings/binomial_ideals.jl
wdecker: @JohnAAbbott please commit suggestion so that we can merge

lgoettgens: I just committed the suggestion myself. So only waiting for the CI to succeed is necessary.

PR #2598, opened 2023-07-27
Bump `DocumenterCitations` to `v1.0.0`

lgoettgens:
WIP

DocumenterCitations (the package that we use for references in the documentation) has received a major update including some breaking changes and a lot of new fancy stuff.
I will first try to get it to work exactly like it does currently with the old version and then try to nice things up.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -3,4 +3,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
 
 [compat]
-DocumenterCitations = "0.2.5"
+DocumenterCitations = "1.0.0"
--- a/docs/citation_style.jl
+++ b/docs/citation_style.jl
@@ -1,0 +1,44 @@
+#
+# This file is included by docs/make_work.jl to define the custom citation style `oscar_style`.
+#
+# It is heavily based upon
+# https://juliadocs.org/DocumenterCitations.jl/v1.0.0/gallery/#Custom-style:-Citation-key-labels
+#
+
+import DocumenterCitations
+
+# we use some (undocumented) internal helper functions for formatting...
+using DocumenterCitations: format_names, tex2unicode, italicize_md_et_al
+
+const oscar_style = :oscar
+
+# The long reference string in the bibliography
+function DocumenterCitations.format_bibliography_reference(::Val{oscar_style}, entry)
+  return DocumenterCitations.format_bibliography_reference(:numeric, entry)
+end
+
+# The label in the bibliography
+function DocumenterCitations.format_bibliography_label(::Val{oscar_style}, entry, citations)
+  return "[$(entry.id)]"
+end
+
+# The order of entries in the bibliography
+DocumenterCitations.bib_sorting(::Val{oscar_style}) = :nyt  # name, year, title
+
+# The type of html tag to use for the bibliography
+DocumenterCitations.bib_html_list_style(::Val{oscar_style}) = :dl
+
+function DocumenterCitations.format_citation(
+  ::Val{oscar_style}, entry, citations; note, cite_cmd, capitalize, starred
+)
+  link_text = isnothing(note) ? "[$(entry.id)]" : "[$(entry.id), $note]"
+  if cite_cmd == :citet
+    et_al = starred ? 0 : 1  # 0: no "et al."; 1: "et al." after 1st author
+    names = tex2unicode(
+      format_names(entry; names=:lastonly, and=true, et_al, et_al_text="*et al.*")
+    )
+    capitalize && (names = uppercasefirst(names))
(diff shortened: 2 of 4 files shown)


lgoettgens:
Notable changes:

1. It now computes the reference strings (e.g. `[AL94]`) itself, instead of using the identifiers from the `oscar-bibliography.bib`.
2. The hyperlink to the DOI or URL has moved from the title to the journal info. This has been made to mimic the [APS Journals](https://journals.aps.org/).
3. There are neat commands for including a citation in a sentence, see https://juliadocs.org/DocumenterCitations.jl/stable/gallery/#alphabetic_style

Points 1. and 2. can be changed by using a custom style instead of the `:alpha` style shipped with `DocumenterCitations`.
After the discussion has been settled here, I could adapt this in a similar fashion to the other repos.

A preview should be available here: https://docs.oscar-system.org/previews/PR2598/
In particular https://docs.oscar-system.org/previews/PR2598/references/

fingolfin:
Lots of issues with this new version
- it failed to include some references, such as
  - BDEPS04
  - BES-E-D21
  - DF20
  - DLRS10
  - DHS98
  - ... I stopped writing down, but there are many more missing
- it doesn't handle multiple papers with the same author list and same year of appearance, such as BJR10 (which occurs twice)
- only the first three authors seems to matter, so e.g. BHMPW20 has become BHM20 (and many similar cases)
- BKR20 is also gone
- `OMdCS00` has become `CS00` but the last name of Ignacio is "Ojeda Martínez de Castilla" and clearly does not start with `C`, it starts with `O` (similar story for C-MLS20 and C-MLS21)
- it doesn't like Umlauts much, maybe? see `[Gr\03]
B. Grünbaum. Convex polytopes.`
- the new placement of links fails for the Atlas, i.e. WWT (formerly WWTSPNNLBA -- perhaps this is an instance where a custom name like `[Atlas]` would be more suitable?)

All in all, let's stay away from alpha mode, I am not sure they *can* fix the above issues.

Also, clearly we must figure out why it drops items.

On the upside, I do like the new styling.

(BTW I am somewhat grumpy that they did not migrate the existing repository, or turn a fork  into the the new "main" repo (GitHub will do that when asked) but instead created a new repo from scratch, thus loosing all the old issues and making it harder to discover that a move occurred unless one surfs to the old repo and reads the README *sigh*)

benlorenz:
> it failed to include some references, such as ...

There is the following entry in the NEWS file:

> Only cited references are included in the main bibliography by default, as opposed to all references defined in the underlying .bib file.

So we might need this (see [here](https://juliadocs.org/DocumenterCitations.jl/stable/syntax/#full_bibliography)):
````
```@bibliography
*
```
````

lgoettgens:
> * it failed to include some references, such as
>   
>   * BDEPS04
>   * BES-E-D21
>   * DF20
>   * DLRS10
>   * DHS98
>   * ... I stopped writing down, but there are many more missing

Those are references that are not cited anywhere. Similarly to biblatex defaults, they are thus removed from the bibliography. But if they should be included, this can be added as a one-liner.

> * it doesn't handle multiple papers with the same author list and same year of appearance, such as BJR10 (which occurs twice)

Known. See https://github.com/JuliaDocs/DocumenterCitations.jl/issues/26

> * only the first three authors seems to matter, so e.g. BHMPW20 has become BHM20 (and many similar cases)

Noticed and subject to change, see https://github.com/JuliaDocs/DocumenterCitations.jl/pull/27

> * `OMdCS00` has become `CS00` but the last name of Ignacio is "Ojeda Martínez de Castilla" and clearly does not start with `C`, it starts with `O` (similar story for C-MLS20 and C-MLS21)
> * it doesn't like Umlauts much, maybe? see `[Gr\03] B. Grünbaum. Convex polytopes.`
> * the new placement of links fails for the Atlas, i.e. WWT (formerly WWTSPNNLBA -- perhaps this is an instance where a custom name like `[Atlas]` would be more suitable?)
> 
> All in all, let's stay away from alpha mode, I am not sure they _can_ fix the above issues.

I would suggest to use the custom style teased in the docs (see https://juliadocs.org/DocumenterCitations.jl/stable/gallery/#Custom-style:-Citation-key-labels), that takes the labels from the `oscar-bibliography.bib` file as before. Then the few things above do not matter for us.

> Also, clearly we must figure out why it drops items.

See above. If we want them, I can easily add them again.

> On the upside, I do like the new styling.
> 
> (BTW I am somewhat grumpy that they did not migrate the existing repository, or turn a fork into the the new "main" repo (GitHub will do that when asked) but instead created a new repo from scratch, thus loosing all the old issues and making it harder to discover that a move occurred unless one surfs to the old repo and reads the README _sigh_)

Yeah... I only noticed when I setup a new module and forgot the compat entry for DocumenterCitations and got weird syntax errors.

lgoettgens:
This gets continued in https://github.com/oscar-system/Oscar.jl/pull/2605, as this PR is somehow broken for me.

PR #2600, opened 2023-07-28
Tweak some docstrings (mostly whitespace)

fingolfin:


experimental/IntersectionTheory/src/Main.jl
```diff
@@ -248,74 +278,92 @@ line_bundle(X::AbstractVariety, D::MPolyDecRingElem) = AbstractBundle(X, 1, 1+D[
...
 Compute the total Chern class of the tangent bundle of $X$.
 """
 total_chern_class(X::AbstractVariety) = total_chern_class(X.T)
 
 @doc raw"""
     chern_class(X::AbstractVariety, k::Int)
     chern_class(X::TnVariety, k::Int)
-Compute the $k$-th Chern class of the tangent bundle of $X$.
+C
+ompute the $k$-th Chern class of the tangent bundle of $X$.
```
afkafkafk13: seems to be an unwanted change...

fingolfin: Indeed, ouch, thanks!
```suggestion

Compute the $k$-th Chern class of the tangent bundle of $X$.
```

afkafkafk13 (approved):

PR #2602, opened 2023-07-28
moved `AlgClosureFp` from `experimental` to `src`

ThomasBreuer:
This is just part of what we need:
- caching is missing (it will be necessary to create the algebraic closure only once for each prime field),
- documentation is missing, and a wrapper function `algebraic_closure` that shall be documented,
- tests are missing, in particular `test_Field_interface` (which is currently used in the tests for `QQAbField`) reports lots of errors,
- `import`/`export` are not clean (same holds for `src/Rings/AbelianClosure.jl`).


src/Rings/AlgClosureFp.jl
```diff
@@ -1,14 +1,26 @@
...
+#
+###############################################################################
+
+# This is an implementation of the algebraic closure of finite fields,
+# which is modelled as the union of finite fields.
+
 module AlgClosureFp
 
-using Oscar
+using ..Oscar
```
fieker, line 12: That I don't know. I've always done using Oscar....

experimental/GModule/GModule.jl
```diff
@@ -878,6 +911,27 @@ function hom_base(C::T, D::T) where T <: GModule{<:Any, <:Generic.FreeModule{<:F
   return b
 end
 
+function hom_base(C::T, D::T) where T <: GModule{<:Any, <:Generic.FreeModule{<:AlgClosureElem{<:FinField}}}
```
fieker, line 914: we'll have to change the signature to replace FreeModule -> FPModule
This is done in the CF-WorkInProgress branch. Feel free to ignore this at this point

src/Rings/AlgClosureFp.jl
```diff
@@ -101,11 +115,11 @@ end
 //(a::AlgClosureElem, b::AlgClosureElem) = AlgClosureElem(op(//, a, b), parent(a))
 divexact(a::AlgClosureElem, b::AlgClosureElem) = AlgClosureElem(op(divexact, a, b), parent(a))
 ==(a::AlgClosureElem, b::AlgClosureElem) = op(==, a, b)
-Oscar.iszero(a::AlgClosureElem) = iszero(data(a))
-Oscar.isone(a::AlgClosureElem) = isone(data(a))
+iszero(a::AlgClosureElem) = iszero(data(a))
```
fieker, line 118: Does this work? as it needs to extend THE iszero...

fieker (approved):

src/Rings/AlgClosureFp.jl
ThomasBreuer: I think THE `iszero` is `Base.iszero`, and this gets imported.

(Do we have a convention whether we better import the functions that shall get extended or extend the functions by using their qualified names?)

fieker: No, but they have to be imported. I had sometimes surprising results in submodules. No preference either. If it works, I am good

PR #2605, opened 2023-07-28
Bump `DocumenterCitations` to `v1.0.0`

lgoettgens:
(Follow-up to https://github.com/oscar-system/Oscar.jl/pull/2598 being broken somehow)

DocumenterCitations (the package that we use for references in the documentation) has received a major update including some breaking changes and a lot of new fancy stuff.
I will first try to get it to work exactly like it does currently with the old version and then try to nice things up.

Preview at  https://docs.oscar-system.org/previews/PR2605/references/

### DIFF
--- a/docs/oscar_references.bib
+++ b/docs/oscar_references.bib
@@ -69,13 +69,6 @@
   pdf           = {https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=6eaea4b5c868c84295a0977ed9fce92d7c960a39}
 }
 
-@Misc{Aut00,
-  author        = {{A}uthors, {T}he {S}tacks {P}roject},
-  title         = {{S}tacks {P}roject},
-  year          = {0000},
-  url           = {https://stacks.math.columbia.edu/}
-}
-
 @Article{BBS02,
   author        = {Bayer, M. and Bruening, A. and Stewart, J.},
   title         = {A Combinatorial Study of Multiplexes and Ordinary Polytopes},
@@ -147,13 +140,13 @@
   location      = {Virtual Event, Russian Federation}
 }
 
-@Misc{BES19,
+@Article{BES19,
   author        = {Backman, Spencer and Eur, Christopher and Simpson, Connor},
   title         = {Simplicial generation of Chow rings of matroids},
-  publisher     = {arXiv},
+  journal       = {arXiv:1905.07114},
   year          = {2019},
-  doi           = {10.48550/ARXIV.1905.07114},
-  url           = {https://arxiv.org/abs/1905.07114}
+  doi           = {10.48550/arxiv.1905.07114},
+  primaryclass  = {math.CO}
 }
 
 @Book{BGV03,
@@ -190,13 +183,13 @@
   doi           = {10.1017/fms.2023.50}
 }
 
-@Misc{BHMPW20,
+@Article{BHMPW20,
   author        = {Braden, Tom and Huh, June and Matherne, Jacob P. and Proudfoot, Nicholas and Wang, Botong},
   title         = {A semi-small decomposition of the Chow ring of a matroid},
-  publisher     = {arXiv},
+  journal       = {arXiv:2002.03341},
   year          = {2020},
-  doi           = {10.48550/ARXIV.2002.03341},
-  url           = {https://arxiv.org/abs/2002.03341}
+  doi           = {10.48550/arXiv.2002.03341},
+  primaryclass  = {math.AG}
 }
 
 @Article{BJRR10,
@@ -217,6 +210,7 @@
   author        = {Blumenhagen, Ralph and Jurke, Benjamin and Rahn, Thorsten and Roschy, Helmut},
   title         = {cohomCalg package},
   note          = {High-performance line bundle cohomology computation based on BJRR10},
+  howpublished  = {Published electronically on GitHub},
   year          = {2010},
   url           = {https://github.com/BenjaminJurke/cohomCalg}
 }
@@ -477,13 +471,14 @@
   zbmath        = {404355}
 }
 
-@Misc{DF20,
+@Article{DF20,
   bibkey        = {{DF20}},
   author        = {De Franceschi, G.},
   title         = {Centralizers and conjugacy classes in finite classical groups},
-  year          = 2020,
-  doi           = {10.48550/ARXIV.2008.12651},
-  url           = {https://arxiv.org/abs/2008.12651}
+  journal       = {arXiv:2008.12651},
+  year          = {2020},
+  doi           = {10.48550/arXiv.2008.12651},
+  primaryclass  = {math.GR}
 }
 
 @Article{DFO13,
@@ -856,7 +851,7 @@
   year          = {2010},
   doi           = {10.1007/s00791-009-0127-3},
   url           = {http://dx.doi.org/10.1007/s00791-009-0127-3},
-  arxiv         = {0711.2397}
+  eprint        = {0711.2397}
 }
 
 @InProceedings{GK14,
@@ -934,7 +929,8 @@
 @Misc{Gat18,
   author        = {Gathmann, Andreas},
   title         = {Class notes „Plane Algebraic Curves” (SS 2018)},
-  year          = 2018,
+  howpublished  = {Published electronically},
+  year          = {2018},
   url           = {https://www.mathematik.uni-kl.de/~gathmann/en/curves.php}
 }
 
@@ -1213,7 +1209,6 @@
   year          = {2015},
   doi           = {10.1007/JHEP01(2015)142},
   eprint        = {1408.4808},
-  archiveprefix = {arXiv},
   primaryclass  = {hep-th},
   reportnumber  = {UPR-1264-T, CERN-PH-TH-2014-171, Bonn-TH-2014-13}
 }
@@ -1227,16 +1222,17 @@
   year          = {2011},
   doi           = {10.1007/JHEP08(2011)094},
   eprint        = {1106.3854},
-  archiveprefix = {arXiv},
   primaryclass  = {hep-th},
   reportnumber  = {UCSB-MATH-2011-09, IPMU11-0107, NSF-KITP-11-110, KCL-MTH-11-13}
 }
 
-@Misc{KO14,
-  author        = {Kelleher, J. and O'Sullivan, B.},
+@Article{KO14,
+  author        = {Kelleher, Jerome and O'Sullivan, Barry},
   title         = {Generating All Partitions: A Comparison Of Two Encodings},
+  journal       = {arXiv:0909.2331},
   year          = {2014},
-  url           = {https://arxiv.org/abs/0909.2331}
+  doi           = {10.48550/arXiv.0909.2331},
+  primaryclass  = {cs.DS}
 }
 
 @Book{KR05,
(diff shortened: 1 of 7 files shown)


benlorenz:
https://www.githubstatus.com/incidents/l59z35rhzdky

> We are investigating an issue which is causing a delay in the time it takes new commits to show up on a pull request.
> Posted Jul 28, 2023 - 14:51 UTC

lgoettgens:
> https://www.githubstatus.com/incidents/l59z35rhzdky
> 
> > We are investigating an issue which is causing a delay in the time it takes new commits to show up on a pull request.
> > Posted Jul 28, 2023 - 14:51 UTC

Thanks. That wasn't there when I checked a minuted ago.

lgoettgens:
This now furthermore includes the following changes:

1. Custom style: Use the labels from the `.bib` file as cite labels.
2. Show all references, even if they are not cited anywhere.
3. Slightly adapted some `.bib` entries that behaved weirdly.

Known problems:

1. When there is no year in a bib entry, there are additional `()` printed. See https://github.com/JuliaDocs/DocumenterCitations.jl/pull/30
2. `author` fields without names behave weirdly (see e.g. OEIS and Stacks). If someone has an idea, please let me know. Showing first names of all authors instead of initials should work, but wouldn't look as nice.

If somebody finds more broken bib entries in the bibliography overview, please lmk.

A preview should be available at https://docs.oscar-system.org/previews/PR2605/references/ once CI finishes.

Pinging @fingolfin as he motivated the above changes.

docs/src/references.md
```diff
@@ -14,4 +14,5 @@ margin-bottom: 0.75em;
 </style>
 ```
 ```@bibliography
+*
```
fingolfin, line 17: Why is this necessary?

fingolfin:
Restarted CI to see if 1.0.1/1.0.2 fix the year issue.

docs/src/references.md
lgoettgens: Without the *, only the references actually being referenced somewhere are included. With the *, the whole bib-file is printed.

lgoettgens: This was a reaction to the first bullet point of your comment https://github.com/oscar-system/Oscar.jl/pull/2598#issuecomment-1655510496

fingolfin: OK, thanks for the explanation.

fingolfin:
Documenter 1.0 has been released -- with the change in this PR, nothing blocks the CI from using Documenter 1.0, and so it does, and so things fail.

I think for now we should modify `docs/Project.toml` to restrict Documenter to 0.27 (?), and sort things out here. Then we can worry about switching to Documenter 1.0 in a separate PR?

lgoettgens:
Thanks @benlorenz! This check and duplication can be removed again once we change the Documenter compat to 1.0.0.

lgoettgens:
> Restarted CI to see if 1.0.1/1.0.2 fix the year issue.

The year issue is fixed.

OEIS and Stacks are still broken.

Furthermore, `@Article` without `journal` are printed a bit weirdly (e.g. `BES19`). This currently happens for arXiv preprints.

fingolfin:
Yeah there is a reason why preprints were marked as `@Misc` -- I also do this in "real" `.bib` files I use for writing books. What's the reason to change those to `@Article`, by the way?

docs/oscar_references.bib
```diff
@@ -1467,8 +1478,9 @@ @Misc{OEIS
   bibkey        = {OEIS},
   author        = {OEIS Foundation Inc.},
   title         = {The {O}n-{L}ine {E}ncyclopedia of {I}nteger {S}equences},
-  note          = {Published electronically at \url{http://oeis.org}},
-  year          = {2023}
+  howpublished  = {Published electronically at http://oeis.org},
+  year          = {2023},
+  url           = {http://oeis.org}
```
fingolfin: While we are at it
```suggestion
  howpublished  = {Published electronically at https://oeis.org},
  year          = {2023},
  url           = {https://oeis.org}
```

docs/oscar_references.bib
```diff
@@ -1467,8 +1478,9 @@ @Misc{OEIS
   bibkey        = {OEIS},
   author        = {OEIS Foundation Inc.},
```
fingolfin, line 1408: Does it help to `{}`-protect the entry? It would in real BibTeX, probably not here, but worth a try?
```suggestion
  author        = {{OEIS Foundation Inc.}},
```

lgoettgens:
> Yeah there is a reason why preprints were marked as `@Misc` -- I also do this in "real" `.bib` files I use for writing books. What's the reason to change those to `@Article`, by the way?

If I remember correctly, `@Misc` was formatter even weirder. I'll check again and provide a screenshot here

fingolfin:
Is there a chance we can get `@Misc` fixed either by a fix to DocumenterCitations, or some custom formatter or wahtever here?

lkastner:
What is the status of this @lgoettgens ? I need the changes by @benlorenz that make the documenter helpers work with newer versions. Should I move this to a separate PR?

lgoettgens:
> What is the status of this @lgoettgens ? I need the changes by @benlorenz that make the documenter helpers work with newer versions. Should I move this to a separate PR?

If you need it urgently, it would be best to move it to a separate PR. I will be on vacation the next week and will only find time to work on this afterwards.

fingolfin:
Just as a note relevant to the `@misc` discussion above: the DocumenterCitations documentation [says this about preprints:](https://juliadocs.org/DocumenterCitations.jl/v1.2/syntax/#Preprint-support)
> For documents that are available only as an arXiv eprint, the best result is obtained with a BibTeX entry using the `@article` class, with, e.g., `arXiv:2003.10132` in the `Journal` field, and, e.g., `10.48550/ARXIV.2003.10132` in the `Doi` field (but no `Eprint` field)

fingolfin:
Also `archiveprefix = {arXiv},` is not necessary as this is the default. Probably a good idea to read all of https://juliadocs.org/DocumenterCitations.jl/v1.2/syntax/#Preprint-support

lgoettgens:
> Just as a note relevant to the `@misc` discussion above: the DocumenterCitations documentation [says this about preprints:](https://juliadocs.org/DocumenterCitations.jl/v1.2/syntax/#Preprint-support)
> 
> > For documents that are available only as an arXiv eprint, the best result is obtained with a BibTeX entry using the `@article` class, with, e.g., `arXiv:2003.10132` in the `Journal` field, and, e.g., `10.48550/ARXIV.2003.10132` in the `Doi` field (but no `Eprint` field)

Thanks! I will look into this in the upcoming week.

docs/oscar_references.bib
```diff
@@ -25,14 +26,15 @@ @Article{ABS97
 }
 
 @Article{AG10,
-  author        = {Arzhantsev, Ivan V. and Ga\v{i}fullin, Sergei A.},
+  author        = {Arzhantsev, Ivan V. and Gaĭfullin, Sergey A.},
```
fingolfin, line 29: This change seems unrelated?

docs/oscar_references.bib
```diff
@@ -1746,6 +1679,14 @@ @Article{Sta79
   year          = {1979}
 }
 
+@Misc{Stacks,
+  bibkey        = {Stacks},
+  author        = {{A}uthors, {T}he {S}tacks {P}roject},
```
fingolfin: The braces lead to all kinds of confusion, and they are not necessary either (unlike BibTeX, there is no automatic change in capitalization being performed here).

fingolfin: The thesis by Johannes Schmitt, and the bib item right after it, also use extra braces in the title, which are rendered badly in this PR, so they should go for both

docs/oscar_references.bib
lgoettgens: Some umlaut/accent escaped were parsed/printed badly. As DocumenterCitations.jl now supports Unicode, my easiest solution was to change all escapes to Unicode symbols. I cannot remember (and didn't check) which exactly had to be changed but just redid all of them.
To ease the review, all of these changes are in well-named commits.

docs/oscar_references.bib
lgoettgens: Stacks and OEIS are still on my todo-list.
For the braces, I certainly forgot to commit yesterday 😅 thanks for pointing out

lgoettgens:
Unfortunately, non-person authors ("corporate names") are currently not supported by `BibInternal.jl` and `DocumenterCitations.jl` (waiting for https://github.com/Humans-of-Julia/BibInternal.jl/pull/21 to get released, and see https://github.com/JuliaDocs/DocumenterCitations.jl/issues/44)

The last commit here monkey-patches some functions from DocumenterCitations.jl to use different printing for `OEIS` and `Stacks` (that does not initialize first and middle names). This should only be a temporary fix until https://github.com/JuliaDocs/DocumenterCitations.jl/issues/44 is available.

lgoettgens:
I also added some guidelines on how DocumenterCitations.jl expects the `.bib` file to look like to the dev docs. 
Maybe one day I might find the time to write some bibtool test for that.

docs/oscar_references.bib
fingolfin: My concern is that this makes it even harder to extract things from the .bib file to use in LaTeX. Maybe not for those using biber, but there are still people using bibtex...

It's not a major problem for me, but it seems a bit gratuitous. Anyway, in the end, I can live with it, but I wonder whether @thofma or @fieker can...

docs/oscar_references.bib
```diff
@@ -1789,14 +1728,16 @@ @Article{Tay87
   volume        = {3},
   pages         = {76--85},
   year          = {1987},
-  url           = {https://arxiv.org/abs/2201.09155}
+  eprint        = {2201.09155},
+  primaryclass  = {math.GR}
```
fingolfin, line 1732: As far as I can tell, the `primaryclass` is only printed for a very bib entries, like this one; for most other arxiv ones it is not. Presumably it is *only* printed for those using the `eprint` field? I guess this is an artifact of the "hack" used to abuse `@article` for use in eprints. I wonder why they are not simply implement `@Misc` suitably?

docs/oscar_references.bib
lgoettgens: Each bib system has its own quirks, so one needs to adapt the bib file for each. 

The machinery available in julia cannot parse generic Latex, but only applies some limited regex replacement. If we decide against Unicode, now and for each future reference that first needs a new Unicode symbol, we need to PR a change to the DocumenterCitations regex list and wait for a release there.

docs/oscar_references.bib
lgoettgens: If we want to have them visible for arxiv only papers as well, it needs to go into the journal field. I can adapt that tomorrow

fingolfin: It is not important from my POV, I was just at first confused why it appeared for some papers but not others.

so, don't waste your time on this :-)

docs/oscar_references.bib
thofma: No objection

docs/oscar_references.bib
lgoettgens: Ok :)

fingolfin (approved):
I am mostly happy with this now, thank you very much for the effort you put into this, @lgoettgens !

One minor question: the sorting is a bit weird. It seems to  be based on the key, but case *in*sensitive? Because it puts `ZS98` before `Zie95` ? If it easily possible to make the sorting case *sensitive* would be great. If it takes a lot of effort, forget it or put it on the TODO stack somewhere far from the top.

lgoettgens:
> One minor question: the sorting is a bit weird. It seems to be based on the key, but case _in_sensitive? Because it puts `ZS98` before `Zie95` ? If it easily possible to make the sorting case _sensitive_ would be great. If it takes a lot of effort, forget it or put it on the TODO stack somewhere far from the top.

Nothing I could manage to get working in roughly 20 minutes -> maybe investigate some time way later.
What I can say: it is a bibtool setting. DocumenterCitations.jl keeps the order from the .bib file.

PR #2611, opened 2023-08-01
small corrections to Hilbert stuff

wdecker:


src/Rings/mpoly-affine-algebras.jl
```diff
@@ -112,7 +112,7 @@ function hilbert_series(A::MPolyQuoRing)
    if iszero(A.I)
       R = base_ring(A.I)
```
ederc, line 113: Don't we also need to check in case of `iszero(A.I) == true` that `base_ring(A.I)` is graded? Otherwise there is no `R.d` to use:

```
julia> R, (x,y,z) = QQ["x", "y", "z"]
(Multivariate polynomial ring in 3 variables over QQ, QQMPolyRingElem[x, y, z])

julia> A, p = quo(R, R(0), ordering=degrevlex(R))
(Quotient of multivariate polynomial ring by ideal with 1 generator, Map from
Multivariate polynomial ring in 3 variables over QQ to A defined by a julia-function with inverse)

julia> hilbert_series(A)
ERROR: type QQMPolyRing has no field d
Stacktrace:
 [1] getproperty
   @ Oscar ./Base.jl:37 [inlined]
 [2] hilbert_series(A::MPolyQuoRing{QQMPolyRingElem})
   @ Oscar ~/repos/julia-depot/dev/Oscar/src/Rings/mpoly-affine-algebras.jl:115
 [3] top-level scope
   @ REPL[28]:1
```

wdecker: @ederc  Yes, thx! In fact, there are other things as well. So please do not merge.

hannes14 (approved):

PR #2614, opened 2023-08-02
Homog ideal zm

JohnAAbbott:
Added documentation with references to the underlying maths justifying the algorithms (earlier references were incomplete/misleading)

PR #2618, opened 2023-08-03
Semicolon before kwargs

JohnAAbbott:
Ensured that all(?) calls (in `mpoly-affine-algebras.jl`) specifying kwargs have a semicolon just before the first kwarg.
Refer to discussion #2615

### DIFF
--- a/docs/src/Groups/permgroup.md
+++ b/docs/src/Groups/permgroup.md
@@ -80,6 +80,7 @@ julia> x(6)
 ## Operations for permutation groups
 
 ```@docs
+cycle_structures(G::PermGroup)
 is_transitive(G::PermGroup, L::AbstractVector{Int} = 1:degree(G))
 transitivity(G::PermGroup, L::AbstractVector{Int} = 1:degree(G))
 is_primitive(G::PermGroup, L::AbstractVector{Int} = 1:degree(G))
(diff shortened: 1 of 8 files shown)


lgoettgens:
One can additionally remove duplicate names if the name of the kwarg and the name of the variable coincide, i.e. `foo(param; something=something)` can be simplified to `foo(param; something)`. In my opinion, this makes the code even easier to read and understand (in particular it shortens lines).

src/Rings/mpoly-graded.jl
```diff
@@ -2184,66 +2188,66 @@ end
...
+  ######  There is a problem/bug with Singular.satstd: so disable this version
+  ######  IH = ideal(R, Hgens);  Y = ideal(R, prod_h_vars);
+  ######  singular_assure(IH); singular_assure(Y); return Singular.satstd(IH.gens.S, Y.gens.S)
 end
 
 
 # ============================================
 # 2023-06-30 START: New impl of homogenization
-# By John Abbott, based on K+R Book vol 2
+# By John Abbott, based on K+R Book "Computational Commutative Algebra" (see "justification" below)
```
fingolfin: This PR seems to contain other changes beyond those advertised in the title?

It also has conflicts now...

Did you generate this with a script? If so, perhaps we can just redo this cleanly from latest `master`, with the script in the commit message (handy for future repetitions of this change)

JohnAAbbott: Should I look to see how scary "resolve conflicts" is?

lgoettgens: It seems to be mostly due to https://github.com/oscar-system/Oscar.jl/pull/2630 and https://github.com/oscar-system/Oscar.jl/pull/2614.

JohnAAbbott:
Ahh!  Not sure what I did.  Hope it wasn't harmful.

PR #2619, opened 2023-08-03
[FTheoryTools] Yet more general bases

HereAround:
~~(For my convenience, includes the changes in https://github.com/oscar-system/Oscar.jl/pull/2561.)~~

### DIFF
--- a/experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
+++ b/experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
@@ -1,0 +1,119 @@
+#####################################################
+# 1 Basic attributes
+#####################################################
+
+@doc raw"""
+    coordinate_ring(f::FamilyOfSpaces)
+
+Return the coordinate ring of a generic member of the family of spaces.
+
+```jldoctest
+julia> ring, (f, g, Kbar, u) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
+
+julia> grading = [4 6 1 0]
+1×4 Matrix{Int64}:
+ 4  6  1  0
+
+julia> d = 3
+3
+
+julia> f = family_of_spaces(ring, grading, d)
+A family of spaces of dimension d = 3
+
+julia> coordinate_ring(f)
+Multivariate polynomial ring in 4 variables f, g, Kbar, u
+  over rational field
+```
+"""
+coordinate_ring(f::FamilyOfSpaces) = f.coordinate_ring
+
+
+@doc raw"""
+    weights(f::FamilyOfSpaces)
+
+Return the grading of the coordinate ring of a generic member of the family of spaces.
+
+```jldoctest
+julia> ring, (f, g, Kbar, u) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
+
+julia> grading = [4 6 1 0]
+1×4 Matrix{Int64}:
+ 4  6  1  0
+
+julia> d = 3
+3
+
+julia> f = family_of_spaces(ring, grading, d)
+A family of spaces of dimension d = 3
+
+julia> weights(f)
+1×4 Matrix{Int64}:
+ 4  6  1  0
+```
+"""
+weights(f::FamilyOfSpaces) = f.grading
+
+
+@doc raw"""
+    dim(f::FamilyOfSpaces)
+
+Return the dimension of the generic member of the family of spaces.
+
+```jldoctest
+julia> ring, (f, g, Kbar, u) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
+
+julia> grading = [4 6 1 0]
+1×4 Matrix{Int64}:
+ 4  6  1  0
+
+julia> d = 3
+3
+
+julia> f = family_of_spaces(ring, grading, d)
+A family of spaces of dimension d = 3
+
+julia> dim(f)
+3
+```
+"""
+dim(f::FamilyOfSpaces) = f.dim
+
+
+#####################################################
+# 2 Advanced attributes
+#####################################################
+
+@doc raw"""
+    stanley_reisner_ideal(f::FamilyOfSpaces)
+
+Return the equivalent of the Stanley-Reisner ideal
+for the generic member of the family of spaces.
+
+```jldoctest
+julia> coordinate_ring, (f, g, Kbar, u) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
+
+julia> grading = [4 6 1 0]
+1×4 Matrix{Int64}:
+ 4  6  1  0
+
+julia> d = 3
+3
+
+julia> f = family_of_spaces(coordinate_ring, grading, d)
+A family of spaces of dimension d = 3
+
+julia> stanley_reisner_ideal(f)
+ideal(f*g*Kbar, f*g*u, f*Kbar*u, g*Kbar*u)
+```
+"""
+@attr MPolyIdeal{QQMPolyRingElem} function stanley_reisner_ideal(f::FamilyOfSpaces)
+  ring = coordinate_ring(f)
+  variables = gens(ring)
+  combis = Oscar.combinations(length(variables), dim(f))
+  ideal_generators = [prod([variables[i] for i in c]) for c in combis]
+  return ideal(ideal_generators)
+end
--- a/experimental/FTheoryTools/src/FamilyOfSpaces/constructors.jl
+++ b/experimental/FTheoryTools/src/FamilyOfSpaces/constructors.jl
@@ -1,0 +1,14 @@
+#####################################################################
+# 1: Some simple constructors
+#####################################################################
+
+family_of_spaces(coordinate_ring::MPolyRing, grading::Matrix{Int64}, dim::Int) = FamilyOfSpaces(coordinate_ring, grading, dim)
+
+
+#####################################################################
+# 2: Display
+#####################################################################
+
+function Base.show(io::IO, f::FamilyOfSpaces)
+  print(io, "A family of spaces of dimension d = $(dim(f))")
+end
--- a/experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
+++ b/experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
@@ -146,7 +146,7 @@ hypersurface_model(base::ToricCoveredScheme, fiber_ambient_space::ToricCoveredSc
 
 
 @doc raw"""
-    hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, D1::Vector{Int64}, D2::Vector{Int64}, p::MPolyRingElem)
+    hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, D1::Vector{Int64}, D2::Vector{Int64}, p::MPolyRingElem; toric_sample = true)
 
 This method constructs a hypersurface model over a base space that is not
 fully specified. In the background, we construct an auxiliary toric base space.
@@ -211,9 +211,14 @@ julia> h = hypersurface_model(auxiliary_base_vars, auxiliary_base_grading, d, fi
 Assuming that the first row of the given grading is the grading under Kbar
 
 Hypersurface model over a not fully specified base
+
+julia> h = hypersurface_model(auxiliary_base_vars, auxiliary_base_grading, d, fiber_ambient_space, D1, D2, p; toric_sample = false)
+Assuming that the first row of the given grading is the grading under Kbar
+
+Hypersurface model over a not fully specified base
 ```
 """
-function hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, D1::Vector{Int64}, D2::Vector{Int64}, p::MPolyRingElem)
+function hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, D1::Vector{Int64}, D2::Vector{Int64}, p::MPolyRingElem; toric_sample = true)
   
   # Is there a grading [1, 0, ..., 0]?
   Kbar_grading_present = false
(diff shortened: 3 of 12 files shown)


experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
```diff
@@ -0,0 +1,119 @@
...
+#####################################################
+
+@doc raw"""
+    coordinate_ring(f::FamilyOfSpaces)
+
+Return the coordinate ring of a generic member of the family of spaces.
+
+```jldoctest
+julia> ring, (f, g, Kbar, v) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
```
apturner: There seems to be inconsistency between "u" and "v" here

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
```diff
@@ -0,0 +1,119 @@
...
+
+
+@doc raw"""
+    weights(f::FamilyOfSpaces)
+
+Return the grading of the coordinate ring of a generic member of the family of spaces.
+
+```jldoctest
+julia> ring, (f, g, Kbar, v) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
```
apturner: "u" and "v" again

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
```diff
@@ -0,0 +1,119 @@
...
+
+
+@doc raw"""
+    dim(f::FamilyOfSpaces)
+
+Return the dimension of the generic member of the family of spaces.
+
+```jldoctest
+julia> ring, (f, g, Kbar, v) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
```
apturner: "u" and "v" again

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
```diff
@@ -0,0 +1,119 @@
...
+
+@doc raw"""
+    stanley_reisner_ideal(f::FamilyOfSpaces)
+
+Return the equivalent of the Stanley-Reisner ideal
+for the generic member of the family of spaces.
+
+```jldoctest
+julia> coordinate_ring, (f, g, Kbar, v) = QQ["f", "g", "Kbar", "u"]
+(Multivariate polynomial ring in 4 variables over QQ, QQMPolyRingElem[f, g, Kbar, u])
```
apturner: "u" and "v" again

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
```diff
@@ -0,0 +1,119 @@
...
+ 4  6  1  0
+
+julia> d = 3
+3
+
+julia> f = family_of_spaces(coordinate_ring, grading, d)
+A family of spaces of dimension d = 3
+
+julia> stanley_reisner_ideal(f)
+ideal(f*g*Kbar, f*g*u, f*Kbar*u, g*Kbar*u)
```
apturner, line 110: Just flagging this as a place that may have to change to be consistent with the choice between "u" and "v" above

experimental/FTheoryTools/src/FamilyOfSpaces/constructors.jl
```diff
@@ -0,0 +1,14 @@
...
+
+family_of_spaces(coordinate_ring::MPolyRing, grading::Matrix{Int64}, dim::Int) = FamilyOfSpaces(coordinate_ring, grading, dim)
+
+
+#####################################################################
+# 2: Display
+#####################################################################
+
+function Base.show(io::IO, f::FamilyOfSpaces)
+  print(io, "A family of spaces of dimension d = $(dim(f))")
```
apturner, line 13: This doesn't have to be now: later we might want to add more info here, like the variables and the grading

experimental/FTheoryTools/src/types.jl
```diff
@@ -37,26 +50,26 @@ end
 
 
 @attributes mutable struct HypersurfaceModel <: AbstractFTheoryModel
-  base_space::AbsCoveredScheme
-  ambient_space::AbsCoveredScheme
+  base_space::FTheorySpace
```
apturner, line 53: I like the inclusion of `FTheorySpace` to cover both `AbsCoveredScheme` and `FamilyOfSpaces`; we should probably make this replacement everywhere in this file, for instance in the definition of `AbstractFTheoryModel`

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
```diff
@@ -282,7 +295,49 @@ function hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_
...
+
+function _construct_generic_sample(base_vars::Vector{String}, base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, D1::Vector{Int64}, D2::Vector{Int64}, p::MPolyRingElem)
+  # Convert Weierstrass sections into polynomials of the auxiliary base
+  base_space = family_of_spaces(PolynomialRing(QQ, base_vars, cached = false)[1], base_grading, d)
   
+  # Construct ambient space
+  ambient_space_vars = vcat(base_vars, coordinate_names(fiber_ambient_space))
+  coordinate_ring_ambient_space = PolynomialRing(QQ, ambient_space_vars, cached = false)[1]
+  w = vcat([k.coeff for k in cox_ring(fiber_ambient_space_space).d])
+  ambient_space_grading = zero_matrix(Int, nrows(base_grading)+nrows(S),ncols(base_grading)+ncols(w))
```
apturner: Perhaps I'm just missing it, but where is `S` defined within `_construct_generic_sample`? I see it defined below this chunk of code. Should this be `nrows(w)`?

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
```diff
@@ -282,7 +295,49 @@ function hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_
...
+
+
+function _construct_generic_sample(base_vars::Vector{String}, base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, D1::Vector{Int64}, D2::Vector{Int64}, p::MPolyRingElem)
+  # Convert Weierstrass sections into polynomials of the auxiliary base
+  base_space = family_of_spaces(PolynomialRing(QQ, base_vars, cached = false)[1], base_grading, d)
   
+  # Construct ambient space
+  ambient_space_vars = vcat(base_vars, coordinate_names(fiber_ambient_space))
+  coordinate_ring_ambient_space = PolynomialRing(QQ, ambient_space_vars, cached = false)[1]
+  w = vcat([k.coeff for k in cox_ring(fiber_ambient_space_space).d])
```
apturner: This has `fiber_ambient_space_space` which should be `fiber_ambient_space`

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
```diff
@@ -282,7 +295,49 @@ function hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_
...
+  end
+  for i in 1:nrows(w)
+    for j in 1:ncols(w)
+      ambient_space_grading[nrows(base_grading)+i,ncols(base_grading)+j] = w[i,j]
+    end
+  end
+  for i in 1:rows(w)
+    ambient_space_grading[i,ncols(base_grading)+1] = D1[i]
+    ambient_space_grading[i,ncols(base_grading)+2] = D2[i]
+  end
```
apturner: Should this not be
```
for i in 1:ncols(base_grading)
  ambient_space_grading[nrows(base_grading)+1, i] = D1[i]
  ambient_space_grading[nrows(base_grading)+2, i] = D2[i]
end
```

Before going through this loop, we have a matrix with a block in the top left that is the base_grading, and a block in the bottom right that is the ambient fiber space grading. What I think we want to do with this loop is modify the lower left block to include the twisting of the fiber ambient space over the base. We only need to specify the twisting of two of the fiber coordinates, which is what D1 and D2 are for, and they should have the same length as the number of columns of `base_grading`. Please correct me if I'm wrong.

Regardless, if I try to reconstruct the above three loops for the Krause, Mayrhofer, Weigand model, this last loop gives me an out-of-bounds error, so something here seems to be wrong.

apturner: Assuming my above comment is correct, I think the result of the above loops could instead be achieved as follows, starting from the definition of `w`:
```
w = Matrix{Int64}(vcat([k.coeff for k in cox_ring(fiber_ambient_space).d]))
z_block = zeros(Int64, nrows(base_grading), ncols(w))
D_block = [D1'; D2'; zeros(Int64, nrows(w) - 2, ncols(base_grading))]

ambient_space_grading = hvcat(2, base_grading, z_block, D_block, w)
```
I leave it up to you if this is better.

experimental/FTheoryTools/src/TateModels/constructors.jl
```diff
@@ -239,6 +254,37 @@ function global_tate_model(auxiliary_base_ring::MPolyRing, auxiliary_base_gradin
 end
 
 
+function _construct_generic_sample(base_grading::Matrix{Int64}, base_vars::Vector{String}, d::Int, ais::Vector{T}) where {T<:MPolyRingElem}
```
apturner: To avoid duplication of code and thus potential future errors upon edits, should we instead have a separate `_construct_toric_sample` and `_construct_generic_sample` function that each of HypersurfaceModels, TateModels, and WeierstrassModels calls upon in their own way?

apturner: Similar comments apply here as to the version of this function in HypersurfaceModels. The code replacing the loop is probably less important, since so many of the entries are set by hand in this case.

experimental/FTheoryTools/src/WeierstrassModels/constructors.jl
```diff
@@ -212,17 +226,49 @@ function weierstrass_model(auxiliary_base_ring::MPolyRing, auxiliary_base_gradin
...
+  # Construct the model
   pw = _weierstrass_polynomial(f, g, cox_ring(auxiliary_ambient_space))
   model = WeierstrassModel(f, g, pw, toric_covered_scheme(auxiliary_base_space), toric_covered_scheme(auxiliary_ambient_space))
   set_attribute!(model, :base_fully_specified, false)
   return model
 end
 
 
-#######################################
+function _construct_generic_sample(base_grading::Matrix{Int64}, base_vars::Vector{String}, d::Int, weierstrass_f::MPolyRingElem, weierstrass_g::MPolyRingElem)
```
apturner: As above

apturner (changes requested):
Thank for this @HereAround! I added a bunch of comments throughout.

Overall, I think this is a good move, to just create a simple FamilyOfSpaces type that only carries around what we need it to carry around. Ultimately, I can imagine we may shift entirely to this and drop auxiliary toric base spaces, if this ends up having all of the functionality we need.

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
apturner: If you want to be even more terse, the last line is also achieved with
```ambient_space_grading = [base_grading z_block; D_block w]```

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
HereAround: Thank you. Fixed.

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
HereAround: Thank you. Fixed.

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
HereAround: Thank you. Fixed.

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
HereAround: Thank you. Fixed.

experimental/FTheoryTools/src/FamilyOfSpaces/attributes.jl
HereAround: Thank you. Fixed.

experimental/FTheoryTools/src/FamilyOfSpaces/constructors.jl
HereAround: Agreed. Let us please postpone until later.

experimental/FTheoryTools/src/types.jl
HereAround: Agreed. Changed accordingly.

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
HereAround: Thanks. Good pointer!

I use `toric_sample = true` currently as the default option for constructions over arbitrary bases. We can eventually change this default. I reserve this for a next step/separate PR.

In any case, I of course must call the tests with `toric_sample = false` to test the new constructions. By accident however, I added tests which invoked `toric_sample = true`. And so the above did (not yet at least) lead to errors.

Have just fixed this glitch and will see how the tests "act" up as a consequence. ;)

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
HereAround: Thank you. Fixed.

HereAround:
@apturner I just attended to (most) of your comments and fixed them as you suggested. Thank you for the many pointers!

The following are still open:
- Print more details of general bases: Maybe leave it for a future PR.
- Testing if the shorter code for creating the grading matrix for hypersurface models works (running right now). If so, I will squash the current two commits into one. Otherwise, I will need to investigate more.
- Your last two comments above concern better organization of the code and avoiding code duplication. I fully agree that this should be avoided. I am waiting for the test to pass (a.k.a. that we have a working code), then I will optimize it further.

HereAround:
@apturner Regarding the above points:
 
* Printing more details of general bases: On second thought, I prefer the brief printing method. If used interactively, this "eats up" less space in the terminal. Another point to consider are the doctests. The print-out typically shows in the doctests (unless we suppress it). Hence, the more information we print, the more prone these tests are to fail if we make changes. If we truly want a method that prints all the details of these spaces, then maybe we could add a new method for this (say e.g. `full_information`) or we could pass an optional parameter. Long story short: Maybe we should discuss again (e.g. in slack) and address this point in a separate PR. (Not unrelated: There were a lot of changes on the printing a while back. I will have to refresh my mind on the new policy/guidelines.)
* Constructing the grading matrix of hypersurface models as block matrix: This seems to work as discussed on slack.
* Less code/avoiding code duplication: I will check on this again tomorrow and see what can be done.

HereAround:
@apturner I just pushed changes which aim to organize the code better.

HereAround:
@apturner I think this is ready for the next round of review. Please let me know your thoughts.

apturner (approved):

PR #2621, opened 2023-08-04
next round for `algebraic_closure`

ThomasBreuer:
This is a follow-up to #2602.
Some of the open items mentioned there are now done:
- caching,
- tests (in particular, `test_Field_interface` succeeds now; several additional methods had to be added for that),
- new function `algebraic_closure`, and
- a first bit of documentation is available
  (Is it o.k. to put the documentation behind the section on finite fields? It would be tempting to put it *into* the section about finite fields, but this is just copied from Nemo. Is this perhaps an argument to move the whole code to Nemo?)

There are still open questions:
- Currently the code is not type-stable in several places. For example, the type of the field returned by `ext_of_degree` cannot be derived from the type of the given algebraic closure. My understanding is that one wants to admit the field `F = GF(p)` as input for `algebraic_closure`, but then the extension fields cannot have the same type as `F`. Similarly, the `data` field of an `AlgClosureElem` has no concrete type.
- The `minpoly` method for `AlgClosureElem` returns a polynomial over a perhaps unsuitable field. (The underlying `minpoly` method for the finite field element chooses a field; here we should insist in a prime field that fits to the given element.)

### DIFF
--- a/docs/doc.main
+++ b/docs/doc.main
@@ -57,6 +57,7 @@
         "Nemo/qadic.md",
      ],
      "Nemo/finitefield.md",
+     "Fields/algebraic_closure_fp.md",
   ],
 
   "Linear Algebra" => [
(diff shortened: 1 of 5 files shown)


fieker (approved):
If we use fq_default, the types are stable...

Issue #2019, opened 2023-03-08
equality of Gröbner bases


wdecker:
**Describe the bug**
I am not yet sure whether this is a bug, but in the example below the
two computed Gröbner bases are not considered equal despite the
fact that there elements are pairwise equal:

**To Reproduce**
Steps to reproduce the behavior, please provide a code snippet that triggers
the bug.
```
using Oscar
# Julia code
R, (x, y, z) = polynomial_ring(GF(32003), ["x", "y", "z"]);
f1 = x^2*y+169*y^21+151*x*y*z^10;
f2 = 6*x^2*y^4+x*z^14+3*z^24;
f3 = 11*x^3+5*x*y^10*z^10+2*y^20*z^10+y^10*z^20;
I = ideal(R, [f1, f2,f3]);
W = [10, 1, 1];
GB1 = groebner_basis_hilbert_driven(I, destination_ordering = lex(R), weights = W);
S, t = polynomial_ring(ZZ, "t")
hn = -t^75 + t^54 + t^51 + t^45 - t^30 - t^24 - t^21 + 1
GB2 = groebner_basis_hilbert_driven(I, destination_ordering = lex(R), weights = W, hilbert_numerator = hn);
GB1 == GB2
for i in 1:length(GB1)
 println(GB1.gens.S[i] == GB2.gens.S[i])
end
```

wdecker:
@jankoboehm   Any idea?

fieker:
On Wed, Mar 08, 2023 at 09:55:01AM -0800, Wolfram Decker wrote:
> **Describe the bug**
> I am not yet sure whether this is a bug, but in the example below the
> two computed Gröbner bases are not considered equal despite the
> fact that there elements are pairwise equal:
> 
> **To Reproduce**
> Steps to reproduce the behavior, please provide a code snippet that triggers
> the bug.
> ```
> using Oscar
> # Julia code
> R, (x, y, z) = polynomial_ring(GF(32003), ["x", "y", "z"]);
> f1 = x^2*y+169*y^21+151*x*y*z^10;
> f2 = 6*x^2*y^4+x*z^14+3*z^24;
> f3 = 11*x^3+5*x*y^10*z^10+2*y^20*z^10+y^10*z^20;
> I = ideal(R, [f1, f2,f3]);
> W = [10, 1, 1];
> GB1 = groebner_basis_hilbert_driven(I, destination_ordering = lex(R), weights = W);
[...]

PR #2622, opened 2023-08-04
Comparison of Groebner bases

jankoboehm:
This is to address Issue #2019 about comparison of ideal generating systems and Groebner bases, along the lines discussed there.

### DIFF
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -287,23 +287,25 @@ end
 
 function show(io::IO, I::IdealGens)
   if I.isGB
-    if is_global(I.ord)
-      print(io, "Gröbner basis with elements")
-    else
-      print(io, "Standard basis with elements")
-    end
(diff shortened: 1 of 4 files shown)


src/Rings/mpoly.jl
```diff
@@ -705,6 +765,27 @@ function map_entries(R, M::Singular.smatrix)
   return matrix(S, s[1], s[2], elem_type(S)[R(M[i,j]) for i=1:s[1] for j=1:s[2]])
 end
 
+@doc raw"""
+    generating_system(I::MPolyIdeal) 
+
+Return the system of generators of `I`.
+
```
afkafkafk13, line 772: This is not just a generating system as vector of polynomials, but as an IdealGens, right?

Maybe you want to add a note, that this contains additional information on ordering and standard basis, if available. We should avoid to confuse users by giving the impression that this is just a vector of polynomials which will be compared as such.

wdecker (approved):

PR #2624, opened 2023-08-04
Free resolution fixes

ederc:
1. Removes usage of `sres` (not needed at the moment).
2. Computes a Gröbner basis of the input for `fres`.
3. Adjusts tests correspondingly.


jankoboehm:
Think, what happens is that computing the Gröbner basis before going to the resolution changes the first syzygy matrix, so the resolution does not attach to the presentation leading to homology, i.e. the presentation matrix has to be adjusted. Might occur at several places. Can sort it out on Monday.

Just as a note: We do not need the change of basis matrix between the generating system and the Gröbner basis for this.

ederc:
No longer needed due to PR #2634 .

PR #2625, opened 2023-08-04
docu `monomials_of_degree`

wdecker:


fieker (approved):

afkafkafk13:
I may be missing a point here:

This seems to cover the rather straight forward case of a graded polynomial ring -- maybe closest to the functionality of maxideal(d) in Singular. 
But I do not see the more interesting case of a quotient of a graded polynomial ring by a graded ideal -- closely related to the functionality of Singular's kbase command. In this latter case, the description of the task of finding some monomial basis for the homogeneous component of the given degree needs to be documented and implemented carefully. I expect this to be in a later PR, right?

wdecker:
@afkafkafk13 The Singular commands you mention address the standard ZZ-grading. Here we consider ARBITRARY gradings. The corresponding information for quotient rings can be easily deduced from the information provided by `monomials_of_degree`.  See the new code for `multivariate_hilbert_series`. You are right, however, in that we should extract this as a function for the user. A different function can take the roIe of kbase in the zero-dimensional kase. I will address this in a future PR. For now, this PR can be merged.

afkafkafk13:
> @afkafkafk13 The Singular commands you mention address the standard ZZ-grading. Here we consider ARBITRARY gradings. The corresponding information for quotient rings can be easily deduced from the information provided by `monomials_of_degree`. See the new code for `multivariate_hilbert_series`. You are right, however, in that we should extract this as a function for the user. A different function can take the roIe of kbase in the zero-dimensional kase. I will address this in a future PR. For now, this PR can be merged.

I completely agree that this PR should be merged and the rest is for a future PR. My reference to the Singular-commands was just to clearly point out the difference I was referring to. I am aware of the much more general grading.

PR #2627, opened 2023-08-05
Rational points of varieties/schemes

simonbrandhorst:
A first draft for rational points @JHanselman @fieker @afkafkafk13 @jankoboehm 

If this goes in the right direction, I will continue with projective points (over fields, over rings they seem to be quite complicated and not always given by a vector of coordinates in fact). 
I won't treat weighted projective points yet, because we do not have a working weighted projective space as a scheme as far as I know.

Here are some thoughts:
A k-rational point is a vector of coordinates that knows where it lives. 
Therefore it should have a parent.  This parent is an affine scheme. It can be affine space or any affine subscheme of it. For instance an elliptic curve. 

The k-point lives in affine space too. Therefore two k-points of two varieties in the same ambient affine space can be compared. But if the ambient affine spaces are different, then they compare as not equal, e.g.
```
X1 = affine_space(QQ,[:x,:y]);
X2 = affine_space(QQ,[:s,:t]);
X1([1,1]) == X1([1,1])
```
returns false because `X1` and  `X2` are not equal.
 
A k-point P of X defines a maximal ideal. Where should this ideal live? From a theoretical point of view, the coordinate ring of its parent X is a valid choice. But from a computational/practical point of view I would say in the coordinate polynomial ring of its ambient affine space is better. This is how I implemented it for the moment.

Note that the points  `[1, 1]` in affine space over QQ and the point `[1, 1]` over QQ[i] compare as not equal.
This is because their ambient affine spaces compare as not equal and consistent with the fact that we do not 
identify QQ[x,y] as a subset of QQ[i][x,y] automatically either.


src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -0,0 +1,171 @@
+@doc raw"""
+    AffineRationalPoint{CoeffType<:RingElem, ParentType<:AbsSpec}
+
+A $k$-rational point ``P`` of an affine scheme ``X``.
```
afkafkafk13: For the reader of this documentation this $k$ lacks context.

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -0,0 +1,171 @@
...
+
+# Examples
+```jldoctest
+julia> A2 = affine_space(GF(2), [:x, :y]);
+Rational point
+  of Affine 2-space over GF(2) with coordinates [x, y]
+  with coordinates (1, 0)
+
+julia> A2([1, 0])
+
```
afkafkafk13: Cut&Paste Error: misplaced output

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -0,0 +1,171 @@
...
+struct AffineRationalPoint{S<:RingElem, T<:AbsSpec} <: AbsAffineRationalPoint{S, T}
+  # not mutable since this should be efficient
+  coordinates::Vector{S}
+  parent::T
+  function AffineRationalPoint(parent::T, coordinates::Vector{S}; check::Bool=true) where {S <: RingElem, T<:AbsSpec}
+    r = new{S, T}(coordinates, parent)
+    @check begin
+      n = dim(ambient_space(parent))
+      k = base_ring(parent)
+      n == length(coordinates) || error("$(coordinates) must have length $(n)")
```
afkafkafk13: Is dim(ambient_space(parent)) safe in the localized setting or should we explicitly exclude (certain types of) localization? 
( The example which came to my mind is: localization of k[x,y,z] at the complement of the prime ideal generated by [x,y] )

afkafkafk13 (commented):
I agree with the choice of specifying coordinates in the ambient affine space. This coincides with the literature.

Overall I agree with your approach, but there are a few details which I would like to think through (and then -- in case of doubt -- ask you based on specifically constructed examples) to make sure that we do not mess up in corner cases.

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
simonbrandhorst: `ambient_space` should work always. In the example you mentioned I double checked.

afkafkafk13: I tested several corner cases and I am content, too.

simonbrandhorst:
From my point of view this is ready to go. Rational points in projective space will be a different issue.

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -169,3 +182,83 @@ function (f::AbsSpecMor)(P::AbsAffineRationalPoint)
...
+Return the Zariski tangent space of the parent of `P` at its point `P`.
+
+See also [`tangent_space(X::AbsSpec{<:Field}, P::AbsAffineRationalPoint)`](@ref)
+"""
+tangent_space(P::AbsAffineRationalPoint{<:FieldElem}) = tangent_space(parent(P), P)
+
+@doc raw"""
+    is_du_val_singularity(P::AbsAffineRationalPoint{<:Field})
+
+Return if the parent of `P` has hat most a Du Val singularity at `P`.
```
afkafkafk13: whether

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -169,3 +182,83 @@ function (f::AbsSpecMor)(P::AbsAffineRationalPoint)
...
+
+Note that this includes the case that ``P`` is a smooth point.
+"""
+is_du_val_singularity(P::AbsAffineRationalPoint{<:FieldElem}) = is_du_val_singularity(parent(P), ideal(P))
+
+@doc raw"""
+    decide_du_val_singularity(P::AbsAffineRationalPoint{<:Field})
+
+Return if the parent of `P` has a Du Val singularity at `P`.
```
afkafkafk13: whether

afkafkafk13 (approved):
Feel free to merge, as soon as the tests pass.

afkafkafk13:
P.S.: My Milnor-PR is behind yours, because I would like to also allow your rational points in the input.

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -0,0 +1,265 @@
...
+Return the Zariski tangent space of the parent of `P` at its point `P`.
+
+See also [`tangent_space(X::AbsSpec{<:Field}, P::AbsAffineRationalPoint)`](@ref)
+"""
+tangent_space(P::AbsAffineRationalPoint{<:FieldElem}) = tangent_space(parent(P), P)
+
+@doc raw"""
+    is_du_val_singularity(P::AbsAffineRationalPoint{<:Field})
+
+Return whether the parent of `P` has hat most a Du Val singularity at `P`.
```
lgoettgens: This sentence looks somehow linguistically wrong.

simonbrandhorst: Uhm, I am not a native english speaker. Do you have a suggestion?

lgoettgens:
```suggestion
Return whether the parent of `P` has at most one Du Val singularity at `P`.
```
I am not familiar with the mathematical context. But maybe something like this (if this is what you want to state here)?

afkafkafk13: mathematically correct:
Return whether the parent of `P` does not have a worse singularity than a du Val singularity at `P`.
(But this is not a good phrase for a software documentation because of the negation !!!)

Background: du Val singularities are the mildest surface singularities you can encounter. As long as you do not have worse than du Val singularities, many things behave nicely and are under control (ranging from resolution of singularities to moduli questions). Please bear in mind that literature also views a smooth point 'not worse than du Val singularity', so there need not be a singularity at `P` at all.

afkafkafk13:
Please close and reopen the PR to use the tests without the ones from duVal tests and FTheory tests affected by random error effects, which are currently being investigated.

simonbrandhorst:
After some discussion with @JHanselman we decided that point should have a parent point set. 
To get a roughly consistent interface with the schemes, one may also view a rational point as a homomorphism.
But some of the point terminology is kept.

simonbrandhorst:
Rewrite with sets of rational points.

simonbrandhorst:
Let `X` be an affine / projective scheme. Should `X([1,1])` return a rational point of `X`? Or a scheme point of `X`, i.e. a maximal ideal? I am not quite decided on this.

JHanselman:
> Let `X` be an affine / projective scheme. Should `X([1,1])` return a rational point of `X`? Or a scheme point of `X`, i.e. a maximal ideal? I am not quite decided on this.

Mathematically it should return a scheme point of X (as X is a scheme). But I think in practice (and for computational purposes) it would be easier if it returned a rational point. I think that is also how Magma does it.

afkafkafk13:
> Let `X` be an affine / projective scheme. Should `X([1,1])` return a rational point of `X`? Or a scheme point of `X`, i.e. a maximal ideal? I am not quite decided on this.

I need to e.g. to localize at a specified point or evaluate at it. @JHanselman on the other hand might have completely different operations in mind. So the 'right' choice will always be influenced by what the specific user expects. On the other hand, passage from one to the other should be available in any case and make this a minor question.

A more general remark:
Your change of viewpoint in commit [48336bc](https://github.com/oscar-system/Oscar.jl/pull/2627/commits/48336bc59f1a5186f5cde1490d4f77aaed19c5cc) is a major change to the whole notion introduced in this PR. 
Allow me to ask the standard question, which we always need to ask ourselves for design decisions in OSCAR: Which is the underlying textboot and what is closest to its spirit (mathematically/philosophically)? 
In particular: are we in the scheme world or in the elliptic curve world and how are the two planned to interact?

afkafkafk13:
I have been pondering your change of the meaing of `parent` and I need to express one doubt (for which I will be happy to hear an argument dispelling it):
Should `parent` be a simple getter without computing anything? In this case, the rational_point_set might pose some problem, because quite often you might know and want to use some rational point of a given scheme `X`, but not know the full rational point set of the (possibly higher dimensional) scheme `X`.
I agree with @simonbrandhorst and @JHanselman that it makes sense to see a rational point in the context of the rational point set. I am simply worried whether the way to access it should be a function as common as `parent` or some more specialized name.

simonbrandhorst:
The rational point set is basically an almost empty hull just like our rings, e.g. `ZZ`, or `QQ[x]`, creating them does not compute anything. I mostly did it because magma and sage do it too. 
In case one wants to iterate on the rational points points one can make 
`RationalPointSet` iterable or add `rational_points(::Scheme)`. 

- A reason to have a rational point set is to have a unique parent-child relation. 
- Another one is to allow `P`-points on a `k`-scheme `X/k` without having to compute the base changed scheme `X/P`.
- Rational points behave differently from closed points - closed points push forward; rational points pull back under base change.
- We could give the set of rational points `E(k)` of an elliptic curve `E/k` a group structure.

It would of course still be possible to work around all of this, by storing everything directly in the points/scheme.
I am not sure about the implications though and trusted on our competitors.

simonbrandhorst:
We are kind of in both worlds. Silverman in his arithmetic of elliptic curves, does it in coordinates and can hence avoid the language of schemes a k-rational point is a tuple of elements of k. Hartshorne defines them in a coordinate free fashion as a homset.

afkafkafk13:
Your argument about the point_set being designed as an empty hull dispells my doubt. Thanks a lot !!!

I am aware of the difference between rational points and closed points under base change (one of the reasons why I am happy to have access to both). 

The most important argument for the point_set for me is considering P-points of X/k without fully creating X/P. 

Summing this up: This is green light for the changed perspective from me.

-----------------------------------
Concerning the differences in literature between Silverman and Hartshorne:

As we expect to have users from both worlds, we need to cater to their different needs anyway. We will probably end up with a few extra aliases for functions to allow people from the respective other world to intuitively find in their terminology, what they are looking for. (E.g. 'codomain' is far from intuitive for someone in the terminology of Silverman.) But we can leave this for a later PR.

JHanselman (approved):

afkafkafk13 (approved):

src/AlgebraicGeometry/RationalPoint/Types.jl
```diff
@@ -0,0 +1,78 @@
...
+@doc raw"""
+    ProjectiveRationalPoint{CoeffType<:RingElem, ParentType<:AbsProjectiveScheme}
+
+Type for rational points in projective varieties.
+
+# Examples
+```jldoctest
+julia> P2 = projective_space(QQ, 2)
+
+julia> P2([4, 0 , 2//3])
```
lgoettgens, line 57: This doctest is missing the return values.

PR #2632, opened 2023-08-07
Modular Gröbner bases

ederc:
Adds modular Gröbner basis computations using Singular underneath. Initially implemented by @fieker in `experimental`, moved to `src` by @RafaelDavidMohr.

Note: This is still probabilistic and misses a certification tests, to be done as next steps.


fieker:
the runtime seems to be a 1.9.2 vs 1.9.1 problem. 1.9.1

PR #2634, opened 2023-08-07
Resfix

jankoboehm:
Fixes free resolutions.

### DIFF
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -6399,15 +6399,26 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     cc_complete = true
   end
 
-  dom = domain(maps[1])
   codom = codomain(maps[1])
-  SM = SubModuleOfFreeModule(codom, gbpres)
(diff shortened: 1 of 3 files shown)


src/Modules/UngradedModules.jl
```diff
@@ -6386,13 +6384,11 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
   singular_kernel_entry = Singular.Module(base_ring(singular_free_module),
                               [singular_free_module(repres(g)) for g in gens(kernel_entry)]...)
 
-  singular_kernel_entry.isGB = true
+  gbpres = Singular.std(singular_kernel_entry)
```
ederc: Only `fres`  and `sres` need a precomputed Gröbner basis of the input. Maybe we should not globally do this since other `algorithm`s like `mres` will be added.

jankoboehm: OK will shift the line for future cases.

hannes14 (approved):
duplicate free_resolution in ModulesGraded around line 545

PR #2635, opened 2023-08-07
Rename `monomials_of_degree` into `monomial_basis`, create related me…

wdecker:
…thods

### DIFF
--- a/src/Rings/mpoly-affine-algebras.jl
+++ b/src/Rings/mpoly-affine-algebras.jl
@@ -58,8 +58,8 @@ function vector_space_dimension(A::MPolyQuoRing)
     error("vector_space_dimension requires a coefficient ring that is a field")
   end
   I = A.I
-  @req dim(I) == 0 "The ideal must be zero-dimensional"
   G = groebner_assure(I)
+  @req dim(I) == 0 "The ideal must be zero-dimensional"
   singular_assure(G)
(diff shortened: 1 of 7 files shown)


src/Rings/mpoly-affine-algebras.jl
```diff
@@ -58,11 +58,114 @@ function vector_space_dimension(A::MPolyQuoRing)
     error("vector_space_dimension requires a coefficient ring that is a field")
   end
   I = A.I
+  @req dim(I) == 0 "The ideal must be zero-dimensional"
```
afkafkafk13: I would avoid this, as Singular's vdim reacts to non-zerodimensional ideals in a well-defined (but mathematically strange) way -- if not done already, singular.jl has to pass this on -- and we can use the return value -1 to detect the situation a posteriori. The dim call already triggers a groebner basis computation.
Moreover, just as a question: Does it make sense to throw an error for the non-zerodimensional case, if ensuring the zero-dimensional case could cost the calling function a groebner basis computation and can be detected a posteriori?

afkafkafk13: suggestion: compute a leading ideal beforehand and do the "@req" and the Singular.vdim with this leading ideal. Then you are sure to not accidentally trigger a second standard basis computation

afkafkafk13 (approved):

PR #2636, opened 2023-08-07
Fix `Oscar.build()`

### LINKED ISSUE #2631: `Oscar.build()` is broken
**Describe the bug**

Currently, `Oscar.build()` is broken, as it runs tests by simply including the `runtests.jl` file. This currently poses multiple problems : 

 - The path to tests is wrong. As the `runtests.jl` file assumes the current working directory is `Oscar/test`, we get a `No such file or directory` error if one runs `Oscar.build()` from any other working directory.
 - We have a different `Project.toml` for the tests and main Oscar. So, we get a `LoadError` when trying to run the Aqua tests, as that is a dependency only for the Oscar tests, not for Oscar itself. Thus, one needs to manually add Aqua to the current Project before `Oscar.build()` can be used.
 - We can't just include Aqua in the Oscar `Project.toml`, as that causes the Aqua test to fail.

The current workarounds to build a system image is to either activate an environment which already includes Aqua (but not as a direct dependency of Oscar), or comment out the Aqua test.

**To Reproduce**

In a clean environment (such that `Aqua.jl` is not already in the `Project.toml` file)

```
julia> using Oscar
julia> Oscar.build()
```

**Expected behavior**
All the tests should run successfully, and then a system image should be generated which includes Oscar.

aaruni96:
This commit splits the testlist into two, to optionally skip some tests if building a sysimage.

ToDo : also deal with the path issue.

Will eventually solve #2631

### DIFF
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -14,6 +14,7 @@ CO = joinpath(tmp, "CompileOscar.jl")
 write(CO, """
 using Oscar
 using Pkg, Test
+buildingSysImage=true
 Oscar.system("precompile.jl")
 """)
 
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -2,7 +2,16 @@ if VERSION < v"1.9.0-DEV"
   error("Julia >= 1.9 required")
 end
 
+oscarpath = pkgdir(Oscar)
+
+println("=====================")
+println(oscarpath)
+println("=====================")
+
 using Pkg
+Pkg.activate(temp=true)
+Pkg.add(path="$(oscarpath)")
+using Oscar
 Pkg.add("PackageCompiler")
 Pkg.add("Libdl")
 
@@ -11,10 +20,14 @@ using PackageCompiler, Libdl
 tmp = mktempdir(cleanup = false)
 CO = joinpath(tmp, "CompileOscar.jl")
 
+
 write(CO, """
+using Pkg
+Pkg.add(path="$(oscarpath)")
+Pkg.precompile()
 using Oscar
-using Pkg, Test
-buildingSysImage=true
+using Test
+println("oscarpath is $(oscarpath) , while Oscar path is $(pkgdir(Oscar))")
 Oscar.system("precompile.jl")
 """)
 
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -4,13 +4,9 @@ end
 
 oscarpath = pkgdir(Oscar)
 
-println("=====================")
-println(oscarpath)
-println("=====================")
-
 using Pkg
 Pkg.activate(temp=true)
-Pkg.add(path="$(oscarpath)")
+Pkg.develop(path="$(oscarpath)")
 using Oscar
 Pkg.add("PackageCompiler")
 Pkg.add("Libdl")
@@ -23,11 +19,9 @@ CO = joinpath(tmp, "CompileOscar.jl")
 
 write(CO, """
 using Pkg
-Pkg.add(path="$(oscarpath)")
-Pkg.precompile()
+Pkg.develop(path="$(oscarpath)")
 using Oscar
 using Test
-println("oscarpath is $(oscarpath) , while Oscar path is $(pkgdir(Oscar))")
 Oscar.system("precompile.jl")
 """)
 
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -23,3 +23,11 @@ PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_exe
 
 println("(re)start julia as")
 println("\tjulia -J $(sysimage)")
+
+# Remove the packages we added, so that 
+# Aqua doesn't complain in future testing
+Pkg.rm("PackageCompiler")
+Pkg.rm("Libdl")
+Pkg.rm("Documenter")
+Pkg.rm("PrettyTables")
+Pkg.rm("Printf")
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -23,11 +23,3 @@ PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_exe
 
 println("(re)start julia as")
 println("\tjulia -J $(sysimage)")
-
-# Remove the packages we added, so that 
-# Aqua doesn't complain in future testing
-Pkg.rm("PackageCompiler")
-Pkg.rm("Libdl")
-Pkg.rm("Documenter")
-Pkg.rm("PrettyTables")
-Pkg.rm("Printf")
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -7,7 +7,7 @@ import Random
 numprocs_str = get(ENV, "NUMPROCS", "1")
 
 oldWorkingDirectory = pwd()
-cd( joinpath(pkgdir(Oscar), "test") )
+cd(joinpath(pkgdir(Oscar), "test"))
 
 if !isempty(ARGS)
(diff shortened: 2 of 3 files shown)


thofma:
How long does building the system image take (roughly)?

aaruni96:
It takes upwards of an hour (~30 minutes to run through the tests, then, 40+ minutes to compile the system image).

Edit: I just ran it again, and it took only ~24 minutes in the compilation phase. Its possible I remember the 40 minute figure from running it on a different(slower) machine.

system/Build.jl
```diff
@@ -22,3 +23,11 @@ PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_exe
...
 println("(re)start julia as")
 println("\tjulia -J $(sysimage)")
+
+# Remove the packages we added, so that 
+# Aqua doesn't complain in future testing
+Pkg.rm("PackageCompiler")
+Pkg.rm("Libdl")
+Pkg.rm("Documenter")
+Pkg.rm("PrettyTables")
+Pkg.rm("Printf")
```
lgoettgens:
```suggestion
Pkg.rm("PackageCompiler")
Pkg.rm("Libdl")
```
I would instead remove them in `precompile.jl` if that is possible. Then add and rm would be in the same file.

test/runtests.jl
```diff
@@ -6,6 +6,9 @@ import Random
 
 numprocs_str = get(ENV, "NUMPROCS", "1")
 
+oldWorkingDirectory = pwd()
+cd( joinpath(pkgdir(Oscar), "test") )
```
lgoettgens:
```suggestion
cd(joinpath(pkgdir(Oscar), "test"))
```

lgoettgens (commented):
After seeing how other test-only dependencies are handled here, I would disagree with myself (https://github.com/oscar-system/Oscar.jl/issues/2631#issuecomment-1667594751) in that we should handle all test-only dependencies the same. So this would mean for Aqua to be added in `precompile.jl`, and we do not need the split of the testlist.

aaruni96:
> So this would mean for Aqua to be added in precompile.jl, and we do not need the split of the testlist.

We cannot do this, because Aqua complains about adding packages to the `Project.toml`, and therefore the first test fails. (Or at least, we cannot do this if the active project is Oscar's `Project.toml`).

```julia
Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13]: Test Failed at /home/akaushik/.julia/packages/Aqua/HpgB8/src/stale_deps.jl:23
  Expression: result ⊜ true
   Evaluated: ⟪result: 😭 FAILED: Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13]
    Some package(s) in `deps` of Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] are not loaded during via `using Oscar`.
    * Documenter [e30172f5-a6a5-5a46-863b-614d45cd2de4]
    * Aqua [4c88cf16-eb10-579e-8560-4a9242c79595]
    * PackageCompiler [9b87118b-4619-50d2-8e1e-99f35a4d4d9d]
    * PrettyTables [08abe8d2-0d0c-5749-adfa-8a2ac140af0d]
    
    To ignore from stale dependency detection, pass the package name to `ignore` keyword argument of `Aqua.test_stale_deps`
⟫ ⊜ true

...

Test Summary:                                    | Pass  Fail  Total   Time
Aqua.jl                                          |    5     2      7  11.9s
  Method ambiguity                               |              None   0.0s
  Unbound type parameters                        |              None   0.0s
  Undefined exports                              |    1            1   0.0s
  Compare Project.toml and test/Project.toml     |    1            1   0.1s
  Stale dependencies                             |          1      1  10.0s
    Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] |          1      1   0.2s
  Compat bounds                                  |          1      1   0.2s
    Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] |          1      1   0.0s
  Project.toml formatting                        |    3            3   0.1s
  Piracy                                         |              None   0.0s
```

system/Build.jl
```diff
@@ -22,3 +23,11 @@ PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_exe
 
 println("(re)start julia as")
 println("\tjulia -J $(sysimage)")
+
+# Remove the packages we added, so that 
+# Aqua doesn't complain in future testing
```
fingolfin: Can't we create a temporary environment, install packages just in there, and discard it later? That seems cleaner and simpler to me, and also reduces the risk of leaving things in a partially messed up state in case the user ctrl-C's at a bad moment.

aaruni96: I am unable to make this idea work, and I'm not sure why it doesn't work.

As of commit `3b1f1abb` , trying to `Oscar.build()` fails, with one of the problems being that `Oscar.system("precompile.jl")` includes `precompile.jl` from the "system" installation of Oscar, not the the place pointed to by the curent `Project.toml`, but I don't understand *why* it chooses the wrong path.

(There are also other problems relating to missing source files of the extra packages I'm trying to include, but that might be easier to fix once I understand why we are using the wrong path to begin with.)

fingolfin: Maybe we can sit together and debug this?

system/Build.jl
```diff
@@ -2,7 +2,16 @@ if VERSION < v"1.9.0-DEV"
...
 
+oscarpath = pkgdir(Oscar)
+
+println("=====================")
+println(oscarpath)
+println("=====================")
+
 using Pkg
+Pkg.activate(temp=true)
+Pkg.add(path="$(oscarpath)")
```
fingolfin: I think we want 
```suggestion
Pkg.develop(path="$(oscarpath)")
```
as I think `Pkg.add` will copy stuff. Indeed:
```
julia> Pkg.add(path="/Users/mhorn/Projekte/OSCAR/AbstractAlgebra.jl")
    Updating git-repo `/Users/mhorn/Projekte/OSCAR/AbstractAlgebra.jl`
   Resolving package versions...
    Updating `/private/var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T/jl_N9fIWI/Project.toml`
  [c3fe647b] + AbstractAlgebra v0.31.1 `/Users/mhorn/Projekte/OSCAR/AbstractAlgebra.jl#master`
    Updating `/private/var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T/jl_N9fIWI/Manifest.toml`
...
julia> using AbstractAlgebra

julia> pkgdir(AbstractAlgebra)
"/Users/mhorn/.julia/packages/AbstractAlgebra/zuBl1"
```

In contrast:
```
julia> Pkg.develop(path="/Users/mhorn/Projekte/OSCAR/AbstractAlgebra.jl")
   Resolving package versions...
    Updating `/private/var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T/jl_ezgpmP/Project.toml`
  [c3fe647b] + AbstractAlgebra v0.31.1 `~/Projekte/OSCAR/AbstractAlgebra.jl`
    Updating `/private/var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T/jl_ezgpmP/Manifest.toml`
...

julia> using AbstractAlgebra

julia> pkgdir(AbstractAlgebra)
"/Users/mhorn/Projekte/OSCAR/AbstractAlgebra.jl"
```

aaruni96: This seems to have fixed it

fingolfin (approved):

PR #2637, opened 2023-08-07
Removes duvalsing and FTheory tests which randomly fail

ederc:
Some tests for duvalsing randomly fail due to absolute factorisation. This hinders several pull requests at the moment, thus remove those tests for the time being.


afkafkafk13 (approved):
Makes sense in the current situation, as these failures block everyone.

PR #2640, opened 2023-08-08
Fix minor triangulation bug

lkastner:


HereAround (approved):
Thank you @lkastner . Great!

@ederc these changes should fix the FTheory tests.

PR #2641, opened 2023-08-08
change printing of complexes, resolutions and presetations

fieker:
requires https://github.com/thofma/Hecke.jl/pull/1181


afkafkafk13:
Looking just at the diffs, the changes make a lot of sense to me. Thank you!

However, there seem to be Hecke prerequisites which are not present in the testing environment. Thus I do not do a formal review for now.

PR #2644, opened 2023-08-09
LieAlgebras: Homomorphisms for Lie algebras and modules

lgoettgens:
New features:

- Morphisms of Lie algebras
  - image (of ideals, subalgebras)
  - kernel
- Morphisms of Lie algebra modules
- Constructor for abelian Lie algebras
- Constructor for modules with trivial action
- Solvability and nilpotency checks for Lie algebras
- Update to the printing of Lie algebra modules, in particular, it is a bit more compact

Docs and tests for everything new are included as well.


lgoettgens:
@thofma Could you please have a look if this is how the AA/Hecke map types are supposed to be used? That would be great to know before I put more time into this direction.

thofma:
Looks good. Alternatively one can also not implement the "image function", but directly overload `image(f::LieAlgebraModuleHom, ...)` and `preimage(f::LieAlgebraModuleHome,....)`.

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -335,6 +335,50 @@ function centralizer(L::LieAlgebra, x::LieAlgebraElem)
...
+    push!(series, next)
+    curr = next
+  end
+  return series
+end
+
+@doc raw"""
+    descending_central_series(L::LieAlgebra) -> Vector{LieAlgebraIdeal}
+
+Return the descending central series of `L`, i.e. the sequence of ideals
```
fingolfin: There are many descending centrals series of a Lie algebra in general, but here you seem to look at the one that descends quickest, i.e., the lower central series. I'd recommend to call the function that, then it will also be in line with the corresponding function for groups (which I just noticed is missing, I'll make a PR adding it)

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -335,6 +335,50 @@ function centralizer(L::LieAlgebra, x::LieAlgebraElem)
...
+  while true
+    next = bracket(L, curr)
+    dim(next) == dim(curr) && break
+    push!(series, next)
+    curr = next
+  end
+  return series
+end
+
+@alias lower_central_series descending_central_series
```
fingolfin: Ah, I see you added this as an alias. But again, I don't think this is a good choice of terminology for the reason I outlined above.

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -358,11 +420,21 @@ Return `true` if `L` is simple, i.e. `L` is not abelian and has no non-trivial i
 !!! warning
     This function is not implemented yet.
 """
-function is_simple(L::LieAlgebra)
+@attr Bool function is_simple(L::LieAlgebra)
   is_abelian(L) && return false
+  !is_perfect(L) && return false
   error("Not implemented.") # TODO
```
fingolfin, line 427: Note that GAP implements this and much more, so perhaps interfacing in more places with GAP would be beneficial?

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -346,10 +390,28 @@ end
 
 Return `true` if `L` is abelian, i.e. $[L, L] = 0$.
 """
-function is_abelian(L::LieAlgebra)
+@attr Bool function is_abelian(L::LieAlgebra)
   return all(iszero, x * y for (x, y) in combinations(basis(L), 2))
```
fingolfin: note part of this PR, but perhaps better as
```suggestion
  b = basis(L)
  n = length(b)
  return all(iszero, b[i] * b[j] for i in 1:n for j in i+1:n)
```

Just to say:
```
julia> b = [0,1,2,3];

julia> function f1(b)
         n = length(b)
         count(iszero, b[i] * b[j] for i in 1:n for j in i+1:n)
       end
f1 (generic function with 1 method)

julia> function f2(b)
         count(iszero, x * y for (x, y) in Oscar.combinations(b, 2))
       end
f2 (generic function with 1 method)

julia> @btime f1(b)
  51.735 ns (0 allocations: 0 bytes)
3

julia> @btime f1($b)
  9.217 ns (0 allocations: 0 bytes)
3

julia> @btime f2(b)
  693.878 ns (23 allocations: 1.70 KiB)
3

julia> @btime f2($b)
  658.019 ns (23 allocations: 1.70 KiB)
3
```

experimental/LieAlgebras/src/LieAlgebraModuleHom.jl
```diff
@@ -0,0 +1,289 @@
...
+) where {T1<:LieAlgebraModule,T2<:LieAlgebraModule,T3<:LieAlgebraModule}
+  @req codomain(f) === domain(g) "Composition: Maps are not compatible"
+  h = LieAlgebraModuleHom(domain(f), codomain(g), f.matrix * g.matrix; check=false)
+  if isdefined(f, :inverse_isomorphism) && isdefined(g, :inverse_isomorphism)
+    h.inverse_isomorphism = LieAlgebraModuleHom(
+      codomain(g),
+      domain(f),
+      g.inverse_isomorphism.matrix * f.inverse_isomorphism.matrix;
+      check=false,
+    )
```
fingolfin, line 150: Should this also be here?
```suggestion
    )
    h.inverse_isomorphism.inverse_isomorphism = h
```

experimental/LieAlgebras/src/LieAlgebraModuleHom.jl
```diff
@@ -0,0 +1,289 @@
...
+    is_isomorphism(h::LieAlgebraModuleHom) -> Bool
+
+Return `true` if `h` is an isomorphism.
+This function tries to invert the transformation matrix of `h` and caches the result.
+The inverse isomorphism can be cheaply accessed via `inv(h)` after calling this function.
+"""
+function is_isomorphism(h::LieAlgebraModuleHom)
+  isdefined(h, :inverse_isomorphism) && return true
+  fl, invmat = is_invertible_with_inverse(h.matrix)
+  fl || return false
```
fingolfin, line 177: So if the matrix is *not* invertible, the result is not cached, is it?

experimental/LieAlgebras/src/LieAlgebra.jl
lgoettgens: Yes, I will need to look into what GAP offers. On my todo-list, but just not right now (for this PR).

experimental/LieAlgebras/src/LieAlgebraModuleHom.jl
lgoettgens: I added `@attr` caching to `is_isomorphism`. That should be sufficient, right?

fingolfin (approved):

PR #2647, opened 2023-08-10
Update Project.toml to Hecke 0.20

fieker:


thofma:
Since this breaks the doctests, I think you have to adjust them at the same time.

PR #2648, opened 2023-08-10
Minimal betti tables

HechtiDerLachs:
As requested by @wdecker .


wdecker:
Please export function

test/Modules/ModulesGraded.jl
```diff
@@ -1091,3 +1091,36 @@ end
...
+  M = cokernel(inc)
+  A, _ = quo(R, I)
+  # To reproduce the string on the right hand side, evaluate 
+  #   `"$(oscar.minimal_betti_table(M))"` 
+  # and insert the result here; after verification of the result!
+  @test "$(oscar.minimal_betti_table(A))" == "       0  1  2  3  4  \n0    : 1  -  -  -  -  \n1    : -  -  -  -  -  \n2    : -  7  10 5  1  \n3    : -  -  -  -  -  \n4    : -  -  -  -  -  \n---------------------\ntotal: 1  7  10 5  1  \n"
+
+  @test "$(oscar.minimal_betti_table(M))" == "       0  1  2  3  4  \n0    : 1  -  -  -  -  \n1    : -  -  -  -  -  \n2    : -  7  10 5  1  \n3    : -  -  -  -  -  \n4    : -  -  -  -  -  \n---------------------\ntotal: 1  7  10 5  1  \n"
+  @test "$(oscar.minimal_betti_table(I))" == "$(oscar.minimal_betti_table(sub_F))"
+
```
afkafkafk13: I would suggest to put one slightly non-trivial example in the doctests and only compare content here, not strings.

afkafkafk13 (approved):

fieker:
currently I think it 's a documenter failure: the output:
```
Subexpression:
│ 
│ minimal_betti_table(free_resolution(A))
│ 0  1  2  3
│ 
```
implies that the 0 1 2 3 are treated as part of the input rather than pat of the output. That also explains the  suprious 1
Why this happens and how to avoid that: currently no idea
S.th. for next week

PR #2649, opened 2023-08-10
admit `has_preimage` for embedding into abelian closure

ThomasBreuer:
- moved the code that creates the embeddings from the function `character_field` to `src/Rings/AbelianClosure.jl`,
- added `has_preimage` for such embeddings, using a keyword argument in the defining function for computing preimages; note that the auxiliary data needed to decide whether preimages exist are known to the mapping only inside the function -- I do not really like this code but I do not have a better idea
- added `reduce` for p-modular reductions of `QQAbElem` to finite fields

### DIFF
--- a/src/Rings/AbelianClosure.jl
+++ b/src/Rings/AbelianClosure.jl
@@ -25,13 +25,15 @@ module AbelianClosure
 
 using ..Oscar
 
-import Base: +, *, -, //, ==, zero, one, ^, div, isone, iszero, deepcopy_internal, hash
+import Base: +, *, -, //, ==, zero, one, ^, div, isone, iszero,
+             deepcopy_internal, hash, reduce
 
 #import ..Oscar.AbstractAlgebra: promote_rule
 
-import ..Oscar: addeq!, is_unit, parent_type, elem_type, gen, root_of_unity,
-                root, divexact, mul!, roots, is_root_of_unity, promote_rule,
-                AbstractAlgebra, parent
+import ..Oscar: AbstractAlgebra, addeq!, elem_type, divexact, gen,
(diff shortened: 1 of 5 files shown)


src/Rings/AbelianClosure.jl
```diff
@@ -843,6 +846,86 @@ function Oscar.order(a::QQAbElem)
...
+    end
+
+    finv = function(x::QQAbElem; check::Bool = false)
+      if n % conductor(x) == 0
+        return Hecke.force_coerce_cyclo(F, data(x))
+      elseif check
+        return
+      else
+        error("element has no preimage")
+      end
```
fingolfin, line 879: What are the intended semantics of `check` here (and is this stated somewhere)? I am just surprised to see that `check == true` *disables* the error, and instead we return `nothing` (making the return value of this function somewhat type unstable)

ThomasBreuer: The point is that in the case that the given element has no preimage, the code is intended to be used in two different situations:
An exception shall be thrown if one has just asked for a preimage, but the function shall signal the nonexistence of a preimage if one has asked to check for the existence of a preimage; here this is done via a `nothing` return value.

I know that this looks dirty, but the problem is that the `MapFromFunc` objects have access only to their stored data (domain, codomain, and the two Julia functions). Note that the function for computing preimages needs some objects from the context of its creation, thus we cannot simply duplicate the code "outside the creation of the mapping".

(Would it be better to implement `has_preimage` via `try`/`catch` and `preimage`?)

fieker (approved):
I think the reduce needs to be re-visited as we discussed. But for the time being, its fine

PR #2654, opened 2023-08-10
Change all occurrences of `Hecke.Map` to `Map`

lgoettgens:
This semantically changes nothing, as
```julia
julia> Hecke.Hecke.Map === Hecke.Map === AbstractAlgebra.Map === Map
true
```
but in my opinion, it makes it clearer what happens. In particular, I always get confused between `Hecke.Map` (which is `=== Map`) and `Hecke.HeckeMap` (which has more stuff behind it).

I am asking all people who last edited these functions (@HechtiDerLachs, @HereAround, @thofma, @simonbrandhorst) to review this PR. Please check, that you did not get mixed up with `Map` and `Hecke.HeckeMap`. Since I am not that deep into the internals of your parts of Oscar, I will leave that to you.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/Methods.jl
@@ -6,7 +6,7 @@ we compute ``X' = X ×ₖ Spec(R)`` and return a pair `(X', f)` where
 ``f : X' → X`` is the canonical morphism.
 
 !!! note
-    We do not restrict `phi` to be a `Hecke.Map` so that one can also use coercion, anonymous functions, etc. 
+    We do not restrict `phi` to be a `Map` so that one can also use coercion, anonymous functions, etc. 
 """
 function base_change(phi::Any, X::Scheme)
   error("base_change not implemented for input of type $(typeof(X))")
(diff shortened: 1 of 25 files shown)


src/AlgebraicGeometry/Schemes/Methods.jl
```diff
@@ -6,7 +6,7 @@ we compute ``X' = X ×ₖ Spec(R)`` and return a pair `(X', f)` where
 ``f : X' → X`` is the canonical morphism.
 
 !!! note
-    We do not restrict `phi` to be a `Hecke.Map` so that one can also use coercion, anonymous functions, etc. 
+    We do not restrict `phi` to be a `Map` so that one can also use coercion, anonymous functions, etc.
```
simonbrandhorst:
```suggestion
    We do not restrict `phi` to be of type `Map` so that one can also use coercion, anonymous functions, etc. 
```

src/AlgebraicGeometry/Schemes/Methods.jl
```diff
@@ -24,7 +24,7 @@ where ``a : X' → X`` is the morphism from `base_change(phi, X)`,
 morphism on those fiber products.
 
 !!! note
-    We do not restrict `phi` to be a `Hecke.Map` so that one can also use coercion, anonymous functions, etc. 
+    We do not restrict `phi` to be a `Map` so that one can also use coercion, anonymous functions, etc.
```
simonbrandhorst:
```suggestion
    We do not restrict `phi` to be of type `Map` so that one can also use coercion, anonymous functions, etc. 
```

simonbrandhorst (approved):

fingolfin (approved):
Fine by me

HereAround:
I do not see changes on the toric geometry part/major flaws. In this sense, fine by me.

HereAround (approved):

PR #2656, opened 2023-08-11
avoid calling `FamilyPcgs`

ThomasBreuer:
Calling GAP's `FamilyPcgs` with a proper subgroup of a pc group yields a pcgs for the full pc group, see gap-system/gap/issues/5483.

If I remember right then the idea of `relations(G)` for a group `G` was that the relations hold w.r.t. the generators stored for `G`, and then the proposed change will in general violate this assumption.

Perhaps it would be safer to change `relations` such that also the list of generators is returned w.r.t. which the relations hold.

### DIFF
--- a/experimental/GModule/Cohomology.jl
+++ b/experimental/GModule/Cohomology.jl
@@ -443,7 +443,7 @@ function Oscar.relations(G::Oscar.GAPGroup)
 end
 
 function Oscar.relations(G::PcGroup)
-   f = GAP.Globals.IsomorphismFpGroupByPcgs(GAP.Globals.FamilyPcgs(G.X), GAP.Obj("g"))
+   f = GAP.Globals.IsomorphismFpGroupByPcgs(GAP.Globals.Pcgs(G.X), GAP.Obj("g"))
    @req f != GAP.Globals.fail "Could not convert group into a group of type FPGroup"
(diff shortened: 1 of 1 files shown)


fieker:
I need the relations relative to exactly the generators of the group, getting differnt ones will not be helpful

ThomasBreuer:
@fieker Then what about the following: `Oscar.relations(G::PcGroup)` checks whether the stored generators actually are a pcgs. If yes then it calls GAP's `IsomorphismFpGroupByPcgs`, otherwise it calls GAP's `IsomorphismFpGroupByGenerators`.
(In fact, one could then do this in `Oscar.relations(G::Oscar.GAPGroup)`, and no special method for pc groups is needed.)

fieker:
Will I get the pc relations this way?

ThomasBreuer:
Suppose the given group `G` is a pc group but the stored generators are not a pcgs for `G`.
Then we want relations w.r.t. the stored generators, thus we cannot call `IsomorphismFpGroupByPcgs`, and we will not get pc relations.

If the stored generators are a pcgs for `G` and if `IsomorphismFpGroupByPcgs` is called with this pcgs then it returns the pc relations w.r.t. this pcgs.

fingolfin (approved):

PR #2660, opened 2023-08-11
truncate ideals, modules

wdecker:


src/Rings/mpoly-graded.jl
```diff
@@ -2746,3 +2746,78 @@ forget_grading(I::MPolyIdeal{<:MPolyDecRingElem}) = forget_decoration(I)
...
+       if degree(Int, V[i]) < d
+          if degree(Int, V[i]) > s
+             s = degree(Int, V[i])
+	     B = monomial_basis(R, d-s)
+	  end
+	     append!(RES, B .*V[i])
+       else
+           push!(RES, V[i])
+       end
+  end
```
afkafkafk13, line 2820: Please fix indentation.  The way it is typeset currently, it is a bit confusing. (same holds in the module case -- as the block was obviously copied)

src/Modules/ModulesGraded.jl
```diff
@@ -1926,3 +1926,96 @@ function hom(F::FreeMod_dec, G::FreeMod_dec)
...
+5 -> x*z^2*e[1]
+6 -> x*y*z*e[1]
+7 -> x*y^2*e[1]
+8 -> x^2*z*e[1]
+9 -> x^2*y*e[1]
+10 -> x^3*e[1]
+by submodule of F generated by
+1 -> x*e[1]
+2 -> y^4*e[1]
+3 -> z^5*e[1]
```
afkafkafk13, line 1984: The output makes sense, but it is for F/V, where you basically do nothing else , but writing down the monomials of the desired lowest degree in F as generators. Maybe you want to illustrate truncate also with sub(F,V)[1], to parallel the explanations in the ideal case, where we really see something happening.

src/Rings/mpoly-graded.jl
wdecker: I am not sure what precisely you want. Please feel free to make the changes.

PR #2661, opened 2023-08-11
`QuadFormWithIsom` patch 1

StevellM:
First patch of `QuadFormWithIsom`:

- Re-enable some previously deleted tests;
- Extend the `primitive_extensions` methods for general even lattices;
- Fix the output of `image_centralizer_in_Oq` to be consistent with its cousin function `image_in_Oq`;
- Apply further suggestions of the last PR: namely add the semi-colon for optional arguments, remove the instances of `Hecke.` whenever possible, etc...;
- Add missing signatures and add more type stability;
- Try to modify some parts of the codes to improve performance and memory allocations. Hopefully this should help to improve testing of the code, outside the compilation issues;
- Fix some minors issues found using the package.

NB: I forgot to change and create a new branch, so all the old commits appear here... I will delete the current branch after this PR.


experimental/QuadFormAndIsom/src/enumeration.jl
```diff
@@ -461,7 +465,7 @@ function representatives_of_hermitian_type(Lf::ZZLatWithIsom, m::Int = 1)
   ok, rk = divides(rk, euler_phi(n*m))
   ok || return reps
 
-  gene = Hecke.HermGenus[]
+  gene = HermGenus[]
```
thofma: `HermGenus` is not a concrete type if I remember correctly. There should be a `hermitian_genus_type(E)` or something like that in Hecke.

StevellM: Ah yes, indeed, thanks! I forgot about that function..

thofma (approved):

PR #2662, opened 2023-08-11
Singular var name fix

### LINKED ISSUE #1905: Choice of variable names interferes with variables in Singular libraries
The absolute primary decomposition is broken:
```
julia> R, (psi, ) = PolynomialRing(QQ, ["psi"])
(Multivariate Polynomial Ring in psi over Rational Field, fmpq_mpoly[psi])

julia> absolute_primary_decomposition(ideal([psi]))
   skipping text from `)` error at token `)`
ERROR: Could not call function
Stacktrace:
 [1] call_singular_library_procedure
   @ ~/.julia/packages/CxxWrap/IdOJa/src/CxxWrap.jl:618 [inlined]
 [2] low_level_caller_rng(lib::String, name::String, ring::Singular.PolyRing{Singular.n_Q}, args::Singular.sideal{Singular.spoly{Singular.n_Q}})
   @ Singular ~/.julia/dev/Singular/src/caller.jl:374
 [3] absPrimdecGTZ(ring::Singular.PolyRing{Singular.n_Q}, args::Singular.sideal{Singular.spoly{Singular.n_Q}})
   @ Singular.LibPrimdec ~/.julia/dev/Singular/src/Meta.jl:49
 [4] absolute_primary_decomposition(I::MPolyIdeal{fmpq_mpoly})
   @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-ideals.jl:574
 [5] top-level scope
   @ REPL[7]:1
```

Here it is more clear from the error what is happening:
```
julia> R, (p, psi) = PolynomialRing(QQ, ["p", "psi"])
(Multivariate Polynomial Ring in p, psi over Rational Field, fmpq_mpoly[p, psi])

julia> absolute_primary_decomposition(ideal([2//27*p^3 + psi]))
`int` * `map` failed
expected `int` * `int`
expected `int` * `matrix`
expected `int` * `bigintmat`
expected `int` * `intvec`
expected `int` * `intmat`
...parse error
error occurred in or before primdec.lib::absPrimdecGTZ_i line 5872: `K=a,p^3+(27*psi)/2,a;`
```
We have a variable called `psi`, but this interferes with the `psi` in the Singular library function, which seems to be a map.

I know that we had naming problems before and I can of course add `"psi"` to the list here: https://github.com/oscar-system/Singular.jl/blob/master/src/MessyHacks.jl#L164. The problem I see is that this works now for this function. But I can imagine a situation where we are interfering with things in the Singular library and the result will silently be wrong (without error). Maybe there is a better way?  @hannes14? @fingolfin? Not sure who to ask.

Edit: Reported by @apturner

mjrodgers:
This should hopefully fix issues with libSingular giving errors due to bad variable names.

Whenever Oscar creates a singular ring, it will only use generic variable names (x, y, z if there are 3 or less variables, x1, x2, ... otherwise).

This should in particular fix #1905

### DIFF
--- a/src/Rings/FunctionField.jl
+++ b/src/Rings/FunctionField.jl
@@ -26,7 +26,7 @@ end
 # multivariate
 function Oscar.singular_coeff_ring(F::AbstractAlgebra.Generic.FracField{T}) where T <: Union{QQMPolyRingElem, fpMPolyRingElem}
   R = base_ring(F)
-  return Singular.FunctionField(singular_coeff_ring(base_ring(R)), string.(R.S))[1]
+  return Singular.FunctionField(singular_coeff_ring(base_ring(R)), _variables_for_singular(symbols(R)))[1]
 end
 
 function (F::Singular.N_FField)(x::AbstractAlgebra.Generic.Frac{T}) where T <: Union{QQMPolyRingElem, fpMPolyRingElem}
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -567,10 +567,10 @@ julia> gens(groebner_basis(J))
  a*c - 2*a + c
 
 julia> SR = singular_poly_ring(base_ring(J))
-Singular Polynomial Ring (QQ),(a,b,c),(dp(3),C)
+Singular Polynomial Ring (QQ),(x,y,z),(dp(3),C)
 
 julia> I = Singular.Ideal(SR,[SR(-1+c+b+a^3),SR(-1+b+c*a+2*a^3),SR(5+c*b+c^2*a)])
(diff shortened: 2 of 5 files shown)


thofma:
What about the generator name of it is an algebraic extension? Does this also have to be adjusted?

fieker:
On Fri, Aug 11, 2023 at 05:09:18AM -0700, Tommy Hofmann wrote:
> What about the generator name of it is an algebraic extension? Does this also have to be adjusted?
>
Yes. Could be unicode and thus killing singular..

mjrodgers:
Any repeated symbol names generated by this should be handled by `Singular.jl`, I don't think it will make a difference on the Oscar end.

PR #2664, opened 2023-08-11
elliptic surface maintainance

simonbrandhorst:


simonbrandhorst:
@HechtiDerLachs  please shepherd this to a merge
Please check that the decomposition info is still correct.
The rest should be o.k.
The `weierstrass_chart_on_minimal_model` is for you.

HechtiDerLachs:
It turned out that there is a problem with our new Weierstrass chart. 

We did the following: Let U be the classical Weierstrass chart with coordinates `(x, y, t)`. We determined all critical values of `t` which have singular fibers. Then we removed those fibers by passing to the complement via `U_sub = PrincipalOpenSubset(U, h)` with `h` vanishing on the critical fibers and added `U_sub` to the `patches` of the covering. 

When designing coverings and their refinements I made the following assumption. We have implicit refinement trees by the ancestry relation from `PrincipalOpenSubset` and `SimplifiedSpec`. This ancestry is also used for `inherit_glueings!` and the like. For the whole framework to work, it is essential that every `Covering` is completely transversal to this tree, i.e. a `Covering` must not contain two patches `U` and `V` where `U` is an ancestor of `V`. 

But for us `U_sub` is an ancestor of the original Weierstrass chart `U` by construction. We could fix this by creating `U_sub` as a `Spec` rather than a `PrincipalOpenSubset`, but then we would have to glue it with all the other patches manually which is also rather tedious. 

I think the solution is to simply not create and/or use this "non-singular Weierstrass chart". The reason we wanted to introduce it, was that we wanted the rational functions for the fibration hop to be easily visible in the resolutions. Non-singular charts survive the resolution process and so we would naturally find them upstairs. Now with the recent advances in #2681 we do not have this problem anymore: We can simply pull back and push forward rational functions along blowups and their restrictions to strict transforms. 

I suggest to extract the changes in the elliptic parameter section into another pr. I will modify the vinberg script in another branch to use my new technologies and see whether we can get around the Weierstrass chart.

PR #2666, opened 2023-08-12
Catelnuovo-Mumford regularity

wdecker:


src/Modules/UngradedModules.jl
```diff
@@ -9089,3 +9089,48 @@ function has_monomials_on_all_axes(M::SubquoModule)
...
+1 -> x^2*e[1]
+2 -> y^2*e[1]
+3 -> z^2*e[1], F -> M
+e[1] -> e[1]
+Homogeneous module homomorphism)
+
+julia> regularity(M)
+3
+
+julia> minimal_betti_table(M);
```
lgoettgens: Why would you add the betti table to the doctest, if it doesn't get shown?

src/Modules/UngradedModules.jl
```diff
@@ -9089,3 +9089,48 @@ function has_monomials_on_all_axes(M::SubquoModule)
...
+```
+"""
+function regularity(M::ModuleFP)
+ error("Not implemented for the given type")
+end
+
+function regularity(M::FreeMod)
+   return 0
```
lgoettgens: Shouldn't this check the base_ring to be graded as in the Subquo case?

src/Modules/UngradedModules.jl
```diff
@@ -9089,3 +9089,48 @@ function has_monomials_on_all_axes(M::SubquoModule)
...
+end
+
+function regularity(M::SubquoModule)
+   @assert is_standard_graded( base_ring(M))
```
lgoettgens:
```suggestion
   @req is_standard_graded(base_ring(M)) "The base ring is not graded"
```
Adding a custom error message will be helpful for users.

src/Rings/mpoly-ideals.jl
```diff
@@ -1647,3 +1647,32 @@ function grassmann_pluecker_ideal(ring::MPolyRing,
...
+# Examples
+```jldoctest
+julia> R, (w, x, y, z) = graded_polynomial_ring(QQ, ["w", "x", "y", "z"]);
+
+julia> I = ideal(R, [y^2*z − x^2*w, z^4 − x*w^3]);
+
+julia> regularity(I)
+6
+
+julia> minimal_betti_table(I);
```
lgoettgens: Same as above.

src/Rings/mpoly-ideals.jl
```diff
@@ -1647,3 +1647,32 @@ function grassmann_pluecker_ideal(ring::MPolyRing,
...
+```
+"""
+function regularity(I::MPolyIdeal)
+   @assert is_standard_graded( base_ring(I))
```
```suggestion
   @req is_standard_graded(base_ring(I)) "The base ring is not graded"
```
Same as above.

lgoettgens (commented):
Some minor remarks

src/Modules/UngradedModules.jl
wdecker: There is a problem with the documenter in printing betti tables. The ; will be removed as soon as this problem is fixed.

src/Rings/mpoly-ideals.jl
wdecker: as above

src/Modules/UngradedModules.jl
wdecker: Yes, thanks, I will upload  a new file.

micjoswig:
Our decision against submodules in OSCAR requires long names throughout.  Please rename regularity to CM_regularity or similar.

PR #2667, opened 2023-08-12
Gorenstein and Picard index

justus-springer:
Add functions `gorenstein_index` and `picard_index` for normal toric varieties. Also adds `map_from_torusinvariant_cartier_divisors_to_class_group` and `map_from_picard_group_to_class_group`.


src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
```diff
@@ -833,6 +833,38 @@ GrpAb: Z^3
...
+Codomain:
+=========
+Abelian group with structure: Z
+```
+"""
+@attr GrpAbFinGenMap function map_from_torusinvariant_cartier_divisor_group_to_class_group(v::AbstractNormalToricVariety)
+    # check input
+    if has_torusfactor(v)
+        throw(ArgumentError("Group of the torus-invariant Cartier divisors can only be computed if the variety has no torus factor"))
+    end
```
HereAround: Can this also be achieved with the following?
```suggestion
    @req has_torusfactor(v) "Group of the torus-invariant Cartier divisors can only be computed if the variety has no torus factor"
```

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
```diff
@@ -861,10 +893,8 @@ Abelian group with structure: Z
         throw(ArgumentError("Group of the torus-invariant Cartier divisors can only be computed if the variety has no torus factor"))
     end
     
-    # compute mapping
-    map1 = map_from_torusinvariant_cartier_divisor_group_to_torusinvariant_weil_divisor_group(v)
-    map2 = map_from_torusinvariant_weil_divisor_group_to_class_group(v)
-    return restrict_codomain(map1*map2)
+    f = restrict_codomain(map_from_torusinvariant_cartier_divisor_group_to_class_group(v))
+    return f * inv(snf(codomain(f))[2])
```
HereAround, line 895: Why is `inv(snf(codomain(f))[2]` necessary? It seems the original code did not have this.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
```diff
@@ -886,6 +916,72 @@ GrpAb: Z
...
+
+############################
+# Gorenstein and Picard Index
+############################
+
+@doc raw"""
+    gorenstein_index(v::AbstractNormalToricVariety)
+
+Return the Gorenstein index of an abstact normal toric variety `v`.
+
```
HereAround: We use CLS11 as our standard reference for notation, implementation, algorithms. Hence, I just checked CLS11 and could not find a definition of this index in there - at least not yet. Please correct me if I overlooked it.

If indeed not introduced in CLS11, I would suggest to add a reference to the OSCAR bibliography which introduces this index and link it here. Possibly also add a few explaining words.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
```diff
@@ -886,6 +916,72 @@ GrpAb: Z
...
+    f = cokernel(map_from_picard_group_to_class_group(v))[2]
+    order(f(c))
+end
+
+
+@doc raw"""
+    picard_index(v::AbstractNormalToricVariety)
+
+Return the index of the Picard group in the class group of an abstact normal toric variety `v`.
+
```
HereAround: Similar to the above comment on the Gorenstein index - consider adding a few explaining words and/or a reference (if not introduced in CLS11 - I could not yet find it in there).

HereAround (changes requested):
Thank you for this great PR @justus-springer!

Generally speaking, your PR looks great. It adds good new code, documentation and valuable quick tests. Exactly what we aim for. Very nice!

I noticed a couple of things while checking your PR and left comments below.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
justus-springer: You're right, the original code did not include `snf` here. But it did include it for the [class group](https://github.com/oscar-system/Oscar.jl/blob/4690df274f4113fc9e5898c839a133b3033b5c35/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl#L731). The class group was always returned in smith normal form but the Picard group was not. I propose return both of them in snf, especially since I also add `map_from_picard_group_to_class_group`.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
justus-springer: It is defined under the name "index" in Exercises 8.3.10 and 8.3.11 in CLS11 (only for surfaces, but it works equally well in arbitrary dimension). There are many other references, I can try to find a better one if needed.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
justus-springer: Unlike the Gorenstein Index, there are indeed not many references for this index (there is at least one though). Though it's very easy to define, since it's just the index of the Picard group in the class group, under the well-known embedding. I added it since it's so easy to compute and it relates to the Gorenstein index (the latter divides the former).

justus-springer:
Thanks for the review @HereAround ! I replied to your comments and added a few more words explaining `gorenstein_index` and `picard_index`. If you think `picard_index` is too obscure (it does not occur in CLS11), I'd be fine with removing it from the PR.

lkastner:
> If you think `picard_index` is too obscure (it does not occur in CLS11), I'd be fine with removing it from the PR.

Could you maybe just add the reference? https://docs.oscar-system.org/dev/DeveloperDocumentation/documentation/#Updating-the-bibliography If you need this functionality it makes sense to have it in Oscar.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
HereAround: I see. Good catch! That is appreciated.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
HereAround: Thank you for the pointer.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
```diff
@@ -886,6 +914,76 @@ GrpAb: Z
...
+############################
+# Gorenstein and Picard Index
+############################
+
+@doc raw"""
+    gorenstein_index(v::AbstractNormalToricVariety)
+
+Return the Gorenstein index of a $\mathbb{Q}$-Gorenstein normal toric variety `v`. 
+This is the smallest positive integer $l$ such that $lK$ is Cartier, where $K$
+is a canonical divisor on `v`.
```
HereAround: If I read exercise 8.3.10 in CLS correct, then the condition is that `-l * K` is Cartier. Also, I would suggest we link to this example in CLS so that it is clear where this index is defined/what is known (according to CLS). Proposal:
```suggestion
Return the Gorenstein index of a $\mathbb{Q}$-Gorenstein normal toric variety `v`. 
This is the smallest positive integer $l$ such that $-l K$ is Cartier, where $K$
is a canonical divisor on `v`. See exercise 8.3.10 and 8.3.11 in [CLS11](@cite) for more details.
```

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
HereAround: Thank you explaining this index in the doc string. This is definitely sufficient to know what this index is.

I agree with @lkastner that it would be nice to have a reference and link it here. However, if you really cannot find anything, I think this should not be a leg breaker. Unless @lkastner objects, I am in this case good to go without reference.

HereAround:
@justus-springer For your information, the failures of the tests are (very very likely) not due to the changes in this PR. A fix is being worked on...

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl
justus-springer: I have added a reference in the docstring of `picard_index`, let me know what you think. It doesn't really go into detail about it, but it is a reference where the Picard index was used in a classification.

HereAround:
@justus-springer Could you please rebase on the current master, so we can see if the tests complete? (There were significant changes this week.) Thank you!

justus-springer:
Oh sorry, this doesn't look right. I tried rebasing, but I don't know what happened.
Edit: Should be fixed now.

HereAround (approved):
Thank you for making those changes @justus-springer . From my perspective, this PR is good to go. (Of course @lkastner can correct me, should I have overlooked something.)

One subtlety seems to be the `Run tests / doctest (1.9, macOS-latest) (pull_request) `, which times out after 150 minutes. I rerun the tests already, but to no avail. Maybe @benlorenz , @thofma or @fingolfin can propose a way forward?

benlorenz:
> One subtlety seems to be the `Run tests / doctest (1.9, macOS-latest) (pull_request) `, which times out after 150 minutes. I rerun the tests already, but to no avail. Maybe @benlorenz , @thofma or @fingolfin can propose a way forward?

#2685 will probably fix that timeout.

HereAround:
> > One subtlety seems to be the `Run tests / doctest (1.9, macOS-latest) (pull_request) `, which times out after 150 minutes. I rerun the tests already, but to no avail. Maybe @benlorenz , @thofma or @fingolfin can propose a way forward?
> 
> #2685 will probably fix that timeout.

I understand that @lkastner is adding a bit more functionality to this other PR (hence, it is still a draft). So this might take a little while longer.

lkastner (approved):

PR #2669, opened 2023-08-14
Make Cox rings of linear quotients work in greater generality

joschmitt:
Adjusts the function `cox_ring(::LinearQuotient)` to also work for groups which are not subgroups of $SL_n$.
Further minor improvements of `experimental/LinearQuotients`.


experimental/LinearQuotients/src/misc.jl
```diff
@@ -1,27 +1,38 @@
...
-  for h in gens(G)
-    push!(g, map_entries(K, h.elm))
-  end
-  return matrix_group(g)
-end
+@doc raw"""
+    is_reflection(g::MatrixGroupElem)
 
+Return `true` if `g` is a reflection, `false` otherwise. By a reflection, we mean
+a matrix with fixed space of codimension 1.
```
fingolfin: I am not sure that's a generally accepted definition for reflection; I'd also expect the final eigenvalue to be -1...

joschmitt: Should I change it to 'pseudoreflection'?

thofma: I think this would make sense (it is also the definition in Bourbaki).

joschmitt: For what it's worth, in Lehrer--Taylor "Unitary reflection groups" it is called reflection. But I can't argue against the authority of Bourbaki :)

fingolfin (approved):
Nobody in our Wednesday meeting had objections to this today, so let's merge it already :-)

experimental/LinearQuotients/src/misc.jl
```diff
@@ -1,29 +1,40 @@
...
       append!(g, collect(c))
     end
   end
-  return matrix_group(base_ring(G), degree(G), g)
+  return sub(G, g, check = false)
 end
 
-# Check if G contains reflections
+@doc raw"""
+    is_small_group(G::MatrixGroup)
```
fingolfin, line 30: This name is also a bit problematic, seeing as we have the "small groups" library, but it can be changed at another point.

joschmitt: I agree that it is a terrible terminology, but some people use it.

PR #2677, opened 2023-08-14
Graded ring morphisms

### LINKED ISSUE #2657: Special morphisms for graded rings
I ran into the following difficulties when implementing `multi_hilbert_series` for graded modules.

Consider the following setup:
```julia
R, _ = polynomial_ring(QQ, [:x, :y])

# Create a grading group
Z3 = free_abelian_group(3)
Z3_sub, inc = sub(Z3, [3*Z3[1] - Z3[2] + 7*Z3[3]])
G, pr = cokernel(inc)

# G is abstractly isomorphic to ZZ^2. For the `hilbert_series` command 
# we need to make that isomorphisms explicit along the following lines: 
S, (x, y) = grade(R, [5*G[1]-3*G[2] + 9*G[3], G[2]])
H, iso = snf(G)
iso_inv = hom(G, H, [preimage(iso, x) for x in gens(G)])

# Now we can recreate the ring `S` with the "new" grading:
P, (x, y) = grade(R, [iso_inv(degree(x)) for x in gens(S)])

# ...and try to write down the identifying isomorphism:
phi = hom(S, P, gens(P))

# Let's consider a graded free module over `S`:
FG = graded_free_module(S, 2, [-5*G[1], 13*G[2]-G[3]])

# We would like to be able to do the following: 
FH, phi_F = change_base_ring(phi, FG)

# Now `FH` should be isomorphic to `FG`. 
```
The first problem is: With the current implementation neither `phi` nor `change_base_ring` sees that `FG` is graded. Hence, also the output `FH` is not graded. 

In principle, one could go to the implementation of the method of `change_base_ring` and insert some `if is_graded(FG)`-block for the second argument `FG` (deciding based on the type is not possible due to the architecture of the graded modules). But even then, one runs into further problems: If `FG` has non-trivial shifts for its generators, say `d_1, d_2` in `G`, then we need to recover these shifts over `P` via `iso(d_1), iso(d_2)`. But only from `phi` (which is the only argument passed on to `change_base_ring`) we are unable to recover `iso`. 

The solution would be to implement a new concrete type for homomorphisms from/to graded rings which stores an accompanying homomorphism of the grading groups. Further down the road, we will then also need to fix this up for homomorphisms of graded modules over a non-trivial homomorphisms of base rings, etc., so widening the homomorphisms of rings seems to really only be the first step. 

What do you think?

CC: @jankoboehm, @thofma, @wdecker

HechtiDerLachs:
Suggestion to resolve #2657 .

### DIFF
--- a/src/Rings/MPolyMap/DecoratedRings.jl
+++ b/src/Rings/MPolyMap/DecoratedRings.jl
@@ -1,0 +1,208 @@
+#const _DecoratedRingTypes = Union{MPolyDecRing, MPolyQuoRing{<:MPolyDecRingElem}}
+#const _DecoratedRingElemTypes = Union{MPolyDecRingElem, MPolyQuoRingElem{<:MPolyDecRingElem}}
+
+# Since the interface regarding the decorations for decorated rings and their quotients is not 
+# yet coherent, we only allow free polynomial rings for the moment. 
+#
+# For instance `forget_grading` does not work for quotient rings.
+const _DecoratedRingTypes = Union{MPolyDecRing}
+const _DecoratedRingElemTypes = Union{MPolyDecRingElem}
+
+@attributes mutable struct DecoratedRingMorphism{
+    DomainType<:_DecoratedRingTypes,
+    CodomainType<:_DecoratedRingTypes,
+    RingMapType<:MPolyAnyMap,
+    GradingGroupMorphismType,
+    ImgGensType
+  } <: Map{
+    DomainType, CodomainType, Hecke.Hecke.Map, DecoratedRingMorphism
+  }
+
+  domain::DomainType
+  codomain::CodomainType
+  ring_map::RingMapType
+  decoration_map::GradingGroupMorphismType
+  img_gens::Vector{ImgGensType}
+
+
+  function DecoratedRingMorphism(
+      R::_DecoratedRingTypes, # the domain
+      S::_DecoratedRingTypes, # the codomain
+      g::Vector{<:_DecoratedRingElemTypes}, # the images of the generators
+      phi::Map, # the associated homomorphism of the grading groups
+      coeff_map::Any # the map on the coefficient rings
+    )
+    @assert all(x->ishomogeneous(x), g) "images of generators must be homogeneous"
+    @assert all(i->degree(g[i]) == phi(degree(gen(R, i))), 1:ngens(R)) "morphism on grading groups is not compatible with the images of the generators"
+
+    Phi = hom(forget_grading(R), forget_grading(S), coeff_map, forget_grading.(g))
+    return new{typeof(R), typeof(S), typeof(Phi), typeof(phi), elem_type(S)}(R, S, Phi, phi, g)
+  end
+
+  function DecoratedRingMorphism(
+      R::_DecoratedRingTypes, # the domain
+      S::_DecoratedRingTypes, # the codomain
+      g::Vector{<:_DecoratedRingElemTypes}, # the images of the generators
+      phi::Map # the associated homomorphism of the grading groups
+    )
+    @assert length(g) == ngens(R) "number of images does not coincide with the number of variables"
+    @assert all(x->parent(x)===S, g) "images are not elements of the codomain"
+    @assert all(x->ishomogeneous(x), g) "images of generators must be homogeneous"
+    @assert all(i->degree(g[i]) == phi(degree(gen(R, i))), 1:ngens(R)) "morphism on grading groups is not compatible with the images of the generators"
+
+    Phi = hom(forget_grading(R), forget_grading(S), forget_grading.(g))
+    return new{typeof(R), typeof(S), typeof(Phi), typeof(phi), elem_type(S)}(R, S, Phi, phi, g)
+  end
+  function DecoratedRingMorphism(
+      R::_DecoratedRingTypes, # the domain
+      S::_DecoratedRingTypes, # the codomain
+      Phi::Map, # the homomorphism of the rings without decoration
+      phi::Map # the associated homomorphism of the grading groups
+    )
+    @assert domain(Phi) === R "domains are incompatible"
+    @assert codomain(Phi) === S "codomains are incompatible"
+    g = S.(Phi.(gens(domain(Phi))))
+    @assert all(i->degree(g[i]) == phi(degree(gen(R, i))), 1:ngens(R)) "morphism on grading groups is not compatible with the images of the generators"
+    return new{typeof(R), typeof(S), typeof(Phi), typeof(phi), elem_type(S)}(R, S, Phi, phi, g)
+  end
+end
+      
+domain(psi::DecoratedRingMorphism) = psi.domain
+codomain(psi::DecoratedRingMorphism) = psi.codomain
+forget_grading(psi::DecoratedRingMorphism) = psi.ring_map
+decoration_map(psi::DecoratedRingMorphism) = psi.decoration_map
+
+# Note: The codomain of the map below does not necessarily coincide with the codomain of 
+# psi, but elements can be coerced in the latter. 
+coefficient_map(psi::DecoratedRingMorphism) = coefficient_map(forget_grading(psi))
+
+########################################################################
+# User facing constructors
+########################################################################
+
+function hom(R::_DecoratedRingTypes, S::_DecoratedRingTypes, g::Vector;
+    decoration_map=begin
+      G = grading_group(R)
+      H = grading_group(S)
+      # try to infer the morphism of grading groups from g
+      gg_imgs = elem_type(H)[]
+      @assert all(x->parent(x)===S, g) "images of generators do not have the correct parent"
+      for v in gens(G)
+        i = findfirst(x->degree(x)==v, gens(R))
+        i === nothing && error("morphism of grading groups can not be inferred from the ring map")
+        push!(gg_imgs, degree(g[i]))
+      end
+      hom(G, H, gg_imgs)
+    end, 
+    coefficient_map=nothing
+  )
+  # Deflect to coercion if necessary
+  all(x->parent(x) === S, g) || return hom(R, S, S.(g))
+
+  # Since we have to use keyword arguments for the coefficient_map, we can not do anything better here:
+  if coefficient_map === nothing
+    return DecoratedRingMorphism(R, S, g, decoration_map)
+  else 
+    return DecoratedRingMorphism(R, S, g, decoration_map, coefficient_map)
+  end
+end
+
+########################################################################
+# Composition of morphisms of decorated rings
+#
+# In particular, we need to take care of the different constellations 
+# with (non-)trivial coefficient maps, so there are four cases to cover
+########################################################################
+const _trivial_coeff_map = MPolyAnyMap{<:_DomainTypes, <:NCRing, Nothing, <:Any}
+const _non_trivial_coeff_map = MPolyAnyMap{<:_DomainTypes, <:NCRing, <:Any, <:Any}
+const _hecke_coeff_map = MPolyAnyMap{<:_DomainTypes, <:NCRing, <:Map, <:Any}
+
+function compose(
+    psi1::DecoratedRingMorphism{
+              <:_DecoratedRingTypes, 
+              <:_DecoratedRingTypes, 
+              <:_trivial_coeff_map
+            },
+    psi2::DecoratedRingMorphism{
+              <:_DecoratedRingTypes, 
+              <:_DecoratedRingTypes, 
+              <:_trivial_coeff_map
+            }
+  )
+  g = psi2.(psi1.(gens(domain(psi1))))
+  coeff_ring_map = coefficient_map(psi1)
+  dec_map = compose(decoration_map(psi1), decoration_map(psi2))
+  return DecoratedRingMorphism(domain(psi1), codomain(psi2), g, dec_map)
+end
+
+function compose(
+    psi1::DecoratedRingMorphism{
+              <:_DecoratedRingTypes, 
+              <:_DecoratedRingTypes, 
+              <:_trivial_coeff_map
+            },
+    psi2::DecoratedRingMorphism{
+              <:_DecoratedRingTypes, 
+              <:_DecoratedRingTypes, 
+              <:_non_trivial_coeff_map
+            }
+  )
+  g = psi2.(psi1.(gens(domain(psi1))))
+  dec_map = compose(decoration_map(psi1), decoration_map(psi2))
+  return DecoratedRingMorphism(domain(psi1), codomain(psi2), g, dec_map, x->coefficient_map(psi2)(x))
+end
+
+function compose(
+    psi1::DecoratedRingMorphism{
+              <:_DecoratedRingTypes, 
+              <:_DecoratedRingTypes, 
+              <:_non_trivial_coeff_map
+            },
+    psi2::DecoratedRingMorphism{
+              <:_DecoratedRingTypes, 
+              <:_DecoratedRingTypes, 
+              <:_trivial_coeff_map
+            }
+  )
+  g = psi2.(psi1.(gens(domain(psi1))))
+  coeff_ring_map = coefficient_map(psi1)
+  dec_map = compose(decoration_map(psi1), decoration_map(psi2))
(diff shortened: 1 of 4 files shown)


joschmitt:
If we overload `hom` like this, we have the following:
```
julia> R, (x, y) = graded_polynomial_ring(QQ, ["x", "y"], [ 1, 1 ]);

julia> S, (u, v, w) = graded_polynomial_ring(QQ, ["u", "v", "w"], [ 2, 2, 2 ]);

julia> hom(S, R, [ x^2, x*y, y^2 ])
ERROR: morphism of grading groups can not be inferred from the ring map
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] hom(R::MPolyDecRing{QQFieldElem, QQMPolyRing}, S::MPolyDecRing{QQFieldElem, QQMPolyRing}, g::Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
   @ Oscar ~/.julia/dev/Oscar/src/Rings/MPolyMap/DecoratedRings.jl:92
 [3] top-level scope
   @ REPL[10]:1
```
This is the standard situation in invariant theory, so I would really want it to work ~with~ without having to specify the map of the grading groups (which is the identity) by hand.
Right now, I would say that we either need to make the inference of the map on grading groups better (it feels like linear algebra to me, but I haven't really thought about it) or introduce a new method for the graded morphisms, say `graded_hom`, so that `hom` still works as before. In the latter case, I would say that we can remove the inference; if the user wants a graded morphism, they should specify both maps.

HechtiDerLachs:
Thanks for pointing this out @joschmitt ! I think it's important to collect such use cases so that people with a good overview can decide on the preferred user interface. We can do it however people like in the end, but someone has to take a decision. And we do need that special type for morphisms of graded rings.

afkafkafk13:
What is the status here? Is this abandoned/discontinued or just put aside because of more urgent things?

A more general remark: Taking a theorists point of view, having the objects implemented calls for appropriate morphisms. So I am in favor of discussing this topic to get a good solution **at some point**, not necessarily now.

fingolfin:
We discussed this a couple weeks back but I forgot to write a summary here. But we really need this feature. So I'll try to summarize what I remember, with help from some people here who'll correct me:
- the main contentious point was when to "guess" a `decoration_map`
- if the two rings have *identical* grading groups, then it is OK for us to assume that the identity map is meant and use that
- if the grading groups are not isomorphic, of course you must specify a grading map
- this leaves the the case of isomorphic but not identical grading groups. Problem is that isomorphisms are in general not unique.
  - One could maybe make a case for the special case of `ZZ` as there are only two automorphisms, and one reverses signs, so the "identity" is kinda canonical. But we need to decide. Do we want this for this special case?
  - For $Z^n$ with $n>1$ I don't think we should try anything (but again, it's a matter to be decided)

One way out would be to change `graded_polynomial_ring` and `grade` to created *cached* graded groups by default. Then if one specifies a grading via an integer vector or matrix, then the gradings will all be in the same (identical) copy of `Z^n`. But then still people who use custom grading groups would be forced to specify any "nontrivial" grading map...

So we need a decision here to finally get this merged. I think @wdecker wanted to contemplate the options some more and then perhaps propose a solution? I'd also be happy to discuss this some more with people who are interested.

If I don't hear anything within the next 14 days, I'll just make an executive decision and ask @HechtiDerLachs to please implement that, and anyone who doesn't it like can complain to me about it.

simonbrandhorst:
I suppose that our abelian groups come with a distinguished finite set of generators? `gens`
In this case let me propose the following:
If $(G, (g_1,...,g_n))$ $(H, h_1,..., h_k)$  are two grading groups with their fixed generators $g_i$ and $h_i$, then we infer the grading map if and only if
$n = k$ and $g_i \mapsto h_k$ defines a group homomorphism. If this is not the case, then the user has to specify what they actually want.

HechtiDerLachs:
The main design issues that this PR raises have never been discussed and decided. On the contrary, we have established internal workarounds now.

wdecker:
In the light of Pr #4891, let me reopen the discussion here.

To begin with, let me make two proposals:

1) For `hom(R, S, ...)`, we only allow that either both rings are ungraded or that both rings are graded.
2) With respect to inferring the group homomorphism in the graded case, we follow the suggestion of @simonbrandhorst above.

@HechtiDerLachs, @joschmitt, @fingolfin, @simonbrandhorst, @afkafkafk13, @jankoboehm, @thofma, @tom111, and others:

What is your opinion? If you disagree with 1), what are your use cases for "mixed situations"?

thofma:
If `S = grade(R, ...)` with 1), we cannot make the canonical maps `R -> S` and `S -> R` work as in the following example, right?
```
julia> R, = QQ[:x1, :x2]; S, = grade(R, [2, 2]);

julia> h = hom(R, S, gens(S))
Ring homomorphism
  from multivariate polynomial ring in 2 variables over QQ
  to graded multivariate polynomial ring in 2 variables over QQ
defined by
  x1 -> x1
  x2 -> x2

julia> hh = hom(S, R, gens(R))
Ring homomorphism
  from graded multivariate polynomial ring in 2 variables over QQ
  to multivariate polynomial ring in 2 variables over QQ
defined by
  x1 -> x1
  x2 -> x2
```

wdecker:
> If `S = grade(R, ...)` with 1), we cannot make the canonical maps `R -> S` and `S -> R` work as in the following example, right?
> 
> ```
> julia> R, = QQ[:x1, :x2]; S, = grade(R, [2, 2]);
> 
> julia> h = hom(R, S, gens(S))
> ```

Yes, but you can use this (this is, in fact, how grade is implemented):

`julia> h = map(S, gens(R))`
`2-element Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}:`
 `x1`
 `x2`

Again, the question is "what use cases do we have in mind?". Like "do we need kernel computations in case `R` is graded, and `S` is not?".

Oooops,

`julia> map(R, gens(S))`

does not work.

joschmitt:
I thought about this again and I am with Wolfram. I don't see why we need a morphism between a non-graded and a graded ring or a non-graded morphism between graded rings. I use(d) graded rings quite a lot, also with different gradings on the "same" ring.

* If one has mathematically identical rings with different gradings, one can use `forget_grading` and type casting to move elements between the rings (efficiently!). Maybe there is room for improvement here as Wolfram discovered in his comment.
* If the rings are different with incompatible gradings (like in #4891), I feel that it is safer to not allow the creation of any morphism instead of falling back to a non-graded morphism. Quite probably something went wrong and the user should think about what they want. One can create the non-graded morphism between the non-graded rings (via `forget_grading`) and this is mathematically precise.

tom111:
I have a hard time understanding all the consequences of such choices for the design of OSCAR, so I will leave that to you. It's the classical problem of symbolic computation that we humans when we write R=k[x,y] on the blackboard, we keep in mind that this thing is simultaneously a vector space, a ring, an N-graded ring, an N^2-graded ring, a polynomial ring, an algebra, a monoid algebra, a Z^2-graded module, etc. 

Mapping all of this to the type system and then requiring the user to always be explicit about the type might be too much, but I also see the benefits.

As a user, I would prefer a relatively clean interface where I don't need to typecast back and forth all the time and I don't want to fall into traps like issue #4891. ( I think the solution for that issue should be that `hom` fails, not `kernel`. )

afkafkafk13:
I agree with Johannes, but we need to make this clear to the users. This means:
- meaningful error messages (e.g. hom complaining not kernel in #4891 and pointing the user to the grading problem)
- examples showing how to proceed within this philosophy

question: What short cuts would occasional users need to not get their hands frustratingly dirty with moving stuff around?

tom111:
> * If the rings are different with incompatible gradings

The gradings *are* compatible in this case, but the data for the map does not define a homomorphism of graded algebras, therefore I agree that `hom` should fail. But this will require some computation in `hom`, checking if the data is correct to define an arrow in the respective category.

So for example, in the injective resolutions project we started implementing the type `MonoidAlgebra` and then part of defining that new type must be implementing the respective checks in `hom` for that new type.

thofma:
> * If one has mathematically identical rings with different gradings, one can use `forget_grading` and type casting to move elements between the rings (efficiently!). Maybe there is room for improvement here as Wolfram discovered in his comment.

Having the canonical maps not be proper maps is a constant source for trouble and frustration (e.g. in the implementation of polynomial ring maps themselves). For example, one cannot compose, check for domain or codomain, apply it to an ideal, etc. We probably need a better way to represent those canonical maps.

joschmitt:
> > * If one has mathematically identical rings with different gradings, one can use `forget_grading` and type casting to move elements between the rings (efficiently!). Maybe there is room for improvement here as Wolfram discovered in his comment.
> 
> Having the canonical maps not be proper maps is a constant source for trouble and frustration (e.g. in the implementation of polynomial ring maps themselves). For example, one cannot compose, check for domain or codomain, apply it to an ideal, etc. We probably need a better way to represent those canonical maps.

I'm wondering: Let's say we implement some version of a canonical map $\varphi : R\to S$, where $S$ is a graded version of $R$, (in whatever fashion, probably $\varphi$ does not support, for example, `kernel`). But then somebody has a graded morphism $\psi: S\to T$ to another graded ring $T$ and now expects to be able to concatenate the two maps. So, it seems to me that implementing the canonical maps 'right' means ending up with ungraded morphisms between graded rings, but the long way round.

I want to mention the alternative to everything up to here:
* We leave `hom` as it is, so morphisms between graded and non-graded rings are allowed as are non-graded morphisms between graded rings.
* We add a signature `hom(R, S, f, images)`, where `f` is a map of the grading groups, which then gives a graded morphism.

The advantages are that this would allow us to implement the canonical maps as normal ring morphisms and #2657 would be solved. The disadvantages are that #4891 stays as it is and there is some artificial distinction between not-explicitly-graded and graded morphisms.

Or we do Wolfram's suggestion and then add a keyword to disable all checks. Then we can hack in the canonical maps internally when needed. I honestly don't know what I prefer at this point.

jankoboehm:
Here is one more opinon to our problems (following some of the statements above):

We are dealing with two problems, which are transversal.

1) About generating morphisms between graded and ungraded rings. If we think of dehomogenization we of course can also have a hyperplane at infinity which is not the standard one (this is more or less occuring in the above discussion in form of the question of composition of @joschmitt ). I think it is important that we can write down maps of this kind, compose them and support all infrastructure for homomorphisms, in particular the canonical maps as stated by @thofma above. One further reason is that we have to support any higher level structure which should support e.g. dehomogenization, and there are already a lot there and will come further in the future which we do not even know. If theses implementations can just make use e.g. of the canonical homomorphisms as true homomorphisms everything will be fine. In cany case we should avoid to have to cook up a special soluation in each case.

2) With regard to the homomorphism of the grading groups, I would take the identity if both rings are graded with the same group. Otherwise I would require the homomorphism to be given. We can write in the docu that the user should make sure to first create the grading group and then use it in constructing the graded rings, so that no problem will arise. Of course one can still give a non-trivial homomorphims then in case of identical grading groups.

wdecker:
> We are dealing with two problems, which are transversal.

@jankoboehm The two problems are not transversal. If we would stick to your solution of 2, how can we then realize your solution to 1: Entering `h = hom(R, S, images)` if both `R` and `S are graded, Oscar will either infer a homomorphism between the grading groups or throw an error. So how to realize a non-graded morphism between graded rings?

Should we think of having a separate function `graded_hom`?

afkafkafk13:
Having a separate `graded_hom` feels rather unnatural, as we are dealing with graded objects and should respect their structure (true to what I understood of Oscar's philosophy). One might argue that this only works in the easy cases, but we can throw an error if there is neither a very obvious choice like identity nor an optional argument specified.
 
On the other hand, having some `ungraded_hom` for handling graded objects forgetting the grading makes sense to me, because there are indeed cases where you want to map between graded an ungraded objects. But should you then be able to `grade` the resulting morphism by specifying a map of the grading groups?

I am still undecided in what direction we should go....

simonbrandhorst:
I agree with Janko and Tommy, that we need morphisms that behave as such. 
Especially from a user perspective it is nice to have morphisms.  

To address Johannes concern one can give the canonical morphisms a special type which is extra efficient. 

In general the homomorphism should "know" if it is graded or not.

The method `hom` is ambiguous in some cases, because the mathematics is: it does not specify the category in which the homomorphism is defined. We make an educated guess based on the types of the input and I think this design is good. 

In the case of homomorphisms between graded rings I am in favor of inferring the gradings as in my suggestion 2). 

Homomorphisms between graded and ungraded rings should be allowed and are ungraded. 
We will need ungraded homomorphisms between graded rings, because they will/may come up as composition of morphisms. If one want to construct such a thing, then one should explicitly define it and I would say via an extra argument.
`hom(R,S,images; :ungraded)` or something similar?

jankoboehm:
> > We are dealing with two problems, which are transversal.
> 
> @jankoboehm The two problems are not transversal. If we would stick to your solution of 2, how can we then realize your solution to 1: Entering `h = hom(R, S, images)` if both `R` and `S are graded, Oscar will either infer a homomorphism between the grading groups or throw an error. So how to realize a non-graded morphism between graded rings?
> 
> Should we think of having a separate function `graded_hom`?

We could do that, but I tend to agree with @simonbrandhorst

1. The homomorphisms should know whether they are graded, and know their degree (at least in a lazy way)
2. hom should give an ungraded homomorphims if one of the rings is not graded.
3. If both are graded we check homogeniety (and compute the degree) except one sets a flag not to check. If the degree is not computed on construction, it will be computed in a lazy way.
4. if we want to make an ungraded homomorphism of graded objects, then we can have an option to create that, as @simonbrandhorst suggests.
5. We can either leave the current errors in mixed situations or allow the exstence of non-homogenous ideals in graded rings. In this case, if we add, intersect, quotient a homogeneous with an inhomogeneous one we get an inhomogeneous one. The ideal then needs a flag if it is homogeneous.

fingolfin:
We discussed in triage and this is not easy. Some points from here:
- the type of a map should indicate whether it is a homomorphism (including in which category, so e.g. whether it is graded or not)
- there should be a convenient way to test that in a concise way, e.g. functions `is_graded_hom` and `is_graded_hom_type`
- `hom` between graded rings/modules/whatever should return a graded map and otherwise an error (perhaps with a `check=false` option to turn that off for performance reasons, as usual)
- *if* we really need maps between graded and ungraded rings etc., then these are not really morphisms in the sense of category theory. If we really need to support these, then they would have a different type (but note that "different" could include a different set type parameter, so if we have a type `RingMap` it could have a type parameter `IsHom` resp. `IsGraded` etc. which can be `true` or false`, i.e. it does not necessarily require to copy&paste everything)
  - possibly better would be to encode the category in which it is a morphism as a type parameter (so e.g. type param `Category` set to `GradedRing` vs. `Ring` or so)
  - to "upgrade" a non-graded hom between graded rings which actually is preserving the grading to a graded "hom", you need to do something (e.g. construct a new graded hom)


@fieker quite reasonably suggests that we should first try some mock code examples showing how this would play out.

For this it would e.g. be very helpful to know a bit more about these dehomogenization examples, like maybe a minimal example and how it would use such ungraded maps between graded rings etc., just to get a better feeling of how things fit together (or not). Or perhaps we only need maps from graded to ungraded objects or the other way around, then that would be much easier to agree on and handle.

-----

Two more points:
- morphisms between two polynomial rings might actually also be "algebra morphisms" resp. "graded algebra morphism" ; or not; the "Category type param" idea might be able to help with that as well
- @HechtiDerLachs reminds the people discussing in triage that there is also the business of mapping between different grading groups; this has been discussed above already; we won't discuss it now as we've run out of time.

wdecker:
> For this it would e.g. be very helpful to know a bit more about these dehomogenization examples, like maybe a minimal example and how it would use such ungraded maps between graded rings etc., just to get a better feeling of how things fit together (or not). Or perhaps we only need maps from graded to ungraded objects or the other way around, then that would be much easier to agree on and handle.

Having convincing explicit examples would indeed be very helpful. For example, how do ungraded homomorphisms of graded rings arise through composition? Looking e.g. at dehomogenization and homogenization in the context of covered schemes, I read that a `dehomogenization_map` is an instance of type `Map`, while a `homogenization_map` is not.

PR #2686, opened 2023-08-16
Add Coxeter Groups and Root Systems

felix-roehrich:
I am currently working on implementing https://github.com/ulthiel/CoxeterGroups.jl (see #2217). To have compatibility with root systems and since many efficient algorithms for Weyl groups rely on geometric properties, I have also written a first draft for root systems in OSCAR.
Currently this is still under heavy development, especially distributing logic between root systems and Weyl groups.
At the moment Cartan matrices are implemented as ZZMatrix, but it might make sense to implement it as a sparse matrix instead.
Documentation and tests are also missing.

For simplicity everything lives in `experimental/LieAlgebras` but everything from Coxeter groups that does not depend on root systems should probably moved to the Groups folder at some point.


felix-roehrich:
Accidentally tried to merge from master. New PR is #2689.

PR #2687, opened 2023-08-16
Adv/abstract serializer

antonydellavecchia:
Major refactor of load and save mechanisms. Involves storing a type tree that is parsed before the data tree.
The restructuring no longer stores an internal tree in a dict, and can be generalized to other serializer types.
There is currently no upgrade script. I maybe include in this pull request, otherwise I will include in the follow up pull request.


antonydellavecchia:
closes #1507

micjoswig (approved):
With most recent additions, looks good.

It makes sense to put the upgrade script into a different pull request in order to keep things managable.

benlorenz (approved):

PR #2688, opened 2023-08-16
Betti diagrams in Docu

wdecker:


src/Modules/ModulesGraded.jl
```diff
@@ -1318,12 +1346,13 @@ function betti_table(F::FreeResolution; project::Union{GrpAbFinGenElem, Nothing}
...
 
-function betti(b::FreeResolution; reverse_direction::Bool = false)
-  return betti_table(b, project = nothing, reverse_direction = reverse_direction)
+function betti(b::FreeResolution; project::Union{GrpAbFinGenElem, Nothing} = nothing, reverse_direction::Bool = false)
+  return betti_table(b; project, reverse_direction)
 end
 
 function as_dictionary(b::BettiTable)
   return b.B
+  return betti_table(b; project = nothing, reverse_direction = reverse_direction)
```
fingolfin: This line is never executed because of the `return` in the line before it. Is that intentional?

wdecker: thx! Too many copy and paste errors today.

fingolfin (approved):

PR #2697, opened 2023-08-17
Unary minus for toric divisors

justus-springer:


HereAround (approved):
Thank you @justus-springer . This looks good to me!

Issue #2674, opened 2023-08-14
multi_hilbert_series: does not detect infinite dimensional case


JohnAAbbott:
**Describe the bug**
The normal Hilbert function (and Hilbert series) are not defined if there are homogeneous components of infinite dimension, and this can occur with a mixture of positive and negative weights.  Currently a meaningless result is produced.

**To Reproduce**
Steps to reproduce the behaviour, please provide a code snippet that triggers the bug.
```
using Oscar
P,(x,y) = graded_polynomial_ring(QQ, ["x","y"], [1, -1]);
I = ideal(P,[]); # zero ideal
PmodI,_ = quo(P,I);
multi_hilbert_series(PmodI)
```

**Expected behavior**
An error should be signalled, _e.g._ `there are homogeneous components of infinite dimension`

**System (please complete the following information):**
```
julia> Oscar.versioninfo(full=true)
OSCAR version 0.13.0-DEV
...
```

**Additional context**
Claus Fieker pointed out that the situation could be complicated: _e.g._ if in the example above use the ideal `I = ideal(y)` then `P/I` is canonically isomorphic to `QQ[x]` which does have a well-defined Hilbert series.  Using the actual implementation we obtain a result with a common factor between numerator and denominator: numer `1-t^(-1)` and denom `(1-t^(-1))*(1-t)`.

Probably unrelated: should `hilbert_series` and `multi_hilbert_series` give errors if the chosen output ring does not have characteristic zero?

JohnAAbbott:
The solution seems to require checking whether there is a positive integer linear combination of the weights which gives a zero grading.  Claus Fieker reports that this should be possible in Oscar (_e.g._ using some variant of `solve_mixed` or similar).

wdecker:
@JohnAAbbott, @HechtiDerLachs You just forgot to require `is_positively_graded`. I will add this to your code, where needed, and discuss the new functionality for modules in the docu. Just give me a day or two.

JohnAAbbott:
I have put a temporary fix inside the function `_hilbert_series_denominator` (in `src/Rings/hilbert.jl`).  **NB** This low level function cannot use `is_positively_graded` because the ring is not passed as a parameter.

JohnAAbbott:
After talking to @HechtiDerLachs  we think that it makes most sense if the result of `is_positively_graded` is stored inside the `graded_polynomial_ring` so that it is not computed more than once -- perhaps this is already done?

[...]

PR #2698, opened 2023-08-17
Hilbert series integration

JohnAAbbott:
Minor revisions (see also discussion #2674)


src/Rings/hilbert.jl
```diff
@@ -927,6 +927,25 @@ function separate_simple_pps(gens::Vector{PP})
...
+function _hilbert_series_check_weights(W::Vector{Vector{Int}})
+  # assumes W is rectangular (and at least 1x1)
+  # Transpose while converting:
+  ncols = length(W);
+  nrows = length(W[1]);
+  A = zero_matrix(FlintZZ, nrows,ncols);
+  for i in 1:nrows  for j in 1:ncols  A[i,j] = W[j][i];  end; end;
+  b = zero_matrix(FlintZZ, nrows,1);
+  try
+    solve_non_negative(A, b); # any non-zero soln gives rise to infinitely many, which triggers an exception
```
fingolfin: So `b` is a zero vector here. But then isn't the existence of a non-trivial solution simply determined by the rank of `A`? In that case you also don't need to transpose, I guess...

JohnAAbbott: The check is for the existence of **non-negative** solutions, so checking the rank alone is not sufficient.

src/Rings/hilbert.jl
```diff
@@ -927,6 +927,25 @@ function separate_simple_pps(gens::Vector{PP})
...
+  catch _
+    # solve_non_negative must have thrown because there is a non-zero soln
+    error("given weights permit infinite dimensional homogeneous spaces")
+  end
```
thofma: The catched error should be checked against the "expected" error and rethrown otherwise.

(The function could also throw, because your input is garbage or the interface changed or ...)

JohnAAbbott: Corrected!  Thanks for pointing that out.  Anyway that code is obsolescent - hopefully `is_positively_graded` is better written.

src/Rings/hilbert.jl
```diff
@@ -927,6 +927,29 @@ function separate_simple_pps(gens::Vector{PP})
...
 
 
+# !!!OBSOLESCENT!!!   2023-08-17 this fn will no be needed after Wolfram's PR is merged
+# Returns nothing; throws if ker(W) contains a non-zero vector >= 0
+function _hilbert_series_check_weights(W::Vector{Vector{Int}})
+  # assumes W is rectangular (and at least 1x1)
+  # Transpose while converting:
+  ncols = length(W);
+  nrows = length(W[1]);
+  A = zero_matrix(FlintZZ, nrows,ncols);
```
fingolfin:
```suggestion
  ncols = length(W)
  nrows = length(W[1])
  A = zero_matrix(FlintZZ, nrows,ncols)
```

src/Rings/hilbert.jl
```diff
@@ -927,6 +927,29 @@ function separate_simple_pps(gens::Vector{PP})
...
+  for i in 1:nrows  for j in 1:ncols  A[i,j] = W[j][i];  end; end;
```
fingolfin:
```suggestion
  for i in 1:nrows, j in 1:ncols
    A[i,j] = W[j][i]
  end
```

fingolfin:
@HechtiDerLachs (and perhaps also @wdecker ) can you have a look at this? Shall we merge it (perhaps up to the minor tweaks I suggested)?

fingolfin:
And just to clarify: does this resolve issue #2674 ?

wdecker:
@fingolfin The crucial part is marked # !!!OBSOLESCENT!!!  in Oscar.jl/src/Rings/hilbert.jl. This should be removed and be replaced by the already updated `is_positiely_graded` function wherever it is called.
@HechtiDerLachs Can you please do this? In particular, you know better than me how this is handled in Oscar.jl/src/Modules/hilbert.jl

HechtiDerLachs:
I still get the following:
```julia
x, y) = graded_polynomial_ring(QQ, ["x", "y"], [1, -1])

julia> is_positively_graded(R)
true
```
That's not what's supposed to happen, is it?

Should I really try to fix something already?

HechtiDerLachs:
Currently tests fail. Probably due to the bug described above.

wdecker:
I will take a look

wdecker:
@HechtiDerLachs There is an unmotivated `break` in John's `is_positive_grading_matrix` function. Do you agree?

HechtiDerLachs:
No, I think if the comment prior to that method is correct, then the break is fine. Earlier zero entries are caught by the `continue` and trailing entries can be skipped.

wdecker:
Ooops, the whole `is_positive_grading_matrix` function is scrap. We have to rewrite it in a seperate PR.

wdecker:
Sorry @JohnAAbbott, the mistake is on my side, your matrix is the transpose of mine. Things will be changed correspondingly.

src/Rings/mpoly-affine-algebras.jl
```diff
@@ -572,7 +573,7 @@ function multi_hilbert_series(
 
   # Get the weights as Int values: W[k] contain the weight(s) of x[k]
   W = [[ Int(R.d[i][j])  for j in 1:m]  for i in 1:n]
-  fac_denom = _hilbert_series_denominator(HSRing, W)
+  fac_denom = _hilbert_series_denominator(HSRing, W) # also calls _hilbert_series_check_weights
```
HechtiDerLachs:
```suggestion
  fac_denom = _hilbert_series_denominator(HSRing, W)
```

HechtiDerLachs (approved):
Looks fine from my side now.

PR #2700, opened 2023-08-18
Attempt for toric schemes

HereAround:
cc @HechtiDerLachs @lkastner 

This a new approach at the old PR https://github.com/oscar-system/Oscar.jl/pull/2375. Massive changes have occurred in the mean time, which made it necessary to start anew. Certainly, this old PR can be closed once this PR here has been merged.

(For similar reasons: It might be necessary to start fresh on https://github.com/oscar-system/Oscar.jl/pull/2376.)

closes #2375


src/PolyhedralGeometry/PolyhedralFan/fan_types.jl
benlorenz, line 1: This file looks very sketchy, the inheritance `PolyhedralFan{T} <: PolyhedralObject{T}` is now gone (why?) but this is needed for compatibility with the SubObjectIterator and various other interfaces.

The access functions defined below refer to points and vertices which does not make sense for fans and they look like to be copied from somewhere else?

Several of the comments are also wrong.

HereAround: @benlorenz Thank you for the comment. I understand this is a subtle, and so am more than happy to hear about ways that I can improve this PR.

Here is the challenge: In this PR, we want to give the toric varieties the combined functionality of polyhedral fans and schemes. As such, there are currently two abstract types `AbsCoveredScheme` as well as `PolyhedralObject`, that we would like ToricVarieties to be derived of. This is - to my knowledge - not possible in Julia.

In an attempt to resolve this, and yet achieve the full functionality, I have here mimicked the commit https://github.com/oscar-system/Oscar.jl/pull/2375/commits/1199f5b08ee6af90e49dd6668e9ec018f92401be by @lkastner in the original PR linked above. This PR resolved the above "double derivation problem" by freeing `_FanLikeTypes` from the polyhedral context.

The ties of `_FanLikeType` to the `PolyhedralGeometry` grew stronger in the meantime (by virtue of 4efad306ea7), which derived this type from the abstract type `PolyhedralObject`. I can see, that my PR here somewhat reverts this old PR and am happy to hear about other/better ways.

Regarding the access functions. I suppose you mean the following:
```julia
_empty_access() = nothing

function _empty_subobjectiterator(::Type{T}, Obj:: _FanLikeType) where T
    return SubObjectIterator{T}(Obj, _empty_access, 0, NamedTuple())
end

for f in ("_point_matrix", "_vector_matrix", "_generator_matrix")
    M = Symbol(f)
    @eval begin
        function $M(::Val{_empty_access}, P::_FanLikeType; homogenized=false)
            scalar_regexp = match(r"[^<]*<(.*)>[^>]*", String(Polymake.type_name(pm_object(P))))
            typename = scalar_regexp[1]
            T = _scalar_type_to_polymake(scalar_type_to_oscar[typename])
            return Polymake.Matrix{T}(undef, 0, Polymake.polytope.ambient_dim(pm_object(P)) + homogenized)
        end
    end
end

for f in ("_facet_indices", "_ray_indices", "_vertex_indices", "_vertex_and_ray_indices")
    M = Symbol(f)
    @eval begin
        $M(::Val{_empty_access}, P::_FanLikeType) = return Polymake.IncidenceMatrix(0, Polymake.polytope.ambient_dim(P))
    end
end

for f in ("_linear_inequality_matrix", "_linear_equation_matrix")
    M = Symbol(f)
    @eval begin
        function $M(::Val{_empty_access}, P::_FanLikeType)
            scalar_regexp = match(r"[^<]*<(.*)>[^>]*", String(Polymake.type_name(P)))
            typename = scalar_regexp[1]
            T = _scalar_type_to_polymake(scalar_type_to_oscar[typename])
            return Polymake.Matrix{T}(undef, 0, Polymake.polytope.ambient_dim(P))
        end
    end
end

for f in ("_affine_inequality_matrix", "_affine_equation_matrix")
    M = Symbol(f)
    @eval begin
        function $M(::Val{_empty_access}, P::_FanLikeType)
            scalar_regexp = match(r"[^<]*<(.*)>[^>]*", String(Polymake.type_name(P)))
            typename = scalar_regexp[1]
            T = _scalar_type_to_polymake(scalar_type_to_oscar[typename])
            return Polymake.Matrix{T}(undef, 0, Polymake.polytope.ambient_dim(P) + 1)
        end
    end
end

_matrix_for_polymake(::Val{_empty_access}) = _point_matrix
_empty_access() = nothing

function _empty_subobjectiterator(::Type{T}, Obj:: _FanLikeType) where T
    return SubObjectIterator{T}(Obj, _empty_access, 0, NamedTuple())
end

for f in ("_point_matrix", "_vector_matrix", "_generator_matrix")
    M = Symbol(f)
    @eval begin
        function $M(::Val{_empty_access}, P::_FanLikeType; homogenized=false)
            scalar_regexp = match(r"[^<]*<(.*)>[^>]*", String(Polymake.type_name(pm_object(P))))
            typename = scalar_regexp[1]
            T = _scalar_type_to_polymake(scalar_type_to_oscar[typename])
            return Polymake.Matrix{T}(undef, 0, Polymake.polytope.ambient_dim(pm_object(P)) + homogenized)
        end
    end
end

for f in ("_facet_indices", "_ray_indices", "_vertex_indices", "_vertex_and_ray_indices")
    M = Symbol(f)
    @eval begin
        $M(::Val{_empty_access}, P::_FanLikeType) = return Polymake.IncidenceMatrix(0, Polymake.polytope.ambient_dim(P))
    end
end

for f in ("_linear_inequality_matrix", "_linear_equation_matrix")
    M = Symbol(f)
    @eval begin
        function $M(::Val{_empty_access}, P::_FanLikeType)
            scalar_regexp = match(r"[^<]*<(.*)>[^>]*", String(Polymake.type_name(P)))
            typename = scalar_regexp[1]
            T = _scalar_type_to_polymake(scalar_type_to_oscar[typename])
            return Polymake.Matrix{T}(undef, 0, Polymake.polytope.ambient_dim(P))
        end
    end
end

for f in ("_affine_inequality_matrix", "_affine_equation_matrix")
    M = Symbol(f)
    @eval begin
        function $M(::Val{_empty_access}, P::_FanLikeType)
            scalar_regexp = match(r"[^<]*<(.*)>[^>]*", String(Polymake.type_name(P)))
            typename = scalar_regexp[1]
            T = _scalar_type_to_polymake(scalar_type_to_oscar[typename])
            return Polymake.Matrix{T}(undef, 0, Polymake.polytope.ambient_dim(P) + 1)
        end
    end
end

_matrix_for_polymake(::Val{_empty_access}) = _point_matrix
```
Those are defined in `PolyhedralGeometry/iterators.jl` for objects of type `PolyhedralObject`. Initially, I tried to replace this type by `Union{PolyhedralObject, _FanLikeType}`. This lead to an error about ambigious calls. So I defined those functions anew for `_FanLikeType`. It is possible that some of these functions need not be defined for the `_FanLikeTypes` specifically. But still, since they were defined for any object of type `PolyhedralObject`, I was under the impression that all of them should be available for polyhedral fans. Please feel free to correct me.

What comments are wrong?

HereAround: Regarding the `SubObjectIterator` defined in `PolyhedralGeometry/iterators.jl` - have a look at the appearences of `_FanLikeType` in `iterators.jl`. In essence, I have extended the definition of the `SubObjectIterator` to accept `Union{PolyhedralObject, _FanLikeType}` as object, together with one constructor function for `_FanLikeType` and one similar adjustment to empty iterators.

Again - I am more than happy to hear of better ways/alternative ways.

benlorenz: > What comments are wrong?

This one:
```
# We introduce this abstract (hidden) type to allow for other objects to be
# used like polyhedral fans without duplicating too much code, concretely we
# want to be able to directly access rays, maximal_cones, etc for
# NormalToricVariety's.
# abstract type _FanLikeType{T} end
# abstract type AbstractNormalToricVariety end
```
The other one seems correct, I misread some parts of it.

Have you tried defining it like this?
```
struct PolyhedralFan{T} <: PolyhedralObject{T}
```

Regarding the SubObjectIterators I need to have a closer look next week. But I think this should be done and also possible without the duplication. And if code needs to be copied there should be cross-references to make sure this is maintainable.

HereAround: I see. Yes, this comment needs fixing.

I have not tried `struct PolyhedralFan{T} <: PolyhedralObject{T}` yet, but will give it a shot. More shortly.

HereAround: I have been pondering a bit about having or not having a type parameter `T`. Long story short, I am uncertain.

So that I can stare at it better, and hear people's thoughts, let me restate the puzzle.

Currently, we have the following:
```julia
abstract type PolyhedralObject{T} end

abstract type _FanLikeType{T} <: PolyhedralObject{T} end

struct PolyhedralFan{T} <:_FanLikeType{T}
    pm_fan::Polymake.BigObject
    parent_field::Field
   
    PolyhedralFan{T}(pm::Polymake.BigObject, f::Field) where T<:scalar_types = new{T}(pm, f)
    PolyhedralFan{QQFieldElem}(pm::Polymake.BigObject) = new{QQFieldElem}(pm, QQ)
end

abstract type AbstractNormalToricVariety <: _FanLikeType{QQFieldElem} end
```
A.k.a. `AbstractNormalToricVariety <: _FanLikeType <: PolyhedralObject`.

The proposed change is to remove the dependency on the abstract type `PolyhedralObject` and "replace" this dependency with one on the abstract types for schemes. The upshot being that the toric varieties can eventually be called with all methods applicable for polyhedral fans but also any methods that exist for (covered) schemes. Specifically, as long as types are concerned, the proposed change looks like this:

```julia
###########################################
# (1) Types for schemes
###########################################

abstract type Scheme{BaseRingType} end
abstract type AbsSpec{BaseRingType, RingType<:Ring} <: Scheme{BaseRingType} end
abstract type AbsCoveredScheme{BaseRingType} <: Scheme{BaseRingType} end


###########################################
# (2) Types for toric varieties
###########################################

@attributes mutable struct NormalToricVariety <: AbsCoveredScheme{QQField}
           polymakeNTV::Polymake.BigObject
           NormalToricVariety(polymakeNTV::Polymake.BigObject) = new(polymakeNTV)
end

@attributes mutable struct AffineNormalToricVariety <: AbsSpec{QQField, MPolyQuoRing}
           polymakeNTV::Polymake.BigObject
           AffineNormalToricVariety(polymakeNTV::Polymake.BigObject) = new(polymakeNTV)
end

@attributes mutable struct CyclicQuotientSingularity
    polymakeNTV::Polymake.BigObject
end

NormalToricVarietyType = Union{NormalToricVariety, AffineNormalToricVariety, CyclicQuotientSingularity}


###########################################
# (3) Types for polyhedral fans
###########################################

struct PolyhedralFan{T}
    pm_fan::Polymake.BigObject
    parent_field::Field
    PolyhedralFan{T}(pm::Polymake.BigObject, f::Field) where T<:scalar_types = new{T}(pm, f)
    PolyhedralFan{QQFieldElem}(pm::Polymake.BigObject) = new{QQFieldElem}(pm, QQ)
end

const _FanLikeType = Union{NormalToricVarietyType, PolyhedralFan}
const _FanLikeTypeQQ = Union{NormalToricVarietyType, PolyhedralFan{QQFieldElem}}
```
So for instance we have `NormalToricVariety <: AbsCoveredScheme{QQField}`, while `NormalToricVariety` is an element in the union `NormalToricVarietyType`, which in turn is an element in the union `_FanLikeType` as well as `_FanLikeTypeQQ`. The latter contain also the polyhedral fans as elements.

My understanding is that this change works because only one abstract type (e.g. `AbsCoveredScheme{QQField}`) is used for derivations. In particular, `PolyhedralFan{T}` is not derived from `PolyhedralObject`. However, this in turn made all sorts of changes in the internals of `PolyhedralGeometry` necessary (e.g. to the `SubObjectIterators`), which people may or may not like (and I can understand that).

I am still pondering if there is a better way, but currently (in particular based on past conversations) I am not sure I would know a (feasible) alternative.

To complete the picture, it might be worth mentioning that I am super excited that @HechtiDerLachs and I got this running in the first place. We were after this for quite a while and side tracked by many other things (maybe also in part due to the expected amount of technical challenges). Given that this is running now after only about two days of work is super exciting.

benlorenz: I have pushed a commit adding a `PolyhedralObjectUnion` which includes the toric types. This is used in various places instead of PolyhedralObject and thus avoids the code duplication.
Tests currently refuse to start due to conflicts with master.

HereAround: Thank you. Let me look into the conflicts.

HereAround: Fixed all conflicts, let us see what the tests tell us.

HereAround:
I have tried to fix all issues/failures after rebasing this PR to the latest master. It seems the tests now fail due to a type parameter that is being used in the current master in the function `transform(F::_FanLikeType, A::Union{AbstractMatrix{<:Union{Number, FieldElem}}, MatElem{U}}; check=true) where {U<:FieldElem}` in `src/PolyhedralGeometry/PolyhedralFan/standard_constructions.jl`. I removed this dependency on the type parameter `T` in the first commit of this PR but failed to fix the internal dependency of this function on `T`. Help appreciated.

HereAround:
Thank you @benlorenz .

The next errors come from the serialization of toric varieties:
```julia
no method matching load_object(::Oscar.DeserializerState, ::Type{NormalToricVariety}, ::Dict{Symbol, Any})
```
It seems that line 10 in `/src/Serialization/ToricGeometry.jl` defines exactly this function. (For all it matters: This file is exactly as it was before the rebase earlier today, and in this form the test completed last weekend. )

Long story short: I am confused as to what causes this error. And ideas? Presumably it is something rather trivial, but let me ping our expert @antonydellavecchia anyways.

src/exports.jl
```diff
@@ -1356,12 +1356,14 @@ export tetrahedron
 export tighten_simulation
 export to_elementary_symmetric
 export tor
+#export toric_covered_scheme
```
benlorenz:
```suggestion
```

src/exports.jl
```diff
@@ -1356,12 +1356,14 @@ export tetrahedron
...
 export toric_divisor
 export toric_divisor_class
 export toric_ideal
 export toric_identity_morphism
 export toric_line_bundle
 export toric_morphism
+#export toric_spec
```
benlorenz:
```suggestion
```

benlorenz (approved):

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
```diff
@@ -0,0 +1,136 @@
...
+  
+  for i in 1:(length(patch_list)-1)
+    for j in i+1:length(patch_list)
+      
+      X = patch_list[i]
+      Y = patch_list[j]
+      
+      # Step 1: Only glue if the cone of intersection has the "right" dimension:
+      facet = intersect(cone(X), cone(Y))
+      (dim(facet) == dim(cone(X)) - 1) || continue
```
lkastner, line 77: Then you need to assume pureness.

Furthermore this intersection is expensive and it can be read off easily from the Hasse diagram.

src/exports.jl
HechtiDerLachs: I agree: Let's remove these exports.

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
```diff
@@ -0,0 +1,136 @@
...
+      
+      # Step 7: Assign the gluing isomorphisms
+      set_attribute!(gres, :inverse, fres)
+      set_attribute!(fres, :inverse, gres)
+      f = SpecMor(U, V, fres)
+      g = SpecMor(V, U, gres)
+      set_attribute!(g, :inverse, f)
+      set_attribute!(f, :inverse, g)
+      G = SimpleGlueing(X, Y, f, g)
+      add_glueing!(cov, G)
```
lkastner, line 128: Why one giant for loop? Why not put the different steps into smaller helper functions?

src/Oscar.jl
```diff
@@ -189,11 +189,13 @@ include("exports.jl")
 @alias characteristic_polynomial charpoly  # FIXME
 @alias minimal_polynomial minpoly  # FIXME
 
+
```
lkastner:
```suggestion
```
Why?

src/exports.jl
lkastner:
```suggestion
```
I do not think it makes sense to keep this commented code.

src/exports.jl
lkastner:
```suggestion
```

lkastner (approved):
At this point I think it makes sense just to merge and to repair the larger issues later.

src/exports.jl
HereAround: Agreed. Just removed it by accepting Benjamin's suggestion above.

src/exports.jl
HereAround: Agreed. Just removed it by accepting Benjamin's suggestion above.

src/Oscar.jl
HereAround: This must have happened by accident. Thank you for spotting this.

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
HereAround: Thank you for spotting this. Shall we then demand `pureness`  of the fan to begin with and error if violated (like "underlying_scheme is currently only supported for toric varieties with pure fans")?

I recall that we had a discussion of the Hasse diagrams almost one year back. How can we read this off from the Hasse diagram (as in, what code do we have to execute for this)?

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
HereAround: Sounds like a good idea to me. @HechtiDerLachs do you agree?

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
HereAround: I have just pushed https://github.com/oscar-system/Oscar.jl/pull/2700/commits/7930b7c72a7c57bfa7a8b75a375cae9c8ee03778, which requires the fan to be pure for the construction of `underlying_scheme` of a normal toric variety.

As discussed on slack, I will open an issue now, so we do not forget to use the Hasse diagram as you are suggesting.

HereAround: Issue is here: https://github.com/oscar-system/Oscar.jl/issues/2718

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
HereAround: As discussed on slack, added to issue https://github.com/oscar-system/Oscar.jl/issues/2718 so we can discuss and address this in the future.

Issue #2705, opened 2023-08-19
Superfluous line(s) and column(s) in minimal Betti Table


wdecker:
@HechtiDerLachs: Please adjust code.
```
julia> R, (w, x, y, z) = graded_polynomial_ring(QQ, ["w", "x", "y", "z"]);

julia> I = ideal(R, [-x*z+y^2, x*y-w*z, x^2-w*y]);

julia> FI = free_resolution(I)
Free resolution of I
R^3 <---- R^3 <---- R^1 <---- 0
0         1         2         3

julia> betti(FI)
       0 1 2 
------------
2    : 3 2 1 
3    : - 1 - 
------------
total: 3 3 1 


julia> minimal_betti_table(I)
       0 1 2 
------------
2    : 3 2 - 
3    : - - - 
------------
total: 3 2 0 
```

wdecker:
Additional remark: In connection with the way `cm_regularity` is implemented, this causes actually wrong results of the latter function. So may be, this is rather a bug than an issue.

wdecker:
There are also empty spaces.

jankoboehm:
The empty spaces just come from the show function. The additional lines should only occur in the minimal Betti tables and should come from keys mapped to rank zero after minimizing. I would suggest to consider such entries as illegal since we also not include all the infinitely many other zeros, and remove them when generating the Betti table.

wdecker:
@jankoboehm Do you already know what has to be changed with the show function?

HechtiDerLachs:
I think nothing has to be changed about the show function in the end. I just removed the zero entries from the hash-table and it seems to be printing just fine -- at least to all criteria that I am aware of.

wdecker:
This example still produces spaces for BOTH betti_table and minimal_betti_table on my machine:

R, (w, x, y, z) = graded_polynomial_ring(QQ, ["w", "x", "y", "z"]);
I = ideal(R, [w^2-x*z, w*x-y*z, x^2-w*y, x*y-z^2, y^2-w*z]);
A, _ = quo(R, I)
FA = free_resolution(A)
betti_table(FA)
minimal_betti_table(FA)

jankoboehm:
That is to be expected. To remove the zero rows (which can only occur for minimal Betti tables), only the function generating the Betti table, not the show function, had to be changed. To fix the empty spaces at the end of the lines (both minimal and non-minimal case), only the show function has to be fixed.

fingolfin:
[...]

PR #2707, opened 2023-08-21
Remove zero entries.

HechtiDerLachs:
#2705


wdecker (approved):

lgoettgens:
The test failures in 1.6 and 1.9 seem to be connected with betti tables, and thus probably with the changes here.

PR #2715, opened 2023-08-23
Fibration hop functionality

HechtiDerLachs:
I add some ported functionality for @simonbrandhorst from earlier python scripts.


experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,165 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+  return g2, phi2
+end
+
+
+#    r"""
+#    y^2 - quartic(x)
+#   TODO: Use pen and paper to calculate the inverse in the else clause
+#    """
+@doc raw"""
+    transform_to_weierstrass(g::MPolyElem, x::MPolyElem, y::MPolyElem, P::Vector{<:RingElem}; return_inverse::Bool=false)
```
HechtiDerLachs: @simonbrandhorst : Do we need the `return_inverse` option? I did not find it being used and it is also not fully functional (one of the `else` clauses throws an error). In any case: I return directly an endomorphism of the fraction field. We could add an inverse for that.

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
```diff
@@ -76,3 +76,15 @@
...
+  R, (x, y) = polynomial_ring(QQ, [:x, :y])
+  P, (u, v) = polynomial_ring(QQ, [:u, :v])
+  f = (3*x^2 - 5)^2*(y - 5*x^3 + 30*x - 5)^2 - (x^2 -5*x + 2)
+  g, trans = oscar._normalize_quartic(f, parent=P)
+  @test trans(f) == g
+
+  R, (x, y) = polynomial_ring(QQ, [:x, :y])
+  f = y^2 - 4*x^4 + 5*x^3 - 3*x^2 + 7*x - 4
+  trans = oscar.transform_to_weierstrass(f, x, y, QQ.([0, 2]))
+  @test trans(f//1) == (16*x^6 + 1//8*x^5 + 14*x^4*y - 16383//4096*x^4 - 16*x^3*y^2 + 647//128*x^3*y)//y^4
```
HechtiDerLachs, line 89: @simonbrandhorst : Could you have a look here? Frankly, I don't know what that function is really supposed to do. But I compared the output with the previous version (which was accepted, apparently) and the outputs coincide, so I assume the output is correct. 

Maybe you could also make a suggestion for a better docstring?

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,165 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+function _normalize_quartic(g::MPolyRingElem; parent::Union{MPolyRing, Nothing}=parent(g))
+  R = oscar.parent(g)
+  @assert ngens(R) == 2 "polynomial must be bivariate"
+  F = fraction_field(R)
+  kt = coefficient_ring(R)
+  (x, y) = gens(R)
+
+  # Prepare the output ring
+  if parent===nothing
+    R1, (x1, y1) = polynomial_ring(kt, [:x, :y], cached=false)
```
simonbrandhorst: Since it is the output ring and not an intermediate tool,  it should probably be cached? How do other functions behave?

simonbrandhorst: If no parent is given, we could just recycle the parent of g?

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,165 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+  @assert degree(G) == 2 "polynomial must be of degree 2 in its second variable"
+
+  #complete the square
+  h, b, a = collect(coefficients(G))
+  h = -h
+  u = unit(factor(a))
+  a = inv(u)*a
+  b = inv(u)*b
+  @assert issquare(a) "leading coefficient as univariate polynomial in the second variable must be a square"
+  sqa = sqrt(a)
```
simonbrandhorst: looks like a use case for `is_square_with_sqrt`

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,165 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+    R1, (x1, y1) = polynomial_ring(kt, [:x, :y], cached=false)
+  else
+    R1 = parent
+    @assert coefficient_ring(R1) == coefficient_ring(R) "coefficient ring of output is incompatible with input"
+    (x1, y1) = gens(R1)
+  end
+
+  # Get the coefficients of g as a univariate polynomial in y
+  ktx, X = polynomial_ring(kt, :X)
+  ktxy, Y = polynomial_ring(ktx, :y)
```
simonbrandhorst: Probably no need to cache these, since they are just used for intermediate computations.

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,160 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+    y1 = evaluate(y1, [-x//c, y//c])
+    x1 = x1+px
+    #@assert x == evaluate(x1, [x2, y2])
+    #@assert y == evaluate(y1, [x2, y2])
+  end
+  F = fraction_field(R)
+  @assert F === parent(x1) "something is wrong with caching of fraction fields"
+  # TODO: eventually add the inverse.
+  trans = MapFromFunc(F, F, f->evaluate(numerator(f), [x1, y1])//evaluate(denominator(f), [x1, y1]))
+  return trans(F(g)), trans
```
simonbrandhorst: Here we actually care about the factor of `F(g)`  which is in weierstrass form. 
So you would 
1. factor `numerator(F(g))`
2. get the factor h of g which is in weierstrass form and discard the other factors. 
3. Return h, trans

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,160 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+########################################################################
+# Internal functionality for Weierstrass transformation 
+########################################################################
+
+
+@doc raw"""
+    _normalize_quartic(g::MPolyRingElem, parent=nothing)
+
+Transform ``a(x)y^2 + b(x)y - h(x)`` in ``K(t)[x,y]`` to ``y'^2 - h(x')``
+"""
```
simonbrandhorst:
```suggestion
@doc raw"""
    _normalize_hyperelliptic_curve(g::MPolyRingElem, parent=nothing)

Transform ``a(x)y^2 + b(x)y - h(x)`` in ``K[x,y]`` to ``y'^2 - h'(x')``  such that h'(x') is squarefree.
The transformation is a (birational) change of coordinates.
"""
```

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,160 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+  g2 = divexact(g2, evaluate(c, x1))
+  return g2, phi2
+end
+
+
+@doc raw"""
+    transform_to_weierstrass(g::MPolyElem, x::MPolyElem, y::MPolyElem, P::Vector{<:RingElem})
+
+Transform a bivariate polynomial `g` of the form `y^2 - Q(x)` with `Q(x)` of degree ``≤ 4``
+to some better form. This returns a pair `(f, trans)` where `trans` is an endomorphism of the
```
simonbrandhorst:
```suggestion
to weierstrass form. This returns a pair `(f, trans)` where `trans` is an endomorphism of the 
```

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1186,3 +1186,160 @@ function simplify(f::MPolyQuoRingElem{<:Union{<:MPolyRingElem, <:MPolyQuoLocRing
...
+  return g2, phi2
+end
+
+
+@doc raw"""
+    transform_to_weierstrass(g::MPolyElem, x::MPolyElem, y::MPolyElem, P::Vector{<:RingElem})
+
+Transform a bivariate polynomial `g` of the form `y^2 - Q(x)` with `Q(x)` of degree ``≤ 4``
+to some better form. This returns a pair `(f, trans)` where `trans` is an endomorphism of the 
+`fraction_field` of `parent(g)` and `f` is the transform.
```
simonbrandhorst:
```suggestion
`fraction_field` of `parent(g)` and `f` is the transform. The input `P` must be a rational point, i.e.
`g(P) == 0`.
```

experimental/Schemes/elliptic_surface.jl
HechtiDerLachs: How do we find the factor in Weierstrass form? I just extrapolated from our vinberg script here. Is this 100% mandatory what your write here?

simonbrandhorst: I would say that we look at its monomials and make sure that  they 
contain [y^2, x^3] and are all contained in [y^2, x*y, y, x^3, x^2, x, 1].
Yes, I would say that it is 95% mandatory. Because otherwise we will have to fish it out by hand each time and  cannot automatize anything  or fish it out later in the automatization.  This seems like the best place for fishing for the (long) weierstrass equation.

simonbrandhorst:
Tests will need to be adjusted to the new names too.

simonbrandhorst:
Could you please add a test for each method over a function field? Just to make sure it runs and we do not need to fix it later. Because that is where we will actually use it.

simonbrandhorst (approved):

PR #2724, opened 2023-08-24
Weierstrass chart on minimal model

HechtiDerLachs:
As discussed with @simonbrandhorst : This adds back in a smooth "Weierstrass chart" to the minimal model of an elliptically fibered K3-surface. I found the mistake in the previous version and it should now be corrected.

### DIFF
--- a/experimental/Schemes/elliptic_surface.jl
+++ b/experimental/Schemes/elliptic_surface.jl
@@ -93,7 +93,14 @@ end
 Return the generic fiber as an elliptic curve.
 """
 generic_fiber(S::EllipticSurface) = S.E
+
 weierstrass_chart(S::EllipticSurface) = S.Weierstrasschart
+
+function weierstrass_chart_on_minimal_model(S::EllipticSurface)
+  # make sure the minimal model is known
+  relatively_minimal_model(S)
+  return get_attribute(S, :weierstrass_chart_on_minimal_model)
+end
 
 @doc raw"""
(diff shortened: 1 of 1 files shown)


simonbrandhorst:
No need for the decomposition info anymore?

HechtiDerLachs:
That's automatically done by the call to `inherit_decomposition_info!` in the very end.

experimental/Schemes/elliptic_surface.jl
```diff
@@ -93,8 +93,15 @@ end
 Return the generic fiber as an elliptic curve.
 """
 generic_fiber(S::EllipticSurface) = S.E
+
 weierstrass_chart(S::EllipticSurface) = S.Weierstrasschart
 
+@attr AbsSpec function weierstrass_chart_on_minimal_model(S::EllipticSurface)
```
fingolfin, line 99: Just a minor side note: since `AbsSpec` is not a concrete type (it is missing type parameters, and it is abstract), this hint, while not completely useless, still won't make this accessor type stable. Perhaps there is no better type you can specify here, then so be it. Just wanted to make sure you are aware :-)

HechtiDerLachs: Yes, I'm aware of that. Either way, this function is sufficiently high-level that type stability shouldn't be an issue of speed. The type assertion is just here as a plausibility check and to improve readability of the code.

simonbrandhorst (approved):

PR #2728, opened 2023-08-24
Add functions to get more series of subgroups

fingolfin:
This is not yet done, I need to complete the docstrings. But since I mentioned `lower_central_series` elsewhere to @lgoettgens I thought I should at least show this here for now.

### DIFF
--- a/src/Groups/sub.jl
+++ b/src/Groups/sub.jl
@@ -108,6 +108,12 @@ end
 
 Return the trivial subgroup of `G`,
 together with its embedding morphism into `G`.
+
+# Examples
+```jldoctest
+julia> trivial_subgroup(symmetric_group(5))
+(Permutation group of degree 5 and order 1, Hom: permutation group -> permutation group)
+```
 """
 @gapattribute trivial_subgroup(G::GAPGroup) = _as_subgroup(G, GAP.Globals.TrivialSubgroup(G.X)::GapObj)
 
@@ -121,7 +127,15 @@ together with its embedding morphism into `G`.
 """
     index(::Type{I} = ZZRingElem, G::T, H::T) where I <: IntegerUnion where T <: Union{GAPGroup, GrpAbFinGen}
 
-Return the index of `H` in `G`, as an instance of `I`.
+Return the index of `H` in `G`, as an instance of type `I`.
+
+# Examples
+```jldoctest
+julia> G = symmetric_group(5); H, _ = derived_subgroup(G);
+
+julia> index(G,H)
+2
+```
 """
 index(G::T, H::T) where T <: Union{GAPGroup, GrpAbFinGen} = index(ZZRingElem, G, H)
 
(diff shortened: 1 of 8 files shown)


lgoettgens:
Thanks!

If #2644 gets merged before this, the im-/exporting between `Oscar` and `Oscar.LieAlgebras` for `lower_central_series` needs to be redone here.

fingolfin:
This is now ready for review. I've also added examples to a bunch of existing docstrings that had none.

gap/OscarInterface/gap/misc.gi
```diff
@@ -0,0 +1,5 @@
+# When a permutation group gets a stabilizer chain,
+# then we can read of its size from that for basically free.
```
ThomasBreuer:
```suggestion
# then we can read off its size from that for basically free.
```

src/Groups/sub.jl
```diff
@@ -199,6 +292,24 @@ among the nontrivial normal subgroups.
 
 Return the list of characteristic subgroups of `G`,
```
ThomasBreuer, line 293:
```suggestion
Return all characteristic subgroups of `G`,
```

src/Groups/sub.jl
```diff
@@ -617,8 +986,15 @@ end
 """
     derived_subgroup(G::GAPGroup)
 
-Return the derived subgroup of `G`, i.e.,
-the subgroup generated by all commutators of `G`.
+Return the derived subgroup `G'` of `G`, i.e.,
+the subgroup generated by all commutators of `G`,
+together with an embedding `G'` into `G`.
```
ThomasBreuer, line 991:
```suggestion
together with an embedding of `G'` into `G`.
```

ThomasBreuer (approved):
Nice.
(Just a few suggestions.)

As far as I understand, you want to omit the phrase "the vector of" from the description of the return value in those cases where the ordering of the entries does not have a meaning (for example in `normal_subgroups`), but to use "the vector of" for return values describing series.

fingolfin:
> As far as I understand, you want to omit the phrase "the vector of" from the description of the return value in those cases where the ordering of the entries does not have a meaning (for example in `normal_subgroups`), but to use "the vector of" for return values describing series.

Yes, that's roughly the idea. Then for e.g. `subgroups` in the future we maybe can return a special iterator which only computes those groups on the fly. This won't be a vector anymore.

src/Groups/sub.jl
```diff
@@ -199,6 +292,24 @@ among the nontrivial normal subgroups.
...
+ Permutation group of degree 3 and order 2
+ Permutation group of degree 3 and order 2
+ Permutation group of degree 3 and order 3
+ Permutation group of degree 3 and order 6
+
+julia> characteristic_subgroups(quaternion_group(8))
+3-element Vector{PcGroup}:
+ Pc group of order 1
+ Pc group of order 2
+ Pc group of order 8
```
ThomasBreuer, line 309: The result of the doctest says that the groups should appear in decreasing order.
I have checked with several GAP versions that this behaviour is reproducible.
How did the claimed result with increasing group orders arise?

fingolfin: Apparently the order changes depending on whether the CRIPS package is loaded or not.

PR #2685, opened 2023-08-16
ToricGeometry: Fix well definedness check; Allow disabling

lkastner:

PR #2729, opened 2023-08-24
CI: avoid timeouts on macOS

fingolfin:
To conserve resources, we were running both the regular test suites and the doctest on a single macOS instance. That was good until the test suite became so slow that this started routinely time out. That means that now those macOS jobs run ~150 minutes just to fail.

That's a waste, as those macOS runner are strictly limited and often pose the bottleneck for our CI tests to run.

So until the test suite gets faster again (if ever), revert to just running the doctests. Together with the regular tests being run under macOS with Julia nightly, plus several of us developing on macOS, this should hopefully still catch the worst and most obvious macOS regressions.

If this turns out problematic, we'll have to consider alternatives, including:
1. using a third macOS runner again,
2. increasing the timeout
3. using a custom GitHub runner, using a macOS machine we own and control,

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -100,15 +100,17 @@ jobs:
         if: runner.os == 'macOS'
         # restrict number of openMP threads on macOS due to oversubscription
         run: echo "OMP_NUM_THREADS=1" >> $GITHUB_ENV
-      - name: "Run tests"
-        # HACK: since macOS runners are so expensive, we reuse this one to run
-        # both regular tests and doctests
-        if: runner.os == 'macOS'
-        uses: julia-actions/julia-runtest@latest
-        with:
-          annotate: ${{ matrix.julia-version == '1.9' }}
-          coverage: ${{ matrix.julia-version == '1.9' }}
-          depwarn: error
+#       - name: "Run tests"
+#         # HACK: since macOS runners are so expensive, we reuse this one to run
+#         # both regular tests and doctests
+#         # FIXME: since the tests got too slow, we can't afford this HACK anymore,
+#         # so disable it again....
+#         if: runner.os == 'macOS'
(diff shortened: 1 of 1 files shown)


benlorenz:
I think the 150 minutes should be ok right now, where did you see the last timeout of this job?
#2685 should have made the doctests about 20 minutes faster.
In the last two days there were some other crashes of the macos tests during the process cleanup that should be fixed with Polymake.jl 0.11.4, released a few hours ago.

fingolfin:
You are right. I somehow thought those were timeouts, but that was because I failed at converting minutes to hours ☹️ and somehow thought they all failing after roughly 1:30 was the timeout... but it's not.

From the last 48 hours, just on master
- a ton are failing in Perl, which I assume are fixed by your Polymake.jl update
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5961699720/job/16171436551
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5961429839/job/16174275324
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5961429839/job/16174275324
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5956565943/job/16157583012
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5954064446/job/16149745823
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5949953203/job/16136782392
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5938873726/job/16117927773
- some just segfault ... "somewhere"
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5956585289/job/16157642939
  - https://github.com/oscar-system/Oscar.jl/actions/runs/5952946464/job/16146058535
- a few failed due to the temporary regression in the serialization test suite which Anthony has since then fixed
- a bunch just passed

PR #2734, opened 2023-08-24
Sort src/exports.jl

fingolfin:
Using 'sort < src/exports.jl > dummy ; mv dummy src/exports.jl'

### DIFF
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -1,3 +1,5 @@
+# Entries are sorted with uppercase before lowercase. To resort it,
+# execute:  sort < src/exports.jl > dummy ; mv dummy src/exports.jl
 export *
 export @check
 export @pbw_relations
@@ -6,8 +8,8 @@ export @permutation_group
 export @tropical
 export ANTIC
 export AbsAffineAlgebraicSet
-export AbsAffineRationalPoint
 export AbsAffineCurve
+export AbsAffineRationalPoint
 export AbsAffineVariety
 export AbsCoveredCurve
 export AbsCoveredScheme
@@ -31,14 +33,10 @@ export AffineHalfspace
 export AffineHyperplane
 export AffineNormalToricVariety
 export AffineVariety
-export affine_halfspace
-export affine_hyperplane
-export affine_normal_toric_variety
-export algebraic_set
 export AutomorphismGroup
 export AutomorphismGroupElem
-export BorcherdsCtx
 export BettiTable
+export BorcherdsCtx
 export ClosedEmbedding
 export ClosedSubvarietyOfToricVariety
 export CohomologyClass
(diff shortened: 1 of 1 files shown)


joschmitt:
When I first looked into this file, it took me some time to realize that in the used ordering we have Z < a, so everything starting with a capital letter comes first. (I would have sorted with the rule A == a.)
From the changes, it looks like at least one other person didn't notice this, so maybe we can add a comment explaining it?

fingolfin:
Done, including instructions for resorting.

And yes, I carefully phrased this so the comment stays correct even after sorting ;-).

joschmitt:
> Done, including instructions for resorting.
> 
> And yes, I carefully phrased this so the comment stays correct even after sorting ;-).

Great, thank you!

lgoettgens (approved):
Looks good to me. I noticed that you additionally removed a duplicate export for `affine_normal_toric_variety`.
I think most of them are due to some renaming scripts or just tiny mistakes. Does it make sense to you to add some CI job testing this file for being sorted? We could maybe add one that tests multiple "bad patterns" that are consensus to not be used one after the other. This could perhaps include the bibtool thing as well. I could write such a prototype tomorrow.

lgoettgens:
Is this some kind of new ci error? https://github.com/oscar-system/Oscar.jl/actions/runs/5964602260/job/16180184142?pr=2734#step:8:17

joschmitt:
> Looks good to me. I noticed that you additionally removed a duplicate export for `affine_normal_toric_variety`. I think most of them are due to some renaming scripts or just tiny mistakes. Does it make sense to you to add some CI job testing this file for being sorted? We could maybe add one that tests multiple "bad patterns" that are consensus to not be used one after the other. This could perhaps include the bibtool thing as well. I could write such a prototype tomorrow.

~Can't we just add a small script to the CI that sorts the file and checks whether the output is the same, like we already have for the references?~ (I just learned that one can also remove duplicates with the `uniq` command line tool, so `sort < src/exports.jl | uniq > dummy`.)

EDIT: I think I (partially) misunderstood you. Sorry!

lgoettgens:
Yes, basically that. To save a bit of resources, i would have gathered these basic tests into one job. Otherwise multiple containers get spun up with each of them cloning the repository

fingolfin:
I don't think that sorting this file is that important (it is mostly done to take away some "mental burden" from those of us), so I wouldn't want to block PRs just over this, contributors already have to jump through quite some hoops.

joschmitt:
Too late, but I just learned that there is a difference between the ordering used by `sort` on Mac and on Linux systems regarding lower- and uppercase, see https://apple.stackexchange.com/questions/335741/sort-with-case-sensitivity-lowercase-before-uppercase-in-macos-in-the-terminal . On my Ubuntu computer, `sort < src/exports.jl' produces a completely different file.

PR #2735, opened 2023-08-24
Fix another trivial type stability issue

fingolfin:


fingolfin:
@StevellM this is your code, just a heads up (this fix won't affect anything, but since I noticed it while debugging something else, I figured I might as well submit it)

StevellM:
Ah great, thanks! And it looks cleaner like this. I am working on a next batch of functionalities, I might as well do a little round of type stability.

thofma (approved):

PR #2737, opened 2023-08-25
small additions, mainly about class fusions

ThomasBreuer:
- added `atlas_description`,
- added `init_class_fusion`,
- support `fusionmap` parameter in `possible_class_fusions`,
- changed `__init_group_libraries` such that it fits also to `all_character_table_names` (The tests for that cannot be made part of the CI tests because this would slow down the tests in the situation that no Browse package is available.)

(some parts are needed by oscar-system/OSCARBinder/pull/13)

### DIFF
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -19,7 +19,7 @@
 ##
 ##  Atlas irrationalities
 ##
-"""
+@doc raw"""
     atlas_irrationality([F::AnticNumberField, ]description::String)
 
 Return the value encoded by `description`.
@@ -30,6 +30,7 @@ if `F` is not given then the result has type `QQAbElem`.
 `description` is assumed to have the format defined in
 [CCNPW85](@cite), Chapter 6, Section 10.
 
+# Examples
 ```jldoctest
 julia> Oscar.with_unicode() do
          show(atlas_irrationality("r5"))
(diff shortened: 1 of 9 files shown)


src/Groups/group_characters.jl
```diff
@@ -1264,10 +1293,48 @@ end
 
 class_multiplication_coefficient(tbl::GAPGroupCharacterTable, i::Int, j::Int, k::Int) = class_multiplication_coefficient(ZZRingElem, tbl, i, j, k)
 
+@doc raw"""
+    init_class_fusion(subtbl::GAPGroupCharacterTable,
```
fingolfin: Why this name? Is `init` short for `initialize`? What is the rationale behind this name? Does this mean something is also stored in `tbl` and/or `subtbl`?

fingolfin (approved):

src/Groups/group_characters.jl
ThomasBreuer: The corresponding function in GAP is called `InitFusion`, I did not try to find a completely independent new name. 
The idea is an "initial approximation" of what is possible, which is cheap to compute; other functions then try to refine the information by excluding possible images.
Would `approximate_class_fusion` sound better?
(And no, nothing gets stored in the character tables.)

PR #2738, opened 2023-08-25
Adv/serialization cleanup

### LINKED ISSUE #2720: Tune `@registerSerializationType` macro
Summarizing some things I talked about with @antonydellavecchia 
- we should rename the macro to follow naming conventions, so e.g. `@register_serialization_type`
- we should change it to use a "kwarg" style for passing arguments to make code more self-documenting (see [here](https://stackoverflow.com/a/64116235/928031) for how to do that)
  - e.g. `@registerSerializationType(Int64, false, "Base.Int")` could become `@registerSerializationType Int64 name="Base.Int")` or so
- then we can easily add more parameters -- such as `need_params=true/false`, to replace manual setting of `type_needs_params`

So e.g. this
```julia
@registerSerializationType(MatSpace, true)
@registerSerializationType(MatElem)
type_needs_params(::Type{<:MatElem}) = true
```
could become this (whether we use the parenthesis is up to debate, I don't mind either way)
```julia
@registerSerializationType MatSpace uses_id=true
@registerSerializationType MatElem needs_params=true
```
or even this:
```julia
@registerSerializationType MatSpace uses_id
@registerSerializationType MatElem needs_params
```

antonydellavecchia:
Adds a detailed description of serialization mechanism, + register serialization cleanup.
closes #2720 . Also gets read of is_basic-serialization and parent ref arrays` in favour of a single parent ref


src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+# There are three saving and loading functions that are used during serialization
+# they are save_typed_object, load_typed_object, save_type_params, load_type_params
+# and save_object, load_object.
+
+################################################################################
+# save_type_object / load_type_object
+
+# For the most part these functions should not be touched, they are high level
+# functions and are used to (de)serialize the object with its
+# type information as well as it's data. They can be used on a first attempt at
```
```suggestion
# type information as well as its data. They can be used on a first attempt at
```

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
+################################################################################
+# Description of the saving and loading mechanisms
+
+# There are three saving and loading functions that are used during serialization
+# they are save_typed_object, load_typed_object, save_type_params, load_type_params
+# and save_object, load_object.
```
```suggestion
# There are three pairs of saving and loading functions that are used
# during serialization:
# 1. save_typed_object, load_typed_object;
# 2. save_type_params, load_type_params;
# 3. save_object, load_object.
#
# In the following, we discuss each pair in turn.
```

src/Serialization/main.jl
fingolfin: Trying to put myself in the shoes of someone learning about the serialization: what is meant with "a first attempt"?

What I'd really would want to know is: what do I need to do to support serialization of my datatype? So I need to learn about `@register_serialization_type` and which methods I need to overload, and which I *might* want to overload (and why)

If you tell me "you might want to use it, but probably not", that raises more questions than it helps me. It also is not clear what "use" means: call this? Or add new methods for this? Or both? Or neither?

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+
+################################################################################
+# save_type_object / load_type_object
+
+# For the most part these functions should not be touched, they are high level
+# functions and are used to (de)serialize the object with its
```
fingolfin, line 19: Are they "high level functions"? To me it sounds more as if they are "internal" functions: someone who wants to *use* our serialization framework doesn't need them; and someone who wants to extend the framework by support for a new type should not use them either?

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+# type information as well as it's data. They can be used on a first attempt at
+# serializing an object, but in general this will lead to a verbose format and
+# should ultimately only be used for the root object (top level object) i.e. should
+# in general only be called from the save/load function.
+
+################################################################################
+# save_object / load_object
+
+# These functions are at the core of the serialization and are the first functions
+# that should be implemented when working on the serialization of a new type.
```
fingolfin, line 31: OK, that's definite and helpful!

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+# Here is where one should use functions data_dict and data_array to structure
+# the serialization.
```
fingolfin: Can you add an explanation what they are about?

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+
+#  Examples
+#  function save_object(s::SerializerState, obj::NewType)
+#    data_array(s) do
```
fingolfin: Maybe `save_data_array(s)` would be a more fitting name?

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+#      "key1": obj.3,
+#      "key2": obj.4
+#    }
+#  ]
+
+#  function save_object(s::SerializerState, obj::NewType)
+#    data_dict(s) do
+#      save_object(s, obj.1, :key1)
+#      s.key = :key2
+#      data_array(s) do
```
fingolfin: How about
```suggestion
#      data_array(s, :key2) do
```
or even
```suggestion
#      save_data_array(s, :key2) do
```

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+#  }
+#
+# This is ok
+# function save_object(s::SerializerState, obj:NewType)
+#   save_object(s, obj.1)
+# end
+
+# This is bad
+# function save_object(s::SerializerState, obj:NewType)
+#   save_object(s, obj.1, :key)
```
fingolfin, line 89: Does this work? What happens? I would expect this to raise an error.

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+# end
+# if you insist on having a key you should
+# first open a data_dict
+#
+################################################################################
+# save_type_params / load_type_params
+
+# Avoiding type information inside the data branch will lead to a more
+# efficient serialization format. When serialize_with_params(MyType) = true
```
fingolfin: I think this needs to be expanded. For starters, what is the "data branch"?

fingolfin: You should also point out when `serialize_with_params(MyType)` is true, and whether this is something to be done by the person who implements serialization for a new type, or whether it is automatic, or what

src/Serialization/main.jl
```diff
@@ -1,9 +1,104 @@
...
+# type information will be stored as a Dict. This can be set by using the
+# uses_params flag of register_serialization. In general implementing
+# a save_type_params and load_type_params should not happen frequently
```
fingolfin: "should not happen" as in "it will have bad consequences, so don't do it" or "we don't expect it to be necessary often", or ... ? Clarify the language?

How you clarify it depends a bit on whom you target.... As it is, I feel this text needs to decide whom it targets: someone who wants to implement serialization for a new type? Or someone who wants to understand the system? Those are two different things. We probably also need two different texts to cover both.

src/Serialization/upgrades/main.jl
```diff
@@ -66,6 +66,7 @@ sort!(upgrade_scripts; by=version)
 
 ################################################################################
 # Loading with upgrade checks on dict
+const backref_sym = "#backref"
```
fingolfin: Is the change from a Symbol to a string intentional?

src/Serialization/main.jl
```diff
@@ -70,98 +160,49 @@ function registerSerializationType(ex::Any,
...
+macro register_serialization_type(ex::Any, args...)
+  uses_id = false
+  uses_params = false
+  str = nothing
+  for el in args
+    if el isa String
+      str = el
+    elseif el == Symbol("uses_id")
+      uses_id = true
+    elseif el == Symbol("uses_params")
```
fingolfin: More idiomatic would be this:
```suggestion
    elseif el == :uses_id
      uses_id = true
    elseif el == :uses_params
```

src/Serialization/upgrades/0.11.3.jl
```diff
@@ -100,7 +100,7 @@ push!(upgrade_scripts_set, UpgradeScript(
     U = decode_type(dict_type)
 
     # Upgrades QQFieldElem serialization
```
fingolfin: maybe this comment should also be adjusted?

data/Surfaces/abelian_d15_pi21_quintic_3
```diff
@@ -1 +1 @@
...
...2","0","1"],"22114"],[["0","1","3","1","0"],"1"],[["0","1","0","2","2"],"22114"],[["0","0","1","3","1"],"1"]],[[["3","0","1","1","0"],"9877"],[["2","1","2","0","0"],"31990"],[["2","0","0","2","1"],"31990"],[["1","2","0","0","2"],"31990"],[["1","1","1","1","1"],"22458"],[["1","1","0","3","0"],"9877"],[["1","0","3","0","1"],"9877"],[["0","3","0","1","1"],"9877"],[["0","2","1","2","0"],"31990"],[["0","1","1","0","3"],"9877"],[["0","0","2","1","2"],"31990"]],[[["5","0","0","0","0"],"31990"],[["3","0","1","1","0"],"11996"],[["2","2","0","1","0"],"9533"],[["2","0","1","0","2"],"9533"],[["1","1","1","1","1"],"5"],[["1","1","0","3","0"],"11996"],[["1","0","3","0","1"],"11996"],[["1","0","2","2","0"],"9533"],[["0","5","0","0","0"],"31990"],[["0","3","0","1","1"],"11996"],[["0","2","2","0","1"],"9533"],[["0","1","1","0","3"],"11996"],[["0","1","0","2","2"],"9533"],[["0","0","5","0","0"],"31990"],[["0","0","0","5","0"],"31990"],[["0","0","0","0","5"],"31990"]]],"refs":{"412b7d25-2fb3-488d-9818-5d2e0fbb55ce":{"type":"MPolyRing","data":{"base_ring":{"type":"Nemo.fpField","data":"31991"},"symbols":["x","y","z","u","v"]}}},"_ns":{"Oscar":["https://github.com/oscar-system/Oscar.jl","0.13.0-DEV"]}}
```
fingolfin, line 1: Meta comment (not specific to this PR): it seems most of the type information comes after the data. But don't we need the type information before the data when loading?

fingolfin (commented):
Thanks, a great improvement! Some remarks and questions

data/Surfaces/abelian_d15_pi21_quintic_3
antonydellavecchia: the refs come after the data. The type comes before data when there is a ref the load wll jump to that section.

src/Serialization/upgrades/main.jl
antonydellavecchia: Ah, it wasn't, thanks

src/Serialization/main.jl
antonydellavecchia: I agree

src/Serialization/main.jl
antonydellavecchia: I refactor for this and added an @req for when a key is getting overridden

data/Surfaces/abelian_d15_pi21_quintic_3
fingolfin: My point is that when loading data, I cannot just "jump ahead" -- I have to buffer everything in between, then once I reached the type information, need to go back and handle the buffered data.

In this case, even the information that this is an OSCAR serialized file is a the very end -- so the whole file has to be parsed just to determine whether we can load it at all, and whether an upgrade script has to be applied. This makes no difference in the current design because we always load all the data at once anyway. But with other formats we have the potential for arranging things so that we can access parts without parsing everything.

For other formats and data, I see why it may be written last. But why do we write the `_ns` data last, not first? And why isn't the ring written before the ideal? I guess some of this may get clearer once I understand better what you are doing now... Nevertheless, I hope we can change it to something more amenable to serial treatment.

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
+################################################################################
+# Description of the saving and loading mechanisms
+#
+# We require that any types serialized through OSCAR are registered using the
+# @register_serialization_type macro, like so
```
fingolfin: I think this should say
```suggestion
# @register_serialization_type macro. For more information, see its docstring.
```
And then move the explanation below into the docstring.

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
...
+
+# @register_serialization_type NewType "String Representation of type" uses_id uses_params
+
+# The last three arguments are optional and can arise in any order. Passing a String
```
```suggestion
# The last three arguments are optional and can arise in any order. Passing a string
```

src/Serialization/Combinatorics.jl
```diff
@@ -3,14 +3,14 @@ using JSON
 ###############################################################################
 ## Graphs
 ###############################################################################
-@registerSerializationType(Graph{Directed}, "Graph{Directed}")
-@registerSerializationType(Graph{Undirected}, "Graph{Undirected}")
+@register_serialization_type Graph{Directed} "Graph{Directed}"
+@register_serialization_type Graph{Undirected} "Graph{Undirected}"
```
fingolfin, line 7: I think this can now be changed, can't it?
```suggestion
@register_serialization_type Graph{Directed}
@register_serialization_type Graph{Undirected}
```
At least I think `string(Graph{Directed})` now returns `"Graph{Directed}"` when used inside Oscar?

src/Serialization/Rings.jl
```diff
@@ -495,15 +469,15 @@ end
 
 ################################################################################
 # Laurent Series
-@registerSerializationType(Generic.LaurentSeriesRing, true, "LaurentSeriesRing")
-@registerSerializationType(Generic.LaurentSeriesField, true, "LaurentSeriesField")
-@registerSerializationType(ZZLaurentSeriesRing)
+@register_serialization_type Generic.LaurentSeriesRing  "LaurentSeriesRing" uses_id
+@register_serialization_type Generic.LaurentSeriesField "LaurentSeriesField" uses_id
```
fingolfin, line 473: Maybe we should add a comment which explains why the optional string argument is used here -- namely because the type otherwise would be serialized as `AbstractAlgebra.Generic.LaurentSeriesField` (I think)

src/Serialization/Rings.jl
```diff
@@ -524,28 +498,24 @@ function load_object(s::DeserializerState,
...
 # elements
-@registerSerializationType(Generic.LaurentSeriesFieldElem, "LaurentSeriesFieldElem")
-@registerSerializationType(Generic.LaurentSeriesRingElem, "LaurentSeriesRingElem")
-@registerSerializationType(ZZLaurentSeriesRingElem)
-type_needs_params(::Type{<: Union{ZZLaurentSeriesRingElem,
-                                  Generic.LaurentSeriesFieldElem,
-                                  Generic.LaurentSeriesRingElem}}) = true
+@register_serialization_type Generic.LaurentSeriesFieldElem "LaurentSeriesFieldElem" uses_params
+@register_serialization_type Generic.LaurentSeriesRingElem "LaurentSeriesRingElem" uses_params
+@register_serialization_type ZZLaurentSeriesRingElem uses_params
```
fingolfin, line 503: Meta-question (not a blocker for this PR): I wonder, do we really want to serialize `ZZLaurentSeriesRingElem` under that name? Shouldn't a "laurent ring over the integers" automatically be mapped to `ZZLaurentSeriesRingElem`?

I am thinking here of how this might look longterm, in a FAIR format -- the existence of a separate `ZZLaurentSeriesRingElem` type seems like an OSCAR specific implementation detail to me...

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
...
+# arguement will override how the type is store as a string. The last two are boolean
```
```suggestion
# argument will override how the type is stored as a string. The last two are boolean
```
I think the explanation for the optional string needs to be extended. How about this:

> Each serializable type is associated to a unique id that is written and read during initialization. By default this identifier is `string(NewType)` but in some cases it can be desirable to customize it. This is what the optional string argument does, it overrides the id.

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
...
+# flags. When setting uses_id the object will be stored as a reference and will be
```
```suggestion
# flags. When `uses_id` is given then instances of this type will be stored as a reference and will be
```
This explanation is unclear to me. Surely *at some point* the object needs to be stored. I think what is meant is that the object is only serialized once, and all places using it (e.g. if it occurs in a container) reference it indirectly.

And then I would also explain what happens *otherwise*. Namely the object is written immediately whenever it occurs. And if it is referenced in multiple places (e.g. it is twice in a vector), then we write its data twice (and so when loading, we end up with two objects).

Is that right?

If so, some alternative names for this flag come to mind:
- `by_ref` ("by reference" is how we access it)
- `unique` (as we preserve the property that this object is unique -- if serialized multiple times, then after deserialization in all places that referenced it, the same (identical) object is used
- `preserve_identity` (variation of the previous point)

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
...
+# referred to throughout the serialization using a UUID. This should typically only
+# be used for types that do not have a fixed normal form for example PolyRing and MPolyRing.
```
fingolfin: I don't understand this "fixed normal form" bit...

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
...
+# Using the uses_params flag will serialize the object with a more structured type
+# description which will make the serialization more efficient see the discussion on
```
fingolfin: What happens otherwise?

This description makes me think "gee, if this option optimizes things, shouldn't I always use it???"

src/Serialization/main.jl
```diff
@@ -1,9 +1,135 @@
...
+# any information that is stored in a node that is a child of a type node.
+# Avoiding type information inside the data branch will lead to a more
+# efficient serialization format. When the uses_params is set while calling
+# @register_serialization_type (de)serialization will use 
+# save_type_params / load_type_params to format the type information.
+# In general we expect that implementing a save_type_params and load_type_params
+# should not always be necessary. Many types will serialize their types
+# in a similar fashion for example serialization of a FieldElem will
+# use the save_type_params / load_type_params from RingElem since in both cases
+# the only parameter needed for such types is their parent.
```
fingolfin, line 127: OK, I still have questions. But it's fine by me to merge this already, and improve it later.

fingolfin (approved):
Overall this is a great improvement. I left a few more comments, and I also still have questions I have not yet written out, but I am also fine with merging this ASAP, so we can benefit from it and you can continue working.

I'll leave it to you to decide if you want to address my new comments first, or merge first.

data/Surfaces/abelian_d15_pi21_quintic_3
antonydellavecchia: ah I agree about the namespace first, that's an easy fix. The refs have no choice but to be saved last, since we are in the middle of storing an obj when a the need to store a ref pops up. This is different in other serialization where we can potentially open up a second stream  (or context ) for reference. The ring is after because it is a reference, essentially we start saving the ideal, we get to the part where the ideal has a Ring it belongs to but since this ring is stored with an id we just write the id and save the ring so that it is serialized in the refs section. If we had a have an io context we would just get the serializer to just write it out to that context immediately

src/Serialization/main.jl
antonydellavecchia: this is not exactly right. This actually has to do with not having a fixed normal form, essentially thats when it should be used. ids are not for elemen types they are for parent types. i.e. types that have encodings, for example some quotient ring will have a normal form that depends on the ideal. You can figure out when two elements are the same given the how they are encoded with respect to their parent. But I see your point I will add more about this in the next PR.

src/Serialization/main.jl
antonydellavecchia: I'll talk about this more in another PR

src/Serialization/main.jl
antonydellavecchia: otherwise the type information is just a string. Ultimately I do think it should be used whenever possible. I am still trying to figure out when that is exactly and maybe not even exposing this to the user but for now it is a tool for the people who understand the mechanism (1.5 people?) I agree needs a bit more work

data/Surfaces/abelian_d15_pi21_quintic_3
micjoswig: Let me remind you that a JSON object is a dictionary.  There is *no* ordering of the individual properties.

micjoswig: Moreover, it is exactly the UUIDs which allow for additional efficiency.  If the context is known, it is syntactically legal to omit the references in the JSON object.  Of course, the load function needs to cope with this.

src/Serialization/main.jl
micjoswig: PRs and code are not a good place to discuss this.  There is a paper (by @antonydellavecchia @benlorenz and me) coming up on the arXiv by the end of this week.  That should hopefully answer these questions (or needs to be expanded, if not).

micjoswig:
As far as streaming is concerned, I would like to see a *specific* use case (not a hypothetical one).  Then we explain how it is supposed to work.

PR #2741, opened 2023-08-29
[ToricVarieties] Enhance blow_up method

HechtiDerLachs:
@HereAround : Can you take it from here?

### DIFF
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1219,3 +1219,89 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
   return IdealSheaf(X, K, check=false)
 end
 
+
+@doc raw"""
+    IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
+
+Create a sheaf of ideals on a toric variety ``X`` from a homogeneous ideal 
+`I` in its `cox_ring`.
+
+# Examples
+```jldoctest
+julia> P3 = projective_space(NormalToricVariety, 3)
+Normal, non-affine, smooth, projective, gorenstein, fano, 3-dimensional toric variety without torusfactor
+
+julia> (x1,x2,x3,x4) = gens(cox_ring(P3))
+4-element Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}:
+ x1
+ x2
+ x3
+ x4
+
+julia> I = ideal([x2,x3])
+ideal(x2, x3)
+
+julia> IdealSheaf(P3, I);
+```
+"""
+function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
+  @req base_ring(I) === cox_ring(X) "ideal must live in the cox ring of the variety"
+
+  # We currently only support this provided that the following conditions are met:
+  # 1. All maximal cones are smooth, i.e. the fan is smooth/X is smooth.
+  # 2. The dimension of all maximal cones matches the dimension of the fan.
+  @req is_smooth(X) "Currently, ideal sheaves are only supported for smooth toric varieties"
+  @req all(m -> dim(m) == dim(X), maximal_cones(X)) "Currently, ideal sheaves require that all maximal cones have the dimension of the variety"
+
+  ideal_dict = IdDict{AbsSpec, Ideal}()
+
+  # We need to dehomogenize the ideal I in the Cox ring S to the local
+  # charts U_sigma, with sigma a cone in the fan of the variety.
+  # To this end we use Proposition 5.2.10 from Cox-Little-Schenck, p. 223ff.
+  # Abstractly, the chart U_sigma is isomorphic to C^n with n the number
+  # of ray generators of sigma, owing to the assumptions 1 an 2 above.
+  # Note however that the coordinates of C^n are not one to one to
+  # the homogeneous coordinates of the Cox ring. Rather, the coordinates 
+  # of the affine pieces correspond to the `hilbert_basis(polarize(sigma))`.
+
+  # TODO: In the long run we should think about making the creation of the 
+  # following dictionary lazy. But this requires partial rewriting of the 
+  # ideal sheaves as a whole, so we postpone it for the moment.
+
+  for (k, U) in enumerate(affine_charts(X))
+
+    # We first create the morphism \pi_s* from p. 224, l. 3.
+    indices = [j for j in 1:nrays(X) if ray_indices(maximal_cones(X))[k,j]]
+    help_ring, x_rho = polynomial_ring(QQ, ["x_$j" for j in indices])
+    imgs_phi_star = [j in indices ? x_rho[findfirst(k->k==j, indices)] : one(help_ring) for j in 1:nrays(X)]
+    phi_s_star = hom(cox_ring(X), help_ring, imgs_phi_star)
+
+    # Now we need to create the inverse of alpha*.
+    imgs_alpha_star = elem_type(help_ring)[]
+    for m in hilbert_basis(dual_cone(U))
+      img = one(help_ring)
+      for j in 1:length(indices)
+        u_rho = matrix(ZZ,rays(X))[indices[j],:]
+        expo = ZZ(_my_dot(m, u_rho))
+        img = img * x_rho[j]^expo
+      end
+      push!(imgs_alpha_star, img)
+    end
+    alpha_star = hom(OO(U), help_ring, imgs_alpha_star)
+
+    # TODO: There should be better ways to create this map!
+    # Presumably, one can invert the matrix with the `expo`s as entries?
+    # @larskastner If you can confirm this, let's change this.
+    alpha_star_inv = inverse(alpha_star)
+    ideal_dict[U] = ideal(OO(U), [alpha_star_inv(phi_s_star(g)) for g in gens(I)])
+  end
+
+  return IdealSheaf(X, ideal_dict, check=true) # TODO: Set the check to false eventually.
+end
+
+function _my_dot(v::PointVector{ZZRingElem}, u::ZZMatrix)
+  n = length(v)
+  @req n == length(u) "matrix sizes do not match"
+  return sum(v[i]*u[1,i] for i in 1:n; init=zero(QQ))
+end
+
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1251,7 +1251,7 @@ function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
   # 1. All maximal cones are smooth, i.e. the fan is smooth/X is smooth.
   # 2. The dimension of all maximal cones matches the dimension of the fan.
   @req is_smooth(X) "Currently, ideal sheaves are only supported for smooth toric varieties"
-  @req all(m -> dim(m) == dim(X), maximal_cones(X)) "Currently, ideal sheaves require that all maximal cones have the dimension of the variety"
+  @req ispure(X) "Currently, ideal sheaves require that all maximal cones have the dimension of the variety"
 
   ideal_dict = IdDict{AbsSpec, Ideal}()
 
@@ -1282,7 +1282,7 @@ function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
       img = one(help_ring)
       for j in 1:length(indices)
         u_rho = matrix(ZZ,rays(X))[indices[j],:]
-        expo = ZZ(_my_dot(m, u_rho))
+        expo = (u_rho*m)[1]
         img = img * x_rho[j]^expo
       end
       push!(imgs_alpha_star, img)
@@ -1297,12 +1297,6 @@ function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
   end
 
   return IdealSheaf(X, ideal_dict, check=true) # TODO: Set the check to false eventually.
-end
-
-function _my_dot(v::PointVector{ZZRingElem}, u::ZZMatrix)
-  n = length(v)
-  @req n == length(u) "matrix sizes do not match"
-  return sum(v[i]*u[1,i] for i in 1:n; init=zero(QQ))
 end
 
 function _generic_blow_up(v::NormalToricVarietyType, I::MPolyIdeal)
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1231,12 +1231,7 @@ Create a sheaf of ideals on a toric variety ``X`` from a homogeneous ideal
 julia> P3 = projective_space(NormalToricVariety, 3)
 Normal, non-affine, smooth, projective, gorenstein, fano, 3-dimensional toric variety without torusfactor
 
-julia> (x1,x2,x3,x4) = gens(cox_ring(P3))
-4-element Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}:
- x1
- x2
- x3
- x4
+julia> (x1,x2,x3,x4) = gens(cox_ring(P3));
 
 julia> I = ideal([x2,x3])
 ideal(x2, x3)
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1273,7 +1273,7 @@ function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
 
     # Now we need to create the inverse of alpha*.
     imgs_alpha_star = elem_type(help_ring)[]
-    for m in hilbert_basis(dual_cone(U))
+    for m in hilbert_basis(weight_cone(U))
       img = one(help_ring)
       for j in 1:length(indices)
         u_rho = matrix(ZZ,rays(X))[indices[j],:]
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1263,10 +1263,11 @@ function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
   # following dictionary lazy. But this requires partial rewriting of the 
   # ideal sheaves as a whole, so we postpone it for the moment.
 
+  IM = maximal_cones(IncidenceMatrix, X)
   for (k, U) in enumerate(affine_charts(X))
 
     # We first create the morphism \pi_s* from p. 224, l. 3.
-    indices = [j for j in 1:nrays(X) if ray_indices(maximal_cones(X))[k,j]]
+    indices = [k for k in row(IM, k)]
     help_ring, x_rho = polynomial_ring(QQ, ["x_$j" for j in indices])
     imgs_phi_star = [j in indices ? x_rho[findfirst(k->k==j, indices)] : one(help_ring) for j in 1:nrays(X)]
     phi_s_star = hom(cox_ring(X), help_ring, imgs_phi_star)
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1305,3 +1305,6 @@ function _my_dot(v::PointVector{ZZRingElem}, u::ZZMatrix)
   return sum(v[i]*u[1,i] for i in 1:n; init=zero(QQ))
 end
 
+function _generic_blow_up(v::NormalToricVarietyType, I::MPolyIdeal)
+  return blow_up(IdealSheaf(v, I))
+end
--- a/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
@@ -321,16 +321,33 @@ Multivariate polynomial ring in 5 variables over QQ graded by
   x3 -> [0 1]
   x4 -> [1 0]
   e -> [1 -1]
+
+julia> I2 = ideal([x2 * x3])
+ideal(x2*x3)
+
+julia> b2P3 = blow_up(P3, I2);
+
+julia> codomain(b2P3) == P3
+true
 ```
 """
 function blow_up(v::NormalToricVarietyType, I::MPolyIdeal; coordinate_name::String = "e", set_attributes::Bool = true)
     @req base_ring(I) == cox_ring(v) "The ideal must be contained in the cox ring of the toric variety"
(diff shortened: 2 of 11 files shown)


src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
```diff
@@ -49,7 +107,7 @@ with default covering
     1: normal, affine toric variety
     2: normal, affine toric variety
     3: normal, affine toric variety
-  in the coordinate(s)
+  in the coordinate(s)Od
```
HechtiDerLachs: This doesn't look right.

HereAround: Yes that looks weird. I am surprised it did not lead to errors. I fixed it.

HereAround:
> @HereAround : Can you take it from here?

Working on it.

HereAround:
@lkastner This is the next "round" of toric schemes and their applications in the FTheoryTools.

In this PR, @HechtiDerLachs and I extend the existing `blow_up` functionality for toric varieties. In the current master, these blowups only work provided that the blowup can be conducted (to the best of my knowledge) torically. This PR addresses the complement. Namely, if the blowup cannot (to my knowledge) be conducted torically, then we call the corresponding schemes functionality to perform the blowup with the techniques that @HechtiDerLachs implemented. This is what @apturner is very much looking forward to making use of for the FTheoryTools.

Once the tests complete, this is - as I see it - ready for review.

HechtiDerLachs:
So far, it doesn't seem to be our fault, that the tests fail. One is due to singular conversion of finite fields and the other is due to some doctest in the GIT fans. @jankoboehm : Do you maybe see what's going on?

HereAround:
> So far, it doesn't seem to be our fault, that the tests fail. One is due to singular conversion of finite fields and the other is due to some doctest in the GIT fans. @jankoboehm : Do you maybe see what's going on?

At least one failure happened because of me - I wrote `test` instead of `testset` by accident. Let see what the tests say now.

HereAround:
It seems that now all required tests pass but the doctest (cf. https://github.com/oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354055?pr=2741). I am not sure (yet) why the latter failed...

benlorenz:
> It seems that now all required tests pass but the doctest (cf. [oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354055?pr=2741](https://github.com/oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354055?pr=2741)). I am not sure (yet) why the latter failed...

There is a whitespace difference in the output here: https://github.com/oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354443?pr=2741#step:8:2020

```
┌ Error: doctest failure in ~/work/Oscar.jl/Oscar.jl/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl:300-353
...
│ Subexpression:
│ 
│ I2 = ideal([x2 * x3])
│ 
│ Evaluated output:
│ 
│ ideal(x2*x3)
│ 
│ Expected output:
│ 
│ ideal(x2 * x3)
│ 
│   diff = ideal(x2 * x3)ideal(x2*x3)
└ @ Documenter.DocTests 
```

Unfortunately that error block is now somewhat hard to find due to the extra output. I did not expect the doctests to continue running when encountering an error...
I will have a look if we can improve this.

HereAround:
> > It seems that now all required tests pass but the doctest (cf. [oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354055?pr=2741](https://github.com/oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354055?pr=2741)). I am not sure (yet) why the latter failed...
> 
> There is a whitespace difference in the output here: https://github.com/oscar-system/Oscar.jl/actions/runs/6027670947/job/16353354443?pr=2741#step:8:2020
> 
> ```
> ┌ Error: doctest failure in ~/work/Oscar.jl/Oscar.jl/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl:300-353
> ...
> │ Subexpression:
> │ 
> │ I2 = ideal([x2 * x3])
> │ 
> │ Evaluated output:
> │ 
> │ ideal(x2*x3)
> │ 
> │ Expected output:
> │ 
> │ ideal(x2 * x3)
> │ 
> │   diff = ideal(x2 * x3)ideal(x2*x3)
> └ @ Documenter.DocTests 
> ```
> 
> Unfortunately that error block is now somewhat hard to find due to the extra output. I did not expect the doctests to continue running when encountering an error... I will have a look if we can improve this.

Thank you. Fixed and just updated the code.

For my curiosity: How did you find this? I looked at the log from top to bottom but did not spot this. Did I just overlook it?

benlorenz:
I searched for failure in the box on the top right (with the corresponding log block expanded).

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+```
+"""
+function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
+  @req base_ring(I) === cox_ring(X) "ideal must live in the cox ring of the variety"
+
+  # We currently only support this provided that the following conditions are met:
+  # 1. All maximal cones are smooth, i.e. the fan is smooth/X is smooth.
+  # 2. The dimension of all maximal cones matches the dimension of the fan.
+  @req is_smooth(X) "Currently, ideal sheaves are only supported for smooth toric varieties"
+  @req all(m -> dim(m) == dim(X), maximal_cones(X)) "Currently, ideal sheaves require that all maximal cones have the dimension of the variety"
```
benlorenz:
```suggestion
  @req ispure(X) "Currently, ideal sheaves require that all maximal cones have the dimension of the variety"
```
That should create less intermediate objects.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+  # correspond to the `hilbert_basis(dual_cone(U_s))`.
+
+  # TODO: In the long run we should think about making the creation of the 
+  # following dictionary lazy. But this requires partial rewriting of the 
+  # ideal sheaves as a whole, so we postpone it for the moment.
+
+  for (k, U) in enumerate(affine_charts(X))
+
+    # We first create the morphism Φₛ* from p. 224, l. 3.
+    indices = [j for j in 1:nrays(X) if ray_indices(maximal_cones(X))[k,j]]
```
lkastner: This is again an abuse of `IncidenceMatrix`.
```
julia> nf = normal_fan(cube(2))
Polyhedral fan in ambient dimension 2

julia> IM = maximal_cones(IncidenceMatrix, nf)
4×4 IncidenceMatrix
[1, 3]
[2, 3]
[1, 4]
[2, 4]


julia> row(IM, 1)
Set{Int64} with 2 elements:
  3
  1

julia> 1 in row(IM, 1)
true

julia> 2 in row(IM, 1)
false
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+    help_ring, x_rho = polynomial_ring(QQ, ["x_$j" for j in indices])
+    imgs_phi_star = [j in indices ? x_rho[findfirst(k->k==j, indices)] : one(help_ring) for j in 1:nrays(X)]
+    phi_s_star = hom(cox_ring(X), help_ring, imgs_phi_star)
+
+    # Now we need to create the inverse of α*.
+    imgs_alpha_star = elem_type(help_ring)[]
+    for m in hilbert_basis(dual_cone(U))
```
lkastner: Why is there something called `dual_cone`? It was agreed upon that the function taking cones to their (polar) dual is called `polarize`. But this function cannot even be applied to cones, only to affine normal toric varieties. So it is even worse, it will confuse users who only want to dualize a polyhedral cone and users having an affine normal toric variety will not know that this might be the function they are looking for.

There are better names that one can choose, please have a look in CLS. In polymake we called this the `WEIGHT_CONE`, but also something like `semigroup_cone` or `character_cone` could probably work. `dual_cone` is definitely not ok.

lkastner: What if this cone has lineality?
```
julia> c = positive_hull([1 0; -1 0])
Polyhedral cone in ambient dimension 2

julia> hilbert_basis(c)
ERROR: ArgumentError: Cone not pointed
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/Hecke/xt25Z/src/Assertions.jl:564 [inlined]
 [2] hilbert_basis(C::Cone{QQFieldElem})
   @ Oscar ~/oscars/master/src/PolyhedralGeometry/Cone/properties.jl:529
 [3] top-level scope
   @ REPL[14]:1

```
I think you probably also want to require at least `isfulldimensional` and maybe also `ispointed`.

experimental/Schemes/IdealSheaves.jl
HereAround: Thank you.

HereAround:
@HechtiDerLachs, based on our conversations in slack, I now take the liberty to mark this as ready for review. Please correct me if needed.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+
+function _my_dot(v::PointVector{ZZRingElem}, u::ZZMatrix)
+  n = length(v)
+  @req n == length(u) "matrix sizes do not match"
+  return sum(v[i]*u[1,i] for i in 1:n; init=zero(QQ))
+end
+
+function _generic_blow_up(v::Any, I::Any)
+  error("Not yet supported")
+end
```
lkastner:
```suggestion
```
Why is this needed?

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+  @req all(m -> dim(m) == dim(X), maximal_cones(X)) "Currently, ideal sheaves require that all maximal cones have the dimension of the variety"
+
+  ideal_dict = IdDict{AbsSpec, Ideal}()
+
+  # We need to dehomogenize the ideal I ⊂ S = ℂ[xᵨ| ρ ∈ Σ(1)] to the local 
+  # charts Uₛ, s ⊂ Σ a cone. To this end we use Proposition 5.2.10 from 
+  # Cox-Little-Schenck, p. 223ff. Abstractly, the chart 
+  # Uₛ is isomorphic to ℂˢ⁽¹⁾ = Spec(ℂ[xᵨ | ρ ∈ s(1)]). But these are 
+  # not the coordinates we use for the affine charts here. They 
+  # correspond to the `hilbert_basis(dual_cone(U_s))`.
```
lkastner: https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/
> the use of Unicode should be kept to a minimum.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+```jldoctest
+julia> P3 = projective_space(NormalToricVariety, 3)
+Normal, non-affine, smooth, projective, gorenstein, fano, 3-dimensional toric variety without torusfactor
+
+julia> (x1,x2,x3,x4) = gens(cox_ring(P3))
+4-element Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}:
+ x1
+ x2
+ x3
+ x4
```
lkastner:
```suggestion
julia> (x1,x2,x3,x4) = gens(cox_ring(P3));
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1210,3 +1210,105 @@ function saturation(I::IdealSheaf, J::IdealSheaf)
...
+  on scheme over QQ covered with 4 patches
+    1: [x_1_1, x_2_1, x_3_1]   normal, affine toric variety
+    2: [x_1_2, x_2_2, x_3_2]   normal, affine toric variety
+    3: [x_1_3, x_2_3, x_3_3]   normal, affine toric variety
+    4: [x_1_4, x_2_4, x_3_4]   normal, affine toric variety
+with restrictions
+  1: ideal(x_2_1, x_3_1)
+  2: ideal(x_2_2, x_3_2)
+  3: ideal(1, x_3_3)
+  4: ideal(x_2_4, 1)
```
lkastner: I don't think this output is sustainable for larger examples.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
```diff
@@ -321,19 +321,55 @@ Multivariate polynomial ring in 5 variables over QQ graded by
...
+    2a: [x_1_2, x_2_2, x_3_2]   spec of quotient of multivariate polynomial ring
+    3a: [x_1_3, x_2_3, x_3_3]   spec of quotient of multivariate polynomial ring
+    4a: [x_1_4, x_2_4, x_3_4]   spec of quotient of multivariate polynomial ring
+and exceptional divisor
+  effective cartier divisor defined by
+    sheaf of ideals with restrictions
+      1a: ideal(x_2_1*x_3_1)
+      2a: ideal(x_2_2*x_3_2)
+      3a: ideal(x_3_3)
+      4a: ideal(x_2_4)
```
lkastner: This escalated quickly...

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
```diff
@@ -1,5 +1,96 @@
...
+  end
+
+  # Recreate all the glueings with the new patches
+  for U in affine_charts(X)
+    for V in affine_charts(X)
+      glue = default_covering(X)[U, V]
+      new_glue = restrict(glue, inverse(iso_dict[U]), inverse(iso_dict[V]), check=true)
+      add_glueing!(cov, new_glue)
+    end
+  end
```
lkastner: https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/#Code-structure

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
```diff
@@ -321,19 +321,55 @@ Multivariate polynomial ring in 5 variables over QQ graded by
...
 """
 function blow_up(v::NormalToricVarietyType, I::MPolyIdeal; coordinate_name::String = "e", set_attributes::Bool = true)
     @req base_ring(I) == cox_ring(v) "The ideal must be contained in the cox ring of the toric variety"
     indices = [findfirst(y -> y == x, gens(cox_ring(v))) for x in gens(I)]
-    @req length(indices) == ngens(I) "All generators must be indeterminates of the cox ring of the toric variety"
-    rs = matrix(ZZ, rays(v))
-    new_ray = vec(sum([rs[i,:] for i in indices]))
-    new_ray = new_ray ./ gcd(new_ray)
-    return blow_up(v, new_ray; coordinate_name = coordinate_name, set_attributes = set_attributes)
+    if length(indices) == ngens(I) && (nothing in indices) == false
```
lkastner:
```suggestion
    if length(indices) == ngens(I) && !(nothing in indices)
```

lkastner (changes requested):
This is moving a bit too fast for me to keep up with. There is already a lot to unpack below the mathematical level. Try to add more tests for boundary cases.

experimental/Schemes/IdealSheaves.jl
HereAround: Ok, let me simply suppressed the output for now:
```julia
julia> IdealSheaf(P3, I);
```

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
HereAround: I guess the following is better?
```julia
julia> b2P3 = blow_up(P3, I2);

julia> codomain(b2P3) == P3
true
```

experimental/Schemes/IdealSheaves.jl
HereAround: Let me modify this... (cc @HechtiDerLachs).

src/AlgebraicGeometry/ToricVarieties/ToricSchemes/attributes.jl
HereAround: Thank you for the pointer. Adjusted to
```julia
for U in affine_charts(X), V in affine_charts(X)
  ...
end
```
I tried also
```julia
for U,V in affine_charts(X)
  ...
end
```
But this errored with "invalid iteration specification".

experimental/Schemes/IdealSheaves.jl
HereAround: I am rebasing to work in those changes and will push shortly. Please inspect again and let me know if this is better.

My understanding is that @HechtiDerLachs uses Unicode for the documentation and comments in the schemes code. Certainly, I have only touched the lines about the ideal sheaves added in this PR.

experimental/Schemes/IdealSheaves.jl
HereAround: The code on ideal sheaves is still experimental, while the `blow_up` function in the toric varieties is in `src`.

I understand that there was a rule "`src` must never call `experimental`" or similar. I discussed this with @fingolfin yesterday, and he proposed this work-around. It ensure that even if you disabled all `experimental` code, the toric varieties code would still run except for raising the error "Not yet supported" exactly where the call into `experimental` would have happened.

lkastner: I don't understand. If this wasn't here you would also get an error, right? Also this function is in `experimental`, so it would be disabled too if `experimental` is gone?

experimental/Schemes/IdealSheaves.jl
HereAround: My recollection is a bit vague - I believe this dates back probably half a year or even longer, when we first prototyped the toric schemes. The ideas did not really go very deep, but focused on organizing the code with some helper functions. Nothing more than that. This is I think why this exists in the first place.

Certainly, I am more than happy to rename it. Based on what you are saying, maybe it would be best to align it with the polymake name, i.e. `weight_cone`. Let me prepare that...

experimental/Schemes/IdealSheaves.jl
HereAround: I also noticed this and adjusted it. Let me push the changes that I have by now. One sec.

HereAround: If I did it correctly, then this function should now be in `standard_constructions.jl`, right below the `blow_up` function.

HereAround: And yes, if this function was not there, we would also get an error ("something undefined" or alike). Maybe the "Not yet supported" error is better? Maybe not? Maybe the error message could be adjusted...

As per usual - I am happy to hear other/better alternatives. :)

experimental/Schemes/IdealSheaves.jl
HereAround: Proposal is here 0f159c61f68590d9efcaea5bdf8769b45c4f5c59. What do you think?

HereAround: Thank you for the pointers. Proposal for resolution is here: 0cc754cb79d328ba1f8332d2fd5cb28cb63a3ae5

experimental/Schemes/IdealSheaves.jl
HereAround: Sorry for this abuse. I of course read your example and adjusted the code a bit. But I am not sure if I did grasp the full scale of how this can be optimized with your proposal. Please have a look at 730c7e0dcede5abd68c8ec21baa551960688147f and let me know what you think.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: > https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/
> 
> > the use of Unicode should be kept to a minimum.

I'm using the notation from the book and tried to stick as close to it as possible so people do not have to go through the hustle to write down a dictionary to translate the comment to the setup of CLS. I think this is a justified use of unicode.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: > I don't think this output is sustainable for larger examples.

This has been thought through by @simonbrandhorst and thoroughly worked out by @StevellM . If you have doubts about this, we can contact them, but let's not open this discussion on this pr.

simonbrandhorst: If an object is complicated, it may take much space to print. If the output is too large, one can still use a semicolon to suppress it or ask for compact printing.

lkastner (approved):
I still wonder what the use case for the `forget_*` function is. It is a very technical thing and I do not think that users should worry about such a thing. Furthermore I see it more as a constructor, in the same way one can call `polyhedral_fan` on a toric variety, I would think the name `covered_scheme(*)` or something similar would be better suited.

HechtiDerLachs:
Thank you for the review @lkastner ! 

I understand that `forget_toric_structure` might seem unintuitive. But it was on @HereAround 's wishlist in this specific form, so I did it. Calling it `covered_scheme(...)`, however, is not an option: As of now, an `AbstractNormalToricVariety` is already an `AbsCoveredScheme`, so asking for one should return the same object. `underlying_scheme` on the other hand did not satisfy @HereAround 's requirements.

HereAround:
Thank you for the review @lkastner and approving this complicated PR.

Regarding `forget_toric_structure`, I do not have a use case just yet. In essence, I asked @HechtiDerLachs in order to (hopefully soon) satisfy my curiosity. As I see it, it should be interesting to compare how the schemes technology operates (e.g. on blowups) without remembering at all that the scheme/variety is/was toric, and then to compare this with the toric procedure. I expect that exactly this might be useful for @apturner to move forward with a sort of abstract blowup, that is yet to be drafted, discussed etc.

Long story short: No use case just yet. However, hopefully soon to be used as inspiration for further developments.

Based on the approval, I assume we can live with this for now. However, should this becomes relevant on a larger scale (i.e. implementing forgetful functors throughout OSCAR), then we can discuss again, adjust the name etc. As per usual, I am happy to adjust the name etc.

PR #2744, opened 2023-08-30
fixes saving a filename without a dirname

antonydellavecchia:
quick fix

### DIFF
--- a/src/Serialization/main.jl
+++ b/src/Serialization/main.jl
@@ -484,7 +484,11 @@ function save(io::IO, obj::T; metadata::Union{MetaData, Nothing}=nothing) where 
 end
 
 function save(filename::String, obj::Any; metadata::Union{MetaData, Nothing}=nothing)
-  temp_file = tempname(dirname(filename))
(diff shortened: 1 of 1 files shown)


src/Serialization/main.jl
```diff
@@ -484,7 +484,11 @@ function save(io::IO, obj::T; metadata::Union{MetaData, Nothing}=nothing) where
 end
 
 function save(filename::String, obj::Any; metadata::Union{MetaData, Nothing}=nothing)
-  temp_file = tempname(dirname(filename))
+  dir_name = dirname(filename)
+  if dir_name == ""
+    dir_name = pwd()
+  end
+  temp_file = tempname(dir_name)
```
benlorenz:
```suggestion
  dir_name = dirname(filename)
  # julia dirname does not return "." for plain filenames without any slashes
  temp_file = tempname(isempty(dir_name) ? pwd() : dir_name)
```

benlorenz (approved):

PR #2747, opened 2023-08-30
Refactor `canonical_{in/pro}jection`

lgoettgens:
Adds methods to get all in/projections for direct sums of modules, like already in Hecke for `GrpAbFinGen`. Some more readability enhancements.

This needs https://github.com/thofma/Hecke.jl/pull/1196 to be available, thus the bump in the Hecke compat in anticipation of a future merge.


lgoettgens:
For `DirectProductGroup`, the corresponding functions are named `embedding` and `projection`. I think it would be good to have the names aligned here, i.e. rename them to the ones already in place for `GrpAbFinGen` and modules. I would furthermore add methods for returning all in/projections, as above.

Maybe @fingolfin and @ThomasBreuer have an opinion on this?

src/Modules/FreeModules-graded.jl
```diff
@@ -1214,25 +1214,32 @@ function direct_product(G::ModuleFP_dec...; task::Symbol = :none)
...
+end
 
-function Hecke.canonical_injection(G::ModuleFP_dec, i::Int)
+function canonical_injection(G::ModuleFP_dec, i::Int)
   H = get_attribute(G, :direct_product)
-  if H === nothing
-    error("module not a direct product")
-  end
-  0<i<= length(H) || error("index out of bound")
-  j = i == 1 ? 0 : sum(ngens(H[l]) for l=1:i-1) -1
```
lgoettgens, line 1224: The `-1` here at the end of this line seems wrong to me, so I removed it. It has been introduced by @fieker in https://github.com/oscar-system/Oscar.jl/commit/ecfbd92a771ba141f50732aa633cfbcf7345a8e6 (a long time ago). Could you please have a look if that has been a typo (that will be fixed by this) or if this introduces faulty behavior?

experimental/GModule/GaloisCohomology.jl
```diff
@@ -1081,15 +1081,15 @@ function Oscar.completion(I::IdelParent, P::NfAbsOrdIdl)
...
+  inj = canonical_injection(I.M, p+1) #units are first
+  pro = canonical_projection(I.M, p+1)
 
 
   @assert domain(inj) == codomain(pro)
 
   J = components(I.M)[p+1]
   if mKp.P == P #easy case
-    return Kp, mKp, mGp, mUp,  pro * Hecke.canonical_projection(J, 1) ,  Hecke.canonical_injection(J, 1)*inj
+    return Kp, mKp, mGp, mUp,  pro * canonical_projection(J, 1) ,  canonical_injection(J, 1)*inj
```
fingolfin: Not your "fault", but while we are at it:
```suggestion
    return Kp, mKp, mGp, mUp, pro * canonical_projection(J, 1), canonical_injection(J, 1)*inj
```

src/Modules/FreeModules-graded.jl
fingolfin: @fieker ping

fingolfin (approved):
Fine by me, but I think @fieker should have a look.

lgoettgens:
@fingolfin What do you think about renaming the functions for `DirectProductGroup` (see https://github.com/oscar-system/Oscar.jl/pull/2747#issuecomment-1700714017)? If that should be tackled, I could include it in this PR.

fingolfin:
I think renaming the functions for `DirectProductGroup` would make sense, but it shouldn't be in this PR -- we don't want to hold it up unnecessarily...

fingolfin (approved):

lgoettgens:
> I think renaming the functions for `DirectProductGroup` would make sense, but it shouldn't be in this PR -- we don't want to hold it up unnecessarily...

Fine for me, I'll provide that in a new pr.

Did you talk to fieker concerning the index?

PR #2748, opened 2023-08-30
some methods for computing orth. discriminants

ThomasBreuer:
- add `od_from_order`, `od_from_eigenvalues`, `od_for_specht_module`
- add utilities `character_of_entry`, `order_omega_mod_N`, `reduce_mod_squares`

### DIFF
--- a/experimental/OrthogonalDiscriminants/src/theoretical.jl
+++ b/experimental/OrthogonalDiscriminants/src/theoretical.jl
@@ -1,0 +1,174 @@
+
+# character-theoretical methods
+
+
+@doc raw"""
+    od_from_order(chi::GAPGroupClassFunction)
+
+Return `(flag, val)` where `flag` is `true` if the order of the group
+of `chi` divides only one of the orders of the two orthogonal groups
+[`omega_group`](@ref)`(+/-1, d, q)`, where `d` is the degree of `chi`
+and `q` is the order of the field of definition of `chi`.
+
+In this case, `val` is `"O+"` or `"O-"`.
+
+```jldoctest
+julia> t = character_table("L3(2)");
+
+julia> Oscar.OrthogonalDiscriminants.od_from_order(mod(t, 3)[4])
+(true, "O-")
+
+julia> Oscar.OrthogonalDiscriminants.od_from_order(mod(t, 2)[4])
+(false, "")
+```
+"""
+function od_from_order(chi::GAPGroupClassFunction)
+  characteristic(chi) == 0 && return (false, "")
+  d = numerator(degree(chi))
+  q = order_field_of_definition(chi)
+  tbl = ordinary_table(chi.table)
+  ord = order(ZZRingElem, tbl)
+
+  # Compute the order of the subgroup that shall embed into
+  # the perfect group `omega_group(epsilon, d, q)`.
+  n = sum(class_lengths(tbl)[class_positions_of_solvable_residuum(tbl)])
+  flag1, flag2 = order_omega_mod_N(d, q, n)
+  if flag1
+    return flag2 ? (false, "") : (true, "O+")
+  else
+    return flag2 ? (true, "O-") : (false, "")
+  end
+end
+
+
+@doc raw"""
+    od_from_eigenvalues(chi::GAPGroupClassFunction)
+
+Return `(flag, val)` where `flag` is `true` if there is a conjugacy class
+on which representing matrices for `chi` have no eigenvalue $\pm 1$.
+In this case, if `chi` is orthogonally stable (this is not checked here)
+then `val` is a string that describes the orthogonal discriminant of `chi`.
+
+If `flag` is `false` then `val` is equal to `""`.
+
+This criterion works only if the characteristic of `chi` is not $2$,
+`(false, "")` is returned if the characteristic is $2$.
+
+# Examples
+```jldoctest
+julia> t = character_table("A5");
+
+julia> Oscar.OrthogonalDiscriminants.od_from_eigenvalues(t[4])
+(true, "5")
+
+julia> Oscar.OrthogonalDiscriminants.od_from_eigenvalues(mod(t, 3)[4])
+(true, "O-")
+
+julia> Oscar.OrthogonalDiscriminants.od_from_eigenvalues(mod(t, 2)[4])
+(false, "")
+```
+"""
+function od_from_eigenvalues(chi::GAPGroupClassFunction)
+  p = characteristic(chi)
+  p == 2 && return (false, "")
+
+  tbl = chi.table
+  ord = orders_class_representatives(tbl)
+  for i in 2:length(chi)
+    n = ord[i]
+    ev = multiplicities_eigenvalues(chi, i)
+    if ev[end] != 0 || (iseven(n) && ev[divexact(n, 2)] != 0)
+      continue
+    end
+
+    F, z = cyclotomic_field(n)
+    od = prod(x -> x[1]^x[2], [(z^i-z^-i, ev[i]) for i in 1:n])
+    if mod(degree(chi), 4) == 2
+      od = -od
+    end
+
+    K, _ = abelian_closure(QQ)
+    if p == 0
+      # Coerce `od` into the character field of `chi`.
+      F, emb = character_field(chi)
+      od = preimage(emb, K(od))
+
+      # Reduce the representative `od` mod obvious squares
+      # in the character field.
+      od = reduce_mod_squares(F, od)
+
+      # Embed this value into the alg. closure.
+      od = emb(od)
+
+      # Turn the value into a string (using Atlas notation).
+      str = atlas_description(od)
+    else
+      # Decide if the reduction mod `p` is a square in the char. field.
+      str = is_square(reduce(K(od), character_field(chi)[1])) ? "O+" : "O-"
+    end
+
+    return true, str
+  end
+
+  return false, ""
+end
+
+@doc raw"""
+    od_for_specht_module(chi::GAPGroupClassFunction)
+
+Return `(flag, val)` where `flag` is `true` if `chi` is an ordinary
+irreducible character of a symmetric group or of an alternating group
+such that `chi` extends to the corresponding symmetric group.
+In this case, if `chi` is orthogonally stable (this is not checked here)
+then `val` is a string that describes the orthogonal discriminant of `chi`;
+the discriminant is computed using the Jantzen-Schaper formula,
+via [`gram_determinant_specht_module`](@ref).
+
+`(false, "")` is returned in all cases where this criterion is not applicable.
+
+# Examples
+```jldoctest
+julia> t = character_table("A5");
+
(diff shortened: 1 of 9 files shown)


experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+
+julia> Oscar.OrthogonalDiscriminants.reduce_mod_squares(F, 4*z^0)
+1
+
+julia> Oscar.OrthogonalDiscriminants.reduce_mod_squares(F, -8*z^0)
+-2
+```
+"""
+function reduce_mod_squares(F::AnticNumberField, val::nf_elem)
+  @req parent(val) === F "val must have parent F"
```
fingolfin: Why is `F` even an argument when the only place using it is this check?

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+  is_zero(val) && return val
+  d = denominator(val)
+  if ! isone(d)
+    val = val * d^2
```
fingolfin, line 78:
```suggestion
    val *= d^2
```

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+  end
+  if is_integer(val)
+    val = ZZ(val)
```
fingolfin: It may not matter, but just FYI, this breaks type stability (and I think possibly even of the whole function?) for `val` if it is not already a `ZZRingElem`.

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+    sgn = sign(val)
+    good = filter(x -> is_odd(x[2]), collect(factor(val)))
+    return F(prod([x[1] for x in good], init = sgn))
```
fingolfin: The following might avoid the second allocation. Dunno if it makes things better or not, though...
```suggestion
    good = [x[1] for x in collect(factor(val)) if is_odd(x[2])]
    return F(prod(good, init = sgn))
```

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+  c = map(numerator, coefficients(val))
+  s = 1
+  for (p, e) in collect(factor(gcd(c)))
+    if iseven(e)
+      s = s * p^e
+    elseif e > 1
+      s = s * p^(e-1)
+    end
+  end
+  return F(c // s)
```
fingolfin: I am confused, isn't `c` a vector at this point?

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+    sgn = sign(val)
+    good = filter(x -> is_odd(x[2]), collect(factor(val)))
+    return F(prod([x[1] for x in good], init = sgn))
+  end
+  # Just get rid of the square part of the gcd of the coefficients.
+  c = map(numerator, coefficients(val))
+  s = 1
+  for (p, e) in collect(factor(gcd(c)))
+    if iseven(e)
+      s = s * p^e
```
fingolfin, line 92:
```suggestion
      s *= p^e
```

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+    elseif e > 1
+      s = s * p^(e-1)
```
fingolfin, line 94:
```suggestion
      s *= p^(e-1)
```

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+  end
+  if m*(m-1) < exp
+    # A group of order `N` does not embed in any candidate.
+    return (false, false)
+  end
+
+  q2 = ZZ(q)^2
+  q2i = ZZ(1)
+  for i in  1:(m-1)
+    q2i = q2 * q2i
```
fingolfin, line 35:
```suggestion
    q2i *= q2
```

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+julia> Oscar.OrthogonalDiscriminants.order_omega_mod_N(4, 5, 60)
+(true, true)
+```
+"""
+function order_omega_mod_N(d::IntegerUnion, q::IntegerUnion, N::IntegerUnion)
+  @req is_even(d) "d must be even"
+  m = div(d, 2)
+  exp = 0
+  while mod(N, q) == 0
+    exp = exp + 1
```
fingolfin:
```suggestion
    exp += 1
```

experimental/OrthogonalDiscriminants/src/theoretical.jl
```diff
@@ -0,0 +1,174 @@
...
+  chipos == nothing && return (false, "")
+  para = character_parameters(tbl)[chipos]
+  isa(para, Vector{Int}) || return (false, "")
+
+  # Now we know that `chi` belongs to Sym(n) or extends to Sym(n)
+  gramdet = gram_determinant_specht_module(partition(para))
+  res = 1
+  for pair in gramdet
+    if is_odd(pair[2])
+      res = res * pair[1]
```
fingolfin, line 166:
```suggestion
      res *= pair[1]
```

experimental/OrthogonalDiscriminants/src/theoretical.jl
```diff
@@ -0,0 +1,174 @@
...
+  name == "A$n" || name == "A$n.2" || name == "A6.2_1" || return (false, "")
+
+  chipos = findfirst(isequal(chi), tbl)
+  chipos == nothing && return (false, "")
+  para = character_parameters(tbl)[chipos]
+  isa(para, Vector{Int}) || return (false, "")
+
+  # Now we know that `chi` belongs to Sym(n) or extends to Sym(n)
+  gramdet = gram_determinant_specht_module(partition(para))
+  res = 1
```
fingolfin: Is this type stable?

fingolfin (approved):
Fine by me, but some questions and suggestions -- but all optional, feel free to ignore some or all.

experimental/OrthogonalDiscriminants/src/utils.jl
ThomasBreuer: Yes.
I still have to get used to the fact that the elements carry the relevant context.

experimental/OrthogonalDiscriminants/src/utils.jl
ThomasBreuer: Do you want to say that this syntax is preferable?
If yes then should this be stated in the "Deveoper Style Guide"? (There are quite some places in the code where one could switch to this syntax.)

experimental/OrthogonalDiscriminants/src/utils.jl
ThomasBreuer: I do not understand this comment. In the end, an element of `F` is returned.

experimental/OrthogonalDiscriminants/src/utils.jl
ThomasBreuer: Yes, we want the vector where each entry is divided by `s`.
Perhaps it is better to divide `val` by `s`.

experimental/OrthogonalDiscriminants/src/theoretical.jl
ThomasBreuer: Better `ZZ(1)`.
(However, for the return value of the function, this does not matter because we call `string(res)`.)

thofma:
How unique does your reduction modulo squares need to be? There is an (unexported) `Hecke.reduce_mod_powers(::nf_elem, ::Int)`, which also tries to reduce something modulo powers.

ThomasBreuer:
@thofma Thanks for the hint.
> How unique does your reduction modulo squares need to be?

If the value is in fact in the prime field then I need that the reduction is represented by a squarefree rational integer.
Otherwise I cannot expect uniqueness, and what `Hecke.reduce_mod_powers` promises seems to be useful.

experimental/OrthogonalDiscriminants/src/utils.jl
fingolfin: From my POV either is fine, this is a matter of taste. Pick whichever you prefer :-)

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+    N = div(N, q)
+  end
```
fingolfin: Actually, I think it might be possible to replace this by
```suggestion
  exp, N = remove(N, q)
```

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+  facts = collect(factor(q))
+  p = facts[1][1]
+  if mod(N, p) == 0
+    exp = exp + 1
```
fingolfin, line 24: Huh, so first `exp` was defined as "`q^exp` is largest power of `q` dividing the original `N`. But after this loop, `exp = exp_q + exp_p`, where `original_N = q^exp_q * p^exp_p * N'` and `N'` is coprime to `p`. Is that intentional? It jus seems so unusual.

fingolfin: Oh no, I totally misread that: if there are `p` divisors left, you only increment `exp ` once. Nevermind.

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -0,0 +1,103 @@
...
+    while mod(N, p) == 0
+      N = div(N, p)
+    end
```
fingolfin:
```suggestion
    _, N = remove(N, p)
```

experimental/OrthogonalDiscriminants/src/utils.jl
fingolfin: The code inside this function is unstable: Julia cannot determine a concrete type for `val` anymore: `val` was initially an `nf_elem`, but now is a `ZZRingElem` -- so the best it can do is to say that `val` is a `Union{nf_elem, ZZRingElem}` in this function. You can see this in `@code_warntype`.

This kind of issue can percolate: Julia may or may not be able to deduce which `sign` method is called, and thus may or may not be able to predict the type of `sgn`, and then `good`, and finally for the return value.

In this case, it *seems* to be OK as the union splitting code in the Julia compiler can deal with it. But in general I would recommend to avoid re-using a variable for a value of a different type (instead call it `val2` or whatever)

PR #2750, opened 2023-08-31
serialize ZZModRing (note the capital ZZ)

antonydellavecchia:
This type was missing only had the int version

### DIFF
--- a/src/Serialization/Rings.jl
+++ b/src/Serialization/Rings.jl
@@ -63,11 +63,13 @@ end
 @register_serialization_type ZZRing
 
 ################################################################################
-#  non simpleton base rings
+#  Mod Rings
 @register_serialization_type Nemo.zzModRing
+@register_serialization_type Nemo.ZZModRing
+const ModRingUnion = Union{Nemo.zzModRing, Nemo.ZZModRing}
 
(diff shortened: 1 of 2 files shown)


src/Serialization/Rings.jl
```diff
@@ -63,27 +63,36 @@ end
...
 
-function save_object(s::SerializerState, x::zzModRingElem)
+function save_object(s::SerializerState, x::ModRingElemUnion)
   save_data_basic(s, string(x))
 end
 
-function load_object(s::DeserializerState, ::Type{zzModRingElem},
-                                 str::String, parent_ring::Nemo.zzModRing)
+function load_object(s::DeserializerState, ::Type{<:ModRingElemUnion},
+                                 str::String, parent_ring::T) where T <: ModRingUnion
```
fingolfin, line 95: Shouldn't this be
```suggestion
function load_object(s::DeserializerState, ::Type{T},
                                 str::String, parent_ring::T) where T <: ModRingUnion
```

antonydellavecchia: no we want to load the element, which should belong to the parent_ring being passed so we need to pass the elem type as the second argument. Just for some reassurance the tests wouldn't pass if this wasn't correct

benlorenz (approved):

PR #2754, opened 2023-09-01
type -> _type, refs -> _refs

antonydellavecchia:
renaming of properties. using the underscore helps when sorting to reorder underscore properties at the top.
This also aligns with the polymake format


src/Serialization/upgrades/0.13.0.jl
```diff
@@ -67,38 +67,36 @@ push!(upgrade_scripts_set, UpgradeScript(
...
     # type has already been updated
-    if dict[:type] isa Dict
+    if haskey(dict, :_type) && dict[:_type] isa Dict
       return dict
     end
     if T <: Vector
       # this section wasn't necessary for upgrading the folder of surfaces
       # which was our primary focus, this may be updated upon request
-      throw(Error("The upgrade script needs an update"))
+      throw(error("The upgrade script needs an update"))
```
benlorenz:
```suggestion
      error("The upgrade script needs an update")
```

src/Serialization/upgrades/0.13.0.jl
```diff
@@ -67,38 +67,36 @@ push!(upgrade_scripts_set, UpgradeScript(
...
     elseif T <: Union{PolyRingElem, MPolyRingElem}
       if dict[:data] isa Dict && haskey(dict[:data], :parents)
         upgraded_parents = dict[:data][:parents][end]
         params = upgraded_parents
         upgraded_dict[:data] = upgrade_terms(dict[:data][:terms])
       else
-        # this section wasn't necessary for upgrading the folder of surfaces
-        # which was our primary focus, this may be updated upon request
-        throw(Error("The upgrade script needs an update"))
+        println(json(dict, 2))
```
benlorenz: Leftover from some debugging?

benlorenz (commented):
Maybe we should make these const variables (and use those instead of repeating the symbols everywhere):
```
const type_key = :_type
const refs_key = :_refs
```

benlorenz (approved):

PR #2757, opened 2023-09-03
Format proj() to the Oscar standard

mgemath:
Here I format the file containing the function proj to the Oscar format standard.
Moreover, I cleaned the documentation a little bit.
Finally, I add the function proj() [with no arguments] so the user is invited to add some arguments.


HereAround (approved):
Thank you @mgemath . Looks good to me!

PR #2760, opened 2023-09-04
Two neighbor cleanup

HechtiDerLachs:
CC: @simonbrandhorst

### DIFF
--- a/experimental/Schemes/Auxiliary.jl
+++ b/experimental/Schemes/Auxiliary.jl
@@ -96,6 +96,14 @@ function pullback(f::AbsCoveredSchemeMorphism, II::IdealSheaf)
   return IdealSheaf(X, ID, check=false)
 end
 
+function pullback(f::CompositeCoveredSchemeMorphism, C::EffectiveCartierDivisor)
+  result = C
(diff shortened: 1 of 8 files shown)


simonbrandhorst:
Thanks for doing this! Looks good to me once the tests pass.

simonbrandhorst (approved):

HechtiDerLachs:
Note that now the `weierstrass_chart` of an `EllipticSurface` `X` does not live on the `weierstrass_model`, but on the `relatively_minimal_model` -- which is nothing but `X` itself by now. This second thing is the main inconsistency which has been removed here: We already returned the `relatively_minimal_model` as the `underlying_scheme` of `X`, but didn't use `X` as the `domain` of the blowdown morphism. This has been corrected. However, we might want to reconsider whether a getter like `relatively_minimal_model` is then still suitable. I think what we really need there is something like  `projection_to_weierstrass_model`.

HechtiDerLachs:
Ready for merge from my side.

simonbrandhorst:
Yea, I have pondered about this myself too. I agree that `projection_to_weierstrass_model` is probably better.

PR #2763, opened 2023-09-04
Make `check` as a keyword argument

StevellM:
It should have been a semicolon since the implementation of the code, but looks like I missed it.


thofma:
Is this function exported?

StevellM:
Yes, I just saw it somewhere in the `exports.jl` file

thofma:
I am afraid we have to add a deprecation. You know how to do that?

StevellM:
I guess I do. Though, I am convinced I have been the only using it locally so it should be fine ?

thofma:
Yes, but don't forget the million other Oscar users, which we don't know about.

fingolfin (approved):

PR #2766, opened 2023-09-04
Fix dim for affine schemes (#2369)

### LINKED ISSUE #2639: dim(X) gives wrong answer for X=Spec(A) and A localized at a prime ideal
julia> R,(x,y,z) = QQ["x","y","z"];

julia> U=complement_of_prime_ideal(ideal(R,[x,y]));

julia> S,_ = Localization(R,U);

julia> X = Spec(S)
Spectrum
  of localization
    of multivariate polynomial ring in 3 variables over QQ
    at complement of prime ideal(x, y)

julia> dim(X)
3   WRONG

`X =Spec(S)` and `S` is clearly a regular local ring of Krull-dimension 2 (as the maximal ideal is generated by 2 elements x and y). so `X` should have dimension 2.
(Caution: Do not be misled by trdeg_QQ ( Quot(S) ) = 3) 

Suggestion for ad hoc fix (to be improved later on): forbid dim for localizations at prime ideals for the time being.

paemurru:
Implemented dimension for localizations along

  * MPolyPowersOfElement,
  * MPolyComplementOfPrimeIdeal, and
  * MPolyComplementOfKPointIdeal

Not implemented for other localizations.
Enabled check for primeness in creating localized rings.
@HechtiDerLachs

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -391,12 +391,10 @@ dim(X::AbsSpec)
 end
 
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyPowersOfElement}})
-  println("QuoLoc PowersOfElement")
   return dim(closure(X))
 end
 
-@attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyComplementOfPrimeIdeal}})
-  println("QuoLoc ComplementOfPrimeIdeal")
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:Union{MPolyComplementOfPrimeIdeal, MPolyComplementOfKPointIdeal}}})
   # Spec (R / I)_P
   R = OO(X)
   P = prime_ideal(inverted_set(R))
@@ -409,13 +407,11 @@ end
 end
 
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyPowersOfElement}})
-  println("Loc PowersOfElement")
   # zariski open subset of A^n
   return dim(closure(X))
 end
 
-@attr function dim(X::AbsSpec{<:Ring, <:MPolyLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyComplementOfPrimeIdeal}})
-  println("Loc ComplementOfPrimeIdeal")
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:Union{MPolyComplementOfPrimeIdeal, MPolyComplementOfKPointIdeal}}})
   P = prime_ideal(inverted_set(OO(X)))
   return codim(P)
 end
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -387,13 +387,37 @@ julia> dim(Y) # one dimension comes from ZZ and two from x1 and x2
 dim(X::AbsSpec)
 
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing})
-  return dim(saturated_ideal(modulus(OO(X))))
+  error("Not implemented")
+end
+
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyPowersOfElement}})
+  println("QuoLoc PowersOfElement")
+  return dim(closure(X))
+end
+
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyComplementOfPrimeIdeal}})
+  println("QuoLoc ComplementOfPrimeIdeal")
+  # Spec (R / I)_P
+  R = OO(X)
+  P = prime_ideal(inverted_set(R))
+  I = saturated_ideal(modulus(R))
+  return dim(I) - dim(P)
 end
 
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyLocRing})
-  # the following line is supposed to refer the problem to the
-  # algebra side of the problem
-  return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))]))
+  error("Not implemented")
+end
+
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyPowersOfElement}})
+  println("Loc PowersOfElement")
+  # zariski open subset of A^n
+  return dim(closure(X))
+end
+
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyComplementOfPrimeIdeal}})
+  println("Loc ComplementOfPrimeIdeal")
+  P = prime_ideal(inverted_set(OO(X)))
+  return codim(P)
 end
 
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyRing})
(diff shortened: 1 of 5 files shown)


src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -387,13 +387,33 @@ julia> dim(Y) # one dimension comes from ZZ and two from x1 and x2
...
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing})
-  return dim(saturated_ideal(modulus(OO(X))))
+  error("Not implemented")
+end
+
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:MPolyPowersOfElement}})
+  return dim(closure(X))
+end
+
+@attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing{<:Any,<:Any,<:MPolyRing,<:MPolyRingElem, <:Union{MPolyComplementOfPrimeIdeal, MPolyComplementOfKPointIdeal}}})
```
afkafkafk13, line 397: Please do not do the two cases together. They differ significantly, which is otherwise hidden in the calls to other functions.
For MPolyComplementOfPrimeIdeal this version is fine.
In the case of MPolyComplementOfKPointIdeal, you do not need to adjust at all as dim(P) is guaranteed to be zero; you can just use "return dim(saturated_ideal(modulus(OO(X))))", which is the line you (correctly) removed above.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -387,13 +387,33 @@ julia> dim(Y) # one dimension comes from ZZ and two from x1 and x2
 dim(X::AbsSpec)
 
 @attr function dim(X::AbsSpec{<:Ring, <:MPolyQuoLocRing})
-  return dim(saturated_ideal(modulus(OO(X))))
+  error("Not implemented")
+end
```
afkafkafk13, line 391: One comment beforehand: Most of this should not happen on the geometric side, but on the algebra side. We need all these fixes on the algebra side as well and should only have them in one spot. The geometric side should directly refer dim (X) to dim(OO(X)) in the local and affine cases.

HechtiDerLachs (approved):
Thanks! I fixed some minor things, but overall, I think, this is good. 

Treating both, the case of prime ideals and ideals of rational points with one signature is good in terms of avoiding code duplication. But on the other hand deciding primeness and dimension, etc. is much easier for the latter ideals so that writing code of its own for that case also seems worthwhile. I chose the solution via setting the necessary attributes to avoid unnecessary computation for the moment.

HechtiDerLachs:
I guess the tests for primeness had been disabled due to @wdecker ? In my opinion it's OK to switch them on. I mean: We already saw in the tests for the `MPolyQuo`s where not testing this leads to (see the fix).

src/Rings/mpoly-localizations.jl
```diff
@@ -355,13 +373,13 @@ Complement
...
 @doc raw"""
-    complement_of_prime_ideal(P::MPolyIdeal; check::Bool=false)
+    complement_of_prime_ideal(P::MPolyIdeal; check::Bool=true)
 
 Given a prime ideal ``P`` of a polynomial ring ``R``, say,
 return the multiplicatively closed subset ``R\setminus P.``
 
 !!! note
-    If  `check` is set to `true`, the function checks whether ``P`` is indeed a prime ideal. 
+    Since  `check` is set to `true`, the function checks whether ``P`` is indeed a prime ideal.
```
afkafkafk13, line 392: This should still be an "if", as it can be overruled by input.

afkafkafk13 (commented):
Looks good to me. But to avoid duplicate code, the case by case distinction should happen on the algebra side, not on the geometry side. (see comment in the code)

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
HechtiDerLachs: Ok, agreed. It seems that a lot of this is still missing on the algebra side, so this might be a good occasion to add it.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
HechtiDerLachs: Yes. Now that I try to fix the case of `MPolyComplementOfKPointIdeal` over the integers, I also start to have my doubts whether putting the cases together is a good idea.

src/Rings/mpoly-localizations.jl
```diff
@@ -332,6 +333,33 @@ mutable struct MPolyComplementOfKPointIdeal{
   end
 end
 
+# A function with this name exists with a method for 
+# MPolyComplementOfPrimeIdeal. In order to treat them in 
+# parallel, we introduce the analogous method here. 
+function prime_ideal(S::MPolyComplementOfKPointIdeal) 
+  kk = coefficient_ring(ambient_ring(S))
+  # TODO: Test whether kk is an integral domain
```
HechtiDerLachs, line 341: I didn't find a way to test this. If anybody knows, let me know, too!

src/Rings/mpoly-localizations.jl
```diff
@@ -337,19 +337,29 @@ end
...
   if !isdefined(S, :m)
     R = ambient_ring(S)
     x = gens(R)
     n = ngens(R)
     m = ideal(R, [x[i]-a for (i, a) in enumerate(point_coordinates(S))])
     set_attribute!(m, :is_prime=>true)
-    set_attribute!(m, :is_maximal=>true)
-    set_attribute!(m, :dim=>0)
+    set_attribute!(m, :is_maximal=>iszero(dim(kk)))
```
afkafkafk13: I doubt that this will work for kk = ZZ/<4>.

src/Rings/mpoly-localizations.jl
afkafkafk13: This is a primatlity test for the zero ideal of kk.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: Yes. But nothing really does, so I think, it cannot be a real issue here. It wouldn't be mathematically incorrect here, would it?

src/Rings/mpoly-localizations.jl
HechtiDerLachs: Yes, sure. I tried it. It fails over `ZZ`: The required methods are not there and we do not have a coherent interface. We need one method like `is_integral_domain` which we expect to be implemented for the ring.

src/Rings/mpoly-localizations.jl
afkafkafk13: This attribute is_maximal is mathematically incorrect here, as a maximal ideal in this example would be m+<2>, which strictly contains m.

src/Rings/mpoly-localizations.jl
afkafkafk13: I agree and I would explicitly test for the following cases:
- field
- integers
and not set is_maximal for the other cases.

Edit: This is of course for the lowest level. As long as kk=R[x] for some R, we can still go one level further down and decide there.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
simonbrandhorst: How serious are we about doing things on the algebra side? 
If we are serious, then there would be only a single method:
`dim(X::AbsSpec) = dim(OO(X))`
and that is it.  But yea, in the long run this is maybe the best way to go.

afkafkafk13: This is definitely the way to go. Otherwise there will be tons of duplicate code.

(My issue was meant as a marker not to forget this (large and necessary) cleanup. I wrote it on the geometric side because of my own lazyness, as I had stumbled into it on that side. But I always expected the solution to be implemented on the algebraic side and only a simple call on the geometric one.

fingolfin:
Hi @paemurru, thank you for your contribution. I assume the algebraic geometry people here know who you are, but right now your GitHub user profile is empty and so to an outsider you are a bit mysterious. In general we appreciate if we know for anyone who contributes at least the full name, ideally also affiliation. Perhaps you would be amenable to adding this to your GitHub user profile?

fingolfin:
(I guess technically this would also be nice for @simonbrandhorst but there the GitHub username at least is the full name, and we know him, so it's not *really* a double standard...)

paemurru:
> your GitHub user profile is empty

Thanks, I added my full name and affiliation to my GitHub user profile.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
simonbrandhorst: You write that `dim(P)` is guaranteed to be zero. Is it also zero over the integers?

src/Rings/mpoly-localizations.jl
simonbrandhorst: By default it is true, so I think that `since` fits.

simonbrandhorst (approved):
Looks good to me for now. 
Moving everything to the commutative algebra side could be done later?

lgoettgens:
> Looks good to me for now. Moving everything to the commutative algebra side could be done later?

I have added #2771 to keep that in mind.

PR #2768, opened 2023-09-04
More polish for elliptic surfaces

simonbrandhorst:
This replaces `relatively_minimal_model(X::EllipticSurface)` by `weierstrass_contraction(X::EllipticSurface)` 

@HechtiDerLachs  `two_neighbor_step(X, ff) ` is still buggy.  See the added line in the test file which fails. 
Kumar in his paper assumes that `a(t)` and `b(t)` are elemens of the function field k(t). You seem to assume that they are polynomials, which is not true in general. 
Probably the other cases of `_elliptic_parameter_conversion` are affected too. 
Please fix.

### DIFF
--- a/experimental/Schemes/elliptic_surface.jl
+++ b/experimental/Schemes/elliptic_surface.jl
@@ -458,16 +458,17 @@ This triggers the computation of the `underlying_scheme` of ``X``
 as a blowup from its Weierstrass model. It may take a few minutes.
 """
 function weierstrass_contraction(X::EllipticSurface)
-  if isdefined(E, :blowup)
-    return E.Y, E.blowup
+  Y = X
+  if isdefined(Y, :blowup)
+    return Y.blowup
   end
-  S, inc_S = weierstrass_model(E)
-  Crefined = _separate_singularities!(E)
+  S, inc_S = weierstrass_model(Y)
+  Crefined = _separate_singularities!(Y)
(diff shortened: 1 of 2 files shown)


experimental/Schemes/elliptic_surface.jl
```diff
@@ -528,26 +548,26 @@ function relatively_minimal_model(E::EllipticSurface)
...
+  inc_Y0_wrap = CoveredClosedEmbedding(Y, codomain(inc_Y0), covering_morphism(inc_Y0), check=false)
+  Y.inc_Y = inc_Y0_wrap
 
-  set_attribute!(E, :is_irreducible=> true)
-  set_attribute!(E, :is_reduced=>true)
-  set_attribute!(E, :is_integral=>true)
-  return E, piY
+  set_attribute!(Y, :is_irreducible=> true)
+  set_attribute!(Y, :is_reduced=>true)
+  set_attribute!(Y, :is_integral=>true)
```
HechtiDerLachs, line 589: This must be `E`! Otherwise, the attribute is set only for the `underlying_scheme` and that will lead to massive tests and failures for further methods.

simonbrandhorst: I renamed E to Y. The final blown up surface is Y0

HechtiDerLachs (approved):

PR #2769, opened 2023-09-05
Fix printing guidelines

fingolfin:
Properly explain how to reach each printing mode, and also explain what 'io' is

### DIFF
--- a/docs/src/DeveloperDocumentation/printing_details.md
+++ b/docs/src/DeveloperDocumentation/printing_details.md
@@ -7,15 +7,17 @@ found in the [Developer Style Guide](@ref).
 
 ## Implementing show functions
 
-Here is the translation between `:detail`, `one line` and `:supercompact`.
+Here is the translation between `:detail`, `one line` and `:supercompact`,
+where `io` is an `IO` object (such as `stdout` or an `IOBuffer`):
 
 ```
-print(io, "text/plain", x)                 # detailed printing
-print(io, x)                               # one line printing
-print(IOContext(:supercompact => true), x) # supercompact printing
+show(io, MIME"text/plain"(), x)                # detailed printing
+print(io, x)                                   # one line printing
+print(IOContext(io, :supercompact => true), x) # supercompact printing
 ```
 
-For reference, string interpolation `"$(x)"` will also use `print(io, x)`.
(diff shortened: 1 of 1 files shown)


docs/src/DeveloperDocumentation/printing_details.md
```diff
@@ -7,15 +7,17 @@ found in the [Developer Style Guide](@ref).
 
 ## Implementing show functions
 
-Here is the translation between `:detail`, `one line` and `:supercompact`.
+Here is the translation between `:detail`, `one line` and `:supercompact`,
+where `io` is an `IO` object (such as `stdout` or an `IOContext`):
```
fingolfin:
```suggestion
where `io` is an `IO` object (such as `stdout` or an `IOBuffer`):
```

docs/src/DeveloperDocumentation/printing_details.md
```diff
@@ -173,7 +175,7 @@ end
 
 function Base.show(io::IO, b::B)
   io = AbstractAlgebra.pretty(io)
-  print(io, LowercaseOff(), "Hilbert thing")
+  print(io, AbstractAlgebra.LowercaseOff(), "Hilbert thing")
```
simonbrandhorst, line 178: I believe this is exported? Then `AbstractAlgebra.` is not necessary.

ThomasBreuer: As far as I understand, `LowercaseOff`, `Indent`, `Dedent`, `pretty` are imported from AbstractAlgebra into Oscar, thus one need not write `AbstractAlgebra.` in code inside Oscar; but they are not exported from Oscar.

thofma:
Looks good, once the the namespace has been figured out.

docs/src/DeveloperDocumentation/printing_details.md
fingolfin: The example code uses an explicit `AbstractAlgebra.` everywhere else, and it should be consistent. With it there, a user can copy&paste the example code and start working with it, whether or not they have done `using AbstractAlgebra` (which is dangerous inside a package) or `import AbstractAlgebra`.

Anyway, my primary concern was to make it consistent. If someone prefers to drop the `AbstractAlgebra.` everywhere, that should be a seperate PR, IMHO

fieker (approved):

PR #2770, opened 2023-09-05
CI: use buildpkg@v1

benlorenz:
(instead of the outdated `latest` tag)


fingolfin:
I am fine with this, but should we perhaps also open an issue there and suggest to them to update the `latest` tag?

fingolfin (approved):

benlorenz:
> I am fine with this, but should we perhaps also open an issue there and suggest to them to update the `latest` tag?

done: https://github.com/julia-actions/julia-buildpkg/issues/35

Issue #2772, opened 2023-09-05
Adjust printing of GAP groups & group homomorphism to follow OSCAR printing guidelines


fingolfin:
... as described in https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/ (with fixes in PR #2769).

Right now I see this:
```
julia> H, phi = derived_subgroup(symmetric_group(5))
(Alt( [ 1 .. 5 ] ), Group homomorphism from
Alt( [ 1 .. 5 ] )
to
Sym( [ 1 .. 5 ] ))

julia> phi
Group homomorphism from
Alt( [ 1 .. 5 ] )
to
Sym( [ 1 .. 5 ] )
```
Note that `phi` prints the same in all "modes" (detailed, oneline, supercompact)

Based on [this discussion](https://hackmd.io/cwOoBrTaReihNtmFjSdtTg) I think it rather should look something like this:
```
julia> phi  # detailed
Group homomorphism
  from Alt( [ 1 .. 5 ] )
  to Sym( [ 1 .. 5 ] )

julia> print(phi)  # oneline
Hom: permutation group -> permutation group

julia> print(IOContext(stdout, :supercompact => true), phi)
Group homomorphism
```

Of course the above example also assumes that we implement the three modes for permutation groups, and then logically also for other kinds of groups. So that should be tackled at the same time.

lgoettgens:
I would really like the oneline printing to state the category we are in in the beginning. So e.g. `Group hom: ...`, as this is already the beginning of the other two printing styles.

ThomasBreuer:
I support the comment
> I would really like the oneline printing to state the category we are in in the beginning. So e.g. `Group hom: ...`, as this is already the beginning of the other two printing styles.

As long as all homomorphisms respect the structure of domain and codomain, one could argue that the shown description of domain and codomain says what kind of homomorphism the map in question is.
I think we will have to support also homomorphisms for which domain and codomain have different structures, and then the meaning of "homomorphism" is not obvious.

(Same for #2773.)

fingolfin:
> I think we will have to support also homomorphisms for which domain and codomain have different structures, and then the meaning of "homomorphism" is not obvious.

One could argue that a homomorphism by definition is a map between two objects in a single category. Of course I can look at maps from e.g. a group into a matrix algebra, but that's then not a homomorphism.


Anyway: the HackMD says "Hom:" and that was already a compromise after a lot of discussion.

[...]

PR #2774, opened 2023-09-05
Revise printing of GAP groups and homomorphisms

fingolfin:
First stab at resolving #2772.

TODO:
- add resp. adjust tests
- adjust printing of groups
- ...?


src/Groups/homomorphisms.jl
```diff
@@ -1,10 +1,27 @@
+function Base.show(io::IO, ::MIME"text/plain", x::GAPGroupHomomorphism)
+   println(io, "Group homomorphism")
+   io = AbstractAlgebra.pretty(io)
+   print(io, AbstractAlgebra.Indent())
```
lgoettgens:
```suggestion
   io = pretty(io)
   print(io, Indent())
```
all of this gets imported in `src/imports.jl`

src/Groups/homomorphisms.jl
```diff
@@ -1,10 +1,27 @@
...
+   println(IOContext(io, :supercompact => true), "from ", domain(x))
+   print(IOContext(io, :supercompact => true), "to ", domain(x))
+   print(io, AbstractAlgebra.Dedent())
```
lgoettgens:
```suggestion
   print(io, Dedent())
```

src/Groups/GAPGroups.jl
```diff
@@ -640,14 +689,14 @@ julia> G = symmetric_group(3);
...
 julia> subgroup_reps(G)
 4-element Vector{PermGroup}:
- Group(())
- Group([ (2,3) ])
- Group([ (1,2,3) ])
- Group([ (1,2,3), (2,3) ])
+ Permutation group of degree 3 and order 1
+ Permutation group of degree 3 and order 2
+ Permutation group of degree 3 and order 3
+ Permutation group of degree 3 and order 6
```
fingolfin, line 713: I've also experimented with changing how we print groups, following some of the suggestions in the HackMD. At first I did not show the order (because we don't want to trigger its computation just by printing; and it seems potentially confusing to only sometimes show it).

But even that is a bit bland to my liking, we really loose information compared to before.

Perhaps at least the "detailed mode" for permutation groups should still show generators?

src/Groups/GAPGroups.jl
```diff
@@ -1110,7 +1159,8 @@ julia> (length(h), order(h[1]))
 (1, 6)
 
 julia> g = GL(3, 2)
-GL(3,2)
+Matrix group of degree 3
+  over finite field of degree 1 over GF(2)
```
fingolfin: I guess the custom printing here should be preserved? I.e. the "name" `GL(3,2)`. Might be as simple as calling `set_name!(G, "GL(3,2)")` in the constructor or so.

src/Groups/GAPGroups.jl
```diff
@@ -612,7 +661,7 @@ Return the vector of all conjugacy classes of subgroups of G.
 # Examples
 ```jldoctest
 julia> G = symmetric_group(3)
-Sym( [ 1 .. 3 ] )
+Permutation group of degree 3 and order 6
```
fingolfin, line 682: It would be kinda nice to preserve GAP assigned names to some extent; in this case printing this group as `Sym(3)` or `Symmetric group of degree 3` or anything like would be nice

fingolfin: Of course we can also take inspiration from Magma and Sage and GAP...

Magma:
```
> Sym(6);
Symmetric group acting on a set of cardinality 6
Order = 720 = 2^4 * 3^2 * 5
```
so a special name, and the order is also printed factorized, which I guess can be quite handy.
Or here:
```
> H, map := PermutationGroup< 9 | (1,2,4)(5,6,8)(3,9,7), (4,5,6)(7,9,8) >;
> H;
Permutation group H acting on a set of cardinality 9
    (1, 2, 4)(3, 9, 7)(5, 6, 8)
    (4, 5, 6)(7, 9, 8)
```
So the generators are still printed in this detailed form... But they are hidden in "compact" form, e.g. when printing the map:
```
> map;
Mapping from: GrpPerm: H to GrpPerm: $, Degree 9, Order 2^7 * 3^4 * 5 * 7
```
Note the use of `H` and `$` which are the "names" of the group in the session. We can see this when we assign a name to the codomain
```
> Codomain(map);
Symmetric group acting on a set of cardinality 9
Order = 362880 = 2^7 * 3^4 * 5 * 7
> U := Codomain(map);
> map;
Mapping from: GrpPerm: H to GrpPerm: U
```
(In our system, assigning a name would drop the `GrpPerm` / `Permutation group` bit)


(Should look at how they handle pc and fp groups later, too; and perhaps also look at Sage... well, and GAP, but I know that / that's basically what we are using right now)

src/Groups/GAPGroups.jl
lgoettgens: There is already a competing implementation for Matrix Groups in https://github.com/oscar-system/Oscar.jl/blob/37eec92a19fb9304e2fbf8c688ab5c261ab2f647/src/Groups/matrices/MatGrp.jl#L121
This should be removed, but you could maybe copy-paste some bits of it to your new implementation.

src/Groups/GAPGroups.jl
ThomasBreuer: I would distinguish assigning a name to the group from assigning the group object to a variable. Why shall the name of the first variable to which the group gets assigned determine how the group is printed?
(Yes, this happens already for certain Oscar objects, but I find this unnatural.)

Concerning the idea to assign a name to a group and then use this name for printing: This is in principle reasonable. However, I find it irritating that calling `describe` (or rather GAP's `StructureDescription`) *automatically* assigns a name, as in the following GAP session.
```
gap> l:= [ Group( (1,2) ), SmallGroup( 2, 1 ), Group( [[Z(3)]] ) ];
[ Group([ (1,2) ]), <pc group of size 2 with 1 generator>, 
  Group([ [ [ Z(3) ] ] ]) ]
gap> List( l, StructureDescription );
[ "C2", "C2", "C2" ]
gap> l;
[ C2, C2, C2 ]
```
(The same happens in `all_primitive_groups`.)
Thus the user does not have control over the assigned names.

src/Groups/GAPGroups.jl
```diff
@@ -261,8 +261,57 @@ Return the identity of the parent group of `x`.
 Base.one(x::GAPGroupElem) = one(parent(x))
 
 Base.show(io::IO, x::GAPGroupElem) = print(io, String(GAPWrap.StringViewObj(x.X)))
+
+
+# Printing GAP groups
 Base.show(io::IO, x::GAPGroup) = print(io, String(GAPWrap.StringViewObj(x.X)))
```
ThomasBreuer: This generic method, i. e., delegate to `GAPWrap.StringViewObj` is currently used by f.p. groups. It could be replaced by
```
function Base.show(io::IO, x::PermGroup)
  @show_name(io, x)
  @show_special(io, x)
  print(io, "Group")
  if !get(io, :supercompact, false)
    if has_order(x)
      print(io, " of order ", order(x))
    end
  end
end
```

src/Groups/GAPGroups.jl
ThomasBreuer: Looking at other structures in Oscar, showing the generators seems to be not usual. For example, polynomial rings print only the number of generators.

src/Groups/GAPGroups.jl
ThomasBreuer: When should the name be set, only in the construction or also later on when the structure in question has been found out? (The same would hold for symmetric groups.)

src/Groups/group_characters.jl
```diff
@@ -283,31 +283,31 @@ then `nothing` is returned.
...
+
 χ₁  1 -1  1
 χ₂  2  . -1
 χ₃  1  1  1
 
 julia> Oscar.with_unicode() do
          show(stdout, MIME("text/plain"), character_table(symmetric_group(3), 2))
        end;
-Sym( [ 1 .. 3 ] ) mod 2
+Permutation group of degree 3 and order 6 mod 2
```
ThomasBreuer: Perhaps putting the `mod 2` in brackets would suffice to "disambiguate" the description. If not then we need a more explicit description such as `2-modular Brauer table of permutation group of degree 3`.

ThomasBreuer (approved):
This looks good.
I am not sure that we want to print groups via their stored names.

src/Groups/group_characters.jl
fingolfin: I think I would "2-modular Brauer table of permutation group of degree 3"

src/Groups/GAPGroups.jl
fingolfin: OK

src/Groups/GAPGroups.jl
```diff
@@ -435,6 +435,35 @@ julia> length(small_generating_set(abelian_group(PermGroup, [2,3,4])))
    return res
 end
 
+"""
+    minimal_generating_set(G::GAPGroup)
```
lgoettgens: I am a bit confused by the name and the docstring of this function. If it does not guarantee to return a minimal generating set, why not call it `small_generating_set`? That function already exists for different types.

fingolfin: That was a copy&paste mistake. Fixed in #2816; I hope that one will be merged quickly then I'll rebase this PR.

For this PR I still need to adjust some tests, working on that.

fingolfin: (And thanking you for catching this)

src/Groups/group_characters.jl
```diff
@@ -892,9 +893,11 @@ function Base.show(io::IO, ::MIME"text/plain", tbl::GAPGroupCharacterTable)
     emptycol = ["" for i in 1:n]
 
     if isdefined(tbl, :group)
-      headerstring = string(group(tbl))
+      headerstring = lowercasefirst(string(group(tbl)))
       if characteristic(tbl) != 0
-        headerstring = "$headerstring mod $(characteristic(tbl))"
+        headerstring = "$(characteristic(tbl))-module Brauer table of $(headerstring)"
```
ThomasBreuer:
```suggestion
        headerstring = "$(characteristic(tbl))-modular Brauer table of $(headerstring)"
```

src/Groups/group_characters.jl
```diff
@@ -283,31 +283,31 @@ then `nothing` is returned.
...
+
 χ₁  1 -1  1
 χ₂  2  . -1
 χ₃  1  1  1
 
 julia> Oscar.with_unicode() do
          show(stdout, MIME("text/plain"), character_table(symmetric_group(3), 2))
        end;
-Sym( [ 1 .. 3 ] ) mod 2
+2-module Brauer table of permutation group of degree 3 and order 6
```
ThomasBreuer:
```suggestion
2-modular Brauer table of permutation group of degree 3 and order 6
```

ThomasBreuer (changes requested):
typo

docs/src/InvariantTheory/finite_groups.md
```diff
@@ -89,16 +89,18 @@ julia> M2 = matrix(K, [1 0 0; 0 a 0; 0 0 -a-1])
...
+Matrix group of degree 3
+  over cyclotomic field of order 3
 
 julia> IR = invariant_ring(G)
 Invariant ring of
-Matrix group of degree 3 over K
+  Matrix group of degree 3 over cyclotomic field of order 3
 with generators
-AbstractAlgebra.Generic.MatSpaceElem{nf_elem}[[0 0 1; 1 0 0; 0 1 0], [1 0 0; 0 a 0; 0 0 -a-1]]
+  AbstractAlgebra.Generic.MatSpaceElem{nf_elem}[[0 0 1; 1 0 0; 0 1 0], [1 0 0; 0 a 0; 0 0 -a-1]]
```
fingolfin, line 99: maybe this could be printed properly as matrix...

lgoettgens: I do not think that this stays readable if we print it as a full-blown matrix and there are multiple generators.
Removing the type would be great

joschmitt: I don't see a problem with printing proper matrices (neatly one over the other like Magma does). Sure, this will quickly fill the screen (just type `ShephardTodd(37);` in Magma), but if the user (that would be me) asks for how the group `G` in the REPL was defined again, then they should get a readable answer. I personally don't consider the "folded up" way to print a matrix in one line readable. We could of course put `⋮` in a longer list of generators, like it was proposed in the HackMD (is implemented?) for ideals in polynomial rings.

joschmitt: Okay, I just noticed that this is about the printing of `InvRing`. `MatrixGroup` still does not print any generators.

fingolfin: But I was thinking about MatrixGroup

ThomasBreuer: Ah, now I understand this:
The `show` method for the invariant ring `IR` prints `group(IR)` and then `action(IR)`, and the latter refers not to the generators of the matrix group but to the underlying matrices.
The confusion about the type would disappear if the type information would not be shown, as suggested by @lgoettgens .
If `show` for matrix groups would start showing generators then probably `show` for invariant rings must be adjusted.

src/Groups/GAPGroups.jl
simonbrandhorst: Yea, and it started to really annoy me. So I would argue that detailed should certainly print the generators.

fingolfin:
This can certainly still be improved but I think it already is a good bit closer to what we want in the end, so perhaps we can merge it and then improve it in future PRs.

src/Groups/GAPGroups.jl
fingolfin: @simonbrandhorst I tend to agree, feel free to open an issue suggesting this (or submit a PR)

fingolfin:
@joschmitt I plan to revise the printing of generators in detailed mode, including for matrix groups, but this PR already improves things, so I'd like to get it merged ASAP, and then someone can make a follow-up PR with better detailed printing (I might do it if as soon as I find time or need to procrastinnate, but if you are interested... 😁 )

@ThomasBreuer please approve this PR if you think it's OK as an intermediate step (again: I fully expect us to iterate on this, but I'd like to get the ball rolling...)

ThomasBreuer (approved):

joschmitt:
> @joschmitt I plan to revise the printing of generators in detailed mode, including for matrix groups, but this PR already improves things, so I'd like to get it merged ASAP, and then someone can make a follow-up PR with better detailed printing (I might do it if as soon as I find time or need to procrastinnate, but if you are interested... 😁 )

:+1: Let's see what my future boss wants me to do during my OSCAR time...

PR #2775, opened 2023-09-05
Revert "CI: use buildpkg@v1"

lgoettgens:
Reverts oscar-system/Oscar.jl#2770.
The issue to be circumvented by this is now solved.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
       - name: "Build package"
-        uses: julia-actions/julia-buildpkg@v1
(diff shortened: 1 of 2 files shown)


benlorenz:
I think in general using a specific version is better. `latest` might switch to v2 at some point which could introduce breaking changes. (Not that I really expect anything like this for this action ...)

_Edit:_ The [readme](https://github.com/julia-actions/julia-buildpkg#usage) also suggests using `v1`.

fingolfin:
I tend to a agree that we are better of using v1

lgoettgens:
Ok, fine with me. But then we should check/change this for our other packages as well. I'll keep this PR open as a reminder.

PR #2779, opened 2023-09-07
Toric morphisms to morphisms of covered schemes

HechtiDerLachs:
Continuation of #2741 . To be rebased eventually.

### DIFF
--- a/docs/src/AlgebraicGeometry/ToricVarieties/ToricMorphisms.md
+++ b/docs/src/AlgebraicGeometry/ToricVarieties/ToricMorphisms.md
@@ -49,6 +49,7 @@ morphism_on_torusinvariant_weil_divisor_group(tm::ToricMorphism)
 morphism_on_torusinvariant_cartier_divisor_group(tm::ToricMorphism)
 morphism_on_class_group(tm::ToricMorphism)
 morphism_on_picard_group(tm::ToricMorphism)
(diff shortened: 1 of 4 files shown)


HereAround:
Rebased just now, since https://github.com/oscar-system/Oscar.jl/pull/2741 has been merged.

HechtiDerLachs:
Looks like we should override the print method for toric morphisms once again. For now it seems to default to the one for morphisms of covered schemes which entails actually computing the whole thing. We definitely want to avoid that.

HereAround:
> Continuation of #2741 . To be rebased eventually.

Just rebased to the latest master.

HereAround:
Based on conversations with @HechtiDerLachs, this is ready for review. As the title says, this PR allows to convert a toric morphism to a morphism of (covered) schemes. In this regard, this PR is again at the interface between toric geometry and algebraic geometry. Hence, it seems reasonable to bring expertise from both areas to check on this PR. Let me thus call in @lkastner @simonbrandhorst - what do you think of this PR?

(The upshot of all this is to enable blowup functionality, including proper/strict transforms in the toric setting and then to make use of these tools in the context of the FTheoryTools. I am sure @apturner will be very happy with the changes in this PR and the next couple that @HechtiDerLachs and I are working on/discussing.)

simonbrandhorst (approved):
I was looking forward to this! Well done.

PR #2780, opened 2023-09-07
Interface for blowups and blowdown morphisms

HechtiDerLachs:
This is a draft to fix an interface for blowups in general. Please let me know your thoughts, ideas, and objections here.

CC: @afkafkafk13 , @HereAround @apturner .

As this is derived from #2779 , you only need to consider the last commit.


HechtiDerLachs:
@HereAround : I would like to introduce a new type for toric blowups via fan subdivision. Up to now, the `blow_up` method returns a usual `toric_morphism`. But this morphism does not know about its `exceptional_divisor`, its `center`, etc., so it is not yet suited to implement the interface formulated here. A new type could store this extra information and provide the required entry points for the entailing functionality. I will try to work out a prototype and push it here for discussion.

afkafkafk13:
This looks good to me. Do you expect weighted blow ups to comply with these interface settings as well sometime in the farther future, when they are implemented? 
Edit: no it is not sufficiently general for this, weighted blow-ups live in a stacky world!

HechtiDerLachs:
My understanding of weighted blowups does not go very deep as for now. But yes: I hope that they will also fit in here. The natural next step, however, would be sequences of simple blowups as they occur in a resolution process. That would be my first question here: Do you think this can accommodate a resolution of singularities?

afkafkafk13:
Basically yes, the sequence of blow-ups in a Hironaka-style resolution is just a means to construct the proper birational covered-scheme morphism which is the resolution of singularities. Only the command `center` is not straight forward and I would not make it part of the mandatory set of commands to implement.
(In this case, we also need a means to access the individual steps)

Let us now think about a Lipman-style resolution of surfaces: Here we have two different kinds of birational morphisms, namely blowups and normalization. Then we would need a more general type of proper birational morphism of schemes. In this case, it does not make sense to talk about exceptional divisors or exceptional loci either. so I would say that we ignore the case for now.

If we think about a weighted/log-weighted resolution, the design fits again, but the morphisms will not be morphisms of schemes, but of stacks.

experimental/Schemes/ToricBlowups.jl
```diff
@@ -0,0 +1,67 @@
+@attributes mutable struct ToricBlowdownMorphism{
+                                                 DomainType <: NormalToricVariety, 
+                                                 CodomainType <: NormalToricVariety
+    } <: AbsSimpleBlowdownMorphism{DomainType,
+                                   CodomainType,
+                                   Nothing,
+                                   ToricBlowdownMorphism
+                                  }
```
HechtiDerLachs: @HereAround : This is the new type I was talking about. Let me know what you think.

experimental/Schemes/ToricBlowups.jl
```diff
@@ -0,0 +1,67 @@
...
+end
+
+# The user facing constructor and candidate to replace the current method of 
+# `blow_up`, eventually. Called with the weird name for the time being.
+function wrapped_blow_up(v::NormalToricVarietyType, new_ray::AbstractVector{<:IntegerUnion}; coordinate_name::String = "e", set_attributes::Bool = true)
+  return ToricBlowdownMorphism(v, new_ray, coordinate_name, set_attributes)
+end
+
+# Forwarding all the essential functionality for morphisms of schemes 
+underlying_morphism(bl::ToricBlowdownMorphism) = bl.toric_morphism
```
HechtiDerLachs: This could, for instance, be used to get back the familiar toric morphism. All the special dispatch we already implemented for that one is automatically forwarded.

experimental/Schemes/ToricBlowups.jl
```diff
@@ -0,0 +1,67 @@
...
+
+# Implement the interface for AbsSimpleBlowdownMorphism
+function center(bl::ToricBlowdownMorphism)
+  if !isdefined(bl, :center)
+    v = bl.new_ray
+    Y = codomain(bl)
+    S = cox_ring(Y)
+    x = gens(S)
+    # TODO: Construct the ideal in the cox_ring corresponding to the new ray
+    #I = ideal(S, [x[i]^v[i] for i in 1:n if !iszero(v[i])])
```
HechtiDerLachs: I do not know how to do this right now, but only because for today I'm too tired to start reading the books. This shouldn't be too difficult, in the end.

experimental/Schemes/ToricBlowups.jl
```diff
@@ -0,0 +1,67 @@
...
+end
+
+function exceptional_divisor(bl::ToricBlowdownMorphism)
+  if !isdefined(bl, :exceptional_divisor)
+    v = bl.new_ray
+    X = domain(bl)
+    S = cox_ring(X)
+    x = gens(S)
+    # TODO: Construct the sheaf of principal ideals on the `default_covering` of X 
+    # which corresponds to the new ray
```
HechtiDerLachs: Same here.

test/AlgebraicGeometry/ToricVarieties/toric_schemes.jl
```diff
@@ -21,10 +21,82 @@ using Test
...
+    bl = Oscar.wrapped_blow_up(IP2, [1, 1])
+    f = underlying_morphism(bl)
+    S = cox_ring(IP2)
+    x, y, z = gens(S)
+    I = ideal(S, [x-y])*ideal(S, [z])
+    II = IdealSheaf(IP2, I)
+    pb_II = pullback(II)
+    # The following is not yet functional, because `exceptional_divisor` is not yet implemented.
+    # However, this should run using the generic code!
+    #J = strict_transform(bl, II)
```
HechtiDerLachs: This is what I'm finally aiming at: The signature for `strict_transform` has been widened so that the code is supposed to handle input of type `<:AbsSimpleBlowdownMorphism`. The `ToricBlowdownMorphism` now falls into this category. Once we have a fully working implementation of `exceptional_divisor` for this concrete type, we should be able to uncomment this line and have it run out of the box!

HereAround:
To keep working on this and after talking to @HechtiDerLachs, I have just squashed the previous commits and rebased this on the current master.

HereAround:
Another attempt on the rebase and squash... hopefully it works this time.

HereAround:
Pushed a bit of clean-up, but more to come...

HechtiDerLachs:
The failing tests in `ubuntu-latest` do not seem to be our fault. @HereAround : Can we do something about the failing doctests?

HereAround:
I will look into this once I am back next week. But this definitely needs more cleanup as we discussed (documentation, missing functions to be added etc.)

HereAround:
Unless somebody beats me to it, that is.

experimental/Schemes/ToricBlowups.jl
HereAround: Is good (as discussed in person).

experimental/Schemes/ToricBlowups.jl
HereAround: I like it!

experimental/Schemes/ToricBlowups.jl
HereAround: Resolved by in-person discussion.

experimental/Schemes/ToricBlowups.jl
HereAround: Also resolved by in-person discussion.

test/AlgebraicGeometry/ToricVarieties/toric_schemes.jl
HereAround: I like it!

HereAround:
As per usual, the tests may show something unexpected. That put aside, I have completed the improvements that I wanted to complete. Those changes include:
- Code improvements (no need to remember `new_ray` in the type `ToricBlowDownMorphism`, only a single internal constructor for `ToricBlowdownMorphism` is needed, ...)
- Code documentation (2 new pages for the toric varieties, one concerning all of the blowup technology and one other about the ideal sheaves - both are marked as experimental)
- Organized the experimental code to align it with the code structure of the toric varieties (so eventually, we merely have to copy it to `/src`)
- Optimized the tests (some tests were executed multiple times/created the same object repeatedly, grouped them according to subject).

Our current implementation of ideal sheaves only works for smooth toric varieties. (Reason: No time yet to look beyond it.) The previous implementation attempted to set the center of the blowup, and this failed for non-smooth examples appearing in the FTheoryTools. Since I cannot implement ideal sheaves for non-smooth toric varieties right now, I have altered this such that the blowdown morphism can be computed, but its attributes `center` and `exceptional_divisor` cannot be computed in such non-smooth settings. We (or I) will have to come back to this eventually.

Other than this... should be good to go once the tests go green.

Thank you for setting up the prototype @HechtiDerLachs !

HereAround:
Based on conversation with @HechtiDerLachs, this seems ready for review.

cc @afkafkafk13 @lkastner @simonbrandhorst

HereAround (approved):

HereAround:
Since I hear no objections (yet) and this PR concerns experimental code, I will merge now.

Thank you @HechtiDerLachs !

PR #2782, opened 2023-09-07
Put weierstrass_chart back on the weierstrass_model

HechtiDerLachs:
As discussed with @simonbrandhorst. 

Could you have a look? I had to mess a bit with the internals of the linear systems. So far this didn't break any test. But I'm not 100% sure I did all the things correctly.


simonbrandhorst (approved):
looks reasonable to me

PR #2787, opened 2023-09-12
Add filter to `convention(::TropicalSemiring)` doctest

lgoettgens:
A changing number of methods for `max` and `min` should not break the doctest.


lgoettgens:
Having this merged would reduce the number of errors in https://github.com/oscar-system/Oscar.jl/pull/2796.

thofma (approved):

PR #2788, opened 2023-09-12
Repair quotient rings

HechtiDerLachs:
CC: @wdecker , @fingolfin @fieker

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -101,6 +101,53 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
   return MPolyQuoRingElem(Base.deepcopy_internal(a.f, dict), a.P, a.simplified)
 end
 
+########################################################################
+# Representatives of elements in quotient rings and normal forms 
+#
+# Elements [a] ∈ P/I in quotients of polynomial rings 
+# P = 𝕜[x₁,…,xₙ] by ideals I = ⟨f₁,…,fᵣ⟩ admit unique representatives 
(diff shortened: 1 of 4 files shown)


thofma:
Was it decided at one of the Kaiserslautern meetings, that we'd rather have a hash functions which errors than a slow default one?

HechtiDerLachs:
Yes. This was @fieker 's proposal.

lgoettgens:
> Was it decided at one of the Kaiserslautern meetings, that we'd rather have a hash functions which errors than a slow default one?

This is similar to https://github.com/oscar-system/GAP.jl/pull/891.

HechtiDerLachs:
Looks like the code coverage report is nothing that can be taken serious these days. All coverage changes occur in places that I did not touch; not even indirectly! I suppose this is due to the various failing tests? 

Either way: I would like to point to #2789 if anyone feels like doing a proper review.

src/Rings/MPolyQuo.jl
```diff
@@ -807,9 +842,20 @@ true
...
-  return f.f == g.f
+  Q = parent(f)::MPolyQuoRing
+  P = base_ring(Q)::MPolyRing
+  kk = coefficient_ring(P)::Ring
+
+  # The following bracket is for the default assumption
+  if is_coefficient_type_with_groebner_basis_algorithm_in_singular(kk)
+    f.f == g.f && return true
+    hash(f) == hash(g) && return true # calls simplify already
+    return f.f == g.f
```
simonbrandhorst, line 853: This looks wrong to me. Please double check

HechtiDerLachs: Indeed. What was I thinking...? Corrected in #2789 .

PR #2789, opened 2023-09-12
Repair quotient rings with traits

HechtiDerLachs:
I felt this was a good occasion to finally learn about [traits](https://ahsmart.com/pub/holy-traits-design-patterns-and-best-practice-book/). So here it is: My alternative solution to #2788 (hence closes #2788). Obviously I prefer this one. But let me know what you think. Also, I might still have gotten some things wrong with the traits (I tried to do something like "subtraits"?), so better have a look there, please.

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -136,7 +136,7 @@ struct HasNoSingularGroebnerAlgorithm <: HasSingularGroebnerAlgorithmTrait end
 
 # By default we do not expect the Singular backend to handle the case
 HasSingularNormalFormTrait(a::Any) = HasSingularNormalFormTrait(typeof(a))
-HasSingularNormalFormTrait(::Type) = HasNoSingularNormalForm()
+HasSingularNormalFormTrait(::Type{T}) where {T} = HasNoSingularNormalForm()
 
 # Same for the Groebner bases. But if a normal form algorithm exists, then 
 # this works, too.
@@ -151,8 +151,8 @@ HasSingularNormalFormTrait(::Type{<:FieldElem}) = HasSingularNormalForm()
 HasSingularNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
 
 # For polynomial rings over the integers we only allow global orderings
-HasSingularGroebnerAlgorithmTrait(::Type{<:ZZRingElem}) = HasSingularGroebnerAlgorithm()
-HasSingularGroebnerAlgorithmTrait(::Type{<:ZZRing}) = HasSingularGroebnerAlgorithm()
+HasSingularGroebnerAlgorithmTrait(::Type{ZZRingElem}) = HasSingularGroebnerAlgorithm()
+HasSingularGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
 
 # This list can (and should) be extended by eventual new types which 
 # are supposed to make use of the Singular backend.
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -154,35 +154,10 @@ HasSingularNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
 HasSingularGroebnerAlgorithmTrait(::Type{<:ZZRingElem}) = HasSingularGroebnerAlgorithm()
 HasSingularGroebnerAlgorithmTrait(::Type{<:ZZRing}) = HasSingularGroebnerAlgorithm()
 
-### Deprecated code below
-
-is_coefficient_type_with_singular_normal_form(a::Any) = is_coefficient_type_with_singular_normal_form(typeof(a))
-
-# By default we can not expect that there is a singular backend
-is_coefficient_type_with_singular_normal_form(::Type{T}) where {T} = false
-
-# There might be rings for which groebner basis algorithms exist only 
-# for global orderings, so we make a refinement here
-is_coefficient_type_with_groebner_basis_algorithm_in_singular(a::Any) = is_coefficient_type_with_groebner_basis_algorithm_in_singular(typeof(a))
-
-function is_coefficient_type_with_groebner_basis_algorithm_in_singular(::Type{T}) where {T}
-  is_coefficient_type_with_singular_normal_form(T) && return true
-  return false
-end
-
-# For polynomial rings over fields we expect singular to be able to handle the case
-is_coefficient_type_with_singular_normal_form(::Type{T}) where {T<:FieldElem} = true
-is_coefficient_type_with_singular_normal_form(::Type{T}) where {T<:Field} = true
-
-# Singular can also handle groebner bases over the integers
-is_coefficient_type_with_groebner_basis_algorithm_in_singular(::Type{T}) where {T<:ZZRingElem} = true
-is_coefficient_type_with_groebner_basis_algorithm_in_singular(::Type{T}) where {T<:ZZRing} = true
-
 # This list can (and should) be extended by eventual new types which 
 # are supposed to make use of the Singular backend.
 # In particular, this decides whether a reasonable Hash function for 
 # elements in the quotient ring exists.
-
 
 ##############################################################################
 #
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -119,7 +119,7 @@ end
 # provide the following functionality to decide the existence and use 
 # of a Singular backend depending on the type.
 #
-# Since the concept of traits in julia is not uniform, we briefly 
+# Since the concept of traits in Julia is not uniform, we briefly 
 # describe how to set up your own type of coefficients with this
 # framework. 
 #
@@ -177,10 +177,12 @@ struct HasRingFlattening <: HasGroebnerAlgorithmTrait end
 struct HasNoGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
 
 # By default we do not expect the Singular backend to be able to compute normal forms
+HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()
+
+# for convenience, allow passing in rings, ring elements or ring element types
 HasNormalFormTrait(a::Ring) = HasNormalFormTrait(typeof(a))
 HasNormalFormTrait(a::RingElem) = HasNormalFormTrait(parent_type(a))
 HasNormalFormTrait(::Type{T}) where {T <: RingElem} = HasNormalFormTrait(parent_type(T))
-HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()
 
 # Same for the Groebner bases. But if a normal form algorithm exists, then 
 # Groebner bases can be computed, too.
@@ -196,7 +198,8 @@ end
 # For polynomial rings over fields we expect Singular to be able to compute normal forms
 HasNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
 
-# For polynomial rings over the integers we only allow global orderings
+# For polynomial rings over the integers we can compute Gröbner bases but
+# in general can not compute normal forms.
 HasGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
 HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm()
 
@@ -204,7 +207,7 @@ HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm
 # are supposed to make use of the Singular backend; see above.
 
 # In particular, it is decided based on this trait whether a reasonable 
-# Hash function for elements in the quotient ring exists.
+# hash function for elements in the quotient ring exists.
 
 # In some cases it is useful to know that we can do a RingFlattening 
 # and carry out certain operations in the de-nested ring
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -870,7 +870,7 @@ end
 # By default we refer to the generic ideal membership routine which 
 # might be implemented by other means, for instance via a `RingFlattening`.
 function _is_equal(::HasNoSingularGroebnerAlgorithm, f::MPolyQuoRingElem{T}, g::MPolyQuoRingElem{T}) where T
-  return f.f - g.f in modulus(Q)
+  return f.f - g.f in modulus(parent(f))
 end
 
 @doc raw"""
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -117,13 +117,54 @@ end
 # To make sure that the data structure for quotient rings can 
 # nevertheless also accomodate more exotic coefficient rings we 
 # provide the following functionality to decide the existence and use 
-# of a singular backend depending on the type.
+# of a Singular backend depending on the type.
+#
+# Since the concept of traits in julia is not uniform, we briefly 
+# describe how to set up your own type of coefficients with this
+# framework. 
+#
+# Say you have a new type `MyType` for the `coefficient_ring` of a 
+# polynomial ring `P` and you would like to make the `MPolyQuo` 
+# structure useful for quotients of the form `P/I`. Then you would 
+# declare 
+#
+#   HasGroebnerAlgorithmTrait(::Type{MyType}) = HasSingularGroebnerAlgorithm()
+#
+# in case you are sure that the generic code to use Singular as a 
+# backend can also digest polynomial rings whose coefficient ring 
+# is of type `MyType`. 
+#
+# If you need to implement your own backend, you do the following.
+# You declare
+#
+#   HasGroebnerAlgorithmTrait(::Type{MyType}) = HasMyCustomBackend()
+#
+# where `HasMyCustomBackend` is a name of your choice and then implement
+#
+# function _simplify(::HasMyCustomBackend, f::MPolyQuoRingElem)
+#   # Do whatever has to be done to achieve a unique representative
+#   ...
+# end
+#
+# function _hash(::HasMyCustomBackend, f::MPolyQuoRingElem, u::UInt)
+#   # Implement a hash which is unique for the class, NOT the 
+#   # representative!
+#   ...
+# end
+#
+# function _is_equal(::HasMyCustomBackend, f::MPolyRingElem, g::MPolyQuoRingElem)
+#   # Implement an equality check.
+#   ...
+# end
+#
 ########################################################################
 
-# The trait of the coefficient ring and its elements to decide on 
-# whether or not we do have a singular backend for groebner basis 
-# computations
+# The trait of the coefficient ring and its elements to decide 
+# which backend to use for Groebner basis and normal form computations.
 abstract type HasGroebnerAlgorithmTrait end
+# Having a normal form algorithm available is strictly stronger than 
+# having a Grobner basis algorithm. Thus we derive one type from the
+# other here. 
 abstract type HasNormalFormTrait <: HasGroebnerAlgorithmTrait end
 
 # A normal form algorithm (for possibly non-global orderings) requires 
@@ -135,14 +176,14 @@ struct HasSingularGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
 struct HasRingFlattening <: HasGroebnerAlgorithmTrait end
 struct HasNoGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
 
-# By default we do not expect the Singular backend to handle the case
+# By default we do not expect the Singular backend to be able to compute normal forms
 HasNormalFormTrait(a::Ring) = HasNormalFormTrait(typeof(a))
 HasNormalFormTrait(a::RingElem) = HasNormalFormTrait(parent_type(a))
 HasNormalFormTrait(::Type{T}) where {T <: RingElem} = HasNormalFormTrait(parent_type(T))
 HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()
 
 # Same for the Groebner bases. But if a normal form algorithm exists, then 
-# this works, too.
+# Groebner bases can be computed, too.
 HasGroebnerAlgorithmTrait(a::Ring) = HasGroebnerAlgorithmTrait(typeof(a))
 HasGroebnerAlgorithmTrait(a::RingElem) = HasGroebnerAlgorithmTrait(parent_type(a))
 HasGroebnerAlgorithmTrait(::Type{T}) where {T <: RingElem} = HasGroebnerAlgorithmTrait(parent_type(T))
@@ -152,21 +193,23 @@ function HasGroebnerAlgorithmTrait(::Type{T}) where {T}
   return HasNoGroebnerAlgorithm()
 end
 
-# For polynomial rings over fields we expect Singular to handle the case
+# For polynomial rings over fields we expect Singular to be able to compute normal forms
 HasNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
 
 # For polynomial rings over the integers we only allow global orderings
 HasGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
 HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm()
 
+# This list can (and should) be extended by eventual new types which 
+# are supposed to make use of the Singular backend; see above.
+
+# In particular, it is decided based on this trait whether a reasonable 
+# Hash function for elements in the quotient ring exists.
+
 # In some cases it is useful to know that we can do a RingFlattening 
 # and carry out certain operations in the de-nested ring
 # These declarations happen in the file src/Rings/MPolyMap/flattenings.jl
-
-# This list can (and should) be extended by eventual new types which 
-# are supposed to make use of the Singular backend.
-# In particular, this decides whether a reasonable Hash function for 
-# elements in the quotient ring exists.
+# and need to be postponed due to limitations in inclusion orders.
 
 ##############################################################################
 #
@@ -846,7 +889,7 @@ end
 # like equality and hashing will throw an error.
 function _simplify(::HasNoGroebnerAlgorithm, f::MPolyQuoRingElem)
   return f
-  # error("no groebner backend available for simplification; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
+  # error("no groebner backend available for simplification; you can specify a groebner backend by implementing the `HasGroebnerAlgorithmTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
 end
 
 @doc raw"""
@@ -892,7 +935,7 @@ function _is_equal(::HasRingFlattening, f::MPolyQuoRingElem{T}, g::MPolyQuoRingE
 end
 
 function _is_equal(::HasNoGroebnerAlgorithm, f::MPolyQuoRingElem{T}, g::MPolyQuoRingElem{T}) where T
-  error("no groebner backend available for equality check; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
+  error("no groebner backend available for equality check; you can specify a groebner backend by implementing the `HasGroebnerAlgorithmTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
 end
 
 @doc raw"""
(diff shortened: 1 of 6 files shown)


src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,64 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# strictly more, so we make it a subcase.
+struct HasSingularNormalForm <: HasSingularNormalFormTrait end
+struct HasNoSingularNormalForm <: HasSingularNormalFormTrait end
+
+struct HasSingularGroebnerAlgorithm <: HasSingularGroebnerAlgorithmTrait end
+struct HasNoSingularGroebnerAlgorithm <: HasSingularGroebnerAlgorithmTrait end
+
+# By default we do not expect the Singular backend to handle the case
+HasSingularNormalFormTrait(a::Any) = HasSingularNormalFormTrait(typeof(a))
+HasSingularNormalFormTrait(::Type) = HasNoSingularNormalForm()
```
thofma: Shoud also use `::Type{T}`

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,64 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+  return HasNoSingularGroebnerAlgorithm()
+end
+
+# For polynomial rings over fields we expect Singular to handle the case
+HasSingularNormalFormTrait(::Type{<:FieldElem}) = HasSingularNormalForm()
+HasSingularNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
+
+# For polynomial rings over the integers we only allow global orderings
+HasSingularGroebnerAlgorithmTrait(::Type{<:ZZRingElem}) = HasSingularGroebnerAlgorithm()
+HasSingularGroebnerAlgorithmTrait(::Type{<:ZZRing}) = HasSingularGroebnerAlgorithm()
```
thofma:
```suggestion
HasSingularGroebnerAlgorithmTrait(::Type{ZZRingElem}) = HasSingularGroebnerAlgorithm()
HasSingularGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
```

HechtiDerLachs:
Thanks for the catches!

simonbrandhorst:
Any speed difference between the two approaches?

HechtiDerLachs:
> Any speed difference between the two approaches?

I suppose not. If I understand @fingolfin correctly, the compiler should be able to optimize all if-clauses away (Can you say that in English?) . This should similarly be the case with traits, so the actual machine code shouldn't differ too much. What is improved compared to #2788 is the readability and the fact that we are implementing a well understood and accepted pattern instead of messing around with our own recipes for `if`-`else`-spaghetti.

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,64 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# strictly more, so we make it a subcase.
+struct HasSingularNormalForm <: HasSingularNormalFormTrait end
+struct HasNoSingularNormalForm <: HasSingularNormalFormTrait end
+
+struct HasSingularGroebnerAlgorithm <: HasSingularGroebnerAlgorithmTrait end
+struct HasNoSingularGroebnerAlgorithm <: HasSingularGroebnerAlgorithmTrait end
+
+# By default we do not expect the Singular backend to handle the case
+HasSingularNormalFormTrait(a::Any) = HasSingularNormalFormTrait(typeof(a))
+HasSingularNormalFormTrait(::Type{T}) where {T} = HasNoSingularNormalForm()
```
lgoettgens:
```suggestion
HasSingularNormalFormTrait(R::NCRing) = HasSingularNormalFormTrait(typeof(a))
HasSingularNormalFormTrait(a::NCRingElem) = HasSingularNormalFormTrait(parent(a))
HasSingularNormalFormTrait(::Type{T}) where {T} = HasNoSingularNormalForm()
```
Wouldn't this make it a bit easier, as down below you would only need to specify traits for the container types, and not for element types? This seems like a good place for future errors (having differences between traits of containers and its elements).

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,64 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+  return HasNoSingularGroebnerAlgorithm()
+end
+
+# For polynomial rings over fields we expect Singular to handle the case
+HasSingularNormalFormTrait(::Type{<:FieldElem}) = HasSingularNormalForm()
+HasSingularNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
+
+# For polynomial rings over the integers we only allow global orderings
+HasSingularGroebnerAlgorithmTrait(::Type{ZZRingElem}) = HasSingularGroebnerAlgorithm()
+HasSingularGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
```
lgoettgens:
```suggestion
HasSingularNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()

# For polynomial rings over the integers we only allow global orderings
HasSingularGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
```
Like this

fieker (approved):

afkafkafk13:
To me this is a lot more readable than #2788 .

src/Rings/MPolyQuo.jl
HechtiDerLachs: Alright, this seems reasonable.

simonbrandhorst (approved):

afkafkafk13 (approved):

src/Rings/MPolyQuo.jl
lgoettgens:
```suggestion
HasSingularNormalFormTrait(a::Any) = HasSingularNormalFormTrait(typeof(a))
HasSingularNormalFormTrait(::Type{T}) where {T <: NCRingElem} = HasSingularNormalFormTrait(parent_type(T))
HasSingularNormalFormTrait(::Type{T}) where {T <: NCRing} = HasNoSingularNormalForm()
```
Or even like this. Then, it works for all four of container, container type, element, element type. While one still only has to define it for the container type (as in https://github.com/oscar-system/Oscar.jl/pull/2789#discussion_r1324235900).

afkafkafk13:
@HechtiDerLachs : What is the status here? Is this one ready for merging and the other alternative PR #2788 is planned to be closed at the same time?

HechtiDerLachs:
Yes.

src/Rings/MPolyQuo.jl
```diff
@@ -1320,10 +1403,25 @@ function grading_group(A::MPolyQuoRing{<:MPolyDecRingElem})
...
+  return _hash(HasGroebnerAlgorithmTrait(kk), w, u)
+end
+
+function _hash(::HasSingularGroebnerAlgorithm, w::MPolyQuoRingElem, u::UInt)
   simplify(w)
   return hash(w.f, u)
 end
 
+function _hash(::HasRingFlattening, w::MPolyQuoRingElem, u::UInt)
+  error("hash function not implemented due to lack of unique representatives; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(w)))))`")
```
fingolfin: There is no `HasGroebnerBackendTrait`. I assume you mean `HasGroebnerAlgorithmTrait`?
```suggestion
  error("hash function not implemented due to lack of unique representatives; you can specify a groebner backend by implementing the `HasGroebnerAlgorithmTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(w)))))`")
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+#
+# In general, we can not expect a normal form algorithm to exist and, 
+# in particular, that it is implemented in Singular. However, we wish 
+# to use Singular as a default backend and this also drives the design 
+# of the quotient rings to begin with. 
+#
+# To make sure that the data structure for quotient rings can 
+# nevertheless also accomodate more exotic coefficient rings we 
+# provide the following functionality to decide the existence and use 
+# of a singular backend depending on the type.
```
fingolfin:
```suggestion
# of a Singular backend depending on the type.
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+########################################################################
+
+# The trait of the coefficient ring and its elements to decide on 
+# whether or not we do have a singular backend for groebner basis
```
fingolfin:
```suggestion
# whether or not we do have a Singular backend for Groebner basis 
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# to use Singular as a default backend and this also drives the design 
+# of the quotient rings to begin with. 
+#
+# To make sure that the data structure for quotient rings can 
+# nevertheless also accomodate more exotic coefficient rings we 
+# provide the following functionality to decide the existence and use 
+# of a singular backend depending on the type.
+########################################################################
+
+# The trait of the coefficient ring and its elements to decide on
```
fingolfin: There is no uniformly agreed concept of what a "trait" is in Julia, and how exactly to use it. Thus just saying "this is a trait" is not really enough to let people know how to use it.

In this case, one can try and stare at the code for a while and perhaps guess the patter, but it would be really helpful if it was actually explained how to set it up, and also what the various options means.

I am also not so sure if it is a the best idea to use the type `HasGroebnerAlgorithmTrait` also as the main dispatch function -- this could just as well be done by a seperate helper, say `groebner_algorithm_trait` or whatever (the "Has" to me suggests a function returning a boolean, so I'd drop it here).

That has the advantage of untangling the concept of "base type for a hierarchy of 'trait' types" versus "function to determine for a given type which traits it has"

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# strictly more, so we make it a subcase.
+struct HasSingularNormalForm <: HasNormalFormTrait end
+struct HasNoNormalForm <: HasNormalFormTrait end
+
+struct HasSingularGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
+struct HasRingFlattening <: HasGroebnerAlgorithmTrait end
+struct HasNoGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
+
+# By default we do not expect the Singular backend to handle the case
+HasNormalFormTrait(a::Ring) = HasNormalFormTrait(typeof(a))
```
fingolfin, line 183: It seems `HasNormalFormTrait` is a second top level trait getter, separate from `HasGroebnerAlgorithmTrait` yet "derived" from it. I find that confusing. So is `HasGroebnerAlgorithmTrait` allowed to return `HasSingularNormalForm` or `HasNoNormalForm`? Is it supposed to?

All in all, perhaps it would be better if `HasNormalFormTrait` was its own top level type, and not derived from `HasGroebnerAlgorithmTrait`? Or is there a  reason for this connection after all, which I am missing? If so please explain.

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+HasNormalFormTrait(::Type{T}) where {T <: RingElem} = HasNormalFormTrait(parent_type(T))
+HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()
+
+# Same for the Groebner bases. But if a normal form algorithm exists, then 
+# this works, too.
+HasGroebnerAlgorithmTrait(a::Ring) = HasGroebnerAlgorithmTrait(typeof(a))
+HasGroebnerAlgorithmTrait(a::RingElem) = HasGroebnerAlgorithmTrait(parent_type(a))
+HasGroebnerAlgorithmTrait(::Type{T}) where {T <: RingElem} = HasGroebnerAlgorithmTrait(parent_type(T))
+
+function HasGroebnerAlgorithmTrait(::Type{T}) where {T}
```
fingolfin, line 193: Should this perhaps be
```suggestion
function HasGroebnerAlgorithmTrait(::Type{T}) where {T <: Ring}
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+  HasNormalFormTrait(T) isa HasSingularNormalForm && return HasSingularGroebnerAlgorithm()
```
fingolfin, line 194: So if nothing else is said about a type, then we inspect its `HasNormalFormTrait`. So that's an implication resp. delegation between traits. OK. Though I think it would be helpful to add a comment to that effect

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+  return HasNoGroebnerAlgorithm()
+end
+
+# For polynomial rings over fields we expect Singular to handle the case
```
fingolfin: What is "the case"? Perhaps
```suggestion
# For polynomial rings over fields we expect Singular to be able to even
# compute normal forms
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+HasNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
+
+# For polynomial rings over the integers we only allow global orderings
```
fingolfin: I am rather confused by this comment; what has the ordering to do with the `HasGroebnerAlgorithmTrait`?

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# A normal form algorithm (for possibly non-global orderings) requires 
+# strictly more, so we make it a subcase.
+struct HasSingularNormalForm <: HasNormalFormTrait end
+struct HasNoNormalForm <: HasNormalFormTrait end
+
+struct HasSingularGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
+struct HasRingFlattening <: HasGroebnerAlgorithmTrait end
+struct HasNoGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
+
+# By default we do not expect the Singular backend to handle the case
```
fingolfin: What is "the case"? Maybe you meant something like this?
```suggestion
# By default we do not expect the Singular backend to be able to compute normal forms
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+HasNormalFormTrait(a::Ring) = HasNormalFormTrait(typeof(a))
+HasNormalFormTrait(a::RingElem) = HasNormalFormTrait(parent_type(a))
+HasNormalFormTrait(::Type{T}) where {T <: RingElem} = HasNormalFormTrait(parent_type(T))
+HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()
+
+# Same for the Groebner bases. But if a normal form algorithm exists, then 
+# this works, too.
```
fingolfin: What "same"? And what does "this works" refer to?

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# For polynomial rings over the integers we only allow global orderings
+HasGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
+HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm()
+
+# In some cases it is useful to know that we can do a RingFlattening 
+# and carry out certain operations in the de-nested ring
+# These declarations happen in the file src/Rings/MPolyMap/flattenings.jl
+
+# This list can (and should) be extended by eventual new types which 
+# are supposed to make use of the Singular backend.
```
fingolfin: What is "this list"? That of `HasGroebnerAlgorithmTrait` methods above? In the case I'd move these two lines above those talking about ring flattening

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,73 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# For polynomial rings over fields we expect Singular to handle the case
+HasNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
+
+# For polynomial rings over the integers we only allow global orderings
+HasGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
+HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm()
+
+# In some cases it is useful to know that we can do a RingFlattening 
+# and carry out certain operations in the de-nested ring
+# These declarations happen in the file src/Rings/MPolyMap/flattenings.jl
```
fingolfin, line 214: Should this comment perhaps be near `HasRingFlattening`? At least I assume it refers to that? Or not?

src/Rings/MPolyQuo.jl
```diff
@@ -1320,10 +1403,25 @@ function grading_group(A::MPolyQuoRing{<:MPolyDecRingElem})
...
   return hash(w.f, u)
 end
 
+function _hash(::HasRingFlattening, w::MPolyQuoRingElem, u::UInt)
+  error("hash function not implemented due to lack of unique representatives; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(w)))))`")
+end
+
+function _hash(::HasNoGroebnerAlgorithm, w::MPolyQuoRingElem, u::UInt)
+  error("hash function not implemented due to lack of unique representatives; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(w)))))`")
+end
```
fingolfin: Instead of having multiple identical methods, you can now make use of the type hierarchy you defined, and simply define one method for *all* `HasGroebnerAlgorithmTrait` subtypes. 
```suggestion
function _hash(::HasGroebnerAlgorithmTrait, w::MPolyQuoRingElem, u::UInt)
  error("hash function not implemented due to lack of unique representatives; you can specify a groebner backend by implementing the `HasGroebnerAlgorithmTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(w)))))`")
end
```

fingolfin:
Just to say, overall I like this PR; my comments mostly aim at making sure the resulting code is long-term maintainable ;-)

src/Rings/MPolyQuo.jl
afkafkafk13: Normalform Algorithms for global and local/mixed ordering differ significantly. In particular, Singular has reliable Normalform Algorithms (and StandardBasis Algorithms) for global and local/mixed orderings over (Singular-)fields and for global orderings over the integers and rings of the form ZZ/n, but not for local/mixed orderings. 

So yes, for determining the Trait you need to look at the Ring and its ordering (or more precisely the localization of the polynomial ring in queston at its ordering).

HechtiDerLachs: No, this is about something different. I assume we do have a Groebner backend in singular for global orderings, but I am not aware that we can use arbitrary orderings over the integers. Correct me if I'm wrong.

HechtiDerLachs:
> Just to say, overall I like this PR; my comments mostly aim at making sure the resulting code is long-term maintainable ;-)

Alright. Thanks for having a look and so many constructive comments. However, if there is the need to discuss all these things in this detail (which there apparently is...) I would like to first do this in person at this point. I started this with the idea of just providing a quick fix. Now this is waiting to be merged since more than a week. And since this morning it doesn't look good anymore. Frankly, this is a bit frustrating. And I will not have time for this until after our meeting in Saarbruecken.

src/Rings/MPolyQuo.jl
fingolfin: Whatever the explanation is, my points are that (a) the comment should state them and not require additional outside explanation, and (b) I actually see nothing in the code dealing with / checking for orderings at all, so I am even more confused by the comment :-(

fingolfin:
> Alright. Thanks for having a look and so many constructive comments. However, if there is the need to discuss all these things in this detail (which there apparently is...) I would like to first do this in person at this point. 

I feel there is some miscommunication here. The majority of my comments refers to issues in comments: explanations that are incomplete or seem to reference things that are not there. In several cases I even added suggestions for potential fixes which you could apply with a single click in case my guess was right. OK, in the case of `HasGroebnerBackendTrait` you can't click, but search&replace locally should be easy enough?

So, for these, there is no need to discuss anything, just clarify the wording of the comments -- I would do it myself, but I can't, only you can. That resolves the majority of my comments, and my guess would be that it could be done with maybe 10 minutes of work?

The only things that maybe could need discussion is the "type hierarchy" you used; and whether to separate the type `HasGroebnerAlgorithmTrait` from the function querying the trait. But not even there I am sure we need to discuss much -- I asked questions of the kind "do you have a reason for that?", can be resolved quite easily by you: if you have a reason, add a comment naming the reason; if there isn't a reason, just adjust it.



> I started this with the idea of just providing a quick fix. Now this is waiting to be merged since more than a week.

Yes, I am surprised by this, too. Three people already approved this PR, but nobody thought of merging it? Not my fault... though I am glad because it's easier for me to leave these remarks then to submit them afterwards in an issue.

> And since this morning it doesn't look good anymore. Frankly, this is a bit frustrating. And I will not have time for this until after our meeting in Saarbruecken.

Please note that I specifically did not "request changes" to this PR, so as to not block it from being merged. If you prefer to make the fixes to the typos I pointed out later, also fine by me, though applying them later probably is more work than clicking "commit suggestion" here, but your choice :-). For the design concerns, I think it makes perfect sense to address them later, though I would then ask you to please open an issue as a reminder. The issue can consist of a single line, e.g.: "Resolve `HasGroebnerAlgorithmTrait` design questions from PR #2789".

src/Rings/MPolyQuo.jl
```diff
@@ -783,6 +841,13 @@ function simplify(f::MPolyQuoRingElem)
   return f::elem_type(parent(f))
 end
 
+# Having the simplify method do nothing still allows for arithmetic to operate 
+# (simplify is automatically called for multiplications), but further functionality 
+# like equality and hashing will throw an error.
+function _simplify(::HasNoGroebnerAlgorithm, f::MPolyQuoRingElem)
+  return f
+  # error("no groebner backend available for simplification; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
```
fingolfin:
```suggestion
  # error("no groebner backend available for simplification; you can specify a groebner backend by implementing the `HasGroebnerAlgorithmTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
```
or perhaps even
```suggestion
```

src/Rings/MPolyQuo.jl
```diff
@@ -807,11 +872,29 @@ true
...
 end
 
+# By default we refer to the generic ideal membership routine which 
+# might be implemented by other means, for instance via a `RingFlattening`.
+function _is_equal(::HasRingFlattening, f::MPolyQuoRingElem{T}, g::MPolyQuoRingElem{T}) where T
+  return f.f - g.f in modulus(parent(f))
+end
+
+function _is_equal(::HasNoGroebnerAlgorithm, f::MPolyQuoRingElem{T}, g::MPolyQuoRingElem{T}) where T
+  error("no groebner backend available for equality check; you can specify a groebner backend by implementing the `HasGroebnerBackendTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
```
fingolfin: My guess: you meant
```suggestion
  error("no groebner backend available for equality check; you can specify a groebner backend by implementing the `HasGroebnerAlgorithmTrait(::Type{T}) = MyBackend()` for `T = $(typeof(coefficient_ring(base_ring(parent(f)))))`")
```

src/Rings/MPolyQuo.jl
HechtiDerLachs: That's a good point and I was also not sure about this. I did this following the general principle that types become more and more special the further down the tree they appear. In this case having a normal form algorithm is more special than just having a groebner basis algorithm. But the first implies the latter. But I also see no benefits to introducing this hierarchy here, to be honest.

HechtiDerLachs:
I hope, I now addressed all things. From my side this should be ready for merge finally.

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,116 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+# in particular, that it is implemented in Singular. However, we wish 
+# to use Singular as a default backend and this also drives the design 
+# of the quotient rings to begin with. 
+#
+# To make sure that the data structure for quotient rings can 
+# nevertheless also accomodate more exotic coefficient rings we 
+# provide the following functionality to decide the existence and use 
+# of a Singular backend depending on the type.
+#
+# Since the concept of traits in julia is not uniform, we briefly
```
fingolfin:
```suggestion
# Since the concept of traits in Julia is not uniform, we briefly 
```

Also, thank you for this comment explaining the traits, it is very helpful!

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,116 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+
+struct HasSingularGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
+struct HasRingFlattening <: HasGroebnerAlgorithmTrait end
+struct HasNoGroebnerAlgorithm <: HasGroebnerAlgorithmTrait end
+
+# By default we do not expect the Singular backend to be able to compute normal forms
+HasNormalFormTrait(a::Ring) = HasNormalFormTrait(typeof(a))
+HasNormalFormTrait(a::RingElem) = HasNormalFormTrait(parent_type(a))
+HasNormalFormTrait(::Type{T}) where {T <: RingElem} = HasNormalFormTrait(parent_type(T))
+HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()
```
fingolfin:
```suggestion
# By default we do not expect the Singular backend to be able to compute normal forms
HasNormalFormTrait(::Type{T}) where {T} = HasNoNormalForm()

# for convenience, allow passing in rings, ring elements or ring element types
HasNormalFormTrait(a::Ring) = HasNormalFormTrait(typeof(a))
HasNormalFormTrait(a::RingElem) = HasNormalFormTrait(parent_type(a))
HasNormalFormTrait(::Type{T}) where {T <: RingElem} = HasNormalFormTrait(parent_type(T))
```

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,116 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+
+# For polynomial rings over the integers we only allow global orderings
+HasGroebnerAlgorithmTrait(::Type{ZZRing}) = HasSingularGroebnerAlgorithm()
+HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm()
+
+# This list can (and should) be extended by eventual new types which 
+# are supposed to make use of the Singular backend; see above.
+
+# In particular, it is decided based on this trait whether a reasonable 
+# Hash function for elements in the quotient ring exists.
```
fingolfin:
```suggestion
# hash function for elements in the quotient ring exists.
```

fingolfin (approved):
Thank you, *much* better!!! You left a few of my comments unresolved, let's briefly talk about them and decided whether to merge or firs tweak

src/Rings/MPolyQuo.jl
```diff
@@ -101,6 +101,116 @@ function Base.deepcopy_internal(a::MPolyQuoRingElem, dict::IdDict)
...
+
+function HasGroebnerAlgorithmTrait(::Type{T}) where {T}
+  HasNormalFormTrait(T) isa HasSingularNormalForm && return HasSingularGroebnerAlgorithm()
+  return HasNoGroebnerAlgorithm()
+end
+
+# For polynomial rings over fields we expect Singular to be able to compute normal forms
+HasNormalFormTrait(::Type{<:Field}) = HasSingularNormalForm()
+
+# For polynomial rings over the integers we only allow global orderings
```
fingolfin:
```suggestion
# For polynomial rings over the integers we can compute Gröbner bases but
# in general can not compute normal forms.
```

PR #2792, opened 2023-09-13
add permutation action on `FreeAssAlgElem`s

ThomasBreuer:
(can be merged as soon as Nemocas/AbstractAlgebra.jl/pull/1418 is available in Oscar)

### DIFF
--- a/src/Groups/action.jl
+++ b/src/Groups/action.jl
@@ -284,6 +284,7 @@ end
 """
     on_indeterminates(f::GAP.GapObj, p::PermGroupElem)
     on_indeterminates(f::MPolyRingElem, p::PermGroupElem)
+    on_indeterminates(f::FreeAssAlgElem, p::PermGroupElem)
     on_indeterminates(f::MPolyIdeal, p::PermGroupElem)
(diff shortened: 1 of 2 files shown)


thofma:
I wonder whats up with coverage.

ThomasBreuer:
The failures look different than earlier ones (where either the job got canceled or some specific error was shown). How can I interpret the current failures?

benlorenz:
> I wonder whats up with coverage.

> > Report is 62 commits behind head on master.
> > The diff coverage is `0.00%`.

This seems to be comparing a pretty old commit from master (base commit of this branch, 16 days ago) with the merge commit from this branch with master (since CI runs on a temporary merge)?

benlorenz:
> The failures look different than earlier ones (where either the job got canceled or some specific error was shown). How can I interpret the current failures?

Sometimes it helps to use `View raw logs`, but in this case one of those also doesn't work. The failures for 1.10 and nightly on ubuntu can be ignored and are caused by the usual OOM issues.

thofma:
Let's see if a rebase fixes the coverage

thofma:
Still not working, no clue why

thofma (approved):

PR #2794, opened 2023-09-13
Remove redundant pretty calls, fix various `print` calls missing the `io` argument

fingolfin:
Also fix `Base.show(io::IO, S::MPolyLeadingMonOne)` method to set
`supercompact` to `true` *after* checking whether it is equal to `true`.


lgoettgens (approved):

thofma:
doctests broken?

src/Rings/mpolyquo-localizations.jl
```diff
@@ -219,7 +219,6 @@ function Base.show(io::IO, ::MIME"text/plain", L::MPolyQuoLocRing)
 end
 
 function Base.show(io::IO, L::MPolyQuoLocRing)
-  io = pretty(io)
   if get(io, :supercompact, false)
     print(io, "Localized quotient of multivariate polynomial ring")
   else
```
lgoettgens, line 226: RIght below this line, there is a similar thing wrong with the order of calling `pretty` and setting `supercompact => true` as the one already fixed for `MPolyLeadingMonOne`. The doctest failure seems related to this.

fingolfin: I fixed the doctest, but I couldn't find the spot you are talking about (I think?). If you still see an issue, please pinpoint me. Thanks!

lgoettgens: You already fixed it

fingolfin:
Running into a bug in AA (see https://github.com/Nemocas/AbstractAlgebra.jl/pull/1447) will try to work around it.

PR #2796, opened 2023-09-13
Bump dependencies

lgoettgens:
This prepares Oscar for the breaking releases of AbstractAlgebra v0.32.0, Nemo v0.36.0, and Hecke v0.21.0.

cc @thofma @fingolfin


src/Rings/MPolyMap/MPolyRing.jl
```diff
@@ -59,7 +59,7 @@ defined by
   x -> y
   y -> x
 with map on coefficients
-#1
+  #1
```
lgoettgens, line 62: @simonbrandhorst do you have an idea what's happening here? The change is possibly due to https://github.com/Nemocas/AbstractAlgebra.jl/pull/1436, but the printing was weird before as well.

thofma: Its an anonymous function

lgoettgens:
The CI job reports:
```
Invalidations on default branch: 1823 (1823 via deps)
This branch: 1826 (1826 via deps)
```

This low increase seems to be okay for the large amount of stuff moved from Nemo to AA.

thofma:
This is just https://github.com/Nemocas/AbstractAlgebra.jl/pull/1436

PR #2798, opened 2023-09-14
PolyhedralGeometry: Enhance docs of solve_mixed

lkastner:


HechtiDerLachs:
Seems reasonable, thanks for the work. The original issue which finally lead to this PR was the following: The docstring says we solve a system `A*x = b` under some condition. Here `A` and `B` are both matrices. Hence, given what I was used to from the context of `lift`-like commands from commutative algebra, I was expecting also `x` to be a matrix; probably the one who's `j`-th column multiplied with `A` becomes the `j`-th column of `b`. What was confusing me was that what I got was a matrix `x` which I could not even multiply with `A`! Example:
```julia
julia> A = ZZ[1 0; 1 -1]
[1    0]
[1   -1]

julia> B = ZZ[1; 1]
[1]
[1]

julia> C = ZZ[1 0; 0 1]
[1   0]
[0   1]

julia> x = solve_mixed(ZZMatrix, A, B, C)
[1   0]

julia> A*x
ERROR: Incompatible matrix dimensions
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] *(x::ZZMatrix, y::ZZMatrix)
   @ Nemo ~/.julia/packages/Nemo/Tj33G/src/flint/fmpz_mat.jl:312
 [3] top-level scope
   @ REPL[95]:1
```
This is something that I personally still find confusing. But I understand that a lot of demands have already gone into the design of this function and I am not the only one who is confused about something. The docstring now might have helped me better to overcome my confusion, so I think this is an improvement.

HechtiDerLachs (approved):

PR #2799, opened 2023-09-14
automorphism_group fix for rank 0 matroids + added test

Sequenzer:
Rank zero matroids would cause an error, which has been fixed by this.


src/Combinatorics/Matroids/matroids.jl
```diff
@@ -864,4 +864,7 @@ julia> automorphism_group(M)
 Group([ (3,4), (1,2), (2,3) ])
 ```
 """
-automorphism_group(m::Matroid) = automorphism_group(IncidenceMatrix(bases(m));action=:on_cols)
+function automorphism_group(m::Matroid) 
+  length(m) > 0 || throw(ArgumentError("The matroid should not be empty."))
```
lgoettgens:
```suggestion
  @req length(m) > 0 "The matroid should not be empty."
```

src/Combinatorics/Matroids/matroids.jl
```diff
@@ -864,4 +864,13 @@ julia> automorphism_group(M)
...
+function automorphism_group(m::Matroid) 
+  @req length(m) > 0 "The matroid should not be empty."
+  I = rank(m) < 1 ? IncidenceMatrix(bases(dual_matroid(m))) : IncidenceMatrix(bases(m))
+  resize!(I, nrows(I), length(m))
+  return automorphism_group(I;action=:on_cols)
+end
+
+
+
+
```
benlorenz:
```suggestion
  return automorphism_group(I; action=:on_cols)
end
```

benlorenz:
```suggestion
```

antonydellavecchia (approved):

PR #2801, opened 2023-09-14
Move `test_save_load_roundtrip` to init file

lgoettgens:
When running the tests with multiple workers, the `FTheoryTools` and `QuadFormAndIso` tests may fail, when they are run on a different worker than the serialization tests. This is due to the latter adding a test function `test_save_load_roundtrip` that is used by the former two.
My solution is to move the `test_save_load_roundtrip` function to a file that gets included `@everywhere` right before running any tests. When similar things come up in the future, we may put them there as well.
The name of the file is open for discussion.


test/runtests.jl
```diff
@@ -104,7 +104,7 @@ end
   Base.cumulative_compile_timing(true)
 end
 
-println("Making test list")
+@everywhere include("init_tests.jl")
```
antonydellavecchia, line 107: would it be possible to put this into Serialization/runtests .jl  ?

antonydellavecchia: I think it makes sense to put this line in Serialization/runtests.jl and include the init_tests.jl in that folder as well. That is, if it still works? otherwise I would put at least put the init_tests.jl in the serialization folder

antonydellavecchia (approved):
Looks good, I would just like to see if we can put the init file one level deeper and then merge

test/runtests.jl
lgoettgens: Putting it into `serialization/runtests.jl` unfortunately does not work, as that is already its own parallel job (that might be on a different worker than the FTheoryTools test). Moving the file into the folder is no problem.

benlorenz:
An alternative would be to add a conditional include to the relevant test files:
```julia
isdefined(Main, :test_save_load_roundtrip) || include(joinpath(Oscar.oscardir,"test","Serialization","init_tests.jl"))
```
This would allow running the tests separately from the main runtests without manually including that file.

lgoettgens:
> An alternative would be to add a conditional include to the relevant test files:
> 
> ```julia
> isdefined(Main, :test_save_load_roundtrip) || include(joinpath(Oscar.oscardir,"test","Serialization","init_tests.jl"))
> ```
> 
> This would allow running the tests separately from the main runtests without manually including that file.

That seems like a better idea. I will change this pr accordingly.

benlorenz (approved):

lkastner (approved):

PR #2802, opened 2023-09-14
Reduce the number of global objects during tests

lgoettgens:
by wrapping them into new scopes.

### DIFF
--- a/test/Groups/conjugation.jl
+++ b/test/Groups/conjugation.jl
@@ -116,41 +116,43 @@
 
 end
 
-function TestConjCentr(G,x)
-   Cx = centralizer(G,x)[1]
-   cc = conjugacy_class(G,x)
-   @test index(G,Cx)==length(cc)
-   T=right_transversal(G,Cx)
-   @testset for y in cc
-       @test count(t -> y==x^t, T) == 1
+let
+   function TestConjCentr(G,x)
+      Cx = centralizer(G,x)[1]
+      cc = conjugacy_class(G,x)
+      @test index(G,Cx)==length(cc)
+      T=right_transversal(G,Cx)
+      @testset for y in cc
+         @test count(t -> y==x^t, T) == 1
+      end
+      
+      cs = conjugacy_class(G,Cx)
+      Nx = normalizer(G,Cx)[1]
+      @test index(G,Nx)==length(cs)
+      T=right_transversal(G,Nx)
+      # Set([Cx^t for t in T]) == Set(collect(cs)) does not work
+      @testset for H in collect(cs)
+         @test sum([H==Cx^t for t in T])==1
+      end
    end
-   
-   cs = conjugacy_class(G,Cx)
-   Nx = normalizer(G,Cx)[1]
-   @test index(G,Nx)==length(cs)
-   T=right_transversal(G,Nx)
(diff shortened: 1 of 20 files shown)


test/Groups/conjugation.jl
```diff
@@ -116,41 +116,43 @@
...
 end
 
-function TestConjCentr(G,x)
-   Cx = centralizer(G,x)[1]
-   cc = conjugacy_class(G,x)
-   @test index(G,Cx)==length(cc)
-   T=right_transversal(G,Cx)
-   @testset for y in cc
-       @test count(t -> y==x^t, T) == 1
+let
```
fingolfin, line 119: Is this really necessary? I thought `@testset` starts a new scope and so these functions stay "local" inside that scope anyway. A quick test seems to confirm that:
```
julia> using Test

julia> @testset "bla" begin
         foo(x) = x + 1
       end
Test Summary: |Time
bla           | None  0.0s
Test.DefaultTestSet("bla", Any[], 0, false, false, true, 1.695032056975692e9, 1.695032056992175e9, false)

julia> foo(1)
ERROR: UndefVarError: `foo` not defined
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

test/Modules/FreeModules-graded.jl
```diff
@@ -1,326 +1,328 @@
...
-    f = MPolyBuildCtx(R)
-    for z = 1:3
-      e = rand(0:2, ngens(R))
-      r = base_ring(R)(rand(0:22))
-      push_term!(f, r, e)
-    end
-    g = finish(f)
-    if !iszero(g)
-      push!(polys, g)
+let
```
fingolfin, line 1: Following what I wrote above, I'd just move the start of the testset up:
```suggestion
@testset "Modules" begin
```

(Plus of course the matching removal of line ~48)

test/NumberTheory/nmbthy.jl
```diff
@@ -1,15 +1,18 @@
...
-  return x.unit * prod(p*k for (p,k) = x.fac)
-end
-@testset "Polymake.factorisations" begin
-  k, a = quadratic_field(-5)
-  zk = maximal_order(k)
-  f = factorisations(zk(6))
-  @test length(f) == 2
-  @test all(x -> evalu(x) == 6, f)
+let
+  function evalu(x::Fac)
```
fingolfin, line 5: Again, just move it into the `@testset`?

test/PolyhedralGeometry/cone.jl
```diff
@@ -1,178 +1,179 @@
 const pm = Polymake
+let
+    NF, sr2 = quadratic_field(2)
+    Qx, x = QQ["x"]
+    K, (a1, a2) = embedded_number_field([x^2 - 2, x^3 - 5], [(0, 2), (0, 2)])
```
fingolfin, line 5: I think this should all just be inside a top level `@testset` (probably including the definition of `pm`.

As a variant, it could also be a `@testset ... for` like so (I would need to but can't modify the next couple lines, too, but I hope the idea is clear):

```suggestion
@testset "Cone{$(elem_type(f))} over $f" for f, _ in [
        quadratic_field(2),
        QQ["x"],
        embedded_number_field([x^2 - 2, x^3 - 5], [(0, 2), (0, 2)]),
        ]
```

PR #2805, opened 2023-09-15
fixup #2792: enable test file

benlorenz:
cc: @thofma 
Maybe codecov was not wrong after all.

And maybe we should switch to either including all files using readdir or add some check to make sure all test files are really included.


thofma (approved):

thofma:
sorry codecov

PR #2806, opened 2023-09-16
Sb/print local rings

simonbrandhorst:


experimental/Schemes/BlowupMorphism.jl
```diff
@@ -37,14 +37,14 @@ ideal(x, y, z)
...
   in sheaf of ideals with restriction
     1b: ideal(x, y, z)
 with domain
   scheme over QQ covered with 3 patches
-    1a: [(s1//s0), (s2//s0), x]   spec of quotient of multivariate polynomial ring
-    2a: [(s0//s1), (s2//s1), y]   spec of quotient of multivariate polynomial ring
-    3a: [(s0//s2), (s1//s2), z]   spec of quotient of multivariate polynomial ring
+    1a: [(s1//s0), (s2//s0), x]   V(0, 0, 0)
+    2a: [(s0//s1), (s2//s1), y]   V(0, 0, 0)
+    3a: [(s0//s2), (s1//s2), z]   V(0, 0, 0)
```
afkafkafk13, line 47: That looks a bit weird...

Did you want to say V(0) with variables [(s1/.......]? This is hard to guess from the output

experimental/Schemes/MorphismFromRationalFunctions.jl
```diff
@@ -50,8 +50,8 @@ julia> realizations = Oscar.realize_on_patch(Phi, U);
 
 julia> realizations[3]
 Morphism
-  from [(t//s)]          spec of localized ring
-  to   [(x//z), (y//z)]  spec of multivariate polynomial ring
+  from [(t//s)]          AA^1 \ V()
+  to   [(x//z), (y//z)]  affine 2-space
```
afkafkafk13, line 54: This also looks strange. Did you want to say V(0)?

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
```diff
@@ -241,20 +245,21 @@ julia> (x1,x2,x3) = gens(R)
...
+  from [x1, x2, x3]  V(x1)
   to   [x1, x2, x3]  affine 3-space over QQ
 given by the pullback function
   x1 -> 0
   x2 -> x2
   x3 -> x3
 
 julia> graph(f)
-(Spec of quotient of multivariate polynomial ring, Morphism: spec of quotient of multivariate polynomial ring -> spec of quotient of multivariate polynomial ring, Morphism: spec of quotient of multivariate polynomial ring -> affine 3-space over QQ with coordinates [x1, x2, x3])
+(V(x1, -x1, x2 - x2, x3 - x3), Morphism: V(x1, -x1, x2 - x2, x3 - x3) -> V(x1), Morphism: V(x1, -x1, x2 - x2, x3 - x3) -> affine 3-space over QQ with coordinates [x1, x2, x3])
```
afkafkafk13, line 262: Something is going wrong with the variable names here. This keeps the new output from being more helpful than the old one. 
(I am aware that it is a non-trivial design choice, what to do in the case of products with the same variable names.)

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -535,14 +551,17 @@ Complement
...
-    of quotient of multivariate polynomial ring by ideal with 1 generator
+    of quotient
+      of multivariate polynomial ring in 2 variables x, y
+        over rational field
+      by ideal(x^2 - 2*x*y + y^2)
     at complement of maximal ideal of point (0, 0)
 
 julia> reduced_scheme(X)
-(Spec of quotient of multivariate polynomial ring, Morphism: spec of quotient of multivariate polynomial ring -> spec of quotient of multivariate polynomial ring)
+(V(x^2 - 2*x*y + y^2, x - y), Morphism: V(x^2 - 2*x*y + y^2, x - y) -> V(x^2 - 2*x*y + y^2))
```
afkafkafk13, line 561: Not a comment to the printing, but to its content: 
This reveals redundant generators to the user -- is this intentional (both the redundant generators and revealing them)?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -617,14 +636,15 @@ Spectrum
...
 Spectrum
   of quotient
-    of multivariate polynomial ring in 3 variables over QQ
+    of multivariate polynomial ring in 3 variables x, y, z
+      over rational field
     by ideal(x^2 - y^2 + z^2)
 
 julia> singular_locus(A3)
-(Spec of quotient of multivariate polynomial ring, Morphism: spec of quotient of multivariate polynomial ring -> spec of multivariate polynomial ring)
+(V(1), Morphism: V(1) -> affine 3-space)
```
afkafkafk13, line 644: Again just a remark on the design decision, certainly not a change request: 
`empty set` would be more user friendlc than V(1) -- but has the same mathematical content. But on the other hand large insatisfiable systems of equations might hide an empty set, whereas computing that it is the empty set may be expensive. One way out might be to check whether the set itself knows that it is the empty set and only then print èmpty set` and print the generating set otherwise.

afkafkafk13 (commented):
New printing looks very good: right balance between necessary information and sufficiently compact presentation. Thank you.

I left a few comments on particular cases of output, which struck me as odd or weird at first glance or for which I would have expected something else. These are to be understood as markers for further discussion, not as complaints against this PR.

In particular, we should discuss at some point whether the affine space encoded as Spec(k[x]) and the affine space encoded as Spec(k[x]/<0>) or as V() should be printed in the same or different ways. Same question for the empty set and V(1) etc.

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: Yes, I think that's

experimental/Schemes/MorphismFromRationalFunctions.jl
simonbrandhorst: The defining ideal has no generators

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
simonbrandhorst: One could choose new names for the codomain in case of an automorphism. Say `x,y,t` and `x',y',t'`

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
simonbrandhorst: If there are redundant generators, this fact should not be hidden.

experimental/Schemes/BlowupMorphism.jl
afkafkafk13: ok. I can live with that, but we might need to investigate the reason for having several zeros in there at some point.

experimental/Schemes/MorphismFromRationalFunctions.jl
afkafkafk13: If we want to stick to mathematical notation, we would need a zero here or omit the  \V() completely in the case of no generators of the ideal.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
afkafkafk13: Yes this would be a viable solution for this case, but what about morphisms between rings which only happen to have the same variable names by chance? We should discuss this point directly (not on github).

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
afkafkafk13: ok. This is a clear policy and I can fully support it. We should, however, investigate, why `reduced_scheme` returns something with redundant generators.

experimental/Schemes/MorphismFromRationalFunctions.jl
simonbrandhorst: `V({})` is mathematically sound too. Here it is shortened to `V()`.  ... I am a bit reluctant to add the zero. 
Because we do `ideal()` too and not `ideal(0)`. I see it in the same vein.

afkafkafk13 (approved):

experimental/Schemes/MorphismFromRationalFunctions.jl
afkafkafk13: ok good point.

PR #2807, opened 2023-09-16
twisted modules

wdecker:


src/Modules/ModulesGraded.jl
```diff
@@ -2567,3 +2567,89 @@ function ideal_as_module(I::MPolyIdeal)
...
+    twist(M::ModuleFP{T}, d::IntegerUnion) where {T<:MPolyDecRingElem}
+
+Given a module `M` over a $\mathbb Z$-graded polynomial ring and an integer `d`, 
+convert `d` into an element `g` of the grading group of the ring and proceed as above.
+
+# Examples
+```jldoctest
+julia> R, (x, y) = graded_polynomial_ring(QQ, ["x", "y"]);
+
+julia> I =  ideal(R, [zero(R)])
```
lgoettgens:
```suggestion
julia> I = ideal(R, [zero(R)])
```

src/Modules/ModulesGraded.jl
```diff
@@ -2567,3 +2567,89 @@ function ideal_as_module(I::MPolyIdeal)
...
+
+```
+"""
+function twist(M::ModuleFP{T}, g::GrpAbFinGenElem) where {T<:MPolyDecRingElem}
+ error("Not implemented for the given type")
+end
+
+function twist(M::SubquoModule{T}, g::GrpAbFinGenElem) where {T<:MPolyDecRingElem}
+ R = base_ring(M)
+ @req parent(g) == grading_group(R) "Group element not contained in grading group of base ring"
```
lgoettgens, line 2623: Is `==` here enough or do we need `===`? (same at other places)

test/Modules/ModulesGraded.jl
```diff
@@ -1166,3 +1166,13 @@ end
   b = transpose(a) * a
   @test is_symmetric(b)
 end
+
+@testset "twist" begin
+  R, (x, y) = graded_polynomial_ring(QQ, ["x", "y"], [1 0; 0 1]);
+  I = ideal(R, [x, y])
+  M = quotient_ring_as_module(I)
+  N = twist(M, [1; 2])
```
lgoettgens:
```suggestion
  N = twist(M, [1, 2])
```
I think you want a vector here instead

src/Modules/ModulesGraded.jl
wdecker: @fieker Please clearify!

afkafkafk13: Sideremark: 
If `===` is needed, this problem is not restricted to this PR. The problem then also occurs in `Modules/ModulesGraded.jl` (and possibly other files) as the same kind of comparison also occurs there with `==`.

fingolfin: I think `==` and `===` are the same for rings, so either is fine.

jankoboehm:
Thanks for spotting, it should be ```===```, and it should also be changed in ```set_grading!```, ```graded_free_module```, and ```grade```, every time the same copied line.

src/Modules/ModulesGraded.jl
```diff
@@ -2567,3 +2567,89 @@ function ideal_as_module(I::MPolyIdeal)
...
+1 -> e[1]
+represented as subquotient with no relations
+
+julia> degree(gen(N, 1))
+[-2]
+
+```
+"""
+function twist(M::ModuleFP{T}, g::GrpAbFinGenElem) where {T<:MPolyDecRingElem}
+ error("Not implemented for the given type")
```
fingolfin, line 2618: I am not going to block this PR over this, but, normally we strive to indent code with (at least) two spaces.

(But of course we are already widely inconsistent about this, which is why I am not asking for it to be changed here and now, just wanted to point it out)

src/Modules/ModulesGraded.jl
wdecker: As confirmed by @fieker, it is as I thought: There is no difference between `==` and `===` here. That is, the PR can be merged and no other action is needed.

afkafkafk13 (approved):

PR #2808, opened 2023-09-18
all_small_groups: allow filtering by 'exponent'

fingolfin:


src/Groups/libraries/smallgroups.jl
```diff
@@ -176,6 +176,7 @@ The following functions are currently supported as values for `func`:
 - `is_solvable`
 - `is_supersolvable`
 - `number_conjugacy_classes`
+- `exponent`
```
ThomasBreuer: Up to now, the list was ordered alphabetically. What is the reason to put `exponent` here?

ThomasBreuer (approved):
Good.
(I have left one minor comment.)

src/Groups/libraries/smallgroups.jl
fingolfin: Just didn't think about it. Moved it up now, thanks!

Issue #2691, opened 2023-08-17
Randomized tests failing for `AlgClosure{fqPolyRepField}`


lgoettgens:
Starting with #2684, I noticed some seemingly random fails on the master CI with different julia versions.
All are pretty lengthy, but contain something like the following multiple times:
```julia
Basic functions: Error During Test at /home/runner/.julia/packages/AbstractAlgebra/lDlYU/test/Rings-conformance-tests.jl:98
  Test threw exception
  Expression: equality(a - c, a + -c)
  MethodError: objects of type Nothing are not callable
  Stacktrace:
   [1] (::fqPolyRepField)(b::fqPolyRepFieldElem)
     @ Nemo ~/.julia/packages/Nemo/Tj33G/src/flint/fq_nmod.jl:660
   [2] op(f::typeof(-), a::AlgClosureElem{fqPolyRepField}, b::AlgClosureElem{fqPolyRepField})
     @ Oscar.AlgClosureFp ~/work/Oscar.jl/Oscar.jl/src/Rings/AlgClosureFp.jl:117
   [3] -(a::AlgClosureElem{fqPolyRepField}, b::AlgClosureElem{fqPolyRepField})
     @ Oscar.AlgClosureFp ~/work/Oscar.jl/Oscar.jl/src/Rings/AlgClosureFp.jl:126
   [4] macro expansion
     @ ~/.julia/packages/AbstractAlgebra/lDlYU/test/Rings-conformance-tests.jl:98 [inlined]
   [5] macro expansion
     @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
   [6] macro expansion
     @ ~/.julia/packages/AbstractAlgebra/lDlYU/test/Rings-conformance-tests.jl:70 [inlined]
   [7] macro expansion
     @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Test/src/Test.jl:1151 [inlined]
   [8] test_NCRing_interface(R::AlgClosure{fqPolyRepField}; reps::Int64)
     @ Main ~/.julia/packages/AbstractAlgebra/lDlYU/test/Rings-conformance-tests.jl:31
```
[...]

PR #2809, opened 2023-09-18
Call Oscar.randseed! at start of test suite

fingolfin:
... to make test runs *slightly* easier to reproduce.

However, this is not enough. We also need to reset the RNGs at the start of `@testset` like is done for the global RNG. Indeed, `Test.@testset` does this at the start:
```julia
        local RNG = default_rng()
        local oldrng = copy(RNG)
        local oldseed = Random.GLOBAL_SEED
        try
            # RNG is re-seeded with its own seed to ease reproduce a failed test
            Random.seed!(Random.GLOBAL_SEED)
```
and then this at the end of each testset:
```julia
        finally
            copy!(RNG, oldrng)
            Random.set_global_seed!(oldseed)
```

Ideally something similar would be done for `Oscar.randseed!` -- IMHO it would be fully sufficient to do `Oscar.randseed!(SOME_SESSION_FIXED_VALUE)` at the start of each testset.

There are various hacky ways to achieve that (e.g. we could monkey patch various methods), but a clean solution would be to implement a custom `AbstractTestSet` subtype which implements the desired behavior....

CC @benlorenz

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -36,6 +36,7 @@ end
 @everywhere using Test
 @everywhere using Oscar
 @everywhere Oscar.set_seed!($seed)
+@everywhere Oscar.randseed!($seed)
 
 # hotfix, otherwise StraightLinePrograms returns something which then leads to an error
 module SLPTest


fingolfin:
It seems with this PR, the error from issue #2691 related to errors in `AlgClosureElem{fqPolyRepField}` handling *always* happens in Julia 1.6... at least I've now re-run that job (after rebasing etc.) multiple times and it always fails.

Can we get a better workaround there, I wonder? @fieker ?

PR #2812, opened 2023-09-18
Resolve some JET warnings

fingolfin:
... about captured variables and runtime dispatch

Just the tip of the iceberg, but I figured I should already submit this, as I don't know when/if I'll be able to work through all the output produced by JET.... 

@fieker this touches some of your code, too

### DIFF
--- a/src/Groups/matrices/FiniteFormOrthogonalGroup.jl
+++ b/src/Groups/matrices/FiniteFormOrthogonalGroup.jl
@@ -25,7 +25,6 @@
 function _mod_p_to_a_kernel(G::Union{zzModMatrix, ZZModMatrix}, a, p)
   n = ncols(G)
   R = base_ring(G)
-  E = identity_matrix(R, n)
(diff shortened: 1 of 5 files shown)


src/Groups/matrices/FiniteFormOrthogonalGroup.jl
```diff
@@ -57,15 +56,15 @@ function _mod_p_to_a_kernel(G::Union{zzModMatrix, ZZModMatrix}, a, p)
       end
     end
     for h in gensk
-      g = deepcopy(E)
+      g = identity_matrix(R, n)
```
thofma, line 59: Whats the problem with this one?

fingolfin: In at least one of the functions, it redefined `E` to different things, and these objects are then captured in closures; it is not happy that `E` is redefined then. It seemed simplest to just replace all those `deepcopy` calls by direct calls to `identity_matrix` (IMHO makes it easier to read).

PR #2813, opened 2023-09-18
Reorganize tests to always use runtests.jl files

fingolfin:
... to group things together. Also merge `test/Experimental` content into the `experimental` directory.

This should be mostly orthogonal to PR #2810 by @benlorenz


benlorenz (approved):
thanks, looks good!

test/runtests.jl
```diff
@@ -105,51 +105,34 @@ end
...
+  "Serialization/polymake/runtests.jl",
+  "Serialization/upgrades/runtests.jl",
 
   "Data/runtests.jl",
+  "StraightLinePrograms/runtests.jl",
 
-  "StraightLinePrograms/runtests.jl"
+  # Automatically include tests of all experimental packages following our
+  # guidelines.
+  "../experimental/runtests.jl",
```
lgoettgens, line 132: For the parallelization, it would make sense to not include `experimental/runtests.jl` as the last item, as that file contains a lot of tests that in total need a long time. And parallelization with `pmap` gets more efficient if the larger tasks are at the front and the smaller ones in the end of the list.

test/Rings/runtests.jl
```diff
@@ -1,3 +1,6 @@
+using Oscar
+using Test
```
lgoettgens, line 2: If we are `using Test` in basically every testfile, we do not check that it is already loaded in
https://github.com/oscar-system/Oscar.jl/blob/81b983d1df7ada2aca3641bae806f01722219799/src/utils/tests.jl#L23

test/runtests.jl
fingolfin: Sure, I can move it further up again.

That said, perhaps we should revise how we do this parallelization anyway: it would make more sense to parallelize *all* those include statements (under the assumption that they are all mutually independent, which they really should be).

For this, we could just take all the `*.jl` files under `test` which are not named `runtests.jl`, and then pass all of them to `pmap`. Then the `runtests.jl` files would just be for convenience (and they could all also become identical, with the same code that includes all "*.jl" files in "subdirectories of the current directory except for those named `runtests.jl`).

That would then also render the part of PR #2810 moot which checks for files that we accidentally did not include...

test/Rings/runtests.jl
fingolfin: But we currently don't do it in ever testfile, just in the `runtests.jl` files...

test/runtests.jl
benlorenz: When I prepared #2810 I was considering including everything automatically as well. But I thought it would be easier / safer to just check the includes first.
Luckily there weren't that many missing includes. Automatically including all `jl` files in the test folders would be fine with me as well.

Issue #2672, opened 2023-08-14
Add `SubPcGroup` type


fingolfin:
... as discussed with @ThomasBreuer and @fieker . The plan is that `PcGroup` type is reserved for cases where the generators returned by `gens` form a pc sequence ("pcgs" in GAP lingo).

fingolfin:
Some more notes:
1. the name `SubPcGroup` is preliminary
1. the documentation should spell out the relation between `PcGroup` and `SubPcGroup`, and also point out that one can convert to the former "as usual" (i.e. `PcGroup(G)` or `isomorphism(PcGroup, G)`
1. we also need a `SubPcGroupElem` type (very easy)
1. something similar should be done for `FpGroup` as well, i.e. adding a `SubFpGroup` type, see also #1836
1. given a `SubPcGroup` there should be a way to get the group it lives in; in GAP lingo that would be its "parent" but perhaps that'd be confusing given how `parent` is used so far in OSCAR? So here are some alternative names
   1. `full_group`
   1. `whole_group`
   1. `ambient_group`
1. this new type actually also simplifies some semantical issues we had with serialization support: if one serializes a vector of several `SubPcGroup` instances which all have the same "full group", then this property can now easily be preserved, because (de)serialization of a `SubPcGroup` would (de)serialize its "full group"
1. we must then adjust the code in `_get_type` resp. the vector `_gap_group_types`: currently it simply maps `GAP.Globals.IsPcGroup` to `PcGroup` which of course then isn't enough anymore.

ThomasBreuer:
Some more aspects to be addressed by this change:
- `PcGroup` will then mean a group that knows/stores its polycyclic presentation, whereas a `SubPcGroup` group will initially not know such a presentation.
[...]

PR #2816, opened 2023-09-18
Groups: add minimal_generating_set

fingolfin:


src/Groups/group_constructors.jl
```diff
@@ -475,20 +475,26 @@ where the `i`-th generator is printed as `L[i]`.
 function free_group(n::Int, s::VarName = :f; eltype::Symbol = :letter)
    @req n >= 0 "n must be a non-negative integer"
    if eltype == :syllable
-     return FPGroup(GAP.Globals.FreeGroup(n, GAP.GapObj(s); FreeGroupFamilyType = GapObj("syllable"))::GapObj)
+     G = FPGroup(GAP.Globals.FreeGroup(n, GAP.GapObj(s); FreeGroupFamilyType = GapObj("syllable"))::GapObj)
    else
-     return FPGroup(GAP.Globals.FreeGroup(n, GAP.GapObj(s))::GapObj)
+     G = FPGroup(GAP.Globals.FreeGroup(n, GAP.GapObj(s))::GapObj)
    end
+   GAP.Globals.SetRankOfFreeGroup(G.X, n)
```
ThomasBreuer: I see: `RankOfFreeGroup` belongs to GAP's FGA package, therefore the function `FreeGroup` cannot set this attribute. Thus the situation here is better.

src/Groups/GAPGroups.jl
```diff
@@ -1448,6 +1499,29 @@ false
 @gapattribute is_finitelygenerated(G::GAPGroup) = GAP.Globals.IsFinitelyGeneratedGroup(G.X)::Bool
 
 
+@doc raw"""
+    is_free(G::GAPGroup)
+
+Return whether `G` is a free group.
```
ThomasBreuer: We should say what is meant with "is a free group".
GAP's `IsFreeGroup` means essentially a group created with `free_group` or a subgroup of such a group.
The GAP documentation is also vague about this, but it states that groups consisting of elements in the filter `IsAssocWordWithInverse` are regarded as free groups.

One could have the idea that "free group" means a group that is free on some set of elements, and then each trivial group is free on the empty set, and certain factors of free groups (for example where just some generators are mapped to the identity) are also free.

src/Groups/GAPGroups.jl
```diff
@@ -1424,9 +1452,32 @@ function set_prime_of_pgroup(G::GAPGroup, p::IntegerUnion)
   set__prime_of_pgroup(G, GAP.Obj(p))
 end
 
+"""
+    rank(G::GAPGroup)
+
+Return the rank of the group `G`, i.e., the minimal size of a generating set.
```
ThomasBreuer: `GrpAbFinGen` defines `rank` as the rank of the free part. Thus torsion groups created with `abelian_group` have always rank zero. If we really want to define `rank` for `GAPGroup` as the cardinality of a minimal generating set then perhaps we should mention this.

ThomasBreuer (commented):
Thanks.

The definition of `rank` does not agree with `rank` for `GrpAbFinGen`.
Perhaps the definition of `is_free` should be more explicit.
(See my comments.)

src/Groups/group_constructors.jl
fingolfin: Aye.

Regarding the definition of `rank(::GrpAbFinGen)` differing (it returns the Z-rank, or the "Hirsch length" to use polycyclic group terminology): ahh, that's a shame. Dang. Well, I could rename this function `_rank` for now, so that I can use it in the printing code (that was what motivated this PR), and then possibly again restrict it to support "free fp groups" only. And then worry about how to expose this to users later.

src/Groups/GAPGroups.jl
fingolfin: Thank you for this important observation.

Adding to it, also any infinite cyclic group (e.g. represented as a pcp group) would be "free". And yeah, in GAP we have this mathematically surprising situation then:
```
gap> G:=AbelianGroup([0]);
<fp group of size infinity on the generators [ f1 ]>
gap> IsFreeGroup(G);
false
gap> IsFreeGroup(FreeGroup(1));
true
```

So this is a case were one needs to distinguish between "mathematically free" (on some unknown set) and "technically free" (created as a free group), for lack of better terminology.

Both are of independent interest.... But I think a function `is_free` for groups should aim to check for the mathematical property, though of course it cannot in general (as long as we can't generally decide whether a fp group is trivial, we also can't decide freeness). But we can do it in many other cases, e.g. for finite groups (where `is_free(G) = is_trivial(G)`), pcp groups (either trivial or infinite cyclic, both can be tested for efficiently). For fp groups, we could make a best effort: compute abelian invariants; if any are non-zero, return `false`; otherwise we can read of the potential rank of the "free" group (if it is free); then use `TzGoGo` to try and simply (a copy of) the presentation to see if it is "obviously" free (= no relations left).


But of course what I really am using is the "technical interpretation". I actually thought already that perhaps we should split the `FPGroup` type not just into `FPGroup` and `SubFPGroup` (as we already discussed at PR #2672), but in fact also have `FreeGroup` (and perhaps also `SubFreeGroup`). In that case the "technical" check becomes a type check.

ThomasBreuer: Yes, I think that the "technical interpretation" is used more frequently; also GAP's `IsFreeGroup` is just a filter, there was no aim to support a check for the mathematical property.

So do we need a better name for the "technical interpretation", in order to keep `is_free` for the mathematical one,
or do we decide that we introduce the type `FreeGroup` as mentioned above?

(I guess the name `IsFreeGroup` was chosen because people did not think that checking the mathematical property for arbitrary f. p. groups would be interesting. Moreover, as soon as such a function would be available, somebody would come and ask for a function that checks whether a group is free on a given set, and for a function that returns a set on which a given group is free ...)

fingolfin:
In order to move forward, I have for now disabled the `rank` and `is_free` methods, to be discussed later. But `minimal_generating_set` is kept.

src/Groups/GAPGroups.jl
```diff
@@ -1424,9 +1452,36 @@ function set_prime_of_pgroup(G::GAPGroup, p::IntegerUnion)
   set__prime_of_pgroup(G, GAP.Obj(p))
 end
 
-
-"""
-    is_finitelygenerated(G)
+# TODO/FIXME: the rank method below is disabled because it conflicts
+# with semantics of  the `rank` method for GrpAbGen. We'll have
```
joschmitt:
```suggestion
# with semantics of  the `rank` method for GrpAbFinGen. We'll have
```

fingolfin: thanks

ThomasBreuer (approved):

PR #2817, opened 2023-09-19
removes more singular_assure calls

ederc:
Removes `singular_assure` calls in `mpoly-ideals.jl`.

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -1258,8 +1258,6 @@ julia> dim(I)
   if I.dim > -1
     return I.dim
   end
-  @show I.gb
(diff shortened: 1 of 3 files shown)


src/Rings/mpoly-ideals.jl
```diff
@@ -1286,9 +1258,9 @@ julia> dim(I)
   if I.dim > -1
     return I.dim
   end
-  G = groebner_assure(I, false, true)
-  singular_assure(G)
-  I.dim = Singular.dimension(G.S)
+  @show I.gb
+  @show singular_groebner_generators(I, false, true)
```
joschmitt:
```suggestion
```

fingolfin (approved):

PR #2753, opened 2023-08-31
LieAlgebras: Constructions for module homs

lgoettgens:
~~This builds upon https://github.com/oscar-system/Oscar.jl/pull/2644 and thus includes that PR. Once https://github.com/oscar-system/Oscar.jl/pull/2644 gets merged, I will rebase this PR.~~

Introduces new constructions for Lie algebra module homs:

- [x] injections/projections for direct sums
- [x] `hom_tensor`
- [x] `hom_direct_sum`
- [x] lift hom $V \to W$ to $T^k V \to T^k W$
- [x] lift hom $V \to W$ to $\bigwedge^k V \to \bigwedge^k W$
- [x] lift hom $V \to W$ to $S^k V \to S^k W$

While doing that, refactor the constructions of the modules and put more helpful stuff in attribute storage:

- [x] direct sum
- [x] tensor product
- [x] tensor power
- [x] exterior power
- [x] symmetric power

More changes:
- [x] add `is_welldefined` to be able to check that even if using `check=false` in the `hom` constructor
- [x] add `zero_map`
- [x] change the internals of modules to use right multiplication with a matrix as the action

Missing: 
- [ ] Handle 0-th powers

PR #2818, opened 2023-09-19
LieAlgebras: Follow-up to #2753

lgoettgens:
This includes some small adaptions and fixes that I missed in #2753.

### DIFF
--- a/experimental/LieAlgebras/src/LinearLieAlgebra.jl
+++ b/experimental/LieAlgebras/src/LinearLieAlgebra.jl
@@ -140,21 +140,17 @@ end
 ###############################################################################
 
 @doc raw"""
-    (L::LinearLieAlgebra{C})(m::MatElem{C}) -> LieAlgebraElem{C}
+    coerce_to_lie_algebra_elem(L::LinearLieAlgebra{C}, x::MatElem{C}) -> LinearLieAlgebraElem{C}
 
-Return the Lie algebra element whose matrix representation corresponds to `m`.
-This requires `m` to be a square matrix of size `n > 1` (the dimension of `L`), and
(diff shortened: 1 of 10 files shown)


experimental/LieAlgebras/src/LinearLieAlgebra.jl
```diff
@@ -140,21 +140,17 @@ end
 ###############################################################################
 
 @doc raw"""
-    (L::LinearLieAlgebra{C})(m::MatElem{C}) -> LieAlgebraElem{C}
+    coerce_to_lie_algebra_elem(L::LinearLieAlgebra{C}, x::MatElem{C}) -> LinearLieAlgebraElem{C}
```
fingolfin, line 143: Hmm, so `L(m)` is changed to `coerce_to_lie_algebra_elem(L,m)` -- why?

lgoettgens: Because `L(m)` is already in use for `m` being a coefficient vector (of size `1 x dim(L)`). For dimensions 1 and 0, it is ambiguous which of the two you mean

thofma: Why are coefficient vectors matrices and not vectors?

lgoettgens: That's because I started with `AbstractAlgebra`'s `Generic.FreeModuleElem` and adapted it to my needs.

I should probably change it to `SRow{C}` and change module action matrices and home matrices to sparse matrices as well. But as that is a larger change, I would like to split that into a new PR.
`coerce_to_lie_algebra_elem(L, m)` will then be removed again in favor of `L(m)`.

fingolfin (approved):

PR #2824, opened 2023-09-19
Revise some character table tests to use a doctest

fingolfin:
These tests deal with testing how character tables are printed in various situations. We change this to (ab)use the `jldoctest` system from Documenter. This brings with it the advantage that it is very easy to adjust the tests for desired changes to the printing: it can be done changing

    doctest(nothing, [AuxDocTest_GroupCharacters])

to

    doctest(nothing, [AuxDocTest_GroupCharacters]; fix=true)

and then running this in the Julia REPL:

    include("test/Groups/group_characters.jl")

This PR is motived by PR #2774 where it helps me to adjust the tests for the revised printing. But I think it's also of independent benefit.

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -1,6 +1,7 @@
 [deps]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
@@ -8,3 +9,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
 Aqua = "0.7"
+Documenter = "0.27, 1.0"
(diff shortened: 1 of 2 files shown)


ThomasBreuer (approved):
O.k.

Thus the message is:
Whenever we have tests concerning the printed format of objects then we should use `jldoctest` instead of `@testset` & `@test`.

(I have another testfile that has to be revised in the same spirit.)

test/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
lgoettgens, line 4: please add a compat bound for Documenter. Preferably `0.27` for now, and we can adapt that when we bump it for the docs

fingolfin: I don't think 1.0 behaves differently here.

fingolfin: done

fingolfin:
@ThomasBreuer exactly

fingolfin:
@ThomasBreuer wrote:

> I have another testfile that has to be revised in the same spirit.

Which one is that? I am aware of `test/Groups/MatrixDisplay.jl` also in need of such changes, is that what you had in mind. If you plan to work on that this week, let me know, then I'll wait for it -- I'd rather not duplicate efforts.

ThomasBreuer:
> > I have another testfile that has to be revised in the same spirit.
> 
> Which one is that?

`experimental/OrthogonalDiscriminants/test/data.jl`. I have made the changes already, and will create a pull request for that; at the same time, I will deal with `test/Groups/MatrixDisplay.jl`.

PR #2826, opened 2023-09-19
Fix CI breakage

fingolfin:
... caused by two PRs which each was fine by its own but together they broke


lgoettgens (approved):
This looks indeed weird. If CI succeeds, this change is fine with me.

fingolfin:
Just to explain what's going on: `with_unicode()` used to (accidentally) return the return value of the final `allow_unicode` call in it, which always was `true`. As a result, the doctests introduced in PR #2824 (which invoke `with_unicode`) printed those `true` values without me noticing. But in PR #2822 I refactored `with_unicode` and as an (unintended but IMHO desirable) side effect it now returns nothing. Boom

PR #2827, opened 2023-09-19
Explicitly load various GAP package

fingolfin:
... that we use actively. Future GAP.jl versions will load fewer packages automatically


ThomasBreuer (approved):
Yes, we should make explicit which GAP packages get loaded.

I have thought also about GAP's `-r` option:
Currently we call GAP without this option, thus some (newer) packages stored in the user's `.gap` directory may have the effect that silently more packages get loaded, and perhaps in a different order. This may affect the reproducibility of test results.
On the other hand, having the private GAP root directory is useful when one actually works with GAP in Julia.

PR #2836, opened 2023-09-21
Free resolutions updates

ederc:
1. Adds an attribute `algorithm` to the underlying `ChainComplex` in order to correctly extend a partially computed free resolution.
2. Introduces `algorithm` options `mres` and `nres` for extension code.
3. Removes error message if free resolution is already complete and returns zero module instead.
4. Fixes return of zero module if index is higher than the length of the free resolution.


wdecker (approved):

afkafkafk13:
Please adjust doctest output.

PR #2839, opened 2023-09-21
Add heisenberg_group constructor

fingolfin:
Maybe useful for @wdecker and might make @fieker happier?


wdecker:
Is this the Heisenberg group of upper triangular matrices? This is completely different from the Heisenberg groups I consider.

ThomasBreuer (approved):
Looks good.
Eventually there should be an entry for `heisenberg_group` in the documentation, but I do not see an obvious place for it.
(`cyclic_group` and `dihedral_group` are documented in the section about pc groups. Thus here it would make sense to have a paragraph in the section about matrix groups.)

fingolfin:
@wdecker It is not clear from what you write in the book chapter which group you really want to consider... But the one you describe in terms of matrices is isomorphic to the group called `heisenberg_group(GF(3))` resp. `heisenberg_group(GF(3), 3)` in this PR. Of course it is in a different *representation*...

In any case I think some more details are needed in your write up: the "classic" Heisenberg group $\mathcal{H}$ is defined over $\mathbb{R}$. Its Schrödinger representation is defined on $L^2(\mathbb{R})$ which of course is infinite dimensional.

Perhaps you are considering a discrete subgroup of $\mathcal{H}$ for which the Schrödinger representation has finite sections. E.g. the "natural" integral points (corresponding to `heisenberg_group(ZZ)` in this PR) obviously admit any `heisenberg_group(GF(p))` as quotients. These groups are extra special groups of exponent $p$ and order $p^3$ and admit certain $p$-dimensional irreducible representations. For $p=3$ we get exactly the representation you are studying (actually, there are $p-1$ such representations, depending on the choice of primitive root...). This construction actually works for any modulus $n$, but they are only irreducible when $n$ is a prime:
```julia
  K, a = CyclotomicField(n, "a")
  m1 = diagonal_matrix([a^i for i in 0:n-1])
  m2 = permutation_matrix(K,[ (i%n)+1 for i in 1:n])
  G = matrix_group(m1, m2)
```
But the point remains, these are representations of a discrete subgroup of $\mathcal{H}$, not of $\mathcal{H}$ itself.

PR #2840, opened 2023-09-21
Adv/serialization session bf

antonydellavecchia:
Fixes some session bugs.

For example loading files that use ids at the top level are properly added to session.
Adds test for testing session functionality


ThomasBreuer (approved):
Thanks, looks good.

PR #2842, opened 2023-09-21
deduce finiteness of the normal subgroup in `quo`

ThomasBreuer:
if finiteness of the whole group was set in the computation on the GAP side
(suggested by @wdecker)

### DIFF
--- a/src/Groups/sub.jl
+++ b/src/Groups/sub.jl
@@ -483,6 +483,10 @@ PermGroup
 """
 function quo(G::T, N::T) where T <: GAPGroup
   mp = GAP.Globals.NaturalHomomorphismByNormalSubgroup(G.X, N.X)::GapObj
+  # The call may have found out new information about `G.X`,
+  # for example that `G.X` is finite.
+#FIXME: The GAP function should deal with this situation.
+  GAP.Globals.UseSubsetRelation(G.X, N.X)
   cod = GAP.Globals.ImagesSource(mp)::GapObj
   S = elem_type(G)
   S1 = _get_type(cod)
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
How shall I deal with the segmentation fault in `doctest (1.9, macOS-latest)`?
(And why are two tests running 20 hours after I had opened the pull request?)

src/Groups/sub.jl
```diff
@@ -483,6 +483,10 @@ PermGroup
 """
 function quo(G::T, N::T) where T <: GAPGroup
   mp = GAP.Globals.NaturalHomomorphismByNormalSubgroup(G.X, N.X)::GapObj
+  # The call may have found out new information about `G.X`,
+  # for example that `G.X` is finite.
+#FIXME: The GAP function should deal with this situation.
```
fingolfin, line 488: What GAP function should deal with what situation? You mean, if anything is learned about a group $G$ then all its "known" subgroups and quotients should be informed about it? This is something the Homalg packages implement for their objects, it is kind of like immediate methods on steroids (with all the pros and cons this implies).

Or maybe just that all "related groups" that are known "and at hand / reachable / already handled in our code" ? As in the case of `NaturalHomomorphismByNormalSubgroup`, but potentially others (e.g. `Normalizer`)?

fingolfin (approved):

src/Groups/sub.jl
ThomasBreuer: Sorry for not being precise enough.
I meant that `GAP.Globals.NaturalHomomorphismByNormalSubgroup` (or rather `GAP.Globals.NaturalHomomorphismByNormalSubgroupNC`) is a better place for the call of `UseSubsetRelation`, and I meant only one call for the pair `(G, N)`.

PR #2845, opened 2023-09-21
Invariant rings with an existing polynomial ring

joschmitt:
Add constructors to allow setting the underlying polynomial ring of an invariant ring (as requested by @wdecker ; I also already had this problem).

We can argue about the syntax: Right now the functions are `invariant_ring(::MPolyDecRing, ::Group)` (so the ring as first argument). This slightly collides with the existing `invariant_ring(::Field, ::Group)` which creates an invariant ring with the corresponding coefficient field. Should I change the signature to `invariant_ring([::Field], ::Group, [::MPolyDecRing])`?

### DIFF
--- a/src/InvariantTheory/invariant_rings.jl
+++ b/src/InvariantTheory/invariant_rings.jl
@@ -19,7 +19,16 @@ function invariant_ring(M::Vector{<: MatrixElem})
   return invariant_ring(base_ring(M[1]), M)
 end
 
+function invariant_ring(R::MPolyDecRing, M::Vector{<: MatrixElem})
+  K = coefficient_ring(R)
+  return invariant_ring(R, matrix_group([change_base_ring(K, g) for g in M]))
+end
+
 invariant_ring(matrices::MatrixElem{T}...) where {T} = invariant_ring(collect(matrices))
+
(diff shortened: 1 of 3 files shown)


wdecker:
To me, what you do here, looks fine from the user point of view. `julia` wizards?

`invariant_ring(G::MatrixGroup)`
`invariant_ring(K::Field = QQ, G::PermGroup)`
`invariant_ring(R::MPolyDecRing, G::MatrixGroup)`
`invariant_ring(R::MPolyDecRing, G::PermGroup)`

lgoettgens:
> `julia` wizards?
> 
> `invariant_ring(G::MatrixGroup)` `invariant_ring(K::Field = QQ, G::PermGroup)` `invariant_ring(R::MPolyDecRing, G::MatrixGroup)` `invariant_ring(R::MPolyDecRing, G::PermGroup)`

These four signatures are pairwise unambiguous, as `PermGroup` and `MatrixGroup` are disjoint types, as are `Field` and `MPolyDecRing`.

joschmitt:
Yes, my question was about the user point of view because now the first argument of `invariant_ring` can either be the ring on which the group acts or the coefficient ring of that ring. As we only allow fields as coefficient rings, there is no ambiguity from the julia side.

wdecker (approved):

PR #2846, opened 2023-09-22
Remove functions moved to Hecke

lgoettgens:
This is a counterpart to https://github.com/thofma/Hecke.jl/pull/1221.

Once this is merged, the 1.10 CI should no longer complain about precompilation issues.


src/exports.jl
```diff
@@ -438,7 +438,6 @@ export demazure_character
 export denest
 export denominator
 export denominators
-export dense_row
```
benlorenz, line 441: Shouldn't we keep the export for compatibility (i.e. re-export the Hecke function) ?

lgoettgens: It gets exported by due to thofma/Hecke.jl#1221 and reexported in Oscar in https://github.com/oscar-system/Oscar.jl/blob/aa8e64f19d58cd6e76176ce65d5223da614cd986/src/imports.jl#L176-L180

So this extra export can be safely removed.

benlorenz: Not saying that I like this approach, but I guess than this is fine for now.

benlorenz (approved):

PR #2847, opened 2023-09-23
`QuadFormWithIsom`: further cleanups + preparation for incoming new features

StevellM:
Here are some further minor changes for the package:
- Clean-up of the docstrings and of the html documentation (I was not happy with how it looked);
  * Fix the different bugs in display, the typos and other mistakes
  * Add more internal references to different part of the html documentation
  * Add more descriptions/comments for the internal functions 
 
- More improvements following the previous patch: 
  * Disable more unnecessary checks
  * Avoid the creation of some lists in largely used functions when it can be replace by "smarter math"
  * Avoid to use `cartesian_product_iterator` for products of two lists, and put back `inplace = true`  for products of multiple lists (we might save more time and memory now)
  * Rename some variables which should not have the same type in a same code
  
- Few more functionalities for the soon to be added complement on equivariant primitive embeddings:
  * Recognition of primary, elementary and unimodular lattices with isometry
  * Computation of orthogonal submodules (whenever it makes sense)
  *  Computation of the representation of a group of isometries for a lattice on the discriminant group of that lattice
  * `image_centralizer_in_Oq` extended to the general case using hermitian Miranda-Morrison and gluing of stabilizers along equivariant primitive extensions
  * Make the function `primitive_extensions` part of the available functions and not only a background routine (to be safe I have added a sort of deprecations)

### DIFF
--- a/experimental/QuadFormAndIsom/docs/src/enumeration.md
+++ b/experimental/QuadFormAndIsom/docs/src/enumeration.md
@@ -85,6 +85,6 @@ Note that an important feature from the theory in [BH23](@cite) is the notion of
 *admissible gluings* and equivariant primitive embeddings for admissible triples.
 In the next chapter, we present the methods regarding Nikulins's theory on primitive
 embeddings and their equivariant version. We use this basis to introduce the
-method [`admissible_equivariant_primitive_extension`](@ref) (Algorithm 2 in
+method [`admissible_equivariant_primitive_extensions`](@ref) (Algorithm 2 in
 [BH23](@cite)) which is the major tool making the previous enumeration
 possible and fast, from an algorithmic point of view.
--- a/experimental/QuadFormAndIsom/docs/src/enumeration.md
+++ b/experimental/QuadFormAndIsom/docs/src/enumeration.md
@@ -12,19 +12,19 @@ reference.
 
 ## Admissible triples
 
-Roughly speaking, for a prime number $p$, a *$p$-admissible triple* `(A, B, C)`
-is a triple of integer lattices such that, in certain cases, `C` can be obtained
-as a primitive extension $A \perp B \to C$ where one can glue along
-$p$-elementary subgroups of the respective discriminant groups of `A` and `B`.
+Roughly speaking, for a prime number $p$, a *$p$-admissible triple* $(A, B, C)$
+is a triple of integer lattices such that, in some cases, $C$ can be obtained
+as a primitive extension $A \oplus B \to C$ where one can glue along
+$p$-elementary subgroups of the respective discriminant groups of $A$ and $B$.
 Note that not all admissible triples satisfy this extension property.
 
-For instance, if $f$ is an isometry of an integer lattice `C` of prime order
-`p`, then for $A := \ker \Phi_1(f)$ and $B := \ker \Phi_p(f)$, one has that
-`(A, B, C)` is $p$-admissible (see Lemma 4.15. in [BH23](@cite)).
+For instance, if $f$ is an isometry of an integer lattice $C$ of prime order
+$p$, then for $A := \ker \Phi_1(f)$ and $B := \ker \Phi_p(f)$, one has that
+$(A, B, C)$ is $p$-admissible (see Lemma 4.15. in [BH23](@cite)).
 
-We say that a triple `(AA, BB, CC)` of genus symbols for integer lattices is
-*$p$-admissible* if there are some lattices $A \in AA$, $B \in BB$ and
-$C \in CC$ such that $(A, B, C)$ is $p$-admissible.
+We say that a triple $(G_A, G_B, G_C)$ of genus symbols for integer lattices is
+*$p$-admissible* if there are some lattices $A \in G_A$, $B \in G_B$ and
+$C \in G_C$ such that $(A, B, C)$ is $p$-admissible.
 
 We use Definition 4.13. and Algorithm 1 of [BH23](@cite) to implement the necessary
 tools for working with admissible triples. Most of the computations consists of
@@ -70,7 +70,8 @@ iteratively the order up to $p^dq^e$.
 ### Underlying machinery
 
 Here is a list of the algorithmic machinery provided by [BH23](@cite) used
-previously to enumerate lattices with isometry:
+previously to enumerate lattices with isometry. The following correspond
+respectively to Algorithms 3, 4, 5, 6 and 7 of the aforementioned paper:
 
 ```@docs
 representatives_of_hermitian_type(::ZZLatWithIsom, ::Int)
@@ -84,6 +85,6 @@ Note that an important feature from the theory in [BH23](@cite) is the notion of
 *admissible gluings* and equivariant primitive embeddings for admissible triples.
 In the next chapter, we present the methods regarding Nikulins's theory on primitive
 embeddings and their equivariant version. We use this basis to introduce the
-method `admissible_equivariant_primitive_extension` (Algorithm 2 in
+method [`admissible_equivariant_primitive_extension`](@ref) (Algorithm 2 in
 [BH23](@cite)) which is the major tool making the previous enumeration
 possible and fast, from an algorithmic point of view.
--- a/experimental/QuadFormAndIsom/src/embeddings.jl
+++ b/experimental/QuadFormAndIsom/src/embeddings.jl
@@ -268,7 +268,7 @@ function _subgroups_orbit_representatives_and_stabilizers(Vinq::TorQuadModuleMor
   end
 
   subs = TorQuadModule[s[1] for s in subs]
-  m = gset(O, _on_subgroup_automorphic, subs)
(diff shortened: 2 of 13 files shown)


experimental/QuadFormAndIsom/docs/src/enumeration.md
```diff
@@ -84,6 +85,6 @@ Note that an important feature from the theory in [BH23](@cite) is the notion of
 *admissible gluings* and equivariant primitive embeddings for admissible triples.
 In the next chapter, we present the methods regarding Nikulins's theory on primitive
 embeddings and their equivariant version. We use this basis to introduce the
-method `admissible_equivariant_primitive_extension` (Algorithm 2 in
+method [`admissible_equivariant_primitive_extension`](@ref) (Algorithm 2 in
```
fingolfin:
```suggestion
method [`admissible_equivariant_primitive_extensions`](@ref) (Algorithm 2 in
```
This causes the docs to fail to build

StevellM: Thanks!

fingolfin (approved):
Looks good to me. Perhaps @simonbrandhorst should also approve and/or merge it?

experimental/QuadFormAndIsom/src/embeddings.jl
```diff
@@ -547,29 +573,59 @@ function _classes_automorphic_subgroups(q::TorQuadModule,
...
+  - `classification == :sublat`: $V$ consists of representatives for all isomorphism classes of primitive extensions of $M\oplus N$ of the wanted form, up to the actions of $\overline{O(M)}$ and $\overline{O(N)}$;
+  - `classification == :emb`: $V$ consists of representatives for all isomorphism classes of primitive extensions of $M\oplus N$ of the wanted form, up to the action $\overline{O(N)}$;
+
+Note that the symbols for `classification` are similar to those used for the
+function [`primitive_embeddings(::ZZLat, ::ZZLat)`](@ref): the
+classification methods correspond to the different classes of primitive
+embeddings defined by Nikulin in [Nik79](@cite).
+
+Indeed, by fixing $H$, $M$ and $N$, one fixes a unique genus $G$ of even
+primitive extensions of $M\oplus N$ with a glue map described as above. We can
```
simonbrandhorst: I am not sure this statement is true. Can you give a reference?

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
```diff
@@ -1471,43 +1692,103 @@ julia> order(G)
...
-    UL = ZZMatrix[-matrix(one(OqL)), matrix(fqL)]
-    UL = elem_type(OqL)[OqL(m; check = false) for m in UL]
+    UL = elem_type(OqL)[OqL(m; check = false) for m in ZZMatrix[-matrix(one(OqL)), matrix(fqL)]]
     return sub(OqL, unique!(UL))
   elseif is_definite(L) || rank(L) == 2
     # We can compute the orthogonal groups and centralizer with GAP
     OL = orthogonal_group(L)
     f = isometry(Lf)
+    # The ambient isometry might not be identity on the complement of L, so we
+    # built another ambient isometry extending f by the identity
```
simonbrandhorst: Probably this can be fed into Plesken-Souvignier also directly. This may be much more efficient for 
hermitian lattice for instance.

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
```diff
@@ -1471,43 +1692,103 @@ julia> order(G)
...
+    # and we split the "largest non trivial exponent of the isometry".
+    #
+    # TODO: make a "smart search", for instance for some particular stable
+    # kernel sublattices of small rank or for which rank(L) == euler_phi(n)
+    #
+    # We use the similar method as used for extending stabilizers along
+    # equivariant primitive extensions as seen in the function
+    # `admissible_equivariant_primitive_extensions`.
+    divs = sort!(divisors(n); rev = true)
+    k = divs[findfirst(k -> rank(kernel_lattice(Lf, k)) != 0, divs)]
```
simonbrandhorst: You can also make this work beyond finite order: if the minmal polynomial is separable, and you iterate over its irreducible factors.

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
```diff
@@ -1471,43 +1692,103 @@ julia> order(G)
...
+
+    _, _, graph = _overlattice(phi, HMinD, HNinD, isometry(M), isometry(N); same_ambient = true)
+    disc, stab = _glue_stabilizers(phi, actM, actN, OqMinOD, OqNinOD, graph)
+    qL, fqL = discriminant_group(Lf)
+    OqL = orthogonal_group(qL)
+    phi = hom(qL, disc, TorQuadModuleElem[disc(lift(x)) for x in gens(qL)])
+    @hassert :ZZLatWithIsom 1 is_isometry(phi)
+    @hassert :ZZLatWithIsom 1 qL == disc
+
+    stab = sub(OqL, elem_type(OqL)[OqL(compose(phi, compose(g, inv(phi))); check = false) for g in stab])
```
simonbrandhorst, line 1798: Is the output `stab` of `_glue_stabilizers` a group or a list of generators?
If it is a group, you can iterate over the generators only here.

experimental/QuadFormAndIsom/src/embeddings.jl
StevellM: I might be wrong but I guess this follows from Nikulin. If you have fixed isometry class of subgroups $H_M$ for $D_M$ and $H_N$ for $D_N$ which are anti-isometric, then the isometry class of the discriminant form for the overlattice is independent on the choice of the representatives in the isometry classes of $H_M$ and $H_N$. Or am I using a wrong shortcut ? Otherwise I could just revise the documentation to avoid this statement.

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
StevellM: For the computation of which part we could use Plesken-Souvignier? The computation of the stabilizers ?

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
StevellM: Ah yes good point, I will modify it.

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
StevellM: Those are single elements, but I might have to change the `_glue_stabilizers` function to return unique occurrences (I did not check that I could have several times a same element)

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
```diff
@@ -1653,10 +1653,12 @@ function discriminant_representation(L::ZZLat, G::MatrixGroup;
   geneG = gens(G)
   for g in geneG
     if !ambient_representation
-      g = block_diagonal_matrix(QQMatrix[matrix(g), identity_matrix(QQ, nrows(B2))])
-      g = inv(C)*g*C
+      mg = block_diagonal_matrix(QQMatrix[matrix(g), identity_matrix(QQ, nrows(B2))])
+      mg = inv(C)*mg*C
```
simonbrandhorst: the computation of the inverse is better outside of the loop.

simonbrandhorst (changes requested):

experimental/QuadFormAndIsom/src/embeddings.jl
simonbrandhorst: As discussed, this part of the code needs to be changed to be correct.

experimental/QuadFormAndIsom/src/lattices_with_isometry.jl
```diff
@@ -1578,9 +1883,9 @@ function signatures(Lf::ZZLatWithIsom)
   lambda = C(eig[j])
   Sq = Int[i for i in 1:div(n,2) if gcd(i,n) == 1]
   D = Dict{Integer, Tuple{Int, Int}}()
-  f = change_base_ring(C, f)
+  fC = change_base_ring(C, f)
   for i in Sq
-    M = f + inv(f) - lambda^i - lambda^(-i)
+    M = fC + inv(fC) - lambda^i - lambda^(-i)
```
fingolfin: I guess this inv(fC)` could also be pulled out of the loop

StevellM:
There is more than what was originally planned: we remarked that the code for `primitive_embeddings` was not totally right for some cases that were not reached. It has been modified accordingly.

fingolfin:
What's the status here? Is this ready now, or is more work needed?

StevellM:
For me it is good to go, maybe @simonbrandhorst would like to check whether my fixes are correct.

simonbrandhorst (approved):

PR #2852, opened 2023-09-25
Fix doctest_fix for Documenter 1.0

fingolfin:


src/utils/docs.jl
```diff
@@ -80,12 +80,18 @@ Fixes all doctests for the file `n`, ie. all files in Oscar where
 function doctest_fix(n::String; set_meta::Bool = false)
   doc = get_document(set_meta)
 
+  if isdefined(Main.Documenter, :DocTests)
+    doctest = Main.Documenter.DocTests.doctest
+  else
+    doctest = Main.Documenter._doctest
+  end
```
lgoettgens, line 87:
```suggestion
  if isdefined(Main.Documenter, :_doctest)
    doctest = Main.Documenter._doctest
  else
    doctest = Main.Documenter.DocTests.doctest
  end
```
I would be more inclined to switch up the order, such that the Documenter 1.0 version is the first one. This makes the future dropping of 0.27 support clearer and more readable.

lgoettgens (approved):
LGTM, one comment below you are free to ignore.

src/utils/docs.jl
fingolfin: I don't agree that it is clearer, one still has to know which version is what. If we want this to be "clear" we need to switch this and other existing checks to check for a version. But that seems to be investing 5 minutes of work now to save 5 minutes of work later, so IMHO not worth it :-)

lgoettgens: True :)

PR #2857, opened 2023-09-27
Fix printing of character tables in Jupyter

fingolfin:
The header should not be inside math mode.


ThomasBreuer (approved):
Good, this fixes the Jupyter situation.

Just as an explanation:
My original idea was that the enclosing `$` characters should better be printed by the function that requests LaTeX format because this function might want to combine several LaTeX strings, and then can decide where a `$` is necessary.
I remember there were other cases of this kind in the code, but perhaps this code has been changed in the meantime.

PR #2858, opened 2023-09-27
Nicer detailed printing for RelativeBrauerGroupElem

fingolfin:
I used this in my software demo today. It made the difference between completely unreadable (to me) and at least decipherable.


fieker (approved):

fingolfin:
The failing doctest was due to randomization in an unrelated test involving a Galois context. I've restarted it now in the hopes that this will fix it. But I also submitted PR #2865 to fix the root cause.

PR #2863, opened 2023-09-28
PolyhedralGeometry: restrict signature of visualize, add visualize(SubdivisionOfPoints)

benlorenz:
also use `@req` macro

cc: @lkastner @YueRen

### DIFF
--- a/src/PolyhedralGeometry/visualization.jl
+++ b/src/PolyhedralGeometry/visualization.jl
@@ -27,7 +27,7 @@ end
 
 function visualize(P::SubdivisionOfPoints{T}) where T<:Union{QQFieldElem, Float64}
   d = ambient_dim(P)
-  @req ambient_dim(P) <= 3 "Can not visualize $(typeof(P)) of ambient dimension $d. Supported range: 1 <= d <= 3"
+  @req d <= 3 "Can not visualize $(typeof(P)) of ambient dimension $d. Supported range: 1 <= d <= 3"
   # polymake will by default use 0:n-1 as labels so we assign labels
   # starting from 1 here if there are no labels yet
   # (note: labels are mutable, i.e. they can be changed again later)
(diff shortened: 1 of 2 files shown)


YueRen:
Thanks for the fix!  One unrelated question: Is there a difference between the macros `@req` and `@assert`?  In particular, should one be preferred over the other in OSCAR code?

src/PolyhedralGeometry/visualization.jl
```diff
@@ -1,26 +1,39 @@
...
+  if !Polymake.exists(pm_object(P), "RAY_LABELS")
+    pm_object(P).RAY_LABELS = string.(1:Oscar.pm_object(P).N_RAYS)
+  end
+  pmo = pm_object(P)
+  Polymake.visual(pmo)
+end
+
+function visualize(P::SubdivisionOfPoints{T}) where T<:Union{QQFieldElem, Float64}
+  d = ambient_dim(P)
+  @req ambient_dim(P) <= 3 "Can not visualize $(typeof(P)) of ambient dimension $d. Supported range: 1 <= d <= 3"
```
lgoettgens: `d` isn't used anywhere. so this can be simplified to either
```suggestion
  @req ambient_dim(P) <= 3 "Can not visualize $(typeof(P)) of ambient dimension $d. Supported range: 1 <= d <= 3"
```
or
```suggestion
  d = ambient_dim(P)
  @req d <= 3 "Can not visualize $(typeof(P)) of ambient dimension $d. Supported range: 1 <= d <= 3"
```

benlorenz: Only the second option was valid since `d` is used in the error string.

benlorenz:
> Thanks for the fix! One unrelated question: Is there a difference between the macros `@req` and `@assert`? In particular, should one be preferred over the other in OSCAR code?

I think the `@req` macro is recommended in Oscar code for input validation. The `@assert` checks might be disabled at higher optimization levels.

lkastner (approved):
Thank you.

PR #2864, opened 2023-09-28
Another fix for printing of scheme-related types

StevellM:
Complement to #2856 for scheme objects.

- Fix the `:supercompact` printings for the scheme-related objects (which was too compact and not useful);
- Remove spacings for some morphism types;
- Replace occurrences of `Morphism` by `Hom` in the compact printings of morphisms.

@simonbrandhorst 

(I duplicated the changes of PR-2856 for two of the types to avoid potential conflicts when one of the two PR will be merged.)


experimental/Schemes/BlowupMorphism.jl
```diff
@@ -497,7 +497,7 @@ end
 function Base.show(io::IO, Bl::BlowupMorphism)
   io = pretty(io)
   if get(io, :supercompact, false)
-    print(io, "Morphism")
+    print(io, "Blow-up morphism")
```
fingolfin: Shouldn't this match the detailed printing below? It just uses `"Blow up"`. But perhaps they are intentionally different?

StevellM: Good point. Actually both of them are not as they should be since it should be a unique word. I did not notice that, I will modify it now, thanks!

simonbrandhorst (approved):

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Methods.jl
```diff
@@ -11,7 +11,7 @@ end
 
 function Base.show(io::IO, P::AbsProjectiveScheme{<:Any, <:MPolyQuoRing})
   if get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Projective scheme")
```
fingolfin, line 14: The detailed show message below uses `Projective space of dimension ...` ?

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -281,7 +281,7 @@ function Base.show(io::IO, CPS::CoveredProjectiveScheme)
   n = length(projective_patches(CPS))
   K = base_ring(base_scheme(CPS))
   if get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Covered projective scheme")
```
fingolfin: In the code below this is referred to as either "Empty covered projective scheme" or as "Relative projective scheme" (the latter twice). What gives?

experimental/Schemes/FunctionFields.jl
```diff
@@ -388,7 +388,7 @@ canonical_unit(f::VarietyFunctionFieldElem) = f # part of the ring interface tha
 function Base.show(io::IO, KK::VarietyFunctionField)
   io = pretty(io)
   if get(io, :supercompact, false)
-    print(io, "Field")
+    print(io, "Function field")
```
fingolfin: In the detailed printing method this is referred to as "Field of rational functions" instead

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1029,7 +1029,7 @@ function Base.show(io::IO, I::IdealSheaf, show_scheme::Bool = true)
   if has_attribute(I, :name)
     print(io, get_attribute(I, :name))
   elseif get(io, :supercompact, false)
-    print(io, "Presheaf")
+    print(io, "Ideal sheaf")
```
fingolfin: Below this is "Sheaf of ideals"

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Methods.jl
```diff
@@ -26,7 +26,7 @@ end
 function Base.show(io::IO, X::AffineAlgebraicSet{<:Field,<:MPolyQuoRing})
   io = pretty(io)
   if get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Affine algebraic set")
```
fingolfin, line 29: The "detailed show" method below calls this "Reduced subscheme"

src/AlgebraicGeometry/Schemes/AffineVariety/Objects/Methods.jl
```diff
@@ -49,7 +49,7 @@ end
 function Base.show(io::IO, X::AffineVariety)
   io = pretty(io)
   if get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Affine variety")
```
fingolfin, line 52: Below this is referred to once as "Affine open subset of ..." and once as "Affine space of dimension ..."

src/AlgebraicGeometry/Schemes/ProjectiveAlgebraicSet/Objects/Methods.jl
```diff
@@ -15,7 +15,7 @@ end
 function Base.show(io::IO, X::AbsProjectiveAlgebraicSet{<:Field, <:MPolyQuoRing})
   io = pretty(io)
   if get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Projective algebraic set")
```
fingolfin, line 18: In the "detailed" printing code above this is an "Algebraic set".

And in the "else" case below, no prefix seems to be shown at all?

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Morphisms/Methods.jl
```diff
@@ -98,10 +98,10 @@ end
 
 function Base.show(io::IO, f::ProjectiveSchemeMor)
   if get(io, :supercompact, false)
-    print(io, "Morphism")
+    print(io, "Projective scheme morphism")
```
fingolfin, line 126: Detailed printing still just uses "Morphism"...

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Methods.jl
```diff
@@ -15,7 +15,7 @@ end
 # one line printing
 function Base.show(io::IO, X::AbsProjectiveVariety{<:Field,<:MPolyQuoRing})
   if get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Projective variety")
```
fingolfin, line 18: Else case is missing this prefix and instead prints "V(...)" with "V" hardcoded that seems not good?

src/AlgebraicGeometry/Schemes/SpecOpen/Objects/Methods.jl
```diff
@@ -187,7 +187,7 @@ function Base.show(io::IO, U::SpecOpen, show_coord::Bool = true)
   if isdefined(U, :name) 
     print(io, name(U))
   elseif get(io, :supercompact, false)
-    print(io, "Scheme")
+    print(io, "Zariski open in affine scheme")
```
fingolfin: This is missing a noun ("Zariski open SOMETHING")

Also in the `else` it is `Complement to V(...)`

fingolfin (commented):
I find it really confusing that the supercompact, oneline and detailed printing for these objects often use completely different taglines.

Maybe this is justified in a few cases, but in all??

I think this is partly because there is no systematic testing for this printing and so one does not even see anywhere what changes to these cause, and how the different printing options look when put next to each other.

I've created a setup for making this kind of tests a bit easier to write, you can preview it in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1450/files#diff-249d4721dac939c93f2b04531e4728899bf98414e9ffeff88015a2e9516fd8b5 -- it will hopefully soon be available here (after some PR merging and releases)

StevellM:
I will go through the comments, thank you very much for this!

The differences between printing modes for a same types (with sometimes slight change of signatures) have been made on purpose, thoroughly. We can of course discuss the relevance/consistency, but words are put in a certain context and so the printings have been sometimes adapted to make them as clear as possible and avoid redundancy.

I agree that for some places, viewing some of your comments, there are things to be adapted that I did not notice the first time I made all those changes. For some others I do not agree, and I guess it is important to keep those differences because we do not bring the same amount of information between detailed and non-detailed printings.

The general motto was: we do not compute anything we do not know, but we put as much information as we know about the object in the (super)compact printings. In the case of `AffineVariety` for instance, by the signatures of the type, we can already deduce that it is an affine space, and the other cases we don't know... but in the case where we know it is an affine space, why shouldn't we print that it is an affine space ? 
This explains why we carry those difference of printings (similarly with the `Empty` version of some of the printing: isn't it easier to know directly by the printing that something is empty if the object knows it is empty ?)

experimental/Schemes/CoveredProjectiveSchemes.jl
StevellM: It should be all "Relative projective scheme", thank you.

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Methods.jl
StevellM: The idea of the supercompact printing was to just give the essential, and no dependence on the type parameters. All of them are affine algebraic set: some are given by an ideal, others are not and we "just know" that they are the reduced subscheme of a given object. So for the non supercompact printings, we show what we know. The problem is that for objects of the same type (maybe different type parameters), we do not always have the same amount of information...

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Methods.jl
StevellM: Here was one of the solution to obtain explicit short printing, that could also fit well in nested printings. For a projective variety, we know the defining ideal `I` and it is common to use the notation `V(I)` to describe such objects. This is why the "V" is hardcoded, and then we print the ideal in between parentheses.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -281,10 +269,10 @@ function Base.show(io::IO, CPS::CoveredProjectiveScheme)
   n = length(projective_patches(CPS))
   K = base_ring(base_scheme(CPS))
   if get(io, :supercompact, false)
-    print(io, "Covered projective scheme")
+    print(io, "Relative projective scheme")
```
afkafkafk13: Please check this with @HechtiDerLachs -- "Covered projective scheme" has the wording of the underlying type built into it. This might be on purpose.

src/AlgebraicGeometry/Schemes/SpecOpen/Morphisms/Methods.jl
```diff
@@ -20,7 +20,7 @@ function Base.show(io::IO, ::MIME"text/plain", f::SpecOpenMor)
   kY = length(str)
   push!(co_str, str)
   k = max(length.(co_str)...)
-  println(io, "Morphism")
+  println(io, "Spec open morphism")
```
afkafkafk13, line 23: Not for this PR, but for later discussion:
`Spec open` is not a standard wording. Maybe it should be printed as a typename or with a different wording.

StevellM: Yes I agree. To be honnest I was not sure what to do with this without having very long names. At least the printing looks consistent and it carries the wording of the typename (for now)

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Methods.jl
fingolfin: I just didn't realize that the "detailed printing" *below* this method is not the matching one; the matching one is *above*, and it is consistent. So that is fine!

fingolfin (approved):
OK thanks, this is now much better in terms of consistency!

PR #2865, opened 2023-09-28
Adjust printing of Galois context objects

fingolfin:
- ensure 'Galois' is always uppercase
- ... but 'context' should be lowercase
- adjust some doctest filters that did not match anymore
  (they matched for "Group" where now we had "group")

### DIFF
--- a/src/NumberTheory/GaloisGrp/GaloisGrp.jl
+++ b/src/NumberTheory/GaloisGrp/GaloisGrp.jl
@@ -6,6 +6,8 @@ import Oscar: Hecke, AbstractAlgebra, GAP, extension_field, isinteger,
               upper_bound
 using Oscar: SLPolyRing, SLPoly, SLPolynomialRing, CycleType
 
+import Oscar: pretty, LowercaseOff
+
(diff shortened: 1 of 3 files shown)


src/NumberTheory/GaloisGrp/GaloisGrp.jl
```diff
@@ -485,7 +485,7 @@ mutable struct ComplexRootCtx
 end
 
 function Base.show(io::IO, GC::GaloisCtx{ComplexRootCtx})
-  print(io, "Galois Context for $(GC.f) using complex roots")
+  print(io, LowercaseOff(), "Galois context for $(GC.f) using complex roots")
```
lgoettgens: This is missing a `pretty` call

lgoettgens: Same for others as well

fingolfin: Thanks, fixed now

fieker (approved):
be my guest

PR #2866, opened 2023-09-28
support both documenter versions in doctest helper code

benlorenz:
extracted commit from #2605

cc: @lkastner @lgoettgens


lkastner (approved):
Thank you.

lgoettgens (approved):

PR #2868, opened 2023-09-28
Implement Hessian matrix for polynomials

HechtiDerLachs:
As requested by @wdecker .

### DIFF
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -1300,7 +1300,7 @@ function hessian_matrix(f::MPolyElem)
   R = parent(f)
   n = nvars(R)
   df = jacobi_matrix(f)
-  result = zero(MatrixSpace(R, n, n))
+  result = zero_matrix(R, n, n)
   for i in 1:n
     for j in i:n
       result[i, j] = result[j, i] = derivative(df[i], j)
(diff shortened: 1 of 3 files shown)


src/Rings/mpoly.jl
```diff
@@ -1294,3 +1294,20 @@ end
 
 # assure compatibility with generic code for MPolyQuos:
 lift(f::MPolyRingElem) = f
+
+### Hessian matrix
+function hessian_matrix(f::MPolyElem)
+  R = parent(f)
+  n = nvars(R)
+  df = jacobi_matrix(f)
+  result = zero(MatrixSpace(R, n, n))
```
lgoettgens:
```suggestion
  result = zero_matrix(R, n, n)
```
This saves the creation of the matrix space.

wdecker (approved):
@HechtiDerLachs thx; what about the suggestion by @lgoettgens ?

src/Rings/mpoly.jl
HechtiDerLachs: Thanks! I was not aware of this constructor. Probably this has to do with the rather recent decision to drop parent objects for matrices over rings? Either way, this is reasonable.

thofma: This has been there for a few years

fingolfin: And which "decision to drop parent objects for matrices over rings" ? I am not aware of such a decision.

afkafkafk13:
to be reopened

afkafkafk13:
and reopen

PR #2871, opened 2023-09-29
fix for serializing 0 ideal

antonydellavecchia:
Didn't realise base_ring was the function i was looking for

### DIFF
--- a/src/Serialization/Rings.jl
+++ b/src/Serialization/Rings.jl
@@ -259,7 +259,7 @@ const IdealUnionType = Union{MPolyIdeal, LaurentMPolyIdeal, FreeAssAlgIdeal}
 function save_type_params(s::SerializerState, x::T) where T <: IdealUnionType
   save_data_dict(s) do
     save_object(s, encode_type(T), :name)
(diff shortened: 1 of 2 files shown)


src/Serialization/Rings.jl
```diff
@@ -259,7 +259,7 @@ const IdealUnionType = Union{MPolyIdeal, LaurentMPolyIdeal, FreeAssAlgIdeal}
 function save_type_params(s::SerializerState, x::T) where T <: IdealUnionType
   save_data_dict(s) do
     save_object(s, encode_type(T), :name)
-    ref = save_as_ref(s, parent(gens(x)[1]))
+    ref = save_as_ref(s, base_ring(x))
```
HechtiDerLachs, line 262: Yes, that's exactly what I was thinking. Sorry, I should have spelled out that fix straight away. At some point it was also suggested to support `ring(I::Ideal)` rather than `base_ring(I::Ideal)`, but that does not seem to have gone through, yet.

HechtiDerLachs (approved):

PR #2872, opened 2023-09-29
Switch from FiniteField to finite_field

fingolfin:
Command used to create this PR:

    git grep -l -w FiniteField | xargs perl -pi -e 's;\bFiniteField\b;finite_field;g'


fingolfin:
Hrmpf, failures are once again of this kind (and thus not caused by this PR):
```
Arithmetic for Finite field of degree 1 over GF(3): Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Rings/AlgClosureFp.jl:89
  Test threw exception
  Expression: a * (b + c) == a * b + a * c
  MethodError: objects of type Nothing are not callable
  Stacktrace:
   [1] (::fqPolyRepField)(b::fqPolyRepFieldElem)
     @ Nemo ~/.julia/packages/Nemo/iAwYA/src/flint/fq_nmod.jl:660
   [2] op(f::typeof(+), a::AlgClosureElem{fqPolyRepField}, b::AlgClosureElem{fqPolyRepField})
     @ Oscar.AlgClosureFp ~/work/Oscar.jl/Oscar.jl/src/Rings/AlgClosureFp.jl:129
   [3] +(a::AlgClosureElem{fqPolyRepField}, b::AlgClosureElem{fqPolyRepField})
     @ Oscar.AlgClosureFp ~/work/Oscar.jl/Oscar.jl/src/Rings/AlgClosureFp.jl:132
```
@fieker would be really good if we could do something about this :-(

PR #2877, opened 2023-10-02
Refactor some Map printing code

fingolfin:
Reduce some code duplication and prepare for a further reduction in the future.

The primary goal here is to remove gratuitous differences in the code, which in turn makes it trivial to detect possibly unwanted differences and other problematic behavior. I'll add some inline comments to point out specific changes.

See also https://github.com/Nemocas/AbstractAlgebra.jl/pull/1450 for related changes in AbstractAlgebra.

CC @StevellM


experimental/LieAlgebras/src/LieAlgebraHom.jl
```diff
@@ -71,7 +71,7 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", h::LieAlgebraHom)
   io = pretty(io)
-  println(io, LowercaseOff(), "Lie algebra morphism")
+  println(IOContext(io, :supercompact => true), h)
```
fingolfin, line 74: This ensures that this detailed printing method is identical to that for `LieAlgebraModuleHom` or `GAPGroupHomomorphism`

experimental/Schemes/MorphismFromRationalFunctions.jl
```diff
@@ -122,7 +122,9 @@ function Base.show(io::IOContext, Phi::MorphismFromRationalFunctions)
   if get(io, :supercompact, false)
     print("Morphism from rational functions")
   else
-    print("Hom: ", X, " -> ", Y)
```
fingolfin, line 125: Fixes a bug, `X` and `Y` are not defined in this function

fingolfin: (This also means there is no test for this printing variant, so ideally we should add one)

experimental/Schemes/MorphismFromRationalFunctions.jl
```diff
@@ -122,7 +122,9 @@ function Base.show(io::IOContext, Phi::MorphismFromRationalFunctions)
...
+    io = pretty(io)
+    print(io, "Hom: ")
+    print(io, Lowercase(), domain(Phi), " -> ", Lowercase(), codomain(Phi))
```
fingolfin, line 127: In other functions we print this in "supercompact" mode. But not here and not in some others. One can argue either way, for now I wanted to keep this change "simple" while also aligning it as much as possible with other `show` methods for maps here and in other repos.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Methods.jl
```diff
@@ -19,7 +17,7 @@ function Base.show(io::IO, X::AbsProjectiveVariety{<:Field,<:MPolyQuoRing})
   elseif get_attribute(X, :is_empty, false)
     print(io, "Empty projective variety")
   else
-    print(io, "V(")
+    print(io, LowercaseOff(), "V(")
```
fingolfin, line 20: This ensures the `V` always is in uppercase, so technically this is a bug fix, albeit a very minor one

src/Rings/mpoly-localizations.jl
```diff
@@ -2722,29 +2722,20 @@ julia> (PSI, )
...
-      println(io, "$(R[i]) -> $(psi(R[i]))")
-    end
-    n = ngens(R)
-    print(io, "$(R[n]) -> $(psi(R[n]))")
+  println(io, Dedent(), "defined by", Indent())
+  R = base_ring(domain(phi))
+  psi = restricted_map(phi)
+  to = is_unicode_allowed() ? " ↦ " : " -> "
+  for i in 1:ngens(R)-1
+    println(io, R[i], to, psi(R[i]))
```
fingolfin, line 2735: This change is just about minimizing code duplication and simplification... though it also fixes a bug (in the unicode variant, there was a final `println` instead of a final `print`)

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -208,22 +208,12 @@ function Base.show(io::IO, ::MIME"text/plain", f::AbsSpecMor)
...
-  io = pretty(io)
   if get(io, :supercompact, false)
     print(io, "Affine scheme morphism")
   else
-    X = domain(f)
-    Y = codomain(f)
+    io = pretty(io)
     print(io, "Hom: ")
-    if typeof(X) <: Union{PrincipalOpenSubset, AffineVariety{ <:Field, <: MPolyAnyRing}, <:Spec{<:Field, <:MPolyAnyRing}}  # Take care of the case where the domain is not given as a V(bla)
-      print(io, Lowercase())
```
fingolfin, line 219: This case was checking for specific types of `X` to decide whether or not to issue `Lowercase()` -- that's a bad code smell, don't do that! Here, the proper fix is to ensure that the sub-object issues `LowercaseOff` if necessary (I've done on such change elsewhere in this PR).

@StevellM There is more printing code for Schemes which performs similar checks that should be changed. I'll write to you about this separately on Slack.

lgoettgens (approved):
Changes to Lie algebra stuff look fine. Didn't check the rest.

src/AlgebraicGeometry/Schemes/ProjectiveVariety/Objects/Methods.jl
StevellM: Good to have this function, I was not aware of this one, thanks !

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
StevellM: Good, as mentioned in another comment, I was not aware this was an option. I will add it to the list of things to be changed among others for the schemes' printings.

experimental/Schemes/MorphismFromRationalFunctions.jl
StevellM: I do not understand, which part is printed in "supercompact" in other functions ?

fingolfin: This is "one line" printing mode, and the recommendation (though not a requirement) is that any delegation it performs uses "supercompact" mode, to help ensure the "one line" promise is kept. So for other kinds of maps we do
```
print(IOContext(io, :supercompact => true), Lowercase(), domain(x), " -> ", Lowercase(), codomain(x))
```

PR #2880, opened 2023-10-04
Fix doctest_fix(f::Function)

fingolfin:


fieker (approved):
looks good - abd experimentally seems to help...

PR #2881, opened 2023-10-04
Try to make AlgClosureFp tests fail less often

fingolfin:
Instead of using the global cached versions of the starting fields,
use fresh copies to avoid interference by the global effects of
previous tests.

Of course this does not solve issue #2691, but for now
it should reduce the number of times CI fails randomly due to this.

This contains PR #2809 because that PR seemed to trigger these failures consistently, so if this PR here now passes, it may actually work...


fingolfin:
@fieker unfortunately this doesn't seem to help

fingolfin:
Unfortunately doesn't work, so I made #2888 instead

PR #2882, opened 2023-10-04
Better implementation of `show` methods for schemes

StevellM:
At the time I revised the functions for the printing of `Schemes`, the codes I wrote were bad in some places and I often misused the methods `show`. This is a fix of this, namely:
- Use more `LowercaseOff` for objects we do not want to decapitalize (and remove the `typeof` elsewhere);
- Remove the distinction for `base_ring` to be equal to `QQ` or not (at that time, the compact printing of `QQ` did not use the `LowercaseOff()`, now it does);
- Use more `ItemQuantity` for pluralization of nouns in the printing (whenever it makes sense);
- Revise the `show` methods: I pass the covering, the use of `_show_semi_compact` and others into an `IOContext` for nested printings.

These changes will not affect the printings (except the pluralization with `0`, it was time to learn some English grammar for me) since it is just about code design.

Together with those changes, I have updated the html developer documentation about printings to include the above tools I have used thanks to the suggestions of @fingolfin .


experimental/Schemes/AlgebraicCycles.jl
```diff
@@ -301,6 +301,10 @@ end
 function Base.show(io::IO, ::MIME"text/plain", D::AlgebraicCycle, cov::Covering = get_attribute(scheme(D), :simplified_covering, default_covering(scheme(D))))
   io = pretty(io)
   X = scheme(D)
+  # If the IO context knows about a covering to be used, we used this one.
```
fingolfin:
```suggestion
  # If the IO context knows about a covering to be used, we use this one.
```

experimental/Schemes/AlgebraicCycles.jl
```diff
@@ -301,6 +301,10 @@ end
...
+  # Otherwise, we check whether X has a simplified covering. If not, we use the
+  # default covering of X
+  cov = get(io, :covering, get_attribute(X, :simplified_covering, default_covering(X)))
```
fingolfin: This exact code snippet it used a lot, perhaps makes sense to put it into a helper?
```julia
_covering_for_printing(io, X) = get(io, :covering, get_attribute(X, :simplified_covering, default_covering(X)))
```

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -278,8 +278,7 @@ function Base.show(io::IO, CPS::CoveredProjectiveScheme)
     end
     print(io, Lowercase(), base_scheme(CPS))
     if n != 0
-      print(io, " covered with $n projective patch")
-      n > 1 && print(io, "es")
+      print(io, " covered with ", ItemQuantity(n, "projective patch", "projective patches"))
```
fingolfin: Can be simplified once https://github.com/Nemocas/AbstractAlgebra.jl/pull/1460 lands and is released to
```suggestion
      print(io, " covered with ", ItemQuantity(n, "projective patch"))
```

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Methods.jl
```diff
@@ -24,18 +24,15 @@ end
...
   elseif get_attribute(X, :is_empty, false)
+    io = pretty(io)
     print(io, "Empty affine algebraic set over ")
     K = base_ring(X)
-    if K == QQ
-      print(io, "QQ")
-    else
-      print(IOContext(io, :supercompact => true), Lowercase(), K)
-    end
+    print(IOContext(io, :supercompact => true), Lowercase(), K)
```
fingolfin, line 33: Very nice.

(Though I notice that this is not covered by tests, like many of the other printing routines here... :-( )

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -165,32 +165,10 @@ function Base.show(io::IO, ::MIME"text/plain", f::AbsSpecMor)
...
-                                # spaces (since the coordinates already appear
-                                # before)
-    else
-      print(io, X)
-    end
-    println(io)
-  else
-    println(io, X)
-  end
+  println(IOContext(io, :show_coordinates => false), Lowercase(), X)
```
fingolfin, line 168: So much nicer :-)

fingolfin (approved):
Nice!

experimental/Schemes/AlgebraicCycles.jl
StevellM: Good idea.

experimental/Schemes/CoveredProjectiveSchemes.jl
StevellM: Ok so if I push the changes some doctests will fail but then we re-run everything once the new Nemo is here ?

fingolfin: New AA, actually. Basically, just change the `Project.toml` of Oscar to require AA 0.32.5 and add that to this PR, then push it once https://github.com/JuliaRegistries/General/pull/92810 is merged (probably in 15-20 minutes)

StevellM:
Some tests do not pass and it does not seem to be related to the content of the PR.

PR #2887, opened 2023-10-05
Matroids: Choose edges of graph as ground set for cycle_matroid

lkastner:


bschroter (commented):
The changes look fine to me. But I want to remark the following:

1) It is odd that a cycle matroid gets now names from the edges of the graph while the dual matroid, i.e., bond or coycle matroid doesn't. This will confuse the user and should be a small fix.

2) The code assumes that edges are labeled by two nodes (source and destination) which are of type `Int` Should be more general edge labels and node labels be considered?

lkastner:
>     1. It is odd that a cycle matroid gets now names from the edges of the graph while the dual matroid, i.e., bond or coycle matroid doesn't. This will confuse the user and should be a small fix.
> 
I do not understand this remark. In your code you have
```
bond_matroid(g::Graph) = dual_matroid(cycle_matroid(g))
```
so if `dual_matroid` works in the right way, this should just work. Let me try:
```
julia> cm = cycle_matroid(complete_graph(4))
Matroid of rank 3 on 6 elements

julia> dcm = dual_matroid(cm)
Matroid of rank 3 on 6 elements

julia> matroid_groundset(dcm)
6-element Vector{Edge}:
 Edge(2, 1)
 Edge(3, 1)
 Edge(3, 2)
 Edge(4, 1)
 Edge(4, 2)
 Edge(4, 3)
```
I am not sure what to fix here.

>     2. The code assumes that edges are labeled by two nodes (source and destination) which are of type `Int` Should be more general edge labels and node labels be considered?

Which code are you referring to? The line I changed in the Matroid section is just
```
gs = collect(edges(g))
```
does not assume anything. If you are talking about the conversion, then you are correct. We modelled the Graphs such that they can be used in the same way as Julia's Graphs.jl package in order to make migrating code easier (usually I would call this "compatible", but @lgoettgens frowns up on this). These use only `Int`s and right now our graphs also only use `Int`s. Nevertheless if we decide to change this, this change would be in the Graphs section and the Matroid section should not be affected at all.

bschroter:
Regarding 1) Yes, you are right. What was irritating me is the change in `docs/src/Combinatorics/matroids.md` line 29. The rest should automatically work.

Regarding 2) Currently the ground set will take the edges (pairs of nodes), but one might want to have other names, i.e., a labelling of nodes or edges which should then be taken as ground set instead. It is more an open question the a issue.

benlorenz (approved):

PR #2889, opened 2023-10-05
Prefer graded_polynomial_ring over grade

fingolfin:


src/Rings/mpoly-graded.jl
```diff
@@ -410,14 +410,18 @@ julia> T, (x, y, z) = graded_polynomial_ring(QQ, ["x", "y", "z"])
 ```
 """
 function graded_polynomial_ring(C::Ring, V::Union{Tuple{Vararg{T}}, AbstractVector{T}},
-      W=ones(Int, length(V)); ordering=:lex) where
+      W=ones(Int, length(V)); ordering=:lex, cached::Bool=true) where
       T<:VarName
-   return grade(polynomial_ring(C, V; ordering)[1], W)
+   # HACK: we pass 'cached' only to 'polynomial_ring', but it really should
+   # also affect whether the MPolyDecRing gets cached...
+   return grade(polynomial_ring(C, V; ordering, cached)[1], W)
```
fingolfin, line 417: @fieker FYI

fieker (approved):

HereAround (approved):
Thank you @fingolfin . I like this change!

HereAround:
@fingolfin I have taken the liberty to fix the failed doctests. Also, auto-merge is now enabled. If everything goes according to plan, this PR will thus be merged shortly.

PR #2890, opened 2023-10-05
Fix sheaf_cohomology_bgg docstring, sheafCohTable printing

fingolfin:


fingolfin:
@RafaelDavidMohr please have a look to verify this is OK

RafaelDavidMohr:
Sorry for the delay. Looks good to me.

src/Modules/ModulesGraded.jl
```diff
@@ -1581,7 +1581,7 @@ S^4 <---- S^6 <---- S^4 <---- S^1 <---- 0
 julia> M = cokernel(map(FI, 2));
 
 julia> tbl = sheaf_cohomology_bgg(M, -6, 2)
-       -6  -5  -4  -3  -2  -1   0   1   2
+.      -6  -5  -4  -3  -2  -1   0   1   2
```
fingolfin: @wdecker @RafaelDavidMohr @ederc I've now tweaked the printing of this table to avoid running into the issue with Documenter (if the first line of output is indented by more than 6 spaces, it thinks this is a continuation of the input line above.

As you can see, the tweak is that the first thing printed now is a dot `.` not a space. Not great but it works.

ederc: @wdecker Maybe instead of a dot we could print something like `S` (for shift) or `D` (for degree) for this row?

ederc: Discussed with @wdecker to add `twist: ` as header for the first row, updated PR correspondingly

wdecker (approved):

src/Modules/ModulesGraded.jl
fingolfin: Great, thanks!

HereAround (approved):
Thank you @fingolfin . I also like the word "twist" in the table.

PR #2894, opened 2023-10-06
All monomial for free standard graded modules

HechtiDerLachs:
As discussed with @jankoboehm : An iterator over all monomials in standard graded free modules. To be used as a backend for eventual implementations of `vector_space` and friends. For the non-standard-graded case we will need other functionality, but this is not a contradiction. Nothing is exported and what is added fits with the work by @joschmitt .


src/InvariantTheory/InvariantTheory.jl
```diff
@@ -6,3 +6,6 @@ include("primary_invariants.jl")
 include("secondary_invariants.jl")
 include("fundamental_invariants.jl")
 include("affine_algebra.jl")
+
+# postponed inclusion from src/Modules/Modules.jl due to dependency
+include("../Modules/Iterators.jl")
```
joschmitt, line 11: This is not really about this PR, but it seems to me like we should start thinking about putting all types (`(mutable) struct`s) in a handful of 'central' places or at least make sure that they can be included at first.

Although in this case we could also just move the `all_monomials` for rings to the `Rings/` directory where it arguably belongs.

src/Modules/Iterators.jl
```diff
@@ -0,0 +1,89 @@
...
+function Base.length(amm::AllModuleMonomials)
+  F = underlying_module(amm)
+  r = rank(F)
+  R = base_ring(F)
+  d = degree(amm)
+  result = 0
+  for i in 1:r
+    d_loc = d - Int(degree(F[i])[1])
+    d_loc < 0 && continue
+    result = result + length(all_monomials(R, d_loc))
```
joschmitt, line 32: One should be aware that `all_monomials` completely ignores the grading of a `MPolyDecRing`. But you only consider the standard graded case, so it's fine.

src/Modules/Iterators.jl
```diff
@@ -0,0 +1,89 @@
...
+  d = degree(amm)
+  R = base_ring(F)
+
+  d_loc = d - Int(degree(F[i])[1])
+  while i < ngens(F) && d_loc < 0
+    i = i+1
+    d_loc = d - Int(degree(F[i])[1])
+  end
+  
+  i == ngens(F) && d_loc < 0 && return nothing
```
joschmitt:
```suggestion
  i = findfirst(i -> d - Int(degree(F[i])[1]) >= 0, 1:ngens(F))
  i === nothing && return nothing
  d_loc = d - Int(degree(F[i])[1])
```

src/Modules/Iterators.jl
```diff
@@ -0,0 +1,89 @@
...
+    i == ngens(F) && return nothing
+    i = i+1
+    d_loc = d - Int(degree(F[i])[1])
+
+    while i < ngens(F) && d_loc < 0
+      i = i+1
+      d_loc = d - Int(degree(F[i])[1])
+    end
+
+    i == ngens(F) && d_loc < 0 && return nothing
```
joschmitt:
```suggestion
    i = findnext(i -> d - Int(degree(F[i])[1]) >= 0, 1:ngens(F), i + 1)
    i === nothing && return nothing
    d_loc = d - Int(degree(F[i])[1])
```

src/Modules/Iterators.jl
```diff
@@ -0,0 +1,89 @@
...
+  while i < ngens(F) && d_loc < 0
+    i = i+1
+    d_loc = d - Int(degree(F[i])[1])
+  end
+  
+  i == ngens(F) && d_loc < 0 && return nothing
+
+  mon_it = all_monomials(R, d_loc)
+  res = iterate(mon_it, nothing)
+  res === nothing && i == ngens(F) && return nothing
```
joschmitt, line 49: Can `res === nothing` happen? `mon_it` is an iterator over a component of non-negative degree, so there should always be at least one monomial, no?

src/Modules/Iterators.jl
```diff
@@ -0,0 +1,89 @@
...
+    while i < ngens(F) && d_loc < 0
+      i = i+1
+      d_loc = d - Int(degree(F[i])[1])
+    end
+
+    i == ngens(F) && d_loc < 0 && return nothing
+
+    mon_it = all_monomials(R, d_loc)
+    res_loc = iterate(mon_it, nothing)
+    res_loc === nothing && i == ngens(F) && return nothing
```
joschmitt, line 69: As above.

joschmitt (approved):
Looks good to me, I left some minor comments.

src/InvariantTheory/InvariantTheory.jl
HechtiDerLachs: Yes, this discussion has to take place at some point. But not here, eventually.

src/Modules/Iterators.jl
HechtiDerLachs: Exactly.

src/Modules/Iterators.jl
HechtiDerLachs: Let's see whether this works.

src/Modules/Iterators.jl
HechtiDerLachs: I don't know. Probably not. But the line doesn't hurt.

joschmitt: But what does happen if `res === nothing` and `i != ngens(F)`, so the `return nothing` is not reached?
In the next line, you assign `x, s = res` which throws an error.
Handling this case properly (increase `i` etc.) might be possible, but doesn't seem to be worth the effort.
```suggestion
  @assert res !== nothing
```
Maybe this would be clearer then?

HechtiDerLachs: Yes, you might be right. But I don't want to interrupt the tests again now. Let's first see whether they pass in principle. In either case, the line doesn't seem to be breaking things. Worst case it's confusing to the reader.

HechtiDerLachs:
From my side this is good to be merged. In my opinion, the small change discussed above is not worth spending the time and (electric) energy to running all the tests again.

test/Modules/ModulesGraded.jl
```diff
@@ -1176,3 +1176,15 @@ end
...
+  S, (x, y, u, v, w) = grade(R)
+
+  F = graded_free_module(S, [-1, 2])
+
+  for d in -4:4
+    amm = Oscar.all_monomials(F, d)
+    @test d < -1 || !isempty(amm)
+    @test all(x->degree(x) == grading_group(F)([d]), amm)
+  end
+end
```
afkafkafk13, line 1190: Why is this identical test in ModulesGraded.jl and in GradedModules.jl?
(By the way, is the latter filename the result of a mishap or intentional?)

HechtiDerLachs: I put the tests in one file and ran it locally. Tests failed completely, but not due to the part I added. I suppose that test-file was deprecated, but not removed. Just to be sure, I then put my tests in both, finally. Somebody should tell us what's up with the test files, though.

afkafkafk13 (approved):
For me, this is good to go. My comment was not meant as blocking.

PR #2895, opened 2023-10-08
De Rham complexes

HechtiDerLachs:
Based on the recent work in #2879 and to be rebased, eventually.


src/Combinatorics/OrderedMultiIndex.jl
```diff
@@ -0,0 +1,182 @@
...
+
+Base.eltype(I::OrderedMultiIndexSet) = OrderedMultiIndex
+Base.length(I::OrderedMultiIndexSet) = binomial(bound(I), index_length(I))
+
+function Base.iterate(I::OrderedMultiIndexSet)
+  ind = OrderedMultiIndex([i for i in 1:index_length(I)], bound(I))
+  return ind, ind
+end
+
+function Base.iterate(I::OrderedMultiIndexSet, state::OrderedMultiIndex)
```
afkafkafk13: please put a commenting line at each of the two steps (to ensure that we still see at first glance whats going on in a few years). Maybe something like:
a) find right-most non-maximal entry at position l (and detect if state is the last one) (lines 132-137)
b) increment index at position l and set further indices to minimal value (lines 138-142)

src/Combinatorics/OrderedMultiIndex.jl
```diff
@@ -0,0 +1,182 @@
...
+  for i in indices(ind)
+    print(io, "< $i ")
+  end
+  print(io, "<= $(bound(ind))")
+end
+
+# For an ordered multiindex i = (0 < i₁ < i₂ < … < iₚ ≤ n) this 
+# returns the number k so that i appears at the k-th spot in the 
+# enumeration of all ordered multiindices for this pair 0 ≤ p ≤ n.
+function linear_index(ind::OrderedMultiIndex)
```
afkafkafk13: I am not 100% happy with the name. This returns the linear index of an OrderedMultiIndex in an OrderedMultiIndexSet -- maybe just too many uses of index floating around...
Even if it is really early in the development of this part, maybe a small example would be helpful for the reader.

src/Modules/ExteriorPowers/FreeModules.jl
```diff
@@ -0,0 +1,191 @@
...
+    K = chain_complex([phi])
+    KM = tensor_product(K, M)
+    return kernel(map(KM, 1))[1]
+  end
+
+  if iszero(i) # Homology at the last entry of the complex.
+    phi = wedge_multiplication_map(exterior_power(F, n-1)[1], exterior_power(F, n, cached=cached)[1], v)
+    K = chain_complex([phi])
+    KM = tensor_product(K, M)
+    return cokernel(map(K, 1)) # TODO: cokernel does not seem to return a map by default. Why?
```
afkafkafk13: @jankoboehm or @wdecker: Do you know whether there are reasons for this different behaviour or whether this is just an inconsistency from the evolution of the system?

src/Modules/ExteriorPowers/FreeModules.jl
```diff
@@ -0,0 +1,191 @@
...
+  end
+
+  ext_powers = [exterior_power(F, p, cached=cached)[1] for p in i-1:i+1]
+  boundary_maps = [wedge_multiplication_map(ext_powers[p], ext_powers[p+1], v) for p in 1:2]
+  K = chain_complex(boundary_maps)
+  return homology(K, 1)
```
afkafkafk13: Did I miss something or is the tensoring with M missing here?

src/Modules/ExteriorPowers/Generic.jl
```diff
@@ -0,0 +1,184 @@
...
+      end
+      success, _, q = is_exterior_power(Oscar.parent(v))
+      !success && (q = 1)
+      exterior_power(F, p + q)[1]
+    end
+  )
+  success1, F1, p = is_exterior_power(Oscar.parent(u))
+  if !success1
+    F = Oscar.parent(u) 
+    Fwedge1, _ = exterior_power(F1, 1)
```
afkafkafk13: This looks a bit weird to me. You are setting F without using it (before the corresponding return statement).

src/Modules/ExteriorPowers/Generic.jl
```diff
@@ -0,0 +1,184 @@
...
+    return wedge(Fwedge1(coordinates(u)), v, parent=parent)
+  end
+
+  success2, F2, q = is_exterior_power(Oscar.parent(v))
+  if !success2
+    F = Oscar.parent(v) 
+    Fwedge1, _ = exterior_power(F1, 1)
```
afkafkafk13: same as a few lines above

src/Modules/ExteriorPowers/Generic.jl
```diff
@@ -0,0 +1,184 @@
...
+    parent::ModuleFP=begin
+      success, F, p = is_exterior_power(Oscar.parent(u))
+      if !success 
+        F = Oscar.parent(u)
+        p = 1
+      end
+      success, _, q = is_exterior_power(Oscar.parent(v))
+      !success && (q = 1)
+      exterior_power(F, p + q)[1]
+    end
```
afkafkafk13: For me this makes the initialization of `parent` rather hard to read. Can you convince me, why this is the best way to do this?

afkafkafk13 (commented):
Upt to now, I looked at it more closely as far as Multiindices and ExteriorPowers are concerned. (I will look at the de Rham complexes in the next days.)

The overall design looks good to me. 

I left a few comments -- nothing which changes the structure, but some parts are hard to read or keep track of, until they are documented.

HechtiDerLachs:
@afkafkafk13 : Sorry, but everything you have reviewed has already been merged in #2879 . I rebased this one now to exhibit the truely new parts.

HechtiDerLachs:
Probably we have to delete the `runtests.jl` again. I'll look into that on Monday.

afkafkafk13 (commented):
I now also looked at the deRham part of the orignal PR, which is the only part left after the rebase.  It looks good to me.

As mentioned before: A bit of documentation would be welcome.

src/Modules/deRhamComplexes.jl
```diff
@@ -0,0 +1,186 @@
+# By default we cache the Kaehler differentials and their 
+# exterior powers in the attributes of the ring. This 
+# internal function maintains that structure.
+function _kaehler_differentials(R::Ring)
+  if !has_attribute(R, :kaehler_differentials)
+    set_attribute!(R, :kaehler_differentials, Dict{Int, ModuleFP}())
```
simonbrandhorst, line 6: should these be weak dicts to allow garbage collection?

src/Modules/deRhamComplexes.jl
```diff
@@ -0,0 +1,186 @@
...
+  return ComplexOfMorphisms(typeof(Omega[1]), res_maps, typ=:cochain, seed=0, check=false)
+end
+
+# printing of kaehler differentials
+function show_kaehler_differentials(io::IO, M::ModuleFP)
+  success, F, p = is_exterior_power(M)
+  R = base_ring(F)
+  success && print(io, "Ω^$p($R)")
+  !success && print(io, "Ω^1($R)")
+end
```
simonbrandhorst: what is the purpose of this function, i.e. where is it called from?
Also please check whether unicode printing is allowed.

src/Modules/deRhamComplexes.jl
HechtiDerLachs: They could probably be.

HechtiDerLachs: No, they can't. The keys are integers. Weak dictionaries only work when loosing the key implies loosing all objects associated to the value.

src/Modules/deRhamComplexes.jl
HechtiDerLachs: It's the print function for modules which are Kaehler differentials. It is set as an attribute in the respective modules just like for the exterior powers and tensor products.

HechtiDerLachs:
From my side this is good to go. I think, I addressed all requests made in the discussion, didn't I?

src/Modules/ExteriorPowers/Generic.jl
HechtiDerLachs: How else to do it? I know it's hard to read, but I would have to write an internal initialization function otherwise. Would that make it clearer? Either way: That part has already been merged. So any discussion is about "how to do it next time".

afkafkafk13 (approved):
Thank you for the work you are putting into this.

PR #2898, opened 2023-10-09
GIT-Fans: compute the symmetry group of the Q-matrix

ThomasBreuer:
- Add `subgroup_property` to Oscar.

- Use `subgroup_property` in the GIT-Fans stuff for computing the symmetry group of the Q-matrix.
  (@fieker had suggested to compute the symmetries instead of just prescribing and verifying them.)

- Distinguish between the matrix `Qint` of integers (which is needed for prescribing the grading) and the matrix `Q` of type `QQMatrix` (which is needed everywhere else).

- Require the type `QQMatrix` instead of `Matrix{Int}` for the argument `Q` of `orbit_cones`, `action_on_target`, `git_fan`.

- Change the input syntax of the matrix `Q` such that rows are separated by semicolons.
  (Julia admits the syntax where line breaks separate rows, but this has caused already some confusion.)

Some thoughts about further steps:

- @jankoboehm Is "Q-matrix" the correct technical term, and what would be a better function name than `symmetry_group_q_matrix`?

- Instead of computing just the permutation group, we could in fact regard the isomorphism from this group to the corresponding matrix group (that is acting from the right) as "the symmetry group" of the Q-matrix; this way, we could get rid of the function `action_on_target` whose name was already criticized.

  (With a permutation group `G` as an optional second argument, `symmetry_group_q_matrix(Q, G)` could then just compute the isomorphism from `G`, without trying to compute more symmetries.)

- We could change `git_fan` such that the symmetry group `G` is an optional argument, since we can compute the full symmetry group of `Q`.

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1311,13 +1311,15 @@ end
 ################################################################################
 
 """
-    subgroup_property(prop::Function, G:: PermGroup)
+    subgroup_by_property(prop::Function, G::PermGroup)
 
 Return the subgroup of all those elements `g` in `G` for which `prop(g)`
 is `true`.
 
 For that, `prop(g)` must be `true` or `false` for each `g` in `G`,
 and the set of those `g` with `prop(g) == true` must form a group.
+However, this is *not* checked and if violated, results of operations
+with the resulting "group" object are unpredictable.
 
 # Examples
 ```jldoctest
@@ -1326,11 +1328,11 @@ julia> G = symmetric_group(4);  x = gen(G, 2)
 
 julia> prop = g -> g*x == x*g;  # centralizer of x
 
-julia> describe(subgroup_property(prop, G)[1])
+julia> describe(subgroup_by_property(prop, G)[1])
 "C2 x C2"
 ```
 """
-function subgroup_property(prop::Function, G:: PermGroup)
+function subgroup_by_property(prop::Function, G::PermGroup)
    gapprop = GAP.WrapJuliaFunc(gapelm -> prop(group_element(G, gapelm)))
    S = GAPWrap.SubgroupProperty(G.X, gapprop)
    return _as_subgroup(G, S)
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1306,6 +1306,39 @@ end
 
 ################################################################################
 #
+# Subgroups defined by properties
+#
+################################################################################
+
+"""
+    subgroup_property(prop::Function, G:: PermGroup)
+
+Return the subgroup of all those elements `g` in `G` for which `prop(g)`
+is `true`.
+
+For that, `prop(g)` must be `true` or `false` for each `g` in `G`,
+and the set of those `g` with `prop(g) == true` must form a group.
+
+# Examples
+```jldoctest
+julia> G = symmetric_group(4);  x = gen(G, 2)
+(1,2)
+
+julia> prop = g -> g*x == x*g;  # centralizer of x
+
+julia> describe(subgroup_property(prop, G)[1])
+"C2 x C2"
+```
+"""
+function subgroup_property(prop::Function, G:: PermGroup)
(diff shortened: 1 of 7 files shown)


fieker:
On Mon, Oct 09, 2023 at 02:47:10AM -0700, Thomas Breuer wrote:
> - Add `subgroup_property` to Oscar.
> 
> - Use `subgroup_property` in the GIT-Fans stuff for computing the symmetry group of the Q-matrix.
>   ***@***.*** had suggested to compute the symmetries instead of just prescribing and verifying them.)
> 
> - Distinguish between the matrix `Qint` of integers (which is needed for prescribing the grading) and the matrix `Q` of type `QQMatrix` (which is needed everywhere else).
... add an interface to allow rational gradings? if this is important?

ThomasBreuer:
@fieker

> ... add an interface to allow rational gradings? if this is important?

 I do not think so.
The matrix must be entered into the Julia session in some way, and then we can use this matrix of integers for the grading.
Fine.
Later we want to use the matrix for linear algebra over the Rationals, and then it is better to convert the matrix once and to use the result as input, instead of converting several times inside the functions.

fieker:
On Mon, Oct 09, 2023 at 02:57:04AM -0700, Thomas Breuer wrote:
> @fieker
> 
> > ... add an interface to allow rational gradings? if this is important?
> 
>  I do not think so.
> The matrix must be entered into the Julia session in some way, and then we can use this matrix of integers for the grading.
> Fine.
> Later we want to use the matrix for linear algebra over the Rationals, and then it is better to convert the matrix once and to use the result as input, instead of converting several times inside the functions.
> 
Good.

src/Groups/GAPGroups.jl
```diff
@@ -1304,6 +1304,39 @@ an exception is thrown if `G` is not solvable.
...
+################################################################################
+
+"""
+    subgroup_property(prop::Function, G:: PermGroup)
+
+Return the subgroup of all those elements `g` in `G` for which `prop(g)`
+is `true`.
+
+For that, `prop(g)` must be `true` or `false` for each `g` in `G`,
+and the set of those `g` with `prop(g) == true` must form a group.
```
fingolfin, line 1320:
```suggestion
and the set of those `g` with `prop(g) == true` must form a group.
However, this is *not* checked and if violated, results of operations
with the resulting "group" object are unpredictable.
```

src/Groups/GAPGroups.jl
```diff
@@ -1304,6 +1304,39 @@ an exception is thrown if `G` is not solvable.
...
 
 
+################################################################################
+#
+# Subgroups defined by properties
+#
+################################################################################
+
+"""
+    subgroup_property(prop::Function, G:: PermGroup)
```
fingolfin: I find this name confusing (also in GAP). Maybe `subgroup_by_property`? Though even then it somewhat clashes with the name `sub` (not *subgroup*). But `sub_by_property` doesn't really sound better. Ah well.

src/Groups/GAPGroups.jl
```diff
@@ -1304,6 +1304,39 @@ an exception is thrown if `G` is not solvable.
...
+julia> G = symmetric_group(4);  x = gen(G, 2)
+(1,2)
+
+julia> prop = g -> g*x == x*g;  # centralizer of x
+
+julia> describe(subgroup_property(prop, G)[1])
+"C2 x C2"
+```
+"""
+function subgroup_property(prop::Function, G:: PermGroup)
```
```suggestion
function subgroup_property(prop::Function, G::PermGroup)
```

src/Groups/GAPGroups.jl
ThomasBreuer: O.k.

src/Groups/GAPGroups.jl
ThomasBreuer: O.k.

The proposed statement differs from the one for `GAP.Globals.SubgroupProperty`.
If `prop` does not define a group then the result may depend on the chosen base, thus the result is unpredictable.
However, do we have examples where the returned object is corrupted as a group object?
If yes then the corresponding warning in the GAP Reference Manual should be strengthened.

PR #2901, opened 2023-10-09
LieAlgebras: Restrict to work over fields only 

lgoettgens:
For some functions already defined, I have doubts about the correctness over general rings, in particular in the existence of zero divisors.
So, I would like to restrict everything to work only over fields for now. If someone (maybe @fingolfin ?) with the required knowledge wants to work with Lie rings in the future, after a per-case decision one can widen the signatures again.


fingolfin (approved):
Fine by me.

ThomasBreuer (approved):
Fine.
(I guess that changing the condition from `RingElement` not to `FieldElement` but to `FieldElem` is intended.)

lgoettgens:
> Fine. (I guess that changing the condition from `RingElement` not to `FieldElement` but to `FieldElem` is intended.)

Yes, as I e.g. do not wanna have a `LieAlgebra{Int}` anyway. For scalar multiplication, integer scalar types are still allowed.

PR #2905, opened 2023-10-09
Chow rings of matroids

bschroter:
This commit removes an unnecessary helper function and adds a function which gives the user the volume map of the chow ring.


src/Combinatorics/Matroids/ChowRings.jl
```diff
@@ -210,11 +210,35 @@ function augmented_quadratic_relations(ring::MPolyRing, proper_flats::Vector{Vec
...
 """
-A helper function to select indices of a vector that do `include` elements of a given set and `exclude` another
-"""
-function _select(include::Union{AbstractVector,Set},exclude::Union{AbstractVector,Set},set::Union{AbstractVector,Set})
-    all = union(set...)
-    compl = setdiff(all,exclude)
-    return findall(s->issubset(include,s)&&issubset(s,compl),set);
+function volume_map(M::Matroid, A::MPolyQuoRing)
+    flat = flats(M)[1];
+    prod = one(A);
```
fingolfin:
```suggestion
    flat = flats(M)[1]
    prod = one(A)
```

src/Combinatorics/Matroids/ChowRings.jl
```diff
@@ -210,11 +210,35 @@ function augmented_quadratic_relations(ring::MPolyRing, proper_flats::Vector{Vec
...
+    for i in 2:length(flats(M))-1
+        if(is_subset(flat,flats(M)[i]))
```
fingolfin:
```suggestion
        if is_subset(flat, flats(M)[i])
```

src/Combinatorics/Matroids/ChowRings.jl
```diff
@@ -210,11 +210,35 @@ function augmented_quadratic_relations(ring::MPolyRing, proper_flats::Vector{Vec
...
+            flat = flats(M)[i]
+            prod*= A[i-1]
+        end
+    end
+    volume_map(f) = is_zero(f) ? 0//coeff(prod.f,1) : coeff(f.f,1)//coeff(prod.f,1);
```
fingolfin: This looks wrong: you are defining a method for `volume_map` inside a method for `volume_map`. Did you perhaps mean
```suggestion
    return f -> is_zero(f) ? 0//coeff(prod.f,1) : coeff(f.f,1)//coeff(prod.f,1)
```
Although that still seems weird... It definitely is not good to directly poke into the innards of these `MPolyQuoRingElem` instances `f.f` and `prod.f`. It is not even clear to me from what I see here whether this is computation is well-defined? After all, `f.f` and `prod.f` are just representatives... presumably their degree 1 term is well-defined if one knows what ideal is being divided out?

bschroter: This map should only be applied to the top degree component which is one dimensional. Thus for this component the map is well-defined.

src/Combinatorics/Matroids/ChowRings.jl
```diff
@@ -210,11 +210,35 @@ function augmented_quadratic_relations(ring::MPolyRing, proper_flats::Vector{Vec
...
-function _select(include::Union{AbstractVector,Set},exclude::Union{AbstractVector,Set},set::Union{AbstractVector,Set})
-    all = union(set...)
-    compl = setdiff(all,exclude)
-    return findall(s->issubset(include,s)&&issubset(s,compl),set);
+function volume_map(M::Matroid, A::MPolyQuoRing)
+    flat = flats(M)[1]
+    prod = one(A)
+    for i in 2:length(flats(M))-1
+        if is_subset(flat, flats(M)[i])
+            flat = flats(M)[i]
```
benlorenz:
```suggestion
    mflats = flats(M)
    flat = mflats[1]
    prod = one(A)
    for i in 2:length(mflats)-1
        if is_subset(flat, mflats[i])
            flat = mflats[i]
```
Storing the flats should avoid several unnecessary conversions from polymake data.

src/Combinatorics/Matroids/ChowRings.jl
```diff
@@ -210,11 +210,35 @@ function augmented_quadratic_relations(ring::MPolyRing, proper_flats::Vector{Vec
...
+            prod*= A[i-1]
+        end
+    end
+    return f -> is_zero(f) ? 0//coeff(prod.f,1) : coeff(f.f,1)//coeff(prod.f,1)
```
benlorenz:
```suggestion
    return f -> is_zero(f) ? 0//coeff(lift(prod),1) : coeff(lift(f),1)//coeff(lift(prod),1)
```
Even though using the `.f` here might be well defined (which I cannot tell), I think it would be better to use some kind of accessor function instead of using the field directly.
While undocumented for this type, there is even a function which does exactly what you need:
https://github.com/oscar-system/Oscar.jl/blob/a31ce6c23cf5407265d371ec09cf71fa9823fc68/src/Rings/MPolyQuo.jl#L1014
(and `is_zero(f)` will do a `simplify` anyway)

bschroter: Thank you. I didn't knew the lift function exists

fingolfin (approved):
Much better, thanks!

fingolfin:
Hmm. CI fails... Will try restarting it...

bschroter:
Thank you all for your help (and for restarting the test often enough)

PR #2906, opened 2023-10-09
Implement flattenings of graded polynomial rings

HechtiDerLachs:
The difference to what we have is that we get graded rings again and they have the correct default ordering.


src/Rings/MPolyMap/AffineAlgebras.jl
```diff
@@ -5,9 +5,10 @@
 ################################################################################
 
 const AffAlgHom = MPolyAnyMap{DT, CT, Nothing} where {T <: FieldElem,
-                                                       U <: MPolyRingElem{T},
-                                                       DT <: Union{MPolyRing{T}, MPolyQuoRing{U}},
-                                                       CT <: Union{MPolyRing{T}, MPolyQuoRing{U}}}
+                                                       U1 <: MPolyRingElem{T},
+                                                       U2 <: MPolyRingElem{T}, # types in domain an codomain might differ: one might be decorated, the other not.
+                                                       DT <: Union{MPolyRing{T}, MPolyQuoRing{U1}},
+                                                       CT <: Union{MPolyRing{T}, MPolyQuoRing{U2}}}
```
lgoettgens, line 11:
```suggestion
                                                       DT <: Union{MPolyRing{T}, MPolyQuoRing{<:MPolyRingElem{T}}},
                                                       CT <: Union{MPolyRing{T}, MPolyQuoRing{<:MPolyRingElem{T}}}}
```
For the case of different Us, there is no need to specify them concretely.

HechtiDerLachs: Yes, seems reasonable. I just preferred to stay close to what I found from before, also to not confuse the original authors unnecessarily much in case they want to look and confirm. Either way is fine with me, but personally I wouldn't bother to change this.

src/Modules/ModulesGraded.jl
```diff
@@ -2688,3 +2688,43 @@ function twist(M::ModuleFP{T}, d::IntegerUnion) where {T<:MPolyDecRingElem}
   @assert is_z_graded(R)
   return twist(M, grading_group(R)([d]))
 end
+
+# TODO: implement further base changes. But for this we need more functionality 
+# for potential change of grading groups. See the open pr #2677.
+function change_base_ring(
```
HechtiDerLachs, line 2663: This actually became quite relevant for me in this PR: `change_base_ring` for graded modules and translating the gradings. For special cases where I know the gradings don't change, you can customize this. But in general, I was forced to introduce an internal helper routine below. Maybe this can one day be used in a front-end method as a workhorse for the case of preserved gradings.

src/Modules/ModulesGraded.jl
```diff
@@ -2688,3 +2688,43 @@ function twist(M::ModuleFP{T}, d::IntegerUnion) where {T<:MPolyDecRingElem}
...
+  R = base_ring(M)
+  FS, mapF, iso_inv = f(F) # Makes sure ambient modules are preserved due to caching in flattenings
+  g = ambient_representatives_generators(M)
+  rels = relations(M)
+  MS = SubquoModule(FS, mapF.(g), mapF.(rels))
+  map = SubQuoHom(M, MS, gens(MS), f)
+  return MS, map
+end
+
+function _regularity_bound(M::SubquoModule)
```
HechtiDerLachs, line 2687: I will need this for the algorithm to compute derived pushforwards which I described in [my article here](https://link.springer.com/content/pdf/10.1007/s00208-021-02223-5.pdf).

src/Modules/flattenings.jl
```diff
@@ -0,0 +1,100 @@
...
+    set_attribute!(iso, :inverse, iso_inv)
+    set_attribute!(iso_inv, :inverse, iso)
+    flat_counterparts(flat_map)[F] = (F_flat, iso, iso_inv)
+  end
+    
+  F_flat, iso, iso_inv = flat_counterparts(flat_map)[F]::Tuple{<:ModuleFP, <:ModuleFPHom, <:ModuleFPHom}
+  return F_flat, iso, iso_inv
+end
+
+function flatten(
```
HechtiDerLachs, line 21: I pursued the following approach: There is one user facing method `flatten` that can be called on any reasonable object. The user does not need to care about which flattening is being used and where it is cached, etc. Internally, a `RingFlattening` of `R` is created and stored in the ring `R`. This one is called by default in the `flatten` methods and it maintains all associated flattened objects like ideals, modules, etc. The internal user/programmer can or rather should use a specified flattening and not the cached one when coding. 

Does that make sense?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -35,6 +37,30 @@
                                                     )
   end
 
+  function RingFlattening(
```
HechtiDerLachs, line 40: There's a lot of code duplication here for the graded cases. But I didn't find how to avoid this, really. Maybe the people involved in the development of graded stuff have some ideas?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -244,7 +387,12 @@ end
 
 ### Getters 
 function (phi::RingFlattening)(x::RingElem)
-  return phi.orig_to_flat(x)
+  if !haskey(flat_counterparts(phi), x)
+    result = phi.orig_to_flat(x)
+    flat_counterparts(phi)[x] = result
+    return result
+  end
+ return flat_counterparts(phi)[x]::elem_type(codomain(phi))
```
HechtiDerLachs, line 395: I'm not 100% sure that doing this is really performant. And maybe we should also find a way to turn it off, e.g. using keyword arguments. Apart from that: @fingolfin , do you think this `flat_counterpart` architecture that I set up here is a safe and correct application of the `WeakKeyIdDict`s? I will link again to this below.

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -263,6 +411,30 @@ function map_from_coefficient_ring_to_flattening(phi::RingFlattening)
   return phi.base_ring_to_flat
 end
 
+### flat counterparts of algebraic objects defined over the domain
```
HechtiDerLachs, line 414: Here I explain the philosophy of caching the `flat_counterpart`s. One issue might be that everything with cached flat counterpart must be "mathematically immutable". I think that's more or less safe in Oscar, but one probably has to be careful with things like in-place operations on matrices. Does anyone see problems here?

src/Rings/mpoly-localizations.jl
```diff
@@ -3185,3 +3185,24 @@ end
 
 dim(R::MPolyLocRing{<:Field, <:FieldElem, <:MPolyRing, <:MPolyElem, <:MPolyComplementOfPrimeIdeal}) = nvars(base_ring(R)) - dim(prime_ideal(inverted_set(R)))
 
+########################################################################
+# Localizations of graded rings                                        #
+########################################################################
```
HechtiDerLachs, line 3198: I was happily surprised that localizing graded rings does not seem to pose any problems so far. The code is sufficiently generic so that it just works. But I am wondering whether there are some more functions which we should provide for such localizations, see below.

test/Rings/MPolyAnyMap/AffineAlgebras.jl
```diff
@@ -66,3 +66,12 @@
     @test f(x) == x
   end
 end
+
+@testset "cross-type kernels" begin
```
HechtiDerLachs, line 70: This is the test corresponding to my fix of the `AffAlgHom`.

test/Rings/MPolyAnyMap/flattenings.jl
```diff
@@ -117,3 +115,198 @@ end
...
+  flat = Oscar.flatten(S)
+
+  S_flat = codomain(flat)
+
+  F = graded_free_module(S, [-2, 1])
+  F_flat, iso, iso_inv = Oscar.flatten(F)
+
+  M, _ = quo(F, [u^5*F[1], v*F[1]])
+  M_flat, iso, iso_inv = Oscar.flatten(M)
+  res = free_resolution(M)
```
HechtiDerLachs, line 297: @jankoboehm : Could you maybe have a look at this and tell me whether I implemented this custom free resolution in accordance with how you guys designed this? Maybe just run the code of this test in the REPL once, or something?

HechtiDerLachs (commented):
I left some comments and questions about the code and would be happy to discuss these at some point.

HechtiDerLachs:
I will add a `cached=true/false` option to the mapping of elements. Otherwise, there seem to be no real objections to merging this from @fingolfin 's side, if I understand correctly.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
```diff
@@ -399,13 +399,13 @@ Projective space of dimension 2
 with homogeneous coordinates [x, y, z]
 
 julia> homogeneous_coordinates_on_affine_cone(P)
-3-element Vector{MPolyQuoRingElem{QQMPolyRingElem}}:
+3-element Vector{MPolyQuoRingElem{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}}:
  x
```
afkafkafk13, line 403: Not concerning this PR, but more generally: we need to rethink the printing here. The way it is now, it does not have any added value for the user, whereas it is a lot better in line 399.

src/Modules/flattenings.jl
afkafkafk13: In other words: flat_counterparts is some dictionary which encodes the correspondence between non-flattened and flattened objects -- to me the idea makes sense (provided this does not eat up too much space)

test/Rings/MPolyAnyMap/flattenings.jl
```diff
@@ -117,3 +115,198 @@ end
   @test Oscar.map_from_coefficient_ring_to_flattening(phi).(gens(R)) == phi.(Q.(S.(gens(R))))
 end
 
+@testset "flattenings of graded rings" begin
+  # towers of polynomial rings
+  R, (x, y) = polynomial_ring(QQ, [:x, :y], cached=false)
+  S, (u, v) = grade(polynomial_ring(R, [:u, :v], cached=false)[1])
+
```
afkafkafk13, line 122: Does this creation of S also work with graded_polynomial_ring? 
If so, it might be appropriate to do this. (This is the case I expect)
If not, @fingolfin would certainly like to hear about the problem.

afkafkafk13 (commented):
Overall this looks good. I left you a few minor remarks in the code.

Just a remark on wording: 
In commutative algebra 'flat' is of course a property of a module,
here you use it in the sense of 'flattened'. You might want to mention that somewhere in the file.
(By the way: I fully accept this choice for better readability of the code)

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Attributes.jl
simonbrandhorst: We cannot change the printing of julia Arrays.

simonbrandhorst (approved):
Looks good to me.

PR #2909, opened 2023-10-10
UngradedModules: don't use display(...) in special show methods

benlorenz:
From the help of `display(...)`:

> In general, you cannot assume that `display` output goes to `stdout`...

But all the text surrounding that `display` call goes to `stdout`.

Also it doesn't get captured by Documenter which is used for checking the code for the oscar-book.

The new show call seems to replicate the previous output:
```
julia> show_subquo(H)
Graded subquotient of
[1   0   0   0]
[0   0   0   1]
by image of
[x     0   0     0]
[0   y^2   0     0]
[0     0   x     0]
[0     0   0   y^2]
with ambient free module Rg^4
```

Anyone who knows more about all the pretty-printing features please feel free to add suggestions.

### DIFF
--- a/src/Modules/UngradedModules.jl
+++ b/src/Modules/UngradedModules.jl
@@ -1966,16 +1966,16 @@ function show_subquo(SQ::SubquoModule)
       else
         println("Cokernel of")
       end
-      display(generator_matrix(SQ.quo))
+      show(stdout, "text/plain", generator_matrix(SQ.quo))
     else
       if is_graded(SQ)
         println("Graded subquotient of")
       else
         println("Subquotient of")
       end
-      display(generator_matrix(SQ.sub))
-      println("by image of")
-      display(generator_matrix(SQ.quo))
+      show(stdout, "text/plain", generator_matrix(SQ.sub))
+      println("\nby image of")
+      show(stdout, "text/plain", generator_matrix(SQ.quo))
     end
   else
     if is_graded(SQ)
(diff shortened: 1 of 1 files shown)


fingolfin:
As far as I can tell, all these methods `show_subquo` , `show_morphism_as_map`, `show_morphism` are not called anywhere. It is not clear to me what they are meant to be used for, and by whom. As they are right now, I think they are best internal debugging helpers?

Perhaps @jankoboehm can clarify.

benlorenz:
They are exported and `show_subquo` even has a docstring. `show_subquo` is used in the book chapter on algebraic geometry.

jankoboehm:
`show_subquo` prints a subquo in terms of a dense representation of the vectors and omits the indexing of the vectors i.e. in terms of a generator and relation matrix. This was requested by various people because in some settings this offers a shorter and better understandable presentation. We would like to keep that, but we can rename it or make it available in a different way. The analogous thing is `show_morphism`. 

The function  `show_morphism_as_map` is historical. Remember that a morphism then only printed domain and codomain. This function is what later became the standard way of printing morphisms. So it should be removed.

lkastner (approved):

benlorenz:
I will merge this now as this helps with the Oscar book testing and the new code just uses more appropriate methods for displaying the matrices. The output should be identical.
Feel free to make a new PR to clean up the historical method and/or reorganize this.

PR #2912, opened 2023-10-11
Update to GAP.jl 0.10

fingolfin:
It redefines hash(::GapObj) to produce an error, hence is a breaking update.

Let's see if anything breaks...


lgoettgens (approved):
That looks (surprisingly) good!
And I am happy that the removed conversions in https://github.com/oscar-system/GAP.jl/pull/936 didn't break anything in Oscar.

PR #2913, opened 2023-10-11
Rename `mul` to `*`

lgoettgens:
Oscar side of https://github.com/thofma/Hecke.jl/pull/1243


benlorenz:
The corresponding Hecke version was released and the deprecation broke the Oscar CI.
I merged master to fix the conflicts and let CI run, so we can hopefully fix the CI soon.

lgoettgens:
Thanks for working on this! I restarted all ci jobs that failed (mostly due to overly long build steps). But overall, it seems to be fine after your change.

thofma:
@benlorenz is there something we need to change in the downstream tests to catch this? I am pretty sure they showed success.

lgoettgens:
> @benlorenz is there something we need to change in the downstream tests to catch this? I am pretty sure they showed success.

This failure is expected with the current setup in Oscar. I created #2956 as a try to change that for the future.

thofma:
It is fine that Oscar expects this. It is not fine, that this is not catched in downstream tests. How does #2956 fix this? Shouldn't the fix involve https://github.com/oscar-system/OscarDevTools.jl?

benlorenz:
> It is fine that Oscar expects this. It is not fine, that this is not catched in downstream tests. How does #2956 fix this? 

It doesn't strictly fix this it just avoids breaking the whole CI and instead shows just one failing job when Oscar triggers deprecations.
Using deprecated functions should not block us from merging other PRs, especially since the deprecations might not even come from code in the Oscar ecosystem (last time we discussed this it came from JLLWrappers).

> Shouldn't the fix involve [oscar-system/OscarDevTools.jl](https://github.com/oscar-system/OscarDevTools.jl)?

We could enable depwarn=error there as well but I am not sure if this is the best approach, as explained above.
Fixing deprecations is not really something that needs to be done immediately.

thofma (approved):

PR #2918, opened 2023-10-14
Rename .bib item for ATLAS

fingolfin:


thofma:
Is there a problem with WWTSPNNLBA? I thought everybody knew that this is the ATLAS.

lgoettgens:
Can we wait with this until #2605 is merged? The conflict will be much easier to handle here instead of there.

fingolfin:
I agree 100% this should wait for #2605 !

@thofma err.... I am not quite sure to respond; are you serious?

thofma:
No, it was a (poor) attempt at humour.

lgoettgens (approved):

ThomasBreuer (approved):
No problem with this change.
(The original idea was that contrary to LaTeX, Documenter.jl does not create labels for showing the bib entries but just uses the given programmatic labels.)

PR #2919, opened 2023-10-16
Add some more tests to the bibtool CI job

lgoettgens:
Most of this is either a simple copy-paste from their manual, or it was needed to get #2605 running/looking nice. Thus, I would like to keep the .bib file in its current nice format. Some easy-to-check things are added to the `.bibtoolrsc` to add them to the CI job.

### DIFF
--- a/.github/workflows/BibtoolCI.yml
+++ b/.github/workflows/BibtoolCI.yml
@@ -25,7 +25,9 @@ jobs:
           sudo apt-get install -y bibtool
       - name: Execute bibtool and check for changes
         run: |
-          bibtool docs/oscar_references.bib -o docs/oscar_references.bib
+          bibtool docs/oscar_references.bib -o docs/oscar_references.bib 2>&1 | tee bibtool.log
+          test \! -s bibtool.log
+          rm bibtool.log
           if [ -n "$(git status --porcelain)" ]; then
             echo "We employ bibtool to standardize the bibtex entries of our reference list."
             echo "Your changes to the bibliography do not follow this standard."
--- a/.bibtoolrsc
+++ b/.bibtoolrsc
@@ -62,6 +62,10 @@ key.format =
     { %2d(year) }
 }
 
+# Remove empty fields
+rewrite.rule {"^\" *\"$"}
+rewrite.rule {"^{ *}$"}
+
 # Check that 1800<=year<=2029
 check.rule { year "^[\"{]1[89][0-9][0-9][\"}]$" }
 check.rule { year "^[\"{]20[0-2][0-9][\"}]$" }
--- a/.bibtoolrsc
+++ b/.bibtoolrsc
@@ -79,3 +79,6 @@ check.error.rule { year "" "\@ \$: Year has to be a suitable number" }
 
 # Check that the doi field is not a URL
 check.error.rule { doi "\://" "\@ \$: doi field should not be a URL" }
+
+# Check that the url field is not a DOI
+check.error.rule { url "doi\.org/" "\@ \$: url field should not contain a doi. Use the doi field instead" }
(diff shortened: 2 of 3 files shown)


.github/workflows/BibtoolCI.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           sudo apt-get install -y bibtool
       - name: Execute bibtool and check for changes
         run: |
-          bibtool docs/oscar_references.bib -o docs/oscar_references.bib
+          bibtool docs/oscar_references.bib -o docs/oscar_references.bib |& bibtool.log
```
benlorenz: This is not working (`line 1: bibtool.log: command not found`), either `tee` is missing or it should have been `>&`? But I don't understand what this is supposed to do, that log file is not used anyway?

lgoettgens: This is supposed to write stdout and stderr both to the console and to a file. Console so it is visible in the log. File to make the `git status` below fail in case that bibtool printed something.
I guess this syntax is only supported by my zsh locally and not the shell on github actions. I'll try to change this

benlorenz: I see, so you want it to fail if bibtool prints any warning?
But shell redirection and `|& tee something.log` will always create the file, it might stay empty if nothing is printed. So the next check would always fail.

fingolfin: Tests will already fail if there is any change to the `.bib` file. What other log message that neither correspond to a change in the `.bib` file nor to a non-zero exit code is this meant to catch?

fingolfin (approved):
I like, it very nice -- but of course the issue pointed out by @benlorenz needs to be addressed.

Note: GH Actions by default use `bash` for shell actions; this can in principle be changed to use any other shell including `zsh`, but I'd rather not, as that would probably lead to even more confusion down the road...

.github/workflows/BibtoolCI.yml
lgoettgens: Unfortunately, the `check.error.rule` only print an error message to stderr, but the exit code is still 0. If you have a better idea than writing stderr to a file and check that file for being empty, please let me know.

lgoettgens: This should now be fixed. It first commits the empty `bibtool.log` file, then runs bibtool with pipes and afterwards checks `bibtool.log` for changes

benlorenz (approved):

.github/workflows/BibtoolCI.yml
benlorenz: Still looks somewhat weird but I guess that is ok as long as it works.

.github/workflows/BibtoolCI.yml
```diff
@@ -25,7 +25,12 @@ jobs:
           sudo apt-get install -y bibtool
       - name: Execute bibtool and check for changes
         run: |
-          bibtool docs/oscar_references.bib -o docs/oscar_references.bib
+          git config user.name "Bibtool CI Job"
+          git config user.email "bibtool.ci.job@noreply.github.com"
+          touch bibtool.log
+          git add bibtool.log
+          git commit -m "Add bibtool.log"
+          bibtool docs/oscar_references.bib -o docs/oscar_references.bib 2>&1 | tee bibtool.log
```
fingolfin: If all you want to do is to check whether the log file is empty, you can also do it like this:
```suggestion
          bibtool docs/oscar_references.bib -o docs/oscar_references.bib 2>&1 | tee bibtool.log
          test -s bibtool.log
```

lgoettgens: This is missing a negation, but I now used something very similar

fingolfin (approved):

PR #1173, opened 2022-03-16
Purge AlgHom

thofma:
Waiting for a PR by Wolfram before this should be merged.

Issue #588, opened 2021-07-28
links to source code in the Oscar documentation


ThomasBreuer:
The Oscar documentation does currently not contain all links from docstrings to the source code.
The links seem to be missing for docstrings that belong to AA, Nemo, and Hecke.
(Note that the files in question get copied into the Oscar repository when the documentation gets built,
and then it is not clear to which addresses one should point for the source code.)

Is there an easy trick that would allow Documenter.jl to find the source code?

fingolfin:
I don't think Documenter.jl supports this at this time. And it'd be a bit difficult anyway, because it doesn't know which repository and which commit in that repository the code in question comes from. To be precise: I guess in principle it could be taught to deduced this information, at least in our specific case, but its code simply isn't prepared for this. The source link URLs are computed by this code:
```julia
            # When a source link is available then print the link.
            if !ctx.settings.disable_git
                url = Utilities.url(ctx.doc.internal.remote, ctx.doc.user.repo, result)
                if url !== nothing
                    push!(ret.nodes, a[".docs-sourcelink", :target=>"_blank", :href=>url]("source"))
                end
            end
```
So this delegates to the `Utilities.url` function; as you note, it already received the repo(sitory) as an argument. Looking at the declaration of the relevant `url()` methods, one finds this:
```julia
url(remote, repo, doc) = url(remote, repo, doc.data[:module], doc.data[:path], linerange(doc))

function url(remote, repo, mod, file, linerange)
    file === nothing && return nothing # needed on julia v0.6, see #689
    remote = getremote(dirname(file))
    isabspath(file) && isempty(remote) && isempty(repo) && return nothing
    ...
```

fingolfin:
Documenter.jl 1.0 was just released. Its changelog states among other things:
> In addition to generating source and edit links for the main repository, Documenter can now also be configured to generate correct links for cases where some files are from a different repository (e.g. with vendored dependencies).

I have no idea how that works exactly, but it might be worth it having another look at this.

lgoettgens:
> Documenter.jl 1.0 was just released. Its changelog states among other things:
> 
> > In addition to generating source and edit links for the main repository, Documenter can now also be configured to generate correct links for cases where some files are from a different repository (e.g. with vendored dependencies).
> 
> I have no idea how that works exactly, but it might be worth it having another look at this.

[...]

PR #2922, opened 2023-10-16
Old: Bump `Documenter.jl` to `v1.0`

lgoettgens:


src/Rings/affine-algebra-homs.jl
benlorenz, line 1: I think this file should be deleted (again) instead of changed, it was removed in #1173 re-added (probably accidentally?) in https://github.com/oscar-system/Oscar.jl/pull/1986/files#diff-fa3cdba51919e9867525731f50ece6ca127c1b3df70f997d5123f5439d6ee273, but it is not included at all.

lgoettgens: Can you do this in its own PR? That makes history easier to recap. I will then rebase this on top without the change.

benlorenz: #2923

lgoettgens:
Documenter.jl v1.0 has a lot more features for cross-referencing between docs of different packages, that can be very useful for our copy-pasting hack. But I skipped dealing with that / deactivated it for now and leave the fix to https://github.com/oscar-system/Oscar.jl/issues/588.

lgoettgens:
Closed in favor of https://github.com/oscar-system/Oscar.jl/pull/2927 (same repo -> we get a preview)

PR #2924, opened 2023-10-16
Use graded_polynomial_ring in yet more places

fingolfin:


afkafkafk13 (commented):
Either you replaced instances, which you did not want to replace or you are missing a signature or you got brackets wrong.
`no method matching graded_polynomial_ring(::QQField, ::Pair{String, UnitRange{Int64}}`

In any case: I approve of the general flavour of this change, but won't click approve before the tests pass.

afkafkafk13 (approved):

lkastner:
@fingolfin I deleted the failing tests. Hope that helps.

fingolfin:
@lkastner had helpfully added some more `graded_polynomial_ring` methods to this PR to make some of my tests work -- thank you. But I don't really want to do that in this PR, as we are working on revising that anyway. Instead I dropped the "bad" changes from `polynomial_ring` to `graded_polynomial_ring` in this PR. A later PR will add a new `graded_polynomial_ring` method which will support *all* variants of `polynomial_ring`, and more, and will also convert most remaining uses of `grade` to `graded_polynomial_ring` (including the ones I just deleted from this PR).

lkastner (approved):

test/Modules/ModulesGraded.jl
```diff
@@ -1030,8 +973,7 @@ function randpoly(R::Oscar.Ring,coeffs=0:9,max_exp=4,max_terms=8)
 end
 
 @testset "Basic degree and homogeneity tests for free modules" begin
-    Qx, x = polynomial_ring(QQ, 3)
-    R, x = grade(Qx)
+    R, x = graded_polynomial_ring(QQ, 3)
```
afkafkafk13, line 976: Maybe revert this change. The corresponding signature is not available and you might just resolve the problem with you future PR which will contain a unified solution.

fingolfin:
I've decided to mark this as draft, and wait till the `graded_polynomial_ring` update is ready, that will be less work in total, as this PR here is not urgent.

experimental/PlaneCurve/test/runtests.jl
```diff
@@ -196,8 +193,7 @@ end
 end
 
 @testset "ProjPlaneCurve int_multiplicity functions" begin
-    R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
-    T, _ = grade(R)
+    T, (x, y, z) = graded_polynomial_ring(QQ, ["x", "y", "z"])
     F = ProjPlaneCurve(T((x^2 + y^2) * (x^2 + y^2 + 2 * y * z)))
     G = ProjPlaneCurve(T((x^2 + y^2) * (y^3 * x^6 - y^6 * x^2 * z)))
```
afkafkafk13, line 198: Why are you still explicitly moving the polynomial to T -- it is already in T by definition

experimental/PlaneCurve/test/runtests.jl
```diff
@@ -216,8 +212,7 @@ end
 end
 
 @testset "ProjPlaneCurve singularity functions" begin
-    R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
-    T, _ = grade(R)
+    T, (x, y, z) = graded_polynomial_ring(QQ, ["x", "y", "z"])
     PP = proj_space(QQ, 2)
```
afkafkafk13, line 216: same here

experimental/PlaneCurve/test/runtests.jl
```diff
@@ -270,8 +265,7 @@ end
 end
 
 @testset "ProjCurveDivisor basic functions" begin
-    S, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
-    T, _ = grade(S)
+    T, (x, y, z) = graded_polynomial_ring(QQ, ["x", "y", "z"])
     C = ProjPlaneCurve(T(y^2 + y * z + x^2))
```
afkafkafk13, line 269: and here

afkafkafk13:
The 3 instances I mentioned above are very small nitpicks. I would only like to see them cleaned up, because users tend to use the tests as examples.

experimental/PlaneCurve/test/runtests.jl
fingolfin: Simple: I did most of the changes in this file with search & replace and regular expressions. It is much nastier to handle `T(arbitrary_expression)` correctly esp. if there are nested parenthesis. Arguably the code was strange *before* (why did it work with the non-graded variables in the first place?) Since @wdecker was never quite happy with it either, I am not sure this code will ever graduate out of the "experimental" directory. So my enthusiasm for putting time into cleaning it up further is limited...

afkafkafk13: ok. Makes sense.

PR #2925, opened 2023-10-16
Update: Functionality for matroid realization spaces

bschroter:
Larger contribution:

-- replaced `matroid_strata_grassmannian.jl` by `realization_space.jl`
refactoring and cleaning the code

(work by Dan and Lukas)


src/Combinatorics/Matroids/ChowRings.jl
```diff
@@ -57,7 +57,12 @@ function chow_ring(M::Matroid; ring::Union{MPolyRing,Nothing}=nothing, extended:
             #add the variables for the simplicial generators
             var_names = [var_names; ["h_{" * join(S, ",") * "}" for S in [proper_flats;[Flats[number_flats]]]]]
         end
-        ring, vars = polynomial_ring(QQ, var_names, cached=false)
+        
+        if graded
+            ring, vars = GradedPolynomialRing(QQ, var_names)
```
fingolfin:
```suggestion
            ring, vars = graded_polynomial_ring(QQ, var_names, cached=false)
```

src/Combinatorics/Matroids/JMatroids.jl
```diff
@@ -1,3 +1,4 @@
 include("matroids.jl")
 include("properties.jl")
 include("ChowRings.jl")
+include("realization_space.jl")
```
fingolfin:
```suggestion
include("realization_space.jl")

```
Please ensure every file ends with a newline

src/Combinatorics/Matroids/properties.jl
```diff
@@ -970,19 +973,37 @@ end
...
-Computes the revlex basis encoding and the minimal revlex basis encoding among isomorphic matroids 
+Computes the revlex basis encoding of the matroid M.
 
 # Examples
-To get the revlex basis encoding of the fano matroid and to preduce a matrod form the encoding write:
+To get the revlex basis encoding of the fano matroid and to produce a matrod form the encoding write:
 ```jldoctest
-julia> string1, string2 = revlex_basis_encoding(fano_matroid())
-("0******0******0***0******0*0**0****", "0******0******0***0******0*0**0****")
+julia> string = revlex_basis_encoding(fano_matroid())
```
fingolfin: I would recommend to not overwrite global functions like `string`, even in an example, just in case people copy&paste the example into their session, as then afterwards other examples which want to use the `string` function won't work anymore.

Just call it `str` or `s` or whatever.

Of course this also applies to other examples doing the same

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
+
+
+mutable struct MatroidRealizationSpace
```
fingolfin: The docs added in this PR state:

> Its *realization space* ``\mathcal{R}(M)`` is an affine scheme 

Note that we do have actual affine schemes as objects in Oscar, with all the usual property. But the realization space type defined here is *not* an affine scheme in that sense.

Now I don't say that's necessarily a problem, but it sure is potentially confusing. It may be useful to address this explicitly somewhere?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    defining_ideal::Union{Ideal, NumFieldOrdIdl, Nothing}
+    inequations::Union{Vector{Oscar.RingElem},Nothing}
+    ambient_ring::Union{Oscar.MPolyRing, Ring, Nothing}
+    realization_matrix::Union{Oscar.MatElem,Nothing}
+    realizable::Union{Bool,Nothing} # Nothing if we're not sure yet if M is realizable.
+    F::AbstractAlgebra.Ring
+    char::Union{Int,Nothing}
+    q::Union{Int,Nothing}
+    one_realization::Bool
+end
```
fingolfin: Hmm, no inner construct for such a complex type. Hence no information which information is guaranteed to be there, and which is optional (presumably everything that can also be `nothing`) and in which combination... *And* no comment with any hints about all that either...

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
+
+
+mutable struct MatroidRealizationSpace
+    defining_ideal::Union{Ideal, NumFieldOrdIdl, Nothing}
+    inequations::Union{Vector{Oscar.RingElem},Nothing}
```
fingolfin: You could just as well have this as
```suggestion
    inequations::Vector{Oscar.RingElem}
```
and then in the code use `isdefined(obj,:inequations)` instead of `!isnothing(obj.inequations)`. Same for other fields.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    ambient_ring::Union{Oscar.MPolyRing, Ring, Nothing}
```
fingolfin: Having both `MPolyRing` and `Ring` in there is kinda redundant... Also can't you drop the `Oscar.` here and elsewhere?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            print(io, "The matroid is not realizable over the specified field or characteristic.")
+        end
+    else
+        if RS.one_realization
+            println(io, "One realization is given by")
+        else
+            println(io, "The realizations are parametrized by")
+        end
+        # println isn't ideal as it prints the matrix as one big line
+        display(RS.realization_matrix)
```
fingolfin: This is wrong as it ignores the `io` object.

Please also take a look at the "Printing" section of the OSCAR Developer documentation under [https://docs.oscar-system.org](https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/#Implementing-show-functions)

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        println(io, "in the ", RS.ambient_ring)
+        I = RS.defining_ideal
+        if (typeof(I) <: NumFieldOrdIdl && I.gen != ZZ(0)) || (typeof(I) <: Ideal && !iszero(I))
```
fingolfin: Never write `typeof(a) <: B`, instead write `a isa B`

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            println(io, "within the vanishing set of the ideal\n", RS.defining_ideal)
+        end
+        if length(RS.inequations) > 0
+            println(io, "avoiding the zero loci of the polynomials\n", RS.inequations)
```
fingolfin: `show` methods should *not* end with a newline, so use `print` instead of `println` here and above.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        end
+    end
+end
+
+# Constructors
+function MatroidRealizationSpace(F::AbstractAlgebra.Ring, char::Union{Int,Nothing}, q::Union{Int,Nothing})
```
fingolfin: So what is the meaning of `char` and `q` and are they in any way related to each other or to `F`?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    return MatroidRealizationSpace(nothing, nothing, nothing, nothing, false, F, char, q, false)
+end
+
+function MatroidRealizationSpace(I::Union{Ideal, NumFieldOrdIdl, Nothing},
```
fingolfin: The above two constructors probably should be inner constructors. It is also weird to ask for arguments of type `nothing` like this -- it might make sense if you wish to provide default values for these arguments, but like this is it is strange.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    ineqs::Union{Vector{<:Oscar.RingElem},Nothing},
+    R::Union{Oscar.MPolyRing, Ring, Nothing}, mat::Union{Oscar.MatElem,Nothing},
+    F::AbstractAlgebra.Ring, char::Union{Int,Nothing}, q::Union{Int,Nothing})
+    return MatroidRealizationSpace(I, ineqs, R, mat, nothing, F, char, q, false)
+end
+
+@doc raw"""
+is_realizable(M; char::Union{Int,Nothing}=nothing, q::Union{Int,Nothing}=nothing)
```
fingolfin:
```suggestion
    is_realizable(M; char::Union{Int,Nothing}=nothing, q::Union{Int,Nothing}=nothing)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+* If char = nothing, then this function determines whether the matroid is realizable over some field.
+  
+* If char = 0,then this function determines whether the matroid is realizable over some field of
+    characteristic 0.
+  
+* If char = p is prime, this function determines whether the matroid is realizable
+    of characteristic p.
+  
+* if char = p and q is a power of p, this function determines whether the matroid is realizable over the
+    finite field ``GF(q)}``.
```
fingolfin:
```suggestion
    finite field ``GF(q)``.
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    F::AbstractAlgebra.Ring, char::Union{Int,Nothing}, q::Union{Int,Nothing})
+    return MatroidRealizationSpace(I, ineqs, R, mat, nothing, F, char, q, false)
+end
+
+@doc raw"""
+is_realizable(M; char::Union{Int,Nothing}=nothing, q::Union{Int,Nothing}=nothing)
+
+* If char = nothing, then this function determines whether the matroid is realizable over some field.
+  
+* If char = 0,then this function determines whether the matroid is realizable over some field of
```
fingolfin:
```suggestion
* If `char == 0`, then this function determines whether the matroid is realizable over some field of
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    characteristic 0.
+  
+* If char = p is prime, this function determines whether the matroid is realizable
+    of characteristic p.
+  
+* if char = p and q is a power of p, this function determines whether the matroid is realizable over the
```
fingolfin:
```suggestion
* If `char == p` and `q` is a power of `p`, this function determines whether the matroid is realizable over the
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    finite field ``GF(q)}``.
+"""
+function is_realizable(M::Matroid; char::Union{Int,Nothing}=nothing, q::Union{Int,Nothing}=nothing)::Bool
+    RS = realization_space(M, char=char, q=q)
+    return is_realizable(RS)
+end
+
+function is_realizable(RS::MatroidRealizationSpace)
+    !isnothing(RS.realizable) && return RS.realizable
```
fingolfin: This seems like you should turn make your `MatroidRealizationSpace` type into an [attribute storing type](https://nemocas.github.io/AbstractAlgebra.jl/dev/misc/#Attributes), and then annotate this function by `@attr`.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        return RS.realizable
+    end
+    for p in minimal_primes(RS.defining_ideal)
+        component_non_trivial = true
+        for f in RS.inequations
+            if f in p
+                component_non_trivial = false
+                break
+            end
+        end
```
fingolfin:
```suggestion
        component_non_trivial = all(f -> f in p, RS.inequations)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            return RS.realizable
+        end
+    end
+    RS.realizable = false
+    return RS.realizable
+end
+
+
+@doc raw"""
+defining_ideal(RS::MatroidRealizationSpace)
```
fingolfin:
```suggestion
    defining_ideal(RS::MatroidRealizationSpace)
```
Four spaces indentation here. Applies to many other docstrings, too.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+function find_good_basis_heuristically(M::Matroid)
+    bs = bases(M)
+    cs = circuits(M)
+    min_num_vars = length(cs)*rank(M)
+    min_basis = bs[1]
+    for bi in 1:length(bs)
+        current_num_vars = 0
+        for c in cs
+            for e in c
+                @inbounds if e in bs[bi]
```
fingolfin: Or just do `for b in bs` above and then use
```suggestion
                if e in b
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        if current_num_vars < min_num_vars
+            min_num_vars = current_num_vars
+            min_basis = bs[bi]
+        end
+    end
+    return min_basis
+end
+
+
+# Return the factors of f, but not the exponents
```
fingolfin: Maybe this would be a more fitting description?
```suggestion
# Return the prime divisors of f
```
And then rename the function accordingly, and also the next one?

@fieker @thofma would it make sense to have a function `prime_divisors` as companion for `factor` in general? We currently only offer it for integers but not for polynomials?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+end
+
+# Return the unique factors of the elements of Sgen, again no exponents. 
+function gens_2_factors(Sgens::Vector{<:RingElem})
+    return unique!(vcat([poly_2_factors(f) for f in Sgens]...))
+end
+
+
+function stepwise_saturation(I::MPolyIdeal, Sgens::Vector{<:RingElem})
+    foreach(f -> I = saturation(I,ideal([f])), Sgens)
```
fingolfin: Don't do that, that abuses `foreach` for something it wasn't meant to do at all. Either write a genuine `for` loop:
```suggestion
    for f in Sgens
        I = saturation(I, ideal([f]))
    end
```
Or use `reduce` resp. `mapreduce`... something like this (untested)
```suggestion
    return mapreduce(f->ideal([f]), saturation, Sgens; init=I)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            min_basis = bs[bi]
+        end
+    end
+    return min_basis
+end
+
+
+# Return the factors of f, but not the exponents
+function poly_2_factors(f::RingElem)
+    return collect(keys(Dict(factor(f))))
```
fingolfin: This creates a dictionary (rather expensive, allocates a lot). You can do without:
```suggestion
    return map(first, factor(f))
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+=#
+
+#####################
+# full reduction    #
+#####################
+
+# computes the coefficient of v in monomial m. 
+function coefficient_monomial(v::RingElem, m::RingElem)
+    isone(degree(m,v)) || return "The variable is not a degree 1 factor."
```
fingolfin: Should this be an `error` ? 

You could also use the `@req` macro

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    mf = factor(m)
+    mfdict = Dict(mf)
+    u = unit(mf)
+    not_v = [k^(mfdict[k]) for k in keys(mfdict) if k != v ]
```
fingolfin:
```suggestion
    u = unit(mf)
    not_v = [k^e for (k,e) in mf if k != v ]
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    length(not_v) == 0 ? u : u*prod(not_v)
+end
+
+# computes the coefficient of v in f. 
+function coefficient_v(v::RingElem, f::RingElem)
+    isone(degree(f,v)) || return "degree of variable must be 1"
```
fingolfin: Again: don't you mean to invoke `error`?

fingolfin: This occurs in multiple other places, please fix in all of them

src/Combinatorics/Matroids/realization_space.jl
fingolfin: There are more places that needlessly convert a `Fac` object into a `Dict`, please check them as well.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+                     R::Ring, xs::Vector{<:RingElem}) 
+    preSgens = unique!([sub_v(x, t, f, R, xs) for f in Sgens])
+    return gens_2_factors(preSgens)
+end
+
+function n_new_Igens(x::RingElem, t::RingElem, Igens::Vector{<:RingElem}, 
+                     Sgens::Vector{<:RingElem}, R::Ring, xs::Vector{<:RingElem}) 
+
+    preIgens = unique!([clean(sub_v(x, t, f, R, xs), R, Sgens) for f in Igens])
+    return filter(x-> x!= R(0), preIgens)
```
fingolfin:
```suggestion
    return filter(!iszero, preIgens)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+end
+
+
+
+function matrix_clear_den_in_col(X::Oscar.MatElem, c::Int)
+    Xc = [denominator(f) for f in X[:, c]]
+    t = lcm(Xc)
+    result = multiply_column!(X, t, c)
+    return result
+
```
fingolfin:
```suggestion
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+function reduce_ideal_one_step(MRS::MatroidRealizationSpace, 
+                               elim::Vector{<:RingElem}, 
+                               fullyReduced::Bool)
+
+    Igens = gens(MRS.defining_ideal)
+    Sgens = MRS.inequations
+    R = MRS.ambient_ring
+    FR = FractionField(R)
```
fingolfin:
```suggestion
    FR = fraction_field(R)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    xs = gens(R)
+    X = MRS.realization_matrix
+    nr, nc = size(X)
+    
+    Ivars = ideal_vars(Igens);
+
+    for x in Ivars 
+        t = find_solution_v(x, Igens, Sgens, R)
+        t isa String && continue
```
fingolfin: If you want to signal "no solution found", then return `nothing` or `missing`, but not an error string.

lkastner:
1. Please fix the indentation. Oscar comes with an `.editorconfig` file that you can use to avoid this issue in the future. https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/#Editor-configuration To fix the current code, maybe run JuliaFormatter on the changed files.
2. A lot of code does not adhere to the style guide https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/, i.e. functions are too long and there are too many loops nested into each other.
3. Use the `@attributes` system https://docs.oscar-system.org/dev/AbstractAlgebra/misc/#@attributes You seem to compute a lot of properties on demand and have checks in place whether these things are still `nothing` or something. That is exactly what the `@attributes` system is for.
4. Remove commented unused code. You can always retrieve it from git history.

Also in the future I would appreciate smaller and earlier PRs. There are several things here happening at once:
- realization spaces are introduced
- something with chow rings
- more matroid constructions
- some new properties of matroids

It is not easy to work on so many things at once and it will take a while to properly review this. Which also will drag out the process for this to get merged, which will make lengthy rebases necessary.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            normal_Sgens = Vector{RingElem}()
+        else             
+            normal_Sgens = phi.(Sgens)
+        end        
+        
+        
+    else
+        Rnew, xnew = polynomial_ring(coefficient_ring(R), length(xnew_str)) 
+    
+        zero_elim_var = []
```
fingolfin: Enhance type stability:
```suggestion
        zero_elim_var = elem_type(Rnew)[]
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        for i in 1:length(zero_elim)
+            if xs[i] in elim
+                push!(zero_elim_var, Rnew(0))
+            else
+                push!(zero_elim_var, xnew[j] ) 
+                j+=1
+            end
+        end
+        
+        phi = hom(R, Rnew, a->a, zero_elim_var)
```
fingolfin: Don't specify a trivial coefficient map like that, it'll just prevent optimizations elsewher.
```suggestion
        phi = hom(R, Rnew, zero_elim_var)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    Igens = gens(MRS.defining_ideal)
+    Sgens = MRS.inequations
+
+    
+    zero_elim = []        
+    for i in 1:length(xs)
+        if xs[i] in elim
+            push!(zero_elim, 0)
+        else
+            push!(zero_elim, "x"*string(i) )
```
fingolfin:
```suggestion
            push!(zero_elim, "x$i") 
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    zero_elim = []        
+    for i in 1:length(xs)
+        if xs[i] in elim
+            push!(zero_elim, 0)
+        else
+            push!(zero_elim, "x"*string(i) ) 
+        end
+    end
+        
+    xnew_str = Vector{String}(filter(x -> x!=0,  zero_elim))
```
fingolfin: Is `zero_elim` used for anything else? If not, why do you even bother to push `0` into it when you then just filter it out again?

fingolfin: In that case you could just write
```suggestion
    xnew_str = ["x$i" for i in 1:length(xs) if !(xs[i] in elim)]
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            push!(zero_elim, "x"*string(i) ) 
+        end
+    end
+        
+    xnew_str = Vector{String}(filter(x -> x!=0,  zero_elim))    
+    
+    
+    
+    if length(xnew_str) == 0
+        phi = hom(R, cR, a->a, [cR(0) for i in 1:length(xs)])
```
fingolfin:
```suggestion
        phi = hom(R, cR, [cR(0) for i in 1:length(xs)])
```
or even
```suggestion
        phi = hom(R, cR, zeros(cR, length(xs)))
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        ambR = codomain(phi);
```
fingolfin:
```suggestion
        ambR = codomain(phi)
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        if length(Igens) == 0
+            Inew = ideal(ambR, ambR(0))
+        else
+            Inew = ideal(ambR, phi.(Igens)); 
+        end
```
fingolfin: Huh, doesn't this work? If not then that seems like a bug to me.
```suggestion
        Inew = ideal(ambR, phi.(Igens))
```
Same comment applies to other parts of your code, too.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        
+        if length(Sgens) == 0
+            normal_Sgens = Vector{RingElem}()
+        else             
+            normal_Sgens = phi.(Sgens)
+        end
```
fingolfin: Again, doesn't this just work?
```suggestion
        normal_Sgens = phi.(Sgens)
```

src/exports.jl
```diff
@@ -1482,4 +1495,4 @@ export wreath_product
 export write_as_full
 export write_as_linear_combination
 export zonotope
-export zonotope_vertices_fukuda_matrix
+export zonotope_vertices_fukuda_matrix
```
fingolfin: missing trailing newline

test/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,118 @@
+@testset "matroid realization spaces" begin
```
fingolfin: You also need to include your test file (presumably from `runtests.jl` in the same directory), otherwise it won't be run.

test/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,118 @@
...
+        @test coefficient_ring(R1) == GF(2)
+        @test coefficient_ring(R2) == QQ
+        @test coefficient_ring(R3) == ZZ
+        
+        @test length(x1) == 3
+        @test length(x2) == 8
+        @test length(x3) == 9
+	
+        @test mat_M1 ==  matrix(R1, [1 0 1 0 1 0 1; 0 1 1 0 0 1 x1[2]; 0 0 0 1 1 x1[1] x1[3]])
+	@test mat_M2 ==  matrix(R2, [1 1 0 0 1 1 1 0 1; 1 0 1 1 x2[1] x2[3] x2[5] 0 x2[7]; 0 0 0 1 x2[2] x2[4] x2[6] 1 x2[8]])
```
fingolfin: Pease don't mix tabs and spaces. For OSCAR: just don't use tabs.

fingolfin (changes requested):

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
+
+
+mutable struct MatroidRealizationSpace
+    defining_ideal::Union{Ideal, NumFieldOrdIdl, Nothing}
+    inequations::Union{Vector{Oscar.RingElem},Nothing}
+    ambient_ring::Union{Oscar.MPolyRing, Ring, Nothing}
+    realization_matrix::Union{Oscar.MatElem,Nothing}
+    realizable::Union{Bool,Nothing} # Nothing if we're not sure yet if M is realizable.
```
lkastner: What `M`?

src/Combinatorics/Matroids/realization_space.jl
lkastner: I am not sure about this object design. Does it make sense if all those things that can be `Nothing` also are `Nothing`?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+
+@doc raw"""
+defining_ideal(RS::MatroidRealizationSpace)
+
+The ideal of the matroid realization space `RS`.
+"""
+function defining_ideal(RS::MatroidRealizationSpace)
+    return RS.defining_ideal
+end
```
lkastner:
```suggestion
defining_ideal(RS::MatroidRealizationSpace) = RS.defining_ideal
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+@doc raw"""
+inequations(RS::MatroidRealizationSpace)
+
+Generators of the localizing semigroup of `RS`.
+These are the polynomials that need to be nonzero in any realization.
+"""
+function inequations(RS::MatroidRealizationSpace)
+    return RS.inequations
+end
```
lkastner:
```suggestion
inequations(RS::MatroidRealizationSpace) = RS.inequations
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+@doc raw"""
+ambient_ring(RS::MatroidRealizationSpace)
+
+The polynomial ring containing the ideal `defining_ideal(RS)` and the polynomials in `inequations(RS)`. 
+"""
+function ambient_ring(RS::MatroidRealizationSpace)
+    return RS.ambient_ring
+end
```
lkastner:
```suggestion
ambient_ring(RS::MatroidRealizationSpace) = RS.ambient_ring
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+@doc raw"""
+realization_matrix(RS::MatroidRealizationSpace)
+
+A matrix with entries in ambient_ring(RS) whose columns, when filled in with values satisfying equalities
+from `defining_ideal(RS)` and inequations from `inequations(RS)`, form a realization for the matroid. 
+"""
+function realization_matrix(RS::MatroidRealizationSpace)
+    return RS.realization_matrix
+end
```
lkastner:
```suggestion
realization_matrix(RS::MatroidRealizationSpace) = RS.realization_matrix
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+            if col_det != 0 && col in Bs 
+                if isunit( col_det ) 
+                    continue
+                end
+            elseif col_det != 0 # and col is not a basis
+                # determinant nonzero but set not a basis
+                push!( eqs, col_det )
+            elseif col in Bs 
+                #determinant zero but set is a basis, i.e. M is not realizable
+                return MatroidRealizationSpace(F, char, q)
```
lkastner: So why not pass the information about non realizability on to the `MatroidRealizationSpace`?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        if length(RS.inequations) > 0
+            println(io, "avoiding the zero loci of the polynomials\n", RS.inequations)
+        end
+    end
+end
+
+# Constructors
+function MatroidRealizationSpace(F::AbstractAlgebra.Ring, char::Union{Int,Nothing}, q::Union{Int,Nothing})
+    return MatroidRealizationSpace(nothing, nothing, nothing, nothing, false, F, char, q, false)
+end
```
lkastner: At this point I am sure that this constructor is bad design. Even if something is non-realizable there should be information like the defining ideal and inequations. You should not allow this thing that is non-realizable but cannot tell you why.

src/Combinatorics/Matroids/realization_space.jl
lkastner: Nevermind, I had a look at the corresponding constructor and saw, that it sets the corresponding variable to false. Nevertheless, can there be matroid realization spaces without a defining ideal and equations? I thought this is exactly the information needed to compute realizability. I am very unsure whether I would allow these things to be `Nothing`.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        else
+            char = p
+        end
+    end
+
+    if char == 0
+        F = QQ
+    elseif char != nothing
+        F = GF(char)
+    end
```
lkastner: To me it seems like you can always deduce `char` from `q`. If `q` is some `p^k` then `char` is `p`. Otherwise you could just allow `q=0` and then have `char=0`?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+    if q != nothing
+        for i in 1:polyR.nvars
+            push!(eqs, polyR[i]^q-polyR[i])
+        end
+    end
+
+    def_ideal = ideal(polyR,eqs)
+    def_ideal = ideal(groebner_basis(def_ideal))
+    isone(def_ideal) && return MatroidRealizationSpace(F, char, q)
```
lkastner: Again, why not pass on this thing that definitely seems to be a "defining ideal"? Why starve the `MatroidRealizationSpace` of information?

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+    ineqs = gens_2_factors(ineqs)
+    #=
+    # Only saturate easy inequations as it's too slow otherwise.
+    for i in 1:length(ineqs)
+        if count(k -> k!=0, degrees(ineqs[i])) < 3 && sum(degrees(ineqs[i])) < 3
+            def_ideal = saturation(def_ideal, ideal([ineqs[i]]))
+        end
+    end
+    =#
```
lkastner:
```suggestion
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+
+    # Unclear if we should use this. Can be too slow in some cases.
+    #=
+    if !iszero(def_ideal)
+        for i in 1:length(ineqs)
+            ineqs[i] = Oscar.reduce(ineqs[i], gens(def_ideal))
+        end
+    end
+    (polyR(0) in ineqs) && return MatroidRealizationSpace(nothing, nothing, nothing, nothing, false, F, char, q)
+    =#
```
lkastner:
```suggestion
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+    * `char` specifies the characteristic of the coefficient ring, and is used to determine if the matroid
+      is realizable over a field of this characteristic. The default is `nothing`.
+    
+    * `q` is an integer, and when char = p, this input is used to determine whether the matroid 
+      is realizable over the finite field ``GF(p^{q})``. The default is `nothing`.
+
+    * `reduce` determines whether a reduced realization space is returned which means that the equations
+      are used to eliminate variables as far as possible. The default is `true`.
+
+    * `saturate` determines whether `defining_ideal(M)` should be saturated with respect to the semigroup
```
lkastner: Fix indentation (everywhere)

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        
+        push!(elim, v)
+        phi = update_hom(v, w, phi)
+        Igens = filter(x->x!=0, phi.(Igens))
+        return (Igens, phi, elim, false)
+    end
+    return (Igens, phi, elim, true)
+end
+
+
```
lkastner: Remove all this commented code.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,989 @@
...
+        no_v = [term(f,i) for i in 1:length(f) if !(v in vars(monomial(f,i)))]
+        iszero(length(no_v)) && continue
+              
+        h = R(-1)*sum(no_v)
+        return h//den
+        
+    end
+    return "can't solve for v"
+end
```
lkastner: What kind of return values are these? The signature is unclear. Why not throw errors if no solution can be found? Or return a pair `(solution_found::Bool, solution)` such that other functions can properly deal with this? You probably don't want someone to try `isa String` on the return value to figure out what is going on.

lkastner (changes requested):

src/Combinatorics/Matroids/realization_space.jl
benlorenz: Also please don't use `display`:
> In general, you cannot assume that `display` output goes to `stdout` (unlike `print(x)` or `show(x)`). For example, `display(x)` may open up a separate window with an image.

It also doesn't work properly with `jldoctest`.

docs/src/Combinatorics/matroids.md
```diff
@@ -119,7 +119,36 @@ augmented_chow_ring(M::Matroid)
...
 
-For a matroid ``M``, of rank ``d`` on a ground set ``E`` of size ``n`` realizable over a ring ``K``, 
+
+Let ``M`` be a matroid of rank ``d`` on a ground set ``E`` of size ``n``. Its *realization space* 
+``\mathcal{R}(M)`` is an affine scheme that parameterizes all hyperplane arrangements that realize
+the matroid ``M`` (up to the action of ``PGL(r)``). 
+
+
+```@docs
+is_representable(M; char::Union{Int,Nothing}=nothing, q::Union{Int,Nothing}=nothing)
```
lkastner: I cannot find the implementation of this function.

src/exports.jl
```diff
@@ -975,6 +981,7 @@ export maximal_normal_subgroups, has_maximal_normal_subgroups, set_maximal_norma
 export maximal_polyhedra
 export maximal_subgroup_reps
 export maximal_subgroups, has_maximal_subgroups, set_maximal_subgroups
+export min_revlex_basis_encoding
```
lkastner: This needs to move further down alphabetically.

src/Combinatorics/Matroids/matroids.jl
```diff
@@ -742,13 +742,38 @@ Construct the Pappus matroid.
...
+Construct the Möbius-Kantor matroid.
+"""
+moebius_kantor_matroid() = matroid_from_revlex_basis_encoding("0******0******0**********0********0*******0****0**0*****", 3, 8)
+
+"""
+    perles_matroid() 
+
+Construct the Perles matroid.
+"""
+
```
lkastner:
```suggestion
```
The docs will not show if they are separated from the function in any way.

src/Combinatorics/Matroids/matroids.jl
```diff
@@ -742,13 +742,38 @@ Construct the Pappus matroid.
...
 Construct the Vamos matroid.
 """
 vamos_matroid() = Matroid(Polymake.matroid.vamos_matroid())
 
+"""
+    R10_matroid() 
+
+Construct the R10 matroid.
+"""
+
```
lkastner:
```suggestion
```
Again. Have you tried building the docs? https://docs.oscar-system.org/dev/DeveloperDocumentation/documentation/#build_doc

src/Combinatorics/Matroids/properties.jl
```diff
@@ -1065,3 +1086,35 @@ function is_minor(Minor::Matroid, M::Matroid)
...
+function indicator_vector(S::Vector{Int}, n::Int)
+    return map(x -> x in S ? 1 : 0 , 1:n)
+end
+
+function matroid_base_polytope(Q::Matroid)
+    n = length(matroid_groundset(Q))
+    Q = isomorphic_matroid(Q, [i for i in 1:n])
+    Delta_verts = hcat([indicator_vector(x, n) for x in bases(Q)]...)
+    return convex_hull(Delta_verts') 
+end
```
lkastner:
````suggestion
@doc raw"""
    matroid_base_polytope(M::Matroid) 

The base polytope of the matroid `M`.  

# Examples
```jldoctest
julia> D = matroid_base_polytope(uniform_matroid(2,4));

julia> vertices(D)
6-element SubObjectIterator{PointVector{QQFieldElem}}:
 [1, 1, 0, 0]
 [1, 0, 1, 0]
 [1, 0, 0, 1]
 [0, 1, 1, 0]
 [0, 1, 0, 1]
 [0, 0, 1, 1]
```
"""
function matroid_base_polytope(Q::Matroid)
    n = length(matroid_groundset(Q))
    Q = isomorphic_matroid(Q, [i for i in 1:n])
    Delta_verts = hcat([indicator_vector(x, n) for x in bases(Q)]...)
    return convex_hull(Delta_verts') 
end


function indicator_vector(S::Vector{Int}, n::Int)
    return map(x -> x in S ? 1 : 0 , 1:n)
end
````
Really, the docstring is completely arbitrary to Julia. It will not be connected via some complicated parsing mechanism. It is only connected to the function that it is _directly above_.

**Review by lkastner — CHANGES_REQUESTED** [review]

File: docs/src/Combinatorics/matroids.md (hunk 56 above)
Comment (dcorey2814) [reply docs/src/Combinatorics/matroids.md]: This function should be called `is_realizable` which does exist.

Comment (dcorey2814) [reply docs/src/Combinatorics/matroids.md]: Fixed as indicated above.

Comment (dcorey2814) [reply docs/src/Combinatorics/matroids.md]: (should I resolve this?)

File: src/Combinatorics/Matroids/properties.jl (hunk 3 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/properties.jl]: Changes `string` to `str`

File: src/exports.jl (hunk 57 above)
Comment (dcorey2814) [reply src/exports.jl]: Done

File: src/exports.jl (hunk 39 above)
Comment (dcorey2814) [reply src/exports.jl]: Done

File: test/Combinatorics/Matroids/realization_space.jl (hunk 40 above)
Comment (dcorey2814) [reply test/Combinatorics/Matroids/realization_space.jl]: Added `include("Matroids/realization_space.jl")` to `runtests.jl` in the same directory.

File: test/Combinatorics/Matroids/realization_space.jl (hunk 41 above)
Comment (dcorey2814) [reply test/Combinatorics/Matroids/realization_space.jl]: Tabs removed, think I got them all.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 4 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: I added the line "We provide functions that determine the affine coordinate ring of ``\mathcal{R}(M)``. " in the docs. After the functions, we describe how the coordinate ring can be pieced together from the data computed.

Does this address your suggestion? Or would it be better to explicitly mention that affine schemes exist in Oscar and mention how to construct it.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 42 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: Changed `M` to `the matroid`.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 9 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: fixed throughout the file.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 54 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: Removed the initial reduction commented code.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 21 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: Done throughout

File: src/Combinatorics/Matroids/realization_space.jl (hunk 25 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: Fix here, and in the functions `clean`

File: src/Combinatorics/Matroids/realization_space.jl (hunk 34 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: made the suggested change.

Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: resolved below.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 53 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: I ran `format_file`, did this fix the problem?

Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: like the indentation looks like 2 spaces

File: src/Combinatorics/Matroids/realization_space.jl (hunk 30 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: Changed this to a try/catch. I think it should work. The idea is that if no solution is found, then it just goes to the next variable. If none of the variables can be solved for, then the recursion ends.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 55 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: changed so that an error is thrown. Now strings cannot be a return value.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 26 above)
Comment (dcorey2814) [reply src/Combinatorics/Matroids/realization_space.jl]: I fixed this one to throw an error. I think I caught them all.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 8 above)
Comment (LukasKuehne) [reply src/Combinatorics/Matroids/realization_space.jl]: I used display for the matrix as the other options `show` and `print`printed the matrix  in one big line.
Do you know which command to use to print a matrix nicely?

File: src/Combinatorics/Matroids/realization_space.jl (hunk 10 above)
Comment (LukasKuehne) [reply src/Combinatorics/Matroids/realization_space.jl]: Done.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 11 above)
Comment (LukasKuehne) [reply src/Combinatorics/Matroids/realization_space.jl]: This is explained in the documentation. `char` specifies that we want to work over a specific characteristic, and `q` a specific finite field. With `F` one can pass a specific OSCAR ring; this might be redundant and could be removed. I think we added it in the first place if someone has a ring from some other computation and wants to use this exact ring, then they could pass that ring to our methods. If that's not a real use-case we'll remove it.

File: src/Combinatorics/Matroids/realization_space.jl (hunk 49 above)
Comment (LukasKuehne) [reply src/Combinatorics/Matroids/realization_space.jl]: Not quite. There are two scenarios: You want to compute in a specific finite field say F_9, then you pass `q=9`.
Second you be interested in all solutions over some field of characteristic `3`. Then you pass `char=3`. So in particular `char=3` and `q=3` mean different things. We tried to explain this also in the documentation but maybe wasn't really clear?

File: src/Combinatorics/Matroids/realization_space.jl (hunk 8 above)
Comment (benlorenz) [reply src/Combinatorics/Matroids/realization_space.jl]: Please try `show(io, MIME("text/plain"), RS.realization_matrix)`:

```julia
julia> m = zero_matrix(QQ, 3,4);

julia> show(m)
[0 0 0 0; 0 0 0 0; 0 0 0 0]
julia> show(stdout, MIME("text/plain"), m)
[0   0   0   0]
[0   0   0   0]
[0   0   0   0]
```

LukasKuehne: Ah great thanks. Yes that works.
I changed that so this point is done.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: The problem is that `phi.(Vector{RingElem}())` doesn't preserve the type and returns `Any[]`. Is that supposed to be the case?
The list `Any[]` breaks subsequent parts of our code which always expect to receive lists of ring elements.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Fixed.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Fixed.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Fixed.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Thanks for the suggestion. We did this now with Lars' help.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Thanks. Fixed.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Fixed this by restructuring the code.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: We removed the `Nothing`s for the first fields so it is guaranteed that there is always an ideal, ring and inequations also in the nonrealizable case. In this case, we don't save a matrix, so this field is still nothing.
And `char` and `q` are optional parameters which can be set or set to nothing.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: we removed that constructor.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Yes we restructured it and removed that constructor. Thanks for your feedback.

src/Combinatorics/Matroids/realization_space.jl
LukasKuehne: Yes good point. We changed this.

fingolfin:
This PR has a conflict with `master` and hence CI tests cannot be run.

test/Combinatorics/runtests.jl
```diff
@@ -5,4 +5,5 @@ include("SimplicialComplexes.jl")
 include("Matroids/Matroids.jl")
 include("Matroids/Chow_Ring.jl")
 include("Graph.jl")
-include("Matroids/matroid_strata_grassmannian.jl")
```
fingolfin, line 8: The `runtests.jl` files are basically gone, we now basically just execute all the `*.jl` files in subdirectories of `test/`. So resolving the merge conflict should hopefully just require deleting this file/change.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,810 @@
...
+      println(io, "The realizations are parametrized by")
+    elseif !has_attribute(RS, :is_realizable)
+      println(io, "The realization space is")
+    end
+
+    show(io, MIME("text/plain"), RS.realization_matrix)
+    print(io, "\nin the ", RS.ambient_ring)
+    I = RS.defining_ideal
+    if (I isa NumFieldOrdIdl && I.gen != ZZ(0)) ||
+      (I isa Ideal && !iszero(I))
```
fingolfin: This looks to me as if there should be a `iszero(::NumFieldOrdIdl)` method? @thofma does that make sense? Then this could be rewritten as
```suggestion
    if !iszero(I)
```
which I would greatly prefer.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,810 @@
...
+    return true
+  end
+  for p in minimal_primes(RS.defining_ideal)
+    component_non_trivial = true
+    for f in RS.inequations
+      if f in p
+        component_non_trivial = false
+        break
+      end
+    end
```
fingolfin:
```suggestion
    component_non_trivial = all(!in(p), RS.inequations)
```
or
```suggestion
    component_non_trivial = !any(in(p), RS.inequations)
```

test/Combinatorics/runtests.jl
dcorey2814: I got rid of the `matroid_strata_grassmannian.jl` file, and commented everything in the `runtests.jl` file. (I guess we could also just delete it)

benlorenz: Please delete the `runtests.jl`, this is still conflicting with master and the tests won't run if the file is still there but contains commented stuff.

fingolfin: @dcorey2814 that's the opposite of what is needed: the runtests.jl must be deleted. And your new test file of course should be kept

benlorenz: > And your new test file of course should be kept

`matroid_strata_grassmannian.jl` is the old (obsolete?) test file, so deleting that should be fine.

dcorey2814: Ok, I removed `runtests.jl`. And yes, I believe that `matroid_strata_grassmannian.jl` is now obsolete.

src/Combinatorics/Matroids/realization_space.jl
dcorey2814: looks like this was deleted, as requested.

src/Combinatorics/Matroids/realization_space.jl
dcorey2814: looks like this was deleted, as requested.

src/Combinatorics/Matroids/realization_space.jl
dcorey2814: Looks like this change was made.

src/Combinatorics/Matroids/realization_space.jl
dcorey2814: Has this been resolved?

src/Combinatorics/Matroids/realization_space.jl
thofma: I opened an issue https://github.com/thofma/Hecke.jl/issues/1261, but don't count on it to be added before this is merged here.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,820 @@
...
+    I::Union{Ideal,NumFieldOrdIdl},
+    ineqs::Vector{<:RingElem},
+    R::Ring,
+    mat::Union{MatElem,Nothing},
+    char::Union{Int,Nothing},
+    q::Union{Int,Nothing},
+    ground_ring::Ring)
+    return new(I, ineqs, R, mat, char, q, ground_ring, false)
+  end
+end
```
HechtiDerLachs:
```suggestion
@attributes mutable struct MatroidRealizationSpace{BaseRingType, RingType} <: AbsSpec{BaseRingType, RingType}
  defining_ideal::Union{Ideal,NumFieldOrdIdl}
  inequations::Vector{RingElem}
  ambient_ring::Ring
  realization_matrix::Union{MatElem,Nothing}
  char::Union{Int,Nothing}
  q::Union{Int,Nothing}
  ground_ring::Ring
  one_realization::Bool
  
  # Fields for caching
  underlying_scheme::Spec{BaseRingType, RingType}

  function MatroidRealizationSpace(
    I::Union{Ideal,NumFieldOrdIdl},
    ineqs::Vector{<:RingElem},
    R::Ring,
    mat::Union{MatElem,Nothing},
    char::Union{Int,Nothing},
    q::Union{Int,Nothing},
    ground_ring::Ring)
    BaseRingType = typeof(ground_ring)
    RingType = typeof(R)
    if R isa MPolyRing
      PolyRingType = typeof(R)
      MultSetType = MPolyPowersOfElement{BaseRingType, elem_type(BaseRingType), PolyRingType, elem_type(PolyRingType)}
      RingType = MPolyQuoLocRing{BaseRingType, elem_type(BaseRingType), PolyRingType, elem_type(PolyRingType), MultSetType}
    end
    return new{BaseRingType, RingType}(I, ineqs, R, mat, char, q, ground_ring, false)
  end
end
```

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -0,0 +1,820 @@
...
+"""
+inequations(RS::MatroidRealizationSpace) = RS.inequations
+
+@doc raw"""
+    ambient_ring(RS::MatroidRealizationSpace)
+
+The polynomial ring containing the ideal `defining_ideal(RS)` and the polynomials in `inequations(RS)`. 
+"""
+ambient_ring(RS::MatroidRealizationSpace) = RS.ambient_ring
+
```
HechtiDerLachs:
```suggestion
function underlying_scheme(RS::MatroidRealizationSpace{<:Field, <:MPolyQuoLocRing})
  isdefined(RS, :underlying_scheme) && return RS.underlying_scheme

  P = ambient_ring(RS)::MPolyRing
  I = defining_ideal(RS)::MPolyIdeal
  U = powers_of_element(inequations)::MPolyPowersOfElement
  RS.underlying_scheme = Spec(R, I, U)
  return RS.underlying_scheme
end

function underlying_scheme(RS::MatrixRealizationSpace)
  error("method not implemented for realization spaces of type $(typeof(RS))")
end
```

HechtiDerLachs (commented):
Just a comment on how all this might be integrated in the common schemes framework. 

However, I remember that @dcorey2814 found the localizations of affine algebras to be insufficiently performant for the anticipated purposes here. So the integration into the schemes might just be a little playaround for trivial cases? 

@dcorey2814 : You still didn't send me the examples which didn't work for you!

Edit: I should point out that I don't consider these changes to be required for merging. They can also be done at some later point without problems.

dcorey2814:
> Just a comment on how all this might be integrated in the common schemes framework.
> 
> However, I remember that @dcorey2814 found the localizations of affine algebras to be insufficiently performant for the anticipated purposes here. So the integration into the schemes might just be a little playaround for trivial cases?
> 
> @dcorey2814 : You still didn't send me the examples which didn't work for you!
> 
> Edit: I should point out that I don't consider these changes to be required for merging. They can also be done at some later point without problems.

Apologies for the delay. Here is an example. 

```
M = pappus_matroid()
RS = realization_space(M, ground_ring=QQ, simplify=false)
U = inequations(RS)
S = ambient_ring(RS)
I = defining_ideal(RS)

Uloc = MPolyPowersOfElement(S , [S(a) for a in U])
UinvS , _ = Localization(S, Uloc)
Iloc = ideal(UinvS, I)
quo(UinvS, Iloc)
```
I didn't try for long, but even after ~5min it wouldn't terminate.

HechtiDerLachs:
> I didn't try for long, but even after ~5min it wouldn't terminate.

Alright, that was certainly worth having a look at. Your example has revealed a minor flaw with big consequences which I would probably not have found otherwise. Your example runs instantly now. 

Can I push these minor changes to this PR? 

Edit: So, unless you have something bigger to shoot at my localizations with, I see no problem in turning these matroid realization spaces into schemes.

dcorey2814:
> > I didn't try for long, but even after ~5min it wouldn't terminate.
> 
> Alright, that was certainly worth having a look at. Your example has revealed a minor flaw with big consequences which I would probably not have found otherwise. Your example runs instantly now.
> 
> Can I push these minor changes to this PR?
> 
> Edit: So, unless you have something bigger to shoot at my localizations with, I see no problem in turning these matroid realization spaces into schemes.

Ok, I committed @HechtiDerLachs 's suggestions, but when I tried to compile Oscar I got the error message:
```
[ Info: Precompiling Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13]
ERROR: LoadError: UndefVarError: `Spec` not defined
Stacktrace:
  [1] top-level scope
    @ ~/.julia/packages/AbstractAlgebra/dsta0/src/Attributes.jl:83
  [2] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
  [3] include(x::String)
    @ Oscar ~/Documents/GitHub/oscar-matroids/src/Oscar.jl:20
  [4] top-level scope
    @ ~/Documents/GitHub/oscar-matroids/src/Combinatorics/Matroids/JMatroids.jl:4
  [5] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
  [6] include(x::String)
    @ Oscar ~/Documents/GitHub/oscar-matroids/src/Oscar.jl:20
  [7] top-level scope
    @ ~/Documents/GitHub/oscar-matroids/src/Oscar.jl:224
  [8] include
    @ ./Base.jl:457 [inlined]
  [9] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
    @ Base ./loading.jl:2049
 [10] top-level scope
    @ stdin:3
in expression starting at /Users/danielcorey/Documents/GitHub/oscar-matroids/src/Combinatorics/Matroids/realization_space.jl:2
in expression starting at /Users/danielcorey/Documents/GitHub/oscar-matroids/src/Combinatorics/Matroids/JMatroids.jl:4
in expression starting at /Users/danielcorey/Documents/GitHub/oscar-matroids/src/Oscar.jl:1
in expression starting at stdin:3
ERROR: Failed to precompile Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] to "/Users/danielcorey/.julia/compiled/v1.9/Oscar/jl_h2vDck".
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, keep_loaded_modules::Bool)
   @ Base ./loading.jl:2300
 [3] compilecache
   @ ./loading.jl:2167 [inlined]
 [4] _require(pkg::Base.PkgId, env::String)
   @ Base ./loading.jl:1805
 [5] _require_prelocked(uuidkey::Base.PkgId, env::String)
   @ Base ./loading.jl:1660
 [6] macro expansion
   @ ./loading.jl:1648 [inlined]
 [7] macro expansion
   @ ./lock.jl:267 [inlined]
 [8] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1611
```
I even tried doing `git merge bs/matroids master` and it said it was up to date.

LukasKuehne:
> Ok, I committed @HechtiDerLachs 's suggestions, but when I tried to compile Oscar I got the error message:

In light of the near book deadline, would you @HechtiDerLachs mind if we'd save the integration with the schemes framework for the next PR?

HechtiDerLachs:
Sure, no problem! I didn't test my suggestion, it was just to give you a taste of it and since it was mentioned in the discussion. We've had the same problem with the toric schemes and solved it using `src/forward_declarations.jl`. I will make a working version locally and do a PR to your branch. Then you can still decide whether you want it here now or later. But I understand if you have more urgent things to worry about right now.

HechtiDerLachs:
Since this PR was already messed up by the committing of the suggestions, I allowed myself to push the fixes directly here (and not to a separate PR against your branch as originally announced). I hope you don't mind. If this crosses your plans, please feel free to revert these!

One thing I had to do, is to adjust the printing routine's signature so that printing is not automatically overwritten by the scheme's method. Please check! 

Edit: Let me emphasize that these changes should not have *any* effect on your code regarding speed or functionality which was already there! The scheme-side is created only on request (see the tests) and should not play a role for anything else you have been doing so far.

src/Combinatorics/Matroids/realization_space.jl
```diff
@@ -31,7 +31,7 @@
   end
 end
 
-function Base.show(io::IO, RS::MatroidRealizationSpace)
+function Base.show(io::IO, ::MIME"text/plain", RS::MatroidRealizationSpace)
```
HechtiDerLachs: This is the change in the printing that I was talking about.

HechtiDerLachs:
Relevant tests seem to be passing again, so from my side, this seems to be OK now.

src/Combinatorics/Matroids/realization_space.jl
HechtiDerLachs: What's the type of `phi`? This should become type-stable so that the result not `Any[]`. In either case, you could also do `Inew = ideal(ambR, ambR.(phi.(Igens)))` and avoid the `if`.

LukasKuehne:
> Relevant tests seem to be passing again, so from my side, this seems to be OK now.

Thanks for your work. It's also working on my side again.

fingolfin (approved):
I don't currently have time to review this again, but assuming (resp. trusting) reasonable improvements were made based on our feedback I have no objections to merging this (in any case and as always, it can still be improved later, after all). To be clear, though I did not look at it again, so I'll leave it to e.g. @lkastner or @bschroter or someone else to make the final call.

HechtiDerLachs:
@lkastner : I think your moving the method for `underlying_scheme` to `experimental` will probably not work due to the use of `Spec` in the code. Let's see what the tests say, but if they fail now, I suggest to move it back for the moment.

lkastner:
The Pappus example does not work anymore and makes the doctests fail:
```
julia> realization(pappus_matroid(), q=4)
One realization is given by
nothing
in the Multivariate polynomial ring in 2 variables over GF(2)
within the vanishing set of the ideal
ideal(x2 + 1, x1^2 + x1 + 1)
avoiding the zero loci of the polynomials
RingElem[0]
```
This also failed before my changes. @dcorey2814 @LukasKuehne 

The breaking commit seems to be https://github.com/oscar-system/Oscar.jl/pull/2925/commits/f7e32c1985ad42e2b6993d5e675c08f6c813164f . At a glance it could be caused by `eqs` now being populated, after it is used, while before it first got populated and then used.

LukasKuehne:
> The Pappus example does not work anymore and makes the doctests fail:
> 
> ```
> julia> realization(pappus_matroid(), q=4)
> One realization is given by
> nothing
> in the Multivariate polynomial ring in 2 variables over GF(2)
> within the vanishing set of the ideal
> ideal(x2 + 1, x1^2 + x1 + 1)
> avoiding the zero loci of the polynomials
> RingElem[0]
> ```
> 
> This also failed before my changes. @dcorey2814 @LukasKuehne
> 
> The breaking commit seems to be [f7e32c1](https://github.com/oscar-system/Oscar.jl/commit/f7e32c1985ad42e2b6993d5e675c08f6c813164f) . At a glance it could be caused by `eqs` now being populated, after it is used, while before it first got populated and then used.

Hi,

Thanks for spotting this bug. I just pushed a fix. It's working again for me.

lkastner (approved):

PR #2927, opened 2023-10-17
Bump `Documenter.jl` to `v1.0`

lgoettgens:
See title.

Documenter.jl v1.0 has a lot more features for cross-referencing between docs of different packages, that can be very useful for our copy-pasting hack. But I skipped dealing with that / deactivated it for now and leave the fix to https://github.com/oscar-system/Oscar.jl/issues/588.

Preview: https://docs.oscar-system.org/previews/PR2927/

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -3,5 +3,5 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 DocumenterCitations = "daee34ce-89f3-4625-b898-19384cb65244"
 
 [compat]
-Documenter = "0.27.0"
+Documenter = "1.1"
 DocumenterCitations = "~1.1, ~1.2"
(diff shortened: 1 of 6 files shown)


lgoettgens:
The warning 
```
┌ Warning: Unexpected Julia interpolation of type Symbol in the Markdown.
│   value = :k_
└ @ Documenter.HTMLWriter ~/.julia/packages/Documenter/cxHBW/src/html/HTMLWriter.jl:2215
```
will get resolved by https://github.com/thofma/Hecke.jl/pull/1246.

benlorenz (approved):
The preview looks good, I did not check all the pages but I haven't found any issues.

HereAround (approved):
Thank you @lgoettgens . Similar to @benlorenz , I checked the preview and could not find issues. I believe this is thus ready to go.

PR #2928, opened 2023-10-17
Add stdlib compat bounds to `test/Project.toml`

lgoettgens:
Follow-up to https://github.com/oscar-system/Oscar.jl/pull/2921


benlorenz:
Not sure if this is necessary, we won't register the test-folder separately. Also this project will always be merged with the Oscar project when it is used.

lgoettgens:
Once these stdlib are actually upgradeable, this enables us to require more functionality by setting explicit compat bounds.

benlorenz (approved):

fingolfin (approved):

PR #2931, opened 2023-10-17
Compare 'nothing' using === not ==

fingolfin:
... for optimal code results. While in most cases it won't matter, it is
a good habit to just always adhere to this.

### DIFF
--- a/src/Combinatorics/Matroids/properties.jl
+++ b/src/Combinatorics/Matroids/properties.jl
@@ -161,7 +161,7 @@ function flats_impl(M::Matroid, r::Union{Int,Nothing}, num_flats::Int, pm_flats)
         jl_flats = reverse(jl_flats)
     end
     matroid_flats = [[M.groundset[i] for i in flat] for flat in jl_flats]
-    if r!=nothing
+    if r != nothing
         if r<0 || r>rank(M)
             error("The specified rank needs to be between 0 and the rank of the matroid.")
         end
--- a/src/Groups/matrices/MatGrp.jl
+++ b/src/Groups/matrices/MatGrp.jl
@@ -290,15 +290,14 @@ function lies_in(x::MatElem, G::MatrixGroup, x_gap)
          end
       end
    end
-   if isdefined(G,:descr) && G.descr==:GL
+   if isdefined(G,:descr) && G.descr === :GL
       return det(x)!=0, x_gap
-   elseif isdefined(G,:descr) && G.descr==:SL
+   elseif isdefined(G,:descr) && G.descr === :SL
       return det(x)==1, x_gap
-   else
-      if x_gap==nothing x_gap = map_entries(G.ring_iso, x) end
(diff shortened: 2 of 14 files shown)


src/Combinatorics/Matroids/properties.jl
```diff
@@ -161,7 +161,7 @@ function flats_impl(M::Matroid, r::Union{Int,Nothing}, num_flats::Int, pm_flats)
         jl_flats = reverse(jl_flats)
     end
     matroid_flats = [[M.groundset[i] for i in flat] for flat in jl_flats]
-    if r!=nothing
+    if r != nothing
```
thofma, line 164: something is missing here

src/Groups/matrices/MatGrp.jl
```diff
@@ -341,9 +340,10 @@ function (G::MatrixGroup)(x::MatrixGroupElem; check::Bool=true)
    else
       _is_true, x_gap = lies_in(x.elm,G,nothing)
       @req _is_true "Element not in the group"
-      if x_gap==nothing return MatrixGroupElem(G,x.elm)
-      else return MatrixGroupElem(G,x.elm,x_gap)
+      if x_gap == nothing
```
thofma, line 343: here too

lkastner:
> ... for optimal code results. While in most cases it won't matter, it is a good habit to just always adhere to this.

Why not use `isnothing`? Sure, it does the same thing as `===`, but wouldn't this be the better style, to use the dedicated function?

thofma:
https://docs.julialang.org/en/v1.6/manual/performance-tips/#Checking-for-equality-with-a-singleton

(Should be the same starting with 1.7)

HereAround (approved):
Thank you @fingolfin . Looks good to me.

thofma:
There were a few cases missing, which should have been fixed before merging this

lkastner:
I still don't understand why not to use `isnothing`. The singleton comparison statement did not enlighten me (I know that `Nothing` is a singleton), I am sorry.

fingolfin:
I wanted to minimize the changes, that was very easy with a reggex changing `==nothing` to `===nothing`, would have been much harder to change it to `isnothing`. That said, I also don't see anything "better" about using `isnothing`, seems like a purely subjective matter of taste.

src/Combinatorics/Matroids/properties.jl
fingolfin: Indeed *sigh*. I'll send a follow up PR.

lkastner:
`isnothing` is the dedicated Julia function for figuring out if something is `Nothing`. So if at some point comparing to `Nothing` should be done in a different way than `===` this would automatically be changed for us if we used `isnothing`. Similar to how we are encouraged not to directly access members. But I guess this issue is really irrelevant, so I will stop here.

HereAround:
(My apologies for moving this along too quickly.)

PR #2935, opened 2023-10-18
Double complexes

HechtiDerLachs:
After this morning's discussion a first draft of how I think double complexes and their total complexes can effectively be implemented.


HechtiDerLachs:
I added a first implementation for morphisms of complexes (not complexes of morphisms!), as was suggested by @wdecker . For the proof of concept I implemented a version which lifts a morphism of modules incrementally through their free resolutions. I will link it in the source code below.

experimental/MorphismOfComplexes/test/lift_resolution.jl
```diff
@@ -0,0 +1,30 @@
...
+
+  M1, _ = quo(F, IF)
+  M2, _ = quo(F, JF)
+
+  phi = hom(M2, M1, gens(M1))
+
+  res1 = free_resolution(M1).C
+  res2 = free_resolution(M2).C
+
+  psi = Oscar.lift_morphism_through_free_resolutions(phi, domain_resolution=res2, codomain_resolution=res1);
```
HechtiDerLachs: This is the command to lift a morphism of modules `phi : M -> N` to a morphism of their resolutions.

experimental/MorphismOfComplexes/src/Constructors.jl
```diff
@@ -0,0 +1,10 @@
+function lift_morphism_through_free_resolutions(phi::ModuleFPHom;
+    domain_resolution::ComplexOfMorphisms=free_resolution(domain(phi)).C,
+    codomain_resolution::ComplexOfMorphisms=free_resolution(codomain(phi)).C
+  )
+  factory = LiftingMorphismsThroughResolution(phi)
+
+  return MorphismOfComplexes(domain_resolution, codomain_resolution, factory, 0,
+                             left_bound = -2, right_bound = -1, 
+                             extends_left = false, extends_right = true)
+end
```
HechtiDerLachs: This is the actual implementation of that function's method. It involves creating a "factory" to do the lifting on request.

experimental/MorphismOfComplexes/src/Types.jl
```diff
@@ -0,0 +1,123 @@
...
+  end
+end
+
+mutable struct LiftingMorphismsThroughResolution{T} <: AbsMorphismFactory{T}
+   original_map::ModuleFPHom
+  
+   function LiftingMorphismsThroughResolution(phi::ModuleFPHom)
+    return new{ModuleFPHom}(phi)
+  end
+end
```
HechtiDerLachs: The "factory" is a concrete type which remembers the original morphism. The call signature of this struct will be overloaded to do the lifting.

experimental/MorphismOfComplexes/src/Methods.jl
```diff
@@ -0,0 +1,24 @@
...
+
+  # Build up the maps recursively via liftings
+  psi = phi[i-1]
+  dom_map = map(domain(phi), i)
+  cod_map = map(codomain(phi), i)
+  x = gens(domain(dom_map))
+  y = psi.(dom_map.(x))
+  z = [preimage(cod_map, u) for u in y]
+  return hom(domain(dom_map), domain(cod_map), z)
+end
```
HechtiDerLachs: Here we see the lifting actually being computed.

HechtiDerLachs:
I added some user facing documentation. @wdecker @jankoboehm @thofma @fieker : Could you have a look and let me know whether you approve of this? More documentation for the programmer about the internals will follow.

experimental/DoubleComplex/docs/src/user_interface.md
```diff
@@ -0,0 +1,90 @@
...
+    has_right_bound(D::AbsDoubleComplexOfMorphisms)
+    has_left_bound(D::AbsDoubleComplexOfMorphisms)
+```
+If they exist, these bounds can be asked for using 
+```@docs
+    right_bound(D::AbsDoubleComplexOfMorphisms)
+    left_bound(D::AbsDoubleComplexOfMorphisms)
+    upper_bound(D::AbsDoubleComplexOfMorphisms)
+    lower_bound(D::AbsDoubleComplexOfMorphisms)
+```
```
fingolfin, line 64: Looking at this, @fieker and me wonder what an actual application of these and related methods would be? I didn't see any in this PR (maybe I missed them, pointers appreciated).

experimental/DoubleComplex/docs/src/user_interface.md
```diff
@@ -0,0 +1,90 @@
...
+
+In practice one usually encounters complexes which are *bounded* in the sense 
+that outside some specified area of indices ``(i, j) \in \mathbb Z^2`` the entries 
+``D_{i, j}`` are all zero. Such entries are then usually omitted. 
+
+## Basic getters and attributes
+In OSCAR the generic functionality for double complexes is declared for the 
+abstract type `AbsDoubleComplexOfMorphisms`. These functions comprise
+```julia
+  getindex(D::AbsDoubleComplexOfMorphisms, i::Int, j::Int) # Get the `(i,j)`-th entry of `D`
```
fingolfin, line 23: How about a function to test if a specific index is available, i.e. something like `has_index(D, i, j)` to test whether an object at that position is already computed.

And then separately something like `can_compute_index(D, i, j)` which returns `true` if either the object `D[i,j]` already is there, or if it is not there but could be computed.

experimental/DoubleComplex/docs/src/user_interface.md
```diff
@@ -0,0 +1,90 @@
...
+```
+```@docs
+    horizontal_map(dc::AbsDoubleComplexOfMorphisms, i::Int, j::Int)
+    vertical_map(dc::AbsDoubleComplexOfMorphisms, i::Int, j::Int)
+```
+In which direction the maps in the rows and columns go can be asked with the following methods:
+```@docs
+    horizontal_typ(dc::AbsDoubleComplexOfMorphisms)
+    vertical_typ(dc::AbsDoubleComplexOfMorphisms)
```
fingolfin: I don't like `_typ`, it is easy confuse with `type` and also looks like a typo (no pun intended). How about `_direction`?

experimental/DoubleComplex/docs/src/user_interface.md
```diff
@@ -0,0 +1,90 @@
...
+```
+Double complexes can be bounded or unbounded. But even in the bounded case where only finitely 
+many ``D_{i, j}`` can be non-zero, it is not clear that a concrete bound for the 
+range of indices ``(i, j)`` of possibly non-zero entries is known. Then again, even in that case 
+such a bound might be rather theoretical and it is still encouraged that double complexes be implemented 
+lazy and not fill out the full grid of morphisms on construction. Thus a merely theoretical bound 
+on the range of indices does not seem to be practically relevant.
```
fingolfin: I find this a bit confusing. There seem to be multiple bounds in play: one is the "bound of things that are actually there", then there is a "bound on things that could be computed", and they don't seem to be cleanly separated...

Naively, this approach seems to me both too general and too limited: too general because I can't think of any interesting algorithms that would actually be able to successfully use this APIs; too limited because it seems to presuppose that both of these bounds always form rectangles, and I don't see why that should be a particularly relevant case.

HechtiDerLachs:
I did not feel very confident to introduce a whole new pattern along the lines of `has_something` `can_compute_something`, because this differs significantly from the ordinary `ComplexOfMorphisms`. But since you suggested this yourself now, I am happy to follow this. Feel free to have a look at the current state of the art.

experimental/DoubleComplex/src/Attributes.jl
```diff
@@ -0,0 +1,211 @@
...
+  return D.vertical_maps
+end
+
+### Boundedness
+has_right_bound(D::DoubleComplexOfMorphisms) = isdefined(D, :right_bound)
+has_left_bound(D::DoubleComplexOfMorphisms) = isdefined(D, :left_bound)
+has_upper_bound(D::DoubleComplexOfMorphisms)   = isdefined(D, :upper_bound)
+has_lower_bound(D::DoubleComplexOfMorphisms)   = isdefined(D, :lower_bound)
+
+function is_complete(D::DoubleComplexOfMorphisms)
```
HechtiDerLachs, line 28: @jankoboehm : This should be the generic implementation you asked for. Please check.

experimental/DoubleComplex/src/Attributes.jl
```diff
@@ -0,0 +1,211 @@
...
+
+@doc raw"""
+    finalize!(D::DoubleComplexOfMorphisms)
+
+Compute all non-zero entries `D[i, j]` and maps of connected components of entries 
+of `D` in the grid of the double complex, starting from non-zero entries in 
+the cache which have already been computed.
+
+This can be used to write a full double complex to disc by filling the cache first.
+"""
```
HechtiDerLachs: @jankoboehm : This one the other. I'm not 100% sure how to explain all that stuff with the connected components, though. Maybe better to not export this and leave it to the internal people who know what they're doing.

HechtiDerLachs:
I think I have now talked to all people relevant to this PR except @wdecker . So far, everyone seems to be OK with these changes in general and it is my impression, that this is ready for `experimental`. 

However, we did not yet talk about the `MorphismOfComplexes` part and that certainly needs some cleaning up after all the discussion on the double complexes. Since everything about that is cleanly separated in a folder of its own. I could take that out of this PR and make it a separate one later on.

HechtiDerLachs:
I removed the morphisms of complexes and added more documentation. If anyone still has any objections, please let me know. Otherwise this should be good to go and I would run the tests for merge.

experimental/DoubleComplex/src/Attributes.jl
jankoboehm: Looks good

experimental/DoubleComplex/src/Attributes.jl
jankoboehm: I don't see a reason why we should not give it to the user, since it is clearly formulated and not ambiguous.

jankoboehm:
Think we can go forward with this, if there are not objections, @wdecker ?

wdecker (approved):

experimental/DoubleComplex/docs/src/advice_for_the_programmer.md
```diff
@@ -0,0 +1,155 @@
...
+This generic functionality assumes certain methods to be implemented for the 
+`ChainType` and the `MorphismType` of the double complex `D`. For instance, 
+it must be possible to compose two morphisms of type `<:MorphismType` and 
+get a new object of type `<:MorphismType`. Sometimes, the required functionality 
+is not streamlined throughout OSCAR (and there is little hope to achieve this).
+One example for this are direct sums: For finitely generated modules, the 
+function takes a special keyword argument `task` to indicate whether the inclusion 
+and projection maps should also be returned, while for `TorQuadModule`s, this keyword argument 
+is not even available. To potentially accomodate all these different types in our 
+double complexes, the generic code uses an internal method
```
afkafkafk13: This paragraph is rather hard to follow, because it mixes general remarks and the specific example of total_complex. It might be better for the reader, if you visibly split general remarks and specific example.

experimental/DoubleComplex/docs/src/user_interface.md
```diff
@@ -0,0 +1,83 @@
...
+```@docs
+    horizontal_direction(dc::AbsDoubleComplexOfMorphisms)
+    vertical_direction(dc::AbsDoubleComplexOfMorphisms)
+```
+Double complexes can be bounded or unbounded. But even in the bounded case where only finitely 
+many ``D_{i, j}`` can be non-zero, it is not clear that a concrete bound for the 
+range of indices ``(i, j)`` of possibly non-zero entries is known. Then again, even in that case 
+such a bound might be rather theoretical and it is still encouraged that double complexes be implemented 
+lazy and not fill out the full grid of morphisms on construction. Thus a merely theoretical bound 
+on the range of indices does not seem to be practically relevant.
```
afkafkafk13: If the bounds are not of practical relevance, then why do you spend a whole paragraph on this. Maybe it would be better to state that the double complexes are a priori considered as unbounded, unless a bound is given, inherited or explicitly computed. It should then also be mentioned that lazy evaluation is crucial for handling the bounded and unbounded setting in this unified way.

experimental/DoubleComplex/src/Attributes.jl
afkafkafk13: connected components of a complex? This sounds a bit misleading. What about contiguous pieces?

experimental/DoubleComplex/src/Attributes.jl
```diff
@@ -0,0 +1,210 @@
...
+
+
+function horizontal_range(D::DoubleComplexOfMorphisms) 
+  is_horizontally_bounded(D) || error("complex is not known to be horizontally bounded")
+  if horizontal_direction(D) == :chain
+    return right_bound(D):-1:left_bound(D)
+  elseif horizontal_direction(D) == :cochain
+    return left_bound(D):right_bound(D)
+  end
+  error("typ not recognized")
```
afkafkafk13, line 118: typo? 'type'
(and further occurrences)

experimental/DoubleComplex/src/Methods.jl
```diff
@@ -0,0 +1,106 @@
+### Generic functionality
+@doc raw"""
+    total_complex(D::AbsDoubleComplexOfMorphisms)
+
+Construct the total complex of the double complex `D`. 
+
+Note that `dc` needs to be reasonably bounded for this to work so that the strands
```
afkafkafk13: This help text does not seem to be self-contained, because `dc` appears precisely once and without explanation.

experimental/DoubleComplex/src/Methods.jl
```diff
@@ -0,0 +1,106 @@
...
+    _direct_sum(u::Vector{T}) where {T}
+
+Internal method to return a triple `(s, incs, prs)` consisting of an object 
+`s` representing the direct sum of the entries in `u`, together with vectors 
+of maps `incs` for the inclusion maps `u[i] → s` and `prs` for 
+the projections `s →  u[i]`.
+
+Generically this will default to `direct_sum(u)`. If that does not produce 
+a result with the required output format, you must overwrite this method 
+for your specific type `T`.
```
afkafkafk13, line 31: This sounds like an instruction to the programmer on how to handle _direct_sum. It feels a bit weird in the help text of an internal function.

experimental/DoubleComplex/src/Types.jl
```diff
@@ -0,0 +1,415 @@
...
+@doc raw"""
+    lower_bound(D::AbsDoubleComplexOfMorphisms)
+
+Returns a bound ``B`` such that `D[i, j]` can be assumed to be zero 
+for ``j < B``. Whether or not requests for `D[i, j]` beyond that bound are 
+legitimate can be checked using `can_compute_index`.
+"""
+lower_bound(D::AbsDoubleComplexOfMorphisms) = lower_bound(underlying_double_complex(D))
+
+#= Has become obsolete since has_index etc.
```
afkafkafk13: If this is obsolete, please remove it before merge.

experimental/DoubleComplex/src/Types.jl
```diff
@@ -0,0 +1,415 @@
...
+`D[i, j]` and (co-)boundary maps for ``j ≪  0``.
+"""
+extends_down(D::AbsDoubleComplexOfMorphisms) = extends_down(underlying_double_complex(D))
+=#
+
+@doc raw"""
+    is_complete(dc::AbsDoubleComplexOfMorphisms)
+
+Returns `true` if for all indices `(i, j)` with `has_index(dc, i, j) = true` and 
+`dc[i, j]` non-zero, the vertex `(i, j)` lays on an "island" of non-zero entries
```
afkafkafk13: `is lying` not `lays`

experimental/DoubleComplex/src/Types.jl
```diff
@@ -0,0 +1,415 @@
...
+    has_index(D, i+1, j) || !can_compute_index(D, i+1, j) || return false
+    has_index(D, i-1, j) || !can_compute_index(D, i-1, j) || return false
+    has_index(D, i, j+1) || !can_compute_index(D, i, j+1) || return false
+    has_index(D, i, j-1) || !can_compute_index(D, i, j-1) || return false
+  end
+  set_attribute!(D, :is_complete, true)
+  return true
+end
+
+#= has become obsolete with the advent of has_entry etc.
```
afkafkafk13: pleaes remove obsolete code before merge.

experimental/DoubleComplex/src/Types.jl
```diff
@@ -0,0 +1,415 @@
...
+# Produce the t = (i, j)-th entry which will then be cached.
+function (fac::ChainFactory)(dc::AbsDoubleComplexOfMorphisms, t::Tuple)
+  return fac(dc, t...)
+end
+
+# Test whether the (i, j)-th entry can be computed.
+can_compute(fac::ChainFactory, dc::AbsDoubleComplexOfMorphisms, t::Tuple) = can_compute(fac::ChainFactory, dc::AbsDoubleComplexOfMorphisms, t...)
+
+function can_compute(fac::ChainFactory, dc::AbsDoubleComplexOfMorphisms, i::Int, j::Int)
+  error("testing whether the ($i, $j)-th entry of can be computed using $fac has not been implemented; see the programmer's documentation on double complexes for details")
```
afkafkafk13: mising word/symbol: "entry of can"

experimental/DoubleComplex/src/tensor_products.jl
```diff
@@ -0,0 +1,126 @@
...
+
+Create the tensor product of two complexes `C` and `D` as a double complex.
+
+In order for the generic implementation to work for your specific `ChainType` the following 
+needs to be implemented.
+
+  * `morphism_type(ChainType)` must produce the type of morphisms between objects of type `ChainType`;
+  * the call signature for `function (fac::TensorProductFactory{ChainType})(dc::AbsDoubleComplexOfMorphisms, i::Int, j::Int)` needs to be overwritten for your specific instance of `ChainType` to produce the `(i, j)`-th entry of the double complex, i.e. the tensor product of `C[i]` and `D[j]`;
+  * the call signature for `function (fac::HorizontalTensorMapFactory{ChainType})(dc::AbsDoubleComplexOfMorphisms, i::Int, j::Int)` needs to be overwritten to produce the map on tensor products `C[i] ⊗ D[j] → C[i±1] ⊗ D[j]` induced by the (co-)boundary map on `C` (the sign depending on the `typ` of `C`);
+  * similarly for the `VerticalTensorMapFactory`.
```
afkafkafk13, line 96: This sounds like a manual entry targeting programmers. Do you really want it in the docstring of `tensor_product`?

afkafkafk13 (commented):
some minor comments, mostly nitpicks on obsolete code, unclear formulations and typos

experimental/DoubleComplex/src/Attributes.jl
afkafkafk13: seeing the discussion with @fingolfin above, you might have wanted `direction` instead of `type`

experimental/DoubleComplex/docs/src/advice_for_the_programmer.md
HechtiDerLachs: At this moment, I don't see how, really. Sorry.

experimental/DoubleComplex/docs/src/user_interface.md
HechtiDerLachs: The bounds do have practical relevance. Say you have a double complex  and you want to compute it's total complex. For every strand (direct sum of the `D_{i, j}` with `i + j = k`) there would be infinitely many pairs `(i, j)` that one would have to check to be zero. So bounding the range in which to look, is key here. 

The point here was to say that a *theoretical* bound is not relevant here. Only concretely known bounds are supported here, i.e. `has_upper_bound` is `true` only when a *concrete* bound is known (and not something like "We know a Groebner basis is finite, but it might also be a million elements"). 

I wanted to have all this discussion here so that people do not come forward saying: "Have you thought of this? And that?".

experimental/DoubleComplex/src/Attributes.jl
HechtiDerLachs: `typ` was what I found in the `ComplexOfMorphisms`. Not my choice. But I wanted to stay close to the complexes also with terminology. No strong feelings about changing this from my side.

experimental/DoubleComplex/src/tensor_products.jl
HechtiDerLachs: I wanted to have it here for as long as we're in the phase of people trying to use this stuff for the first time for new types. It's basically implemented for nothing but modules at this point. So for any new user (like those on slack) I wanted to put this out very prominently to give a hint to where to start implementing things for their own types.

HechtiDerLachs:
Changes done. Have a look whether you approve. If yes, I can run the tests again.

experimental/DoubleComplex/docs/src/user_interface.md
HechtiDerLachs: I overhauled this part now.

experimental/DoubleComplex/docs/src/user_interface.md
HechtiDerLachs: It is used through `vertical_range` and `horizontal_range` in the `total_complex` routine.

experimental/DoubleComplex/docs/src/advice_for_the_programmer.md
afkafkafk13: Actually, your small change already made it a lot less confusing.

afkafkafk13 (approved):
Thank you for the changes.

HechtiDerLachs:
Can this be merged, then?

PR #2937, opened 2023-10-19
Relax signature of `==(::MPolyIdeal, ::MPolyIdeal)`

joschmitt:
Small change, but requires a long explanation:
The current signature of `==` for ideals in multivariate polynomial rings is
```
==(::MPolyIdeal{T}, ::MPolyIdeal{T}) where T
```
so this function is not called if one tries to compare ideals in polynomial rings of different type (`T`), instead the generic `==` is called:
```
julia> R, (x, y) = QQ["x", "y"];

julia> S, (u, v) = grade(R, [ 1, 1 ]);

julia> ideal(R, [x]) == ideal(S, [u])
false

julia> @which ideal(R, [x]) == ideal(S, [u])
==(x, y)
     @ Base Base.jl:127
```
The answer `false` is of course not wrong, but in my opinion this should **error** because the comparison doesn't make sense; this is what I added in this PR:
```
julia> ideal(R, [x]) == ideal(S, [u])
ERROR: Base rings must coincide.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] check_base_rings(I::MPolyIdeal{QQMPolyRingElem}, J::MPolyIdeal{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
   @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-ideals.jl:75
 [3] ==(I::MPolyIdeal{QQMPolyRingElem}, J::MPolyIdeal{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
   @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-ideals.jl:993
 [4] top-level scope
   @ REPL[6]:1

julia> @which ideal(R, [x]) == ideal(S, [u])
==(I::MPolyIdeal, J::MPolyIdeal)
     @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-ideals.jl:992
```

### DIFF
--- a/docs/src/CommutativeAlgebra/ideals.md
+++ b/docs/src/CommutativeAlgebra/ideals.md
@@ -167,13 +167,13 @@ is_monomial(f::MPolyRingElem)
 ### Containment of Ideals
 
 ```@docs
-is_subset(I::MPolyIdeal{T}, J::MPolyIdeal{T}) where T
+is_subset(I::MPolyIdeal, J::MPolyIdeal)
 ```
 
 ### Equality of Ideals
 
 ```@docs
-==(I::MPolyIdeal{T}, J::MPolyIdeal{T}) where T
+==(I::MPolyIdeal, J::MPolyIdeal)
 ```
 
 ### Ideal Membership
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -66,11 +66,11 @@ end
 
 # elementary operations #######################################################
 @doc raw"""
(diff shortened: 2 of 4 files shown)


thofma (approved):

joschmitt:
I also added a `check_base_rings` to `is_subset(::MPolyIdeal, ::MPolyIdeal)`; this function would work for rings which "look the same" but aren't (which is quite convenient but also leads to more problems than it solves in my opinion).

fingolfin (approved):
Thank you, good idea!

PR #2938, opened 2023-10-19
Avoid various obsolete function and type names

fingolfin:


afkafkafk13 (approved):
mostly changing Localization to lower case (and some minor leftovers from the renaming)

tedious, but necessary stuff -- thank you for taking care of this

PR #2940, opened 2023-10-19
Revamp `graded_polynomial_ring`

fingolfin:
It now accepts exactly the same variants for specifying variables as `polynomial_ring` (by passing them on unchanged).

Moreover, the weight vector can now be passed as a keyword argument. At some point we could drop support for passing it as final argument.

Finally, weight vectors can now be specified as `AbstractVector` instances, making them slightly more usable.

So this is now valid:

    R, x, y = graded_polynomial_ring(QQ, :x => 1:3, :y => (1:2, 1:2); weights=1:7)

Once this is merged I'll update #2924.

The current plan is to follow up to this change in a *future PR* by adding another way for specifying the weights: instead of a flat vector assigning weights to the variables in sequential order, it will then be made possible to specify a weight *tuple* with entries that correspond to the "variable blocks" (such as `:x => 1:3` or `:y => (1:2, 1:2)` or `[:a, :b, :c]`). For each block one then can specify either...
- an integer or abelian group element: all variables in the block will get that weight
- a vector/matrix/higher array of integers/group elements of length/shape matching the length/shape of the variables in that block
- maybe more... tell me your wishes

This will then make it possible to do e.g. this, with three variable blocks:

    R, (a,b), x, y = graded_polynomial_ring(QQ, [:a, :b], :x => 1:3, :y => (1:2, 1:2);
                                            weights=([1,2], 3, [4 5 ; 6 7]))

In that example,
- `weight(a) == 1`
- `weight(b) == 2`
- `weight(x[i]) == 3` for $1\leq i\leq 3$
- `weight(y[i,j])` is 4, 5, 6 or 7, depending on the position in the weight matrix

CC @mgkurtz (this should be mostly orthogonal to your work, except maybe for the macro bit...)

### DIFF
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -3,7 +3,7 @@
   R::S
   D::GrpAbFinGen
   d::Vector{GrpAbFinGenElem}
-  lt
+  lt::Any
   function MPolyDecRing(R::S, d::Vector{GrpAbFinGenElem}) where {S}
     @assert length(d) == ngens(R)
     r = new{elem_type(base_ring(R)), S}()
@@ -107,7 +107,7 @@ function decorate(R::MPolyRing)
 end
 
 @doc raw"""
-    grade(R::MPolyRing, W::Vector{<:IntegerUnion})
+    grade(R::MPolyRing, W::AbstractVector{<:IntegerUnion})
 
 Given a vector `W` of `ngens(R)` integers, create a free abelian group of type `GrpAbFinGen` 
 given by one free generator, and convert the entries of `W` to elements of that group. Then 
@@ -132,7 +132,7 @@ julia> T, (x, y, z) = grade(R)
 (Graded multivariate polynomial ring in 3 variables over QQ, MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}[x, y, z])
 ```
 """
-function grade(R::MPolyRing, W::Vector{<:IntegerUnion})
+function grade(R::MPolyRing, W::AbstractVector{<:IntegerUnion})
   @assert length(W) == ngens(R)
   A = abelian_group([0])
   set_attribute!(A, :show_elem => show_special_elem_grad) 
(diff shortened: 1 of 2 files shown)


src/Rings/mpoly-graded.jl
```diff
@@ -377,18 +353,15 @@ end
...
-type `MPolyDecRing`, together with the vector of variables.
-
-    graded_polynomial_ring(C::Ring, V; ordering=:lex)
+coefficient ring `C` and variables as described by `args...` (using the exact
+same syntax as `polynomial_ring`), and [`grade`](@ref) this ring
+according to the data provided by the keyword argument `weights`.
+Return the graded ring as an object of type `MPolyDecRing`, together with the variables.
 
-As above, where the grading is the standard $\mathbb Z$-grading.
+If `weights` is omitted then all variables are graded with weight `1`.
```
afkafkafk13:
```suggestion
If `weights` is omitted the grading is the standard $\mathbb Z$-grading, i.e. all variables are graded with weight `1`.

```

afkafkafk13 (approved):
Looks good to me.

Do we want to add a line somewhere in the doumentation saying that creating a polynomial ring and grading it leads to a mathematically identical object as graded_polynomial_ring? 
(Might be an important details for scripting users)

src/Rings/mpoly-graded.jl
```diff
@@ -238,14 +233,9 @@ Return `true` if `R` is $\mathbb Z$-graded, `false` otherwise.
 
 # Examples
 ```jldoctest
-julia> R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
+julia> S, (x, y, z) = graded_polynomial_ring(QQ, ["x", "y", "z"]; weights = [1, 2, 3])
 (Multivariate polynomial ring in 3 variables over QQ, QQMPolyRingElem[x, y, z])
```
afkafkafk13: Please adjust output to match the new setting. Otherwise doctests won't pass.

Issue #2930, opened 2023-10-17
`proj_space` vs `projective_space`


lkastner:
Currently both of these appear and I do not know why. Yet, for `affine_space` there is no `aff_space`. It is definitely bad style to have both and I think that `projective_space` is the one that makes more sense (otherwise we would have to change a lot of stuff)
```
[lars@nomurai src]$ cat exports.jl | grep projective
export complex_projective_plane
export is_projective
export is_projective_space
export morphism_of_projective_schemes
export projective_general_linear_group
export projective_geometry
export projective_omega_group
export projective_orthogonal_group
export projective_plane
export projective_scheme
export projective_space
export projective_special_linear_group
export projective_special_orthogonal_group
export projective_special_unitary_group
export projective_symplectic_group
export projective_unitary_group
export real_projective_plane
export weighted_projective_space
```
Through `git blame` I arrived at b02ea952243 where @HechtiDerLachs changed most of the `projective_space` to `proj_space`. The [styleguide](https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/) even uses `projective_space` as an example. Also see https://github.com/oscar-system/Oscar.jl/issues/922.

Hence I would ask someone familiar with the AlgebraicGeometry section to get rid of `proj_space` and introduce the suitable variant of `projective_space`.

lkastner:
Also there are some `export` statements left in these files? @HereAround why did you leave them there when you were working on the export stuff?

HereAround:
The following may or may not be relevant, but I wanted to mention it:

There is the notion of "projectivization of the total space of a vector bundle". If everything works correct, then this uses the constructor `proj` in the toric setting, and so may be borderline relevant.

As for `export` statements (in which files btw?), there is a good chance that I simply missed them. I.e., this is a mistake and should be corrected.

I support your statement that there should be a unified command across all of the algebraic geometry, and given how close for instance the above mentioned `proj` (for projectivization...) is, I also do support `projective_space` as the unified wording/command for this.

lkastner:
> As for `export` statements (in which files btw?), there is a good chance that I simply missed them. I.e., this is a mistake and should be corrected.

The file is
https://github.com/oscar-system/Oscar.jl/blob/73497a75ae2a1bda3a0a52d32c6f906c2097d4b6/src/AlgebraicGeometry/Miscellaneous/basics.jl#L13

It is in a module (`Geometry`) which is why you left it in.

lkastner:
[...]

PR #2945, opened 2023-10-21
Fix show(io::IO, P::ProjSpc)

fingolfin:
I guess this type is supposed to be removed soon anyway, but
since I just stumbled over this printing issue...


HereAround:
This seems related to https://github.com/oscar-system/Oscar.jl/issues/2930.

simonbrandhorst (approved):

PR #2948, opened 2023-10-23
Clean up PlaneCurve tests

fingolfin:
I just was about to update #2924 with the nitpick fixes after all (it wasn't that much, it turns out) but our actions crossed, @afkafkafk13 :-). Anyway, arguably better that way because the other one already had the CI pass...


afkafkafk13 (approved):
Indeed, these were all the remaining instances.

PR #2952, opened 2023-10-24
Refactor graphs, streamline some polyhedral constructors

lkastner:
- Reduce duplication
- Let `{add,rem}_edge!` return bools whether operation succeeded.
- `polyhedral_fan` and `normal_toric_variety` now use the same order of arguments as `polyhedral_complex`. @YueRen 
- Due to `Matrix{Int}` being used for both encoding incidences or rays in `normal_toric_variety` it was impossible to cleanly deprecate all old constructors.
- Rename `pm_tdivisor` to `pm_object` as this is used everywhere else.

### DIFF
--- a/src/Combinatorics/Graphs/functions.jl
+++ b/src/Combinatorics/Graphs/functions.jl
@@ -63,12 +63,12 @@ Undirected graph with 2 nodes and the following edges:
 """
 graph_from_adjacency_matrix(::Type, G::Union{MatElem, Matrix})
 
-function graph_from_adjacency_matrix(::Type{Undirected}, G::Union{MatElem, Matrix})
+function graph_from_adjacency_matrix(::Type{T}, G::Union{MatElem, Matrix}) where {T <: Union{Directed, Undirected}}
   n = nrows(G)
   @req nrows(G)==ncols(G) "not a square matrix"
-  g = Graph{Undirected}(n)
+  g = Graph{T}(n)
   for i in 1:n
-    for j in 1:i-1
+    for j in 1:(T==Undirected ? i-1 : n)
       if isone(G[i,j])
         add_edge!(g, i, j)
       else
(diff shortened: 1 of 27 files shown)


lgoettgens:
> * Let `{add,rem}_edge!` return bools whether operation succeeded.

Are there situations where this does not succeed? If yes, that would imply that I need to check the return value every time to be sure, right?

src/Combinatorics/Graphs/functions.jl
```diff
@@ -79,68 +79,65 @@ function graph_from_adjacency_matrix(::Type{Undirected}, G::Union{MatElem, Matri
...
-  return g
-end
 
 _has_node(G::Graph, node::Int64) = 0 < node <= nv(G)
 
 @doc raw"""
     add_edge!(g::Graph{T}, s::Int64, t::Int64) where {T <: Union{Directed, Undirected}}
 
 Add edge `(s,t)` to the graph `g`.
+Return `true` if operation was successful, `false` otherwise.
```
fingolfin: What does it mean that it was "not successful"? That the edge was already there? Or can it fail for other reasons? Which?

If the former, then I would suggest to make that explicit, e.g.:
```suggestion
Return `true` if there was no edge there before, `false` otherwise.
```

lkastner:
> > * Let `{add,rem}_edge!` return bools whether operation succeeded.
> 
> Are there situations where this does not succeed? If yes, that would imply that I need to check the return value every time to be sure, right?

yes

YueRen:
Ah, I see that you already changed `groebner_fan` for me, much appreciated!

src/Combinatorics/Graphs/functions.jl
```diff
@@ -214,9 +220,11 @@ julia> nv(g)
...
 """
 function add_vertices!(g::Graph{T}, n::Int64) where {T <: Union{Directed, Undirected}}
-    for i = 1:n
-        add_vertex!(g)
-    end
+  result = 0
+  for i = 1:n
+    result += add_vertex!(g) ? 1 : 0
+  end
+  return result
```
benlorenz:
```suggestion
  return count(_->add_vertex!(g), 1:n)
```

benlorenz (approved):

PR #2936, opened 2023-10-18
Experimental: Add `BasisLieHighestWeight`

lgoettgens:
This is another try to https://github.com/oscar-system/Oscar.jl/pull/2115 and https://github.com/oscar-system/Oscar.jl/pull/2402 (now on my fork so I can manage access rights myself).

Background: This is code written mainly by @BenWilop (student assistant in Aachen) that is needed for and referenced in the book chapter of @gfourier and Xin Fang. I will be helping them a bit to "oscarify" everything.

@BenWilop please continue working on this branch. If you encounter any access rights issues, please write me an email.

~@ everyone please wait with reviews and comments for a bit, this PR is currently just to enable collaboration.~

PR #2953, opened 2023-10-24
Rename `isweighted` -> `is_weighted`

lgoettgens:
This is used in #2936, and I think it should follow our naming scheme. As the function was not exported before, I think we can omit the deprecation.

### DIFF
--- a/src/AlgebraicGeometry/Miscellaneous/basics.jl
+++ b/src/AlgebraicGeometry/Miscellaneous/basics.jl
@@ -45,7 +45,7 @@ function weights(P::ProjSpc)
   return Int[x[1] for x = P.Rx.d]
 end
 
-function isweighted(P::ProjSpc)
+function is_weighted(P::ProjSpc)
   return !all(x->isone(x[1]), P.Rx.d)
 end
 
@@ -173,7 +173,7 @@ function normalize!(a::ProjSpcElem{ZZRingElem})
 end
 
 function Base.hash(a::ProjSpcElem, u::UInt=UInt(123432))
-  if isweighted(parent(a))
+  if is_weighted(parent(a))
     return u
   end
   normalize!(a)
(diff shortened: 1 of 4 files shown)


src/AlgebraicGeometry/Miscellaneous/basics.jl
```diff
@@ -45,7 +45,7 @@ function weights(P::ProjSpc)
   return Int[x[1] for x = P.Rx.d]
 end
 
-function isweighted(P::ProjSpc)
+function is_weighted(P::ProjSpc)
   return !all(x->isone(x[1]), P.Rx.d)
```
fingolfin, line 49: I find this method problematic -- why is a projective space in which all weights are 1 "not weighted" ?

src/Rings/orderings.jl
```diff
@@ -124,7 +124,7 @@ function _unique_var_indices(a::AbstractVector{<:MPolyRingElem})
   return z
 end
 
-function isweighted(ord::Symbol)
+function is_weighted(ord::Symbol)
```
fingolfin: I find this method problematic: in how far is a *symbol* weighted? This is perhaps OK for an internal helper, but not for something we export. I would suggest to call this `_is_weighted`

fingolfin (changes requested):
I don't agree that we should export `is_weighted` at this point, at least not with *those* methods. And also not without any docstrings on them.

src/Rings/orderings.jl
joschmitt: I have the impression that this function should be removed completely. It is not used (apparently it was used originally for a `show` method) and monomial orderings should not be represented by symbols nowadays. So if the functionality is needed, there should be a `is_weighted(::MonomialOrdering)` in my opinion.

It looks like one could easily implement this with
```
is_weighted(::MonomialOrdering) = false
is_weighted(::WSymbOrdering) = true
```
but no guarantee.

lgoettgens: The reason why I opened this PR is that I need this for #2936. Users should be able to provide an ordering via symbol as an input there. So please don't remove

fingolfin (approved):

PR #2832, opened 2023-09-20
yr/tropicalBasics

YueRen:
Do not merge yet, this is just a test to see whether everything is fine.

PR #2955, opened 2023-10-25
Polyhedral: add hash method for Cone + Polyhedron

benlorenz:
These have a custom `==` function, so they need a hash function (#2222).
We cannot reasonably hash the polytopes as we want to avoid triggering a convex hull computation. And even if we did, that would not give us a normal form that we could use for the hash. So we just hash the type and the ambient dimension.

Also add `ambient_dim` for LP and MILP.

x-ref: #2832

### DIFF
--- a/src/PolyhedralGeometry/Cone/constructors.jl
+++ b/src/PolyhedralGeometry/Cone/constructors.jl
@@ -91,6 +91,13 @@ function ==(C0::Cone{T}, C1::Cone{T}) where T<:scalar_types
     return Polymake.polytope.equal_polyhedra(pm_object(C0), pm_object(C1))
 end
 
+# we do not want to force a convex hull computation and even if we did, that
+# would not give a normal form
+function Base.hash(x::T, h::UInt) where {T <: Cone}
+  h = hash(ambient_dim(x), h)
+  h = hash(T, h)
+  return h
+end
 
 @doc raw"""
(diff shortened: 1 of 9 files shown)


src/PolyhedralGeometry/Cone/constructors.jl
```diff
@@ -91,6 +91,13 @@ function ==(C0::Cone{T}, C1::Cone{T}) where T<:scalar_types
     return Polymake.polytope.equal_polyhedra(pm_object(C0), pm_object(C1))
 end
 
+# we do not want to force a convex hull computation and even if we did, that
+# would not give a normal form
```
fingolfin: This explanation seems to be missing something... it seems to presuppose that the reader would expect the hash to compute a convex hull? How about something like this:
```suggestion
# For a proper hash function for cones we should use a "normal form",
# which would require a potentially expensive convex hull computation
# (and even that is not enough). But hash methods should be fast, so we
# just consider the ambient dimension and the precise type of the cone.
```

src/PolyhedralGeometry/Polyhedron/constructors.jl
```diff
@@ -137,6 +137,14 @@ function ==(P0::Polyhedron{T}, P1::Polyhedron{T}) where T<:scalar_types
     Polymake.polytope.equal_polyhedra(pm_object(P0), pm_object(P1))
 end
 
+# we do not want to force a convex hull computation and even if we did, that
+# would not give a normal form
```
fingolfin: ditto

src/PolyhedralGeometry/Cone/constructors.jl
benlorenz: thanks, adjusted.

lkastner (approved):

Issue #2693, opened 2023-08-17
Master CI failing since JLLWrappers.jl v1.4.2


lgoettgens:
Starting at https://github.com/oscar-system/Oscar.jl/actions/runs/5882459447, the OSCAR CI fails on all julia versions, while on the previous master commit it was mostly fine (or better to say "usual"; see https://github.com/oscar-system/Oscar.jl/actions/runs/5875640056).
The changes #2682 do not seem to be the root of the failure, but an update of Singular (see the error log below).
The relevant version changes:
```
Singular.jl               v0.18.9 -> v0.18.10
libsingular_julia_jll     v0.36.0+0 -> v0.39.0+0
Singular_jll              v403.203.202+0 -> v403.206.600+0
```

From https://github.com/oscar-system/Oscar.jl/actions/runs/5882459447/job/15953026725
```julia
ERROR: LoadError: libparse() is deprecated, use the non-do-block form
Stacktrace:
  [1] depwarn(msg::String, funcsym::String; force::Bool)
    @ Base ./deprecated.jl:124
  [2] depwarn(msg::String, funcsym::String)
    @ Base ./deprecated.jl:121
  [3] libparse(f::Function; adjust_PATH::Bool, adjust_LIBPATH::Bool)
    @ Singular_jll ~/.julia/packages/JLLWrappers/nEaFF/src/products/executable_generators.jl:20
  [4] libparse(f::Function)
    @ Singular_jll ~/.julia/packages/JLLWrappers/nEaFF/src/products/executable_generators.jl:19
  [5] (::Singular.Setup.var"#2#7"{Dict{Symbol, Vector{Tuple{String, String}}}, Vector{String}, String})()
    @ Singular.Setup ~/.julia/packages/Singular/NnGm6/src/setup.jl:43
  [6] cd(f::Singular.Setup.var"#2#7"{Dict{Symbol, Vector{Tuple{String, String}}}, Vector{String}, String}, dir::String)
    @ Base.Filesystem ./file.jl:112
  [7] regenerate_libraryfuncdictionary(prefixpath::String)
    @ Singular.Setup ~/.julia/packages/Singular/NnGm6/src/setup.jl:39
  [8] top-level scope
    @ ~/.julia/packages/Singular/NnGm6/src/setup.jl:54
  [9] include(mod::Module, _path::String)
    @ Base ./Base.jl:457
 [10] include(x::String)
[...]

PR #2956, opened 2023-10-26
Add a single ci job per group with depwarn=error

lgoettgens:
The idea came first up in https://github.com/oscar-system/Oscar.jl/issues/2693#issuecomment-1682050751, and now with #2913 present, I think this is a good time to talk about this proposal.
And due to #2913 not merged, we have the chance to test this PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -135,7 +135,7 @@ jobs:
       - name: "Run doctests"
         run: |
           julia ${{ matrix.julia-version == '1.9' && '--code-coverage' || '' }} \
-            --project=docs --depwarn=error --color=yes -e'
+            --project=docs --depwarn=${{ matrix.depwarn == 'depwarn=error' && 'error' || 'no' }} --color=yes -e'
               using Documenter
               include("docs/documenter_helpers.jl")
               using Oscar
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -91,7 +91,12 @@ jobs:
           - 'nightly'
         os:
           - ubuntu-latest
+        depwarn: [ '' ]
         include:
+          # Add a single job with deprecation errors on the most recent stable julia version
+          - julia-version: '1.9'
+            os: ubuntu-latest
+            depwarn: 'error'
           # Add macOS jobs (not too many, the number we can run in parallel is limited)
           - julia-version: '1.9'
             os: macOS-latest
@@ -120,7 +125,7 @@ jobs:
         with:
           annotate: ${{ matrix.julia-version == '1.9' }}
           coverage: ${{ matrix.julia-version == '1.9' }}
-          depwarn: error
+          depwarn: ${{ matrix.depwarn != '' && matrix.depwarn || 'no' }}
       - name: "Setup package"
         run: |
           julia --project=docs --color=yes -e '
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -30,6 +30,15 @@ jobs:
         os:
```
lgoettgens, line 30:
```suggestion
        depwarn: ['']
        os:
```
Is this maybe needed for the new jobs to get created?

lgoettgens:
With this change (if it works), the large part of the testsuite ignores deprecation warnings. These are only catches by the new few jobs.
This helps us to still have the large part of the testsuite working when a dependency deprecated something.

thofma:
OK, I see. Would also be fine with me. Compared to testing deprecations in the downstream tests via OscarDevTools, the solution here is better. What do you think @benlorenz?

benlorenz:
I think separating the deprecations check is better than enabling depwarn for OscarDevTools.

But we need to run doctests with depwarn as well.

.github/workflows/CI.yml
thofma: You could also just add a separate job, if you don't want to do this dance.

benlorenz: > Is this maybe needed for the new jobs to get created?

Yes that seems likely, you can use the `include` block for two things: add extra variables to existing (matching) jobs or to add extra jobs. Which one of these is used depends on whether all variables are already defined in the matrix definition, more details [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations).

lgoettgens:
The jobs are now named correctly. But I expected the depwarn=error tests to fail due to #2913 not merged on this branch.
@benlorenz do you have an idea what's going on?

benlorenz:
The tests run on a temporary merge commit with the current master, so this includes the fix now. The test run before the job / doctest adaptions did fail due to the deprecations for julia 1.9: https://github.com/oscar-system/Oscar.jl/actions/runs/6649926182/job/18069184829

Maybe we can use something like `depwarn: 'depwarn'` for the matrix and then in the job:
```
${{ matrix.depwarn == 'depwarn' && 'error' || 'no' }}
```
This way the label for the job would contain `depwarn` instead of just `error`.

And we should probably adjust either the codecov upload count or whether the new jobs do uploads. The codecov page for the current commit shows 6 uploads: https://app.codecov.io/gh/oscar-system/Oscar.jl/commit/ac1a1855de32b2f104bff0f088105292c5e61393.

lgoettgens:
I decided to name it `depwarn=error` to make clear that deprecations actually error and not just warn.

Furthermore, I disabled codecov for these new jobs. If we expect the to fail kind of often, it would be senseless to add the to the codecov config and wait infinitely for the comments. And they shouldn't bring any new covered lines, or at least not a significant amount.

benlorenz:
The run command for the doctests still has an unconditional `--depwarn=error` (currently line 138):
https://github.com/oscar-system/Oscar.jl/pull/2956/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R138

(The `with` block a few lines above is only for running the normal tests on macos, which is done in the doctests job because we only have 5 parallel macos runners)

lgoettgens:
Oh yeah, true. Thanks for spotting! Should be fixed now.

benlorenz (approved):
thanks!

PR #2958, opened 2023-10-26
Hypercomplexes

HechtiDerLachs:
@jankoboehm : As discussed today. The relevant files are exclusively in `experimental/HyperComplexes`. The rest is in #2935 and can be ignored.


HechtiDerLachs:
@jankoboehm : Did you have time to have a look at this? IMHO it would make sense to subsume double complexes under `AbsHyperComplex` and make them implement the hypercomplex interface. Then we could also introduce a concrete wrapper type for double complexes in parallel to the existing `DoubleComplexOfMorphisms`, which uses a hypercomplex in the background. 

The same could then be done for ordinary `ComplexOfMorphisms` and even `FreeResolution`s. But being defined in Hecke, these can not yet be derived from `AbsHyperComplex`. So we probably do not touch upon this for the moment. What we could do nevertheless, is to rebase `FreeResolution`s on a concrete wrapper type for one-dimensional complexes which is based on the hypercomplexes here.

HechtiDerLachs:
Freshly rebased for better readability.

jankoboehm:
I like it. It does what we discussed. Before interlinking it with double complexes, complexes or resolutions, I think we should get it it. I would then like to test it out in some use cases, also in the one-dimensional setting. We can then discuss carefully about interlinking it with what is there, ensuring compatibility. The goal is that the user does not have to struggle with different types of complexes not interoperating, or operating inconsistently.

HechtiDerLachs:
Rebased again after recent merge of the double complexes.

HechtiDerLachs:
@jankoboehm : As discussed this morning. Hypercomplexes fused with the double complexes. For the fun of it, I also added a concrete type for double complexes which is wrapping a hyper complex. In the example in the test file it seems to perform slightly worse than the native double complex implementation. But since the latter is kept and used, I don't see a problem here.

HechtiDerLachs:
Can this be merged then? Or should we talk about it again on Wednesday?

fieker (approved):

PR #2959, opened 2023-10-27
Polyhedral{Fan,Complex}: improve / add constructors from cones / polytopes

benlorenz:
cc: @lkastner @YueRen 

The construction from `maximal_polyhedra` of an existing polyhedral complex will shortcut and just create a copy but any other list of polyhedra will trigger the `check_fan_objects` in polymake and test for the intersection property.
(Similarly for fans)


YueRen:
Looks good, but is it possible to introduce an optional flag that skips `check_fan_objects`?

I have many applications where I construct a `Vector{Polyhedron}` or `Vector{Cone}` and I know that the input objects are the maximal cells of a well-defined polyhedral complex or fan.

benlorenz:
We can't simply skip `check_fan_objects` because this is needed to canonicalize and identify common rays / vertices. Maybe we can hook into something else to achieve this.

Is there lineality involved? Where do these vectors of cones / polyhedra arise, and is there a common list of vertices / rays that these are built from?

YueRen:
Hm, okay.  Maybe a more concrete question (not necessarily related to this pull request) would be: If I have a list of maximal polyhedra / cones of a polyhedral complex / fan (manually created, possibly with lineality), what is the best way to construct a `PolyhedralComplex` / `PolyhedralFan` out of them?  Via the constructors that use `IncidenceMatrix`?

I don't have a common list of vertices and rays (they were all constructed via inequalities and equations), but it shouldn't be hard (compared to the computations that gave rise to the list) to construct one.

benlorenz:
Right now the most efficient way would be to go via IncidenceMatrix.

The upcoming polymake 4.11 will have a few more functions in addition to `check_fan`: to properly deal with Polytopes and return a PolyhedralComplex and also to skip the intersection checks, just assembling the fan / complex. Once that is built I will update this PR.

lkastner (approved):

benlorenz:
There are now:
```julia
polyhedral_complex(polytopes::AbstractVector{Polyhedron{T}}; non_redundant::Bool=false) 
polyhedral_fan(cones::AbstractVector{Cone{T}}; non_redundant::Bool=false)
```
The `non_redundant` flag is internal (undocumented), similar to the other constructors.
With that flag set to `true` it will only do redundancy elimination for the rays (to find the incidences) and remove non-maximal cones/polytopes.
When the flag is `false` (the default) polymake will also check whether the lineality spaces match and compute intersections for all pairs to check whether it is a face of each cone.

PR #2960, opened 2023-10-27
add `od_from_atlas_group` and helper functions

ThomasBreuer:
- the new optional `min_char` argument in `isomorphic_group_over_finite_field` is used to guarantee that the reduction of an irreducible module is irreducible


lgoettgens:
This PR seems to be kind of broken in the sense that it changes over 100 files (what I think is not intended).
Could you try to rebase it, @ThomasBreuer?

ThomasBreuer:
@lgoettgens thanks for the hint.

Puh, I do not understand why these many commits from others are coming up here; my only explanation would be that this happened *because* of a rebase.

ThomasBreuer:
This pull request got corrupted. I have copied the relevant commits to the new pull request #3005.

PR #2963, opened 2023-10-27
Make toric divisors compatible with Weil divisors on schemes

HechtiDerLachs:
@HereAround @lkastner : After talking to Lars a few minutes ago, I could now produce a first running version. Further improvements will follow.


experimental/Schemes/WeilDivisor.jl
```diff
@@ -672,3 +672,63 @@ function order_on_divisor(
...
+  wd = WeilDivisor(X, ZZ, IdDict{IdealSheaf, ZZRingElem}(II => zero(ZZ)))
+  return wd
+end
+
+# Compute the ideal sheaves of the rays of `fan(X)` according to the 
+# "orbit-cone-correspondence", Theorem 3.2.6 in Cox-Little-Schenk.
+#
+# We would like to use Equation (3.2.7) from page 121 for this.
+#@attr Vector{<:IdealSheaf} function _ideal_sheaves_of_rays(X::NormalToricVariety)
+function _ideal_sheaves_of_rays(X::NormalToricVariety; check::Bool=false)
```
HechtiDerLachs: @lkastner : This is the procedure that caused the trouble. It seems that the code below is producing the correct results and I was aiming at what we talked about. Maybe you see some possibilities for improvements? Or corrections?

HechtiDerLachs:
The tests for the toric divisors run for me locally. However, the tests in `test/AlgebraicGeometry/Schemes/WeilDivisors.jl` are broken due to unknown causes. But they are currently also broken on master with the same error message, so I presume it's a different issue. 

Edit: It turned out to be a problem on my local machine.

experimental/Schemes/WeilDivisor.jl
```diff
@@ -660,15 +705,83 @@ end
...
+    for U in affine_charts(X)
+      if !(tau in cone(U))
+        ideal_dict[U] = ideal(OO(U), one(OO(U)))
+        continue
+      end
+      sigma_dual = weight_cone(U)
+      hb = hilbert_basis(sigma_dual)
+      selection = [v for v in hb if !(-v in tau_dual)]
+      x = gens(OO(U))
+      ideal_dict[U] = ideal(OO(U), [x[i] for i in 1:length(x) if !(-hb[i] in tau_dual)])
```
HereAround: Can selection be removed? Or alternatively, can it be used in creating `ideal_dict[U]`?

experimental/Schemes/WeilDivisor.jl
```diff
@@ -660,15 +710,124 @@ end
...
+  X = scheme(td)
+  generating_divisors = _torusinvariant_weil_divisors(X; check)
+  result = sum(a*D for (a, D) in zip(coefficients(td), generating_divisors))
+  set_attribute!(td, :underlying_divisor=>result)
+  return result
+end
+
+# Compute the ideal sheaves and Weil divisors of the rays of `fan(X)` 
+# according to the "orbit-cone-correspondence", Theorem 3.2.6 in 
+# Cox-Little-Schenk.
```
HechtiDerLachs: @lkastner : @HereAround and I had a look at your comments in slack today. We did not quite get your point there. Instead I made an attempt to prove that my hack is correct, see below. 

Be aware that the method is **only** used for the torus invariant prime divisors. So things like 3*ray do not occur here. 

Could you confirm whether or not everything is correct now? If you have questions, please reach out to us through the usual channels.

HechtiDerLachs:
From my side this is good to go. Can someone approve?

HereAround:
Thank you @HechtiDerLachs! Since you, @lkastner and I were discussing this in slack, I left a few minor nitpicks there regarding the code itself.

Regarding the mathematical sanity, I am starting to feel convinced that this code works. Indeed, @HechtiDerLachs and I discussed this at length recently. However, I also understand that @lkastner wants to discuss more on this on Friday (December 1). So, while good to go on my end too, I suggest we wait until this Friday for feedback by @lkastner. Makes sense?

HereAround:
As discussed on slack, I have prepared the code in this PR, which will eventually be moved into `src/AlgebraicGeometry/ToricVarieties`.

I have also taken the liberty to rebase to the latest master.

experimental/Schemes/ToricDivisors/constructors.jl
```diff
@@ -0,0 +1,2 @@
+# Method ambiguity requires the following
+Base.:*(c::Int, td::ToricDivisor) = toric_divisor(toric_variety(td), [ZZRingElem(c)*x for x in coefficients(td)])
```
HereAround, line 2: Also as mentioned on slack, I would like to understand the cause for this better.

There is already a similar implementation, which accepts `c` a subtype of `IntegerUnion`. If my understand is correct, `IntegerUnion` is the union of `Int` and (what used to be) `fmpz` (now `ZZRingElem`?). If indeed true, then the above line should not be needed.

I noticed one or two similar lines, which could be adjusted/investigated similarly.

thofma: What is the signature of the similar implementation? Or what is the full error message for the ambiguity error?

thofma: What is the signature of the similar implementation? Or what is the full error message for the ambiguity error?

HereAround: The existing implementation looks as follows (`src/AlgebraicGeometry/ToricVarieties/ToricDivisors/constructors.jl`, line 99):
```julia
Base.:*(c::T, td::ToricDivisor) where {T <: IntegerUnion} = toric_divisor(toric_variety(td), [ZZRingElem(c)*x for x in coefficients(td)])
```
Regarding the full error message, maybe @HechtiDerLachs still remembers this?

thofma: This looks rather innocent, but I can't say more without the error message.

HereAround: @HechtiDerLachs and I discussed this yesterday with @fingolfin . I think it is now clear what needs to be done. @HechtiDerLachs is looking into this.

HereAround:
(Rebased to latest master.)

HereAround:
We (that is @lkastner , @HechtiDerLachs and I) are discussing this PR on slack. Do not merge yet.

lkastner (approved):

HereAround (approved):
Thank you @HechtiDerLachs .

PR #2964, opened 2023-10-27
init: warning for outdated manifest

benlorenz:
This prints a warning if a git clone is updated (and the Project.toml changes) but the Manifest was not re-resolved:
```
Version 0.14.0-DEV ... 
 ... which comes with absolutely no warranty whatsoever
Type: '?Oscar' for more information
(c) 2019-2023 by The OSCAR Development Team
┌ Warning: Project dependencies might have changed, please run `]up` or `]resolve`.
└ @ Oscar ~/software/polymake/julia/Oscar.jl/src/Oscar.jl:142
```

Pkg will print a similar warning when running `]st` but this doesn't seem to get notices. And it happens quite often that users / developers forget to do `]up`.

cc: @thofma @HechtiDerLachs @lkastner 

PS: This is only active for `Oscar.is_dev` and works only for julia >= 1.8.

### DIFF
--- a/src/Oscar.jl
+++ b/src/Oscar.jl
@@ -135,6 +135,12 @@ function __init__()
 
     add_assertion_scope(:ZZLatWithIsom)
     add_verbosity_scope(:ZZLatWithIsom)
+
+    # Pkg.is_manifest_current() returns false if the manifest might be out of date
+    # (but might return nothing when there is no project_hash)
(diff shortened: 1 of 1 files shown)


lkastner (approved):
I am sorry again.

thofma (approved):

fingolfin (approved):
Good idea! I am sure people will *still* overlook it, but that's on them then ;-)

PR #2965, opened 2023-10-27
add `schur_multiplier`

### LINKED ISSUE #2962: Schur multiplier for groups
Would be great if we could have a `schur_multiplier` function, which could be implemented via `AbelianInvariantsMultiplier` from GAP. For my own usage, I would be happy if this returns (per default) an ordinary abelian group. Maybe we could add an optional type for the return value, as we do for other such constructions.

I am not a group theorist and not married to the name, but I think it is perfect.

Any thoughts @fingolfin @ThomasBreuer?

ThomasBreuer:
and `abelian_invariants` and `abelian_invariants_multiplier`

resolves #2962

### DIFF
--- a/src/Groups/GrpAb.jl
+++ b/src/Groups/GrpAb.jl
@@ -391,7 +391,7 @@ abelian_invariants(::Type{T}, G::GrpAbFinGen) where T <: IntegerUnion =
 
 abelian_invariants(G::GrpAbFinGen) = abelian_invariants(ZZRingElem, G)
 
-function abelian_invariants_multiplier(::Type{T}, G::GrpAbFinGen) where T <: IntegerUnion
+function abelian_invariants_schur_multiplier(::Type{T}, G::GrpAbFinGen) where T <: IntegerUnion
   # By a theorem of I. Schur,
   # the multiplier of an abelian group with elementary divisors
   # n_1 | n_2 | ... | n_k, with k > 1,
@@ -405,7 +405,7 @@ function abelian_invariants_multiplier(::Type{T}, G::GrpAbFinGen) where T <: Int
   return abelian_invariants_of_vector(T, res)
 end
 
-abelian_invariants_multiplier(G::GrpAbFinGen) = abelian_invariants_multiplier(ZZRingElem, G)
+abelian_invariants_schur_multiplier(G::GrpAbFinGen) = abelian_invariants_schur_multiplier(ZZRingElem, G)
 
 nilpotency_class(G::GrpAbFinGen) = (order(G) == 1 ? 0 : 1)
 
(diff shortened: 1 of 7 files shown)


thofma:
Thanks! Why not `abelian_invariants_schur_multiplier` (for consistency)? I have not strong opinion.

ThomasBreuer:
@thofma I have adjusted the name.

thofma (approved):

thofma:
Thanks, looks good to me. Not sure if @fingolfin wants to have a look too?

src/Groups/GrpAb.jl
```diff
@@ -372,6 +372,41 @@ end
 
 is_pgroup_with_prime(G::GrpAbFinGen) = is_pgroup_with_prime(ZZRingElem, G)
 
+function abelian_invariants_of_vector(::Type{T}, v::Vector) where T <: IntegerUnion
```
fingolfin, line 380: add a comment maybe that explains in 1-2 lines what this is about? My guess: `v` is supposed to be a vector of integers. And this function converts between the two different conventions for what "abelian invariants" are:

1. [elementary divisors](https://en.wikipedia.org/wiki/Elementary_divisors), so prime powers or zero -- this is what GAP uses to represent the "abelian invariants" of a group
2. [invariant factors](https://en.wikipedia.org/wiki/Invariant_factor) -- a list of integers $d_1 \mid d_2 \mid \dots \mid d_k$

I linked to Wikipedia here, but e.g. [Encyclopedia of Mathematics](https://encyclopediaofmath.org/wiki/Elementary_divisors) agrees as does Bourbaki. Wikipedia also cites Lang's "Algebra" in support, but he actually only states an "Elementary Divisors Theorem", but does not say what an "elementary divisor" is -- and his theorem produces what above is called "invariant factors"

But not every does, e.g. Hecke has `elementary_divisors` return what I call "invariant factors", and I just read a paper by Frank Lübeck where he also used this convention. 

So as usual, Math is inconsistent as hell... *sigh*

Anyway, back to this function: it seems to convert from format 2 to format 1 above, a comment should say this.

We now have multiple choices: use format 1 or format 2;  and call the result by one or by the other name... Whatever we do, we should make explicit what we mean, given that the terms alone are ambiguous.


Anyway, I am fine with sticking to either naming convention; and also , as long as we make it clear in the documentation which we use (offering a helper function to convert between them of course also would be nice)
As far as I can tell, Magma has two functions, `PrimaryAbelianInvariants` and `AbelianInvariants` which return the first resp. the second kind of sequence.

thofma: I noticed this math inconsistency a while ago, but I am happy to make our system consistent. It does not need to happen in this PR.

ThomasBreuer: O.k., I will add some comments as suggested, and leave it to the discussion of issue #2968 how to change names of functions if necessary.

PR #2966, opened 2023-10-30
Allow choice of target while building systemimage

aaruni96:
Allows one to choose a CPU target while using `Oscar.build()` to build a system image.

[skip ci]


fingolfin (approved):
Fine by me. I trust you tested it :-)

PR #2974, opened 2023-11-02
Johnson solids over number fields

alexej-jordan:
Some johnson solids in polymake are defined with `Float64` scalars because the roots arising during their construction can not be displayed using polymake's `QuadraticExtension` type.

This PR adds precise definitions for these solids with edge length $1$, using Oscar's number field functionality to set coordinates for the $V$-description.


src/PolyhedralGeometry/Polyhedron/johnson.jl
```diff
@@ -0,0 +1,2427 @@
+function _johnson_solid(::Val{9})
+  NF, sr5 = quadratic_field(5)
+  NFy, y = NF["y"]
+  MF, mr = number_field(y^2 - (5 + sr5)//10, "a")
```
fingolfin, line 4: Maybe all these fields and rings should be constructed with `cached=false`? Or perhaps `_johnson_solid` should have a `cached` parameter to control this?

src/PolyhedralGeometry/Polyhedron/johnson.jl
```diff
@@ -0,0 +1,2427 @@
...
+  sre2 = (qre8^2)//2
+  V = [0 0 (2*sre2 + qre8)//4;
+       1//2 1//2 qre8//4;
+       1//2 -1//2 qre8//4;
+       -1//2 1//2 qre8//4;
+       -1//2 -1//2 qre8//4;
+       0 sre2//2 -qre8//4;
+       0 -sre2//2 -qre8//4;
+       sre2//2 0 -qre8//4;
+       -sre2//2 0 -qre8//4]
```
fingolfin, line 36: If I see it right, this is a 2427 lines file which is essentially a "database" consisting of polyhedra given by coordinates over some number fields.

Wouldn't this be a perfect application for @antonydellavecchia's serialization code? I.e. have a data file instead of all this code which needs to be parsed, compiled and optimized for nothing really...

thofma:
It would be great if we could avoid the `hasmethod` code pattern? As far as I understand, the list should be complete and the number of solids can be hardcoded.

I concur with @fingolfin in the use of the serialization infrastructure.

benlorenz:
It seems reasonable to save those objects, there is one issue with this right now as our save-infrastructure currently can't deal with polymake objects containing number field elements. But we could probably save the vertex-matrices instead (maybe a dict mapping number to a (named-)tuple of matrix and description).

If we create these correctly we might be able to store the types that appear multiple times just once.

benlorenz:
obsoleted by #3035

PR #2975, opened 2023-11-02
Adv/add version warning

### LINKED ISSUE #2873: Improve metadata for serialization and Oscar master
Someone send me an "mrdi" file, but I could not read it because things changed and our versions of Oscar were incompatible. While the serialization is still a moving target, maybe we could do the following. Make the metadata more useful by not only recording `{"Oscar":["https://github.com/oscar-system/Oscar.jl","0.13.1-DEV"]}`, but the actual commit (say `"0.13.1-DEV-f9fb3b54ad976b2a1ca65de54d6c5c39c177c34d"`). This should be done 
of course only in case it is not a proper version.

I guess a proper warning would have been nice ("version mismatch for the serialization format") instead of some obscure error, but I think this will be obsolete once things settle and we have possibility to "update".

What do you think @antonydellavecchia?

antonydellavecchia:
Adds a warning when loading a file stored with a dev version of oscar
closes #2873

### DIFF
--- a/src/Serialization/main.jl
+++ b/src/Serialization/main.jl
@@ -148,13 +148,28 @@ end
 ################################################################################
 # Version info
 
-function get_version_info()
+function file_version_info(dict::Dict{Symbol, Any})
+  ns = dict[:_ns]
+  version_info = ns[:Oscar][2]
+  return version_number(version_info)
+end
+
+function version_number(v_number::String)
+  return VersionNumber(v_number)
+end
+
+# needed for older versions 
+function version_number(dict::Dict)
+  return VersionNumber(dict[:major], dict[:minor], dict[:patch])
+end
+
+function file_version_info()
   result = Dict{Symbol, Any}(
     :Oscar => ["https://github.com/oscar-system/Oscar.jl", VERSION_NUMBER]
   )
   return result
 end
-const oscar_serialization_version = get_version_info()
+const oscar_serialization_version = file_version_info()
 
 ################################################################################
 # (De|En)coding types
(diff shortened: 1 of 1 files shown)


src/Serialization/main.jl
```diff
@@ -148,13 +148,37 @@ end
...
+  return VersionNumber(v_number)
+end
+
+# needed for older versions 
+function version_number(dict::Dict)
+  return VersionNumber(dict[:major], dict[:minor], dict[:patch])
+end
+
+function file_version_info()
+  if contains(string(VERSION_NUMBER), "DEV")
```
```suggestion
  if is_dev
```

src/Serialization/main.jl
```diff
@@ -148,13 +148,37 @@ end
...
+function version_number(v_number::String)
+  return VersionNumber(v_number)
+end
+
+# needed for older versions 
+function version_number(dict::Dict)
+  return VersionNumber(dict[:major], dict[:minor], dict[:patch])
+end
+
+function file_version_info()
```
fingolfin: To me this is an unintuitive name. I would expect it to return the version of some file, but it doesn't.

Maybe it can be called something like `current_version_info()` or `serialization_current_version()` or so?

src/Serialization/main.jl
```diff
@@ -146,15 +146,39 @@ function metadata(;args...)
...
+
+# needed for older versions 
+function version_number(dict::Dict)
+  return VersionNumber(dict[:major], dict[:minor], dict[:patch])
+end
+
+function serialization_version_info()
+  if is_dev
+    path = Oscar.oscardir
+    commit_hash = readchomp(`git -C $path log -n 1 --pretty=format:"%H"`)
```
thofma, line 169: I think we should reuse https://github.com/oscar-system/Oscar.jl/blob/master/src/utils/versioninfo.jl#L20-L35 for this. Right, @benlorenz?

thofma: Like `_lookup_git_branch(Oscar.oscardir; commit=true)` should contain the commit somewhere

thofma (approved):
I have not tested it, but looks good, thanks!

thofma:
Hm, there was a comment about reusing the commit sha retrieving functionality we already have.

antonydellavecchia:
> Hm, there was a comment about reusing the commit sha retrieving functionality we already have.

sorry I overlooked. the lookup git branch function returns something slightly different so I thought it was orthogonal.
forexample the output looks like

julia> Oscar._lookup_git_branch(Oscar.oscardir; commit=true)
" - #master, db1bc1afac -- 2023-11-03 11:16:31 +0100"

where I just wanted the hash. But maybe it makesense to do some refactoring?

thofma:
Hm, right, probably too much hassle. Nevermind

PR #2976, opened 2023-11-02
Use `AbstractAlgebra`s `polynomial_ring(ZZ, :x=>1:3)` etc. if it exists

mgkurtz:
The current definition here conflicts with the new version in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1360.

Waiting for new AA/Nemo/Hecke, see #3071

### DIFF
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -1,5 +1,8 @@
 #module MPolyModule
 
+#TODO: reduce = divrem in Nemo. Should be faster - if we have the correct basis
+
+if !applicable(polynomial_ring, ZZ, :x=>1:1)
 ##############################################################################
 #
 # could/ should be in AbstractAlgebra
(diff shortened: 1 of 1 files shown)


src/Rings/mpoly.jl
```diff
@@ -1,5 +1,8 @@
 #module MPolyModule
 
+#TODO: reduce = divrem in Nemo. Should be faster - if we have the correct basis
```
mgkurtz, line 3: I am unsure where this TODO shall go, but it seems not to belong to the code in question.

fingolfin:
I think I'd prefer to put your new `polynomial_ring` code into a "breaking" release of AA, simply because I am afraid of other unintentional side-effects, as it is a big change to a fairly central part of OSCAR...

mgkurtz:
That is fine. Actually I wanted this PR to allow the integration checks to run and to see what else we break.

PR #2977, opened 2023-11-02
Set-Partitions for Experimental

pinguly:
An implementation of set-partitions as used for easy quantum groups, see for example:
- Section 1.2 & 1.3 of https://arxiv.org/abs/1312.3857
- Section 1 of https://arxiv.org/abs/1509.00988
- Section 2 of https://arxiv.org/abs/1609.02321


experimental/SetPartitions/src/AbstractPartition.jl
```diff
@@ -0,0 +1,18 @@
...
+    composition_loops(p, q)[1]
+end
+
+function ⋅(p::T, q::T) where {T <: AbstractPartition}
+    composition_loops(p, q)[1]
+end
+
+function ⊗(p::T, q::T) where {T <: AbstractPartition}
+    tensor_product(p, q)
+end
```
fingolfin: Missing trailing newline

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
+import Base.hash
+import Base.==
+import Base.copy
+
+"""
+ColoredPartition
```
fingolfin:
```suggestion
    ColoredPartition
```

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
...
+
+# Arguments
+- `partition`: Partition object which is generalized with coloring `color_upper_points` and `color_lower_points`
+- `color_upper_points`: color (in [0, 1]) of upper points
+- `color_lower_points`: color (in [0, 1]) of lower points
+"""
+struct ColoredPartition <: AbstractPartition
+    partition::Partition
+    color_upper_points::Array{Int64, 1}
+    color_lower_points::Array{Int64, 1}
```
fingolfin:
```suggestion
    color_upper_points::Vector{Int64}
    color_lower_points::Vector{Int64}
```
This should be adjusted everywhere else, too

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
...
+
+end
+
+function copy(p::ColoredPartition)
+    return ColoredPartition(copy(p.partition), copy(p.color_upper_points), copy(p.color_lower_points))
+end
+
+
+"""
+tensor_product(p::ColoredPartition, q::ColoredPartition)
```
fingolfin:
```suggestion
    tensor_product(p::ColoredPartition, q::ColoredPartition)
```
These lines should be indented by four spaces, which in Markdown indicates a code block.

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
...
+
+This function applies on p tensor product with q (in O(n)).
```
fingolfin:
```suggestion
This function applies on `p` tensor product with `q` (in O(n)).
```

And I'd drop the "(in O(n))" bit here, it is completely unclear what it means anyway (e.g. what is `n`; and is this meant to refer to the complexity? Why is it part of the function description?) If you really want to mention it, I'd add it in a separate paragraph or note or callout.

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
...
+
+# Arguments
+- `p`: Input colored partition
+- `q`: Second input colored partition
+
+# Returns
+- `p` tensor product `q`
```
fingolfin: This way of explaining arguments and return types is highly discouraged except for cases were there really is no good alternative (say because there are so many or so complicated parameters).

Please read https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation for details.

Here, one could write something like this:

> Return the tensor product of two colored partitions `p` and `q`.

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
...
+
+    if tb && isempty(get_upper_points(p))
+        error("Partition has no top part.")
+    elseif !tb && isempty(get_lower_points(p))
+        error("Partition has no bottom part.")
+    end
+
+    ret::Array = [copy(get_upper_points(p)), copy(get_lower_points(p)), copy(p.color_upper_points), copy(p.color_lower_points)]
+
+        if lr
```
fingolfin: The indentation here seems wrong

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,464 @@
...
+    for i in all_partitions
+        push!(tuple_list_all_partitions, i)
+    end
+    for i in vcat(p, tuple_list_all_partitions)
+        all_partitions_by_size = add_partition_to_dict(all_partitions_by_size, i)
+        all_partitions_by_size_top_bottom = add_partition_to_composition_dict(all_partitions_by_size_top_bottom, i)
+    end
+    
+    # while new were found apply on them unary tensor and composition
+    stop_whole::Bool = false
```
fingolfin: This way of writing code with a variable specifier is rather unidiomatic.
```suggestion
    stop_whole = false
```

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,464 @@
...
+            if !(i in already_u)
+                push!(to_unary, i)
+            end
+        end
+
+        # fist phase: all possible combinations of unary operations
+        (stop_whole, all_partitions, already_u, all_partitions_by_size, all_partitions_by_size_top_bottom, trace) = do_unary(to_unary, all_partitions, stop_whole, already_u, max_length, all_partitions_by_size, all_partitions_by_size_top_bottom, trace, spatial_rotation)
+
+        # store pairs that are candidates to get tensor product
+        to_tens::Set{Array} = Set()
```
fingolfin:
```suggestion
        to_tens = Set{Array}()
```
Ideally this would use a fully concrete type like `Set{Vector{Int}}`...

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,464 @@
...
+        if p in keys(trace)
+            println(p, " : ", get(trace, p, -1))
+            for i in get(trace, p, -1)[1]
+                if !(typeof(i) <: AbstractString) && !(i in track)
+                    push!(track, i)
+                end
+            end
+        end
+    end
+end
```
fingolfin: Add trailing newline

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,464 @@
...
+- `p`: list of partitions
+- `n`: size of partitions in constructing category
+- `tracing`: optinal input: activate tracing and get the output (category, trace)
+- `max_artifical`: optional input: allow partitions to grow greater while construction process
+
+# Returns
+- list of all partitions size n constructed from partitions in p
+
+# Examples
+```julia-repl
```
fingolfin: Better us a `jldoctest`, then your example will automatically serve as a test for your code.

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,464 @@
...
+julia> length(construct_category([Partition([1, 2], [2, 1])], 6, true))
+[<Partition([1, 2], [2, 1])> ∩ P(6), Dict{Partition -> Tuple}]
+```
+"""
+function construct_category(p::Array, n::Int, tracing::Bool = false, max_artificial::Int = 0, spatial_rotation::Function = nothing_function() = nothing)
+    
+    @assert length(methods(spatial_rotation)[1].sig.parameters)-1 == 0 || (p isa Array{SpatialPartition})
+
+    # store all candidates found
+    all_partitions::Set{AbstractPartition} = Set()
```
fingolfin:
```suggestion
    all_partitions = Set{AbstractPartition}()
```
and so on -- your current code is unidiomatic

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,464 @@
...
+
+    # all candidates stored in dict from size to partition
+    all_partitions_by_size::Dict{Int, Set{AbstractPartition}} = Dict()
+
+    # all candidates stored in tuple with a dict for top and bottom size
+    all_partitions_by_size_top_bottom::Array = [Dict(), Dict()]
```
fingolfin: The `Array` is a worse type specifier than what is automatically inferred.
```suggestion
    all_partitions_by_size_top_bottom = [Dict(), Dict()]
```

experimental/SetPartitions/README.md
```diff
@@ -0,0 +1,4 @@
+Implementation of set-partitions as used for easy quantum groups, see for example:
+* Section 1.2 & 1.3 of https://arxiv.org/abs/1312.3857
+* Section 1 of https://arxiv.org/abs/1509.00988
+* Section 2 of https://arxiv.org/abs/1609.02321
```
fingolfin: Please follow the format set by `experimental/ExperimentalTemplate/README.md`. Also, hyperlinks in Markdown should be surrounded by `<` and `>`, i.e.
```suggestion
* Section 1.2 & 1.3 of <https://arxiv.org/abs/1312.3857>
* Section 1 of <https://arxiv.org/abs/1509.00988>
* Section 2 of <https://arxiv.org/abs/1609.02321>
```

experimental/SetPartitions/src/AbstractPartition.jl
```diff
@@ -0,0 +1,18 @@
+"""
+AbstractPartition
```
fingolfin:
```suggestion
    AbstractPartition
```

experimental/SetPartitions/README.md
fingolfin: Also I think a little more explanation should be in here, in particular, that this is *not* what ordinary mathematicians would think of when talking about partitions

experimental/SetPartitions/src/AbstractPartition.jl
```diff
@@ -0,0 +1,18 @@
...
+
+Abstract type of Partition for classical-, colored- and spatial Partitions.
```
fingolfin: Hmm, "Abstract type of Partition" doesn't parse as a valid english sentence for me.  Also, if intended to be used as a noun, it should be in lowercase; if it is meant to refer to a type it should be surrounded by backticks.

fingolfin (dismissed):
Thank you for your contribution!

I left some general comments that should be addressed in all similar places they apply.

experimental/SetPartitions/src/ColoredPartition.jl
thofma: Also `Int` instead of `Int64`?

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,164 @@
...
+"""
+function rotation(p::ColoredPartition, lr::Bool, tb::Bool)
+
+    if tb && isempty(get_upper_points(p))
+        error("Partition has no top part.")
+    elseif !tb && isempty(get_lower_points(p))
+        error("Partition has no bottom part.")
+    end
+
+    ret::Array = [copy(get_upper_points(p)), copy(get_lower_points(p)), copy(p.color_upper_points), copy(p.color_lower_points)]
```
thofma:
```suggestion
    ret = (copy(get_upper_points(p)), copy(get_lower_points(p)), copy(p.color_upper_points), copy(p.color_lower_points))
```
I suggest to use a tuple.

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,254 @@
...
+    p.color_upper_points == q.color_lower_points && is_composable(p.partition, q.partition)
+end
+
+function get_upper_points(p::ColoredPartition)
+    p.partition.upper_points
+end
+
+function get_lower_points(p::ColoredPartition)
+    p.partition.lower_points
+end
```
thofma: Better names (better in the sense of consistency with the rest of Oscar) would be `upper_points` and `lower_points`.

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,460 @@
...
+```julia-repl
+julia> length(construct_category([SetPartition([1, 2], [2, 1])], 6))
+105
+julia> length(construct_category([SetPartition([1, 2], [2, 1])], 6, true))
+[<SetPartition([1, 2], [2, 1])> ∩ P(6), Dict{SetPartition -> Tuple}]
+```
+"""
+function construct_category(p::Vector, n::Int, tracing::Bool = false, max_artificial::Int = 0, spatial_rotation::Function = nothing_function() = nothing)
+    
+    @assert length(methods(spatial_rotation)[1].sig.parameters)-1 == 0 || (p isa Vector{SpatialPartition})
```
benlorenz: see above

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,460 @@
...
+                    stop = false
+                    push!(all_partitions, a)
+                    push!(to_unary, a)
+
+                    # call functions which adds the partition a into the right set in the dict
+                    all_partitions_by_size = add_partition_to_dict(all_partitions_by_size, a)
+                    all_partitions_by_size_top_bottom = add_partition_to_composition_dict(
+                            all_partitions_by_size_top_bottom, a)
+                end
+            elseif length(methods(spatial_rotation)[1].sig.parameters)-1 == 1
```
benlorenz: Checking the internal fields of a julia Method object seems very fragile (and already the `[1]` access fails for julia 1.6).
Maybe just use `spatial_rotation::Union{Function,Nothing}=nothing` as function argument and check `spatial_rotation !== nothing` ?

If you really need to check some function argument types you may be able to use like `hasmethod(f, Tuple{SomeType})`.

experimental/SetPartitions/src/AbstractPartition.jl
pinguly: Are the missing newlines there? If not, what is the best way to display them? All the other comments so far should haven been resolved.

thofma:
Newlines look fine.

In OP, there are three arXiv preprints mentioned, but they are not cited in the documentation/code comments. Is this on purpose?

pinguly:
I replaced the three links with a reference to Sebastian's Bachelor's thesis in the README. In his thesis, you can also find the definitions of the different kinds of partitions and some information on the algorithms we implemented.

thofma:
Thanks for the explanation. The failing doctests (https://github.com/oscar-system/Oscar.jl/actions/runs/6893545019/job/18753368231?pr=2977) look genuine.

pinguly:
The doctests seem to work now.

experimental/SetPartitions/src/AbstractPartition.jl
```diff
@@ -0,0 +1,26 @@
+"""
+    AbstractPartition
+
+Abstract type for classical-, colored- and spatial partitions.
+"""
+abstract type AbstractPartition end
+
+function composition(p::T, q::T) where {T <: AbstractPartition}
```
fingolfin: Why not `compose`?

experimental/SetPartitions/README.md
```diff
@@ -0,0 +1,31 @@
...
+We implemented data structures for partitions and corresponding algorithms as described in [[2]](#2). 
+This includes:
+* basic set-partitions and operations on them (e.g. composition, tensor product, involution)
+* variations like colored partitions and spatial partitions
+* enumeration of partitions which can be constructed from a set of generators
+
+In the future, we plan to implement:
+* linear combinations of partitions
+* examples 
+of tensor categories in the framework of [TensorCategories.jl](https://github.com/FabianMaeurer/TensorCategories.jl)
```
fingolfin:
```suggestion
* examples of tensor categories in the framework of [TensorCategories.jl](https://github.com/FabianMaeurer/TensorCategories.jl)
```

experimental/SetPartitions/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+## References
+* <a id="1">[1]</a>
+Teodor Banica and Roland Speicher. Liberation of orthogonal Lie groups.
```
fingolfin: Please use the `.bib` file and the `[BLA](@cite)` syntax provided by DocumenterCitations (just grep our code base for `@cite` to find examples.

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,141 @@
...
+    return ColoredPartition(copy(p.partition), 
+        copy(p.color_upper_points), 
+        copy(p.color_lower_points))
+end
+
+
+"""
+    tensor_product(p::ColoredPartition, q::ColoredPartition)
+
+Return the tensor product of `p` and `q`.
```
fieker, line 172: is it obvious what a tensor prduct of partitions is? or can we get a reference or definition?

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,498 @@
...
+    end
+    return (all_partitions, already_c, stop_whole, all_partitions_by_size, 
+        all_partitions_by_size_top_bottom, trace)
+end
+
+"""
+    construct_category(p::Vector{AbstractPartition}, n::Int, tracing::Bool = false, 
+        max_artifical::Int = 0, spatial_rotation::Union{Function,Nothing}=nothing)
+
+Return a list of all partitions of size `n` constructed from partitions in `p` without using partitions of size greater than max(`n`, maxsize(`p`), `max_artifical`)
```
fieker: To what extend is this a category?

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,498 @@
...
+        return all_partitions_of_size_n, trace
+    end
+
+    return all_partitions_of_size_n
+end
+
+"""
+    get_trace(trace::Dict, start::Partition)
+
+Print the trace of the partition `start` constructed with `construct_category`.
```
fieker, line 522: trace is any dictionary? Usually I'd expect the function to be called just `trace` as the "get" is automatically implied. But I don't understand the 'Dict'

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,190 @@
...
+
+function size(p::SpatialPartition)
+    size(p.partition)
+end
+
+function is_composable(p::SpatialPartition, q::SpatialPartition)
+    p.dimension == q.dimension && is_composable(p.partition, q.partition)
+end
+
+function is_worth_composition(p::AbstractPartition, q::AbstractPartition, max_length::Int)
```
fieker: definition?

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,190 @@
...
+    for (i, n) in enumerate(p_vector)
+        condition = upper_uneven ? i % 2 == 1 : (i <= upper ? i % 2 == 1 : i % 2 == 0)
+        if condition
+            block_to_size[n] = get(block_to_size, n, -1) - 1
+        else
+            block_to_size[n] = get(block_to_size, n, -1) + 1
+        end
+    end
+
+    return all(i -> i == 0, values(block_to_size))
```
fieker: 'all(iszero, values(block_to_size))`

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+            p_return[2][i] = new_id
+            new_id += 1
+        else
+            p_return[2][i] = get(new_ids, p_return[2][i], -1)
+        end
+    end
+    p_return
+end
+
+function add_partition_to_dict(dict::Dict{Int, Set{AbstractPartition}}, p::AbstractPartition)
```
fieker: why the `to_dict` part? Isn't that obvious from the arguments?
Maybe the 'Dict{Int, Set...}` should actually be a data type linked to the partitions? I'd assume if I cann this with a random Dict of this type, chaos will reign

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    add_apbs = get(dict, size(p), -1)
+    push!(add_apbs, p)
+    dict[size(p)] = add_apbs
+
+    return dict
+end
+
+function add_partition_to_composition_dict(vector::Vector, p::AbstractPartition)
```
fieker: at leat in comment: what vector is this?

experimental/SetPartitions/src/SetPartitions.jl
```diff
@@ -0,0 +1,65 @@
...
+export vertical_reflection
+export is_balanced
+export is_pair
+export is_noncrossing
+export construct_category
+export get_trace
+export size
+export *
+export ⊗
+export ⋅
```
fieker: what is this for?

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,141 @@
...
+
+"""
+    rotation(p::ColoredPartition)
+
+Return the rotation of `p` in the direction given by `lr` and `tb`.
+
+# Arguments
+- `p`: input partition
+- `lr`: whether left (true) or right (false)
+- `tb`: whether top (true) or bottom (false)
```
fingolfin: Reading this I still have no idea what `lr` and `tb` do :-(. In particular, "whether left or right" is not a complete sentence, and it doesn't say how the rotation is defined

experimental/SetPartitions/src/ColoredPartition.jl
```diff
@@ -0,0 +1,141 @@
...
+"""
+function rotation(p::ColoredPartition, lr::Bool, tb::Bool)
+
+    if tb && isempty(upper_points(p))
+        error("Partition has no top part.")
```
fingolfin: No period at end of error messages like that

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,498 @@
+function do_unary(
```
fingolfin: This is a reaaally long argument list and complicated function -- it should have at least *some* comments. At the very, very least a line saying what is is about. Great would be a pointer to a textbook or paper that describes the algorithms. More of course would be better. Otherwise this becomes and unmaintainable blackbox down the road.

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,498 @@
...
+                    push!(to_unary, a)
+
+                    # call functions which adds partition a into the right set in the dict
+                    all_partitions_by_size = 
+                        add_partition_to_dict(all_partitions_by_size, a)
+                    all_partitions_by_size_top_bottom = add_partition_to_composition_dict(
+                            all_partitions_by_size_top_bottom, a)
+                end
+            elseif spatial_rotation !== nothing
+                a = spatial_rotation(pmod)
```
fingolfin, line 64: This part of the code is not covered by the tests; with such a complex function that makes me wary.

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,498 @@
...
+            pop!(to_unary, pp)
+        end
+    end
+
+    return (stop_whole, all_partitions, already_u, all_partitions_by_size,
+        all_partitions_by_size_top_bottom, trace)
+
+end
+
+function do_tensor_products(
```
fingolfin: Same comments as for `do_unary` apply here, too.

experimental/SetPartitions/src/GenerateCategory.jl
```diff
@@ -0,0 +1,498 @@
...
+                    push!(get(without, i, -1), ii)
+                end
+            end
+        end
+    end
+    return (all_partitions, already_t, stop_whole, all_partitions_by_size, 
+        all_partitions_by_size_top_bottom, trace)
+end
+
+function do_composition(
```
fingolfin: Same comments as for `do_unary` apply here, too.

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,190 @@
...
+    # Dictionary from block to size of block
+    block_to_size = Dict{Int, Int}()
+
+    # Initialize dictionary
+    for i in p_vector
+        if !(i in keys(block_to_size))
+            block_to_size[i] = 1
+        else
+            block_to_size[i] = get(block_to_size, i, -1) + 1
+        end
```
fingolfin:
```suggestion
        block_to_size[i] = get(block_to_size, i, 0) + 1
```

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,190 @@
...
+
+    for (i, n) in enumerate(p_vector)
+        if (n in incompleted && 
+            (isempty(last_incompleted) ? -1 : last_incompleted[end]) != n)
+            return false
+        end
+        if !(n in already_seen)
+            push!(already_seen, n)
+            block_to_size[n] = get(block_to_size, n, -1) - 1
+            if (get(block_to_size, n, -1) > 0 &&
```
fingolfin: You literally just stored `block_to_size[n]`, why retrieve it again, and with `get`?
```suggestion
            block_to_size[n] = get(block_to_size, n, -1) - 1
            if (block_to_size[n] > 0 && 
```
or even better, avoid another hash table lookup:
```suggestion
            bs = block_to_size[n] = get(block_to_size, n, -1) - 1
            if (bs > 0 && 
```

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,190 @@
...
+            block_to_size[n] = get(block_to_size, n, -1) - 1
+            if (get(block_to_size, n, -1) > 0 && 
+                (isempty(last_incompleted) ? -1 : last_incompleted[end]) != n)
+                push!(last_incompleted, n)
+                push!(incompleted, n)
+            end
+        else
+            if p_vector[i-1] != n && get(block_to_size, p_vector[i-1], -1) != 0
+                return false
+            else
```
fingolfin: Since you `return` in the line above, you can just `end` here, and then de-indent the following block. Advantage: the code is nested less, which makes it either to read and understand.
```suggestion
            end
```

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,266 @@
+"""
+    SetPartition
+
+Initialize SetPartition object, while transforming object to "normal form".
```
fingolfin: This "explanation" does not explain the most crucial thing: what does `SetPartition` represent?

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,266 @@
...
+end
+
+function copy(p::SetPartition)
+    return SetPartition(copy(p.upper_points), copy(p.lower_points))
+end
+
+"""
+    tensor_product(p::SetPartition, q::SetPartition)
+
+Return tensor product of `p` and `q`.
```
fingolfin: Again, what is the definition of the tensor product? Should either be part of the docstring, or the docstring should state where the definition can be found

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,266 @@
...
+function tensor_product(p::SetPartition, q::SetPartition)
+    
+    q_new = helper_new_point_values_vector([p.upper_points, p.lower_points], [q.upper_points, q.lower_points])
+    SetPartition(vcat(p.upper_points, q_new[1]), vcat(p.lower_points, q_new[2]))
+end
+
+"""
+    involution(p::SetPartition)
+
+Return involution of `p`.
```
fingolfin: What is "the involution" of a set partition? Again, definition should be included, or a pointer to it.

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,266 @@
...
+            end
+        end
+    end
+    
+    # final path compression
+    for ii in keys(new_ids)
+        path = [ii]
+        already_in = Set()
+        push!(already_in, get(new_ids, ii, -1))
+        z = get(new_ids, ii, -1)
```
fingolfin: You get `z` here -- and then in the next line you overwrite it?!

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,266 @@
...
+        while z in keys(new_ids)
+            push!(path, z)
+            push!(already_in, z)
+            z = get(new_ids, z, -1)
+            if z in already_in
+                break
+            end
+        end
```
fingolfin: Accessing entries of a dictionary is comparatively expensive. Don't do it again and again when you can get the data once and then keep it around. 

Also, if you create a `Set` via `Set()` this gives a `Set{Any}` which is not type stable, as it has to deal with *arbitrary* content. But here yo already know the type of the content. So you can write `Set{SomeType}()`. Even better: just create the set directly with the initial value, then Julia can infer the correct type.

So, here i a much more idiomatic way to write this code would be like this 
```suggestion
    for (ii, z) in new_ids
        path = [ii]
        already_in = Set((z,))
        while z in keys(new_ids)
            push!(path, z)
            push!(already_in, z)
            z = get(new_ids, z, -1)
            z in already_in && break
        end
```

experimental/SetPartitions/src/SpatialPartition.jl
```diff
@@ -0,0 +1,70 @@
+"""
+    SpatialPartition
+
+Initialize SpatialPartition object
```
fingolfin: What is that?

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
+"""
+    helper_new_point_values_vector(p::Vector{Vector{Int}}, q::Vector{Vector{Int}})
+
+Return a semantically identical partition in form of an vector, 
+which has new number values.
+"""
+function helper_new_point_values_vector(p::Vector, q::Vector)
```
fingolfin: Function decl does not match the docstring, why?
```suggestion
function helper_new_point_values_vector (p::Vector{Vector{Int}}, q::Vector{Vector{Int}})
```

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    p_points = vcat(copy(p[1]), copy(p[2]))
```
fingolfin: Redundant copy, since you `vcat` anyway?
```suggestion
    p_points = vcat(p[1], p[2])
```

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    if !(isempty(p_points))
+        new_id, index = findmax(p_points)
+        new_id += 1
+    else
+        new_id = 1 
+    end
+    
+    q_points = vcat(copy(q[1]), copy(q[2]))
+    new_ids = Dict()
```
fingolfin: Not type stable. Perhaps like this (I am guessing the types
```suggestion
    new_ids = Dict{Int,Int}()
```

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+
+    for n in q_points
+        new_ids[n] = new_id
+        new_id += 1
+    end
+    upper = []
+    for (i, n) in enumerate(q[1])
+        push!(upper, get(new_ids, n, -1))
+    end
```
fingolfin: You don't need `i`. And `upper` is a `Vector{Any}`. How about
```suggestion
    upper = [get(new_ids, n, -1) for n in q[1]]
```
or perhaps even this (I did not try to reason whether `new_ids[n]` will be defined in all cases):
```suggestion
    upper = [new_ids[n] for n in q[1]]
```

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+    
+    lower = []
+    for (i, n) in enumerate(q[2])
+        push!(lower, get(new_ids, n, -1))
```
fingolfin: same comments are for `upper` apply

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+    end
+    
+    [upper, lower]
+
```
fingolfin:
```suggestion
    return [upper, lower]
```

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+    new_ids = Dict{Int, Int}()
+    p_return = helper_new_point_values_vector([[length(p[1]) + length(p[2])], []], p)
+
+    for (i, n) in enumerate(p_return[1])
+        if !(n in keys(new_ids))
+            new_ids[n] = new_id
+            p_return[1][i] = new_id
+            new_id += 1
+        else
+            p_return[1][i] = get(new_ids, p_return[1][i], -1)
```
fingolfin: Huh, isn't this the sameas
```suggestion
            p_return[1][i] = get(new_ids, n, -1)
```

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,95 @@
...
+        end
+    end
+
+    for (i, n) in enumerate(p_return[2])
+        if !(n in keys(new_ids))
+            new_ids[n] = new_id
+            p_return[2][i] = new_id
+            new_id += 1
+        else
+            p_return[2][i] = get(new_ids, p_return[2][i], -1)
```
fingolfin: same as above

experimental/SetPartitions/test/runtests.jl
```diff
@@ -0,0 +1,6 @@
+using Test
+using Oscar
+
+include("Partition-test.jl")
```
fingolfin: I think this `runtests.jl` file can be deleted, as we automatically include all *.jl files inside `test` subdirectories?

fingolfin (commented):
This code is for experimental, and as such it doesn't have to be "perfect".

That said, at the very least the objects and operations provided by this code should be clear, which to me they currently aren't at all. It is also not clear to me what the intended applications of some of these things are -- who wants to use these "spatial partitions" (whatever they are) for what?

Beyond this, there is a lot more in here that should be improved before it could ever have a chance of getting out of experimental. But it could be done after a merge, if you promise to do so.


I did this review because I was asked to, but I'd really appreciate e if others also did this kind of review which just looks for general issues with the code. I am not particularly interested in set partitions myself.

experimental/SetPartitions/src/SetPartition.jl
pinguly: Should we add a high-level explanation with a reference to a paper or do you prefer a more precise mathematical definition in the docstrings?

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,193 @@
+function size(p::SetPartition)
+    length(p.lower_points) + length(p.upper_points)
+end
+
+function is_composable(p::SetPartition, q::SetPartition)
+    length(p.upper_points) == length(q.lower_points)
```
lgoettgens:
```suggestion
    return length(p.upper_points) == length(q.lower_points)
```
All functions with the keyword `function` in front of should use `return`.

lgoettgens:
Could you click on the little `resolve` button on all the comments that you already took care of? That makes it a bit easier to see what is still in the works.

docs/oscar_references.bib
```diff
@@ -270,6 +270,17 @@ @Article{BN07
   doi           = {10.1016/j.aim.2007.04.012}
 }
 
+@Article{BS09,
+  author        = {Banica, Teodor and Speicher, Roland},
+  title         = {Liberation of orthogonal {L}ie groups},
```
lgoettgens:
```suggestion
  title         = {Liberation of orthogonal Lie groups},
```
Our Bibtex parser doesn't like these

docs/oscar_references.bib
```diff
@@ -915,6 +935,15 @@ @Article{Gat96
   doi           = {10.1007/BF01191379}
 }
 
+@PhDThesis{Gro20,
+  author        = {Gromada, Daniel},
+  title         = {Compact matrix quantum groups and their representation categories},
+  year          = {2020},
+  url           = {https://publikationen.sulb.uni-saarland.de/bitstream/20.500.11880/29789/1/dis.pdf},
```
lgoettgens: please provide a DOI instead of a url. `10.22028/D291-32389` should work in this case

experimental/SetPartitions/README.md
```diff
@@ -0,0 +1,20 @@
+# Set-Partitions
+
+## Aims
+
+The goal of this project is to
+bring set-partitions to OSCAR.
```
lgoettgens:
```suggestion
The goal of this project is to bring set-partitions to OSCAR.
```

experimental/SetPartitions/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+We implemented data structures for partitions and corresponding algorithms as described in [Vol23](@cite). 
+This includes:
+* basic set-partitions and operations on them (e.g. composition, tensor product, involution)
+* variations like colored partitions [TW18](@cite) and spatial partitions [CW16](@cite)
+* enumeration of partitions which can be constructed from a set of generators
+
+In the future, we plan to implement:
+* linear combinations of partitions
+* examples of tensor categories in the framework of [TensorCategories.jl](https://github.com/FabianMaeurer/TensorCategories.jl)
```
lgoettgens, line 19: Could you add some contact information about the people working and maintaining this? For an example. see experimental/FTheoryTools/README.md

experimental/SetPartitions/src/AbstractPartition.jl
```diff
@@ -0,0 +1,26 @@
...
+    return composition_loops(p, q)[1]
+end
+
+function ⋅(p::T, q::T) where {T <: AbstractPartition}
+    return composition(p, q)
+end
```
lgoettgens: This symbol is used nowhere else in Oscar. And I think having * should be enough, right?

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
...
+
+function hash(p::SetPartition, h::UInt)
+    return hash(p.upper_points, hash(p.lower_points, h))
+end
+
+function ==(p::SetPartition, q::SetPartition)
+    return p.lower_points == q.lower_points && p.upper_points == q.upper_points
+end
+
+function copy(p::SetPartition)
```
lgoettgens: Instead of `copy`, you should implement `deepcopy_internal` (for all types)

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
+"""
+    SetPartition
+
+A set-partition is a partition of a set of upper and lower points into disjoint subsets. 
+Such partitions are often depicted as string diagrams where points in the same subset 
+are connected by lines. See also Section 4.1.1 in [Gro20](@cite).
+
+`SetPartition` stores two vectors `upper_points` and `lower_points`
+where same numbers correspond to points in the same subset. Both vectors are 
+automatically converted into normal form as returned by [`normal_form`](@ref).
```
lgoettgens: This is about internals of the struct. Users shouldn't touch/see/modify them, so this information is useless to misleading in a docstring. If you want to give users a way to see/modify this stuff, you should write appropriate functions.

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
...
+
+Return the rotation of `p` in the direction given by `lr` and `tb`. 
+
+Rotating a partition moves the left- or right-most point of the upper points 
+to the lower points or vice verca. See also Section 4.1.2 in [Gro20](@cite).
+
+# Arguments
+- `p`: input partition
+- `lr`: rotating at the left (true) or at the right (false)
+- `tb`: rotating from top to bottom (true) or from bottom to top (false)
```
lgoettgens: These are not present in the signature in Line 100, and are still not explained properly. Having two booleans as positional arguments seems the seed of future confusion for me, as one needs to remember their order. Maybe it would be better to have them as keyword arguments instead.

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
...
+"""
+function rotation(p::SetPartition, lr::Bool, tb::Bool)
+    
+    if tb && isempty(p.upper_points)
+        error("SetPartition has no top part")
+    elseif !tb && isempty(p.lower_points)
+        error("SetPartition has no bottom part")
+    end
+
+    ret = (copy(p.upper_points), copy(p.lower_points))
```
lgoettgens: you should use `deepcopy` whenever you want to create a proper copy.

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
...
+    
+    return SetPartition(ret[1], ret[2])
+end
+
+"""
+    composition_loops(p::SetPartition, q::SetPartition)
+
+Return the composition of `p` and `q` as well as the number of removed loops.
+
+The composition of two partitions is obtained concatenating them horizontally
```
lgoettgens:
```suggestion
The composition of two partitions is obtained by concatenating them horizontally
```

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
...
+julia> composition_loops(SetPartition([1, 1], [2]), SetPartition([1], [2, 2]))
+(SetPartition([1], [2]), 1)
+
+julia> composition_loops(SetPartition([1], [1, 2]), SetPartition([1], [2, 2]))
+ERROR: format not fitting
+```
+"""
+function composition_loops(p::SetPartition, q::SetPartition)
+    
+    !is_composable(p, q) ? error("format not fitting") :
```
lgoettgens:
```suggestion
    @req is_composable(p, q) "Number of points mismatch" 
```
or similar. Please use `@req` instead of `error(...)` for input sanitation

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,293 @@
...
+Such partitions are often depicted as string diagrams where points in the same subset 
+are connected by lines. See also Section 4.1.1 in [Gro20](@cite).
+
+`SetPartition` stores two vectors `upper_points` and `lower_points`
+where same numbers correspond to points in the same subset. Both vectors are 
+automatically converted into normal form as returned by [`normal_form`](@ref).
+
+# Examples
+```jldoctest
+julia> SetPartition([2, 4], [4, 99])
```
lgoettgens: All constructible types should have a snake_case function to construct them. See the first bullet point in https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/#Naming-conventions

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,108 @@
...
+    return (upper, lower)
+end
+
+"""
+    normal_form(p_upper::Vector{Int}, p_lower::Vector{Int})
+
+Return two vectors which are semantically identical to `p_upper` and `p_lower`
+as set-partition and are numbered from 1 to the number of subsets.
+"""
+function normal_form(p_upper::Vector{Int}, p_lower::Vector{Int})
```
lgoettgens: Is this a function that a user needs to call themselves? If not, I would call it `_normal_form` instead to make it clear that it is internal. 
I have seen calls to it in basically all constructors and operations, thus the question above.

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,194 @@
+function size(p::SetPartition)
+    return length(p.lower_points) + length(p.upper_points)
+end
```
lgoettgens: I find all of these concrete implementations weird in this file and would personally look for them in the corresponding PartitionType.jl file instead.

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,194 @@
...
+    return size(p.partition)
+end
+
+function is_composable(p::ColoredPartition, q::ColoredPartition)
+    return p.color_upper_points == q.color_lower_points && 
+        is_composable(p.partition, q.partition)
+end
+
+function upper_points(p::ColoredPartition)
+    return p.partition.upper_points
```
lgoettgens:
```suggestion
    return upper_points(p.partition)
```
avoid direct access as much as possible

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,194 @@
+function size(p::SetPartition)
+    return length(p.lower_points) + length(p.upper_points)
+end
+
+function is_composable(p::SetPartition, q::SetPartition)
+    return length(p.upper_points) == length(q.lower_points)
+end
+
+function upper_points(p::SetPartition)
```
lgoettgens: for my use-case, I would like to have accessors for the number of upper/lower points, e.g.
```julia
function num_upper_points(p::AbstractPartition)
	return length(upper_points(p))
end
```

experimental/SetPartitions/src/PartitionProperties.jl
```diff
@@ -0,0 +1,194 @@
...
+    return p.dimension == q.dimension && is_composable(p.partition, q.partition)
+end
+
+"""
+    is_worth_composition(p::AbstractPartition, q::AbstractPartition, max_length::Int)
+
+Return whether it is worth to compose `p` and `q` such that the result is less
+equal `max_length` and not equal to the empty partition.
+
+Note that this is a helper function for the `construct_category` algorithm.
```
lgoettgens: If it's purely an internal helper, consider prepending the name by a `_`.

lgoettgens (commented):
Some more comments from a quick look over some parts. I will have a look later, once some things are adapted.
As Max, I don't know a lot about the topic, so most comments are just from a code style perspective.

experimental/SetPartitions/src/PartitionProperties.jl
lgoettgens: not just here, but basically everywhere. In the optimal case, you only have direct access in one function per field and everything else refers to that.

experimental/SetPartitions/src/SetPartition.jl
pinguly: Would it be better to split it into four functions like `rotate_top_left`, `rotate_top_right`, ... and then document it separately?

experimental/SetPartitions/src/SetPartition.jl
sebvz777: Hi, thanks for your feedback! This @req macro does not seem to be part of the julia standard library. Do you mean by using @req that I should implement my own req macro or should I use the @req from DiscreteValueIteration (https://github.com/JuliaPOMDP/DiscreteValueIteration.jl) or am I missing sth?

lgoettgens: This is part of Hecke and thus available in Oscar without any extra work. See https://docs.oscar-system.org/dev/Hecke/features/macros/#Miscellaneous for the docs.

lgoettgens: But I agree that this should be added to the Oscar developer docs

sebvz777: ah I see, makes sense, thanks!

experimental/SetPartitions/src/SetPartition.jl
```diff
@@ -0,0 +1,305 @@
...
+    return hash(p.upper_points, hash(p.lower_points, h))
+end
+
+function ==(p::SetPartition, q::SetPartition)
+    return p.lower_points == q.lower_points && p.upper_points == q.upper_points
+end
+
+function deepcopy(p::SetPartition)
+    return SetPartition(deepcopy(p.upper_points), deepcopy(p.lower_points))
+end
```
lgoettgens: Please not that `deepcopy` is indeed the method to call for a copy. When providing a method however, you need to define `deepcopy_internal` (see https://docs.julialang.org/en/v1/base/base/#Base.deepcopy).

experimental/SetPartitions/src/Util.jl
```diff
@@ -0,0 +1,160 @@
...
+    _req_colored_partition(partition::AbstractPartition, 
+                            color_upper_points::Vector{Int}, 
+                            color_lower_points::Vector{Int})
+
+Return true if the format of the attributes for initializing a
+colored partition is valid, false otherwise.
+
+Note that this is a helper function for the constructor of colored partitions.
+"""
+function _req_colored_partition(partition::AbstractPartition,
```
lgoettgens: I, personally, wouldn't have these as a separate function and just inline them into the constructor. This makes it furthermore possible to use multiple `@req` statements with different error messages tho help a user with wrong input.

pinguly:
Is there anything else we should do for this first pull request? We definitely improve the code further and add more tests and documentations in the future.

lgoettgens (approved):
I think we do not have anybody who can review this from a technical pov.
The things that were mentioned here have all been adapted. If the authors want to continue working on this, I see no reason why not to merge this into experimental right now.

lgoettgens (changes requested):
I have one minor change I would like to have in here before merge:
There is an overhaul of method naming currently in action for `numer_of_things` functions (see https://github.com/oscar-system/Oscar.jl/pull/3272). To not make the code more inconsistent again, I would ask you to rename `num_lower_points` to `number_of_lower_points` and similarly for upper.

lgoettgens (approved):
Thanks!

I think we do not have anybody who can review this from a technical pov.
The things that were mentioned here have all been adapted. If the authors want to continue working on this, I see no reason why not to merge this into experimental right now.

thofma:
Will merge this once we got https://github.com/oscar-system/Oscar.jl/pull/3288/ merged.

lgoettgens:
> Will merge this once we got #3288 merged.

Bump

PR #2978, opened 2023-11-02
fixes regression loading vectors parametrized types

antonydellavecchia:
Doesn't force containers to be of a potentially more general type when non empty.


fingolfin:
Do you have a simple test case we could add that reproduce the issue / verifies that this fixes it?

fingolfin (approved):
Great way to "add" a test case, I love it :joy:

PR #2980, opened 2023-11-02
Deprecate a bunch of aliases

fingolfin:
... as a first step to weening us off them...

Note that `Base.@deprecate_binding` is undocumented but has been around for a long time and is unlikely to vanish soon. Besides, we only intend to use it temporarily anyway, as an intermediate step before removing these aliases completely; so there is no need to worry about long-term availability.

This and further upcoming changes that deprecate aliases may break downstream code, in particular in the OSCAR book -- this is a feature; we don't want the book to reference `fmpz` and `fmpq` etc.

In OSCAR 0.15 (or 1.0) we should then remove these aliases completely. 

See also https://github.com/Nemocas/AbstractAlgebra.jl/pull/1469 and https://github.com/Nemocas/Nemo.jl/pull/1557


lgoettgens:
I would prefer this PR to not break the deprecation CI, e.g. OSCAR itself shouldn't use these aliases. CI currently reports uses of `issimplicial`.

fingolfin:
fixed the `issimplicial`, and also noticed a use of `issmooth` next to those.

I'll add the `depwarn=error` to the list of required tests -- I never had the intention of breaking here, to the contrary, I rely on our CI to point out things I missed :-)

lgoettgens:
Next deprectation errors are `isbounded` and `representative_action`

PR #2981, opened 2023-11-03
Replace type declarations

fingolfin:
Type declarations insert `convert` calls. Instead, use type assertions
(which just verify the object has the right type to start with), or
fix the code to actually be type stable and return the right type.

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/methods.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/methods.jl
@@ -122,5 +122,5 @@ function integrate(c::CohomologyClass)
     end
     n = AbstractAlgebra.leading_coefficient(top_form.f)
     m = AbstractAlgebra.leading_coefficient(polynomial(volume_form(toric_variety(c))).f)
-    return QQFieldElem(ZZ(n),ZZ(m))
+    return n//m
 end
--- a/src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/methods.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/methods.jl
@@ -122,5 +122,5 @@ function integrate(c::CohomologyClass)
     end
     n = AbstractAlgebra.leading_coefficient(top_form.f)
     m = AbstractAlgebra.leading_coefficient(polynomial(volume_form(toric_variety(c))).f)
-    return QQFieldElem(n,m)
+    return QQFieldElem(ZZ(n),ZZ(m))
 end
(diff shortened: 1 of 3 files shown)


src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/methods.jl
```diff
@@ -97,30 +97,30 @@ function integrate(c::CohomologyClass)::QQFieldElem
...
     end
     top_form = dict[elem]
     if iszero(top_form)
-        return 0
+        return zero(QQ)
     end
     n = AbstractAlgebra.leading_coefficient(top_form.f)
     m = AbstractAlgebra.leading_coefficient(polynomial(volume_form(toric_variety(c))).f)
-    return QQFieldElem(n//m)
+    return QQFieldElem(n,m)
```
HereAround: @fingolfin Unless I am mistaken, this change leads to the CI-failures in 1.6 ubuntu-latest: `no method matching QQFieldElem(::QQFieldElem, ::QQFieldElem)`.

Generally speaking, your changes look good to me. In particular, I would hope that once the above line is fixed, the tests pass again.

HereAround: Maybe `QQFieldElem(ZZ(n), ZZ(m))` might work.

HereAround: I just tested this fix, and it seems to work. So I have taken the liberty to push a corresponding fix.

(Origin: `m`, `n` are coefficients of polynomials with rational coefficients. So by construction, they are rational numbers.)

HereAround (approved):
Modulo the minor change above, this PR looks good to me. Thank you @fingolfin . (But as always, let us see if the tests pass...)

thofma:
Tests are not passing with a genuine error

src/AlgebraicGeometry/ToricVarieties/CohomologyClasses/methods.jl
```diff
@@ -97,30 +97,30 @@ function integrate(c::CohomologyClass)::QQFieldElem
...
     end
     top_form = dict[elem]
     if iszero(top_form)
-        return 0
+        return zero(QQ)
     end
     n = AbstractAlgebra.leading_coefficient(top_form.f)
     m = AbstractAlgebra.leading_coefficient(polynomial(volume_form(toric_variety(c))).f)
-    return QQFieldElem(n//m)
+    return QQFieldElem(ZZ(n),ZZ(m))
```
thofma:
```suggestion
    return n//m
```

thofma: @HereAround I guess `top_form.f` and the other one are polynomials over QQ?

PR #2983, opened 2023-11-03
Serialization of quotient rings

HechtiDerLachs:
CC: @antonydellavecchia


HechtiDerLachs:
We should extend the serialization of monomial orderings a bit one day. But for a start this works and should be good to go, right?

antonydellavecchia (approved):
Still a lot to do, more orderings and so on. For a first attempt this is suffcient.

PR #2987, opened 2023-11-03
Try different implementation for min/max

thofma:
@YueRen This is the alternative implementation of the min/max convention I was thinking about. If I were to implement the tropical semiring again, I would do it this way. (I really dislike abusing the `min` and `max` function.) I am just opening it to hear your opinion.

(There is one failing test, which I would have to look into, if we want to go down this road.)

Maybe @lkastner also has some input?


YueRen:
Thanks for the suggestion!  What is so bad about abusing the `min` / `max` function?

I don't mind switching to `Min` / `Max`, but it is less user-friendly in my opinion as it requires users to distinguish between `min` / `max` and `Min` / `Max`, so I think it requires a good reason.

thofma:
I would say it is "cleaner", since using the function as a type parameter is very cumbersome (see the millions of `typeof(min)` everywhere) and I find the following irritating:
```
julia> convention(P)
max (generic function with 12 methods)
```
Why is the convention a function? It also allows us to add a pretty docstring for `Max` and `Min`:

```julia
help?> Min
search: Min min minor minmax minors minpoly minimum minimum! min_weights minkowski_sum minkowski_map minimal_model minimal_faces

  Min

  Indicating that we are in the tropical semiring (ℝ ∪ {∞}, ⊕, ⊙) with the min convention.

help?> Max
search: Max max maxes maximum maximum! max_order max_fields maxintfloat MaximalOrder maximal_order maximal_cones maximal_cells

  Max

  Indicating that we are in the tropical semiring (ℝ ∪ {-∞}, ⊕, ⊙) with the max convention.
```
We could not do that for `max` or `min`.

My last argument is that Sascha did it here: https://github.com/saschatimme/TropicalSemiring.jl and he knows what he is doing (as long it agrees with what I am trying to do ;)).

thofma:
I will close this for now. I still believe using `min/max` is a suboptimal way and might yield increased maintenance or inconveniences in the future.

YueRen:
@thofma Feel free to raise this topic when it becomes an issue in the future!

I agree 100% that `typeof(min)/typeof(max)/Union{typeof(min),typeof(max)}` is cumbersome to use, but we can work around it by introducing shorter and cleaner aliases for them.

And regarding pretty prints and docstrings:  Unlike Sascha's code, `min/max` do not play a very visible role in OSCAR's tropical package.  Instead, we have `TropicalSemiringMap` that encodes both the convention as well as the (by default trivial) valuation on the coefficient ring.

PR #2988, opened 2023-11-03
Plane curves

simonbrandhorst:
This is a first attempt at bringing the experimental PlaneCurves module of Delphine Pol to src and make its terminology and interface consistent with the rest of the schemes.


simonbrandhorst:
@HechtiDerLachs thanks for your support!

experimental/Schemes/CoveredScheme.jl
```diff
@@ -101,6 +101,24 @@ affine_refinements(C::Covering) = C.affine_refinements
 ########################################################################
 
 @attr function standard_covering(X::AbsProjectiveScheme{<:Ring, <:MPolyQuoRing})
+  return _standard_covering(X, identity)
+end
+
+
+@attr function standard_covering(X::AbsProjectiveAlgebraicSet)
```
fingolfin: Is there *any* type hint we could give for the return type? E.g.
```suggestion
@attr Covering function standard_covering(X::AbsProjectiveAlgebraicSet)
```
Same for the other uses of `@attr`

src/AlgebraicGeometry/RationalPoint/ProjectiveRationalPoint.jl
```diff
@@ -123,7 +139,12 @@ end
 
 function is_smooth_at(X::AbsProjectiveScheme{<:Field}, P::AbsProjectiveRationalPoint)
   @req P in X "not a point on X"
-  error("not implemented")
+  X = codomain(P)
+  S = standard_covering(X)
+  n = length(coordinates(P))
+  i = findfirst(i->!iszero(P[i]),1:n)
```
fingolfin: Could perhaps also do this? (Feel free to ignore)
```suggestion
  i = findfirst(!iszero,coordinates(P))
```

experimental/Schemes/CoveredScheme.jl
HechtiDerLachs: @simonbrandhorst and I talked about this beforehand and I will overhaul this part in general.

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
```diff
@@ -0,0 +1,333 @@
+export multiplicity
+export tangent
+export tangent_lines
+export curve_intersect
+export geometric_genus
+export intersection_multiplicity
+export are_transverse
```
afkafkafk13: From the general naming scheme, I would start looking for this with tab-completion by entering `is_tran` expecting something like `is_transversal_intersection`. Maybe we could alias a variant which starts with `is_`.

src/AlgebraicGeometry/RationalPoint/ProjectiveRationalPoint.jl
simonbrandhorst: thanks for the hint

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
```diff
@@ -0,0 +1,333 @@
...
+
+function irreducible_components(C::AffinePlaneCurve)
+  return [AffinePlaneCurve(i[1]) for i in factor(defining_equation(C))]
+end
+
+################################################################################
+# Helping function:
+# change of variable to send a point at the origin.
+
+function curve_map_point_origin(C::AffinePlaneCurve, P::AbsAffineRationalPoint)
```
afkafkafk13, line 133: Do you also envision to allow points to be specified in a syntax like [0,1] or by means of a maximal ideal (similar to the input to the localization framework)?

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
```diff
@@ -0,0 +1,333 @@
...
+julia> intersection_multiplicity(C, D, Q)
+1
+
+```
+"""
+function intersection_multiplicity(C::AffinePlaneCurve, D::AffinePlaneCurve, P::AbsAffineRationalPoint)
+  P in ambient_space(C) || error("The point needs to be in the same affine plane as C")
+  P in C && P in D || return 0
+  S,_ = stalk(intersect(C,D), P)
+  return vector_space_dimension(S)
```
afkafkafk13, line 267: You need to check that the dimension is zero. The (possibly reducible) curves might have a common component passing through P. A return value of  `-1` for that case is not sufficiently clear for the user. An error message would be better.

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
```diff
@@ -0,0 +1,333 @@
...
+
+julia> are_transverse(C, D, Q)
+true
+
+```
+"""
+function are_transverse(C::AffinePlaneCurve, D::AffinePlaneCurve, P::AbsAffineRationalPoint)
+  P in C && P in D || return false
+  return is_smooth_at(C, P) && is_smooth_at(D, P) && intersection_multiplicity(C, D, P) == 1
+end
```
afkafkafk13: It suffices to compute the intersection multiplicity here. If one of the curves is not smooth at P, then it is of multiplicity at least 2 itself at P and the intersecton multiplicity clearly exceeds 1.

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
```diff
@@ -0,0 +1,333 @@
...
+
+
+function projective_closure(C::AffinePlaneCurve)
+   F = defining_equation(C)
+   K = base_ring(C)
+   R, (x, y, z) = polynomial_ring(K, ["x", "y", "z"])
+   T, _ = grade(R)
```
afkafkafk13: Please use graded_polynomial_ring for generating a graded polynomial ring.

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
```diff
@@ -0,0 +1,323 @@
...
+  X::ProjectiveAlgebraicSet{BaseRingType, RingType}
+  defining_equation::MPolyDecRingElem
+
+  function ProjectivePlaneCurve(X::ProjectiveAlgebraicSet{S,T}, check::Bool=true) where {S,T}
+    @check begin
+      dim(X) == 1 || error("not of dimension one")
+      dim(ambient_space(X)) == 2 || error("not a plane curve")
+    end
+    new{S,T}(X)
+  end
```
afkafkafk13, line 28: You do not want to set the defining_equation here?

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
```diff
@@ -0,0 +1,323 @@
...
+  a = evaluate(L[1], coordinates(P))
+  b = evaluate(L[2], coordinates(P))
+  c = evaluate(L[3], coordinates(P))
+  if iszero(a) && iszero(b) && iszero(c)
+    error("This is a singular point of the curve")
+  else
+    V = gens(ambient_coordinate_ring(C))
+    T = a*V[1] + b*V[2] + c*V[3]
+    return ProjectivePlaneCurve(T)
+  end
```
afkafkafk13: Just out of curiosity: Why don't you use jacobian_matrix here (staying closer to textbooks)?

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
```diff
@@ -0,0 +1,323 @@
...
+  Ca = S[i]
+  Q = Ca(P)
+  return multiplicity(Ca, Q)
+end
+
+
+################################################################################
+# homogeneization for lines
+
+function help_homogene_line(R::MPolyRing, r::MPolyRing, F::MPolyRingElem, i::Int)
```
afkafkafk13, line 143: This is not self-explanatory. Maybe add a line in the comment, what this should do. In particular, guessing  the roles of r and R requires a bit of imagination.

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
```diff
@@ -0,0 +1,323 @@
...
+
+function geometric_genus(C::ProjectivePlaneCurve)
+  # buggy
+  A = homogeneous_coordinate_ring(C)
+  L = normalization(A)
+  m = length(L)
+  pa = zero(ZZ)
+  for i in 1:m
+    J = L[i][1].I
+    T, _ = grade(parent(J[1]))
```
afkafkafk13, line 271: Shouldn't `normalization` take care of grading in a reasonable way, if the input is graded?

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
```diff
@@ -0,0 +1,323 @@
...
+    V = T.(gens(J))
+    JJ = ideal(T, V)
+    B = quo(T, JJ)
+    H = hilbert_polynomial(B[1])
+    pa = pa - ZZ(coeff(H, 0))
```
afkafkafk13, line 276: Do we have something like `constant_coefficient` as a getter here to make this more readable?

src/Rings/MPolyMap/flattenings.jl
```diff
@@ -510,6 +510,23 @@ function ideal_membership(
...
+          }
+  parent(x) === base_ring(I) || return base_ring(I)(x) in I
+  phi = flatten(parent(x))
+  return radical_membership(phi(x), phi(I))
+end
+
+# TODO: Move to appropriate place eventually.
+function radical_membership(x::MPolyQuoRingElem, I::MPolyQuoIdeal)
+  return lift(x) in saturated_ideal(I)
+end
```
afkafkafk13: Shouldn't this be `radical_membership(lift(x), saturated_ideal(I))`?

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
```diff
@@ -220,3 +241,9 @@ function decide_du_val_singularity(P::AbsAffineRationalPoint{<:FieldElem,<:Any})
   d = decide_du_val_singularity(codomain(P), ideal(P))
   return d[1][1],d[1][3]
 end
+
+function stalk(X::AbsSpec, P::AbsAffineRationalPoint)
```
afkafkafk13, line 244: The notion of stalk is defined for sheaves, not for spectra. You want the stalk of the structure sheaf here. Is this accepted abuse of notation?

afkafkafk13 (commented):
Overall it looks good.

I left you a few questions and comments.

HechtiDerLachs:
@afkafkafk13 : A lot of the code is neither due to @simonbrandhorst , nor me. It's Delphine's old stuff that does not adhere to our coding conventions in many ways. However, the purpose of this PR is to integrate that into the schemes make it run *somehow*. I think nobody has the appetite to do further cosmetics at this point. If it works, it works. 

For some other parts of the code you are right with your suggestions and I will look into them.

afkafkafk13:
@HechtiDerLachs I am aware of this. Feel free to ignore the remarks which you consider inappropriate for the present pass through the code. I only typed up everything which caught my eye.

HechtiDerLachs:
I did what I could to fix up the creation of covered schemes from projective ones. Currently the tests still fail locally on my machine, but that seems to be due to non-visible methods of plane curves being tested. @simonbrandhorst : Can you take it again from here?

HechtiDerLachs:
I can confirm that all tests on the schemes are running again, except the newly added ones for curves. Hence, I return this PR to the hands of @simonbrandhorst again.

simonbrandhorst:
Thank you! I will take it from here.

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
simonbrandhorst: No. But the syntax to create such a point is `C(I)` or `C([0,1])`

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
simonbrandhorst: Done

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
simonbrandhorst: done

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
simonbrandhorst: It will be computed in the method `defining_equation`. (I hope at least)

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
simonbrandhorst: it is an internal funciton. not worth the effort.

src/AlgebraicGeometry/Curves/ProjectivePlaneCurve.jl
simonbrandhorst: apparently normalization is not implemented for graded rings

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
simonbrandhorst: I would say so. Formally a scheme X is a ringed space (X, O_X) one just drops O_X from notation. 
And we do not have O_X .

experimental/Experimental.jl
```diff
@@ -10,6 +10,7 @@ const oldexppkgs = [
   "ModStd",
   "Rings",
   "Schemes",
+  "PlaneCurve",
```
simonbrandhorst: I will not be able to carry all the plane curve stuff over any time soon. 
This disables the experimental package. Otherwise I get a lot of warnings of identifier conflics. (see below). 
But disabling the package entirely seems drastic. @benlorenz do we have any other options? 
Keep it in a module and not export anything?
```
WARNING: using PlaneCurveModule.invert_birational_map in module Oscar conflicts with an existing identifier.
WARNING: using PlaneCurveModule.rat_normal_curve_It_Proj_Even in module Oscar conflicts with an existing identifier.
... and so on
```

src/AlgebraicGeometry/RationalPoint/AffineRationalPoint.jl
afkafkafk13: Indeed, I agree with this argument.

src/AlgebraicGeometry/Curves/AffinePlaneCurve.jl
afkafkafk13: looks good.

simonbrandhorst:
@jankoboehm  @wdecker  I added plane curves to the documentation and adapted some names to Oscar conventions. For instance `parametrization_plane_curve`  is now `parametrization(::PlaneProjectiveCurve)`.

@fieker  I put the experimental `PlaneCurve` code in a new module `PlaneCurve`,
so that it can still be accessed and to avoid name conflicts.

afkafkafk13:
Something seems to be going seriously wrong with your reenabling of the old plane curve package. 

Your new package looks good to me, but I won't finish my review, before this problem is solved (one way or another).

simonbrandhorst:
O.K. no clue how to include the plane curve package without breaking Oscar and the tests. 
So I disabled it for now. Hope the tests pass then.

experimental/Experimental.jl
benlorenz: Sorry about the delay, enabling the old code should work when removing/commenting the block at the bottom of the `PlaneCurve.jl` containing the `using .PlaneCurveModule` and the export statements. (It already lives in a submodule)

The difficult part would be getting the tests to run since we cannot do `using Oscar` and `using Oscar.PlaneCurveModule` because of the conflicting identifiers with the new code.

Maybe it would work to add a module around the tests:
```
module PlaneCurveTest

using ..Oscar
using ..Test
import Oscar.PlaneCurveModule
const PC = Oscar.PlaneCurveModule

...

end
```
And then one needs to change all calls that use functions from the module to `PC.functionname`.

docs/src/AlgebraicGeometry/Curves/ProjectivePlaneCurves.md
```diff
@@ -48,36 +69,24 @@ curves only once. Its individual steps are interesting in their own right:
...
 
 ```@docs
-rational_point_conic(D::ProjPlaneCurve{QQFieldElem})
+rational_point_conic(D::ProjectivePlaneCurve{QQField})
 ```
 ## Parametrizing Rational Plane Curves
 
 ```@docs
- parametrization_plane_curve(C::ProjPlaneCurve{QQFieldElem})
+parametrization_plane_curve(C::ProjectivePlaneCurve{QQField})
```
afkafkafk13: Isn't that called parametrization by now?

afkafkafk13 (approved):

afkafkafk13:
Did you also disable the tests of the old package (experimental/PlaneCurve/test/runtests.jl), when disabling the package?

simonbrandhorst:
If no one objects, I will merge this on Wednesday.

afkafkafk13:
No objections from my side, as soon as all required tests pass.

wdecker:
The  `Singular` counterpart of the `OSCAR` function `invert_birational_map` does not only work for curves. As we now have also higher-dimensional varieties , it might make sense to also allow more general type signatures.

afkafkafk13:
> The `Singular` counterpart of the `OSCAR` function `invert_birational_map` does not only work for curves. As we now have also higher-dimensional varieties , it might make sense to also allow more general type signatures.

Good point. Maybe we should keep this in mind as an issue, but not block this PR by it?

simonbrandhorst:
@wdecker Thanks for the remark. I will open an issue.

wdecker:
@simonbrandhorst Thx, I agree with not blocking the PR at this point.

PR #2989, opened 2023-11-04
Use find_morphism from Hecke

thofma:


fieker (approved):
Wel done, thanks

thofma:
@benlorenz just for your entertainment: 1.10 short takes longer than 1.10 long (but at least it is working again, yay)

benlorenz:
> @benlorenz just for your entertainment: 1.10 short takes longer than 1.10 long (but at least it is working again, yay)

It is more the group of long tests vs the group of short tests, but great that it is working again and `31min` for the long tests is better than I have ever seen (but there is usually quite some fluctuation in these numbers). Lets see how this develops further.

PR #2992, opened 2023-11-05
Switch to AA.allow_unicode etc.

fingolfin:
Waiting for new AA/Nemo/Hecke, see #3071


thofma:
Do we need https://github.com/Nemocas/AbstractAlgebra.jl/pull/1495?

fingolfin:
I didn't expect that we'd need it, but it sure looks that way :-/.

thofma:
Contained in #3071

PR #3002, opened 2023-11-08
Derived pushforward of coherent sheaves (i.e. cohomology of families of varieties and sheaves on them)

HechtiDerLachs:
A first glimpse of what double complexes, exterior powers, Koszul complexes, flattenings of graded modules, and `all_monomials` of graded free modules were made for. 

The implementation is a dummy algorithm which works rather slow at the moment. But it's a proof of concept to be improved, eventually.


test/AlgebraicGeometry/Schemes/DerivedPushforward.jl
```diff
@@ -0,0 +1,51 @@
...
+  # We define a family of elliptic curves over a 1-dimensional base
+  A, (t,) = polynomial_ring(QQ, [:t])
+  P = projective_space(A, [:x, :y, :z])
+  S = homogeneous_coordinate_ring(P)
+  n = ngens(S)
+  f1 = sum(x^n for x in gens(S))
+  f0 = prod(gens(S))
+  #f0 = S[1]^n
+  # For t=0 this degenerates to three lines and we expect jumps in cohomology there
+  f = (1-t)*f0 + t*f1
```
HechtiDerLachs, line 11: We compute the `H^1(\Omega^1_C)` for this family of elliptic curves varying with parameter `t`. For `t=0` this degenerates to three lines and we find jumps in the cohomology groups.

simonbrandhorst:
That looks quite impressive!

test/AlgebraicGeometry/Schemes/DerivedPushforward.jl
simonbrandhorst: What morphism are you pushing forward along? In the signature I just see a projective space and a module?
Or is it the structure morphism to Spec K?

HechtiDerLachs: Yes, it's the structure morphism. But that's a dummy argument really, because `K` is already present as the coefficient ring of the base ring of the module. So we might want to turn this into an internal method taking only one argument. 

Keep in mind that `K` need not be a field. That's the whole point of this more general machine. In this example, it's `QQ[:t]`.

HechtiDerLachs:
I managed to get the code running for a family of quartic surfaces (K3s), degenerating to the four coordinate hyperplanes. By running I mean: Finishing within 5-10 minutes on a local machine with < 16 GB RAM. Still, I would very much like to see how this compares to BGG. 

@jankoboehm : I added some routine to simplify complexes. This should be similar to your method for `simplify` for SubQuos. But I didn't find anything for complexes, yet. In particular, this should minimize a complex over a local ring. For the moment this is just functional, but we should probably look at this and discuss the final shape that this should take.

Even though this is just a prototype, I think this is good to go for experimental. Also this work contains various small fixes for issues I encountered during the work. So once I slim and fix up the tests a bit, I would be happy if this could be merged.

HechtiDerLachs:
Rebased on #3021. The original contributions of this PR are rather minor, see the few last commits. 

Once we have #3021 in, I will rebase this again and then we can hopefully discuss merging it.

experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
```diff
@@ -0,0 +1,579 @@
...
+      k1 = k - 1
+    else # then j < i
+      k0 = k + 1
+    end
+  end
+  return false, nothing
+end
+
+# This addresses the TODO in Heckes src/Sparse/Row.jl:290
+function _get_index_fast(a::SRow, i::Int)
```
HechtiDerLachs, line 564: @fieker @thofma : Should this find its way into Hecke? I get a significant speedup from that. See also the methods below and above whether you would like to have such things. The point was to avoid allocating zeroes for return values in cases where you can not iterate over the entries of an `SRow`.

thofma: I don't understand the comment about "allocating zeroes".  The original code does
```julia
function Base.getindex(A::SRow{T}, i::Int) where {T <: RingElem}
  i < 1 && error("Index must be positive")
  p = findfirst(isequal(i), A.pos)
  if p === nothing
    return zero(base_ring(A))
  else
    return A.values[p]
  end
end
```
Where does that allocate more?

And is your code faster than using `searchsorted` or some of those related functions?

HechtiDerLachs: I didn't try `searchsorted` or anything. My code is probably a naive implementation of that. 

My usecase for the `_has_index` was when collecting a single column of a sparse matrix `A` into one `SRow`. I had to go to every row `v` of `A` and ask for the `q`-th index of `v`. If it was zero, I didn't add it to my result. Thus `v[q]` was always allocating zeroes everytime, even though they were discarded the next moment. Now I can replace this by a call to `_has_index(A, i, q)` and directly proceed if the first part of the answer is `false`. 

Does that make more sense?

thofma: OK, makes sense. You can keep it here. We can move it to Hecke later with some proper name.

test/AlgebraicGeometry/Schemes/DerivedPushforward.jl
```diff
@@ -0,0 +1,236 @@
...
+
+  # Compute the higher direct images of \Omega^1
+  M1 = Oscar.relative_cotangent_module(IPX)
+  M1, _ = pushforward(inc_X, M1)
+  st = Oscar._derived_pushforward(M1);
+  stsp = simplify(st);
+
+  @test iszero(stsp[0])
+  @test rank(stsp[-1]) == 20
+  @test iszero(stsp[-2])
```
HechtiDerLachs, line 234: @Lax202: The benchmark is now ~20 seconds for these computations here. How about BGG in this example?

Interestingly my code runs equally fast in the relative setting, i.e. when the base ring is not a field, but e.g. `QQ[t]`. That gives hope for more.

test/AlgebraicGeometry/Schemes/DerivedPushforward.jl
```diff
@@ -0,0 +1,263 @@
...
+  C = simplify(Oscar._derived_pushforward(M1));
+
+  @test iszero(C[0])
+  @test rank(C[-1]) == 1
+  @test rank(C[-2]) == 2
+  @test iszero(C[-3])
+  @test iszero(C[-4])
+  @test iszero(C[-5])
+
+  @test iszero(map(C, 1, (-1,)))
```
HechtiDerLachs, line 271: This is an example of a search for so-called "mild log singularities". Assuming that in this example `\theta^1_X/S` is indeed just `T^0_X/S` (i.e. log derivations coincide with the classical derivations), we compute the derived pushforward of `W^1_X/S`. The result `C` is non-trivial only in degrees `-1` and `-2` (corresponding to cohomological degrees `1` and `2` with the sign occuring due to internal reasons) and that the map between them is the zero map independent of the base parameter `t`. Thus the relative cohomology groups are indeed free `QQ[:t]`-modules in this example. 

Runtime on my laptop: ~2min.

This is still much, especially given the speed of bgg and local cohomology witnessed in #3058 . However, that machinery does not yet work over coefficient rings like `QQ[:t]`, but only over fields and it also does not yet allow you to get hold of the actual representatives of the cohomology classes. But in the long run, I think, we should switch to some BGG backend nevertheless.

test/AlgebraicGeometry/Schemes/DerivedPushforward.jl
```diff
@@ -0,0 +1,272 @@
...
+  X, inc_X = sub(IP5, I)
+
+  # Testing the structure sheaf
+  M0, _ = pushforward(inc_X, graded_free_module(homogeneous_coordinate_ring(X), [0]))
+
+  C = simplify(Oscar._derived_pushforward(M0));
+
+  @test rank(C[0]) == 1 # Free of rank one
+  @test all(k->iszero(C[-k]), 1:5)
+
```
HechtiDerLachs, line 258: Here we now also test the structure sheaf.

(Comment made for visibility to certain externals)

PR #3003, opened 2023-11-09
Bump DocumenterCitations to 1.3

lgoettgens:
In particular, the workaround for corporate authors is no longer needed. Furthermore, `\url` in `note` fields is now supported.

The large part of the changes here is due to change of internals of `DocumenterCitations`. Once they have been declared public API (see https://github.com/JuliaDocs/DocumenterCitations.jl/pull/57#issue-1969586171), we can change the compat bound from `~1.3.1` to `1.3.1` and allow minor releases.


benlorenz (approved):
Looks good, a bunch of references now show the volume / series that was previously missing and the link is often in a better place (i.e. at the title instead of the series).

fingolfin (approved):

PR #3004, opened 2023-11-09
Clarify the right action on polynomials

joschmitt:
This is almost only comments, but I just thought a lot about left and right actions on polynomials for The Book and wanted to clarify the choice in OSCAR also in the code. (In particular, since `right_action` de facto computes the product of the matrix by the column vector of variables, which made me question several times whether left is right.)

### DIFF
--- a/src/Groups/action.jl
+++ b/src/Groups/action.jl
@@ -286,18 +286,8 @@ end
     on_indeterminates(f::MPolyRingElem, p::PermGroupElem)
     on_indeterminates(f::FreeAssAlgElem, p::PermGroupElem)
     on_indeterminates(f::MPolyIdeal, p::PermGroupElem)
-    on_indeterminates(f::GAP.GapObj, p::MatrixGroupElem)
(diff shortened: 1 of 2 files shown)


src/Groups/action.jl
```diff
@@ -291,9 +291,13 @@ end
     on_indeterminates(f::MPolyIdeal, p::MatrixGroupElem)
 
 Return the image of `f` under `p`.
-If `p` is a `PermGroupElem` then it acts via permuting the indeterminates,
-if `p` is a `MatrixGroupElem` then it acts via evaluating `f` at the
+
+If `p` is a `PermGroupElem` then it acts via permuting the indeterminates.
```
joschmitt: Should we split this doc string for `on_indeterminates` between `PermGroupElem` and `MatrixGroupElem`?

afkafkafk13 (approved):
Thank you! The clarifications are very helpful.

joschmitt:
I split up the doc string now (and removed the `no-ci` so that the required tests can run).

PR #3005, opened 2023-11-10
add `od_from_atlas_group`, `od_from_p_subgroup`, and helpers

ThomasBreuer:
- `od_from_p_subgroup(chi, p)` tries to use the restriction of `chi` to the Sylow `p`-subgroup
- add `known_class_fusions`
- add optional argument to `induced_cyclic`
- add helper `possible_permutation_characters_from_sylow_subgroup`
- the new optional `min_char` argument in `isomorphic_group_over_finite_field` is used to guarantee that the reduction of an irreducible module is irreducible

(This branch tries to copy the relevant commits from the corrupted branch  #2960.)


experimental/OrthogonalDiscriminants/src/direct.jl
```diff
@@ -0,0 +1,221 @@
...
+    end
+
+    if ! haskey(info, :constituents)
+      # We do not yet know that `G` has character `chi`.
+      # Check whether the representation is abs. irred.
+      if p == 0
+        # We can check abs. irred. if we can reduce the repres.
+        # modulo a prime l that does not divide the group order.
+        l = next_prime(max([x[1] for x in factor(order(tbl))]...))
+        Gbar, _ = Oscar.isomorphic_group_over_finite_field(G, min_char = Int(l))
```
fieker, line 195: Ideally this should be 
```M = gmodule(Gbar)
is_abs_irr(M)
```
or similar. The "natural" gamodule of a matrix group is the missing bit...

experimental/OrthogonalDiscriminants/src/theoretical.jl
```diff
@@ -172,3 +175,153 @@ function od_for_specht_module(chi::GAPGroupClassFunction)
...
+  if cand != nothing
+    res = collect(Set([od_from_p_subgroup(chi, p, pi) for pi in cand]))
+    length(res) == 1 && return res[1]
+  end
+  return false, ""
+end
+
+function od_from_p_subgroup(chi::GAPGroupClassFunction, p::Int,
+                            pi::GAPGroupClassFunction)
+  tbl = chi.table
```
fieker, line 224: parent(chi) or character_table(chi)?

src/Rings/AbelianClosure.jl
```diff
@@ -459,6 +469,7 @@ function minimize(::typeof(CyclotomicField), a::nf_elem)
   return minimize(CyclotomicField, [a])[1]
 end
 
+#TODO: document this!
```
fieker, line 472: this function is ambigous: the element defines an order - which has a conductor as an order, and a field. The notions are different. So a comment/ doc would be important

fieker (approved):

ThomasBreuer:
@fieker Thanks for the feedback. I am going to address your comments.
(No need to ask for not merging, since some required tests fail.)

PR #3006, opened 2023-11-10
PolyhedralGeometry: Only consider relevant facets in star_subdivision

lkastner:
If we also consider irrelevant facets, we might get non-containment, where there is containment in reality.


HereAround (approved):
Thank you @lkastner for the super quick fix! Wonderful!

PR #3008, opened 2023-11-11
FTheoryTools: First extension to literature models over concrete bases

HereAround:


apturner (approved):
Thanks @HereAround, this looks good.

Ultimately I hope to allow, in some form, the specification of the aij as particular sections. But the signature you've created could allow for this via the `model_sections` argument, with a later change to the code, so we can delay that discussion for now.

Also I feel that we should later change `desired_base_space` -> `base_space`, but this is also fine for now. Approved!

PR #3012, opened 2023-11-13
FTheoryTools: Overhaul resolve/blowup of global Tate models

HereAround:
@apturner As discussed.


apturner (approved):
Thanks @HereAround. As you mention in the comments, there are some things for us to fix later, but this is great for now.

The one other thing I will mention that needs addressing in the future is that the strict transform code assumes the order of the variables of the original and the new polynomial rings is the same. This is probably guaranteed by the existing code for toric blowups, but this does mean our code relies on a convention choice made elsewhere in OSCAR. That said, this problem probably becomes moot once we generalize beyond the blowup center being generated by indeterminates.

PR #3015, opened 2023-11-14
ToricGeometry: Eliminate set_attributes also from deprecations

lkastner:


HereAround (approved):
Thank you @lkastner. Great spot!

PR #3017, opened 2023-11-14

HereAround:
1. Bugfix - literature model returned model prematurely
2. Introduce attribute partially_resolved and display this via the print method
3. ~~Change desired_base_space argument for literature model to base_space~~ I get the error `objects of type NormalToricVariety are not callable`. Let us address this separately.

@apturner: As discussed


apturner (approved):
Thank you, @HereAround!

PR #3018, opened 2023-11-14
different data format for pc group serialization

ThomasBreuer:
Use arrays of tuples `(i, word)` resp. `(j, i, word)` instead of two parallel arrays
This looks  more natural (would apply also for other objects such as sparse matrices).

### DIFF
--- a/src/Serialization/GAP.jl
+++ b/src/Serialization/GAP.jl
@@ -98,7 +98,7 @@ install_GAP_serialization(:IsFreeGroup,
         save_object(s, wfilt, :wfilt)
         # rank and names of generators
         Xnames = GAP.getbangproperty(elfam, :names)::GapObj
-        if GAP.Globals.Length(Xnames)::GAP.Obj == GAP.Globals.infinity
+        if length(Xnames) == GAP.Globals.infinity
           # store the initial names and the prefix
           prefix = GAP.getbangindex(Xnames, 1)::GapObj
           save_object(s, string(prefix), :nameprefix)
@@ -133,10 +133,10 @@ install_GAP_deserialization(
         fam = GAPWrap.ElementsFamily(GAPWrap.FamilyObj(F))
         Ggens = [GAPWrap.ObjByExtRep(fam, GapObj(x, true)) for x in generators]
         # Create the subgroup.
-        G = GAP.Globals.SubgroupNC(F, GapObj(Ggens))
+        G = GAP.Globals.SubgroupNC(F, GapObj(Ggens))::GapObj
       else
         # Create a new full free group.
-        wfilt = getproperty(GAP.Globals, load_object(s, Symbol, :wfilt))
+        wfilt = getproperty(GAP.Globals, load_object(s, Symbol, :wfilt))::GapObj
         if haskey(s, :nameprefix)
           # infinite rank
           prefix = load_node(s, :nameprefix) do nameprefix
@@ -145,12 +145,12 @@ install_GAP_deserialization(
           init = load_node(s, :names) do names
             GapObj([GapObj(x) for x in names], true)
           end
-          G = GAP.Globals.FreeGroup(wfilt, GAP.Globals.infinity, prefix, init)
+          G = GAP.Globals.FreeGroup(wfilt, GAP.Globals.infinity, prefix, init)::GapObj
         else
           init = load_node(s, :names) do names
             GapObj([GapObj(x) for x in names], true)
           end
-          G = GAP.Globals.FreeGroup(wfilt, init)
+          G = GAP.Globals.FreeGroup(wfilt, init)::GapObj
         end
       end
       return G
@@ -204,7 +204,7 @@ install_GAP_deserialization(
         gens = [GAPWrap.ObjByExtRep(elfreefam, GapObj(x, true)) for x in generators]
         Ggens = [GAPWrap.ElementOfFpGroup(elfam, x) for x in gens]
         # Create the subgroup.
-        G = GAP.Globals.SubgroupNC(F, GapObj(Ggens))
+        G = GAP.Globals.SubgroupNC(F, GapObj(Ggens))::GapObj
       else
         # Create a new full f.p. group.
         F = load_typed_object(s, :freeGroup)
@@ -223,7 +223,7 @@ install_GAP_deserialization(
 #   we do not support (de)serialization of the stored rws,
 #   thus we need not (de)serialize its underlying free group etc.
 
-# utility, turn an expomemt vector `[a_1, a_2, ..., a_n]`
+# utility, turn an exponent vector `[a_1, a_2, ..., a_n]`
 # into `[1, a_1, 2, a_2, ..., n, a_n]`
 function _free_group_extrep_from_exponents(exps::Vector{Int})
   res = Int[]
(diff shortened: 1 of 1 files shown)


antonydellavecchia:
can you post a comparison of the files? There might be a way to use save_object and load_object if there is a way to know the type from some point in tree closer to the root? This is essentially what save_type_params / load_type_params tries to do. For example if you know the type you are serializing will always use some subtype to serialize, you dont need to document that in the file and you can just use save_object which will store the data that needs to be passed to that type. Then on load since you know the structure of the subtype you can just force the data to have that subtype.

some sort of minmal example would be this
```
julia> data = [(1, 2), (3, 4)]
julia> Oscar.save_object(s, data)
[["1","2"],["3","4"]]

```
where you store data at some key :power_relators

then when loading at :power_relators you know that the data is supposed to fit into some specific type
in this example tuple{Int, Int} but I didn't need to add that to the file.

I am not 100% sure this would work in your case? but maybe?

src/Serialization/GAP.jl
```diff
@@ -233,33 +233,31 @@ install_GAP_serialization(:IsPcGroup,
...
-            push!(rels, _free_group_extrep_from_exponents(
-              Vector{Int}(GAP.Globals.ExponentsOfPcElement(fullpcgs, ne))))
+            push!(rels, (i, _free_group_extrep_from_exponents(
+              Vector{Int}(GAP.Globals.ExponentsOfPcElement(fullpcgs, ne)))))
           end
         end
-        save_object(s, reli, :power_reli)
-        save_object(s, rels, :power_rels)
+        if length(rels) != 0
+          save_typed_object(s, rels, :power_rels)
```
antonydellavecchia:
```suggestion
          save_data_array(s, :power_rels) do
            save_object(s, rels)
          end
```

src/Serialization/GAP.jl
```diff
@@ -282,18 +280,19 @@ install_GAP_deserialization(:IsPcGroup,
             length(relord))
       fam = GAPWrap.ElementsFamily(GAPWrap.FamilyObj(F))
       rws = GAP.Globals.SingleCollector(F, GapObj(relord))::GapObj
-      reli = load_object(s, Vector, d[:power_reli], Int)
-      rels = load_object(s, Vector, d[:power_rels], (Vector{Int}, Int))
-      for k in 1:length(reli)
-        GAP.Globals.SetPower(rws, reli[k],
-          GapObj(GAPWrap.ObjByExtRep(fam, GapObj(rels[k]))))
+      if haskey(d, :power_rels)
+        rels = load_typed_object(s, d[:power_rels])
```
antonydellavecchia, line 284: loaded_rels = Tuple{Int, Vector{Int}[]
for data_rel in rels
  loaded_rel = load_object(s, Tuple{Int, Vector{Int}, data_rel)
  push!(loaded_rels, loaded_rel)
end

antonydellavecchia:
```suggestion
        rels = Tuple{Int, Vector{Int}[]
        for data_rel in d[:power_rels]
          loaded_rel = load_object(s, Tuple{Int, Vector{Int}}, data_rel)
          push!(rels, loaded_rel)
        end
```

src/Serialization/GAP.jl
```diff
@@ -233,33 +233,31 @@ install_GAP_serialization(:IsPcGroup,
...
-                Vector{Int}(GAP.Globals.ExponentsOfPcElement(fullpcgs, ne))))
+              push!(rels, (j, i, _free_group_extrep_from_exponents(
+                Vector{Int}(GAP.Globals.ExponentsOfPcElement(fullpcgs, ne)))))
             end
           end
         end
-        save_object(s, reli, :comm_reli)
-        save_object(s, rels, :comm_rels)
+        if length(rels) != 0
+          save_typed_object(s, rels, :comm_rels)
```
antonydellavecchia: this can be dealt with similar to above situation

src/Serialization/GAP.jl
```diff
@@ -282,18 +280,19 @@ install_GAP_deserialization(:IsPcGroup,
...
+        end
       end
-      reli = load_object(s, Vector, d[:comm_reli], (Vector{Int}, Int))
-      rels = load_object(s, Vector, d[:comm_rels], (Vector{Int}, Int))
-      for k in 1:length(reli)
-        (j, i) = reli[k]
-        GAP.Globals.SetCommutator(rws, j, i,
-          GapObj(GAPWrap.ObjByExtRep(fam, GapObj(rels[k]))))
+      if haskey(d, :comm_rels)
+        rels = load_typed_object(s, d[:comm_rels])
```
antonydellavecchia: similar to above I believe?

src/Serialization/GAP.jl
ThomasBreuer: Ah, I see:
If the data file does not specify the type of the array entries then I can run over the entries and set the type for each of them, instead of trying to specify the type of the whole array.
I will try this.

src/Serialization/GAP.jl
antonydellavecchia:
```suggestion
          save_object(s, rels, :power_rels)
```
this should actually work better

src/Serialization/GAP.jl
ThomasBreuer: Wait: I get the following.
```
julia> rel = Any["2", Any["3", "1"]];

julia> @which Oscar.load_object(s, Tuple{Int, Vector{Int}}, rel)
ERROR: no unique matching method found for the specified argument types
```
The code in `Serialization/containers.jl` suggests that I should try
```
julia> Oscar.load_object(s, Tuple{Int, Vector{Int}}, rel, [Int, [Vector{Int}, Int]])
(2, [3, 1])
```
Looks complicated but seems to work. I will update the pull request.

antonydellavecchia: ah ya ok. Ya this is due to the fact that arbitrary tuples can get complicated. I will hopefully get round to simplifying this at some point.

ThomasBreuer: The message for me is:
- Avoid nested vectors.
- Vectors of tuples are fine (although they look like nested vectors in the data files), but deserialize each tuple separately. This way, it is not necessary to store type information about these objects in the data files.

src/Serialization/GAP.jl
```diff
@@ -282,18 +276,17 @@ install_GAP_deserialization(:IsPcGroup,
             length(relord))
       fam = GAPWrap.ElementsFamily(GAPWrap.FamilyObj(F))
       rws = GAP.Globals.SingleCollector(F, GapObj(relord))::GapObj
-      reli = load_object(s, Vector, d[:power_reli], Int)
-      rels = load_object(s, Vector, d[:power_rels], (Vector{Int}, Int))
-      for k in 1:length(reli)
-        GAP.Globals.SetPower(rws, reli[k],
-          GapObj(GAPWrap.ObjByExtRep(fam, GapObj(rels[k]))))
+      for data_rel in d[:power_rels]
+        (i, elm) = load_object(s, Tuple{Int, Vector{Int}}, data_rel,
```
antonydellavecchia: I am not sure if this work?
```suggestion
        (i, elm) = load_object(s, Tuple, data_rel, [Int, Vector{Int}])
```

in any case you can certainly just replace Tuple{Int, Vector{Int}} with Tuple which should help a bit

src/Serialization/GAP.jl
```diff
@@ -282,18 +276,17 @@ install_GAP_deserialization(:IsPcGroup,
...
+                     [Int, [Vector{Int}, Int]])
+        GAP.Globals.SetPower(rws, i,
+          GapObj(GAPWrap.ObjByExtRep(fam, GapObj(elm))))
       end
-      reli = load_object(s, Vector, d[:comm_reli], (Vector{Int}, Int))
-      rels = load_object(s, Vector, d[:comm_rels], (Vector{Int}, Int))
-      for k in 1:length(reli)
-        (j, i) = reli[k]
+      for data_rel in d[:comm_rels]
+        (j, i, elm) = load_object(s, Tuple{Int, Int, Vector{Int}}, data_rel,
```
antonydellavecchia: similarly to above
```suggestion
        (j, i, elm) = load_object(s, Tuple, data_rel,
```

src/Serialization/GAP.jl
```diff
@@ -282,18 +276,17 @@ install_GAP_deserialization(:IsPcGroup,
...
+                        [Int, Int, [Vector{Int}, Int]])
```
antonydellavecchia: Maybe? this needs to be checked
```suggestion
                        [Int, Int, Vector{Int}])
```

antonydellavecchia (approved):
some minor details on simplify a bit, otherwise looks good

src/Serialization/GAP.jl
antonydellavecchia: or it might be this, not sure?
```suggestion
                        [Int, Int, [Vector, Int]])
```

antonydellavecchia: or?
```suggestion
                        [Int, Int, [Vector, Int]])
```

ThomasBreuer: Yes, the latter works. Thanks.

Concerning the proposal to specify just `Tuple` as the second argument of `load_object`: Yes, this also works.
On the other hand, the second argument is logically the right place for specifying the desired type of the result (if it would be possible to omit the fourth argument).

I am going to update the pull request.

antonydellavecchia: yes I also agree, I think the issue is there isn't currently a function that can break down the second argument into all its params in the right way so that I can load an arbitrary tuple. I will make an attempt at dealing with this in a later pull request.

ThomasBreuer:
@antonydellavecchia I have adjusted the proposed changes to the changed syntax, the code became simpler now.

antonydellavecchia (approved):
looks good

PR #3021, opened 2023-11-15
Further functionality for hyper complexes

HechtiDerLachs:
This PR contains new functionality for hypercomplexes.

1. Morphisms of hypercomplexes `f : c -> d`. These produce morphisms `c[i] -> d[i + offset]` on request. A template for how to implement your own morphism is provided. 
2. As a first example we implement a class which lifts a given map `c[0] -> d[0]` to a map of complexes `c -> d` (if possible). In particular, this implements liftings through projective resolutions.
3. Free resolutions of modules: You can now ask for a free resolution of type `SimpleFreeResolution` using `free_resolution(SimpleFreeResolution, M)` on your favorite module `M`. This will provide you with a hypercomplex of dimension 1 which is quasi-isomorphic to your module considered as a complex in degree 0. The augmentation map is also returned as a morphism of hypercomplexes where the codomain is a complex with just one entry `M` at `0`. All of the exported functionality for `FreeResolution` (the type used up to now) is now also available for `SimpleFreeResolution`s, i.e. `betti_table` and `minimal_betti_table`. 
4. For hypercomplexes of graded free modules over a standard-graded ring one can now take strands, i.e. slices of a fixed degree. These come with the natural inclusion map given as a morphism of complexes.
5. For any `AbsSimpleComplex` (hypercomplexes of dimension 1) `c` of free modules you can now call `simplify(c)` and you get a simplified complex where there are no units in the representing matrices of the (co-)boundary maps. These simplification come with maps in both directions inducing a homotopy equivalence. 
6. Koszul complexes can now also be built lazy using `koszul_complex(KoszulComplex, ...)`, i.e. by specifying the desired type of the output to be `KoszulComplex`. 
7. You can take the total complex of any reasonably bounded hypercomplex using `total_complex`. 
8. For any two hypercomplexes `c` and  `d` you can call `hom(c, d)` to produce a new hypercomplex of corresponding hom-modules. For the case where either `c` or `d` is a single module concentrated in degree 0, this is the same as the usual hom. 
9. For `C = hom(c, d)` as above, any element `v` in the `k`-th homology of the total complex of `C` corresponds to a morphism of 1-dimensional complexes `phi : tot(c) -> tot(d)`. We provide an interpretation map which produces that morphism from `v`. This reduces to various classical ext-computations for the particular cases. 
10. Views of complexes: From any complex `c` one can now take ranges and slices. For instance `c[2, 4:7]` will produce a 1-dimensional complex with range `4:7` which is precisely the slice of `c` at index `2` in the first dimension in the range `4:7`. This can also be used to consider an `n`-dimensional hypercomplex as a hypercomplex of dimension `n+1`. 
11. Shifted complexes: Any hypercomplex `c` can be shifted in it's homological degrees. 
12. We provide a template for implementing your own new hypercomplex class. 
13. Tensor products of arbitrary hypercomplexes have also been implemented. We should think about whether or not we want the tensor products of `ComplexOfMorphism`s to be rerouted to that generic implementation.
14. Linear strands of complexes according to Eisenbud: The Geometry of Syzygies, chapter 7. We provide methods to compute a linear strand and its inclusion and then also form the cokernel of this inclusion with the associated projection. The specialty is that this cokernel is again a complex of free modules and it is also realized as such so that one can proceed with linear strands inductively. 

See the various test files to get a taste of the new functionality. Other than that, the documentation is rather poor. But everything is similar to the double complexes which have some documentation already.


HechtiDerLachs:
A good place to start checking out the new features are the test files. I added a few comments there. 

Part of the code for the simplification is imported from #3002 . If its form here is convincing, I can refactor that PR.

src/Combinatorics/OrderedMultiIndex.jl
```diff
@@ -183,3 +183,50 @@ function ordered_multi_index(k::Int, p::Int, n::Int)
...
+mutable struct MultiIndicesOfDegree
+  n::Int
+  d::Int
+
+  function MultiIndicesOfDegree(n::Int, d::Int)
+    @assert n >= 0 "length of index must be non-negative"
+    @assert d >= 0 "degree must be non-negative"
+    return new(n, d)
+  end
+end
```
HechtiDerLachs, line 195: @joschmitt : I needed an iterator over all multiindices `I = (i_1, i_2, ..., i_n)` with `i_k >= 0` and `i_1 + i_2 + ... + i_n = d` here. I didn't want to go through your `monomials_of_degree` things, because I would have to create a dummy ring for that and then extract the exponent vectors. 

Question: Are you using such an iterator in the internals of your code? If not, do you think it would make sense to build your iterator on this one?

src/Modules/UngradedModules.jl
```diff
@@ -180,6 +180,22 @@ function (==)(F::FreeMod, G::FreeMod)
...
+
+function (==)(F::FreeMod, G::SubquoModule)
+  base_ring(F) === base_ring(G) || return false
+  ambient_free_module(G) === F || return false
+  if isdefined(G, :quo) 
+    iszero(G.quo) || return false
+  end
+  all(e -> e in G, gens(F)) || return false
+  return true
+end
```
HechtiDerLachs: @jankoboehm : Without these methods `is_surjective` would fail for homomorphisms with a `FreeMod` as codomain. Do you agree with adding these?

src/Modules/UngradedModules.jl
```diff
@@ -1024,11 +1040,11 @@ julia> a2 == b
 true
 ```
 """
-function hom(F::FreeMod, M::ModuleFP{T}, V::Vector{<:ModuleFPElem{T}}) where T
+function hom(F::FreeMod, M::ModuleFP{T}, V::Vector{<:ModuleFPElem{T}}; check::Bool=true) where T
   base_ring(F) === base_ring(M) || return FreeModuleHom(F, M, V, base_ring(M))
   return FreeModuleHom(F, M, V)
 end
```
HechtiDerLachs: @jankoboehm : I added a dummy `check` flag. It's not doing anything, but it allows to call the hom constructor for any kind of modules with the same signature now. I hope this is ok from your side?

src/Modules/UngradedModules.jl
```diff
@@ -6938,6 +6954,24 @@ function *(h::ModuleFPHom{T1, T2, <:Any}, g::ModuleFPHom{T2, T3, <:Any}) where {
...
+end
+
+function *(h::ModuleFPHom{T1, T2, <:Any}, g::ModuleFPHom{T2, T3, Nothing}) where {T1, T2, T3}
+  @assert codomain(h) === domain(g)
+  return hom(domain(h), codomain(g), 
+             Vector{elem_type(codomain(g))}([g(h(x)) for x = gens(domain(h))]), 
+             base_ring_map(h)
+            )
+
+end
```
HechtiDerLachs: These were necessary for composition of module homomorphisms where one has a non-trivial ring map and the other doesn't.

src/Combinatorics/OrderedMultiIndex.jl
joschmitt: I think in combinatorics this is called $d$-multicombinations of $n$ and yes, it makes sense that there is such an iterator that produces the integer vectors and then one can for example build the monomials out of it.
(I didn't implement it like this back then because I hoped that some kind combinatorics expert would come round and do it properly... For the same reason, I also have my own multiset partitions somewhere in the invariant theory code...😞)

Caveat: Your algorithm appears to be different from what I implemented (for which I don't have a reference, maybe I just took what they have in `Combinatorics.jl`, I don't remember), so I would like to compare these to approaches runtime-wise before we delete "my" implementation. I can also do this once this pull request is merged.

HechtiDerLachs: Sounds good! I also simply did a straightforward (and probably not very optimal) implementation. But as long as we don't have it, I think it's better to have *some* implementation rather than none. 

Comparing the speed of either of the two is good! Finding a combinatorics expert to tweak our stuff would be even better. But that can also be done later since the user facing part will not change.

src/Modules/UngradedModules.jl
jankoboehm: What is it doing for other signatures? Does it apply here?

src/Modules/UngradedModules.jl
jankoboehm: Fine.

src/Combinatorics/OrderedMultiIndex.jl
fieker: there is also a native julia version: Iterators.ProductIterator((1:2, 1:3, 1:4))
This is not quite the same, as this does not limit the sum of the values

joschmitt: I don't understand how one gets indices which sum up to a given `d` with this?

HechtiDerLachs:
This has now been rebased on #3041  (and will hence be rebased again once that is merged). The templates are now taken care of for the moment and hopefully the tests pass.

experimental/DoubleAndHyperComplexes/src/Objects/total_complexes.jl
```diff
@@ -0,0 +1,247 @@
...
+  # produces this output and it is an ongoing effort to streamline this throughout 
+  # Oscar. Since such changes tend to happen with severe delay, if ever, we provide 
+  # a little workaround here. If this code does not run for your type of chains, 
+  # you may try two things:
+  #
+  #   1) Overwrite the method for `_direct_sum` below for your type and wrap the 
+  #      original method for `direct_sum` so that the anticipated output is produced
+  #
+  #   2) If that does not work, you may also overwrite the whole method for production 
+  #      of the direct sum here.
```
HechtiDerLachs, line 73: @fingolfin :
This is where we stopped our discussion yesterday. I hope that adding this check and comments resolves the problem. The point is: If the method for `direct_sum` is implemented correctly (meaning it produces the anticipated output), then the code should just run. But we can not assume this and this provides an easy and transparent workaround. 

Edit: A few lines below you can see that the default implementation of `_direct_sum` just deflects to `direct_sum` so that by default everything should work (but won't in 78.5% of all cases..). Same for `_tensor_product` below.

experimental/DoubleAndHyperComplexes/src/Objects/tensor_products.jl
```diff
@@ -196,3 +196,175 @@ function tensor_product(list::Vector{T}) where {T<:ComplexOfMorphisms}
...
+  # produces this output and it is an ongoing effort to streamline this throughout 
+  # Oscar. Since such changes tend to happen with severe delay, if ever, we provide 
+  # a little workaround here. If this code does not run for your type of chains, 
+  # you may try two things:
+  #
+  #   1) Overwrite the method for `_tensor_product` below for your type and wrap the 
+  #      original method for `tensor_product` so that the anticipated output is produced
+  #
+  #   2) If that does not work, you may also overwrite the whole method for production 
+  #      of the tensor products here.
```
HechtiDerLachs, line 244: @fingolfin : same here.

experimental/DoubleAndHyperComplexes/src/Objects/tensor_products.jl
```diff
@@ -196,3 +196,186 @@ function tensor_product(list::Vector{T}) where {T<:ComplexOfMorphisms}
...
+  internal_complex::HyperComplex{ChainType, MorphismType}
+  factors::Vector{<:AbsHyperComplex}
+
+  function HCTensorProductComplex(
+      factors::Vector{<:AbsHyperComplex}
+    )
+    CT = minimal_common_supertype(chain_type.(factors)...)
+    MT = minimal_common_supertype(morphism_type.(factors)...)
+    chain_fac = HCTensorProductChainFactory(CT, factors)
+    map_fac = HCTensorProductMapFactory(MT, factors)
```
HechtiDerLachs: @thofma : This is where I was using `supertype` (in the methods for `minimal_common_supertype` below). I want to make a best guess on the `ChainType` and the `MorphismType` of the resulting complex. Similar to when you put things of different concrete type in a `Vector`, but you don't want the result to be of type `Vector{Any}`, but of some reasonable (abstract) supertype. 

I'd be happy to hear about other solutions!

experimental/DoubleAndHyperComplexes/src/Objects/tensor_products.jl
```diff
@@ -196,3 +196,186 @@ function tensor_product(list::Vector{T}) where {T<:ComplexOfMorphisms}
...
+### The concrete struct
+@attributes mutable struct HCTensorProductComplex{ChainType, MorphismType} <: AbsHyperComplex{ChainType, MorphismType} 
+  internal_complex::HyperComplex{ChainType, MorphismType}
+  factors::Vector{<:AbsHyperComplex}
+
+  function HCTensorProductComplex(
+      factors::Vector{<:AbsHyperComplex}
+    )
+    CT = reduce(typejoin, chain_type.(factors))
+    MT = reduce(typejoin, morphism_type.(factors))
```
fingolfin, line 342: Note that this first allocates a vector to store the result of `chain_type.(factors)`. This can be avoided by e.g. using `mapreduce`
```suggestion
    CT = mapreduce(chain_type, typejoin, factors)
    MT = mapreduce(morphism_type, typejoin, factors)
```

fingolfin: Beyond that, I think usage of `typejoin` warrants a comment justifying its use... I.e. the big picture "why" it is used.

Otherwise I would have no idea how to debug this code if I run into it again in 2 years.

experimental/DoubleAndHyperComplexes/src/Objects/tensor_products.jl
```diff
@@ -196,3 +196,186 @@ function tensor_product(list::Vector{T}) where {T<:ComplexOfMorphisms}
...
+  j = Vector{Vector{Int}}()
+  k = 0
+  for f in fac.factors
+    push!(j, i[k+1:k+dim(f)])
+    k = k + dim(f)
+  end
+  d = length.(j)
+  l = findfirst(l->sum(d[1:l]; init=0)>=p, 1:length(j))
+  l === nothing && error("mapping direction out of bounds")
+  return can_compute_map(fac.factors[l], p - sum(d[1:l-1]; init=0), Tuple(j[l]))
```
fingolfin, line 330: It may not matter (because `length(j)` is "always" small, perhaps???) but the `findfirst` line is needlessly using a quadratic algorithm to solve a problem that's easily solvable in linear time. E.g. inside the `for` loop above. In fact, isn't `k` equal to the partial sums you are looking at here? And you are only using `j[l]` later ... So can't you do something like this:
```suggestion
  i = collect(I)
  # decompose I into its individual indices
  k = 0
  for f in fac.factors
    if k + dim(f) >= p
      r = i[k+1:k+dim(f)]
      return can_compute_map(f, p - k, Tuple(r))
    end
    k = k + dim(f)
  end
  error("mapping direction out of bounds")
```

experimental/DoubleAndHyperComplexes/src/Morphisms/ext.jl
```diff
@@ -0,0 +1,415 @@
...
+  end
+end
+
+function (fac::HomChainFactory)(hc::AbsHyperComplex, i::Tuple)
+  I = collect(i)
+  d = fac.d
+  c = fac.c
+  i1 = I[1:dim(d)]
+  i2 = I[dim(d)+1:end]
+  return hom(d[Tuple(-i1)], c[Tuple(i2)])[1]
```
fingolfin, line 20: Why do you need to `collect(i)` here and elsewhere? Can't you just do
```suggestion
  d = fac.d
  c = fac.c
  i1 = i[1:dim(d)]
  i2 = i[dim(d)+1:end]
  return hom(d[-i1], c[i2])[1]
```

experimental/DoubleAndHyperComplexes/src/Morphisms/cartan_eilenberg_resolutions.jl
```diff
@@ -0,0 +1,59 @@
...
+function (l::MapLifter{T})(Phi::AbsHyperComplexMorphism, I::Tuple) where {T<:ModuleFPHom}
+  i = first(I)
+  i == l.start_index && return l.orig_map
+
+  inc = 1 
+  if is_chain_complex(domain(Phi))
+    inc = -1
+  end
+  psi = Phi[Tuple(i+inc)]
+  dom_map = map(domain(Phi), 1, I)
```
fingolfin, line 33: This line really confused me until I realized that the standard Julia function `map` is being overloaded to retrieve "maps" from hypercomplexes, free resolutions and even `FreeMod` / `MatElem` (in `src/Modules/UngradedModules/SubQuoHom.jl`). I think that's a really bad idea, and would urge y'all to reconsider this name (CC @jankoboehm @fieker @thofma).

experimental/DoubleAndHyperComplexes/src/Morphisms/cartan_eilenberg_resolutions.jl
```diff
@@ -0,0 +1,59 @@
...
+  cod_map = map(codomain(Phi), 1, Tuple(i + l.offset))
+  @assert codomain(dom_map) === domain(psi)
+  @assert codomain(cod_map) === codomain(psi)
+
+  x = gens(domain(dom_map))
+  y = dom_map.(x)
+  z = psi.(y)
+  w = [preimage(cod_map, a) for a in z]
```
fingolfin, line 41: One the primary benefits of broadcast operators is that you can join them to avoid redundant allocations by fusing operations. For example:
```julia
julia> f(v) = cos.(sin.(v));

julia> g(v) = begin w = sin.(v) ; cos.(w) end;

julia> @btime f(v);
  52.063 ns (1 allocation: 80 bytes)

julia> @btime g(v);
  67.810 ns (2 allocations: 160 bytes)
```
So that "naive" version allocates two vectors, the one using chained broadcasting just one.

That would suggest at least doing
```suggestion
  z = psi.(dom_map.(x))
  w = [preimage(cod_map, a) for a in z]
```
But broadcasting also supports handling "scalars" differently to allow using it for multiple arguments as well.

I didn't test it, but maybe even this works:
```suggestion
  w = preimage.(cod_map, psi.(dom_map.(x)))
```

Now of course this is a fairly minor operation, but I am trying to raise awareness of this nevertheless, as all these tiny allocations everywhere in our code base add up -- one of the primary reasons our test suite uses many *gigabytes* for RAM for many computations for which it really shouldn't (the other of course is memory used to compile code and store the result).

src/Combinatorics/OrderedMultiIndex.jl
lkastner:
```
julia> n = 5
5

julia> d = 3
3

julia> I = identity_matrix(ZZ,d)
[1   0   0]
[0   1   0]
[0   0   1]

julia> A = vcat(I, -I)
[ 1    0    0]
[ 0    1    0]
[ 0    0    1]
[-1    0    0]
[ 0   -1    0]
[ 0    0   -1]

julia> b = vcat(n*ones(Int, d),zeros(Int,d))
6-element Vector{Int64}:
 5
 5
 5
 0
 0
 0

julia> C = matrix(ZZ,[ones(Int,d)])
[1   1   1]

julia> dd = [n]
1-element Vector{Int64}:
 5

julia> P = polyhedron((A,b),(C,dd))
Polyhedron in ambient dimension 3

julia> lattice_points(P)
21-element SubObjectIterator{PointVector{ZZRingElem}}:
 [0, 0, 5]
 [0, 1, 4]
 [0, 2, 3]
 [0, 3, 2]
 [0, 4, 1]
 [0, 5, 0]
 [1, 0, 4]
 [1, 1, 3]
 [1, 2, 2]
 [1, 3, 1]
 [1, 4, 0]
 [2, 0, 3]
 [2, 1, 2]
 [2, 2, 1]
 [2, 3, 0]
 [3, 0, 2]
 [3, 1, 1]
 [3, 2, 0]
 [4, 0, 1]
 [4, 1, 0]
 [5, 0, 0]
```

lkastner: Probably it is fastest if you just compute the `partitions` of your number of the right size(s) and then extend via orbit computation.

joschmitt: I quickly edited the existing function `monomials_of_degree` to not create monomials, but only return the exponent vectors, and compared the runtime:
```
julia> function via_polymake(a::Int, b::Int)
          # a is the number of variables/indices, b the degree/sum
           I = identity_matrix(ZZ, a)
           A = vcat(I, -I)
           bb = vcat(b*ones(Int, a), zeros(Int, a))
           C = matrix(ZZ, [ones(Int, a)])
           dd = [b]
           P = polyhedron((A, bb), (C, dd))
           return lattice_points(P)
       end
via_polymake (generic function with 1 method)

julia> a = 5; b = 5; # a = number of variables, b = sum or degree

julia> R, x = polynomial_ring(QQ, a);

julia> @btime collect(monomials_of_degree(R, b)); # edited so it doesn't create monomials
  2.742 μs (255 allocations: 16.97 KiB)

julia> @btime collect(Oscar.MultiIndicesOfDegree(a, b));
  5.748 μs (449 allocations: 32.53 KiB)

julia> @btime via_polymake(a, b);
  14.657 ms (16320 allocations: 136.42 KiB)

julia> a = 5; b = 10;

julia> @btime collect(monomials_of_degree(R, b));
  21.440 μs (2005 allocations: 133.28 KiB)

julia> @btime collect(Oscar.MultiIndicesOfDegree(a, b));
  35.163 μs (2939 allocations: 209.47 KiB)

julia> @btime via_polymake(a, b);
  15.915 ms (61071 allocations: 486.05 KiB)

julia> a = 10; b = 10;

julia> R, x = polynomial_ring(QQ, a);

julia> @btime collect(monomials_of_degree(R, b));
  2.284 ms (184760 allocations: 16.21 MiB)

julia> @btime collect(Oscar.MultiIndicesOfDegree(a, b));
  5.424 ms (347856 allocations: 34.76 MiB)

julia> @btime via_polymake(a, b);
  367.535 ms (13538535 allocations: 103.31 MiB)

julia> a = 15; b = 10;

julia> R, x = polynomial_ring(QQ, a);

julia> @btime collect(monomials_of_degree(R, b));
  114.180 ms (3922516 allocations: 404.01 MiB)

julia> @btime collect(Oscar.MultiIndicesOfDegree(a, b));
  410.142 ms (8706790 allocations: 1.10 GiB)

julia> @btime via_polymake(a, b);
  16.894 s (660509300 allocations: 4.92 GiB)

julia> a = 15; b = 15;

julia> @btime collect(monomials_of_degree(R, b));
  8.642 s (155117524 allocations: 15.60 GiB)

julia> @btime collect(Oscar.MultiIndicesOfDegree(a, b));
  21.106 s (297865757 allocations: 36.89 GiB)

julia> @btime via_polymake(a, b);
# Killed; 32GB RAM were not enough :(
```
So, in my opinion we should use what is implemented in `monomials_of_degree` for `MultiIndicesOfDegree` (and make `monomials_of_degree` use that iterator internally). I can do this once this pull request here is merged; I think I don't have the rights to push on other people's forks.
Besides the timings, for my application in invariant theory the code in `monomials_of_degree` has the advantage that it is computing things one by one. So if I only need the first 3 monomials nothing else is computed.

experimental/DoubleAndHyperComplexes/src/Objects/tensor_products.jl
HechtiDerLachs: That seems neat. Thanks.

experimental/DoubleAndHyperComplexes/src/Morphisms/ext.jl
HechtiDerLachs: I must have thought that the indexing with a range doesn't work for `Tuple`s. But apparently I was wrong.

experimental/DoubleAndHyperComplexes/src/Morphisms/cartan_eilenberg_resolutions.jl
HechtiDerLachs: I took this from the `ComplexOfMorphisms`. But I agree! Actually, this has already lead to quite confusing error messages, because sometimes it would try to use a method of the original `map` on my input.

@fieker : Your call for the `ComplexOfMorphisms`. I will then follow your lead.

src/Combinatorics/OrderedMultiIndex.jl
HechtiDerLachs: Nicely done! 

In principle you should be able to push directly to this PR if you check it out with `gh pr checkout 3021`. If you run the tests for the modules and those in `experimental/DoubleAndHyperComplexes/test/` before you push, then I think it's fine if you pushed your work here.

But it's also fine with me to do that later.

experimental/DoubleAndHyperComplexes/src/Morphisms/cartan_eilenberg_resolutions.jl
HechtiDerLachs: Yes, I get it. Thanks for the hints.

src/Combinatorics/OrderedMultiIndex.jl
joschmitt: I'm fairly certain I can't do this because I only have 'triage rights', so I can't 'Push to (write) the person or team's assigned repositories' (see https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization ). I never tried though. But I don't want to complain about this; I'm not sure I want the responsibility :)

experimental/DoubleAndHyperComplexes/test/free_resolutions.jl
```diff
@@ -0,0 +1,16 @@
...
+  F = graded_free_module(Pn, 1)
+  dualFIC = hom(FI, F);
+  # Duals invert the range from 0:10 to 0:-1:-10 etc. 
+  @test dualFIC[0] isa FreeMod
+  @test dualFIC[-1] isa FreeMod
+  @test dualFIC[-10] isa FreeMod
+  @test is_graded(dualFIC[0])
+  @test is_graded(dualFIC[-1])
+  @test is_graded(dualFIC[-10])
+  @test !Oscar.can_compute_index(dualFIC, 1)
```
HechtiDerLachs: @wdecker : Your requested example is now up and running here. 

I was mistaken in that the `free_resolution` command also returns the augmentation map. I guess, I wanted to stay consistent with what `free_resolution` usually does and did not yet work out the augmentation map.

HechtiDerLachs: Now the augmentation map is there.

experimental/DoubleAndHyperComplexes/test/linear_strands.jl
```diff
@@ -0,0 +1,42 @@
+@testset "linear strands of complexes of graded modules" begin
```
HechtiDerLachs, line 1: @wdecker : First version of linear strands are now here. I'm working on the induced maps.

experimental/DoubleAndHyperComplexes/test/linear_strands.jl
```diff
@@ -0,0 +1,63 @@
...
+  S, (x, y, z) = graded_polynomial_ring(QQ, [:x, :y, :z])
+
+  S1 = graded_free_module(S, [0])
+  I, inc = sub(S1, [f*S1[1] for f in gens(S)])
+  M = cokernel(inc)
+
+  res, aug = free_resolution(Oscar.SimpleFreeResolution, M)
+
+  str, inc_str = Oscar.linear_strand(res)
```
HechtiDerLachs, line 10: Here you get the linear strand together with its inclusion morphism. There is an optional keyword argument `degree` to specify a different initial degree.

experimental/DoubleAndHyperComplexes/test/linear_strands.jl
```diff
@@ -0,0 +1,63 @@
...
+
+  @test rank(str[0]) == 1
+  @test rank(str[1]) == 3
+  @test rank(str[2]) == 3
+  @test rank(str[3]) == 1
```
HechtiDerLachs, line 15: You get a homogeneous Koszul complex as expected. In fact, everything is equal to the linear strand here.

experimental/DoubleAndHyperComplexes/test/linear_strands.jl
```diff
@@ -0,0 +1,63 @@
...
+  S2 = graded_free_module(S, [0, 1])
+  I1, inc1 = sub(S2, [f*S2[1] for f in gens(S)])
+  I2, inc2 = sub(S2, [f*S2[2] for f in gens(S)])
+
+  I = I1 + I2
+  M, pr = quo(S2, I)
+
+  res, aug = free_resolution(Oscar.SimpleFreeResolution, M)
+
+  str, inc_str = Oscar.linear_strand(res)
```
HechtiDerLachs, line 41: This case is more interesting: A direct sum of the previous Koszul complex and a shift in degree of the same complex. The linear strand will only be the first one and the second will appear as the cokernel of the inclusion.

experimental/DoubleAndHyperComplexes/test/linear_strands.jl
```diff
@@ -0,0 +1,63 @@
...
+  @test rank(str[2]) == 3
+  @test rank(str[3]) == 1
+  @test iszero(str[4])
+  @test !Oscar.can_compute_index(str, (-1,))
+
+  @test compose(inc_str[(1,)], map(res, 1, (1,))) == compose(map(str, 1, (1,)), inc_str[(0,)])
+  @test compose(inc_str[(2,)], map(res, 1, (2,))) == compose(map(str, 1, (2,)), inc_str[(1,)])
+  @test compose(inc_str[(3,)], map(res, 1, (3,))) == compose(map(str, 1, (3,)), inc_str[(2,)])
+
+  K, pr = cokernel(inc_str)
```
HechtiDerLachs, line 54: This call to `cokernel` will dispatch to a special method which creates all cokernels again as free modules. It returns the resulting complex and the projection map. In particular, you can ask for the `linear_strand` of `K` in the next appropriate degree.

HechtiDerLachs:
Now even printing of free resolutions, strands, and duals should look half way nice. We can probably improve that also later.

HechtiDerLachs:
@wdecker : You should now be able to call `linear_strand` directly on your preferred `free_resolution` `res`. Note, however, that the inclusion map for that linear strand does not take values in `res`, because there are no maps for `ComplexOfMorphism`s and/or `FreeResolution`s. What happens internally is that `res` gets wrapped as an `AbsHyperComplex` and the inclusion takes values in that wrapper.

HechtiDerLachs:
@wdecker : Methods for `betti_table` and `minimal_betti_table` have been worked out and I didn't find anything else that seemed relevant and has been exported for `FreeResolution`. Hence the types here should be at least as good as what we had regarding their user facing functionality.

But compared to `FreeResolution` all the other functionality like hom, tensor, dualizing, and `simplify` (!) now runs out of the box for `SimpleFreeResolution` while one has to duplicate every such method for `FreeResolution`. In my opinion this is a clear advantage. 

What remains to be done is to hook up the Schreyer resolution and the like. But that's a different topic which has been postponed: @jankoboehm asked to look into this  in order to also learn about how the things work out here. In particular this will then probably also allow to compute resolutions only up to a certain step on request.

simonbrandhorst (approved):
in a conversation with @jankoboehm  and @HechtiDerLachs  
we concluded that this is ready for experimental.

jankoboehm (approved):

PR #2823, opened 2023-09-19
Draft - InvRing for SLn

Lax202:
For internal discussion. This new type - InvariantRing is meant to serve as a type for invariant rings of linearly reductive groups. As of now we can compute invariant rings of SLn in some representation. In the future we expect to have direct sums and tensor products as well (which does not work correctly as of now). 

Some feedback that I've already received:

1. The user interface should be changed so that the representation of SLn is accessible to the user.  In general the reductive group and its representation should have its own type that can be constructed by the user. 
2. As of now the types InvRing and InvariantRing are neither compatible nor compatibly named. (Suggestions for compatibility are welcome.)

PR #3025, opened 2023-11-17
Rename, extend, document and test `all_monomials`

joschmitt:
* I renamed `all_monomials`, which gives an iterator over all monomials of a polynomial ring of a given degree, to the more appropriate `monomials_of_degree` and exported (+test, document) this function.
* I added the variant `monomials_of_degree(::MPolyRing, ::Int, ::UnitRange)` which only works on a subset of the variables specified by the range and which should cover the use case in #2823 .
* I decided to make it throw an error for non-standard graded rings. So far it works for those rings by ignoring the grading. @HechtiDerLachs is this fine for you? I don't want to break your code or use case.

### DIFF
--- a/src/InvariantTheory/iterators.jl
+++ b/src/InvariantTheory/iterators.jl
@@ -1,23 +1,87 @@
-
 ################################################################################
 #
 #  All Monomials
 #
 ################################################################################
 
-# Return an iterator over all monomials of R of degree d
-all_monomials(R::MPolyRing, d::Int) = AllMonomials(R, d)
+@doc raw"""
+    monomials_of_degree(R::MPolyRing, d::Int)
+    monomials_of_degree(R::MPolyRing, d::Int, vars::Vector{Int})
+    monomials_of_degree(R::MPolyRing, d::Int, r::UnitRange{Int})
+
+Return an iterator over all monomials in `R` of degree `d`.
+An additional vector or range of integers may be given to specify a subset of
+variables of `R` to be used.
+
+# Examples
+```jldoctest
+julia> R, (x, y, z) = QQ["x", "y", "z"];
+
+julia> collect(monomials_of_degree(R, 3))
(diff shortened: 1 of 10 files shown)


src/InvariantTheory/iterators.jl
```diff
@@ -1,23 +1,85 @@
...
+ y*z^2
+ z^3
+
+julia> sum(monomials_of_degree(R, 4, 1:2))
+x^4 + x^3*y + x^2*y^2 + x*y^3 + y^4
+
+julia> sum(monomials_of_degree(R, 4, [1, 3]))
+x^4 + x^3*z + x^2*z^2 + x*z^3 + z^4
+```
+"""
```
thofma, line 39:
```suggestion
"""
monomials_of_degree

```

thofma: Then the docstring is attached to the function and not one particular method.

thofma:
Looking at the coverage, there seem to be a few (changed) lines missing. Did you test this locally or do we hope that nothing breaks?

HechtiDerLachs:
Fine with me. Thanks for the notification!

src/InvariantTheory/iterators.jl
joschmitt: Thanks, I did not know about this.

joschmitt:
> Looking at the coverage, there seem to be a few (changed) lines missing. Did you test this locally or do we hope that nothing breaks?

I added more tests now so that the actual iterator should be covered completely. Otherwise it is the renaming of a so far not exported function, so I hope it is fine?

thofma:
> I added more tests now so that the actual iterator should be covered completely. Otherwise it is the renaming of a so far not exported function, so I hope it is fine?

Sure. I will merge once CI shows green.

thofma (approved):

PR #2982, opened 2023-11-03
Add docs and tests for CartanMatrix.jl

felix-roehrich:
This is a follow up of #2689 and adds docs and tests for the CartanMatrix.jl file added in this PR. It also adds four new functions, name `cartan_symmterizer`, `cartan_bilinear_form`, `cartan_type` and `cartan_type_with_ordering`.
Work isn't completely finished, but I want to open the discussion for the naming of functions.

For one @lgoettgens suggested to rename `cartan_to_coxeter_matrix` to something like `coxeter_from_cartan_matrix` (to follow the convention of dst_src if I remember correctly).

If time permits I will add a function for Cartan matrices of affine type. My suggestion would be `cartan_matrix(fam::Symbol, rk::Int, aff::Int)`. I do not plan on handling the hyperbolic case, What is you opinion @fingolfin ?

PR #3027, opened 2023-11-20
LieAlgebras: Add some small helper functions around simple modules

lgoettgens:
This branch is based on https://github.com/oscar-system/Oscar.jl/pull/2982 and will be rebased once https://github.com/oscar-system/Oscar.jl/pull/2982 is merged.

This PR adds some small helpers (character, dominant character, dim, tensor product decomposition) for simple modules of ss Lie algebras (via highest weight) without explicitly constructing them.
@gfourier wished for this as an alternative to the broken [LiE CAS webpage](http://wwwmathlabo.univ-poitiers.fr/~maavl/LiE/form.html).


experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1242,3 +1228,154 @@ function tensor_power(V::LieAlgebraModule{C}, k::Int) where {C<:FieldElem}
...
+
+Construct the simple module of the Lie algebra `L` with highest weight `hw`.
+The actual construction is done in GAP.
+"""
+function simple_module(L::LieAlgebra, hw::Vector{Int})
+  @req is_dominant_weight(hw) "Not a dominant weight."
+  struct_consts = lie_algebra_simple_module_struct_consts_gap(L, hw)
+  dimV = size(struct_consts, 2)
+  V = abstract_module(L, dimV, struct_consts; check=false)
+  set_attribute!(V, :is_simple => true, :highest_weight => hw)
```
felix-roehrich: Instead of `set_attribute!(V, :is_simple => true, :highest_weight => hw)` myabe it would be better to just save the maximal weights (generically) and check for `dim < \infty` and `length(max_weights) == 1`.

lgoettgens: This line currently conveys no meaning at all, as there is no `is_simple` method for modules yet. So this will be changed again in the future either way.

thofma: Can you remove the sentence mentioning "GAP" from the docstring while you are at it?

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1242,3 +1228,154 @@ function tensor_power(V::LieAlgebraModule{C}, k::Int) where {C<:FieldElem}
...
+MSet{Vector{Int64}} with 6 elements:
+  [0, 0]
+  [1, 1] : 2
+  [2, 2]
+  [3, 0]
+  [0, 3]
+```
+"""
+function tensor_product_decomposition(L::LieAlgebra, hw1::Vector{Int}, hw2::Vector{Int})
+  @req is_dominant_weight(hw1) && is_dominant_weight(hw2) "Not a dominant weight."
```
felix-roehrich:
```suggestion
  @req is_dominant_weight(hw1) && is_dominant_weight(hw2) "Both weights must be dominant."
```

experimental/LieAlgebras/src/CartanMatrix.jl
```diff
@@ -442,3 +431,15 @@ function cartan_type_with_ordering(gcm::ZZMatrix; check::Bool=true)
 
   return type, ord
 end
+
+function is_cartan_type(S, n)
```
felix-roehrich: Please add type annotations and maybe also rename `S->fam` and `n->rk` to keep in line with the rest of the file. This functions is also exported, so please add some documentation.

experimental/LieAlgebras/src/DynkinDiagram.jl
```diff
@@ -0,0 +1,90 @@
+@doc raw"""
+    show_dynkin_diagram(fam::Symbol, rk::Int) -> Nothing
+
+Prints a string representation of the Dynkin diagram of the root system of
+the given cartan type.
+"""
+function show_dynkin_diagram(fam::Symbol, rk::Int)
```
felix-roehrich, line 7: Please add doctests/tests.

experimental/LieAlgebras/src/DynkinDiagram.jl
```diff
@@ -0,0 +1,90 @@
...
+
+  elseif fam == :F
+    if rk == 4
+      D = "1 - 2 >=> 3 - 4"
+    else
+      error("This root system doesn't exist.")
+    end
+  elseif fam == :G
+    if rk == 2
+      D = "1 >>> 2"
```
felix-roehrich:
```suggestion
      D = "1 <<< 2"
```
The second root is long (please double check).

experimental/LieAlgebras/src/LieAlgebraModule.jl
felix-roehrich: > This line currently conveys no meaning at all, as there is no `is_simple` method for modules yet. So this will be changed again in the future either way.

Then just remove the line ?

experimental/LieAlgebras/src/DynkinDiagram.jl
lgoettgens: Done. See cb099e2386099b33b7b5037937766ae74205f25e

felix-roehrich (approved):
lgtm

felix-roehrich:
Because of the order in which things were merged this needs an update.

lgoettgens:
This is ready to be merged.

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -371,6 +371,9 @@ function lie_algebra(R::Field, S::Symbol, n::Int; cached::Bool=true)
 
   # start temporary workaround # TODO: reenable code above
   type = only(root_system_type(rs))
+  if type == (:F, 4)
+    @warn "GAP uses a non-canonical order of simple roots for F4. Be prepared for unexpected results."
```
thofma: This warning should be removed:
- We do not do warnings in general.
- Why is it relevant what GAP does? This is OSCAR.

If there is some important information that the user should know about some choices that are different to something else (what exactly?), it should be noted in the docstring.

lgoettgens: When I computed some test-cases in different Cartan types in different CAS, I got different results for F4 when comparing GAP, LiE and Magma. After some research, I found out that GAP orders the simple roots differently than basically everyone else (and thus all highest weights look differently etc.). (And currently we call GAP underneath to construct the Lie algebra.) As this is the only function, that is called in all cases when one wants to compute with F4, I added the warning here. I don't think that people will look in the docstring of exactly this function, when they experience unexpected/weird behaviour later down the road when computing with a type F4 Lie algebra.
As already commented four lines above, all of this code extracting stuff from GAP is about to go in the near future (including the weirdness with F4 and the warning).

If, after my explanation of the situation, you still think that removing the warning (and maybe add it to the docstring instead) is the best solution, I am happy to change it.

thofma: It is in particular annoying for functions that trigger the warning indirectly. It is just unactionable noise, which in most cases is irrelevant.

What is the "near future"?

Also, "GAP" should be replaced by "OSCAR". Otherwise it is not clear what the relevance of GAP is or why one should care.

lgoettgens: "near future" = next months

OK, I'll then move this warning to the docstring and reformulate it in a way the message is clearer.

lgoettgens: After talking to @thofma on slack, we decided that throwing a "Not implemented" is the way forward here (for now at least).

lgoettgens:
@thofma Any other objections?

thofma (approved):

PR #3028, opened 2023-11-20
Fix `doctest_fix` for submodules

### LINKED ISSUE #2714: `doctest_fix` broken for submodules
**Describe the bug**
`doctest_fix` does not try to fix any doctests, if the file containing the doctest is loaded into a submodule.

**To Reproduce**
Change any doctest output in `experimental/LieAlgebras/src/LinearLieAlgebra.jl`.
```
using Oscar
Oscar.doctest_fix("LieAlgebras")
```
The above call should have reverted the before change, but it doesn't.

**Expected behavior**
`doctest_fix` working independently of submodules.

**System**
current master (ae19865edbb00b2554bdf2e1ed00c3b4c6a2321e)

lgoettgens:
Fixes https://github.com/oscar-system/Oscar.jl/issues/2714 by recursively iterating over all submodules.

### DIFF
--- a/src/utils/docs.jl
+++ b/src/utils/docs.jl
@@ -81,22 +81,38 @@ end
     doctest_fix(n::String; set_meta::Bool = false)
 
 Fixes all doctests for the file `n`, ie. all files in Oscar where
-  `n` occurs in the full pathname of.
+`n` occurs in the full pathname of.
 """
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
I'll trust this works as intended

PR #3032, opened 2023-11-21
Adv/refactor load

antonydellavecchia:
Major Refactor of load to allow the loading of large files. We add the JSON3 dependancy to allow for file streaming this way we do not load the entire file into a Dict. We also refactor load functions to not depend on a Vector or Dict argument being passed to allow for other types of serialization to also use these methods. 

The use of JSON3 was to avoid writting a JSON file streamer. The downside of JSON3 is that it is read only
meaning we cannot use it for updating files. The current upgrade code attempts to load the file to
a Dict and then do the upgrades. Seeing as we can now handle large files that no longer necessarily fit
in memory, we will run into issues when such files need an update.

There seems to be a bug when using JSON3 inter process communication, so when the IPC Serializer is used
the Deserializer state uses a Dict. We can get back to this in a later pull request when we perhaps have an in order
(de) serialization and remove the use of Dict then.

Also adds metadata reading.

The new load mechanisms include the following functions

load_node and load_array_node 
which could be renamed to load_data and load_data_array?

 some examples of how they are used are described in main.jl


test/Serialization/PolynomialsSeries.jl
```diff
@@ -106,7 +106,7 @@ cases = [
 
       # Tropical Semirings currently can't have formal power series
       filter!(case-> case[4] != "Tropical Semiring", cases)
-
+
```
antonydellavecchia, line 109:
```suggestion

```

antonydellavecchia:
@ThomasBreuer I've added changes to the groups code, but seems like it might need an update to gap itself?
```
ERROR: Error thrown by GAP: Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `DeserializeInOscar' on 3 arguments at /home/antonydv/.julia/artifacts/b5c2f0f824457e5c391fb24916f94d5d91c62c4f/share/gap/lib/methsel2.g:249 called from
<function "HANDLE_METHOD_NOT_FOUND">( <arguments> )
 called from read-eval loop at *defin*:0
```

ThomasBreuer:
@antonydellavecchia No, GAP itself need not be adjusted. The point is that GAP's method selection is used to decide which code shall be called. This is done by installing the functions in question as GAP methods for the two GAP operations `SerializeInOscar` and `DeserializeInOscar`, which are part of the interface to Oscar. Up to now, the methods for `DeserializeInOscar` needed four arguments, the last of which was the dictionary in question. With your changes, this dictionary is no longer needed, thus the GAP code dealing with the method installations has to be adjusted. Once I have a working version of your branch, I will tell you what is needed.

src/Serialization/GAP.jl
```diff
@@ -117,31 +121,42 @@ install_GAP_serialization(:IsFreeGroup,
...
+        fam = GAPWrap.ElementsFamily(GAPWrap.FamilyObj(F))
+        Ggens = [GAPWrap.ObjByExtRep(fam, GapObj(x, true)) for x in generators]
+        # Create the subgroup.
+        G = GAP.Globals.SubgroupNC(F, GapObj(Ggens))
       else
-        names = GapObj(d[:names], true)
-        G = GAP.Globals.FreeGroup(wfilt, names)
+        # Create a new full free group.
+        wfilt = getproperty(GAP.Globals, load_object(s, Symbol, :wfilt))
+        if :nameprefix keys(d)
```
ThomasBreuer: An `in` is missing here.
(I had thought that `haskey(d, :nameprefix)` is better, since one need not create `keys(d)`.)

antonydellavecchia: d is now a JSON3.Object and no longer has the haskey method. Thanks for spotting that, weird that I didn't get a revise error

src/Serialization/GAP.jl
```diff
@@ -117,31 +121,42 @@ install_GAP_serialization(:IsFreeGroup,
...
+            GapObj(nameprefix)
+          end
+          
+          init = load_node(s, :names) do names
+            GapObj(names, true)
+          end
+          G = GAP.Globals.FreeGroup(wfilt, GAP.Globals.infinity, prefix, init)
+        else
+          init = load_node(s, :names) do names
+            GapObj(names, true)
```
ThomasBreuer: Here we have a problem:
The object `names` constructed by `load_node` is now a `JSON3.Array` object, which is apparently not supported by the recursive conversion to GAP. For the moment, we can fix this by writing `GapObj([GapObj(x) for x in names])`.

src/Serialization/GAP.jl
```diff
@@ -117,31 +121,42 @@ install_GAP_serialization(:IsFreeGroup,
...
+          end
+          names = GapObj(init, true)
```
ThomasBreuer: Ah, and here `init` is already a GAP object, `GapObj` should not be called.

ThomasBreuer:
O.k., besides the workaround for the `GapObj` calls with first argument a `JSON3.Array` object,
the following changes are needed:
- In `gap/OscarInterface/gap/OscarInterface.gd`, replace the line
```
DeclareConstructor( "DeserializeInOscar", [ IsObject, IsObject, IsObject, IsObject ] );
```
by
```
DeclareConstructor( "DeserializeInOscar", [ IsObject, IsObject, IsObject ] );
```
- In `gap/OscarInterface/gap/OscarInterface.gi`, replace the line
```
             [ JuliaToGAP( IsString, entry[1] ), "IsObject", "IsObject", "IsObject" ],
```
by
```
             [ JuliaToGAP( IsString, entry[1] ), "IsObject", "IsObject" ],
```
- In `src/Serialization/GAP.jl`, rewrite also the code for `install_GAP_deserialization(:IsSubgroupFpGroup, ...)`

benlorenz (approved):

PR #3035, opened 2023-11-22
Precise Johnson solids using serialization

alexej-jordan:
This PR adds precise Johnson solids over fields not supported by `polymake`'s `QuadraticExtension`. As the construction uses the V-description, pre-computed matrices are loaded. This required additional serialization methods for `EmbeddedField`, `EmbeddedFieldElem` and underlying embedding types.


thofma:
Regarding rewriting

```julia
function save_object(s::SerializerState, E::FieldEmbeddingTypes)
```

I am bit worried about using `_absolute_index`. This relies on some ordering of zeros of roots of polynomials. This is consistent within one session, but we do not guarantee some persistent behaviour across sessions. (It will probably work now, but we don't guarantee that it will work in the future.)

Was the idea to have only one implementation for all four different types?

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -251,6 +315,11 @@ A Johnson solid is a 3-polytope whose facets are regular polygons, of various go
 It is proper if it is not an Archimedean solid.  Up to scaling there are exactly 92 proper Johnson solids.
 """
 function johnson_solid(index::Int)
+  if haskey(_johnson_names, index)
+    vertices = load(abspath(@__DIR__, "..", "..", "..", "data", "JohnsonMatrices", string("j", index, ".mat")))
```
antonydellavecchia:
```suggestion
    vertices = load(joinpath(oscardir, "data", "JohnsonMatrices", string("j", index, ".mat")))
```

antonydellavecchia (approved):
looks good, one small comment

alexej-jordan:
> Regarding rewriting
> 
> ```julia
> function save_object(s::SerializerState, E::FieldEmbeddingTypes)
> ```
> 
> I am bit worried about using `_absolute_index`. This relies on some ordering of zeros of roots of polynomials. This is consistent within one session, but we do not guarantee some persistent behaviour across sessions. (It will probably work now, but we don't guarantee that it will work in the future.)

That was what we were worried about, too, hence we need your insight.

> 
> Was the idea to have only one implementation for all four different types?

The idea was to at least have serialization working for all four types. The newly added types usually could not be constructed with a `complex_embedding` method because none is defined for the corresponding number field types.

thofma:
I will think about it.

thofma:
It appears I need to write a few lines of code to make this a bit easier. I will come back to you once this is done.

thofma:
I pushed the improved serialization with extensive tests. This needs https://github.com/thofma/Hecke.jl/pull/1296, which should be available in a few hours. I will restart CI once it is.

thofma:
I guess we need to recompute those `*.mat` files to work with the modified serialization?

alexej-jordan:
Thank you for helping out on this! I managed to successfully save and load the matrices. Unfortunately, there is an exception:
```
julia> Qx, x = QQ["x"]

julia> NF, qr8 = number_field(x^4 - 8, "a")
(Number field of degree 4 over QQ, a)

julia> ENF, qre8 = Hecke.embedded_field(NF, real_embeddings(NF)[2])
(Embedded field
Number field of degree 4 over QQ
at
Complex embedding corresponding to 1.68 of number field, a)

julia> emb = embedding(ENF)
Complex embedding corresponding to 1.68
  of number field with defining polynomial x^4 - 8
    over rational field

julia> save("emb.emb", emb)

julia> Oscar.reset_global_serializer_state()
Dict{Base.UUID, Any}()

julia> load("emb.emb")
┌ Warning: Attempting to load file stored using a DEV version with commit 7dbce862a671f43911ccffb1c4c5ada5738db1a0
└ @ Oscar ~/.julia/dev/Oscar/src/Serialization/main.jl:657
ERROR: Given approximation not close enough to a root. Possible roots are:
   -1.68 + i * 0.00, 1.68 + i * 0.00, 0.00 + i * 1.68, 0.00 + i * -1.68

Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] _find_nearest_complex_embedding(K::AnticNumberField, x::acb)
    @ Hecke ~/.julia/packages/Hecke/78SWd/src/NumField/ComplexEmbeddings/NfAbs.jl:272
  [3] complex_embedding(K::AnticNumberField, c::acb)
    @ Hecke ~/.julia/packages/Hecke/78SWd/src/NumField/ComplexEmbeddings/NfAbs.jl:283
  [4] load_object(s::Oscar.DeserializerState, #unused#::Type{Hecke.NumFieldEmbNfAbs}, dict::Dict{Symbol, Any})
    @ Oscar ~/.julia/dev/Oscar/src/Serialization/Fields.jl:455
  [5] load_typed_object(s::Oscar.DeserializerState, dict::Dict{Symbol, Any}; override_params::Nothing)
    @ Oscar ~/.julia/dev/Oscar/src/Serialization/main.jl:312
  [6] load_typed_object
    @ ~/.julia/dev/Oscar/src/Serialization/main.jl:298 [inlined]
  [7] load(io::IOStream; params::Nothing, type::Nothing, serializer_type::Type)
    @ Oscar ~/.julia/dev/Oscar/src/Serialization/main.jl:646
  [8] load
    @ ~/.julia/dev/Oscar/src/Serialization/main.jl:589 [inlined]
  [9] #5572
    @ ~/.julia/dev/Oscar/src/Serialization/main.jl:665 [inlined]
 [10] open(f::Oscar.var"#5572#5573"{Nothing, Nothing}, args::String; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Base ./io.jl:395
 [11] open
    @ ./io.jl:392 [inlined]
 [12] #load#5571
    @ ~/.julia/dev/Oscar/src/Serialization/main.jl:664 [inlined]
 [13] load(filename::String)
    @ Oscar ~/.julia/dev/Oscar/src/Serialization/main.jl:663
 [14] top-level scope
    @ REPL[9]:1
```
it seems that the following check is `true` for the two conjugate non-real embeddings, thus throwing an error:
https://github.com/thofma/Hecke.jl/blob/d05f266efbfb605fcd8bc1f7b718576b373a75e0/src/NumField/ComplexEmbeddings/NfAbs.jl#L262

thofma:
Thanks, I will have a look. I think I know what the problem is.

thofma:
Fix is on the way. I will ping you once you can try again.

thofma:
I pushed the fix. Please try again.

alexej-jordan:
Everything works now, thank you again! I pushed the new serialized data and hopefully CI tests will succeed now.

thofma (approved):

test/PolyhedralGeometry/polyhedron.jl
```diff
@@ -577,3 +577,15 @@
   end
 
 end
+
+@testset "Johnson solids" begin
+  
+  for i in [9, 10, 13, 16, 17, 18, 20, 21, 22, 23, 24, 25, 30, 32, 33, 34, 35, 36, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59, 60, 61, 64, 68, 69, 70, 71, 72, 73, 74, 75, 77, 78, 79, 82, 84, 85, 86, 87, 88, 89, 90, 92]
```
benlorenz:
```suggestion
  for i in keys(Oscar._johnson_names)
```

src/Serialization/Fields.jl
```diff
@@ -413,43 +426,85 @@ end
...
+      end
+    else
+      a = gens(K)
+      data = E.(a)
+      if any(all(overlaps(t[1], t[2]) for t in zip(data, e.(a))) for e in complex_embeddings(K) if e != E && restrict(E, k) == restrict(e, k))
+        error("Internal error in internal serialization.")
+      end
+      data = tuple(data...)
+    end
+    save_typed_object(s, data, :data)
```
antonydellavecchia, line 459: so I am pretty sure you should be able to use 
save_object here, since you should be able to figure out what the type of data is based on what K is.

Then you can do something like

```
if is_simple(K)
  load_object(s, NotTuple, dict[:data]) ...
else
  load_object(s, Tuple, dict[:data], tuple_params)
end
```
for the load, might seem a bit more complicated, but it will cut down the size of the files, seeing as there are ~50? files, it could make a big difference.

The reason I bring this up now is to avoid writing an update script later to shrink file size.

alexej-jordan: This has proven to be a bit more complex than expected - the `acb` in the `data` field needs an instantiated `AcbField` when loading this way. So the trade-off would be not as big and thus not rewarding enough for the changes needed.

benlorenz (approved):

PR #3041, opened 2023-11-23
Some improvements on the modules

HechtiDerLachs:
@jankoboehm : I wanted to look a bit into the morphisms of modules for the reasons we talked about today. But before I could actually start doing that, I needed to split that gigantic file into more digestable parts. For me personally this was so that syntax highlighting would not crash all the time. But also in general, I think, it makes sense to split this into thematically ordered smaller chunks. 

Do you agree?

Edit: splitting now takes place in #3042 . This PR will eventually be rebased. Until then check out the single last commit to see the changes for the improvements only.


HechtiDerLachs:
I tried to improve the code a bit, especially regarding the speed of evaluation of mappings. So far running the same test-suite for modules gives 

before:
```
 45.207242 seconds (433.82 M allocations: 20.793 GiB, 12.69% gc time, 17.17% compilation time)
 45.368074 seconds (433.65 M allocations: 20.792 GiB, 13.04% gc time, 17.48% compilation time)
```
after
```
 37.961641 seconds (265.70 M allocations: 12.739 GiB, 10.21% gc time, 21.95% compilation time)
 38.276709 seconds (266.08 M allocations: 12.747 GiB, 7.61% gc time, 22.06% compilation time)
 37.249246 seconds (265.81 M allocations: 12.744 GiB, 7.62% gc time, 21.19% compilation time)
```
So especially in terms of memory consumption, I think this is useful and should be merged one way or the other. 

Edit: The main upshots were:
 1. `SubquoModuleElem`s can now store either their `coordinates` **or** their representative. Thus, access via the fields `a.coeffs` and `a.repres` is now forbidden and the getters `coordinates(a)` and `repres(a)` should be used instead. 
 2. Addition and multiplication of `SubQuoElem`s uses whatever of the two is there, but doesn't fill out the other side all the time. 
 3. `SubquoElem`s now have a field `is_reduced` similar to the elements in quotient rings.
 4. Maps store a list of the images of the generators and this can be called via `images_of_generators(f)` (and not through creation of the inclusion of the image and accessing the generators of that submodule as before...)
 5. Evaluation of maps use the images of generators and `coordinates` of the element to be mapped.

thofma:
Why not separate the code reorganization from the improvements PR wise? The spitting should definitely be happening.

HechtiDerLachs:
I ran some more tests locally. Compared to current master I now get the following timings:

before:
```
46.521581 seconds (434.29 M allocations: 20.809 GiB, 12.73% gc time, 16.09% compilation time)
 47.341869 seconds (433.96 M allocations: 20.798 GiB, 9.42% gc time, 16.86% compilation time)
```
after:
```
 39.683286 seconds (273.27 M allocations: 12.962 GiB, 10.36% gc time, 21.05% compilation time)
 38.410556 seconds (273.16 M allocations: 12.958 GiB, 7.44% gc time, 21.16% compilation time)
 39.105477 seconds (273.33 M allocations: 12.962 GiB, 7.44% gc time, 20.39% compilation time)
```
The difference to the above runs is due to some tests being disabled (on both sides) before. These runs here really take the whole test suite for the modules.

HechtiDerLachs:
@jankoboehm : I double checked again. There is no internal constructor for neither `FreeModuleHom` nor `SubQuoHom` which does not take the images of the generators as input. So computing and caching them only on request has never been an option. 

Edit: Besides: Now that the `SubquoModuleElem`s are allowed to also store the information on their coordinates only, this should be just as good as what you had in mind: For every image of a generator store the coordinates in the generators of the codomain. Just make sure that the list of images you are passing to the constructor do not contain the fields on their representatives.

HechtiDerLachs:
Timings now go down to
```
 38.644607 seconds (268.27 M allocations: 12.779 GiB, 7.30% gc time, 20.29% compilation time)
 38.324128 seconds (268.55 M allocations: 12.780 GiB, 7.56% gc time, 21.69% compilation time)
```
No major improvement anymore. But still something. I think I can stop here. 

Could maybe someone confirm the changes with the hashing?

HechtiDerLachs:
Can this be merged, then?

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -216,14 +216,26 @@ function (==)(a::AbstractFreeModElem, b::AbstractFreeModElem)
...
+function hash(a::AbstractFreeModElem{<:MPolyElem}, h::UInt)
   b = 0xaa2ba4a32dd0b431 % UInt
   h = hash(typeof(a), h)
   h = hash(parent(a), h)
   h = hash(coordinates(a), h)
   return xor(h, b)
 end
 
+function hash(a::AbstractFreeModElem{<:MPolyQuoRingElem}, h::UInt)
+  simplify!(a)
```
fingolfin: We just discussed how this call resp. this `hash` method is somewhat dangerous:
- it assumes that `simplify!(a)` modifies `a` in-place and also "remembers" that it is simplified (so it is fast "amortized", i.e. you only compute a GB for it once)
- it must produce a unique normal form -- but we don't really have a specification for `simplify!`. We really need one otherwise we are building on quicksand
- perhaps it would be better to just do `error("not supported")` here and handle using these in dicts differently??
- however the old `hash(a::AbstractFreeModElem, h::UInt)` method just was wrong...

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -216,14 +216,26 @@ function (==)(a::AbstractFreeModElem, b::AbstractFreeModElem)
...
+  b = 0xaa2ba4a32dd0b431 % UInt
+  h = hash(typeof(a), h)
+  h = hash(parent(a), h)
+  h = hash(coordinates(a), h)
+  return xor(h, b)
+end
+
+simplify(a::AbstractFreeModElem{<:MPolyQuoRingElem}) = return parent(a)(map_entries(simplify, coordinates(a)))
```
fingolfin:
```suggestion
simplify(a::AbstractFreeModElem{<:MPolyQuoRingElem}) = parent(a)(map_entries(simplify, coordinates(a)))
```

src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -553,9 +559,9 @@ Homogeneous module homomorphism)
 ```
 """
 function image(h::FreeModuleHom)
-  si = [x for x = map(h, basis(domain(h))) if !iszero(x)]
+  si = filter(x->!iszero(x), images_of_generators(h))
```
fingolfin:
```suggestion
  si = filter(!iszero, images_of_generators(h))
```

fingolfin (approved):
While there issues with the `hash` method, the existing method is just as broken, so this doesn't really make it worse... We could merge this PR, and open an issue to remind us to resolve the `hash` function (and also to specify `simplify!`)

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -216,14 +216,30 @@ function (==)(a::AbstractFreeModElem, b::AbstractFreeModElem)
   return a.coords == b.coords
 end
 
-function hash(a::AbstractFreeModElem, h::UInt)
+# A special method for the case where we can safely assume 
+# that the coordinates of elements allow hashing.
+function hash(a::AbstractFreeModElem{<:MPolyElem{<:FieldElem}}, h::UInt)
```
fingolfin, line 228: Note that a field element does not necessarily have correct hashing implemented either -- though as you correctly just pointed out to me when we talked, such fields probably won't be supported by Singular either.

But on the long run we probably want a bunch of type traits to model whether hashing, normal forms, etc. are supported:
```julia
can_hash(::Type) = false
can_compute_normal_form(::Type) = false
can_hash(::Type{<:MPolyElem{T}}) where T = can_hash(T)
can_compute_normal_form(::Type{<:MPolyElem{T}}) where T = can_compute_normal_form(T)
```
etc.

(Perhaps should turn this into an issue?)

src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -553,9 +559,11 @@ Homogeneous module homomorphism)
 ```
 """
 function image(h::FreeModuleHom)
-  si = [x for x = map(h, basis(domain(h))) if !iszero(x)]
+  si = filter(!iszero, copy(images_of_generators(h)))
```
fingolfin: Why the `copy`? The `filter` function returns a new list anyway?
```suggestion
  si = filter(!iszero, images_of_generators(h))
```

HechtiDerLachs:
Setting the generic hash function to error was not an option in the end. Even the `unique` function uses hashes and I think we can not pull all of such uses out of the code for the moment.

PR #3042, opened 2023-11-23
Module migration

HechtiDerLachs:
Takes care of the splitting of UngradedModules.jl as in #3041 .  Should this be merged, we can rebase the other.


thofma:
Thanks. I tried to look at this file recently and I would love to see it split up!

jankoboehm:
Thanks a lot, we will do still a few adjustments and then we can got ahead.

ederc:
Thank you @HechtiDerLachs. We should also think about renaming the files resp. folder (maybe not in this PR) since there are many calls to `is_graded()`.

HechtiDerLachs:
For me personally it would be best if any changes to this pr could be kept to a minimum for the moment, since I will have to rebase #3041 on this. Thus I would be in favor to postpone renamings and the like.

jankoboehm (approved):

PR #3044, opened 2023-11-24
Adjust invariant theory code to allow polynomial rings with ordering different from `:degrevlex`

joschmitt:
This removes the requirement that the underlying polynomial ring of an invariant ring needs to be created with `ordering = :degrevlex` which was always a hack.

I also adjusted the various `reduce*` functions, so that they can handle an empty array of divisors.

Related: Is there any chance to rename/alias/whatever `reduce` to `rem` and `reduce_with_quotients` to `divrem`? I understand that this is not that easy because functions under these names exist, but I also don't think it is impossible. And it would improve the discoverability of `reduce` and friends significantly in my opinion.

EDIT: The comment is mostly obsolete.

### DIFF
--- a/src/InvariantTheory/fundamental_invariants.jl
+++ b/src/InvariantTheory/fundamental_invariants.jl
@@ -84,7 +84,9 @@ function fundamental_invariants_via_king(RG::InvRing, beta::Int = 0)
         continue
       end
 
-      push!(S, inv(leading_coefficient(f))*f)
+      # Cancelling the leading coefficient is not mathematically necessary and
+      # should be done with the ordering that is used for the printing
+      push!(S, inv(AbstractAlgebra.leading_coefficient(f))*f)
       push!(GO, g)
     end
 
--- a/src/InvariantTheory/fundamental_invariants.jl
+++ b/src/InvariantTheory/fundamental_invariants.jl
@@ -23,10 +23,8 @@ end
 function fundamental_invariants_via_king(RG::InvRing, beta::Int = 0)
   @assert !is_modular(RG)
 
-  Rgraded = polynomial_ring(RG)
+  Rgraded = _internal_polynomial_ring(RG)
   R = forget_grading(Rgraded)
-  # R needs to have the correct ordering for application of divrem
-  @assert ordering(R) == :degrevlex
   ordR = degrevlex(gens(R))
 
   S = elem_type(R)[]
@@ -74,7 +72,7 @@ function fundamental_invariants_via_king(RG::InvRing, beta::Int = 0)
     end
 
     for m in mons
-      f = forget_grading(reynolds_operator(RG, Rgraded(m)))
+      f = forget_grading(_cast_in_internal_poly_ring(RG, reynolds_operator(RG, _cast_in_external_poly_ring(RG, Rgraded(m)))))
       if is_zero(f)
         continue
       end
@@ -84,9 +82,7 @@ function fundamental_invariants_via_king(RG::InvRing, beta::Int = 0)
         continue
       end
 
-      # Cancelling the leading coefficient is not mathematically necessary and
-      # should be done with the ordering that is used for the printing
-      push!(S, inv(AbstractAlgebra.leading_coefficient(f))*f)
+      push!(S, f)
       push!(GO, g)
     end
 
@@ -94,7 +90,10 @@ function fundamental_invariants_via_king(RG::InvRing, beta::Int = 0)
   end
 
   invars_cache = FundamentalInvarsCache{elem_type(Rgraded), typeof(Rgraded)}()
-  invars_cache.invars = [ Rgraded(f) for f in S ]
+  polys_ext = [ _cast_in_external_poly_ring(RG, Rgraded(f)) for f in S ]
+  # Cancelling the leading coefficient is not mathematically necessary and
+  # should be done with the ordering that is used for the printing
+  invars_cache.invars = [ inv(AbstractAlgebra.leading_coefficient(f))*f for f in polys_ext ]
   invars_cache.via_primary_and_secondary = false
   invars_cache.S = graded_polynomial_ring(coefficient_ring(R), [ "y$i" for i = 1:length(S) ], [ total_degree(f) for f in S ])[1]
   return invars_cache
--- a/src/InvariantTheory/invariant_rings.jl
+++ b/src/InvariantTheory/invariant_rings.jl
@@ -1,4 +1,3 @@
-
 ################################################################################
 #
 #  Field access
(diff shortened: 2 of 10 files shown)


thofma:
Have you measured performance impact of not using the fast, native, `divrem`, but the `reduce` function, which creates bazillion unnecessary objects?

joschmitt:
> Have you measured performance impact of not using the fast, native, `divrem`, but the `reduce` function, which creates bazillion unnecessary objects?

Yes, it is slower now. But I don't see a way around this as long as `divrem` does not accept an ordering. Before #2845 there wasn't really a problem because the polynomial ring would only be created internally, but now this is different. So, for example this would fail:
```
julia> K, a = quadratic_field(2);

julia> R, (x, y) = graded_polynomial_ring(K, ["x", "y"]);

julia> M = matrix(K, [1 1; -1 1])*(1//a);

julia> N = matrix(K, [0 1; 1 0]);

julia> G = matrix_group(M, N);

julia> IR = invariant_ring(R, G);

julia> primary_invariants(IR);

julia> secondary_invariants(IR)
ERROR: AssertionError: ordering(R) == :degrevlex
Stacktrace:
 [1] secondary_invariants_nonmodular(RG::Oscar.InvRing{AnticNumberField, MatrixGroup{nf_elem, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}}, MPolyDecRingElem{nf_elem, AbstractAlgebra.Generic.MPoly{nf_elem}}, MPolyDecRing{nf_elem, AbstractAlgebra.Generic.MPolyRing{nf_elem}}, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}})
   @ Oscar ~/.julia/dev/Oscar/src/InvariantTheory/secondary_invariants.jl:166
 [2] _secondary_invariants
   @ ~/.julia/dev/Oscar/src/InvariantTheory/secondary_invariants.jl:283 [inlined]
 [3] secondary_invariants(IR::Oscar.InvRing{AnticNumberField, MatrixGroup{nf_elem, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}}, MPolyDecRingElem{nf_elem, AbstractAlgebra.Generic.MPoly{nf_elem}}, MPolyDecRing{nf_elem, AbstractAlgebra.Generic.MPolyRing{nf_elem}}, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}})
   @ Oscar ~/.julia/dev/Oscar/src/InvariantTheory/secondary_invariants.jl:323
 [4] top-level scope
   @ REPL[10]:1

julia> fundamental_invariants(IR)
ERROR: AssertionError: ordering(R) == :degrevlex
Stacktrace:
 [1] fundamental_invariants_via_king(RG::Oscar.InvRing{AnticNumberField, MatrixGroup{nf_elem, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}}, MPolyDecRingElem{nf_elem, AbstractAlgebra.Generic.MPoly{nf_elem}}, MPolyDecRing{nf_elem, AbstractAlgebra.Generic.MPolyRing{nf_elem}}, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}}, beta::Int64)
   @ Oscar ~/.julia/dev/Oscar/src/InvariantTheory/fundamental_invariants.jl:29
 [2] #fundamental_invariants#5258
   @ ~/.julia/dev/Oscar/src/InvariantTheory/fundamental_invariants.jl:226 [inlined]
 [3] fundamental_invariants
   @ ~/.julia/dev/Oscar/src/InvariantTheory/fundamental_invariants.jl:219 [inlined]
 [4] fundamental_invariants(IR::Oscar.InvRing{AnticNumberField, MatrixGroup{nf_elem, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}}, MPolyDecRingElem{nf_elem, AbstractAlgebra.Generic.MPoly{nf_elem}}, MPolyDecRing{nf_elem, AbstractAlgebra.Generic.MPolyRing{nf_elem}}, AbstractAlgebra.Generic.MatSpaceElem{nf_elem}})
   @ Oscar ~/.julia/dev/Oscar/src/InvariantTheory/fundamental_invariants.jl:219
 [5] top-level scope
   @ REPL[11]:1
```
I don't have a better solution to solve this problem. (Actually, already the creation of the invariant ring should have triggered an assertion, that's what I fixed with the first commit here.)

src/InvariantTheory/fundamental_invariants.jl
```diff
@@ -79,12 +77,12 @@ function fundamental_invariants_via_king(RG::InvRing, beta::Int = 0)
...
       end
 
-      _, g = divrem(f, GO)
+      g = reduce(f, GO, ordering = ordR)
       if is_zero(g)
         continue
       end
 
-      push!(S, inv(leading_coefficient(f))*f)
+      push!(S, inv(AbstractAlgebra.leading_coefficient(f))*f)
```
thofma: Isn't this incompatible with the change to use `reduce`?

joschmitt: Canceling the leading coefficient is mostly eye candy so that the polynomials look nicer, but there is no mathematical necessity. It should be done with the ordering that is used for the printing.

joschmitt:
> > Have you measured performance impact of not using the fast, native, `divrem`, but the `reduce` function, which creates bazillion unnecessary objects?
> 
> Yes, it is slower now. [...]
> I don't have a better solution to solve this problem. [...]

Another solution would be to use a polynomial ring with `:degrevlex` internally, so that `divrem` works, and then only coerce the results in the polynomial ring supplied by the user if necessary (which is of course also not for free in terms of runtime, but should be fine). I don't really like this solution.

Would it be possible to extend the "generic" `divrem` code in AbstractAlgebra to work with an arbitrary monomial ordering (it don't expect it to be as fast as in the "native" ordering)? I just had a quick look and it looks pretty arcane what's implemented there.

thofma:
It seems that as a first step one already forgets the grading. So switching to a faster internal polynomial ring should not make much of a difference? One has to change the ring anyway.

How much is the runtime effected by this change here?

src/InvariantTheory/fundamental_invariants.jl
thofma: Makes sense. Could you add a comment explaining this?

joschmitt:
> How much is the runtime effected by this change here?

It's hard to make a general statement, but I have one example ($S_3\times S_3\times S_3\times S_3$) where the computation of the secondary invariants is 24 times slower with `reduce` than with `divrem`. 😞
Let's try the other approach...

src/InvariantTheory/invariant_rings.jl
```diff
@@ -15,6 +14,46 @@ group(I::InvRing) = I.group
...
+end
+
+# Return f as an element of R. Assumes that ngens(R) == ngens(parent(f))
+# and coefficient_ring(R) === coefficient_ring(parent(f)). This is not checked.
+function __cast_forced(R::MPolyRing{T}, f::MPolyRingElem{T}) where T
+  F = MPolyBuildCtx(R)
+  for (c, e) in zip(AbstractAlgebra.coefficients(f), AbstractAlgebra.exponent_vectors(f))
+    push_term!(F, c, e)
+  end
+  return finish(F)
```
joschmitt, line 31: Is there a faster way to cast a polynomial in another polynomial ring that is "the same" apart from the ordering?

joschmitt:
I switched back to `divrem`. Now we need to handle an additional "internal" polynomial ring with `:degrevlex` if the user supplies one with a different ordering.
This creates a bit of overhead for moving polynomials between the rings, but is think it is fine.

thofma (approved):

src/InvariantTheory/invariant_rings.jl
thofma: No

PR #3045, opened 2023-11-24
Allow optional parent argument for homogenization.

HechtiDerLachs:
CC: @wdecker

### DIFF
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1603,9 +1603,12 @@ function _homogenization(f::MPolyRingElem, S::MPolyDecRing, start_pos::Int)
    return finish(F)
 end
 
-function _homogenization(f::MPolyRingElem, W::ZZMatrix, var::VarName, pos::Int)
+function _homogenization(
+    f::MPolyRingElem, W::ZZMatrix, var::VarName, pos::Int;
+    parent::Union{Nothing, <:MPolyDecRing}=nothing
+  )
     # ASSUME pos is in 1:1+ngens(parent(f))
-  R = parent(f)
+  R = Oscar.parent(f)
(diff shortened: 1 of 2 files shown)


src/Rings/mpoly-graded.jl
```diff
@@ -1603,9 +1603,12 @@ function _homogenization(f::MPolyRingElem, S::MPolyDecRing, start_pos::Int)
...
+function _homogenization(
+    f::MPolyRingElem, W::ZZMatrix, var::VarName, pos::Int;
+    parent::Union{Nothing, <:MPolyDecRing}=nothing
+  )
     # ASSUME pos is in 1:1+ngens(parent(f))
-  R = parent(f)
+  R = Oscar.parent(f)
   A = copy(symbols(R))
   l = length(A)
   @req pos in 1:l+1 "Index out of range."
```
fingolfin, line 1614: The code below always constructs `A` and `WH`, but they are not used when `parent !== nothing`. We should avoid that.

Note that `var` is not even used in this case. And `pos` is only passed to `_homogenization`.

Perhaps there should then be checked that e.g. `nvars(S) = nvars(R) + 1`... Of course on could then also verify that e.g. weights match, variable names etc. -- or one could deliberately ignore this...


The "clean" alternative would be to have a morphism from `parent(R)` to `S`...

JohnAAbbott:
I believe this is now no longer relevant after the change to the new approach using `Homogenizer` objects.
I suggest **closing** this PR.

fieker:
On Wed, Feb 21, 2024 at 03:22:36AM -0800, JohnAAbbott wrote:
> I believe this is now no longer relevant after the change to the new approach using `Homogenizer` objects.
> I suggest **closing** this PR.
> 
please close it

PR #3048, opened 2023-11-24
Fix several things around root systems

lgoettgens:
While writing a jupyter notebook for an exercise class, I found several things that weren't working or just weren't there. This is the first batch of fixes and additions from that.

### DIFF
--- a/experimental/LieAlgebras/src/RootSystem.jl
+++ b/experimental/LieAlgebras/src/RootSystem.jl
@@ -175,6 +175,15 @@ function roots(R::RootSystem)
   return [[r for r in positive_roots(R)]; [-r for r in positive_roots(R)]]
 end
 
+function simple_root(R::RootSystem, i::Int)
+  @req 1 <= i <= rank(R) "Invalid index"
+  return positive_root(R, i)
+end
+
+function simple_roots(R::RootSystem)
+  return [positive_root(R, i) for i in 1:num_simple_roots(R)]
+end
+
 @doc raw"""
     weyl_group(R::RootSystem) -> WeylGroup
 
--- a/experimental/LieAlgebras/src/RootSystem.jl
+++ b/experimental/LieAlgebras/src/RootSystem.jl
@@ -300,6 +300,15 @@ function is_negative_root_with_index(r::RootSpaceElem)
   end
 end
 
+function is_simple_root_with_index(r::RootSpaceElem)
+  i = findfirst(==(r), simple_roots(r.root_system))
+  if isnothing(i)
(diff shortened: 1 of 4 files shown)


simonbrandhorst (approved):

lgoettgens:
@felix-roehrich please review the changes here, even though it has already been merged. I can react to your comments in a future PR.

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -164,6 +175,15 @@ function roots(R::RootSystem)
   return [[r for r in positive_roots(R)]; [-r for r in positive_roots(R)]]
 end
 
+function simple_root(R::RootSystem, i::Int)
+  @req 1 <= i <= rank(R) "Invalid index"
+  return positive_root(R, i)
+end
+
+function simple_roots(R::RootSystem)
+  return [positive_root(R, i) for i in 1:num_simple_roots(R)]
```
felix-roehrich, line 184:
```suggestion
  return positive_roots(R)[1:rank(R)]
```

experimental/LieAlgebras/src/WeylGroup.jl
```diff
@@ -141,6 +141,24 @@ function Base.:(*)(x::WeylGroupElem, w::WeightLatticeElem)
...
+  end
+
+  word = deepcopy(x.word)
+  for _ in 2:n
+    for s in Iterators.reverse(x.word)
+      _lmul!(x.parent.refl, word, s)
+    end
+  end
+
+  return WeylGroupElem(x.parent, word)
```
felix-roehrich, line 159:
```suggestion
  px = deepcopy(x)
  for _ in 2:n
    for s in Iterators.reverse(x.word)
      lmul!(px, Int(s)) # make lmul! more general to make casting unnecessary
    end
  end

  return px
```

experimental/LieAlgebras/src/RootSystem.jl
lgoettgens: The current solution is more efficient than your suggestion. Your suggestion first creates a vector of all positive roots and then copies the first `rk` many to a new vector. As there are many more positive than simple roots, this is mostly useless work.

felix-roehrich: No, the implementation is just an accessor with a type assertion. The suggestion instead prevents creating a new underlying array each time.

lgoettgens: Oh yeah, you are correct.

PR #3050, opened 2023-11-26
Address comments from #3048

lgoettgens:
https://github.com/oscar-system/Oscar.jl/pull/3048 has been merged before the review by @felix-roehrich took place. Here are his suggestions.

### DIFF
--- a/experimental/LieAlgebras/src/RootSystem.jl
+++ b/experimental/LieAlgebras/src/RootSystem.jl
@@ -181,7 +181,7 @@ function simple_root(R::RootSystem, i::Int)
 end
 
 function simple_roots(R::RootSystem)
-  return [positive_root(R, i) for i in 1:num_simple_roots(R)]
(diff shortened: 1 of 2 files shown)


felix-roehrich:
Can please update `lmul` and `lmul!` while you are at it ?
This should do the trick:
```julia
function lmul(x::WeylGroupElem, i::Integer)
  return lmul!(deepcopy(x), i)
end

function lmul!(x::WeylGroupElem, i::Integer)
  @req 1 <= i <= rank(root_system(parent(x))) "Invalid generator"
  _lmul!(parent(x).refl, word(x), UInt8(i))
  return x
end
```

Then you can remove the comment in `^`.

felix-roehrich (approved):
Thanks.

lgoettgens:
@simonbrandhorst As you already merged https://github.com/oscar-system/Oscar.jl/issues/3048, could you merge this as well?

thofma (approved):

PR #3052, opened 2023-11-28

HereAround:
* Add model sections to all literature model files.
* Reorganize that data in the literature model json files, so that constructions over concrete bases can be simplified.
* Streamlined the corresponding constructor for literature Tate models over concrete bases.
* Add warning that constructing a literature Tate model over a concrete base may lead to unwanted enhancements.


apturner (approved):
This looks good @HereAround, thank you!

PR #3057, opened 2023-11-29
remove an import to suppress a warning

simonbrandhorst:
@fieker @micjoswig  
Hopefully this will fix your warning.


thofma:
@lgoettgens was faster than you, and he also kept `is_effective` in #3056. Which one should we take?

lgoettgens:
is_effective has methods both in and outside the PlaneCurve submodule. So we should keep it

simonbrandhorst:
Alright. I close this one then.

micjoswig:
Thanks.  Works now.

PR #3058, opened 2023-11-30
adds local cohomology method for computing sheaf cohomology

RafaelDavidMohr:
@wdecker I tried to change the documentation accordingly, please have a look.

### DIFF
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -1772,7 +1772,7 @@ function _weights_and_sing_mod(M::ModuleFP{T}) where {T <: MPolyDecRingElem}
     cokern_gens = [zero(ambient_free_module(cokern_repr))]
   end
   sing_mod = singular_generators(ModuleGens(cokern_gens))
-  weights = [Int(d[1]) for d in degrees_of_generators(free_mod)]
+  weights = [Int(d[1]) for d in degrees_of_generators(p[0])]
   return sing_mod, weights
 end
   
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -1533,6 +1533,10 @@ The values of the returned table can be accessed by indexing it
 with a cohomological index and a value between `l` and `h` as shown
 in the example below.
 
+The keyword `algorithm` can be set to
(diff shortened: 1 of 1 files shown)


ederc (approved):

thofma:
@HechtiDerLachs you had problems when trying out this PR (reported on slack). Did they get solved?

src/Modules/ModulesGraded.jl
```diff
@@ -1694,6 +1759,23 @@ function _ndigits(val::Int)
...
+
+  # get a cokernel presentation of M
+  p = presentation(M)
+  cokern_repr = image(map(p, 1))[1]
+  cokern_gens = ambient_representatives_generators(cokern_repr)
+  if isempty(cokern_gens)
+    cokern_gens = [zero(ambient_free_module(cokern_repr))]
+  end
+  sing_mod = singular_generators(ModuleGens(cokern_gens))
+  weights = [Int(d[1]) for d in degrees_of_generators(free_mod)]
```
HechtiDerLachs:
```suggestion
  weights = [Int(d[1]) for d in degrees_of_generators(p[0])]
```
This fixed the issue for me. The input module is not so easy to reproduce, because it uses stuff which is not yet merged. But I think, it should be easy to confirm that this change is the correct one to do, either way.

RafaelDavidMohr: It's a bit strange that this fixes the issue, I pushed your suggested change for now.

HechtiDerLachs: Actually, I think it's pretty straightforward, why this solves the issue. It has to do with the 'Subquotient' concept in Oscar which does not have a counterpart in Singular. The translation goes via `presentation` so that the given module `M` is identified with a quotient module `Q` of a free module `F`. But `F` need not be the `ambient_free_module` of the original `M`. In fact, in my example even the number of generators differed! In the fix we now use `F` instead of `ambient_free_module(M)` which should be the correct thing to do. 

Does that make sense? 

I just saw this when @thofma asked again yesterday and I took a closer look.

RafaelDavidMohr: That makes sense, yeah. Thanks for taking a look! I thought there might be something wrong with `ambient_free_module`.

Issue #3059, opened 2023-11-30
Streamline methods for `direct_sum` and `tensor_product` throughout Oscar 


HechtiDerLachs:
As discussed yesterday with @fingolfin : At the moment it is not possible to write generic code which uses functions like `direct_sum`, `tensor_product`, or even `hom`. The reason is that, depending on the type of the objects considered, `direct_sum` will sometimes just return the object and in other cases a triple `(S, inj, pr)` consisting of the object `S`, the injections of the summands `inj`, and the projections `pr`. Currently we find for `direct_sum`:

 * `ModuleFP` -> only the object
 * `FPModule` -> object and maps
 * Lie Algebras -> ? 
 * Abelian groups -> ?

Similar things happen for `tensor_product` which is supposed to return a pair `(P, mult_map)` consisting of the object and a multiplication map. Here we find 

 * `ModuleFP` -> only the object
 * `FPModule` -> only the object
 * LieAlgebras -> ?
 * Abelian groups -> ?
 * polynomial rings -> ?

I came across this again in the work on hyper complexes #3021 and @fingolfin criticized my workarounds for not being too sustainable and no issue being raised from my side. Hence here comes the issue to keep track of such developments.

thofma:
We (@simonbrandhorst and @StevellM) had similar discussions for lattices, where also direct sum, direct products and "both" exist. In the end, we did the (in my opinion) right thing by disentangling and using proper mathematical language (https://github.com/thofma/Hecke.jl/pull/985). The summary is

> [...] The general idea is:
>
> Those objects have finite direct sums and finite direct products which agree, and which are called biproducts. We leave as an interface to the user the following choices:
> - construct the biproduct as a `direct_sum` and only get as extra output the injection maps;
> - construct the biproduct as a `direct_product` and only get as extra output the projection maps;
> - construct the biproduct as a `biproduct` and return the injection maps as well as the projection maps.

P.S.: I am not a fan of `:task`, as you might have guessed.

lgoettgens:
I strongly dislike `:task` as it makes the code intrinsically type unstable. Thus, I would like a consistent way throughout Oscar.

`LieAlgebra`s do not have any sums or products yet. `LieAlgebraModule` returns only the object in both cases.

jankoboehm:
I agree with @thofma, @simonbrandhorst and @StevellM in what we should get mathematically.
In fact for modules (ModuleFP) this is the case, we 
[...]

PR #3060, opened 2023-11-30
Resolve some issues of unbound type parameters

lgoettgens:
Once all of them are resolved, we can enable the Aqua check `unbound_args`.

Most fixes were just minor reordering of a method header, such that the type parameter no longer gets instantiated. In some cases with varargs, it seemed more reasonable to split the method into two (one with 0 args, and one with at least 1 arg).


lgoettgens:
This is done for now. All remaining instances have to do with `direct_sum` and `tensor_product`, and I do not want to interfere with #3059.

src/InvariantTheory/invariant_rings.jl
```diff
@@ -63,10 +63,12 @@ function invariant_ring(R::MPolyDecRing, M::Vector{<: MatrixElem})
...
 
-invariant_ring(matrices::MatrixElem{T}...) where {T} = invariant_ring(collect(matrices))
+function invariant_ring(m::MatrixElem{T}, ms::MatrixElem{T}...) where {T} 
+  return invariant_ring([m, ms...])
+end
 
-function invariant_ring(R::MPolyDecRing, matrices::MatrixElem{T}...) where {T}
-  return invariant_ring(R, collect(matrices))
+function invariant_ring(R::MPolyDecRing, m::MatrixElem{T}, ms::MatrixElem{T}...) where {T} 
+  return invariant_ring(R, [m, ms...])
```
joschmitt, line 71: Dumb question: is this faster/better/prettier than what was there before?
Looking at this signature, I would now expect that `m` and `ms` have different meanings in this function and not that they are just put together in the same vector.
(If it is only about the `T` we could do
```suggestion
function invariant_ring(R::MPolyDecRing, matrices::MatrixElem...)
  return invariant_ring(R, collect(matrices))
```
as the `T` is not used... If somebody puts in matrices of different types, this will error if the entries cannot be cast in the coefficient ring of the polynomial ring.)

lgoettgens: This is about the existence of a dispatch where the T is not instantiated.

Your suggestion would be (mostly) fine as well. The single drawback is, that the type of `collect(matrices)` gets inferred as `Vector{Any}` for the 0-arg case.

joschmitt: (I think we should actually throw a meaningful error if there are no matrices, but that's off-topic for now.)
I can't say I can see the ramification of this function not having a completely well-defined return type. I find the new signature confusing (for the human); if it is less confusing for the computer, that's fine by me.

benlorenz:
The documentation build is still failing, probably one of the signatures there needs to be updated:
```
┌ Error: no docs found for 'intersect(V::T...) where T<:GAPGroup' in `@docs` block in src/Groups/subgroups.md:30-41
```

benlorenz (approved):

PR #3061, opened 2023-12-01
add algorithm keyword to isometry_group

simonbrandhorst:
@StevellM


simonbrandhorst:
@StevellM  will take it from here.

PR #3063, opened 2023-12-01
Cleanup Schemes

simonbrandhorst:
This moves some helpers out of experimental and deletes two files.


afkafkafk13 (approved):
Looks good.

This is basically a cleanup of experimental/Schemes/SpecialTypes.jl, in which only some auxilliary routines were left, which should go to the algebraic side anyway.

PR #3065, opened 2023-12-01
Make `_isometry_group_via_decomposition` optional

StevellM:
Add a keyword argument to choose how to compute isometry group for `ZZLat` and change a bit the code of `_isometry_group_via_decomposition` to handle short vectors differently.

Should replace #3061


src/Groups/matrices/form_group.jl
```diff
@@ -788,38 +805,38 @@ function _isometry_group_via_decomposition(L::ZZLat; closed = true, direct=true)
...
+  return lattice(V, basis_matrix(L) * matrix(g); check=false)
 end
 
 """
     on_matrix(x, g::MatrixGroupElem{QQFieldElem,QQMatrix})
 
 Return `x*g`.
 """
-function on_matrix(x, g::MatrixGroupElem{QQFieldElem,QQMatrix})
+function on_vector(x, g::MatrixGroupElem{QQFieldElem,QQMatrix})
```
simonbrandhorst: this does not match with the docstring anymore.

simonbrandhorst: restrict the signature to vectors?

StevellM: Yes indeed, I will restrict the signature and change the docstring. Thanks!

simonbrandhorst (approved):

PR #3066, opened 2023-12-04

HereAround:
* Support construction of Weierstrass models over concrete bases.
* Optimize constructors for literature models.
* Check that all divisor classes of (internal) model sections are effective.
* Found bug in constructor of zero sections: For literature models over concrete bases, the coordinates of said base must be taken properly into account. For now, reverted the code to simply return strings and not polynomials (yet).


apturner (approved):
Looks good, @HereAround. We’ll have to come back and fix the sections in the future, but this is good for now

PR #3068, opened 2023-12-05
Remove instance dispatches for `{elem,parent}_type`

lgoettgens:
As they already exist via https://github.com/Nemocas/AbstractAlgebra.jl/blob/9dd22199b5262f23ad071b877f5d4fb5ce9ad436/src/fundamental_interface.jl#L53 and are only inconsistently used throughout Oscar.


thofma:
Do we have a universal `parent_type(x) = parent_type(typeof(x))` or why can we remove them all?

lgoettgens:
> Do we have a universal `parent_type(x) = parent_type(typeof(x))` or why can we remove them all?

The case for `elem_type` is linked in my initial message. `parent_type` is in the same file ~20 lines below.

thofma (approved):

PR #3051, opened 2023-11-27
More functionality for Weyl groups

felix-roehrich:
- make WeylGroups iterable
- add WeylOrbit functionality
- add more tests
- fix bugs in CartanMatrix.jl and RootSystem.jl

PR #3070, opened 2023-12-05
Enable group conformance tests

lgoettgens:
Now that finally GroupsCore v0.4.1 has been released, we can enable the group conformance tests again.

@felix-roehrich Once this is merged, you can rebase https://github.com/oscar-system/Oscar.jl/pull/3051 on top and have the conformance test available there as well.


lgoettgens:
Please wait with reviews until this is marked as ready. Sorry for pinging you too early.

thofma:
I thought the idea was to drop GroupsCore eventually. Or am I confusing it with some other package?

fingolfin:
Indeed: I don't really see any benefit from having GroupsCore, only drawbacks.

lgoettgens:
I am fine with dropping it, IF we provide a similar group interface in AA.

thofma:
We did not use it before, so what do we lose by continuing with not using it?

lgoettgens:
We are possibly using it for a lot of group things (it's a dependency of AA). The stuff that was not used before were the conformance tests

lgoettgens:
> I thought the idea was to drop GroupsCore eventually. Or am I confusing it with some other package?

If we want to drop it in favor of our own Group interface e.g. in AA, this has to happen before Oscar 1.0, right?

fingolfin:
We "use" it only in the sense that we derive our Group type from theirs (which is one of the things causing pain) and I think one exception type. I can't think of anything else. The main idea always was to agree on some conventions and names for various functions and then use them across different packages implementing groups, but that never really happened.

lgoettgens:
> We "use" it only in the sense that we derive our Group type from theirs (which is one of the things causing pain) and I think one exception type. I can't think of anything else. The main idea always was to agree on some conventions and names for various functions and then use them across different packages implementing groups, but that never really happened.

For example
```julia
julia> @which istrivial(generic_group([0,1], (x,y) -> (x+y)%2)[1])
istrivial(G::GroupsCore.Group)
     @ GroupsCore ~/.julia/packages/GroupsCore/xRZbK/src/groups.jl:144
```
falls back to some GroupsCore functionality, and I cannot imagine right now how much else as well.
Nonetheless, even changing the parent type of `AA.Group` from `GroupsCore.Group` to `AA.Set` should be considered a breaking change. Everyone depending on both AA/Oscar and GroupsCore will notice that the generic functions are no longer applicable.
Thus, I think that this change should be on the 1.0 Roadmap.

PR #3071, opened 2023-12-06
Bump dependencies

thofma:
Needs more adjustments.

Closes #2976
Closes #2992


fingolfin:
Updated to Singular 0.21, and added a couple commits adjusting a few things

thofma:
Why delete `_variables_for_singular`. Is there something in Singular for this now? If so, this must be changed in Oscar.

fingolfin:
I will push a fixed version soon, but am in a call right now

thofma:
I reverted it for now, because I want to work on it now.

fingolfin:
OK.

FYI, we may have to also restore this function, which was in `mpoly-nested.jl` -- that file was migrated to AA, but this one function was not because it references Singular:
```
# if R2 is a singular ring, keep it that way since limited types can be factored
function _add_variables(R2::Singular.PolyRing, n::Int)
  n += nvars(R2)
  x = map(i->"x"*string(i), 1:n)
  return Singular.polynomial_ring(base_ring(R2), x, cached=false)[1]
end
```

test/Rings/orderings.jl
```diff
@@ -189,7 +189,8 @@ end
 end
  
 @testset "Polynomial Orderings comparison" begin
-   R, (x, y, z) = @inferred polynomial_ring(QQ, ["x", "y", "z"])
+   #R, (x, y, z) = @inferred polynomial_ring(QQ, ["x", "y", "z"])
+   R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
```
fingolfin: This can be reverted, I think?
```suggestion
   R, (x, y, z) = @inferred polynomial_ring(QQ, ["x", "y", "z"])
```

thofma:
OK, I am now at the point, where I am running into
```
Rings/mpoly.jl:29
      From worker 5:	  Expression: sprint(show, "text/plain", x[i, j]) == "x[$(i), $(j)]"
      From worker 5:	   Evaluated: "x[3,3]" == "x[3, 3]"
```
The new polynomial rings do `x[i,j]` instead of `x[i, j]` for the variable names. What do we want? The old or the new?

fingolfin:
Argh. Another unplanned (from my side at least) difference. Whatever we do, we should also add (an issue requesting that we add) a test case to AA which checks for the exact variable names and their printing, so that we catch any further changes there earlier.

I can happily live with either version. I guess I slightly prefer the version without the spaces. Keeping it would also mean we don't have to make yet another AA release. However, downside of course is that we might have to adjust a bunch (?) of docstrings...  And another AA release is not *that* bad *sigh*.

thofma:
I would be in favour of adjusting it in AA (as a "bug fix"), because it is that status quo that was probably decided on some time ago.

fingolfin:
also fine by me. perhaps @fieker can just make a decision

thofma:
Claus leans towards `x[i, j]`. Modulo https://github.com/Nemocas/AbstractAlgebra.jl/pull/1521, this PR here should be good to go.

thofma:
@fingolfin which stuff should still be removed? (related to the revert of your commit earlier)

thofma:
I will merge once everything is green. Stuff can be removed later and this here is prone to merge conflicts.

PR #3073, opened 2023-12-06
Adv/serialization for multigraded implicitization

antonydellavecchia:
Adds some functions that help with the setup of parallelization, that will then allow
users to use pmap.

waiting for https://github.com/oscar-system/Oscar.jl/pull/3032 to merged before merging this pr.
adds some serialization necessary to setup and run a parallel process for https://github.com/bkholler/MultigradedImplicitization.jl


antonydellavecchia:
see https://github.com/oscar-system/Oscar.jl/pull/3079

PR #3076, opened 2023-12-07
New method to convert toric divisors into Weil divisors

HechtiDerLachs:
@lkastner : After your explanations this morning I tried to implement what we talked about. However, when building the divisor locally, the MODULE_GENERATORS seem to not lay in the `weight_cone` of the affine variety! Then `solve_mixed` returns empty results and I can't proceed. 

I left this as a work in progress. Could you take a look?


lkastner:
Can you post a testcase? In general:
```
julia> ntv = normal_toric_variety(positive_hull([1]))
Normal toric variety

julia> td1 = toric_divisor(ntv, [-1])
Torus-invariant, non-prime divisor on a normal toric variety

julia> td1.polymake_divisor.MODULE_GENERATORS
pm::Matrix<pm::Integer>
1


julia> td0 = toric_divisor(ntv, [1])
Torus-invariant, prime divisor on a normal toric variety

julia> td0.polymake_divisor.MODULE_GENERATORS
pm::Matrix<pm::Integer>
-1
```
This means that the divisors we talked about with `+1` will correspond to fractional ideals, and those with `-1` correspond to actual ideals. Maybe that clears up the confusion?

experimental/Schemes/NormalToricVarieties/attributes.jl
```diff
@@ -77,3 +77,56 @@ function _torusinvariant_weil_divisors(X::NormalToricVariety; check::Bool=false)
...
+      @show i
+      # Manually convert the rows of A to a ZZMatrix so that solve_mixed will take it
+      # Why don't we do A all at once? Because `solve_mixed` won't accept real matrices 
+      # as second argument. Duh. 
+      b = zero_matrix(ZZ, ncols(A), 1)
+      for j in 1:ncols(A)
+        b[j, 1] = ZZ(A[i, j])
+      end
+      @show transpose(hb)
+      @show b
```
HechtiDerLachs: This is the part of the code I was talking about. The example I was running was:
```julia
IP = projective_space(NormalToricVariety, 2)
Oscar._weil_divisor_via_polymake(IP, elem_type(ZZ)[1, 0, 0])
```

HechtiDerLachs:
@lkastner : Yes, thanks! That must have been the issue. Sorry for not posting the example code earlier. My connection was lost when I wrote my original messages. 

I will try again with your suggestions this afternoon.

HechtiDerLachs:
Oops... Looks like we need to go through the sign convention once more. Tomorrow, then.

experimental/Schemes/NormalToricVarieties/attributes.jl
```diff
@@ -51,29 +51,119 @@
...
-  for tau in ray_list
-    tau_dual = polarize(cone(tau))
-    ideal_dict = IdDict{AbsSpec, Ideal}()
-    for U in affine_charts(X)
-      if !(tau in cone(U))
-        ideal_dict[U] = ideal(OO(U), one(OO(U)))
-        continue
+function _torusinvariant_weil_divisors(X::NormalToricVariety; check::Bool=false, algorithm::Symbol=:via_polymake)
+  return get_attribute!(X, :_torusinvariant_weil_divisors) do
+    ray_list = rays(polyhedral_fan(X))
```
HereAround, line 56: I understand this function works as long as `X` is simplicial? So maybe:
```suggestion
    @req is_simplicial(X) "Variety must be simplicial"
    ray_list = rays(polyhedral_fan(X))
```

experimental/Schemes/NormalToricVarieties/attributes.jl
```diff
@@ -51,29 +51,119 @@
...
+    return result
+  end::Vector{<:AbsWeilDivisor}
+end
+
+function _ideal_sheaf_via_polymake(X::NormalToricVariety, i::Int; check::Bool=false)
+  return _ideal_sheaf_via_polymake(X, [i==j ? one(ZZ) : zero(ZZ) for j in 1:nrays(polyhedral_fan(X))])
+end
+
+function _ideal_sheaf_via_polymake(X::NormalToricVariety, c::Vector{ZZRingElem}; check::Bool=false)
+  # Input:
```
HereAround, line 90: Likewise here:
```suggestion
  @req is_simpicial(X) "Variety must be simplicial"
  # Input: 
```

experimental/Schemes/ToricDivisors/attributes.jl
```diff
@@ -53,13 +53,24 @@ function forget_toric_structure(td::ToricDivisor)
...
-  if has_attribute(td, :underlying_divisor)
-    return get_attribute(td, :underlying_divisor)::WeilDivisor
-  end
-  X = scheme(td)
-  generating_divisors = _torusinvariant_weil_divisors(X; check)
-  result = sum(a*D for (a, D) in zip(coefficients(td), generating_divisors))
-  set_attribute!(td, :underlying_divisor=>result)
-  return result
+function underlying_divisor(td::ToricDivisor; check::Bool=false, algorithm::Symbol=:direct)
+  return get_attribute!(td, :underlying_divisor) do
```
HereAround, line 57: We could likewise have a check here for a simplicial toric variety, but this will be triggered by the other methods later down this method anyways, unless I am mistaken.

test/AlgebraicGeometry/ToricVarieties/toric_schemes.jl
```diff
@@ -114,9 +114,42 @@ end
...
+  K0 = Oscar.underlying_divisor(w, algorithm=:via_polymake)
+  delete!(w.__attrs, :underlying_divisor)
+  K1 = Oscar.underlying_divisor(w, algorithm=:direct)
+  delete!(w.__attrs, :underlying_divisor)
+  K2 = Oscar.underlying_divisor(w, algorithm=:via_oscar)
+  delete!(w.__attrs, :underlying_divisor)
+  
+  @test K1 == K2
+  @test K1 == K0
+  @test K2 == K0
```
HereAround, line 152: Is this last line necessary? In theory, K1 == K2 and K0 == K1 implies K0 == K2. Maybe this is also true in `OSCAR` (for deeper reasons)?

HereAround (approved):
Thank you for working on this @HechtiDerLachs . It seems that the doctests now complete (other than nightly), so I assume the sign error has been fixed. Great!

I left a few minor comments (nitpicks) above.

experimental/Schemes/NormalToricVarieties/attributes.jl
HechtiDerLachs: I think @lkastner is the right person to ask. If this applies, we should add an `@check`.

test/AlgebraicGeometry/ToricVarieties/toric_schemes.jl
HechtiDerLachs: These trigger different calls for the comparison. So they test different code, indeed.

PR #3077, opened 2023-12-07
First attempts at running CI on a self hosted server (in parallel)

aaruni96:
[skip ci] until someone double checks that the yaml I wrote makes sense

cc @fingolfin @benlorenz

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,8 +49,6 @@ jobs:
             # right ?
             os: rptu-1
             group: long
-            env:
-              NUMPROCS: '6'
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


benlorenz:
The yaml looks ok and yes the `os` value is used for runs-on. If something is wrong github should refuse to run it and report an error.

aaruni96:
I think I need to merge this into `master` before the new CI takes effect. Can someone review? (Or, should I force merge?)

benlorenz:
This should be able to run on the PR as well but you need to remove the `[skip ci]` and then force push.

.github/workflows/CI.yml
```diff
@@ -43,6 +43,12 @@ jobs:
           # Add a few macOS jobs (not too many, the number we can run in parallel is limited)
           - julia-version: 'nightly'
             os: macOS-latest
+          # Add a job on the local runner
+          - julia-version: '1.9'
+            #os doesn't need to be a real OS. Its just the variable which is evaluated for runs-on
+            # right ?
+            os: rptu-1
+            group: long
```
benlorenz, line 51:
```suggestion
            group: ''
```
I think it would make sense to set it to empty for this job. Then all tests are run instead of just one of the two subsets.
(Especially the long test-group doesn't parallelize that well as it contains only 17 files)

fingolfin:
@aaruni96 can this be closed now?

aaruni96:
Sorry, I missed your notification from two weeks ago. Yes this has been superseded by the already merged PR #3650

PR #3079, opened 2023-12-07
Adv/serialization for multigraded implicitization

antonydellavecchia:
Adds some functions that help with the setup of parallelization, that will then allow
users to use pmap.

waiting for https://github.com/oscar-system/Oscar.jl/pull/3032 to merged before merging this pr.
adds some serialization necessary to setup and run a parallel process for https://github.com/bkholler/MultigradedImplicitization.jl


src/exports.jl
```diff
@@ -609,7 +609,6 @@ export graded_free_module
 export graded_image
 export graded_map
 export graded_polynomial_ring
-export grading
```
benlorenz, line 612: Just a note for anyone else looking at this:
We removed the export as this function was broken (it never worked) and also undocumented.

benlorenz (approved):

PR #3081, opened 2023-12-07
Streamline powers of Lie algebra modules

lgoettgens:
Make the code very similar to the exterior powers of ModuleFP by @HechtiDerLachs.

- [x] exterior powers
- [x] symmetric powers
- [x] tensor powers
- [x] cleanup of generic "power code"

Furthermore, fix some type instabilities in the exterior power code of ModuleFP.


experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -987,14 +1014,11 @@ function exterior_power(V::LieAlgebraModule{C}, k::Int) where {C<:FieldElem}
     end
     return E(mat)
   end
-  function pure(as::Tuple)
-    return pure(as...)::LieAlgebraModuleElem{C}
-  end
-  function pure(as::Vector{LieAlgebraModuleElem{C}})
-    return pure(as...)::LieAlgebraModuleElem{C}
+  function my_mult(as::Tuple)
+    return my_mult(as...)::LieAlgebraModuleElem{C}
```
HechtiDerLachs: Recently @thofma pointed out to me that splatting a vector potentially induces more compilation overhead: For every length of the input, a new method has to be compiled. I guess here it's not so bad as the input length is fixed, but in general, we should think of routing things the other way round: collect a tuple and go through the method with a vector as input.

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1006,18 +1030,28 @@ function exterior_power(V::LieAlgebraModule{C}, k::Int) where {C<:FieldElem}
...
     :embedding_tensor_power_projection => T_to_E,
   )
-  return E
+
+  cached && (_exterior_powers(V)[k] = (E, mult_map))
+
+  return E, mult_map
+end
+
+@attr Dict{Int,Tuple{typeof(V),MapFromFunc}} function _exterior_powers(V::LieAlgebraModule)
```
HechtiDerLachs:
```suggestion
@attr Dict{Int, <:Tuple{typeof(V), <:MapFromFunc}} function _exterior_powers(V::LieAlgebraModule)
```
I think in the constructor of dicts you can use abstract types. But for type assertion, I always needed the `<:`. Did this code even work before?

src/Modules/ExteriorPowers/Generic.jl
```diff
@@ -1,14 +1,14 @@
 # We need to cache eventually created exterior powers.
-@attr Dict{Int, Tuple{T, <:Map}} function _exterior_powers(F::T) where {T<:ModuleFP}
-  return Dict{Int, Tuple{typeof(F), Map}}()
+@attr Dict{Int, Tuple{T, MapFromFunc}} function _exterior_powers(F::T) where {T<:ModuleFP}
```
HechtiDerLachs: Same here, I guess.

HechtiDerLachs (dismissed):
Looks good. As already said: At some point we need to streamline the multiplication map.

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens: I changed it both here and for ring modules.

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens: I played around with it. Having abstract types in the dict constructor makes type inference nearly impossible. Changing the type assertion didn't change anything for me. But intuitively, I would say, that the type assertion should exactly match the type of the dict created here. So I would keep it as it currently is.

lgoettgens:
This PR is now ready to be reviewed and merged from my POV. There is still some stuff do do around direct sums and tensor products, but I would like to postpone that to a future PR that tackles #3059 for Lie algebra modules.

@HechtiDerLachs could you please in particular have a look at the changes in `src/`? It should only be the addition of type assertions where they are needed for inference to be more successful. You can see the result when calling `@code_warntype exterior_power(F, 2)` where `F` is a free module and comparing that to the latest master.

lgoettgens:
bump @HechtiDerLachs

src/Modules/ExteriorPowers/FreeModules.jl
```diff
@@ -9,42 +9,44 @@
 
 # User facing constructor for ⋀ ᵖ F.
 function exterior_power(F::FreeMod, p::Int; cached::Bool=true)
-  (p < 0 || p > rank(F)) && error("index out of bounds")
+  @req 0 <= p <= rank(F) "Exponent out of bounds"
```
HechtiDerLachs:
```suggestion
  @req 0 <= p <= rank(F) "exponent out of bounds"
```
This should be the convention: Don't capitalize.

src/Modules/ExteriorPowers/FreeModules.jl
```diff
@@ -9,42 +9,44 @@
...
 
   if cached
     powers = _exterior_powers(F)
-    haskey(powers, p) && return powers[p]::Tuple{typeof(F), <:Map}
+    haskey(powers, p) && return powers[p]
```
HechtiDerLachs, line 16: Why was this changed?

src/Modules/ExteriorPowers/FreeModules.jl
```diff
@@ -9,42 +9,44 @@
...
   end
 
-  R = base_ring(F)
+  R = base_ring(F)::base_ring_type(F)
   n = rank(F)
-  result = FreeMod(R, binomial(n, p))
+  result_ = free_module(R, binomial(n, p))
```
HechtiDerLachs: I'm not sure that this is the right thing to do. `free_module` has not been completely compatible with `FreeMod` and this means that you might get the wrong type (`FPModule` rather than `ModuleFP`). Are you sure this is correct here?

src/Modules/ExteriorPowers/Generic.jl
```diff
@@ -1,14 +1,14 @@
 # We need to cache eventually created exterior powers.
-@attr Dict{Int, Tuple{T, <:Map}} function _exterior_powers(F::T) where {T<:ModuleFP}
-  return Dict{Int, Tuple{typeof(F), Map}}()
+@attr Dict{Int, Tuple{typeof(F), MapFromFunc}} function _exterior_powers(F::ModuleFP)
+  return Dict{Int, Tuple{typeof(F), MapFromFunc}}()
```
HechtiDerLachs, line 3:
```suggestion
@attr Dict{Int, Tuple{typeof(F), <:MapFromFunc}} function _exterior_powers(F::ModuleFP)
  return Dict{Int, Tuple{typeof(F), MapFromFunc}}()
```
This might be a bit too restrictive. Maybe, one day we change the return type. And they are both abstract, so no harm is done staying with just `Map` right?

src/Modules/ExteriorPowers/SubQuo.jl
```diff
@@ -1,16 +1,15 @@
...
 
   if cached
     powers = _exterior_powers(M)
     haskey(powers, p) && return powers[p]
   end
 
-  result = M # Initialize variable
   if iszero(p)
-    F = FreeMod(R, 1)
+    F = free_module(R, 1)
```
HechtiDerLachs: Same as above.

src/Modules/ExteriorPowers/FreeModules.jl
lgoettgens: This didn't help type inference. Iirc it was even making it even worse in some cases

src/Modules/ExteriorPowers/FreeModules.jl
lgoettgens: I didn't know about that
```suggestion
  result_ = FreeMod(R, binomial(n, p))
```

src/Modules/ExteriorPowers/SubQuo.jl
lgoettgens:
```suggestion
    F = FreeMod(R, 1)
```

src/Modules/ExteriorPowers/Generic.jl
lgoettgens: Tuples are invariant in julia, thus the `<:` shouldn't be there. 
As all of this is purely internal, we can change that to Map back anytime we want. But making it as concrete as possible helps type inference massively. In this particular case, it can use some of the annotations in the `MapFromFunc` code in Hecke

HechtiDerLachs:
Tests still fail. Other than that, I can approve for my part. Let me know when you're ready.

src/Modules/ModulesGraded.jl
```diff
@@ -513,7 +513,7 @@ function degree(f::FreeModElem)
   iszero(f) && return A[0]
   f.d = isa(f.d, GrpAbFinGenElem) ? f.d : determine_degree_from_SR(coordinates(f), degrees(parent(f)))
   isa(f.d, GrpAbFinGenElem) || error("The specified element is not homogeneous.")
-  return f.d
+  return f.d::GrpAbFinGenElem
```
lgoettgens, line 516: btw this was the culprit

lgoettgens:
I found a small fix for the failure, that made it possible to revert some other changes to enhance readability.
Let's hope that I broke nothing else.

lgoettgens:
> Tests still fail. Other than that, I can approve for my part. Let me know when you're ready.

CI now looks good

src/Modules/ExteriorPowers/SubQuo.jl
```diff
@@ -1,14 +1,13 @@
 function exterior_power(M::SubquoModule, p::Int; cached::Bool=true)
   n = rank(ambient_free_module(M))
-  R = base_ring(M)
-  ((p >= 0) && (p <= n)) || error("exponent out of range")
+  R = base_ring(M)::base_ring_type(M)
+  @req 0 <= p <= n "Exponent out of bounds"
```
HechtiDerLachs, line 4:
```suggestion
  @req 0 <= p <= n "exponent out of bounds"
```

HechtiDerLachs (approved):
I think the style adjustment is not worth running the tests again. So if there's nothing else, go ahead with the merge from my side.

fingolfin (approved):

PR #3085, opened 2023-12-08
WIP on enumeration of definite integer genera

StevellM:
[skip ci]

Work in progress on a new interface for enumeration of definite `ZZGenus`.

What is missing:
- Save and load partial results;
- Use enumeration of isometries to complete genera;
- Allow to construct neighbour graph;
- Incoming improvements on isometry group and isometry test.

The current plan for the interface has been discussed previously. Nothing is final for the names and the shape of the functions.

Currently, random enumeration is the most efficient, though its efficiency is... random.

Todo:
- Put the "core" parts of the algorithms in a separate box instead of duplicating it;
- Benchmark the code and compare with actual Hecke/Magma implementations;
- Try to re-enumerate known genera in a decent amount of time;
- Implement heuristics to choose the "best" algorithm in the case of unknown user preference.


StevellM:
This will be moved to Hecke at the end, so I close the PR.

PR #3084, opened 2023-12-08
docu sheaf cohomology

wdecker:
@RafaelDavidMohr There are a couple of issues here:

1) The numbers in the first column between `twist` and `chi` are upside down. That is , `0:` must stand directly above `chi` etc. The examples in `Modules_Graded.jl` must be changed accordingly (including things like `tbl[3, 2]`).

2) For the convenience of the reader, please assert that
    a) The base ring is standard ZZ-graded,
    b) the coefficent ring of the base ring is a field, and
    c) the considered module is graded.

Please add the changes to this PR or merge this PR and make a separate PR.

PR #3087, opened 2023-12-08
Fixes sheaf cohomology printing, adds check that coefficient ring is a field

RafaelDavidMohr:
Changes as requested in https://github.com/oscar-system/Oscar.jl/pull/3084


wdecker (approved):

src/Modules/ModulesGraded.jl
```diff
@@ -2052,6 +2084,8 @@ end
 
 elem_type(::Type{FreeMod_dec{T}}) where {T} = FreeModElem_dec{T}
 parent_type(::Type{FreeModElem_dec{T}}) where {T} = FreeMod_dec{T}
+elem_type(::FreeMod_dec{T}) where {T} = FreeModElem_dec{T}
+parent_type(::FreeModElem_dec{T}) where {T} = FreeMod_dec{T}
```
HechtiDerLachs, line 2088: Sorry, but I became curious: Where do these types come from and where are they used? I thought that the graded modules are the same as the usual modules. Did anything change there?

RafaelDavidMohr: This is beyond my scope to be honest, maybe @jankoboehm knows.

wdecker: @HechtiDerLachs Please resolve this in a direct conversation with @jankoboehm

thofma, line 2088: Why are those methods necessary? https://github.com/oscar-system/Oscar.jl/pull/3068 went to great lengths to remove such methods.

CC: @lgoettgens

jankoboehm: Don't use for gradings, we plan to use this for filtrations, once they are sorted out for rings.

PR #3090, opened 2023-12-08
Remove some dead code

fingolfin:


src/Rings/hilbert.jl
```diff
@@ -1099,7 +1099,7 @@ function _hilbert_series_ring(parent::Union{Nothing, Ring}, m::Int)
   if m == 1
     VAR = [:t]
   else
-    VAR = [_make_variable("t", i) for i = 1:m]
```
thofma, line 1102: Why was this so complicated before?

thofma: And why not just use a string nowadays?

fingolfin: One thing is type stability of `VAR`. And in the end the var names are converted to symbols, so why not provide symbols from the start?

thofma (approved):

Issue #3083, opened 2023-12-08
`primary_decomposition` won't finish on very simple input


HechtiDerLachs:
I have a principal ideal for which the generator can easily be decomposed with `factor`. If I run `primary_decomposition`, however, the computation simply won't finish. 

I wrote out the ideal using the serialization, the file should be here:
[difficult_ideal.txt](https://github.com/oscar-system/Oscar.jl/files/13613084/difficult_ideal.txt)

Download the file, start Oscar and type `I = load("difficult_ideal.txt")`. 
Now try `primary_decomposition(I)` and in contrast `factor` for the one non-zero generator. 

Any ideas what might be going on here?

HechtiDerLachs:
CC: @hannes14, @afkafkafk13, @wdecker .

HechtiDerLachs:
Here's another case with a simple algebraic extension of QQ as coefficient ring: 
[another_difficult_ideal.json](https://github.com/oscar-system/Oscar.jl/files/13613865/another_difficult_ideal.json)
`minimal_primes` takes ages to finish while `factor` on the one generator returns instantly.

hannes14:
Singular should report an error here: it does not anything about the coefficient ring (beside basic arithmetic),
so for example Groebner base computation will work, but for example factorization not.

afkafkafk13:
For me this gives the impression that Singular is being called with insufficiently prepared input. 

We know very well about the limitations of Singular, when working over algebraic field extensions and in particular of the primary decomposition functionality. This could be taken into account in the respective Oscar method.

A quick and dirty (i.e. rather suboptimal) solution could be to disallow the call to  Singular's primary decomposition in those cases which cannot be handled by Singular directly. A more elaborate and more satisfactory solution would be to rewrite the polynomial ideal over the (possibly repeated) algebraic extension of QQ into a polynomial ideal in a larger polynomial ring over QQ taking the minimal polynomials of the (tower of) extensions as additional generators of the ideal -- like the Singular user currently does, when such a situation arises.

HechtiDerLachs:
> Singular should report an error here: it does not anything about the coefficient ring (beside basic arithmetic),
> so for example Groebner base computation will work, but for example factorization not.

Makes sense. But the CPU is running on 100% either way. 

Is there any way I would get a primary decomposition to work in my cases?

HechtiDerLachs:
[...]

PR #3091, opened 2023-12-08
Primary decomposition rerouting

HechtiDerLachs:
A prototype for what @afkafkafk13 suggested and @wdecker requested in #3083 .

### DIFF
--- a/src/Rings/MPolyMap/MPolyAnyMap.jl
+++ b/src/Rings/MPolyMap/MPolyAnyMap.jl
@@ -250,7 +250,14 @@ end
 # Julia functions in both maps
 function compose(F::MPolyAnyMap{D, C, <: Function}, G::MPolyAnyMap{C, E, <: Function}) where {D, C, E}
   @req codomain(F) === domain(G) "Incompatible (co)domain in composition"
-  return hom(domain(F), codomain(G), x -> coefficient_map(G)(coefficient_map(F)(x)), G.(_images(F)), check=false)
+  b = coefficient_map(F)(one(coefficient_ring(domain(F))))
+  if parent(b) === domain(G)
+    return hom(domain(F), codomain(G), x -> G(coefficient_map(F)(x)), G.(_images(F)), check=false)
+  elseif parent(b) === coefficient_ring(domain(G))
+    return hom(domain(F), codomain(G), x -> coefficient_map(G)(coefficient_map(F)(x)), G.(_images(F)), check=false)
+  else
+    error("coefficient map is not admissible")
+  end
 end
 
 # Now compose with arbitrary maps
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -475,6 +475,9 @@ Pfister, Sadiq, and Steidel. See [GTZ88](@cite), [SY96](@cite), and [PSS11](@cit
 !!! warning
     The algorithm of Gianni, Trager, and Zacharias may not terminate over a small finite field. If it terminates, the result is correct.
 
+!!! warning
+    If computations are done in a ring over a number field, then the output may contain redundant components.
+
 If `cache=false` is set, the primary decomposition is recomputed and not cached.
 
 # Examples
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -706,7 +706,10 @@ end
     h = first([h for (h, _) in factor(f_kk)])
     kk_ext, zeta = extension_field(h)
     iso_kk_ext = hom(L, kk_ext, zeta)
-    P_prime_ext = map_coefficients(iso_kk_ext, P_prime)
+    br = base_ring(P_prime)
+    LoR, to_LoR = change_base_ring(kk_ext, R)
+    help_map = hom(br, LoR, iso_kk_ext, to_LoR.(iso.(gens(R_exp))))
+    P_prime_ext = ideal(LoR, help_map.(gens(P_prime)))
     push!(full_res, (P, Q, P_prime_ext, degree(h)))
   end
   return full_res
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -600,6 +600,9 @@ defined over a number field of degree $d_{ij}$ whose generator prints as `_a`.
 The implementation combines the algorithm of Gianni, Trager, and Zacharias for primary
 decomposition with absolute polynomial factorization.
 
+!!! warning
+    Over number fields this proceduce might return redundant output.
+
 # Examples
 ```jldoctest
 julia> R, (y, z) = polynomial_ring(QQ, ["y", "z"])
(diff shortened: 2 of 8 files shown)


src/Rings/MPolyMap/MPolyAnyMap.jl
```diff
@@ -250,7 +250,14 @@ end
...
   @req codomain(F) === domain(G) "Incompatible (co)domain in composition"
-  return hom(domain(F), codomain(G), x -> coefficient_map(G)(coefficient_map(F)(x)), G.(_images(F)), check=false)
+  b = coefficient_map(F)(one(coefficient_ring(domain(F))))
+  if parent(b) === domain(G)
+    return hom(domain(F), codomain(G), x -> G(coefficient_map(F)(x)), G.(_images(F)), check=false)
+  elseif parent(b) === coefficient_ring(domain(G))
+    return hom(domain(F), codomain(G), x -> coefficient_map(G)(coefficient_map(F)(x)), G.(_images(F)), check=false)
+  else
+    error("coefficient map is not admissible")
+  end
```
HechtiDerLachs, line 260: @thofma : I think there was a bug here. Please check that this is now correct.

thofma: Which input made it fail?

thofma:
Why do you reroute `nf_elem`? Doesn't Singular automatically do something useful if you have a simple extension over Q or k?

src/Rings/mpoly-ideals.jl
```diff
@@ -531,6 +531,23 @@ function primary_decomposition(I::T; algorithm::Symbol=:GTZ, cache::Bool=true) w
   end::Vector{Tuple{T,T}}
 end
 
+function primary_decomposition(
+    I::MPolyIdeal{T}; 
+    algorithm::Symbol=:GTZ, cache::Bool=true
+  ) where {U<:Union{nf_elem, <:Hecke.NfRelElem}, T<:MPolyRingElem{U}}
```
thofma, line 540: The signature needs to be narrowed down:
```
julia> K, a = number_field([x - 1, x - 2]);

julia> Kt, t = K["t"];

julia> L, b = number_field(t - 1, "b");

julia> M, = number_field(t - 1, "b");

julia> Mx, = polynomial_ring(M, 2);

julia> primary_decomposition(ideal(Mx, [gen(Mx, 1)]))
ERROR: MethodError: no method matching _flatten_ring(::MPolyQuoRing{AbstractAlgebra.Generic.MPoly{NfAbsNSElem}})

Closest candidates are:
```

thofma:
I am not a big fan of the name `_flatten_ring`, because we already have `flatten`. Maybe we can find a more descriptive name.

afkafkafk13:
> Why do you reroute `nf_elem`? Doesn't Singular automatically do something useful if you have a simple extension over Q or k?

Yes and no. In general (in particular in the groebner bases context), it does. 
But in the primary decomposition context, it is more desirable to go down all the way to QQ or k for internal reasons.

HechtiDerLachs:
@ederc : There seems to be some hickup with algebraic solving in the tests here. I suppose it does not have to do with my changes. Do you agree?

HechtiDerLachs:
> I am not a big fan of the name `_flatten_ring`, because we already have `flatten`. Maybe we can find a more descriptive name.

Agreed. We will figure it out next week.

src/Rings/MPolyMap/MPolyAnyMap.jl
HechtiDerLachs: When the coefficient_map of `F` takes values in the domain of `G`. This scenario was legitimate as far as I recall. The tests added here fail in case you remove this fix.

src/Rings/mpoly-ideals.jl
HechtiDerLachs: It seems that there is another type `NfAbsNSElem` lurking around which I was not aware of. We'll settle this next week, I guess.

thofma: There is also a fourth type. It was not clear to me whether this PR was intended to handle all types or just towers of simple extensions?

src/Rings/MPolyMap/MPolyAnyMap.jl
thofma: Looks good. Can you add a test for the else branch to `test/Rings/MPolyAnyMap/MPolyRing`? It seems to not be covered.

thofma:
Since you will have to touch the code anyway, maybe you could also add some documentation in form of a comment (just the general idea, maybe at the top of `src/Rings/primary_decomposition_helpers.jl`).

ederc:
I tested locally with Julia 1.6 and I cannot recreate the failing test. Also in PR #3090 the test does not fail which is strange.

src/Rings/mpoly-ideals.jl
HechtiDerLachs: I don't know either. I think we're fine with the towers of simple extensions for the moment. The rest can be done in a later step. I will leave a TODO comment in the code.

HechtiDerLachs:
If tests pass, this should be ready to go.

thofma (changes requested):
I still get incomprehensible errors for
```
julia> K, a = number_field([x - 1, x - 2]);

julia> Kt, t = K["t"];

julia> L, b = number_field(t - 1, "b");

julia> M, = number_field(t - 1, "b");

julia> Mx, = polynomial_ring(M, 2);

julia> primary_decomposition(ideal(Mx, [gen(Mx, 1)]))
ERROR: MethodError: no method matching _expand_coefficient_field(::MPolyQuoRing{AbstractAlgebra.Generic.MPoly{NfAbsNSElem}}; rec_depth::Int64)
```

This is a regression, since it is working fine on current master:
```
julia> Qx, x = QQ["x"]

julia> K, a = number_field([x - 1, x - 2]);

julia> Kt, t = K["t"];

julia> L, b = number_field(t - 1, "b");

julia> M, = number_field(t - 1, "b");

julia> Mx, = polynomial_ring(M, 2);

julia> primary_decomposition(ideal(Mx, [gen(Mx, 1)]))
1-element Vector{Tuple{MPolyIdeal{AbstractAlgebra.Generic.MPoly{Hecke.NfRelElem{NfAbsNSElem}}}, MPolyIdeal{AbstractAlgebra.Generic.MPoly{Hecke.NfRelElem{NfAbsNSElem}}}}}:
 (ideal(x1), ideal(x1))
```

Catching this cannot be handled using the signature, so you need to use something like
```
_is_a_tower_we_can_handle(K::QQField) = true

_is_a_tower_we_can_handle(K::NumField) = (K isa AnticNumberField || K isa Oscar.Hecke.NfRel) && _is_a_tower_we_can_handle(base_field(K))
```

And then check inside the function and call the ordinary primary decomposition in case it returns `false`.

Alternatively, methods need to be added for conversion of the other field types.

HechtiDerLachs:
What now? Can this be merged?

thofma:
As I said on slack, I will review it today.

src/Rings/MPolyMap/MPolyAnyMap.jl
HechtiDerLachs: Done.

src/Rings/mpoly-ideals.jl
HechtiDerLachs: I implemented all types of number fields now. For the moment, I still use the `.data` field of the elements. If there is a better and systematic way to get that information, then we can (and probably should) change that. Otherwise, I think, it can stay like this for now. If the internals of the number fields change in a way that this is affected, we will see it in the tests failing.

thofma (approved):

thofma:
It's a shame we (I) did not implement `hom(K::NumField, R::Ring)` for arbitrary rings. It would have turned those `x -> evaluate(x.data, theta)` into neat `hom(K, R_exp, theta)`.

HechtiDerLachs:
Yes, that's what I would have needed. Let me know, should that change one day.

thofma:
I will adjust it once https://github.com/thofma/Hecke.jl/pull/1324 is merged.

HechtiDerLachs:
Very good, thanks!

PR #3096, opened 2023-12-12
Add sqrt and cbrt functions for QQAbElem

### LINKED ISSUE #3011: Provide working `sqrt` for `QQAbElem`
Consider this example:
```
julia> QQab, t = abelian_closure(QQ)
(Abelian closure of Q, Generator of abelian closure of Q)

julia> a = QQab(2)
2

julia> sqrt(a)
ERROR: StackOverflowError:

julia>
```
That error is *not* helpful... but the method it ends up invoking is this from AA:
```
function Base.sqrt(a::FieldElem; check::Bool=true)
  R = parent(a)
  R, t = polynomial_ring(R, "t", cached = false)
  f = factor(t^2 - a)
  for (p, e) in f
    if !check || degree(p) == 1
      return -divexact(coeff(p, 0), coeff(p, 1); check=check)
    end
  end
  error("Element $a does not have a square root")
end
```

Anyway, I hope we can do better and either implement a nice `sqrt` method (and maybe also `cbrt`), or else at least provide a proper error.

aaruni96:
Adds `sqrt` and `cbrt` functions for QQAbElem. Resolves #3011 .

### DIFF
--- a/test/Rings/AbelianClosure.jl
+++ b/test/Rings/AbelianClosure.jl
@@ -264,6 +264,10 @@ end
     @test length(roots(x^15-2^15)) == 15
 
     @test order(z(5)) == 5
+
+    @test cbrt(K(8)) == 2
+    @test_throws ErrorException("Element 4 does not have a cube root") cbrt(K(4))
+
   end
 
   @testset "Automorphism" begin
@@ -291,6 +295,9 @@ end
       x, y, n = Oscar.AbelianClosure.quadratic_irrationality_info(a)
       @test ((a - x)//y)^2 == n
     end
+
+    @test sqrt(K(2)) == -z(8)^3 + z(8)
+    @test_throws ErrorException("Element zeta(4) + 1 does not have a square root") sqrt(z(4)+1)
 
     @test Oscar.AbelianClosure.quadratic_irrationality_info(z(5)) === nothing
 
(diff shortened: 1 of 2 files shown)


thofma:
`roots(a, n)` returns a vector, whereas `sqrt` and `cbrt` should return a single element. Also tests would be good :)

test/Rings/AbelianClosure.jl
```diff
@@ -264,6 +264,10 @@ end
     @test length(roots(x^15-2^15)) == 15
 
     @test order(z(5)) == 5
+
+    @test cbrt(K(8)) == 2
```
thofma:
```suggestion
    @test cbrt(K(8))^3 == K(8)
```

test/Rings/AbelianClosure.jl
```diff
@@ -292,6 +296,9 @@ end
       @test ((a - x)//y)^2 == n
     end
 
+    @test sqrt(K(2)) == -z(8)^3 + z(8)
```
thofma:
```suggestion
    @test sqrt(K(2))^2 == K(2)
```

thofma (commented):
Thanks. I suggest two minor changes to the tests to make them more robust (we do not guarantee which root is returned).

May @fieker also wants to have a look?

thofma (approved):

PR #3100, opened 2023-12-12
Rename groebner_basis(::MPolyQuoRing)

### LINKED ISSUE #3086: `groebner_basis(r::MPolyQuoRing)` should be renamed
In https://github.com/oscar-system/Oscar.jl/pull/2658 we added a method
```
groebner_basis(r::MPolyQuoRing)
```
in the wake of the cleanup.

I would suggest that we rename this, given that `groebner_basis` is an exported, well-documented, function.

CC: @wdecker @ederc

thofma:
Fixes #3086.

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -21,6 +21,11 @@
 end
 
 function _groebner_basis(r::MPolyQuoRing)
+  # Assure that the correspondence to Singular is set up
+  # Note that this is the only way to create the Singular quotient ring:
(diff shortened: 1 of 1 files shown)


ederc (approved):

thofma:
@eder I added some comments and also did the same for `groebner_basis(I::MPolyQuoRingIdeal)`, because the return type is not very helpful for a user. It is either `true` or `nothing`. I suggest someone cleans this up later, if we want this functionality.

PR #3101, opened 2023-12-12
Optimize `BasisLieHighestWeight`

gfourier:
Computing the transpose of sparse matrices has been eleminated from the code at two positions. Further, experiments have shown that using the symmetry of partitions of a dominant weight does not speed up the algorithm in this order, so it is reversed.


lgoettgens:
This PR contains some changes from https://github.com/oscar-system/Oscar.jl/pull/3046 as well. Let's try to fix this next time we see each other in the office.

lgoettgens (approved):

lgoettgens:
This is ready to get merged.
@thofma could you do the favor?

lgoettgens:
Bump

PR #3109, opened 2023-12-15
Use proper morphisms in primary decomposition helpers

thofma:


src/Rings/primary_decomposition_helpers.jl
```diff
@@ -71,7 +68,8 @@ function _expand_coefficient_field(
   theta = gens(A_exp)[1:r]
   alpha = gens(coefficient_ring(A))
   to_A = hom(A_exp, A, vcat(A.(alpha), gens(A)))
-  to_A_exp = hom(A, A_exp, x->evaluate(x.data, theta), gens(A_exp)[r+1:end])
+  #to_A_exp = hom(A, A_exp, x->evaluate(x.data, theta), gens(A_exp)[r+1:end])
```
HechtiDerLachs: Is there a reason why you wanted to keep this specific line? I.e. any doubts here? 

Also: Does your new `hom` have a `check` flag which we could/should set to `false` in all these calls?

HechtiDerLachs (approved):
Much appreciated, thanks!

src/Rings/primary_decomposition_helpers.jl
thofma: Oops, just an artifact. Yes, there is a `check` argument, so we "could". I can enable it if you want. I have make another commit anyway.

HechtiDerLachs: Alright. Whether we should depends on how expensive the check is. Since the necessary Groebner bases computations here are carried out automatically with the creation of the quotient rings anyway, I think it's OK to leave it as is.

thofma: I agree. If it turns out to be a bottleneck, we can add it later.

PR #3113, opened 2023-12-15
Add missing ring interface function for AlgClosure

thofma:


src/Rings/AlgClosureFp.jl
```diff
@@ -44,6 +44,8 @@ end
 elem_type(::Type{AlgClosure{T}}) where T = AlgClosureElem{T}
 parent_type(::Type{AlgClosureElem{T}}) where T = AlgClosure{T}
 
+Oscar.canonical_unit(a::AlgClosureElem) = a
```
lgoettgens:
```suggestion
Oscar.canonical_unit(a::AlgClosureElem) = is_zero(a) ? one(a) : a
```
In fields with an easy iszero check we should use that to make sure to return a unit.

thofma: I just went with
```
julia> canonical_unit(QQ(0))
0
```

lgoettgens: There are multiple issues in AA/Nemo about that.
But returning the "unit" 0 can lead to division by zero easily (in not so well-written implementations)

lgoettgens (approved):

src/Rings/AlgClosureFp.jl
thofma: Yes. It should be renamed and redesigned properly.

PR #3114, opened 2023-12-15
Tag 0.14.0

thofma:
There have been some updates upstream, that we should bring to the (casual) user.

One small/big thing, is that we removed a lot of deprecations. In particular `PolynomialRing(QQ, )` now gives an error. Should we announce this somewhere? @fingolfin?


lgoettgens:
This PR misses the version number at multiple places, e.g. the README. If I remember correctly, we have a script somewhere for changing all occurrences.

thofma:
Thanks for the reminder

fingolfin (approved):
Thank you

fingolfin:
Regarding announcement about these changes: OK, sure, but where? I guess in the changelong on the release page (so on https://github.com/oscar-system/Oscar.jl/releases/tag/v0.14.0 analog to the existing https://github.com/oscar-system/Oscar.jl/releases/tag/v0.13.0).

Perhaps also on the mailing list, and on Slack, and ... where else? perhaps in a custom news item on the website?

thofma:
Not sure where to put.

My argument for adding a warning is that someone updating from 0.13.0 to 0.14.0 will never see a deprecation, but just an `MethodError` when calling e.g. `PolynomialRing(ZZ, :x)`. Was this our intention?

fingolfin:
Hmm, no, I don't think that was intentional. I wasn't aware. Gotta take a look

thofma:
(Same is true for `NumberField`)

fingolfin:
It probably affects even more aliases. This is not intentional. We just discussed it. I'll look into a fix after our meeting so we don't break code accidentally in 0.14.0.

Maybe in the release after, we can change things so that a warning when using these bindings are used is always printed... Dunno

thofma:
Yes, I think this would be a good idea. I somehow thought that we already did this at some point.

Apart from https://github.com/oscar-system/Oscar.jl/pull/3132#issuecomment-1868435558 this is looking good.

lgoettgens:
The warning is still there (https://github.com/oscar-system/Oscar.jl/actions/runs/7329236843/job/19958610491?pr=3114#step:7:378)

lgoettgens:
I just pushed a try to circumvent the warning. ~~Let's see if it works~~ According to CI and local testing, it works as expected.

lgoettgens:
@thofma Do you want to get this going before or after the finite field changes?

lgoettgens:
Bump

fingolfin (approved):

PR #3115, opened 2023-12-15
Aftermath of the rerouting of `primary_decomposition` and friends

HechtiDerLachs:
It seems that we have messed some things up. I already mentioned the following ideal in #3083 : 
[decomposition_ideal.json](https://github.com/oscar-system/Oscar.jl/files/13688678/decomposition_ideal.json).
Save that file, check out this PR, and start Oscar in the directory of  `decomposition_ideal.json`. Then try the following:
```julia
J1 = load("decomposition_ideal.json")

@time primary_decomposition(J1) # ~3.8s
@time minimal_primes(J1) # ~22s up to 77s without compilation time!

# manual conversion into a ring with a simple field extension of QQ as coefficient ring
S = base_ring(J1)
kk = coefficient_ring(S)
k_simp, iso = absolute_simple_field(kk)

R, iso_R = change_base_ring(pseudo_inv(iso), S)
iso_R_inv = hom(S, R, iso, gens(S))

J2 = ideal(R, iso_R.(gens(J1)))

# New timings:

@time primary_decomposition(J2) # won't finish!

@time minimal_primes(J2) # ~4.4 - 4.8s 


# Comparison with the old methods before the reouting:

@time Oscar._old_primary_decomposition(J1) # won't finish!

@time Oscar._old_primary_decomposition(J2) # won't finish!

@time Oscar._old_minimal_primes(J1) #  5.611382s / 5.373242s / 5.127290s

@time Oscar._old_minimal_primes(J2) # won't finish!
```
What I see is: While `primary_decomposition` went from impossible to 3.8 seconds we have the opposite for `minimal_primes`. Thus, we should look into the reasons for that. But in the meantime it seems that we can safely take back the rerouting for `minimal_primes` from #3091 . 

I can modify this PR so that it does exactly that; unless somebody with more insight to the internals can tell us something better about the weird behavior above. Let me know your opinions on this.

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -534,6 +534,13 @@ function primary_decomposition(I::T; algorithm::Symbol=:GTZ, cache::Bool=true) w
   end::Vector{Tuple{T,T}}
 end
 
+function _old_primary_decomposition(I::T; algorithm::Symbol=:GTZ, cache::Bool=true) where {T<:MPolyIdeal}
+  !cache && return _compute_primary_decomposition(I, algorithm=algorithm)
+  return get_attribute!(I, :primary_decomposition) do
+    return _compute_primary_decomposition(I, algorithm=algorithm)
+  end::Vector{Tuple{T,T}}
+end
+
 function primary_decomposition(
     I::MPolyIdeal{T}; 
     algorithm::Symbol=:GTZ, cache::Bool=true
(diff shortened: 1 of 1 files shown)


HechtiDerLachs:
@hannes14: Can you confirm these results? I think it's awkward that `_old_minimal_primes(J1)` differs from `_old_minimal_primes(J2)` since it should do nothing but transform `J1` into the format of `J2` and then call the same algorithm. Maybe something gets lost in translation also here in a weird way? 

CC: @wdecker

afkafkafk13:
I am not 100% sure, but I seem to remember that minAssGTZ in Singular passes through facstd under certain (not so rare) circumstances, whereas primdecGTZ never does (but it uses factorization at some other point). If I am right, it makes sense that minAssGTZ does not profit from the change like primdecGTZ does. 

For the differences between J1 and J2 using old_minimal_primes, my suspicion is that internally in primdec.lib some heuristic makes a different decision on the suitable route to take.

fingolfin:
What's the status of this? My understanding is that this PR might fix issue #3083 (does it?).

afkafkafk13:
@fingolfin : This PR does not resolve #3083 , but it investigates it. It seems to have been used for testing, but not been taken anywhere after that. It is @HechtiDerLachs 's call whether the points, for which this PR was created, have been discussed with @wdecker and @hannes14 . If so, the PR can be closed; if not, we should come back to it.

simonbrandhorst:
@HechtiDerLachs  is this still relevant? Can it be closed?

HechtiDerLachs:
This is still relevant, but nobody has the time and/or energy to really look into this.

PR #3119, opened 2023-12-18
[FTheoryTools] New constructor for literature models

HereAround:
New constructor for literature models. (The old constructor is of course still there, but this here is supposed to make things a bit more convenient if one uses the function over and over again, just as I did more recently.)


apturner (approved):
Thanks, @HereAround. I think this is fine for now. I am concerned in the long run that this just orders the models by their sort order in the directory. This means that adding new models can change the index of many of the existing models, as can a change in our internal naming scheme. Ultimately, it would good to add an index parameter to the data of the models themselves so the ordering is fixed once and for all.

HereAround:
> Thanks, @HereAround. I think this is fine for now. I am concerned in the long run that this just orders the models by their sort order in the directory. This means that adding new models can change the index of many of the existing models, as can a change in our internal naming scheme. Ultimately, it would good to add an index parameter to the data of the models themselves so the ordering is fixed once and for all.

Good point. I think it makes sense to fix this here and now. I will modify this PR accordingly (in the hopefully not too distant future).

HereAround:
@apturner I believe the changes should reflect your comment above. Can you confirm?

@thofma There seems to be an error rooted in `AbstractAlgebra`, leading to failure in the `deploy_docs` run. Ideas?

thofma:
Looks related to some recent changes by @fingolfin? If I remember correctly

apturner (approved):
Thanks @HereAround, looks good!

thofma:
The change was here: https://github.com/Nemocas/AbstractAlgebra.jl/pull/1523. @fingolfin maybe there was a reason most of the things are not `@docs` blocks in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1523/files#diff-443bf5c6d07f53397f17f97f5aa8f2808dc0cbe1a62ba1a90c768d1311579321?

thofma:
Max fixed it and I restarted the job here.

HereAround:
> Max fixed it and I restarted the job here.

Great. Thank you!

Issue #2835, opened 2023-09-21
`subgroups` command


fingolfin:
@wdecker just stumbled over this usability issue.

In short, the `subgroups` command is confusing because it does different things for "GAP groups" and abelian groups.

Consider:
```
help?> subgroups
search: subgroups psubgroups subgroup_reps hall_subgroups_representatives normal_subgroups minimal_subgroups

   subgroups(G::GrpGen; order::Int = 0,
                            index::Int = 0,
                            normal::Bool = false,
                            conjugacy_classes::Bool = false)

  Returns subgroups of G with embedding.

  ─────────────────────────────────────────────────────────────────

  subgroups(g::GrpAbFinGen;
            subtype = :all ,
            quotype = :all,
            index = -1,
            order = -1)

  Return an iterator for the subgroups of G of the specific form.

  ─────────────────────────────────────────────────────────────────

  subgroups(G::Group)

  Return the vector of all subgroups of G.
```

While technically the signatures suggest that not every group type supports all, cleaning this requires knowledge of what `GrpAbFinGen` and `GrpGen` are and are not.

The other issues is that the former two methods return iterators, where each entry is a subgroup *together with an embedding*, while the latter just returns a vector of subgroups (which technically is also iterable; but the "embedding" bit is missing). 

Of course the two first methods are also inconsistent with what they support and do not support, and don't document the meaning of the various kwargs...

One way to improve this would be to at least somewhat align these, i.e.,
- teach `subgroups` for GAPGroups to also at least suppor the `index` and `order` kwargs
- maybe also `conjugacy_classes` and `normal` (after figuring out what they really do -- I have guesses but details matter) 
- return pairs of group, embedding after all (???) -- though I'd prefer if we instead finally implemented the `canonical_map` proposal which allows retrieving these later if needed (but of course I am heavily influenced here because for "my" groups and applications, I almost never need them).

Implementation could just be naive, i.e. "compute all subgroups and discard unwanted ones" -- hey, not ideal, but a start.

In any case, the GAP Groups docstring for `subgroups` should also contain a warning that suggests that in many cases it is better to use `conjugacy_classes_subgroups`.

A radical alternative would be to change the `subgroups` method for GAPGroups to take arbitrary kwargs, and then just do `error("not supported for non-abelian groups, use XYZ instead"`).

fingolfin:
Discussed this with @ThomasBreuer today. Since all those extra arguments are undocumented, we decided there is nothing to be done here right now.

[...]

PR #3121, opened 2023-12-19
Change `on_subgroups` action to use the GAP interface

StevellM:
To work with (orbits and) stabilizers of subgroups in `TorQuadModule`, I implemented a very naive approach which was quite expensive, using action on lattices. Since the automorphisms for `TorQuadModule` come from GAP, we change the definition of the action to use the GAP action (I do not have the details here, but on relevant examples we are faster by a factor 8). I also needed `is_conjugate_with_data` for the action on subgroups.

As a remark: a submodule of a `TorQuadModule` do not "remember its parent" (sad). So to work with actions on submodules (on the GAP side), we need to remember the embedding of the underlying subgroup. This is why the `stabilizer` and `is_conjugate_with_data` methods take homomorphism(s) as input(s).

As for the general group interface, I added the action `on_subgroups`: I am ok with suggestions to modify it/change the name, as long as it allows me to make my stabilizer computations faster.

I also deleted what I thought would be useful for incoming new features (i.e. `compute_stab`): it is not used at the end so better removing it. Note that what I have removed was not exported.


experimental/QuadFormAndIsom/src/embeddings.jl
```diff
@@ -267,17 +246,17 @@ function _subgroups_orbit_representatives_and_stabilizers(Vinq::TorQuadModuleMor
     filter!(s -> is_invariant(fV, s[2]), subs)
   end
 
-  subs = TorQuadModule[s[1] for s in subs]
-  m = gset(O, _on_subgroups, subs)
+  to_gap = get_attribute(O, :to_gap)
+  to_oscar = get_attribute(O, :to_oscar)
+
+  qgap = codomain(to_gap)
+  sgap = typeof(qgap)[sub(qgap, elem_type(qgap)[to_gap(q(lift(s[2](a)))) for a in gens(s[1])])[1] for s in subs]
```
simonbrandhorst, line 253: Better create the subgroups directly in gap and only translate representatives to Oscar. 
But in any case the number of subgroups is usually to large to put them in memory.

simonbrandhorst (approved):

experimental/QuadFormAndIsom/src/embeddings.jl
StevellM: I am not sure how to keep track on the form on the GAP side but yes, there are still some possible improvements to be made here. I guess that in practice, this part of the code will be called on small enough examples where we should not have too many (stable) subgroups.

simonbrandhorst: Do you need the form in the orbit/stabliser computation?

simonbrandhorst:
```

    subs = collect(submodules(V; order = ord))
    filter!(s -> is_invariant(fV, s[2]), subs)
```
does not see the form

StevellM: Ah no right indeed, I forgot that the submodules are created as subgroups which inherit the form from the bigger module. So indeed yes, we could do everything on the GAP side. I can try it now and see to make a change in this PR. Thanks!

StevellM: I actually I do not know if it is better. On the GAP side, from what I see, we cannot enumerate subgroups of a given order, or stable subgroups. Which is what we use here in Hecke. I am afraid that on the GAP side we will have to many subgroups: ok we can filter them anyway but if our group has hundreds of subgroups but only 1 of the order we want, Hecke will do things better here (see https://github.com/oscar-system/Oscar.jl/issues/2835)

simonbrandhorst: o.k.

PR #3122, opened 2023-12-19
Random subquotient module elements and homomorphisms

jankoboehm:
Random homogeneous elements of modules, in particular random homomorphisms. Limited to finite base rings, for the moment. What is here already does what we need currently for certain constructions in algebraic geometry. We will expand on it further.

### DIFF
--- a/src/Combinatorics/Compositions.jl
+++ b/src/Combinatorics/Compositions.jl
@@ -1,7 +1,7 @@
 
 function weak_compositions(n, k)
     if k == 0
-        return n == 0 ? [Int[]] : []
+        return n == 0 ? [Int[]] : Vector{Int}[]
     end
     compos = Vector{Int}[]
     for i in 0:n
@@ -14,7 +14,7 @@ end
 
 function compositions(n, k)
     if k == 0
-        return n == 0 ? [Int[]] : []
+        return n == 0 ? [Int[]] : Vector{Int}[]
     end
     if n < k
         return []
--- a/src/Combinatorics/Compositions.jl
+++ b/src/Combinatorics/Compositions.jl
@@ -1,0 +1,29 @@
+
+function weak_compositions(n, k)
+    if k == 0
+        return n == 0 ? [Int[]] : []
+    end
+    compos = Vector{Int}[]
+    for i in 0:n
+        for p in weak_compositions(n - i, k - 1)
+            push!(compos, [i; p])
+        end
+    end
+    return compos
+end
+
+function compositions(n, k)
+    if k == 0
+        return n == 0 ? [Int[]] : []
+    end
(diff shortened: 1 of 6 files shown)


src/Modules/ModulesGraded.jl
```diff
@@ -2882,3 +2882,51 @@ function _regularity_bound(M::SubquoModule)
...
+        return n == 0 ? [Int[]] : []
+    end
+    partitions = Vector{Int}[]
+    for i in 0:n
+        for p in all_partitions_in(n - i, k - 1)
+            push!(partitions, [i; p])
+        end
+    end
+    return partitions
+end
```
lgoettgens: This should not live here, but in `src/Combinatorics` as it can be applied elsewhere. One day, it can then be replaced by the implementation in `experimental/JuLie/src/partitions.jl`.

jankoboehm: Yes, I don't mind to shift it and make it available. We will also expand on it. As far as I understood JuLi does partitiions in the sorted sense. This here is what is usually called compositions. So I renamed it.

src/Combinatorics/Compositions.jl
```diff
@@ -0,0 +1,29 @@
+
+function weak_compositions(n, k)
+    if k == 0
+        return n == 0 ? [Int[]] : []
```
thofma:
```suggestion
        return n == 0 ? [Int[]] : Vector{Int}[]
```

src/Combinatorics/Compositions.jl
```diff
@@ -0,0 +1,29 @@
...
+        for p in weak_compositions(n - i, k - 1)
+            push!(compos, [i; p])
+        end
+    end
+    return compos
+end
+
+function compositions(n, k)
+    if k == 0
+        return n == 0 ? [Int[]] : []
```
thofma:
```suggestion
        return n == 0 ? [Int[]] : Vector{Int}[]
```

src/Modules/ModulesGraded.jl
lgoettgens: Thanks for moving!
However, I think that this should either include a docstring as well, or shouldn't get exported (or maybe even both for now).

lgoettgens: For example, changing this later to ZZRingElem instead of Int would be an unnecessary breaking change, so I would rather keep this internal until an interface has been settled. (With talking to people interested in combinatorics like Uli Thiel)

wdecker (approved):

thofma:
Would have been better to not export this. Not sure this is compatible with the rest of the combinatorics.

thofma:
It also has no docstring.

PR #3129, opened 2023-12-20
Fiber products

HechtiDerLachs:
Needed in particular for the `base_change`s while searching for the famous automorphism.


HechtiDerLachs:
Two things remain to be done:
 1) The glueings need to be made lazy
 2) Various `check=true` need to be set to `check=false`. 
But for the interest of the reviewers, I would leave the code as is for the moment, since wrapping things up in `LazyGlueing`s will make the code even more confusing.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+  B = domain(g)
+  C = codomain(f)
+  h = complement_equations(g)
+  pbh = pullback(f).(h)
+  result = PrincipalOpenSubset(A, pbh)
+  gg = PrincipalOpenEmbedding(SpecMor(result, A, gens(OO(result)), check=false), pbh, check=false)
+  g_res_inv = inverse_on_image(g)
+  ff = compose(restrict(f, result, image(g), check=false), g_res_inv)
+  return result, gg, ff
+end
```
simonbrandhorst:
```suggestion
function fiber_product(f::AbsSpecMor, g::PrincipalOpenEmbedding)
  result, gg, ff == fiber_product(g, f)
  return result, ff, gg
end
```

simonbrandhorst: avoids code duplication

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+In a commutative diagram 
+          b
+   W ------------\
+   |             |
+   |             V
+  a|    X x Y -->Y
+   |      |      | g
+   |      V      V
+   \----->X----> Z
+             f
```
simonbrandhorst: Does this look readable in the documentation? Perhaps enclose in triple backticks.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+this computes the canonical map `W -> XxY`.
```
simonbrandhorst:
```suggestion
this computes the canonical map `W -> X x Y`.
```

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+    fiber_product::Tuple{<:AbsSpec, <:AbsSpecMor, <:AbsSpecMor}=fiber_product(f, g),
+    check::Bool=true
+  )
+  X = domain(f)
+  Y = domain(g)
+  Z = codomain(f)
+  @assert codomain(g) === Z
+  XxY = fiber_product[1]
+  gg = fiber_product[2]
+  ff = fiber_product[3]
```
simonbrandhorst, line 110:
```suggestion
# gg: X x Y ->  X
# ff: X x Y -> Y
  gg = fiber_product[2]
  ff = fiber_product[3]
```

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+  @assert codomain(b) === Y
+  @check compose(a, f) == compose(b, g) "maps do not commute"
+  @check compose(ff, g) == compose(gg, f) "maps do not commute"
+
+  # The ambient scheme of XxY is the actual product of X and Y 
+  # over Spec(k), the coefficient ring. If it is not, then 
+  # this is due to special dispatch which has to also be caught 
+  # with a special method for this function here.
+  img_gens = vcat(pullback(a).(gens(OO(X))), pullback(b).(gens(OO(Y))))
+  return SpecMor(W, XxY, img_gens, check=check)
```
simonbrandhorst: Can you add a check & a useful error message for this?
You seem to assume a certain order of the generators of `XxY` here can we be certain of this? / is it guaranteed?
add a check like 
`@check gens(OO(XxY)) == vcat(pullback(gg).(gens(OO(X)),pullback(ff).(gens(OO(Y))))`?

HechtiDerLachs:
@thofma : Could you have a look at the failing doctests? I don't see what that has to do with any of my changes. I didn't touch any `factor` routine. Thanks!

benlorenz:
> @thofma : Could you have a look at the failing doctests? I don't see what that has to do with any of my changes. I didn't touch any `factor` routine. Thanks!

That was fixed in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1541. I restarted the job and it should turn green soon.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+  ff = fiber_product[3]
+  @assert codomain(ff) === Y
+  @assert codomain(gg) === X
+
+  W = domain(a)
+  @assert W === domain(b)
+  @assert codomain(a) === X
+  @assert codomain(b) === Y
+  @check compose(a, f) == compose(b, g) "maps do not commute"
+  @check compose(ff, g) == compose(gg, f) "maps do not commute"
```
simonbrandhorst: There is a lot of copy paste code duplication here. 
How about moving all the checks to a general method and then resume with a
`_induced_map_to_fiber_product` 
which has a more special signature and does not check anything whatsoever.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+  img_gens = pullback(b).(gens(OO(Y)))
+  return SpecMor(W, XxY, img_gens, check=check)
+end
+
+### Some helper functions
+function _restrict_domain(f::AbsSpecMor, D::PrincipalOpenSubset; check::Bool=true)
+  D === domain(f) && return f
+  ambient_scheme(D) === domain(f) && return SpecMor(D, codomain(f), OO(D).(pullback(f).(gens(OO(codomain(f))))), check=false)
+  @check is_subset(D, domain(f)) "domain incompatible"
+  return SpecMor(D, codomain(f), OO(D).(pullback(f).(gens(OO(codomain(f))))), check=check)
```
simonbrandhorst, line 196:
```suggestion
  return SpecMor(D, codomain(f), OO(D).(pullback(f).(gens(OO(codomain(f))))), check=false)
```
?

simonbrandhorst: can we avoid the allocations happening in `pullback(f).(gens(OO(codomain(f))))` 
by using something like `img_gens` here?

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
```diff
@@ -25,7 +26,240 @@ function fiber_product(
...
+end
+
+function _restrict_codomain(f::AbsSpecMor, D::PrincipalOpenSubset; check::Bool=true)
+  D === codomain(f) && return f
+  if ambient_scheme(D) === codomain(f) 
+    @check is_unit(pullback(f)(complement_equation(D))) "complement equation does not pull back to a unit"
+    return SpecMor(domain(f), D, OO(domain(f)).(pullback(f).(gens(OO(codomain(f))))), check=false)
+  end
+  @check is_subset(D, codomain(f)) "codomain incompatible"
+  @check is_subset(domain(f), preimage(f, D))
```
simonbrandhorst, line 212: Shouldn't we check that  `D` contains the image of `f` instead?

simonbrandhorst: (and likewisein the other methods)

simonbrandhorst: I would probably also suggest separating the computational kernels and the checks?

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Constructors.jl
```diff
@@ -361,6 +361,15 @@ function subscheme(X::AbsSpec, I::Ideal)
...
 
+function sub(X::AbsSpec, a)
+  (a isa RingElem && parent(a) === OO(X)) || return sub(X, OO(X)(a))
+  return sub(X, ideal(OO(X), a))
+end
+
+function sub(X::AbsSpec, a::Vector)
+  all(a->a isa RingElem && parent(a) === OO(X), a) || return sub(X, OO(X).(a))
+  return sub(X, ideal(OO(X), a))
+end
```
simonbrandhorst, line 372: Do we want to support this in general? Then we should perhaps remove the subscheme method all together
and replace it by sub? It seems to be more consistent with the oscar world over all.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
```diff
@@ -159,3 +193,100 @@ function Base.show(io::IO, ::MIME"text/plain", f::AbsCoveredSchemeMorphism)
...
+    # AxCAxCB-----> CAxCB -------> CB
+    # |              |             |
+    # |              |            inc_B
+    # |              |             |
+    # V              V             V
+    # A ---f_cov---> CA ---inc_A-> C 
+    #
+    # TODO: Maybe it's easier to first compose the maps 
+    # on the boundary and do only one square? We should think 
+    # about it!
```
simonbrandhorst, line 278: It would avoid the creation of two morphisms in the end so why not?
After all you compose everything anyways in the end, hence there is no benefit of keeping a factored form?

src/AlgebraicGeometry/Schemes/PrincipalOpenInclusion/Constructors.jl
```diff
@@ -0,0 +1,16 @@
+function complement(X::AbsSpec, h::Vector)
```
simonbrandhorst, line 1: I seem to recall that we have the function `hypersurface_complement` which does what your `complement` does?

src/AlgebraicGeometry/Schemes/PrincipalOpenSubset/Objects/Attributes.jl
```diff
@@ -27,7 +27,7 @@ function inclusion_morphism(U::PrincipalOpenSubset; check::Bool=false)
   if !isdefined(U, :inc)
     X = ambient_scheme(U)
     inc = SpecMor(U, X, hom(OO(X), OO(U), gens(OO(U)), check=check), check=check)
-    U.inc = OpenInclusion(inc, ideal(OO(X), complement_equation(U)), check=check)
+    U.inc = PrincipalOpenEmbedding(inc, [complement_equation(U)], check=check)
```
simonbrandhorst, line 30: It seems that `PrincipalOpenEmbedding` somewhat competes with `OpenInclusion`?
Can this lead to problems later?

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2289,3 +2289,180 @@ function is_homogeneous(a::MPolyQuoLocRingElem{<:Ring, <:RingElem, <:MPolyDecRin
   return is_homogeneous(numerator(a)) && is_homogeneous(denominator(a))
 end
 
+########################################################################
+# Inverses of homomorphisms                                            #
+########################################################################
+
+function inverse(
+    f::Map{<:Union{<:MPolyRing, <:MPolyQuoRing, <:MPolyLocRing, <:MPolyQuoLocRing},
+           <:Union{<:MPolyRing, <:MPolyQuoRing, <:MPolyLocRing, <:MPolyQuoLocRing}
```
simonbrandhorst:
```suggestion
    f::Map{<:MPolyAnyRing, <:MPolyAnyRing}
```

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
```diff
@@ -35,24 +35,58 @@ function compose(f::AbsCoveredSchemeMorphism, g::AbsCoveredSchemeMorphism)
...
-    ref_mor = Y[codomain(cf), domain(cg)] # the refinement `CoveringMorphism`
-    mor_dict = IdDict{AbsSpec, AbsSpecMor}() # TODO: Keep the type of the morphisms?
-    for U in basic_patches(domain(cf))
-      V = codomain(cf[U])
-      W = codomain(ref_mor[V])
-      mor_dict[U] = compose(compose(cf[U], ref_mor[V]), cg[W])
-    end
-    cc = CoveringMorphism(domain(cf), codomain(cg), mor_dict, check=false)
-    return CoveredSchemeMorphism(X, Z, cc)
+    # The most complicated case:
```
simonbrandhorst, line 69: this is not hit by a test. Can you cook one up?

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
simonbrandhorst: The more complicated cases are not hit by tests. Could you cook one up artificially?

simonbrandhorst (commented):
Nice piece of work! I appreciate the comments in the code.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
HechtiDerLachs: Makes sense.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
HechtiDerLachs: That would be nice. But there is no such common getter as of now. We should probably discuss doing this, but not here.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
HechtiDerLachs: > Shouldn't we check that `D` contains the image of `f` instead?

That would be the full mathematical requirement, yes. But that check can not easily be realized. We can only compute the closure of the image. But that can not be used here. So what we really do is just a plausibility check for the moment, a necessary condition. 

I do not fully understand your last comment.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Constructors.jl
HechtiDerLachs: No. `subscheme` stays. But we do want to support the `sub` method more generally. In most instances this will then use `subscheme` in its internals.

src/AlgebraicGeometry/Schemes/PrincipalOpenInclusion/Constructors.jl
HechtiDerLachs: This is similar to `subscheme` vs. `sub`: One returns a map, the other doesn't.

src/AlgebraicGeometry/Schemes/PrincipalOpenSubset/Objects/Attributes.jl
HechtiDerLachs: `OpenInclusion` does not require the image to be a principal open subset. I do not see how this entails conflicts, no.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
HechtiDerLachs: It didn't.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Morphisms/Methods.jl
HechtiDerLachs: I think, I achieved this. Let's see what test coverage says now.

HechtiDerLachs:
All required changes addressed. Let's see how the tests are doing now.

HechtiDerLachs:
Ready for merge from my side.

simonbrandhorst (approved):

PR #3131, opened 2023-12-21
feat: Groebner Walk

welpj:
Hello @ederc ,
I finished the first step of the migration of the Groebner Walk to the Oscar.jl library and already talked to @afkafkafk13 about it. 

This PR includes:
- Gröbner Walk
- Perturbed Gröbner Walk
- Fractal Walk
- Generic Walk
- Tran´s version of the Groebner Walk

There probably are several improvements necessary, for example in the use of IdealGens: I don't know the handling of IdealGens in detail, @jankoboehm do you have any hints for improvements?

At the moment the tests take less than two minutes, but I was thinking of adding another bigger Ideal for testing the whole algorithms. @fingolfin, would that be appropriate? Are there further limits on RAM consumption?

I'm happy for any hint to beautify the code.

Thank you 
Jordi

### DIFF
--- a/src/Rings/walk.jl
+++ b/src/Rings/walk.jl
@@ -1,6 +1,6 @@
 using Oscar
 
-global infoLevel = 1
+global infoLevel = 0
 
 ###############################################################
 # Implementation of different variants of the Groebner Walk.
@@ -8,9 +8,9 @@ global infoLevel = 1
 ###############################################################
 @doc raw"""
 		groebnerwalk(
-			I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I)),
-			startOrder::Symbol = :degrevlex,
-			targetOrder::Symbol = :lex,
+			I::MPolyIdeal; 
+			startOrder::MonomialOrdering = default_ordering(base_ring(I)),
+			targetOrder::Symbol = lex(base_ring(I)),
 			walktype::Symbol = :standard,
 			perturbationDegree::Int = 2,
 		)
@@ -25,7 +25,7 @@ Fractal Walk (:fractalcombined) computes the Walk like it´s presented in Amrhei
 Fractal Walk (:fractal) computes the Walk like it´s presented in Amrhein & Gloor (1998). Pertubes only the target vector.
 Set infoLevel to trace the walk:
 	-'infoLevel=0': no detailed printout,
-	-'infoLevel=1': intermediate weight vectors,
+	-'infoLevel=1': adds intermediate weight vectors,
 	-'infoLevel=2': adds information about the Groebner basis.
 
 #Arguments
@@ -40,10 +40,48 @@ Set infoLevel to trace the walk:
 	- `fractal`: standard-version of the Fractal Walk,
 	- `fractalcombined`: combined Version of the Fractal Walk. Target monomial order needs to be lex,
 - `perturbationDegree::Int=2`: perturbationdegree for the perturbed Walk.
+
+
+# Examples
+
+```jldoctest
+julia> R,(x,y) = polynomial_ring(QQ, ["x","y"], ordering = :degrevlex);
+
+julia> I = ideal([y^4+ x^3-x^2+x,x^4]);
+
+julia> groebnerwalk(I, degrevlex(R), lex(R), :standard)
+standard_walk results
+Crossed Cones in: 
+[4, 3]
+[4, 1]
+[12, 1]
+[1, 0]
+Cones crossed: 4
+Gröbner basis with elements
+1 -> y^16
+2 -> x + y^12 - y^8 + y^4
+with respect to the ordering
+matrix_ordering([x, y], [1 0; 0 1])
+
+julia> groebnerwalk(I, degrevlex(R), lex(R), :perturbed, 2)
+perturbed_walk results
+Crossed Cones in: 
+[4, 3]
+[4, 1]
+[5, 1]
+[12, 1]
+[1, 0]
+Cones crossed: 5
+Gröbner basis with elements
+1 -> y^16
+2 -> x + y^12 - y^8 + y^4
+with respect to the ordering
+matrix_ordering([x, y], [1 0; 0 1])
+```
 """
 function groebnerwalk(
 	I::MPolyIdeal, startOrder::MonomialOrdering = default_ordering(base_ring(I)),
-	targetOrder::MonomialOrdering = :lex,
+	targetOrder::MonomialOrdering = lex(base_ring(I)),
 	walktype::Symbol = :standard,
 	perturbationDegree::Int = 2)
 	S = canonical_matrix(startOrder)
@@ -74,7 +112,7 @@ Fractal Walk (:fractal) computes the Walk like it´s presented in Amrhein & Gloo
 Fractal Walk (:fractalcombined) computes the Walk like it´s presented in Amrhein & Gloor (1998) with multiple extensions. The target monomial order has to be lex. This version uses the Buchberger Algorithm to skip weightvectors with entries bigger than Int32.
 Set infoLevel to trace the walk:
 	-'infoLevel=0': no detailed printout,
-	-'infoLevel=1': intermediate weight vectors,
+	-'infoLevel=1': adds intermediate weight vectors,
 	-'infoLevel=2': adds information about the Groebner basis.
 
 #Arguments
@@ -89,6 +127,43 @@ Set infoLevel to trace the walk:
 	- `fractal`: standard-version of the Fractal Walk,
 	- `fractalcombined`: combined version of the Fractal Walk. The target monomial order needs to be lex,
 -`p::Int=2`: perturbationdegree for the perturbed Walk.
+
+# Examples
+
+```jldoctest
+julia> R,(x,y) = polynomial_ring(QQ, ["x","y"], ordering = :degrevlex);
+
+julia> I = ideal([y^4+ x^3-x^2+x,x^4]);
+
+julia> groebnerwalk(I, [1 1; 0 -1], [1 0; 0 1], :standard)
+standard_walk results
+Crossed Cones in: 
+[4, 3]
+[4, 1]
+[12, 1]
+[1, 0]
+Cones crossed: 4
+Gröbner basis with elements
+1 -> y^16
+2 -> x + y^12 - y^8 + y^4
+with respect to the ordering
+matrix_ordering([x, y], [1 0; 0 1])
+
+julia> groebnerwalk(I, [1 1; 0 -1], [1 0; 0 1], :perturbed, 2)
+perturbed_walk results
+Crossed Cones in: 
+[4, 3]
+[4, 1]
+[5, 1]
+[12, 1]
+[1, 0]
+Cones crossed: 5
+Gröbner basis with elements
+1 -> y^16
+2 -> x + y^12 - y^8 + y^4
+with respect to the ordering
+matrix_ordering([x, y], [1 0; 0 1])
+```
 """
 function groebnerwalk(
 	G::Union{Oscar.IdealGens, MPolyIdeal},
@@ -174,17 +249,17 @@ function standard_walk(
 	tarweight::Vector{Int})
 	while true
 		G = standard_step(G, currweight, T)
-		if infoLevel >= 1
-			println(currweight)
-			if infoLevel == 2
-				println(G)
-			end
-		end
-		raise_step_counter()
 		if currweight == tarweight
 			return G
 		else
 			currweight = next_weight(G, currweight, tarweight)
+		end
+		if infoLevel >= 1
+			raise_step_counter()
+			println(currweight)
+			if infoLevel == 2
+				println(G)
+			end
 		end
 	end
 end
@@ -227,7 +302,7 @@ function generic_walk(
 	ordNew = matrix_ordering(base_ring(G), T)
 	if infoLevel >= 1
 		println("generic_walk results")
-		println("Crossed Cones with: ")
+		println("Facets crossed for: ")
 	end
 	while !isempty(v)
 		G, Lm = generic_step(G, Lm, v, ordNew)
--- a/src/Rings/walk.jl
+++ b/src/Rings/walk.jl
@@ -1,0 +1,1649 @@
+using Oscar
+
+global infoLevel = 1
+
+###############################################################
+# Implementation of different variants of the Groebner Walk.
+# The Groebner Walk is proposed by Collart, Kalkbrener & Mall (1997).
+###############################################################
+@doc raw"""
+		groebnerwalk(
+			I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I)),
+			startOrder::Symbol = :degrevlex,
+			targetOrder::Symbol = :lex,
+			walktype::Symbol = :standard,
+			perturbationDegree::Int = 2,
+		)
+Compute a reduced Groebner basis w.r.t. to a monomial order by converting it using the Groebner Walk.
+The Groebner Walk is proposed by Collart, Kalkbrener & Mall (1997).
+One can choose a strategy of:
+Standard Walk (:standard) computes the Walk like it´s presented in Cox, Little & O´Shea (2005).
+Generic Walk (:generic) computes the Walk like it´s presented in Fukuda, Jensen, Lauritzen & Thomas (2005).
+Perturbed Walk (:perturbed, with p = degree of the perturbation) computes the Walk like it´s presented in Amrhein, Gloor & Küchlin (1997).
+Tran´s Walk (:tran) computes the Walk like it´s presented in Tran (2000).
+Fractal Walk (:fractalcombined) computes the Walk like it´s presented in Amrhein & Gloor (1998) with multiple extensions. The target monomial order has to be lex. This version uses the Buchberger Algorithm to skip weight vectors with entries bigger than Int32.
+Fractal Walk (:fractal) computes the Walk like it´s presented in Amrhein & Gloor (1998). Pertubes only the target vector.
+Set infoLevel to trace the walk:
+	-'infoLevel=0': no detailed printout,
+	-'infoLevel=1': intermediate weight vectors,
+	-'infoLevel=2': adds information about the Groebner basis.
+
+#Arguments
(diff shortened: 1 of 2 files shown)


src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
+using Oscar
```
fingolfin, line 1: This is not the way to go about it if you want to integrate this code into Oscar. Instead, add the code into a subdirectory of `experimental` -- please read <https://docs.oscar-system.org/dev/Experimental/intro/> for details.

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+
+global infoLevel = 0
```
fingolfin, line 3: A global like this is a no-go. 

An alternative would be to use `@vprint` and friends in your code, see <https://www.thofma.com/Hecke.jl/stable/features/macros/#verbosity-macros> for details (and also you can grep through our code to see examples using it).

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+# The Groebner Walk is proposed by Collart, Kalkbrener & Mall (1997).
+###############################################################
+@doc raw"""
+		groebnerwalk(
+			I::MPolyIdeal; 
+			startOrder::MonomialOrdering = default_ordering(base_ring(I)),
+			targetOrder::Symbol = lex(base_ring(I)),
+			walktype::Symbol = :standard,
+			perturbationDegree::Int = 2,
+		)
```
fingolfin, line 16: Should be indented by four spaces, not tabs

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+Compute a reduced Groebner basis w.r.t. to a monomial order by converting it using the Groebner Walk.
```
fingolfin, line 17: There must be an empty line between the function signatures and the text documenting it.

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+The Groebner Walk is proposed by Collart, Kalkbrener & Mall (1997).
+One can choose a strategy of:
+Standard Walk (:standard) computes the Walk like it´s presented in Cox, Little & O´Shea (2005).
+Generic Walk (:generic) computes the Walk like it´s presented in Fukuda, Jensen, Lauritzen & Thomas (2005).
+Perturbed Walk (:perturbed, with p = degree of the perturbation) computes the Walk like it´s presented in Amrhein, Gloor & Küchlin (1997).
+Tran´s Walk (:tran) computes the Walk like it´s presented in Tran (2000).
+Fractal Walk (:fractalcombined) computes the Walk like it´s presented in Amrhein & Gloor (1998) with multiple extensions. The target monomial order has to be lex. This version uses the Buchberger Algorithm to skip weight vectors with entries bigger than Int32.
+Fractal Walk (:fractal) computes the Walk like it´s presented in Amrhein & Gloor (1998). Pertubes only the target vector.
```
fingolfin, line 25:
```suggestion
Standard Walk (:standard) computes the Walk like it´s presented in Cox, Little & O´Shea (2005).
- Generic Walk (:generic) computes the Walk as presented in Fukuda, Jensen, Lauritzen & Thomas (2005).
- Perturbed Walk (:perturbed, with p = degree of the perturbation) computes the Walk ass presented in Amrhein, Gloor & Küchlin (1997).
- Tran's Walk (:tran) computes the Walk like as presented in Tran (2000).
- Fractal Walk (:fractalcombined) computes the Walk like as presented in Amrhein & Gloor (1998) with multiple extensions. The target monomial order has to be lex. This version uses the Buchberger algorithm to skip weight vectors with entries bigger than Int32.
- Fractal Walk (:fractal) computes the Walk as presented in Amrhein & Gloor (1998). Perturbs only the target vector.
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	if infoLevel >= 1
+		println("Cones crossed: ", delete_step_counter())
+	end
+	return Oscar.IdealGens(gens(Gb), matrix_ordering(base_ring(G), T), isGB = true)
+end
+
+###########################################
+# Counter for the steps
+###########################################
+counter = 0
```
fingolfin, line 211: Using a global like this is not really acceptable

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+end
+
+@doc raw"""
+		groebnerwalk(
+			G::Oscar.IdealGens,
+			S::Union{Matrix{T}, MatElem{T}},
+			T::Union{Matrix{T}, MatElem{T}},
+			walktype::Symbol = :standard,
+			p::Int = 2
+		) where T
```
fingolfin, line 104: In this signature `T` is both a type and a variable name...

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+Computes a reduced Groebner basis w.r.t. the monomial order T by converting the reduced Groebner basis G w.r.t. the monomial order S using the Groebner Walk.
```
fingolfin, line 105: `T` is not a monomial order, it is a matrix. Maybe that matrix is supposed to describe a monomial order? Then state it as such?

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+One can choose a strategy of:
+Standard Walk (:standard) computes the Walk like it´s presented in Cox et al. (2005).
+Generic Walk (:generic) computes the Walk like it´s presented in Fukuda et al. (2005).
+Perturbed Walk (:perturbed, with p = degree of the perturbation) computes the Walk like it´s presented in Amrhein et al. (1997).
+Tran´s Walk (:tran) computes the Walk like it´s presented in Tran (2000).
+Fractal Walk (:fractal) computes the Walk like it´s presented in Amrhein & Gloor (1998). Pertubes only the target vector.
+Fractal Walk (:fractalcombined) computes the Walk like it´s presented in Amrhein & Gloor (1998) with multiple extensions. The target monomial order has to be lex. This version uses the Buchberger Algorithm to skip weightvectors with entries bigger than Int32.
```
fingolfin, line 112: This seems to duplicate most of the previous docstring. That makes it harder to maintain (any tweaks need to be applied to both, otherwise they get out of sync and things get very confusing).

I would recommend merging the two docstrings; a docstring can have multiple signatures at the start. 

For this to make sense, you should align parameter names as much as possible, so e.g. `p` should be `perturbationDegree`.

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+###############################################################
+# The Fractal Walk
+###############################################################
+
+##########################################
+# global weightvectors
+##########################################
+pTargetWeights = []
+pStartWeights = []
+firstStepMode = false
```
fingolfin, line 376: No, this is not acceptable.

As an alternative, You could put all the global state you need into a new mutable struct type and pass an instance of that around between your functions.

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	else
+		return [0]
+	end
+end
+
+# returns the next t to compute the next weight vector w(t) = w + t * (tw - w) like it´s done in Cox, Little & O'Sheao (2005).
+function next_weight_fractal_walk(
+	G::Oscar.IdealGens,
+	cweight::Array{T, 1},
+	tweight::Array{K, 1},
```
fingolfin, line 954:
```suggestion
	cweight::Vector{T},
	tweight::Vector{K},
```
Any other use of `Array` should likely be removed, too

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+) where {T <: Number, K <: Number}
```
fingolfin, line 955: So which subtypes of `Number` do you expect here? Can I use floats? Are `T` and `K` really allowed to be of different types? What's the rationale here?

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	if (cweight == tweight)
```
fingolfin, line 956:
```suggestion
	if cweight == tweight
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+		return [0]
```
fingolfin, line 957: Not type stable: you return a `Rational{BigInt}` otherwise, here you return a `Vector{Int}`

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	end
+	tmin = 1
```
fingolfin, line 959: This is not type stable (you later set `tmin` to `cw // (cw - tw)` which presumably is a `Rational{SOMETHING}` ?)

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+			cw = dot(cweight, v)
+			t = cw // (cw - tw)
+			if t < tmin
+				tmin = t
+			end
+		end
+	end
+
+	# BigInt is needed to prevent overflows in the conversion of the weight vectors.
+	return BigInt(numerator(tmin)) // BigInt(denominator(tmin))
```
fingolfin, line 972: In OSCAR we normally don't use `BigInt`, we use `ZZRingElems` (typically created via `ZZ(some_value)`).
Of course we also wouldn't use a `Rational` either, but instead a `QQRingElem` ...

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+end
+
+# returns 'true' if the leading terms of G w.r.t the matrixordering T are the same as the leading terms of G w.r.t the weighted monomial ordering with weight vector of pvecs[p] (pvecs[p-1]) and the matrixordering T.
```
fingolfin, line 975: Please don't use lines more than ~100 characters
```suggestion
# returns 'true' if the leading terms of G w.r.t the matrix ordering T are the
# same as the leading terms of G w.r.t the weighted monomial ordering with
# weight vector of pvecs[p] (pvecs[p-1]) and the matrix ordering T.
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	G::Oscar.IdealGens,
+	T::Matrix{Int},
+	pvecs::Vector{Vector{Int}},
+	p::Int,
+)
+	# pvecs(1) equals the first row of T
+	if p == 1
+		return true
+	end
+	ord = matrix_ordering(base_ring(G), T)
```
fingolfin, line 986: Why do you pass `T` around as a matrix if you then convert it again into an ordering? IMHO the code would be much clearer if you passed around orderings: it would add type safety and make the code slightly more self-documenting, and would probably even optimize some things

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+end
+
+# returns the differences of the exponent vectors of the leading terms and the polynomials of the generators of I.
+function difference_lead_tail(
+	G::Oscar.IdealGens,
+	Lm::Vector{L},
+	T::Union{Matrix{N}, MatElem{N}},
+) where {L <: MPolyRingElem, N}
+	v = Vector{Int}[]
+	for i ∈ 1:length(G)
```
fingolfin, line 1045: Don't use unicode in code, please
```suggestion
	for i in 1:length(G)
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+			if ltu != e
+				push!(v, ltu .- e)
+			end
+		end
+	end
+	return unique!(v)
+end
+
+# returns true if the vector u is parallel to the vector v.
+function isparallel(u::Vector{Int}, v::Vector{Int})
```
fingolfin, line 1057: I would recommend spelling out what you mean by "parallel" -- here it seems that `u` must be a multiple of `v` (except that if `u` is a zero vector, but `v` is not, in your code they are not parallel; while if `v` is a zero vector and `u` is not, they *are* parallel...?!? Maybe `u` and `v` are never supposed to be zero vectors? Then state so. Also, I guess you expect them to have equal length, so perhaps add an `@assert that verifies this..)

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	count = 1
+	x = 0
```
fingolfin, line 1059: Again: not type stable, as later it is set to a `Rational{Int}`

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	return p
+end
+
+# this function interreduces the Groebner basis G w.r.t. the leading terms Lm with tail-reduction.
+function interreduce(
+	G::Vector{T},
+	Lm::Vector{T},
+	ord::MonomialOrdering,
+) where {T <: MPolyRingElem}
+	for i ∈ 1:Singular.length(G)
```
fingolfin, line 1245: Why `Singular.length` ?!

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+			end
+		end
+	end
+	return max
+end
+
+
+function checkInt32(w::Vector{Int})
+	for i ∈ 1:length(w)
+		if tryparse(Int32, string(w[i])) == nothing
```
fingolfin, line 1675: That's a very expensive way to check if something fits into an `Int32`... better:
```suggestion
		if w[i] % Int32 != w[i]
```

(You could also use `unsafe_trunc`, `mod` or `rem` for this)

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+			return false
+		end
+	end
+	return true
```
fingolfin, line 1679:
```suggestion
	return all(x -> x % Int32 == x, w)
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	for i ∈ 1:n
+		for j ∈ 1:n
+			temp = T[i, j]
+			if M < temp
+				M = temp
+			end
+		end
+	end
+	d0 = 0
+	for g in Singular.gens(G)
```
fingolfin, line 1695: Why `Singular.gens` and not `gens`?

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+		if tryparse(Int32, string(w[i])) == nothing
+			return false
+		end
+	end
+	return true
+end
+
+# computes the representation of the matrixorder defined by T.
+function representation_vector(G::Oscar.IdealGens, T::Matrix{Int})
+	n = size(T)[1]
```
fingolfin, line 1684:
```suggestion
	n = nrows(T)
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+			end
+		end
+	end
+	d0 = 0
+	for g in Singular.gens(G)
+		temp = deg(g, n)
+		if d0 < temp
+			d0 = temp
+		end
+	end
```
fingolfin, line 1700:
```suggestion
	d0 = maximum(g -> deg(g, n), gens(g))
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+	n = size(T)[1]
+	M = 0
+	for i ∈ 1:n
+		for j ∈ 1:n
+			temp = T[i, j]
+			if M < temp
+				M = temp
+			end
+		end
+	end
```
fingolfin, line 1693:
```suggestion
	M = maximum(T)
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+		temp = deg(g, n)
+		if d0 < temp
+			d0 = temp
+		end
+	end
+	d = M * (2 * d0^2 + (n + 1) * d0)
+	w = d^(n - 1) * T[1, :]
+	for i ∈ 2:n
+		w = w + d^(n - i) * T[i, :]
+	end
```
fingolfin, line 1705:
```suggestion
	w = sum(i -> d^(n - i) * T[i, :], 1:n)
```
Both versions are somewhat inefficient because they keep recomputing powers of `d`. Better would be something like this (could be slightly buggy but you should get the gist):
```suggestion
    w = T[n, :]
	e = one(d)
	for i in 1:n-1
	    e *= d
		w += e * T[n-i, :]
	end
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+		end
+		if size(collect(Singular.coefficients(g)))[1] == 2
+			counter = counter + 1
+		end
+	end
+	if counter > 1
+		return true
+	end
+	return false
+end
```
fingolfin, line 1724: File should end in a newline
```suggestion
end
```

src/Rings/walk.jl
```diff
@@ -0,0 +1,1724 @@
...
+# checks if Gw is an initialform of an ideal corresponding to a face of the Groebner fan with dimension less than n-1.
+function inSeveralCones(Gw::Vector{T}) where {T <: MPolyRingElem}
+	counter = 0
+	for g in Gw
+		if size(collect(Singular.coefficients(g)))[1] > 2
+			return true
+		end
+		if size(collect(Singular.coefficients(g)))[1] == 2
+			counter = counter + 1
+		end
```
fingolfin, line 1718: Don't compute things twice
```suggestion
		s = size(collect(Singular.coefficients(g)))[1]
		s > 2 && return true
		if s == 2
			counter += 1
		end
```

fingolfin (changes requested):
Thank you for your contribution. A bunch of quick comments added inline. Top level:
- Please do not use tabs at all (see also <https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/>).
- adding multi minute / multi GB tests that are run every time is very problematic. Aren't there easier examples? They don't need to be "mathematically interesting" (interesting slow examples should be there, but not as part of the CI tests).

fingolfin:
@afkafkafk13 just explained that someone else in Berlin (@ooinaruhugh) is working on Gröbner walk stuff and is building on top of this to produce new code (they will probably open a new PR for that, but let's keep this one open for now as reference).

fingolfin:
Closing this in favor of PR #3821 which borrows some code from here but is under active development.

PR #3132, opened 2023-12-21
Re-export deprecated bindings from AA and Nemo

fingolfin:
This restores missing aliases like `PolynomialRing`, as discovered in #3114 (CC @thofma).

In principle I think we could merge this right away, or we could wait for AA/Nemo/Hecke releases to "simplify" this... but I don't see a major advantage, just merging this now and cleaning it up later would allow us to have 0.14.0 out before Christmas, maybe?

### DIFF
--- a/src/aliases.jl
+++ b/src/aliases.jl
@@ -1,3 +1,22 @@
+include(joinpath(pathof(AbstractAlgebra), "..", "Aliases.jl"))
+
+import Nemo: is_cyclo_type
+import Nemo: is_embedded
+import Nemo: is_maxreal_type
+import Nemo: ZZModRing  # FIXME: remove if/once Nemo exports this
+import Nemo: zzModRing  # FIXME: remove if/once Nemo exports this
+import Nemo: FpField  # FIXME: remove if/once Nemo exports this
+import Nemo: fpField  # FIXME: remove if/once Nemo exports this
+include(joinpath(pathof(Nemo), "..", "Aliases.jl"))
+
+#import Hecke: quadratic_genera
+#import Hecke: has_algebra
+#import Hecke: has_embedding
+#import Hecke: has_matrix_action
+#import Hecke: has_root
+#import Hecke: ...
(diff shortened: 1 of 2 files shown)


src/aliases.jl
```diff
@@ -1,3 +1,26 @@
...
+include(joinpath(pathof(Nemo), "..", "Aliases.jl"))
+
+#import Hecke: quadratic_genera
+#import Hecke: has_algebra
+#import Hecke: has_embedding
+#import Hecke: has_matrix_action
+#import Hecke: has_root
+#import Hecke: ...
+#include(joinpath(pathof(Hecke), "..", "Aliases.jl"))
+
```
fingolfin, line 19: I guess we could remove this... I initially thought it would be good to have Hecke compat, too, but there are a *ton* of things in Hecke for which we have deprecated aliases but which never were exported... So it is a lot of work to adjust this. I can do it, but only if someone (@thofma, @fieker, ...) tells me it is important.
```suggestion
```

src/aliases.jl
```diff
@@ -1,3 +1,26 @@
+import AbstractAlgebra: LaurentMPolyRing  # FIXME: remove these imports once AA exports these
```
lgoettgens: There is already an AbstractAlgebra release with them, right? So why don't drop the here and now?

fingolfin: Yeah I just noticed it, too -- fixed now!

thofma (approved):
looks good to me

thofma:
So much green, except for #3136

thofma:
I am seeing
```
julia> using Oscar
[ Info: Precompiling Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13]
WARNING: import of Nemo.is_embedded into Oscar conflicts with an existing identifier; ignored.
```
on master. There is some `is_embedded` used earlier than the `import Nemo: is_embedded` in `aliases.jl`. Should it be added to `imports.jl`? @fingolfin @lgoettgens

PR #3137, opened 2023-12-22
Add `characteristic` for `QQAbField`

lgoettgens:
To reduce the impact of https://github.com/Nemocas/AbstractAlgebra.jl/pull/1540.


thofma (approved):
(I wonder if it should always be a `ZZRingElem` for things on top of Nemo, but this can be changed later.)

fieker:
On Fri, Dec 22, 2023 at 03:09:42AM -0800, Tommy Hofmann wrote:
> @thofma approved this pull request.
> 
> (I wonder if it should always be a `ZZRingElem` for things on top of Nemo, but this can be changed later.)
> 
difficult with current GF(p), p small...

PR #3139, opened 2023-12-22
Excise GroupsCore dependency

fingolfin:
Requires https://github.com/Nemocas/AbstractAlgebra.jl/pull/1528

Closes #3070

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -128,7 +128,7 @@ false
 function order(::Type{T}, x::Union{GAPGroupElem, GAPGroup}) where T <: IntegerUnion
    ord = GAPWrap.Order(x.X)
    if ord === GAP.Globals.infinity
-      throw(GroupsCore.InfiniteOrder(x))
+      throw(InfiniteOrder(x))
    end
    return T(ord)
 end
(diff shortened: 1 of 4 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -128,7 +128,7 @@ false
 function order(::Type{T}, x::Union{GAPGroupElem, GAPGroup}) where T <: IntegerUnion
    ord = GAPWrap.Order(x.X)
    if ord === GAP.Globals.infinity
-      throw(GroupsCore.InfiniteOrder(x))
+      throw(InfiniteOrder(x))
```
lgoettgens, line 131:
```suggestion
      throw(InfiniteOrderError(x))
```
I think you chose a different name

src/Groups/GAPGroups.jl
```diff
@@ -349,8 +349,7 @@ Base.:/(x::GAPGroupElem, y::GAPGroupElem) = group_element(parent(x), (x.X / y.X)
 
 Base.:\(x::GAPGroupElem, y::GAPGroupElem) = group_element(parent(x), (x.X \ y.X)::GapObj)
 
-# Compatibility with GroupsCore interface
-one!(x::GAPGroupElem) = one(parent(x))
+# Compatibility with groups interface
 inv!(out::GAPGroupElem, x::GAPGroupElem) = inv(x)  #if needed later
```
lgoettgens, line 353:
```suggestion
```
All of the following should go if they are equal to the default implementation in AA

lgoettgens:
https://github.com/oscar-system/Oscar.jl/blob/c2a4d7e0459d9e4001953da82e4b13ce24c77831/test/Groups/elements.jl#L91-L130 would need to get adapted as well

PR #3141, opened 2023-12-28
Origami code review (DO NOT MERGE!)

fingolfin:
This is a PR made from your own code, just so that I can easily leave inline comments for you about issues with the code. DO NOT MERGE THIS!

However, if you like you can use the "apply suggestion from code review" feature to apply certain code changes I suggest to your branch (just make sure to remember to `git pull`) the changes to your computer afterwards)


experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    #end
+end
+
+function origami_disconnected(h::PermGroupElem, v::PermGroupElem, d::Integer)
+    return Origami(GAP.Globals.OrigamiNC(GapObj(h), GapObj(v), d), h, v, d)
+end
+
+function Base.:(==)(a::Origami, b::Origami)
+    # TODO rewrite this? for now use Gap equality
+    return GAP.evalstr("$(string(a)) = $(string(b))")
```
fingolfin, line 58: Using `GAP.evalstr` should be a very, *very* last resort. You don't need to convert objects to strings to pass them to GAP, you can do it directly:
```suggestion
    return GapObj(a) == GapObj(b)
```

experimental/Origami/src/special_origigami.jl
```diff
@@ -0,0 +1,36 @@
...
+    h_start = (steps - 1) * (sum_l_h) + 1
+    h_end = (steps - 1) * (sum_l_h) + length
+    h = h * cperm(h_start:h_end)
+
+    v_start = steps * length + (steps - 1) * height
+    v_end = steps * sum_l_h
+    v = v * cperm(v_start:v_end)
+
+    return normal_form(origami(h, v))
+end
```
fingolfin, line 36: All your files in git should end with a newlin
```suggestion
end

```

experimental/Origami/src/types.jl
```diff
@@ -0,0 +1,6 @@
+mutable struct Origami
+    o::GapObj
+    h::PermGroupElem
+    v::PermGroupElem
+    d::Oscar.Integer
```
fingolfin, line 5: Don't use abstract types like `Oscar.Integer` in a struct if you can avoid it -- and here you certainly can: I think `d` is a "degree" and even fits into a "small" machine integer:
```suggestion
    d::Int
```
But perhaps my guess is wrong and this is a value that can be arbitrarily large. Then I would recommend using a `ZZRingElem` (the OSCAR type for big integers, similar to but different from Julia's own `BigInt` type)

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+GapObj(O::Origami) = O.o
+
+function stratum(o::Origami)
+    # cannot use comm from Oscar because of different definitions
+    h = horizontal_perm(o)
+    v = vertical_perm(o)
+    commutator = h * v * h^-1 * v^-1
+    cycle_struc = cycle_structure(commutator)
+    different_cycle_count = length(cycle_struc)
+    stratum::Vector{Integer} = []
```
fingolfin, line 89: This has two problems:
1. this uses a non-concrete type as type argument, loosing type stability later on (thus causing code to be slower than necessary)
2. this create a `Vector{Any}` (namely `[]`) and then converts it to the right type, instead of directly creating a vector of the right type.

Try this instead:
```suggestion
    stratum = ZZRingElem[]
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    for i in 1:different_cycle_count
+        cycle_type_entry = cycle_struc[i]
+        cycle_type_length = cycle_struc[i][1]
+        if(cycle_type_entry[1] == 1)
```
fingolfin, line 93: Weird formatting, should be:
```suggestion
        if cycle_type_entry[1] == 1
```
even better: use the auxiliary variable you introduced above:
```suggestion
        if cycle_type_length == 1
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+            # ignore cycles of length 1
+            continue
+        end
+        for j in 1:cycle_type_entry[2]
+            push!(stratum, ZZ(cycle_type_length - 1))
+        end
+    end
```
fingolfin, line 100: This is not very Julia-esque code. This here is much more idiomatic (but as everything else here: untested!). I also changed the vector `stratum` to use `Int` entries as those are always sufficient. But if you really want big integers instead, you can of course easily change it back.
```suggestion
    stratum = Int[]
    for (len,num) in cycle_structure(commutator)
        # ignore cycles of length 1
        len == 1 && continue
        for j in 1:num
            push!(stratum, len - 1)
        end
    end
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    return stratum
+end
+
+function genus(o::Origami)
+    return ZZ((sum(stratum(o)) + 2) * 0.5)
```
fingolfin, line 105: Multiple issues:
1. this can be implemented more efficiently by using `cycle_structure` directly instead of `stratum`
2. to divide an integer by 2 discarding the remainder, use `div(x,2)` don't multiply by a float (don't do that in GAP either)
3. it is unclear to me whether you really want a "large int" result, but if I am not mistaken the value never exceeds the degree which in turn fits into an `Int`?

Regarding point 1, `sum(stratum(o))` is equivalent to
```suggestion
    sum((len,num) -> num*(len-1), cycle_structure(BLA))
```
So of course the value `BLA` needs to be computed first... But it could be factored out of `stratum` into a common helper. Or if you don't need `stratum` for anything else, just merge this code in here?

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    GAP.Globals.ComputeVeechGroupWithHashTables(GapObj(O))
+end
+
+function index_monodromy_group(o::Origami)
+    GAP.Globals.IndexOfMonodromyGroup(GapObj(o))
+end
+
+function sum_of_lyapunov_exponents(o::Origami)
+    gap_obj = GAP.Globals.SumOfLyapunovExponents(GapObj(o))
+    return GAP.gap_to_julia(gap_obj)
```
fingolfin, line 119: Using `gap_to_julia` is not recommended. Instead specify the desired target type. I can only guess here, but maybe the result is a (potentially big) integer, in that case:
```suggestion
    return ZZRingElem(gap_obj)
```
This has multiple advantages:
1. It makes it obvious what the result type of this function is
2. it allows Julia itself to predict the return type, and hence makes the code type stable code (which is impossible when using `gap_to_julia`)
3. it adds a type check: if something goes wrong and the GAP function returns something that isn't an integer (e.g. `fail) then we get a helpful error here instead of a confusing one elsewhere later on.

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+end
+
+function translations(o::Origami)
+    h = horizontal_perm(o)
+    v = vertical_perm(o)
+    G = normalform_conjugators(o)
+
+    O = (i -> origami(i^-1 * h * i, i^-1 * v * i)).(G)
```
fingolfin, line 127: Conjugating permutations can be (and is!) implemented much more efficiently than with two multiplications and an inversion.
```suggestion
    O = (i -> origami(h^i, v^i)).(G)
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    for i in 1:l
+        positions = findall(item -> item == O[i], O)
+        if length(positions) != 1
+            for j in positions
+                push!(list, G[i] * G[j]^-1)
+            end
+        end
+    end
+
+    return collect(Set(list))
```
fingolfin, line 140: If you just want to remove duplicates, use `unique` or `unique!`

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    n = 2
+    n_inv = 1 / 2
+	x = horizontal_perm(o)
+	y = vertical_perm(o)
+	g = genus(o)
+
+    L = point_reflections(o)
+	L = filter(i -> order(i) == 2, L)
+
+	if L == []
```
fingolfin, line 158: The use of `[]` creates an object, just to check if `L` is empty. Better:
```suggestion
	if isempty(L)
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+        return false
+    end
+
+    # helper to find all elements in x that are not in y
+    function list_diff(x, y)
+        set_1 = Set(x)
+        set_2 = Set(y)
+        diff = setdiff(set_1, set_2)
+        return collect(diff)
+    end
```
fingolfin, line 168: ... This is literally what `setdiff` does. There is no need for the calls to `Set` and `collect`.

But actually I don't think you need either anyway, see below

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+
+    degree_list = collect(1:degree(o))
```
fingolfin, line 170: Why do you `colect` this perfectly fine range? If, say, `degree(o)` is 1000 then the range only takes 16 bytes or so (to store that start and end), while after collect it takes 8000 bytes (to store 1000 integer values, each at 8 bytes).

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    for sigma in L
+        # fixpoints
+        b = 0
+        b = b + length(list_diff(degree_list, moved_points(sigma)))
+        b = b + length(list_diff(degree_list, moved_points(sigma*x)))
+        b = b + length(list_diff(degree_list, moved_points(sigma*y)))
```
fingolfin, line 176: Isn't this the same as this code which is much more efficient?
```suggestion
        b = 3*degree(o) - number_moved_points(sigma)
                        - number_moved_points(sigma*x)
                        - number_moved_points(sigma*y)
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+
+        for i in degree_list
+            if i^(sigma * x^-1 * y^-1) == i^(y*x*(x*y)^-1)
+                b = b + 1
+            end
+        end
+
+        # TODO can n_inv yields floating point errors?
+        if n_inv * (g - 1 - b * 0.5) + 1 == 0
```
fingolfin, line 185: `n` and `n_inv` are constant (namely 2 and 0.5), is that intentional?

Anyway, this check is really strange, I feel I must be missing something... but assuming that `n_inv = 1/n` should always hold, can't you rewrite this as
```suggestion
        if g - 1 - b * 0.5 == -n
```
This is still odd, e.g. the use of `0.5` is fishy (as noted elsewhere). But I'll not make further suggestions here as I don't get what this is supposed to do

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    degree_list = collect(1:degree(o))
+    for sigma in L
+        # fixpoints
+        b = 0
+        b = b + length(list_diff(degree_list, moved_points(sigma)))
+        b = b + length(list_diff(degree_list, moved_points(sigma*x)))
+        b = b + length(list_diff(degree_list, moved_points(sigma*y)))
+
+        for i in degree_list
+            if i^(sigma * x^-1 * y^-1) == i^(y*x*(x*y)^-1)
```
fingolfin, line 179: This is equivalent to
```suggestion
            if i^sigma == i^(y*x*(x*y)^-1*y*x)
```
and then you can precompute `y*x*(x*y)^-1*y*x` into a variable, say `pi`, and end up with
```suggestion
            if i^sigma == i^pi
```
but then obviously the loop just computes the number of fixed points of `sigma/pi`, which we already established can be done very efficiently via `number_of_moved_points`

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+        if n_inv * (g - 1 - b * 0.5) + 1 == 0
+            return true
+        end
+    end
+
+    return false
+end
+
+function cylinder_structure(o::Origami)
+    # TODO returns Vector{Any}, maybe cast to Integer?
```
fingolfin, line 194: `Integer`? Do you mean `Vector{Integer}` That would be bad, though, better: `Vector{Int}` or `Vector{ZZRingElem}`

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    gap_obj = GAP.Globals.CylinderStructure(GapObj(o))
+    return GAP.gap_to_julia(gap_obj)
+end
+
+function veech_group_and_orbit(o::Origami)
+    return GAP.Globals.VeechGroupAndOrbit(GapObj(o))
+end
+
+function veech_group_is_even(o::Origami)
+    return GAP.Globals.VeechGroupIsEven(GapObj(o))
```
fingolfin, line 204: Does this return a boolean? Then I suggest adding a type annotation to make this clear, and to make the code type stable:
```suggestion
    return GAP.Globals.VeechGroupIsEven(GapObj(o))::Bool
```
Also, I would then rename this function to `is_veech_group_even`, to match our general naming conventions.

fingolfin (commented):
Some initial comments. I did not test any of the code, so obviously don't apply any of my suggested changes blindly.

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+function are_equivalent(o1::Origami, o2::Origami)
+    return GAP.Globals.OrigamisEquivalent(GapObj(o1), GapObj(o2))
+end
+
+# TODO why is this not in canonical.jl?
+function normalform_conjugators(o::Origami)
+    x = horizontal_perm(o)
+    y = vertical_perm(o)
+    n = degree(o)
+    G = []
```
fingolfin, line 216:
```suggestion
    G = PermGroupElem[]
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+                push!(Q, wy)
+                seen[wy] = true
+                numSeen = numSeen + 1
+                L[wy] = numSeen
+            end
+        end
+        push!(G, L)
+    end
+
+    return perm.(G)
```
fingolfin, line 251: Just create the permutations right away, that halves the peak memory usage, I think?
```suggestion
        push!(G, perm(L))
    end

    return G
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+end
+
+function point_reflections(o::Origami)
+    if !veech_group_is_even(o)
+        throw("VeechGroup must contain -1")
```
fingolfin, line 256: Never throw strings! Either use `error`, or throw a proper error object
```suggestion
        error("VeechGroup must contain -1")
```

Here you might want to use the `@req` macro (see its docstring for details)

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    result::Vector{PermGroupElem} = []
+
+    l = length(O)
+    for i in 1:l
+        index = findfirst(item -> item == O[i], O1)
+        push!(result, G[i] * (G1[index])^-1)
+    end
+
+    # Remove duplicates - TODO better solution?
+    return collect(Set(result))
```
fingolfin, line 286:
```suggestion
    return unique!(result)
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    # sigma_i *origami *sigma_i^-1=delta_k(i)*origami_1*delta_k(i)^-1
+    O1 = f1.(G1)
+
+    # fitting the permuations together
+    result::Vector{PermGroupElem} = []
+
+    l = length(O)
+    for i in 1:l
+        index = findfirst(item -> item == O[i], O1)
+        push!(result, G[i] * (G1[index])^-1)
```
fingolfin, line 282: "Dividing" two permutations can be done more efficiently than a multiplication plus an inversion
```suggestion
        push!(result, G[i] / G1[index])
```
Although the fastest results are for computing `a^-1*b` -- depending on what your function is supposed to return, exactly, maybe that'd be an option here?
```suggestion
        push!(result, G[i] \ G1[index])
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+    h = horizontal_perm(o)
+    v = vertical_perm(o)
+    o1 = origami(h^-1, v^-1)
+    h1 = horizontal_perm(o1)
+    v1 = vertical_perm(o1)
+    G = normalform_conjugators(o)
+    G1 = normalform_conjugators(o1)
+
+    f(i) = origami(i^-1 * h * i, i^-1 * v * i)
+    f1(i) = origami(i^-1 * h1 * i, i^-1 * v1 * i)
```
fingolfin, line 268:
```suggestion
    f(i) = origami(h ^ i, v ^ i)
    f1(i) = origami(h1 ^ i, v1 ^ i)
```

experimental/Origami/src/Origami.jl
```diff
@@ -0,0 +1,299 @@
...
+
+    # origamis derived from the permutations above
+    O = f.(G)
+    # we need to calculate these to test find k s.t.
+    # sigma_i *origami *sigma_i^-1=delta_k(i)*origami_1*delta_k(i)^-1
+    O1 = f1.(G1)
+
+    # fitting the permuations together
+    result::Vector{PermGroupElem} = []
```
fingolfin, line 277:
```suggestion
    result = PermGroupElem[]
```

experimental/Origami/src/action.jl
```diff
@@ -0,0 +1,41 @@
...
+
+function action_s(o::Origami)
+    h = horizontal_perm(o)
+    v = vertical_perm(o)
+    return origami_disconnected(v^-1, h, degree(o))
+end
+
+function action_t(o::Origami)
+    h = horizontal_perm(o)
+    return origami_disconnected(h, h^-1 * vertical_perm(o), degree(o))
```
fingolfin, line 11:
```suggestion
    return origami_disconnected(h, h \ vertical_perm(o), degree(o))
```

experimental/Origami/src/action.jl
```diff
@@ -0,0 +1,41 @@
...
+function action_s_inv(o::Origami)
+    h = horizontal_perm(o)
+    v = vertical_perm(o)
+    return origami_disconnected(v, h^-1, degree(o))
+end
+
+function action_sl2(A::ZZMatrix, o::Origami)
+    # TODO necessary? loses performance
+    sl2z = special_linear_group(2, ZZ)
+    if !(A in sl2z)
```
fingolfin, line 29:
```suggestion
    if abs(det(A)) != 1
```

experimental/Origami/src/action.jl
```diff
@@ -0,0 +1,41 @@
...
+        throw("Matrix has to be an element of SL2(Z)!")
```
fingolfin, line 30: Again don't throw strings!

experimental/Origami/src/action.jl
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    # TODO implement this in Oscar when implementing the modular subgroup pkg
+    # because this is really slow and ugly
+    gap_origami = GAP.Globals.ActionOfSL2(GAP.julia_to_gap(A), GapObj(o))
+    gap_h = GAP.Globals.HorizontalPerm(gap_origami)
+    gap_v = GAP.Globals.VerticalPerm(gap_origami)
+    h = perm((i-> Integer(i)).(GAP.gap_to_julia(GAP.Globals.ListPerm(gap_h))))
```
fingolfin, line 38: Again multiple issues to note:
1. in this context, `i -> Integer(i)` is a complicated way to say `Integer` :-)
2. but don't use `Integer` here, use a concrete type such as `Int` !
3. but really, just directly convert the list to a `Vector{Int}` instead of first converting to a `Vector{Any}`, i.e.:
```suggestion
    h = perm(Vector{Int}(GAP.Globals.ListPerm(gap_h)))
```
4. but *really* we should be able to just convert (or rather, wrap) this directly into a `PermGroupElem`. I suspect something is missing here in the Oscar, and this also now gives me an idea for the source of the slowdowns with permutations you are seeing... I'll investigate that when I am less sick and get backt to you

experimental/Origami/src/action.jl
```diff
@@ -0,0 +1,41 @@
...
+function action_sl2(A::ZZMatrix, o::Origami)
+    # TODO necessary? loses performance
+    sl2z = special_linear_group(2, ZZ)
+    if !(A in sl2z)
+        throw("Matrix has to be an element of SL2(Z)!")
+    end
+
+    # TODO implement this in Oscar when implementing the modular subgroup pkg
+    # because this is really slow and ugly
+    gap_origami = GAP.Globals.ActionOfSL2(GAP.julia_to_gap(A), GapObj(o))
```
fingolfin, line 35:
```suggestion
    gap_origami = GAP.Globals.ActionOfSL2(GapObj(A), GapObj(o))
```

fingolfin (commented):
Some more comments. This is all I can do right now, fever is raising again.

I also just noticed that I opened this PR on the wrong repository. Dunno if I can transfer it, otherwise I'll just close it here (you can then still ready my comments)

PR #3142, opened 2023-12-30
test: make worker print test name at the beginning

thofma:
This is handy locally to know what is currently taking forever.

(Also remove the splatting)

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -130,7 +130,7 @@ end
 # if many workers, distribute tasks across them
 # otherwise, is essentially a serial loop
 stats = reduce(merge, pmap(testlist) do x
-                        @info "Starting tests for $x"
+                        println("Starting tests for $x")
                         Oscar.test_module(x; new=false, timed=true)
                       end)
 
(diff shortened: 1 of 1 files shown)


test/runtests.jl
```diff
@@ -129,7 +129,10 @@ end
 
 # if many workers, distribute tasks across them
 # otherwise, is essentially a serial loop
-stats = merge(pmap(x -> Oscar.test_module(x; new=false, timed=true), testlist)...)
+stats = reduce(merge, pmap(testlist) do x
+                        @info "Starting tests for $x"
```
benlorenz:
```suggestion
                        println("Starting tests for $x")
```
I think it would be better to use a plain `println`, I don't really know why but `@info` seems to cause the output to be garbled / mixed with the surrounding output (only on nightly):
https://github.com/oscar-system/Oscar.jl/actions/runs/7363413095/job/20042856979?pr=3142#step:7:1936
This effect does also happen a little bit on master but a lot less, with the PR here I found only four correct "Starting tests" printouts in the whole log (vs 100 times "-> Testing" which uses println).

benlorenz (commented):
Thanks, that is indeed helpful.
I added one comment regarding the output on nightly.

test/runtests.jl
thofma: Makes sense, thanks

benlorenz (approved):

PR #3144, opened 2023-12-30
Use julia 1.10 as default in CI

lgoettgens:
Now that julia 1.10 is released, we can drop 1.9 in CI in favor of 1.10.
Once this is merged, we need a change to https://github.com/oscar-system/OscarDevTools.jl/blob/master/src/defaults.jl

Edit: Merging this needs also an update to the list of required CI jobs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,8 +23,7 @@ jobs:
       matrix:
         julia-version:
           - '1.6'
-          - '1.9'
-          - '~1.10.0-0'
+          - '1.10'
           - 'nightly'
         group: [ 'short', 'long' ]
         os:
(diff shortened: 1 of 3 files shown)


benlorenz:
I'm fine with switching the default to 1.10, but I would prefer to keep CI for 1.9 running for a little bit longer to give people some time to upgrade (and not accidentally break their workflow). The Linux workers are rarely at the limit and the tests run rather reliably within one hour no matter if we test just 1.10 or keep 1.9. (The slowest one is always the macos one that runs all tests and doctests.)

lgoettgens:
> I'm fine with switching the default to 1.10, but I would prefer to keep CI for 1.9 running for a little bit longer to give people some time to upgrade (and not accidentally break their workflow). The Linux workers are rarely at the limit and the tests run rather reliably within one hour no matter if we test just 1.10 or keep 1.9. (The slowest one is always the macos one that runs all tests and doctests.)

Done. I suggest just dropping 1.9 in testing once 1.11 gets branched off the master.

thofma:
I agree with Benjamin. Let's wait a few weeks with dropping 1.9 from CI.

lgoettgens:
> I agree with Benjamin. Let's wait a few weeks with dropping 1.9 from CI.

Already changed

benlorenz (approved):
I think we could merge this now since this was changed to just switch the default and still includes 1.9.

lgoettgens:
FYI: The force-push didn't change anything

fingolfin (approved):

PR #3151, opened 2024-01-04
Sort GAP wrappers using 'sort -f'

fingolfin:
I don't really care which sorting to use, as long as we use it consistently, so that I don't have to worry about duplicates or where to insert new wrappers.

### DIFF
--- a/src/GAP/wrappers.jl
+++ b/src/GAP/wrappers.jl
@@ -12,8 +12,9 @@ module GAPWrap
 
 using GAP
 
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
O.k.
Perhaps it would be safer to write "the following list is intended to be sorted according to `sort -f`, i.e. ignoring case", as long as there is no automatic check of this property.

fingolfin:
Sure, done.

PR #3153, opened 2024-01-05
feat: bump dependencies

thofma:
- adjust to finite field changes in Nemo
- adjust to broadcasting changes


docs/src/CommutativeAlgebra/rings.md
```diff
@@ -126,10 +126,10 @@ Rational field
 
 ```jldoctest
 julia> GF(3)
-Finite field of characteristic 3
+Finite field of degree 1 over GF(3)
```
lgoettgens, line 129: the printing of prime fields needs to be re-done, but not a blocker for this

src/Rings/AlgClosureFp.jl
```diff
@@ -104,11 +104,11 @@ function ext_of_degree(A::AlgClosure, d::Int)
...
   k = base_ring(A)
   if isa(k, Nemo.fpField) || isa(k, fqPolyRepField)
-    K = GF(Int(characteristic(k)), d, cached = false)
+    K = Nemo.Native.GF(Int(characteristic(k)), d, cached = false)
   elseif isa(k, FqField)
-    K = Nemo._GF(characteristic(k), d, cached = false)
-  else
     K = GF(characteristic(k), d, cached = false)
+  else
+    K = Nemo.Nativ.GF(characteristic(k), d, cached = false)
```
lgoettgens:
```suggestion
    K = Nemo.Native.GF(characteristic(k), d, cached = false)
```
typo

src/Rings/FunctionField.jl
```diff
@@ -85,10 +85,23 @@ function (F::Nemo.fpField)(x::AbstractAlgebra.Generic.Frac{T}) where T <: Union{
   F(cst_num) // F(cst_denom)
 end
 
+function (F::Nemo.FqField)(x::AbstractAlgebra.Generic.Frac{T}) where T <: Union{FqPolyRingElem, FqMPolyRingElem}
+  num = numerator(x)
+  cst_num = constant_coefficient(num)
+  denom = denominator(x)
+  cst_denom = constant_coefficient(denom)
+  if (num != cst_num || denom != cst_denom) throw(InexactError(:fpFieldElem, fpFieldElem, x)) end
```
lgoettgens:
```suggestion
  if (num != cst_num || denom != cst_denom) throw(InexactError(:FqFieldElem, FqFieldElem, x)) end
```
I think this should be changed, as there is no `fpFieldElem` anywhere in this function

test/GAP/iso_oscar_gap.jl
```diff
@@ -115,7 +115,6 @@ end
    end
    @test GAP.Globals.DefiningPolynomial(codomain(f)) ==
          GAP.Globals.ConwayPolynomial(p, 2)
-   @test F.is_conway == 0
```
lgoettgens, line 118: Why does this get removed?

test/Groups/elements.jl
```diff
@@ -92,7 +92,7 @@ end
    @test eltype(PermGroup)==PermGroupElem
    @test eltype(PcGroup)==PcGroupElem
    @test eltype(FPGroup)==FPGroupElem
-   @test eltype(GL(2,3))==MatrixGroupElem{fqPolyRepFieldElem,fqPolyRepMatrix}
+   #@test eltype(GL(2,3))==MatrixGroupElem{fqPolyRepFieldElem,fqPolyRepMatrix}
```
lgoettgens: Why did this break?
It would be better to change to `@test_broken` instead of commenting if not possible to fix right now.
```suggestion
   @test_broken eltype(GL(2,3))==MatrixGroupElem{fqPolyRepFieldElem,fqPolyRepMatrix}
```

lgoettgens:
```suggestion
   @test eltype(GL(2,3))==MatrixGroupElem{FqFieldElem, FqMatrix}
```
locally running shows this instead, which looks right

lgoettgens (commented):
Most looks good (apart from comments). The singular conversion is just black magic for me, no idea what's happening there, but it seems to just work.

I don't know what I think about the need to add `lift` at so many places. Was this a conscious decision or, if not, can we instead add something of the form `(::FqField)(::ZZModRingElem)` to Nemo instead?

lgoettgens:
I furthermore found some more places that probably/maybe need to be updated:
* https://github.com/oscar-system/Oscar.jl/blob/acc729de2228943dc37ae63eb33a69b9972f45a1/experimental/PlaneCurve/src/AffinePlaneCurve.jl#L539
* maybe https://github.com/oscar-system/Oscar.jl/blob/c2a4d7e0459d9e4001953da82e4b13ce24c77831/experimental/ModStd/ModStdQ.jl#L233
* (multiple places), maybe make the whole thing a jldoctest as there are more old outputs in it  https://github.com/oscar-system/Oscar.jl/blob/c2a4d7e0459d9e4001953da82e4b13ce24c77831/src/InvariantTheory/iterators.jl#L226-L236
* https://github.com/oscar-system/Oscar.jl/blob/c2a4d7e0459d9e4001953da82e4b13ce24c77831/test/GAP/iso_oscar_gap.jl#L310-L312 
* Everything in `data/Surfaces` uses `Nemo.fpField`, should this get updated?
* and similar for the other docstrings in this file https://github.com/oscar-system/Oscar.jl/blob/c2a4d7e0459d9e4001953da82e4b13ce24c77831/src/Groups/matrices/MatGrp.jl#L563

Given that currently, only a non-consistent subset of the "old" finite field types is exported (e.g. `fpFieldElem` is exported but `fpField` is not), and we want to push towards the default field in the most cases, how about no longer exporting all of the `Fp*`, `fp*`, `fqPolyRep*`, and `FqPolyRep*` stuff (should probably be done in Nemo in a new breaking release)?

test/GAP/iso_oscar_gap.jl
thofma: This does not exist anymore.

test/Groups/elements.jl
thofma: I am afraid the test makes little sense and tests things which are not part of the API. These are exactly the tests, which make changing non-API things harder than they need to be.

thofma: I will remove it.

thofma:
> I don't know what I think about the need to add `lift` at so many places. Was this a conscious decision or, if not, can we instead add something of the form `(::FqField)(::ZZModRingElem)` to Nemo instead?

Maybe we can add it at some later point.

I agree with your point about not exporting the old types.

Regarding the other things that you found. Since those things are not tested apparently, it is not clear to me that they actually work. It is a rabbit hole, for which I don't have time for. If things are supposed to not break, they need to be tested.

test/Groups/elements.jl
thofma:
```suggestion
  @test eltype(GL(2,3))==MatrixGroupElem{elem_type(typeof(GF(2))),dense_matrix_type(GF(2))}
```

lgoettgens:
> Regarding the other things that you found. Since those things are not tested apparently, it is not clear to me that they actually work. It is a rabbit hole, for which I don't have time for. If things are supposed to not break, they need to be tested.

For points 4 and 6 it's just some comment that needs to be adjusted. Not doing that will bite our own butt in the future due to inconsistencies. I can deliver the patches here myself 

Points 1 and 3 (converting that to a jldoctest and fixing) can of course be done in a separate PR. 

For 2 and 5 I wasn't sure if they are relevant. If you acknowledge and defer them, that's fine for me.

test/Groups/elements.jl
```diff
@@ -92,7 +92,7 @@ end
    @test eltype(PermGroup)==PermGroupElem
    @test eltype(PcGroup)==PcGroupElem
    @test eltype(FPGroup)==FPGroupElem
-   @test eltype(GL(2,3))==MatrixGroupElem{fqPolyRepFieldElem,fqPolyRepMatrix}
+  @test eltype(GL(2,3))==MatrixGroupElem{elem_type(typeof(GF(2))),dense_matrix_type(GF(2))}
```
lgoettgens:
```suggestion
   @test eltype(GL(2,3))==MatrixGroupElem{elem_type(typeof(GF(2))),dense_matrix_type(GF(2))}
```

thofma:
Thanks. If you are happy, can you give this a positive review?

lgoettgens (approved):
Of course, forgot that last night

Issue #3147, opened 2024-01-02
Basic combinatorics in OSCAR


joschmitt:
I put some ideas together how we could improve the functionality of basic combinatorics in OSCAR, see https://hackmd.io/Ef1t9wjbQwWlkrL4DQgYxQ . Please edit and comment on anything you can think of there (assuming I set the permissions correctly).

For me, the motivation is the discussion in https://github.com/oscar-system/Oscar.jl/pull/3021#discussion_r1399151945 and the subsequent discovery that we have at least three implementations of 'weak compositions of d in n parts' floating around.

See also #1732.

lgoettgens:
Maybe @ulthiel is interested in this as well.

lgoettgens:
Personally, I would like to see most of this upstream, maybe in AbstractAlgebra, so that we don't need additional implementations of some things in AA and Hecke.

thofma:
I would suggest we first get something consistent here in OSCAR and afterwards start thinking about moving things around. The focus before 1.0 should be on a final API (which is independent of where the things reside).

mjrodgers:
I'd be interested in helping work on this, both the basic combinatorial functions and various q-analogs.

I see there is some q-analog stuff in JuLie that hasn't been migrated to experimental, I would vote that if we bring this over, these functions are renamed to not use the term quantum. I think it is most common to just call these q-factorial, q-binomial, etc.

joschmitt:
> I'd be interested in helping work on this, both the basic combinatorial functions and various q-analogs.
> 
> I see there is some q-analog stuff in JuLie that hasn't been migrated to experimental, I would vote that if we bring this over, these functions are renamed to not use the term quantum. I think it is most common to just call these q-factorial, q-binomial, etc.

Great! I don't know much about combinatorics, so any help by somebody who knows this stuff is very welcome.

thofma:
Let me also ping @lkastner: is there anyone in Berlin that has some input on the API question for combinatorics?

jankoboehm:
I would be interested in this. I would like to use it in particular in my class on combinatorics for computer science (which would require to have an easy to use and stable user interface and docu). Next iteration of the class is in the coming summer. Some thoughts: 

First of all, combinatorics is about numbers, so we should also have functions which give you the number of the objects under consideration. Some are given by closed expressions, others through recursions, etc.

[...]

PR #3159, opened 2024-01-08
Promote JuLie to `src`

### LINKED ISSUE #2301: Partition functionality in `src` missing
TL;DR I can haz partitions inside `src`?

Let me quote the third commandment of OSCAR development:

> Thou shalt not covet code from `experimental`

(or for heathens: "Code from `src` must never use code from `experimental`")

Now we stopped importing the partition functionality from AbstractAlgebra to export the partition functionality from `experimental/JuLie`. So we now have `Partition` etc, but we cannot use it inside `src/`. This is a rather unfortunate state. (My motivation comes from using partitions in #1986.)

<s>
I think we should do something about this, so that we can use partitions again in `Oscar/src`. I see two options
1. Move `Partition/partitions` out of `experimental/JuLie` into `src/Combinatorics`.
2. Reenable `AbstractAlgebra.Partition` as the default partition type in Oscar.

I really don't like 2. Are there any objections to 1.?
</s>

We decided to move the parts related to partitions to `src/`.

### LINKED ISSUE #2462: Add display options from `AbstractAlgebra.YoungTableau` to `Tableau`
When displaying a `Tableau`, it gets printed as an array of vectors. Especially for classroom notebooks, it would be nice to have an option to have the proper filled young diagram drawn.

`AbstractAlgebra.YoungTableau` already prints tableaux like such:
```
┌───┬───┬───┐
│ 1 │ 2 │ 3 │
├───┼───┼───┘
│ 4 │ 5 │
├───┼───┘
│ 6 │
└───┘
```
which would be ideal.

**Describe alternatives you've considered**
`Tableau.t` returns the tableau as vector of vectors, which is pretty close but doesn't align properly with large enough numbers.
Using `AbstractAlgebra.YoungTableau` directly would also be an option but both types have different constructors and you can't easily use `standard_tableaux()` etc.

**Additional context**
--

joschmitt:
Resolves #2301 and resolves #2462 as a start on #3147 .
I decided to be so bold to move `experimental/JuLie` to `src/Combinatorics` and did some cosmetic changes.
What else needs to be done to make this ready for `src`? I cannot judge the mathematics. Regarding the source code, I feel that the wording/style of the documentation is often a bit different from what we have elsewhere (e.g., the documentation mostly gives the mathematical definition of the object it returns). Other than that there are some violations of our style guide (whitespaces etc.), but I have seen worse code in `src`...

To do:
* [ ] make `partitions` etc. return an iterator
* [x] tabs -> spaces
* [x] lower case constructor for `Tableau`
* [x] getter functions for `Partition` and `Tableau`
* [x] Rename `Tableau` -> `YoungTableau`
* [x] More (minor) adjustments regarding the style guide
* [x] adjust wording of documentation
* [x] Feature: make tableaux print fancier (even without unicode we should be able to print them better) https://github.com/oscar-system/Oscar.jl/issues/2462
* [x] Put `ascending_partitions` back in experimental for the time being
* [x] Add sanity checks to the constructors and a `check` keyword to disable them
* [x] Adjust `schur_polynomial` to https://github.com/Nemocas/Nemo.jl/pull/1619 once we updated Nemo
* [x] Decide what integer type `number_of_partitions` should return

More?


test/Combinatorics/IntegerCombinatorics/tableaux.jl
```diff
@@ -19,7 +19,6 @@
 	@test is_standard(Tableau([[1,2],[4]])) == false
 	@test is_standard(Tableau([[1,3,2],[4]])) == false
 
-
 	# is_semistandard
 	@test is_semistandard(Tableau([[1,2,4,7,8],[3,5,6,9],[10]])) == true
```
lgoettgens: please address the tabs here

src/Combinatorics/IntegerCombinatorics/types.jl
```diff
@@ -0,0 +1,120 @@
...
+end
+
+################################################################################
+#
+#  Tableau
+#
+################################################################################
+
+@doc raw"""
+    Tableau{T} <: AbstractVector{AbstractVector{T}}
```
lgoettgens: stupid question maybe, but why don't we make the name more explicit? Even the docstring only talks about "Young tableaus".
```suggestion
    YoungTableau{T} <: AbstractVector{AbstractVector{T}}
```

experimental/Experimental.jl
```diff
@@ -16,7 +16,6 @@ const oldexppkgs = [
 # DO NOT USE THIS UNLESS YOU KNOW THE CONSEQUENCES.
 # For more background, see https://github.com/oscar-system/Oscar.jl/issues/2300.
 const orderedpkgs = [
-  "JuLie",
   "IntersectionTheory",
   "OrthogonalDiscriminants",  # needs code from JuLie
```
lgoettgens: If they just depend on JuLie (and not on one another), they can be removed here. (I didn't check, but assume from the comment)

test/Combinatorics/IntegerCombinatorics/tableaux.jl
joschmitt: Right. I'll file that under the "little" things that need to be done for conformance with our style guide.

src/Combinatorics/IntegerCombinatorics/types.jl
joschmitt: I don't have an opinion (I don't work with tableaux). In AbstractAlgebra it is called `YoungTableau` too. Either way, there should be a lower case constructor (`tableau` or `young_tableau` or both) which is so far missing.

experimental/Experimental.jl
joschmitt: I also assumed so, but did not check. I'll check and adjust this.

src/Combinatorics/IntegerCombinatorics/types.jl
thofma: Maybe we can use `YoungTableau` to be on the safe side for now. We could change it later. The other way around would be more difficult.

src/Combinatorics/IntegerCombinatorics/partitions.jl
```diff
@@ -269,10 +209,8 @@ function partitions(n::IntegerUnion)
     push!(P, partition(d[1:k]))
   end
   return P
-
 end
 
-
 @doc raw"""
     ascending_partitions(n::IntegerUnion; algorithm::Symbol=:ks)
```
joschmitt: I'm confused about this function. Reading the doc string, it looks like this was just an experiment: A paper claims that generating ascending partitions is faster than descending ones (which is the usual convention), but this appears not to be the case with the implementation. So, why is this function here (and even exported!)?

If I understand the situation correctly and we want to make all implemented algorithms available to the user, I would suggest to give `partitions` an optional argument to choose an algorithm. If it is one that generates ascending partitions, we can turn them around.

There was also some discussion https://github.com/oscar-system/Oscar.jl/pull/1935#discussion_r1131794777 about the timings. For what it's worth, here are mine:
```
julia> @btime partitions(Int8(90));
  8.120 s (56634195 allocations: 4.53 GiB)

julia> @btime ascending_partitions(Int8(90), algorithm=:ks);
  20.021 s (56634195 allocations: 4.53 GiB)

julia> @btime ascending_partitions(Int8(90), algorithm=:m);
  20.054 s (56634195 allocations: 4.53 GiB)
```

joschmitt:
According to my own plans in #3147 `partitions` etc. should return iterators. This will require a bit more work.

src/Combinatorics/IntegerCombinatorics/partitions.jl
ulthiel: I mentioned elsewhere that ascending partitions should be renamed to ascending_compositions because it's compositions actually. Why I implemented this? Out of curiosity. I think it's interesting. (Also, I should note that this was my private enterprise).

ulthiel:
> According to my own plans in #3147 `partitions` etc. should return iterators. This will require a bit more work.

If I remember correctly, I have implemented an interator, no? 

I checked and recall now: for compositions I implemented an iterator and convinced myself as well that this is the way to go. I wanted to use that as a template for implementing a partition iterator as well but didn't do it. I think it shouldn't be too difficult to modify the code (it should be the same algorithm that I implemented because it's fast).

src/Combinatorics/IntegerCombinatorics/partitions.jl
ulthiel: I remember: compositions need to be ported as well and then I wanted to move ascending_partitions to ascending_compositions in the compositions file. I have to check my code where I stopped with this.

ulthiel: https://github.com/ulthiel/Oscar.jl/tree/ut/compositions was my starting point but I forgot what I did.

ulthiel:
A further thought on iterators (I was thinking a lot about this back then but then forgot everything): as far as possible for any combinatorial structure there should be an interator. There may be many different ways to generate one and the same set of combinatorial objects: different algorithms that also generate in different orders which may be very relevant in applications. Iterators should be implemented per algorithm (ParititionsXYZ, PartitionsABC, ...) and there should be a default choice of iterator. I first tried to have one common type of iterator with optional argument for algorithms. But this doesn't work because different algorithms may use different types of state vectors and one runs in type stability problems (which is bad for performance).

src/Combinatorics/IntegerCombinatorics/partitions.jl
joschmitt: Okay, so for now I would suggest to stop exporting `ascending_partitions` (and maybe remove it from the documentation).

joschmitt:
> as far as possible for any combinatorial structure there should be an interator. 

Good, then we all agree on this. From what I can see, the implementations already work iteratively, so changing this should not be too hard. But doing it efficiently etc. is also not entirely trivial. Let's see.

src/Combinatorics/IntegerCombinatorics/partitions.jl
lgoettgens: and maybe just leave it unexported in experimental for the time being?

joschmitt:
Turning `partitions` etc. into iterators is still missing, but I think this pull request should be reviewed (and merged) like this. That would make it easier to review the iterator stuff later (and also make it easier for other people to help with that).

So far, there is almost no change to functionality. A quick overview of what has been done is given by the check list at the top (or the commit descriptions).

lgoettgens:
While working on https://github.com/oscar-system/Oscar.jl/pull/3272, I noticed that Nemo provides its own `number_of_partitions` with slightly different overflow behavior. It would be great if you could unify your function with this one to reduce the friction between https://github.com/oscar-system/Oscar.jl/pull/3272 and this PR.

joschmitt:
> While working on #3272, I noticed that Nemo provides its own `number_of_partitions` with slightly different overflow behavior. It would be great if you could unify your function with this one to reduce the friction between #3272 and this PR.

Yeah, will do. This pull request isn't relevant for the book or anything so yours has priority.

ulthiel:
Can you explain the open "Figure out why we have several ways of counting partitions (and decide on one)"? The fastest one should be the one from FLINT I think and this one should be used.

joschmitt:
> Can you explain the open "Figure out why we have several ways of counting partitions (and decide on one)"? The fastest one should be the one from FLINT I think and this one should be used.

Apparently there is another `number_of_partitions` in Nemo (which was named differently until the unification of functions with `number_of_...` names yesterday). At first glance, it calls a different C function. I didn't look into it closely yet.

ulthiel:
> > Can you explain the open "Figure out why we have several ways of counting partitions (and decide on one)"? The fastest one should be the one from FLINT I think and this one should be used.
> 
> Apparently there is another `number_of_partitions` in Nemo (which was named differently until the unification of functions with `number_of_...` names yesterday). At first glance, it calls a different C function. I didn't look into it closely yet.

Okay,  maybe @thofma knows? I think we should use the one from the [paper](https://www.cambridge.org/core/journals/lms-journal-of-computation-and-mathematics/article/efficient-implementation-of-the-hardyramanujanrademacher-formula/80A6F53DCB85A4270641920D989A2E8F) by Fredrik Johansson, which should be the one implemented here: https://flintlib.org/doc/partitions.html (and this is the one I have been calling in JuLie).

joschmitt:
> > > Can you explain the open "Figure out why we have several ways of counting partitions (and decide on one)"? The fastest one should be the one from FLINT I think and this one should be used.
> > 
> > 
> > Apparently there is another `number_of_partitions` in Nemo (which was named differently until the unification of functions with `number_of_...` names yesterday). At first glance, it calls a different C function. I didn't look into it closely yet.
> 
> Okay, maybe @thofma knows? I think we should use the one from the [paper](https://www.cambridge.org/core/journals/lms-journal-of-computation-and-mathematics/article/efficient-implementation-of-the-hardyramanujanrademacher-formula/80A6F53DCB85A4270641920D989A2E8F) by Fredrik Johansson, which should be the one implemented here: https://flintlib.org/doc/partitions.html (and this is the one I have been calling in JuLie).

Nemo calls `partitions_fmpz_fmpz` or `partitions_fmpz_ui` https://flintlib.org/doc/partitions.html#c.partitions_fmpz_fmpz . In JuLie, `arith_number_of_partitions` is called https://flintlib.org/doc/arith.html?highlight=arith#c.arith_number_of_partitions . I don't know what the difference is, both documentations talk about the "Hardy-Ramanujan-Rademacher formula".

Apparently, Nemo is slower.
```
julia> for i in 500:100:1500
           @btime number_of_partitions(ZZ($i));
       end
  4.688 μs (21 allocations: 376 bytes)
  5.127 μs (21 allocations: 376 bytes)
  5.667 μs (21 allocations: 440 bytes)
  5.874 μs (26 allocations: 528 bytes)
  5.930 μs (26 allocations: 528 bytes)
  6.924 μs (27 allocations: 552 bytes)
  6.835 μs (27 allocations: 552 bytes)
  6.808 μs (27 allocations: 600 bytes)
  7.600 μs (27 allocations: 600 bytes)
  7.168 μs (27 allocations: 600 bytes)
  7.725 μs (27 allocations: 608 bytes)

julia> for i in 500:100:1500
           @btime Nemo.number_of_partitions(ZZ($i));
       end
  5.277 μs (5 allocations: 3.97 KiB)
  7.004 μs (5 allocations: 4.75 KiB)
  9.002 μs (5 allocations: 5.53 KiB)
  10.784 μs (5 allocations: 6.31 KiB)
  12.817 μs (5 allocations: 7.09 KiB)
  14.713 μs (5 allocations: 7.88 KiB)
  16.906 μs (5 allocations: 8.66 KiB)
  16.946 μs (13 allocations: 472 bytes)
  19.670 μs (13 allocations: 472 bytes)
  18.919 μs (13 allocations: 472 bytes)
  20.664 μs (13 allocations: 472 bytes)

```

joschmitt:
Okay, my previous comment was not correct.

Nemo calls `partitions_fmpz_ui` https://flintlib.org/doc/partitions.html#c.partitions_fmpz_ui **if the input is an `Int`** and forces the result in an `Int` again, although it actually comes as a `ZZRingElem`.
Nemo calls `partitions_fmpz_fmpz` https://flintlib.org/doc/partitions.html#c.partitions_fmpz_fmpz **if the input is a `ZZRingElem`**.
JuLie calls `arith_number_of_partitions` https://flintlib.org/doc/arith.html?highlight=arith#c.arith_number_of_partitions which in flint just calls `partitions_fmpz_ui`. In particular, JuLie can only handle 64-bit input (or whatever a `ulong` is):
```
julia> number_of_partitions(ZZ(2)^64)
ERROR: InexactError: convert(UInt64, 18446744073709551616)
Stacktrace:
 [1] UInt64(a::ZZRingElem)
   @ Nemo ~/.julia/packages/Nemo/dNug3/src/flint/fmpz.jl:2875
 [2] number_of_partitions(n::ZZRingElem)
   @ Oscar ~/.julia/dev/Oscar/src/Combinatorics/IntegerCombinatorics/partitions.jl:146
 [3] top-level scope
   @ REPL[2]:1

```
I would be in favour of having one function that automatically checks whether `partitions_fmpz_ui` or `partitions_fmpz_fmpz` should be called. The open question for me is whether we want to have the Nemo behaviour of forcing the output into the input type (and creating `InexactError`s if it doesn't fit).

lgoettgens:
> I would be in favour of having one function that automatically checks whether `partitions_fmpz_ui` or `partitions_fmpz_fmpz` should be called. The open question for me is whether we want to have the Nemo behaviour of forcing the output into the input type (and creating `InexactError`s if it doesn't fit).

We could have `number_of_partitions(T, n)` that returns the output in type `T`, and `number_of_partitions(n) = number_of_partitions(typeof(n), n)` like for other things like group orders

joschmitt:
> > I would be in favour of having one function that automatically checks whether `partitions_fmpz_ui` or `partitions_fmpz_fmpz` should be called. The open question for me is whether we want to have the Nemo behaviour of forcing the output into the input type (and creating `InexactError`s if it doesn't fit).
> 
> We could have `number_of_partitions(T, n)` that returns the output in type `T`, and `number_of_partitions(n) = number_of_partitions(typeof(n), n)` like for other things like group orders

I think that would be overkill. This is a function that takes an integer and produces an integer. The problem is just that the output will very quickly be much larger than the input, so I see an argument for always returning `ZZRingElem` since starting with input 406 the output won't fit 64 bit anyways (and that's what flint and JuLie do after all). But I have no idea with what numbers people usually work and what they expect.

ulthiel:
I think this (and all other combinatorial functions) should definitely return a ZZRingElem. How's it with factorial?

joschmitt:
> How's it with factorial?

```
julia> factorial(100)
ERROR: OverflowError: 100 is too large to look up in the table; consider using `factorial(big(100))` instead
Stacktrace:
 [1] factorial_lookup
   @ Base ./combinatorics.jl:19 [inlined]
 [2] factorial(n::Int64)
   @ Base ./combinatorics.jl:27
 [3] top-level scope
   @ REPL[2]:1
```
Can't say that I like it.

thofma:
`factorial(100)` is not a method we added.

We do it for example for:
```
julia> rising_factorial(100, 10)
ERROR: OverflowError: 1419367337623872000 * 109 overflowed for type Int64
Stacktrace:
  [1] throw_overflowerr_binaryop(op::Symbol, x::Int64, y::Int64)
    @ Base.Checked ./checked.jl:163
```
or
```
julia> binomial(Int(10900), Int(50))
ERROR: OverflowError: binomial(10900, 50) overflows
Stacktrace:
 [1] binomial(n::Int64, k::Int64)
   @ Base ./intfuncs.jl:1114
 [2] top-level scope
   @ REPL[5]:1
```

joschmitt:
So keeping the input type is established OSCAR practice?

thofma:
As usual, I don't think there is a rule. I can see people doing `for k in 1:binomial(20, 10)`, where one wants to keep everything a machine integer. But I don't see people doing `for k in number_of_partitions(50)` or similar things. So for `number_of_partitions` it might not even matter. So maybe just always return a `ZZRingElem`.

P.S.: I realize my `rising_factorial` example was a bad one, since this works for any ring, so probably does not fully qualify as "combinatorial".

ulthiel:
Return ZZRingElem.

joschmitt:
Okay, I changed it now so that it always returns a `ZZRingElem`.
(Right now, `partitions_fmpz_ui` is slower than `arith_number_of_partitions` at least for "small" examples. But as I said above, at least in the current flint version one calls the other, so I assume that this performance difference is gone, once Nemo uses flint 3.0 or something.)

joschmitt:
I would prefer, if this could be reviewed like this; iterators for partitions and tableaux will follow in a separate pull request (see also https://github.com/oscar-system/Oscar.jl/pull/3159#issuecomment-1917495079).
Of course, #3288 should be merged first.

thofma:
Is there someone who should/could review this?

src/aliases.jl
```diff
@@ -46,7 +45,6 @@ function number_of_vertices end
 @alias nmaxcells number_of_maximal_cells            # decided to have but not used anywhere
 @alias nmaxcones number_of_maximal_cones
 @alias nmaxpolyhedra number_of_maximal_polyhedra    # decided to have but not used anywhere
-@alias npartitions number_of_partitions             # decided to have but not used anywhere
```
lgoettgens: please keep these two lines in this file to provide the alias `npartitions` as documented in this [HackMD](https://hackmd.io/nRnyfrSxTVe5CzXidB1cBQ?both#number-of-%E2%80%A6) (but remove the comment in line 34)

joschmitt: Ah, sorry, somehow I thought that this was the old name from Nemo.

joschmitt:
> Is there someone who should/could review this?

Good question, I don't know. Maybe @mjrodgers could have a look?

docs/doc.main
```diff
@@ -240,9 +240,11 @@
       "Combinatorics/graphs.md",
       "Combinatorics/matroids.md",
       "Combinatorics/simplicialcomplexes.md",
-      "Combinatorics/partitions.md",
-      "Combinatorics/tableaux.md",
-      "Combinatorics/schur_polynomials.md",
+      "Integer combinatorics" => [
+          "Combinatorics/IntegerCombinatorics/partitions.md",
+          "Combinatorics/IntegerCombinatorics/tableaux.md",
+          "Combinatorics/IntegerCombinatorics/schur_polynomials.md",
```
mjrodgers:
```suggestion
      "Enumerative combinatorics" => [
          "Combinatorics/EnumerativeCombinatorics/partitions.md",
          "Combinatorics/EnumerativeCombinatorics/tableaux.md",
          "Combinatorics/EnumerativeCombinatorics/schur_polynomials.md",
```
I think "Enumerative Combinatorics" is a more standard name for this kind of thing - maybe consider this name change? (I know it would have to be changed also a few other places as well.)

mjrodgers:
I have been looking it over, and will finish up and submit a review tonight.

mjrodgers (approved):
There's a lot here and I didn't of course check all of the existing JuLie code, but I think this all looks good to me and adds some important functionality. I'm in favor of merging this into `src`, and we can make some further additions and improvements from here.

docs/doc.main
ulthiel: With Schur polynomials I'm not sure, the rest yes.

mjrodgers: I think it is okay, they are used a lot in generating functions for counting things. I definitely support splitting them off into an `AlgebraicCombinatorics` submodule in the future, if enough stuff gets developed to justify this, it would probably be a better fit for them.

joschmitt: Thank you for the review! I changed the name now. (If we ever find a better place for the Schur polynomials, we can move them.)

joschmitt:
This is good to go from my side (as soon as CI confirms that I didn't miss anything while renaming).

joschmitt:
Still needs an approving review by an adult :)

thofma (approved):

PR #3161, opened 2024-01-09
Split macOS CI jobs

### LINKED ISSUE #3136: macOS CI job times out regularly
For ubuntu, we now have three jobs (short, long, doctests).
On the other hand for macOS, all three of these are combined into one single job. This is unfortunately makes the job no longer succeed in 150 minutes (at least sometimes, re-running helps).
We could either split the job up or increase the allowed time.

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/3136.

~~This currently includes https://github.com/oscar-system/Oscar.jl/pull/3144 and gets rebased once that is merged.~~

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -41,8 +41,18 @@ jobs:
             group: 'long'
             depwarn: 'depwarn=error'
           # Add a few macOS jobs (not too many, the number we can run in parallel is limited)
+          - julia-version: '1.10'
+            os: macOS-latest
+            group: 'short'
+          - julia-version: '1.10'
+            os: macOS-latest
+            group: 'long'
           - julia-version: 'nightly'
             os: macOS-latest
+            group: 'short'
+          - julia-version: 'nightly'
+            os: macOS-latest
+            group: 'long'
 
     steps:
       - uses: actions/checkout@v3
@@ -117,15 +127,6 @@ jobs:
         if: runner.os == 'macOS'
         # restrict number of openMP threads on macOS due to oversubscription
         run: echo "OMP_NUM_THREADS=1" >> $GITHUB_ENV
-      - name: "Run tests"
-        # HACK: since macOS runners are so expensive, we reuse this one to run
-        # both regular tests and doctests
-        if: runner.os == 'macOS'
-        uses: julia-actions/julia-runtest@latest
-        with:
-          annotate: ${{ matrix.julia-version == '1.10' }}
-          coverage: ${{ matrix.julia-version == '1.10' }}
-          depwarn: ${{ matrix.depwarn == 'depwarn=error' && 'error' || 'no' }}
       - name: "Setup package"
         run: |
           julia --project=docs --color=yes -e '
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,26 +33,26 @@ jobs:
         include:
           # Add a single job per group with deprecation errors on the most recent stable julia version
           - julia-version: '1.10'
-            os: ubuntu-latest
             group: 'short'
+            os: ubuntu-latest
             depwarn: 'depwarn=error'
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -23,25 +23,35 @@ jobs:
...
+            group: 'short'
+            os: macOS-latest
+          - julia-version: '1.10'
+            group: 'long'
+            os: macOS-latest
           - julia-version: 'nightly'
+            group: 'short'
+            os: macOS-latest
+          - julia-version: 'nightly'
+            group: 'long'
```
fingolfin, line 54: Do I understand it correctly that this makes us go from 1 to 4 macOS jobs? (Maybe not, I don't know what the "group: short/long" does, so maybe only 2 out of 4 are run?)

Point is, last I checked we can only run 5 macOS jobs concurrently over all of `oscar-system`. So instead of being able to complete about 4-5 full PR builds in parallel, with this patch we could only do 1.25 (?), meaning that completion of our macOS CI jobs would become again a major bottleneck.

Perhaps it suffices to just test with 1.10, not against nightly? Or do one test against 1.10 and one against nightly?


Perhaps @aaruni96 who is looking into setting up a GitHub CI runner in Kaiserslautern can also look into how feasible it is to add a Mac as CI runner (we could easily buy a Mac mini or Mac studio)

benlorenz: `short` and `long` are just two different subsets of our testsuite, `long` contains about [15 long running test groups](https://github.com/oscar-system/Oscar.jl/blob/master/test/runtests.jl#L95-L113) and short all other tests. It would run all 4 jobs.

Currently we do have 2 macos jobs per PR, the full `test+doctest` on 1.9 and a `test` on `nightly`. So it is more like 2.5 PRs in parallel.
The full PR runs currently all need to wait about 150 minutes until the macos test+doctest job is done (or failed due to the timeout). The nightly test varies between 80 minutes and 110 minutes.

If we split this we will have just one PR running (`2×1.10 + 2×nightly + doctest`) but this should finish in slight less than 1 hour (the long jobs usually finish in 30-40 minutes, doctest usually takes less than 30 minutes). To speed it up we could also restrict to just the `short` testgroup on macos+nightly.

In summary, I don't think this will slow down things and if it does I would suggest we remove some tests for nightly until we can run our own macos runners.

lgoettgens: We are testing against stable releases to keep the correctness of our code and notice regressions on our side.
For nightly, this is different: here we test mainly against to notice regression on the julia side. In my opinion, the latter doesn't need to be tested on every commit of every PR. It would be enough to e.g. have an automatic daily (?) run of our master against nightly to notice regressions. Most regressions and breakages are recorded by the ubuntu runners anyway.
This would decrease the "usual PR job set" from 5 to 3 macOS runners, but add some scheduled ones.

thofma: > [...] It would be enough to e.g. have an automatic daily (?) run of our master against nightly to notice regressions. Most regressions and breakages are recorded by the ubuntu runners anyway.

We have tried this and it does not work in practice.

fingolfin: "daily runs" only work if a bunch of people watches those closely -- but that's difficult to ensure. One way could be to set up Slack integration (as we've done for the GAP.jl CI jobs) so that failed jobs ping there -- but that still is far inferior to having this directly in the PR jobs...

That said, it may be sufficient to have one nightly job, as usually the Julia nightly regressions manifest as "Oscar does not start anymore" / "Oscar quickly crashes" ?

Anyway, none of this is a strong objection to this PR -- if you guys thought this through, let's do it, be it with 3 or with 4 macOS jobs. We'll just tweak it as we go along. And we should look into getting our own macOS runners, too...

fingolfin (approved):

PR #3164, opened 2024-01-10
CI: add julia-cache action for github ci

benlorenz:


benlorenz:
When re-running jobs this seems to avoid a bit of download and precompile time. Not sure how much this helps for other PRs but I think this is an improvement, and it is suggested by the other julia actions:
```
[[julia-buildpkg] Caching of the julia depot was not detected](https://github.com/oscar-system/Oscar.jl/actions/runs/7474069735/job/20339475113#step:20:46)
Consider using `julia-actions/cache` to speed up runs https://github.com/julia-actions/cache. To ignore, set input `ignore-no-cache: true`
```

.github/workflows/CI.yml
```diff
@@ -70,6 +70,9 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
       - uses: julia-actions/cache@v1
+        with:
+          cache-name: julia-cache;workflow=${{ github.workflow }};julia=${{ matrix.julia-version }}
```
lgoettgens, line 74: I think we should include the os here. Afaik the compiled stuff is not necessarily compatible between Ubuntu and macos

benlorenz: The OS is [always included automatically](https://github.com/julia-actions/cache#optional-inputs):
>  The key body automatically includes the OS

The currently used key for restore is
```
restore-keys: julia-cache;workflow=Run tests;julia=1.6;os=Linux;
```
With the custom cache name I got rid of the different depwarn settings, long+short groups and put test + doctest in the same cache group. Most of the storage will be artifacts and precompiled code that doesn't depend on these values.

benlorenz: The reason why I wanted to merge more caches is that each cache was about 800MB and once we exceed 10GB in total old caches will get deleted.
With (previously) more or less 1 cache per job we quickly lose caches that might still be useful.

lgoettgens:
The cache job supposedly prints if it found a cache somewhere (cf https://github.com/oscar-system/Oscar.jl/actions/runs/7478508023/job/20353618522#step:4:87), but I cannot find the output. Where would I need to look for that?

.github/workflows/CI.yml
lgoettgens: Ah ok. Merging caches is great!

benlorenz:
> The cache job supposedly prints if it found a cache somewhere (cf [oscar-system/Oscar.jl/actions/runs/7478508023/job/20353618522#step:4:87](https://github.com/oscar-system/Oscar.jl/actions/runs/7478508023/job/20353618522#step:4:87)), but I cannot find the output. Where would I need to look for that?

That was the first run after the changed key so it did not find a cache.

I restarted the job and the second run did find a cache, for example here:
https://github.com/oscar-system/Oscar.jl/actions/runs/7478508023/job/20362240520#step:4:84
```
Received 335544320 of 1024916053 (32.7%), 320.3 MBs/sec
Received 675282944 of 1024916053 (65.9%), 321.5 MBs/sec
Received 981467136 of 1024916053 (95.8%), 311.7 MBs/sec
Cache Size: ~977 MB (1024916053 B)
/usr/bin/tar -xf /home/runner/work/_temp/1031198b-683a-480a-81a0-0e3caa790a08/cache.tzst -P -C /home/runner/work/Oscar.jl/Oscar.jl --use-compress-program unzstd
Received 1024916053 of 1024916053 (100.0%), 244.1 MBs/sec
Cache restored successfully
Cache restored from key: julia-cache;workflow=Run tests;julia=1.10;os=Linux;run_id=7478508023;run_attempt=1
Run pyTooling/Actions/with-post-step@adef08d3bdef092282614f3b683897cefae82ee3
```

buildpkg took only 5 seconds on that second run and precompilation was needed only for 17 packages [here](https://github.com/oscar-system/Oscar.jl/actions/runs/7478508023/job/20362240520#step:8:251): 
```
  17 dependencies successfully precompiled in 55 seconds. 127 already precompiled.
```
(maybe that cache was from a doctest run which doesn't include the test-dependencies)

I am hoping that we can reduce these macos precompilations (from the first run):
```
  135 dependencies successfully precompiled in 579 seconds. 36 already precompiled.
```

lgoettgens (approved):

PR #3166, opened 2024-01-10
introduce `SubPcGroup`

ThomasBreuer:
This is a first attempt to implement what was sketched in #2672.
I am not sure that we really want this:
- Admitting different types for a group and a subgroup was apparently not allowed up to now, several signatures of functions have to be "relaxed" in order to support different types.
- Functions returning arrays of subgroups (typically series such as `derived_series`) will now create the full group (if it occurs) with another type; one has to be aware of different types for the same mathematical object.

If we want to proceed like this then the next step is as follows.
- Fix functions like `intersection` which expect arguments of the same type up to now.
- Adjust the doctests.
- Adjust/extend the documentation.
- Add tests.
- Do the same for `FpGroup`/`SubFpGroup`.

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -494,7 +494,7 @@ in general the length of this vector is not minimal.
 
 # Examples
 ```jldoctest
-julia> length(small_generating_set(abelian_group(PcGroup, [2,3,4])))
+julia> length(small_generating_set(abelian_group(SubPcGroup, [2,3,4])))
 2
 
 julia> length(small_generating_set(abelian_group(PermGroup, [2,3,4])))
@@ -517,7 +517,7 @@ Return a vector of minimal length of elements in `G` that generate `G`.
 
 # Examples
 ```jldoctest
-julia> length(minimal_generating_set(abelian_group(PcGroup, [2,3,4])))
+julia> length(minimal_generating_set(abelian_group(SubPcGroup, [2,3,4])))
 2
 
 julia> length(minimal_generating_set(abelian_group(PermGroup, [2,3,4])))
@@ -1368,7 +1368,7 @@ an exception is thrown if `G` is not solvable.
 end
 
 @doc raw"""
-    complement_classes(G::T, N::T) where T <: GAPGroup
+    complement_classes(G::GAPGroup, N::GAPGroup)
 
 Return a vector of the conjugacy classes of complements
 of the normal subgroup `N` in `G`.
@@ -1393,17 +1393,17 @@ julia> complement_classes(G, center(G)[1])
 GAPGroupConjClass{PcGroup, PcGroup}[]
 ```
 """
-function complement_classes(G::T, N::T) where T <: GAPGroup
+function complement_classes(G::T, N::GAPGroup) where T <: GAPGroup
    res_gap = GAP.Globals.ComplementClassesRepresentatives(GapObj(G), GapObj(N))::GapObj
    if length(res_gap) == 0
-     return GAPGroupConjClass{T, T}[]
+     return GAPGroupConjClass{T, sub_type(T)}[]
    else
      return [conjugacy_class(G, H) for H in _as_subgroups(G, res_gap)]
    end
 end
 
 @doc raw"""
-    complements(G::T, N::T) where T <: GAPGroup
+    complements(G::GAPGroup, N::GAPGroup)
 
 Return an iterator over the complements of the normal subgroup `N` in `G`.
 Very likely it is better to use [`complement_classes`](@ref) instead.
@@ -1416,7 +1416,7 @@ julia> describe(first(complements(G, derived_subgroup(G)[1])))
 "C2"
 ```
 """
-complements(G::T, N::T) where T <: GAPGroup = Iterators.flatten(complement_classes(G, N))
+complements(G::GAPGroup, N::GAPGroup) = Iterators.flatten(complement_classes(G, N))
 
 @doc raw"""
     complement_system(G::Group)
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -779,7 +779,7 @@ function Base.rand(C::GroupConjClass{S,T}) where S where T<:GAPGroup
 end
 
 function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S where T<:GAPGroup
-   return _oscar_group(GAP.Globals.Random(GAP.wrap_rng(rng), C.CC), acting_group(C))
+   return _oscar_subgroup(GAP.Globals.Random(GAP.wrap_rng(rng), C.CC), acting_group(C))
 end
 
 """
@@ -928,7 +928,7 @@ Permutation group of degree 4 and order 3
 """
 function conjugate_group(G::T, x::GAPGroupElem) where T <: GAPGroup
   @req check_parent(G, x) "G and x are not compatible"
-  return _oscar_group(GAPWrap.ConjugateSubgroup(GapObj(G), GapObj(x)), G)
+  return _oscar_subgroup(GAPWrap.ConjugateSubgroup(GapObj(G), GapObj(x)), G)
 end
 
 Base.:^(H::GAPGroup, y::GAPGroupElem) = conjugate_group(H, y)
@@ -1917,7 +1917,7 @@ function map_word(g::PcGroupElem, genimgs::Vector; genimgs_inv::Vector = Vector(
   end
   gX = GapObj(g)
 
-  if GAP.Globals.IsPcGroup(GapObj(G))
+  if GAPWrap.IsPcGroup(GapObj(G))
     l = GAP.Globals.ExponentsOfPcElement(GAP.Globals.FamilyPcgs(GapObj(G)), gX)
   else  # GAP.Globals.IsPcpGroup(GapObj(G))
     l = GAP.Globals.Exponents(gX)
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -245,7 +245,7 @@ Base.:*(x::GAPGroupElem, y::GAPGroupElem) = _prod(x, y)
 
 ==(x::GAPGroup, y::GAPGroup) = GapObj(x) == GapObj(y)
 
-==(x::T, y::T) where T <: BasicGAPGroupElem = GapObj(x) == GapObj(y)
+==(x::BasicGAPGroupElem, y::BasicGAPGroupElem ) = GapObj(x) == GapObj(y)
 
 """
     one(G::GAPGroup) -> elem_type(G)
@@ -848,9 +848,10 @@ julia> maximal_subgroup_classes(G)
 """
 @gapattribute function maximal_subgroup_classes(G::GAPGroup)
   L = Vector{GapObj}(GAP.Globals.ConjugacyClassesMaximalSubgroups(GapObj(G))::GapObj)
-  T = typeof(G)
+  TG = typeof(G)
+  TS = sub_type(TG)
   LL = [GAPGroupConjClass(G, _as_subgroup_bare(G, GAPWrap.Representative(cc)), cc) for cc in L]
-  return Vector{GAPGroupConjClass{T, T}}(LL)
+  return Vector{GAPGroupConjClass{TG, TS}}(LL)
 end
 
 """
@@ -1158,7 +1159,7 @@ Return `N, f`, where `N` is the normalizer of `H` in `G`,
 i.e., the largest subgroup of `G` in which `H` is normal,
 and `f` is the embedding morphism of `N` into `G`.
 """
-normalizer(G::T, H::T) where T<:GAPGroup = _as_subgroup(G, GAPWrap.Normalizer(GapObj(G), GapObj(H)))
+normalizer(G::GAPGroup, H::GAPGroup) = _as_subgroup(G, GAPWrap.Normalizer(GapObj(G), GapObj(H)))
 
 """
     normalizer(G::Group, x::GAPGroupElem)
@@ -2275,7 +2276,12 @@ function describe(G::FPGroup)
 
    if !has_is_finite(G)
       # try to obtain an isomorphic permutation group, but don't try too hard
-      iso = GAP.Globals.IsomorphismPermGroupOrFailFpGroup(GapObj(G), 100000)::GapObj
+#TODO: With GAP 4.13.0, the prescribed bound 100000 will cause a test failure.
+#      This regression will hopefully be fixed in GAP 4.13.1,
+#      see https://github.com/gap-system/gap/issues/5697
+#      and https://github.com/gap-system/gap/pull/5698.
+#     iso = GAP.Globals.IsomorphismPermGroupOrFailFpGroup(GapObj(G), 100000)::GapObj
+      iso = GAP.Globals.IsomorphismPermGroupOrFailFpGroup(GapObj(G))::GapObj
       iso != GAP.Globals.fail && return describe(PermGroup(GAPWrap.Range(iso)))
    elseif is_finite(G)
       return describe(PermGroup(G))
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -330,10 +330,11 @@ function Base.show(io::IO, G::PermGroup)
   end
 end
 
-function Base.show(io::IO, G::PcGroup)
+function Base.show(io::IO, G::Union{PcGroup,SubPcGroup})
   @show_name(io, G)
   @show_special(io, G)
-  print(io, "Pc group")
+  T = typeof(G) == PcGroup ? "Pc group" : "Sub pc group"
+  print(io, T)
   if !is_terse(io)
     if isfinite(G)
       print(io, " of order ", order(G))
@@ -1169,7 +1170,7 @@ Return `C, f`, where `C` is the normal core of `H` in `G`,
 that is, the largest normal subgroup of `G` that is contained in `H`,
 and `f` is the embedding morphism of `C` into `G`.
 """
-core(G::T, H::T) where T<:GAPGroup = _as_subgroup(G, GAPWrap.Core(GapObj(G), GapObj(H)))
+core(G::GAPGroup, H::GAPGroup) = _as_subgroup(G, GAPWrap.Core(GapObj(G), GapObj(H)))
 
 """
     normal_closure(G::Group, H::Group)
@@ -1180,7 +1181,7 @@ and `f` is the embedding morphism of `N` into `G`.
 
 Note that `H` must be a subgroup of `G`.
 """
-normal_closure(G::T, H::T) where T<:GAPGroup = _as_subgroup(G, GAPWrap.NormalClosure(GapObj(G), GapObj(H)))
+normal_closure(G::GAPGroup, H::GAPGroup) = _as_subgroup(G, GAPWrap.NormalClosure(GapObj(G), GapObj(H)))
 
 # Note:
 # GAP admits `NormalClosure` also when `H` is not a subgroup of `G`,
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1390,7 +1390,7 @@ julia> G = dihedral_group(8)
 Pc group of order 8
 
 julia> complement_classes(G, center(G)[1])
-GAPGroupConjClass{PcGroup, PcGroup}[]
+GAPGroupConjClass{PcGroup, SubPcGroup}[]
 ```
 """
 function complement_classes(G::T, N::GAPGroup) where T <: GAPGroup
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -186,7 +186,7 @@ function Base.rand(rng::Random.AbstractRNG, G::GAPGroup)
    return group_element(G, s)
 end
 
-function Base.rand(rng::Random.AbstractRNG, rs::Random.SamplerTrivial{Gr}) where Gr<:Oscar.GAPGroup
+function Base.rand(rng::Random.AbstractRNG, rs::Random.SamplerTrivial{Gr}) where Gr<:GAPGroup
    return rand(rng, rs[])
 end
 
@@ -234,6 +234,9 @@ end
 
 #We need a lattice of groups to implement this properly
 function _prod(x::T, y::T) where T <: GAPGroupElem
+#T not nec. same type,
+#T and for pc subgroups may need to go to the big group
+#T (write tests that model this situation)
   G = _common_parent_group(parent(x), parent(y))
   return group_element(G, GapObj(x)*GapObj(y))
 end
@@ -286,6 +289,7 @@ function Base.show(io::IO, G::FPGroup)
     end
   else
     print(io, "Finitely presented group")  # FIXME: actually some of these groups are *not* finitely presented
+#T introduce SubFPGroup
     if !is_terse(io)
     if has_order(G)
       if is_finite(G)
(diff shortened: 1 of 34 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -1065,7 +1066,7 @@ Return `C, f`, where `C` is the normal core of `H` in `G`,
 that is, the largest normal subgroup of `G` that is contained in `H`,
 and `f` is the embedding morphism of `C` into `G`.
 """
-core(G::T, H::T) where T<:GAPGroup = _as_subgroup(G, GAPWrap.Core(G.X, H.X))
+core(G::T1, H::T2) where {T1<:GAPGroup, T2<:GAPGroup} = _as_subgroup(G, GAPWrap.Core(G.X, H.X))
```
```suggestion
core(G::GAPGroup, H::GAPGroup) = _as_subgroup(G, GAPWrap.Core(G.X, H.X))
```
Your change makes it actually a lot easier

fingolfin:
> This is a first attempt to implement what was sketched in #2672. I am not sure that we really want this:
> 
> * Admitting different types for a group and a subgroup was apparently not allowed up to now, several signatures of functions have to be "relaxed" in order to support different types.

That seems fine to me.

> * Functions returning arrays of subgroups (typically series such as `derived_series`) will now create the full group (if it occurs) with another type; one has to be aware of different types for the same mathematical object.

I also think this is fine, and in fact just reflects the reality in GAP, which tries to paper of it but doesn't really succeed. So I actually think it is *better* that we make this explicit. This way at least the user can immediately tell if a given group is the "full" group or not, and we have a chance of explaining the situation in the docstring for `SubPcGroup`.

> 
> If we want to proceed like this then the next step is as follows.
> 
> * Fix functions like `intersection` which expect arguments of the same type up to now.
> * Adjust the doctests.
> * Adjust/extend the documentation.
> * Add tests.
> * Do the same for `FpGroup`/`SubFpGroup`.

src/Groups/GAPGroups.jl
```diff
@@ -321,10 +321,11 @@ function Base.show(io::IO, G::PermGroup)
   end
 end
 
-function Base.show(io::IO, G::PcGroup)
+function Base.show(io::IO, G::Union{PcGroup,SubPcGroup})
   @show_name(io, G)
   @show_special(io, G)
-  print(io, "Pc group")
+  T = typeof(G) == PcGroup ? "Pc group" : "Sub pc group"
```
fingolfin: Maybe like this?
```suggestion
  T = typeof(G) == PcGroup ? "Pc group" : "Subgroup of pc group"
```

src/Groups/GAPGroups.jl
fingolfin: It makes the signatures easier (with Lars' extra change). Of course we loose a bit of "validation", i.e. previously Julia prevented some situations from occurring (e.g. you could not mix a permutation group with a matrix group), and now we have to catch that ourselves. But that's not so bad in the end, we already have to catch some of these "bad" situations, e.g. when both groups are matrix groups but defined over different rings (perhaps in different characteristic)

src/Groups/group_constructors.jl
```diff
@@ -110,11 +117,12 @@ function cyclic_group(::Type{T}, n::Union{IntegerUnion,PosInf}) where T <: GAPGr
   return T(GAP.Globals.CyclicGroup(_gap_filter(T), GAP.Obj(n))::GapObj)
 end
 
-function cyclic_group(::Type{PcGroup}, n::Union{IntegerUnion,PosInf})
+# special handling for pc groups: distinguish on the GAP side
+function cyclic_group(::Type{T}, n::Union{IntegerUnion,PosInf}) where T <: Union{PcGroup, SubPcGroup}
```
fingolfin, line 121: This creates a new `PcGroup`, which would be the "whole" group, so should we really allow `SubPcGroup` here?

src/Groups/homomorphisms.jl
```diff
@@ -517,6 +524,29 @@ function isomorphism(::Type{T}, G::GAPGroup) where T <: Union{FPGroup, PcGroup,
    end::GAPGroupHomomorphism{typeof(G), T}
 end
 
+# If `G` is not a full pc group then switch to a full pc group in Oscar.
```
fingolfin, line 594: This, by the way, is something that I needed a lot in the past, and was always annoyingly complicated in GAP. So this is a nice bonus I did not even consider (at least I don't recall considering it :-) )

src/Groups/homomorphisms.jl
```diff
@@ -517,6 +524,29 @@ function isomorphism(::Type{T}, G::GAPGroup) where T <: Union{FPGroup, PcGroup,
...
+   # Known isomorphisms are cached in the attribute `:isomorphisms`.
+   isos = get_attribute!(Dict{Type, Any}, G, :isomorphisms)::Dict{Type, Any}
+   return get!(isos, T) do
+     fun = _get_iso_function(T)
+     f = fun(G.X)::GapObj
+     @req f !== GAP.Globals.fail "Could not convert group into a group of type $T"
+     # The codomain of `f` can be a *subgroup* of a full pc group.
+     # In this situation, we have to switch to a full pc group.
+     C = GAP.Globals.Range(f)::GapObj
+     if GAPWrap.GeneratorsOfGroup(C) != GAP.Globals.FamilyPcgs(C)
```
fingolfin: What about `pcp` groups, they don't use `FamilyPcgs` and probably needs separate code?

src/Groups/homomorphisms.jl
```diff
@@ -741,12 +771,14 @@ end
     GrpAbFinGen(G::T) where T <: GAPGroup
     PcGroup(G::T) where T <: Union{GAPGroup, GrpAbFinGen}
     pc_group(G::T) where T <: Union{GAPGroup, GrpAbFinGen}
+    SubPcGroup(G::T) where T <: Union{GAPGroup, GrpAbFinGen}
+    sub_pc_group(G::T) where T <: Union{GAPGroup, GrpAbFinGen}
```
fingolfin: Do we really want/need this? Maybe `SubPcGroup` on a `PcGroup` could be useful, in case one needs the "full group as a subgroup", but other than that?

src/Groups/types.jl
```diff
@@ -197,6 +229,53 @@ as shown in Section [Elements of groups](@ref elements_of_groups).
...
+
+# Examples
+
+```jldoctest
+
+... hier! ...
+
+```
+"""
+const SubPcGroupElem = BasicGAPGroupElem{SubPcGroup}
```
fingolfin, line 280: To me the most annoying part of this PR actually seems to be `SubPcGroupElem`; we probably would want to allow multiplying a `SubPcGroupElem` by a `PcGroupElem` (and vice-versa), and possibly in various other contexts one may have to / wish to support both?

Perhaps a suitable `promote_rule` can help a bit?

src/Groups/homomorphisms.jl
ThomasBreuer: I think GAP is just missing something like `IsomorphismFullPcGroup`.

src/Groups/sub.jl
```diff
@@ -850,7 +858,7 @@ julia> typeof(quo(PermGroup, G, N)[1])
 PermGroup
 ```
 """
-function quo(G::T, N::T) where T <: GAPGroup
+function quo(G::T1, N::T2) where {T1 <: GAPGroup, T2 <: GAPGroup}
```
fingolfin: We could add a helper
```julia
_is_subgroup_type(::Type{T1}, ::Type{T2}) where {T1, T2} = false
_is_subgroup_type(::Type{T}, ::Type{T}) where {T <: GAPGroup} = true
_is_subgroup_type(::Type{PcGroup}, ::Type{SubPcGroup}) = true
```
and then add checks like
```julia
  @req _is_subgroup_type(T1, T2) "argument types do not fit"
```

fingolfin:
Just to document it somewhere in writing: We discussed this some time ago and the verdict was that yes, we want to proceed in this direction; while there are downsides, many of them actually already exists, but just are hidden; and there are also upsides. So...

src/Groups/sub.jl
ThomasBreuer: I thought about introducing something like `_compatibility_check(G1::T1, G2::T2)` which throws an exception if the arguments do not fit together, and otherwise has no effect. Note that checking the types alone is not sufficient, for example in the case of two f.p. subgroups or two pc subgroups which belong to different full groups. If we do not check this compatibility before delegating to GAP then we get just a cryptic error message from GAP.

fingolfin:
@ThomasBreuer I hope this is still on your radar? :-)

ThomasBreuer:
The failure of one of the book examples shows a conceptual problem. (Likely this example can be turned back to the old behaviour by an improvement on the GAP side, but this does not solve the general problem.)

The function `quo` for two groups states that the type of the return value depends on the situation: Anyhow a new group must be created, and GAP decides about a suitable representation; if the result is known to be a finite solvable group then a pc group is returned, otherwise a permutation group or a fp group are possible.

The GAP function `NaturalHomomorphismByNormalSubgroup` is allowed to return a mapping that is not (known to be) surjective. If this happens and if the `Range` of the mapping is a pc group then the quotient on the Oscar side can be only a `SubPcGroup`. (At some extra cost, we can check the surjectivity, and create a `PcGroup` in Oscar if yes.)

It looks as if we have to do something more specific on the GAP side if we want the results of `quo` to be either `PcGroup` or `FPGroup` or `PermGroup`.

(@fieker @fingolfin ?)

ThomasBreuer:
The failure in `test (1.10, book, ubuntu-latest)` is due to an intended change of the output shown: The group in question has now a different type.

benlorenz:
> The failure in `test (1.10, book, ubuntu-latest)` is due to an intended change of the output shown: The group in question has now a different type.

Then you can adjust the output in (`test/book/cornerstones/groups/actions.jlcon`) to the new printing. Only the input must stay identical for 1.1 (i.e. current master).

src/Groups/group_constructors.jl
fingolfin: @ThomasBreuer did you see my question from January there? I don't terribly mind having this (it may make some generic code easier to write), but I wonder: does `cyclic_group(SubPcGroup, 5)` really work correctly, and return something of type `SubPcGroup`? We don't have a test for that, do we?

src/Groups/homomorphisms.jl
```diff
@@ -568,6 +576,57 @@ function isomorphism(::Type{FPGroup}, G::GAPGroup; on_gens::Bool=false)
...
+       if is_finite(G)
+         fam = GAP.Globals.ElementsFamily(GAP.Globals.FamilyObj(G.X))::GapObj
+         Ggens = GapObj([x.X for x in gens(G)])::GapObj
+         Cpcgs = GAP.Globals.PcgsByPcSequence(fam, Ggens)::GapObj
+         CC = GAP.Globals.PcGroupWithPcgs(Cpcgs)::GapObj
+         CCpcgs = GAP.Globals.FamilyPcgs(CC)::GapObj
+         f = GAP.Globals.GroupHomomorphismByImages(G, CC, Cpcgs, CCpcgs)::GapObj
+         return GAPGroupHomomorphism(G, T(CC), f)
+       else
+         f = GAP.Globals.IsomorphismPcpGroup(G.X)::GapObj
```
fingolfin:
```suggestion
         f = GAP.Globals.IsomorphismPcpGroup(GapObj(G))::GapObj
```
possibly also similar (i.e. `FOO.X` to `GapObj(FOO)`) elsewhere?

src/Groups/homomorphisms.jl
```diff
@@ -568,6 +576,57 @@ function isomorphism(::Type{FPGroup}, G::GAPGroup; on_gens::Bool=false)
...
+   # Known isomorphisms are cached in the attribute `:isomorphisms`.
+   isos = get_attribute!(Dict{Tuple{Type, Bool}, Any}, G, :isomorphisms)::Dict{Tuple{Type, Bool}, Any}
+   return get!(isos, (T, false)) do
+     if on_gens
+       # Throw an exception if `gens(G)` is not a polycyclic sequence for `G`.
+       # Create a group in `IsPcGroup` if `G` is finite,
+       # and a group in `IsPcpGroup` otherwise.
+       if is_finite(G)
+         fam = GAP.Globals.ElementsFamily(GAP.Globals.FamilyObj(G.X))::GapObj
+         Ggens = GapObj([x.X for x in gens(G)])::GapObj
```
fingolfin: Huh, does this line work as intended? There is no test case covering this code according to Codecov, and I have some doubts: won't this create a GAP list of OSCAR elements?

Perhaps like this:
```suggestion
         Ggens = GapObj(gens(G); recursive=true)
```

src/Groups/homomorphisms.jl
```diff
@@ -622,18 +681,32 @@ function isomorphism(::Type{T}, A::FinGenAbGroup) where T <: GAPGroup
...
-     # the isomorphic gap group
-     G = abelian_group(T, exponents)
-     # `GAPWrap.GeneratorsOfGroup(G.X)` consists of independent elements
+     # Create an isomorphic GAP group whose `GAPWrap.GeneratorsOfGroup`
+     # consists of independent elements of the orders in `exponents`.
+     # (We cannot guarantee that these generators form a pcgs in the case
+     # `T == PcGroup`, hence we cannot call `abelian_group(T, exponents)`.)
+     if T == PcGroup
+       if 0 in exponents
+         GapG = GAP.Globals.AbelianPcpGroup(length(exponents), GAP.GapObj(exponents, recursive=true))
```
fingolfin, line 709: I guess we can drop the `GAP.` from `GAP.GapObj` here and elsewhere? Then again we still use it in a bunch of places, so we could change that in a later follow-up PR.
```suggestion
         GapG = GAP.Globals.AbelianPcpGroup(length(exponents), GapObj(exponents, recursive=true))
```

src/Groups/homomorphisms.jl
```diff
@@ -811,12 +885,13 @@ end
...
 pc_group(G::T) where {T <: Union{FinGenAbGroup, GAPGroup, MultTableGroup}} = PcGroup(G)
+sub_pc_group(G::T) where {T <: Union{FinGenAbGroup, GAPGroup, MultTableGroup}} = SubPcGroup(G)
 permutation_group(G::T) where {T <: Union{FinGenAbGroup, GAPGroup, MultTableGroup}} = PermGroup(G)
 
 # Now for MultTableGroup
 
-# Remove once Hecke lower bound is >= 0.14.1
-@attributes MultTableGroup
+# # Remove once Hecke lower bound is >= 0.14.1
+# @attributes MultTableGroup
```
fingolfin:
```suggestion
```

src/Groups/matrices/MatGrp.jl
```diff
@@ -435,6 +435,7 @@ function Base.:^(x::MatrixGroupElem, y::MatrixGroupElem)
 end
 
 comm(x::MatrixGroupElem, y::MatrixGroupElem) = inv(x)*conj(x,y)
+#T why needed? GAPGroupElem has x^-1*x^y
```
fingolfin, line 438: If we don't need it, remove it? (Though perhaps in a separate PR?)

Though then again, now I wonder if the general one shouldn't be re-implemented by calling `GAP.Globals.Comm`? And then it might make sense to keep the above method here... Or even change it to, say, `inv(y*x)*(x*y)` (one less inversion, which is arguably the most expensive bit here?)

src/Groups/sub.jl
```diff
@@ -175,13 +188,13 @@ julia> normal_subgroups(symmetric_group(5))
...
- Pc group of order 8
- Pc group of order 4
- Pc group of order 4
- Pc group of order 4
- Pc group of order 2
- Pc group of order 1
+6-element Vector{SubPcGroup}:
+ Subgroup of pc group of order 8
+ Subgroup of pc group of order 4
+ Subgroup of pc group of order 4
```
fingolfin, line 194: Hmmm... this is ambiguous, isn't it? The `4` refers to the order of the subgroup, but a reader might reasonably interpret this as saying the parent group has order 4... :-/.

(How) can we make this clearer? Perhaps We should call it a `Sub-pc group` (and explain that somewhere)? 

Or use a different phrasing: `Subgroup of pc group, has order 4` -- alas that would be different from how this kind of info is printed for almost anything else in OSCAR, wouldn't it?

Perhaps @thofma or @fieker have some brilliant insight on this.

src/Groups/sub.jl
```diff
@@ -792,20 +810,26 @@ julia> typeof(quo(PermGroup, G, N)[1])
...
   # The call may have found out new information about `GapObj(G)`,
   # for example that `GapObj(G)` is finite.
 #FIXME: The GAP function should deal with this situation.
   GAP.Globals.UseSubsetRelation(GapObj(G), GapObj(N))
-  cod = GAP.Globals.ImagesSource(mp)::GapObj
+#T HACK: In order to avoid `SubPcGroup`s as return values of `quo`
+#T       where possible, take the *full* pc group if the range is a pc group
+#T       and if the GAP mapping is surjective.
+  cod = GAP.Globals.Range(mp)::GapObj
+  if !(GAP.Globals.IsPcGroup(cod) && GAP.Globals.IsSurjective(mp))
```
fingolfin:
```suggestion
  if !(GAPWrap.IsPcGroup(cod) && GAPWrap.IsSurjective(mp))
```

src/Groups/sub.jl
```diff
@@ -792,20 +810,26 @@ julia> typeof(quo(PermGroup, G, N)[1])
...
   mp = GAP.Globals.NaturalHomomorphismByNormalSubgroup(GapObj(G), GapObj(N))::GapObj
   # The call may have found out new information about `GapObj(G)`,
   # for example that `GapObj(G)` is finite.
 #FIXME: The GAP function should deal with this situation.
   GAP.Globals.UseSubsetRelation(GapObj(G), GapObj(N))
-  cod = GAP.Globals.ImagesSource(mp)::GapObj
+#T HACK: In order to avoid `SubPcGroup`s as return values of `quo`
+#T       where possible, take the *full* pc group if the range is a pc group
+#T       and if the GAP mapping is surjective.
+  cod = GAP.Globals.Range(mp)::GapObj
```
fingolfin:
```suggestion
  cod = GAPWrap.Range(mp)
```

src/Groups/sub.jl
```diff
@@ -792,20 +810,26 @@ julia> typeof(quo(PermGroup, G, N)[1])
...
+  if !(GAP.Globals.IsPcGroup(cod) && GAP.Globals.IsSurjective(mp))
+    cod = GAP.Globals.ImagesSource(mp)::GapObj
```
fingolfin: Maybe "GAP wrap" `ImagesSource`, too?

src/Groups/types.jl
```diff
@@ -275,8 +357,19 @@ end
...
+# `_oscar_group(obj, G)` is used to create the subgroup of `G`
+# that is described by the GAP group `obj`;
 # default: ignore `G`
-_oscar_group(obj::GapObj, G::T) where T <: GAPGroup = T(obj)
+function _oscar_group(obj::GapObj, G::GAPGroup)
+  S = sub_type(G)(obj)
+  @assert GAP.Globals.FamilyObj(S.X) === GAP.Globals.FamilyObj(G.X)
+  return S
+end
+#T better rename to _oscar_subgroup?
```
fingolfin, line 373: I agree: `_oscar_subgroup` would be better

src/Groups/types.jl
```diff
@@ -302,10 +395,12 @@ function _oscar_group(obj::GapObj, G::MatrixGroup)
   return M
 end
 
+
 ################################################################################
 #
 # "Coerce" an Oscar group `G` to one that is compatible with
 # the given Oscar group `S`.
+#T what does compatible mean?
 compatible_group(G::T, S::T) where T <: GAPGroup = _oscar_group(GapObj(G), S)
```
fingolfin: This is not used anywhere, is it? Can we just remove it?

test/Groups/homomorphisms.jl
```diff
@@ -580,8 +582,9 @@ function test_kernel(G,H,f)
    K,i = kernel(f)
    Im = image(f)[1]
 
-   @test preimage(f,H)==(G,id_hom(G))
-   @test preimage(f,sub(H,[one(H)])[1])==(K,i)
+#TODO: activate these tests as soon as they pass again
```
fingolfin: What is the issue here? Can we perhaps at least record that in the comment?
I.e. do these run into errors? Or return wrong results? Or something else? If there is a relevant issue, link to it here?

test/Groups/libraries.jl
```diff
@@ -139,7 +139,8 @@ end
 
 @testset "Small groups" begin
    L = all_small_groups(8)
-   LG = [abelian_group(PcGroup,[2,4]), abelian_group(PcGroup,[2,2,2]), cyclic_group(8), quaternion_group(8), dihedral_group(8)]
+#  LG = [abelian_group(PcGroup,[2,4]), abelian_group(PcGroup,[2,2,2]), cyclic_group(8), quaternion_group(8), dihedral_group(8)]
```
fingolfin, line 144: Is there a point in keep the "old" copies of the test around, here and below? Otherwise just delete them?

test/Groups/quotients.jl
```diff
@@ -52,14 +53,16 @@ end
    #   otherwise `PcGroup` if finite, `FPGroup` if not
    # - `maximal_abelian_quotient` with prescribed type:
    #    return a group of this type if possible
-   for T in [ PermGroup, PcGroup ]
+#  for T in [ PermGroup, PcGroup ]
+   for T in [ PermGroup, SubPcGroup ]
```
fingolfin, line 58: Why not
```suggestion
   for T in [ PermGroup, PcGroup, SubPcGroup ]
```

fingolfin (commented):
Some remarks. Most are optional (better to get this merged sooner and then polish afterwards), but there are one or two were I wonder if the code is correct as it is...?

src/Groups/homomorphisms.jl
ThomasBreuer: `gens(G)` consists of OSCAR elements, the array of the `x.X` consists of GAP elements.
I am going to add tests.
(And yes, the code you propose is better.)

src/Groups/homomorphisms.jl
ThomasBreuer: O.k., I change all those `.X` introduced in this pull request; if I change also others then there will likely be conflicts.

src/Groups/homomorphisms.jl
ThomasBreuer: Yes, we can write just `GapObj` because meanwhile it is an exported variable.

src/Groups/matrices/MatGrp.jl
ThomasBreuer: Let us leave this to another pull request.

src/Groups/types.jl
ThomasBreuer: Yes.

test/Groups/quotients.jl
ThomasBreuer: The point is that currently `abelian_group(PcGroup, [2, 3, 4])` is not supported.
`abelian_group` promises a result whose generators correspond to the given orders,
and the type `PcGroup` implies that the generators are a polycyclic generating sequence w.r.t. the underlying presentation.
With GAP's `IsPcGroup`, we cannot get this; and `IsPcpGroup` cannot be used because of some bugs.

(I will add a comment.)

test/Groups/libraries.jl
ThomasBreuer: I am still confident that eventually we can support `abelian_group(PcGroup, [2, 4])`.
I will add a comment.

test/Groups/homomorphisms.jl
ThomasBreuer: I will add a comment.

src/Groups/types.jl
ThomasBreuer: O.k., I will rename the two-argument version.

fingolfin:
Unfortunately this has conflicts again now :-(

It would be great if we could merge this tomorrow (before or after triage, whenever you have time to resolve it) so we can move on :-)

src/Groups/sub.jl
fingolfin: To be clear: we can address this in a follow-up PR (but then let's make an issue about this after the PR is merged so so we don't forget)

ThomasBreuer:
> Unfortunately this has conflicts again now :-(

I find it rather surprising that there were not more conflicts.

fingolfin:
@ThomasBreuer there is one unresolved review comment of mine from quite some ago where I ask about `sub_pc_group` -- can you please reply to that?

src/Groups/sub.jl
ThomasBreuer: I have opened issue  #3732.

src/Groups/group_constructors.jl
ThomasBreuer: Yes, `cyclic_group(SubPcGroup, 5)` works.
(I had intended to add a test for this, but somehow this got lost.)

fingolfin (approved):

PR #3172, opened 2024-01-10
Rename `are_algebraically_independent`

lgoettgens:
~~Based on https://github.com/oscar-system/Oscar.jl/pull/3170 to avoid conflicts in `Deprecations.jl`, just ignore the first two commits here.~~

The third commit contains the renaming `are_algebraically_independent` -> `is_algebraically_independent` as discussed in today's coding spring. 
However, this is as `is_*` function that returns additional stuff, and should thus in my opinion be better named `is_algebraically_independent_with_relations`. The fourth commit contains this (and introduces a new `is_algebraically_independent` returning just a boolean).
Applying both commits at once, deprecations work just fine. But if one only applies the `are`->`is` change now, the second change would be a breaking change. The part after the `with_` can be changed later using deprecations as well.

@fingolfin What do you think? Should we make this change right now? A later change to `with_data` (as wished by C) would be possible anyway, so I wouldn't see a reason against it. (but please not before 0.14.0)

Maybe the function should actually have a kwarg for obtaining a ring to construct the relation ideal in, but this can easily added in a non-breaking way once needed, so I didn't bother right now.


fingolfin:
I have no strong opinions on this. But for pragmatic reasons I would rename this to `is_algebraically_independent_with_relations` (or `_with_data`) in one go, with deprecations, and only add `is_algebraically_independent` some time later, if at all (perhaps only if someone asks for it). I.e. if we merge this now, it'd be in 0.15 (hopefully out in a few weeks) and we can replace the deprecation by a separate `is_algebraically_independent` e.g. in 1.0 (or 0.16 or 1.1 or never)

lgoettgens:
Ok. So you want me to remove the newly introduced `is_algebraically_independent`?
As the functionality didn't exist before, there is no need for a deprecation.

fingolfin (approved):

Issue #3054, opened 2023-11-29
Roadmap 1.0 and beyond


fieker:
## Deadlines

- Oscar book needs to be ready **31 January 2024** (or maybe even earlier)
- anything relevant for the book (and examples in it) should be settled before
- the proposal is submitted **1 May 2024**, so 1.0 should be out *before that*
- some improvements can be done afterwards but still should be done ASAP so that referees will see them

## Plan

Kaiserslautern: two regular meetings:
- triage: Wednesday 11:30-12:20 (planning; review and resolve issues and PRs on Oscar, Nemo, AA, ...)
- code sprint: TBD (possibly Wednesday or Thursday 13:30-16:00)
- it will be possible to join either via Zoom

## Things critical for the book (end of January!!)

Examples in the book should match what we get in 1.0 even though 1.0 won't be finished in time for the book deadline. That means a few items should be done before; and if we can't do them, then we may have to "lie" in the book in a few examples (i.e. edit output to match what it is supposed to be in 1.0, not what it is right now)

 - [NEW HackMD for renaming](https://hackmd.io/nRnyfrSxTVe5CzXidB1cBQ)
 - finalize type and function names (at least any user facing)
   - relevant issues:
     - https://github.com/oscar-system/Oscar.jl/issues/1826
     - https://github.com/Nemocas/AbstractAlgebra.jl/issues/1278
     - https://github.com/Nemocas/Nemo.jl/issues/1394
     - https://github.com/Nemocas/AbstractAlgebra.jl/issues/1306
     - [HackMD](https://hackmd.io/0hrBHIUrRZG23xiZZ0NCEA)
   - `Abs..` in type names can indicate "abstract" or "absolute". It should be only one
     - there are many abstract types without the Abs, what is the plan? 
       - abstract type `ProjSpc`, but abstract type `AbsAffineAlgebraicSet`
       - abstract type `ModuleFP`, abstract type `AbstractFreeModElem`
       - abstract type `OscarMap`
       - `AbstractLat`
     - *concrete* types `AbstractVariety` etc.
   - `acb`, `arb` and friends and relatives in Nemo need proper names
   - `CoveringMorphism` -> `CoveringMor`? rules for Mor(phism)?
   - Inv and Invar in type names
   - _dec vs Dec
   - Mod and Module 
   - `FPModule` vs `ModuleFP`
   - GrpAb -> AbGroup? and other Grp in Hecke and below
[...]

PR #3174, opened 2024-01-11
Renaming for localizations

HechtiDerLachs:
Some work for #3054 .

 - un-export a lot of concrete types in the localizations
 - rename `ambient_ring` to simply `ring` 
 - un-export some functions


thofma:
If we have a localized module, how do we get the underlying module back? What is the name of the function?

P.S.: I am just worried about some mismatch, because we have `ring`, but we cannot have a function named `module`.

fingolfin:
Could you say a bit about the motivation for renaming to `ambient_ring` to `ring` (I am not opposed, just wondering)

HechtiDerLachs:
@thofma : There has never been a function to get the underlying module back, really. Do you think we need it? One problem is that the module does not know that it came from a localization in the first place. 

@fingolfin : I just recalled various talks with @fieker suggesting to rename `base_ring` just into `ring`. The name `base_ring` might be too established to do this on such short notice. But at least for `ambient_ring` I thought, I could give it a try.

thofma:
I think `ring` is a rather unfortunate name. It would be similar to calling the $R$ in $R[x]$ *the* ring of $R[x]$.

HechtiDerLachs:
But we're not talking about renaming `base_ring` by `ring` here. I renamed `ambient_ring` for multiplicative sets by `ring`. I think that's unambiguous, isn't it?

thofma:
Sorry, my bad. I was confused.

lgoettgens:
Could you add deprecations for your renaming?

HechtiDerLachs:
There is only one command that has only changed its name. For the rest I simply don't want to export it anymore. Is this legitimate? Back in the days when I wrote the code, I simply didn't have a feeling for it.

lgoettgens:
If there are other dispatches for `ambient_ring`, you should use `@deprecate` with a concrete signature instead. It's best if you have a look at deprecations.jl (and maybe put the deprecation there)

HechtiDerLachs:
@bschroter : It seems that you are using `ambient_ring`  for your realization spaces, too. Could/should this also be changed to `ring`?

HechtiDerLachs:
From my side this can be merged. My question to @bschroter was more a heads up and can be resolved elsewhere.

simonbrandhorst (approved):

dcorey2814:
> @bschroter : It seems that you are using `ambient_ring` for your realization spaces, too. Could/should this also be changed to `ring`?

`ambient_ring` is referring to the polynomial ring where the defining ideal and multiplicative semigroup live. `ring` could be ambiguous since one may think this produces the coordinate ring of the matroid realization space.

PR #3175, opened 2024-01-11
Renaming for schemes

HechtiDerLachs:
Continuation of #3174 .


HechtiDerLachs:
I made the function `SpecMor` available as `affine_morphism`. @afkafkafk13 @simonbrandhorst : Any objections?

Edit: `hom` seemed inappropriate to me, since we're not dealing with algebraic structures. `morphism` alone seemed to general and `morphism_of_affine_schemes` too long. We could later on extend this by `covering_morphism`, `covered_morphism`, etc. to also get rid of the constructors `CoveringMorphism` and `CoveredSchemeMorphism`.

However, we could also unify all of these to different methods of a single `morphism` function. Then we should probably consult with people from other areas whether or not we risk taking away this name from them. 

@fieker @fingolfin : What do you think about `morphism` as something like a "geometric counterpart to `hom`"?

simonbrandhorst:
To me `morphism` and `rational_map` seem like the most natural choice.
If one wants more control on the return types, use the CamelCase constructor coming from the type name?

HechtiDerLachs:
After today's brief discussion in the meeting I switched everything to `morphism` now. We can later look into streamlining this with `hom` via aliases, eventually.

HechtiDerLachs:
@fieker @fingolfin @thofma @afkafkafk13 : Do you have objections introducing `morphism` like it is done here? 

If your answer is "yes" in general: Does the above implementation have any downsides that I am not aware of which forbid eventually fusing it with the existing `hom` later on (in some appropriate way to be discussed on, say, Wednesday)?

Edit: If your answer is "no", then I'd be happy to discuss this on Wednesday, too. Maybe we don't have to go too deep in discussions on GH now in this case, so just give me a heads up that you would like to speak more about it.

afkafkafk13:
I do not have any objections to changing it to 'morphism' -- the name feels very natural to me. From the usability perspective, I am not worried about the potential use of 'morphism' in other parts of OSCAR, because we follow mathematical terminology in all these contexts and the signature will disambiguate, if necessary.

thofma:
Fine with me. Just a quick question for clarification. There are now some docstrings using `Oscar.`. Are these docstrings for internal functions, that is, not exported functions?

HechtiDerLachs:
> Are these docstrings for internal functions, that is, not exported functions?

Yes, they should be. I have un-exported a lot of the types and their associated constructors. Are you aware of any instances where it looks like I have made a mistake? In the diff here I don't see any docstrings changed.

experimental/Schemes/ToricBlowups/attributes.jl
```diff
@@ -13,7 +13,7 @@ Normal toric variety
 julia> blow_down_morphism = blow_up(P3, [0, 1, 1])
 Toric blowdown morphism
 
-julia> underlying_morphism(blow_down_morphism)
```
thofma, line 16: @HechtiDerLachs I mean this here for example. I just wanted to double check that you intend to not export those functions.

HechtiDerLachs: Yes, `underlying_morphism` is internal. @HereAround and I already discussed that the user-facing counterpart should be `forget_toric_structure` in this particular case. `underlying_morphism`, on the other hand, is for forwarding various functionality through the type hiecharchy and applies to far more concrete types than just toric blow-down morphisms.

fingolfin (approved):
I saw no objections but some approvals, so on request of @HechtiDerLachs I am happy to merge this now.

PR #3177, opened 2024-01-11
Replace some instances of 'typeof(a)<:b' by 'a isa b'

fingolfin:
There are more in the F-Theory code but I'll leave it to @HereAround to address them so I don't cause conflicts


lgoettgens (approved):

HereAround (approved):
Looks good to me. Thank you @fingolfin .

PR #3179, opened 2024-01-12
Delete the old plane curves and projective space

simonbrandhorst:
We still have the old plane curve module in experimental.
If there is no one maintaining it, it should be removed.

Thus if someone volunteers to maintain it. They can now speak up.


fingolfin (approved):
Fine by me, but of course algebraic geometers should decide.

simonbrandhorst:
Following our zoom meeting we decided to remove them.

PR #3183, opened 2024-01-12
Catch empty polynomial ring in matroid code

joschmitt:
The tests for the Chow ring of a matroid expect that the polynomial ring in zero variables cannot be constructed.
This is going to change with https://github.com/Nemocas/Nemo.jl/pull/1619 .
I added an explicit check now.

Just to say: with the new code in Nemo, `chow_ring(uniform_matroid(1, 1))` would return a polynomial ring in zero variables modulo the zero ideal, so I'm not even sure we have to throw an error here.


thofma:
This is not changing any functionality, so I will merge this. If the error message is phrased wrongly, this can be adjusted later.

PR #3189, opened 2024-01-15
Degree for finite affine schemes

ooinaruhugh:
I noticed that projective schemes do have a `degree` function, but finite affine schemes don't work with those, because the dimension can't be picked up by the type system. 

I added an overload for affine schemes that's parallel to the degree function for projective schemes.
I did not add a docstring as the version of `degree` for projective schemes didn't have one either (yet).

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -481,8 +481,8 @@ julia> codim(Y)
   return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))])) - dim(X)
 end
 
-@attr Int function degree(X::Spec{BRT, RT}) where {BRT<:Field, RT}
-  @req dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
+@attr Int function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
+  @check dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
   return vector_space_dimension(OO(X))
 end
 
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -482,7 +482,7 @@ julia> codim(Y)
 end
 
 function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
-  @check dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
+  @check dim(X) == 0 "the affine scheme X needs to be zero-dimensional"
   get_attribute!(X, :degree) do
     return vector_space_dimension(OO(X))
   end
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -481,6 +481,10 @@ julia> codim(Y)
   return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))])) - dim(X)
 end
 
+@attr Int function degree(X::Spec{BRT, RT}) where {BRT<:Field, RT}
+  @req dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
+  return vector_space_dimension(OO(X))
+end
 
 @doc raw"""
     name(X::AbsSpec)
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
@@ -481,9 +481,11 @@ julia> codim(Y)
   return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))])) - dim(X)
 end
 
-@attr Int function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
+function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
   @check dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
-  return vector_space_dimension(OO(X))
+  get_attribute!(X, :degree) do
+    return vector_space_dimension(OO(X))
+  end
 end
 
 @doc raw"""
--- a/test/AlgebraicGeometry/Schemes/AffineSchemes.jl
+++ b/test/AlgebraicGeometry/Schemes/AffineSchemes.jl
@@ -358,3 +358,19 @@ end
   @test compose(p1, there) == compose(p2, there)
 end
 
+@testset "degree of zero-dimensional affine schemes" begin
+  R, (x,) = polynomial_ring(QQ, [:x])
+  I = ideal(R, x^2 - 1)
+  X = Spec(R)
+  Q, _ = quo(R, I)
+  Y = Spec(Q)
+  @test dim(Y) == 0
+  @test degree(Y) == 2
+
+  R, (x,y) = QQ[:x,:y]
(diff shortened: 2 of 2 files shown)


HechtiDerLachs:
Thanks for your contribution! 

A few comments:
 1. This is exclusively for zero dimensional schemes over a field. I was wondering whether someone might want to make sense of `degree` also in other cases. @afkafkafk13 ? @simonbrandhorst ? 
 2. I would change the `@req` into an `@check` so that the dimension need not be computed each time; I'll leave a sample in the code. 
 3. Does your code already run for schemes of localizations? Maybe some methods of `vector_space_dimension` are still missing.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -480,6 +480,10 @@ julia> codim(Y)
   return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))])) - dim(X)
 end
 
+@attr Int function degree(X::Spec{BRT, RT}) where {BRT<:Field, RT}
+  @req dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
```
HechtiDerLachs:
```suggestion
@attr Int function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
  @check dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
```

simonbrandhorst:
In general it will be the degree of the projectivisation. 
quoting Wikipedia:
"In [mathematics](https://en.wikipedia.org/wiki/Mathematics), the degree of an [affine](https://en.wikipedia.org/wiki/Affine_variety) or [projective variety](https://en.wikipedia.org/wiki/Projective_variety) of [dimension](https://en.wikipedia.org/wiki/Dimension) n is the number of intersection points of the variety with n [hyperplanes](https://en.wikipedia.org/wiki/Hyperplane) in [general position](https://en.wikipedia.org/wiki/General_position)."

ooinaruhugh:
> Thanks for your contribution!
> 
> A few comments:
> 
> 1. This is exclusively for zero dimensional schemes over a field. I was wondering whether someone might want to make sense of `degree` also in other cases. @afkafkafk13 ? @simonbrandhorst ?
> 2. I would change the `@req` into an `@check` so that the dimension need not be computed each time; I'll leave a sample in the code.
> 3. Does your code already run for schemes of localizations? Maybe some methods of `vector_space_dimension` are still missing.

Regarding 1, degree also makes sense for schemes over ZZ. Spec(ZZ/(p^e)) would be an affine finite-dimensional scheme and one can define the degree as e here. (That's described in Section II.4.5 in "The Geometry of Schemes" by Eisenbud and Harris)
For n general it would be the sum of the exponents in the prime factorization by using the CRT on ZZ/(n). 

Regarding what @simonbrandhorst said, one could also do a case for hypersurfaces taking the total degree of the defining polynomial. Otherwise, would the projective closure of the original scheme work, so if available, take the vanishing ideal, 
homogenize that, pass the resulting projective scheme to the existing degree function?

Regarding 3, did not think of them yet, will check that.

HechtiDerLachs:
I think even for afffine schemes of dimension > 0 one can go via some sort of Hilbert function for a degree ordering. Probably @afkafkafk13 knows more. For the localizations at powers of elements I would then go via the `saturated_ideal` and for localizations at rational points via the `shifted_ideal` and a local ordering. For localizations at prime ideals we have to consult the literature again or simply throw a "not implemented". So no need to go through a projectivization, I think. 

Edit: @ooinaruhugh Do I understand correctly that you will take on this task? If this seems to be too far off what you initially intended to do, feel free to take some shortcut. Then we should probably just throw "not implemented" for all cases to come for the moment.

ooinaruhugh:
@HechtiDerLachs Yeah, I would like to do that definitely. But also, for now, as a working base, I'd keep the code for finite affine schemes (so we'd need to check for dim == 0 and only allow it for fields now) and afterwards, I'd add the other cases.

Also, for now, I cannot use `@check` alongside with `@attr`, or at least I cannot keep that keyword argument for check. That's why the tests fail now. I guess having `@attr` would be nice, but for now I'd need check, or I check the dimension without `@check`, but then one cannot turn that off.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -480,6 +480,10 @@ julia> codim(Y)
   return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))])) - dim(X)
 end
 
+@attr Int function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
+  @check dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
+  return vector_space_dimension(OO(X))
+end
```
ooinaruhugh: So for now, I would keep the check but remove `@attr` even if the output of `degree` is not cached then.
```suggestion
function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
  @check dim(X) == 0 "Degree is not implemented for affine schemes of dimension > 0"
  return vector_space_dimension(OO(X))
end
```

afkafkafk13:
For affine schemes (with fixed embedding into affine space), the degree is defined as the degree of its projectivization as @simonbrandhorst already said. 
The degree of a projective variety is in general computed from the leading coefficient of the Hilbert-polynomial, which counts monomials of a given degree d for sufficiently large d. As the Hilbert-polynomial of the projectivization is precisely the affine Hilbert polynomial of the affine variety, we can also directly look at the leading coefficient of the affine Hilbert-polynomial (which counts monomials of degree up to d for sufficiently large d).

From my point of view, the way to go would be:
- have affine Hilbert polynomial (obtaining it from the already implemented projective one)
- obtain degree from affine Hilbert polynomial 
- handle principal open subsets (localizations at MPolyPowersOfElement) by saturation
- for localizations at maximal ideals a similar concept exists, but should not be supported under the name of degree (for consistency with the mathematical theory)

In the case of localizations at a maximal ideal, the leading coefficient of the Hilbert-Samuel polynomial provides the **multiplicity** in the same way as the leading coefficient of the Hilbert-polynomial provides the degree in the global case. We should hence support it via mulitplicity and throw appropriate error messages pointing to the right command, if degree is called in the local setting or multiplicity in the global one.

afkafkafk13:
And I forgot to say: @ooinaruhugh welcome aboard and thanks for the contribution!

HechtiDerLachs:
> That's why the tests fail now.

There's a workaround for this not using the `@attr` macro. See e.g. `experimental/Schemes/FunctionFields.jl`, it uses some `get_attribute!(...) do... end` structure. 

Edit: Sorry, I just had a look and it seems that the caching of `function_field` in attributes has been lost. Either way: If one manually stores and maintains the attributes, one can use keyword arguments.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
```diff
@@ -480,6 +480,12 @@ julia> codim(Y)
   return dim(ideal(ambient_coordinate_ring(X), [zero(ambient_coordinate_ring(X))])) - dim(X)
 end
 
+function degree(X::Spec{BRT, RT}; check::Bool=true) where {BRT<:Field, RT}
+  @check dim(X) == 0 "The affine scheme X needs to be zero-dimensional"
```
HechtiDerLachs:
```suggestion
  @check dim(X) == 0 "the affine scheme X needs to be zero-dimensional"
```
Really minor thing, but: It's Oscar convention to not use capital letters in the beginning of error messages.

HechtiDerLachs (approved):

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Attributes.jl
joschmitt: (It is?! I didn't know that.)

HechtiDerLachs: I think, I once learned this from @thofma . But things might have changed in the meantime. Either way, I am reluctant to have the tests run once more only for this. But it's something to keep in mind.

Issue #3171, opened 2024-01-10
Automorphisms of groups issue


thofma:
I have been using the automorphism groups of groups more extensively lately and noticed a few things. I will use the following as an example:
```julia
julia> F = free_group(["x", "y"]);

julia> x, y = gens(F);

julia> Q, = quo(F, [x^-3*y^-3*x^-1*y*x^-2,  x*y^-1*x^-1*y*x^-1*y^-1*x^3*y^-1, x*y^-1*x^-1*y^2*x^-2*y^-1*x^2]);

julia> A = automorphism_group(Q);

julia> auts = collect(A)
16-element Vector{AutomorphismGroupElem{FPGroup}}:
 IdentityMapping( <fp group of size 16 on the generators [ x, y ]> )
 [ x*y^-18, y^7 ] -> [ x, y ]
 [ x*y^-12, y^5 ] -> [ x, y ]
 [ x*y^-102, y^35 ] -> [ x, y ]
 [ (y^-4*x^-1)^2*y^-1, y*x*y^3 ] -> [ x, y ]
 [ (y^-10*x^-1)^2*y^-7, y^7*x*y^3 ] -> [ x, y ]
 [ (y^-8*x^-1)^2*y^-5, y^5*x*y^3 ] -> [ x, y ]
 [ (y^-38*x^-1)^2*y^-35, y^35*x*y^3 ] -> [ x, y ]
 [ x*y^8, y^5 ] -> [ x, y ]
 [ x*y^-82, y^35 ] -> [ x, y ]
 [ x*y^-52, y^25 ] -> [ x, y ]
 [ x*y^-502, y^175 ] -> [ x, y ]
 [ (y^-28*x^-1)^2*y^-5, y^5*x*y^23 ] -> [ x, y ]
 [ (y^-58*x^-1)^2*y^-35, y^35*x*y^23 ] -> [ x, y ]
 [ (y^-48*x^-1)^2*y^-25, y^25*x*y^23 ] -> [ x, y ]
 [ (y^-198*x^-1)^2*y^-175, y^175*x*y^23 ] -> [ x, y ]
```

1. Why are the automorphisms "the other way around" aka why are they not `[x, y] -> ...`? At the moment, I can't even evaluate those automorphisms. It will hang for some of them:
  ```julia
julia> for i in 1:length(auts)
         @info i, auts[i](gen(Q, 1))
       end
[ Info: (1, x)
[...]
[...]

Issue #3167, opened 2024-01-10
Something with automorphism groups hanging


thofma:
The following hangs for me:
```julia
julia> F = free_group(["x", "y"])
Free group of rank 2

julia> (x, y) = gens(F);

julia> Q, = quo(F, [x^-3*y^-3*x^-1*y*x^-2,  x*y^-1*x^-1*y*x^-1*y^-1*x^3*y^-1, x*y^-1*x^-1*y^2*x^-2*y^-1*x^2]);

julia> A = automorphism_group(Q);

julia> H, HtoA = inner_automorphism_group(A);

julia> Q, AtoQ = quo(A, H);

julia> [preimage(AtoQ, q) for q in Q] #this will hang forver
```
Any idea @ThomasBreuer? I tried adding `check = false`, but it does not exist for the argument types.

ThomasBreuer:
`Q` has order 16.
I can reproduce that the `primage` computation for some of the elements takes very long.
This happens inside the GAP function `PreImagesRepresentative`. I will try to find a fix for that.

ThomasBreuer:
@thofma Concerning `check = false`, one can enter `[haspreimage(AtoQ, q, check = false) for q in Q]`.
Currently the documentation for `preimage` suggests this if one wants to omit the membership tests.
If you thinkk that it is better to support `check = false` also for `preimage` then we can easily change this.

#3190 yields already some improvement, but there is still the problem from gap-system/gap/issues/5590.

thofma:
Thanks for the explanation. I don't mind too much about the `check = false`. I was just wondering about the inconsistency in the keyword arguments of `preimage/haspreimage`.

thofma:
Should we close this and keep the discussion/progress at #3171?

ThomasBreuer:
@thofma 
> Should we close this and keep the discussion/progress at #3171?

Yes, #3171 is more general.

PR #3190, opened 2024-01-15
fix some problems with group automorphisms

ThomasBreuer:
- provide `preimage` for `AutomorphismGroupElem`, analogous to `preimage` for `GAPGroupHomomorphism` (addresses #3171)
- change the construction of the underlying GAP map of group automorphisms (addresses #3167)


thofma:
I am not too familiar with this part of the code base, but if it works, it works. Thanks!

thofma (approved):

PR #3191, opened 2024-01-16
Add some more functionality, docs and tests for root systems

lgoettgens:


felix-roehrich:
Please add a property test to make sure that the roots of the dual root system correspond to the coroots of the base root system.

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -81,11 +89,41 @@ function cartan_matrix(R::RootSystem)
...
+  return D
+end
+
+@doc raw"""
+    coroot(R::RootSystem, i::Int) -> RootSpaceElem
+
+Returns the `i`-th coroot of `R`, i.e. the `i`-th root of the dual root system of `R`.
+This is a more efficient version for `coroots(R)[i]`.
+
+Also see: `coroots`.
```
fingolfin, line 100: How about using an actual ref?
```suggestion
Also see: [`coroots`](@ref).
```

lgoettgens: This does only work once we have a markdown page collecting the docstrings. That's something @felix-roehrich wants to approach in the near future.

lgoettgens:
> Please add a property test to make sure that the roots of the dual root system correspond to the coroots of the base root system.

done

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -72,6 +78,10 @@ function Base.show(io::IO, R::RootSystem)
   end
 end
 
+@attr ZZMatrix function cartan_bilinear_form(R::RootSystem)
```
felix-roehrich: I think we agreed that here `cartan` is not necessary.

```suggestion
@attr ZZMatrix function bilinear_form(R::RootSystem)
```

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -81,11 +91,34 @@ function cartan_matrix(R::RootSystem)
...
+    coroots(R::RootSystem) -> Vector{RootSpaceElem}
+
+Returns the coroots of `R`, starting with the coroots of positive roots and then the negative roots,
+in the order of `positive_coroots` and `negative_coroots`.
+
+Also see: `coroot`.
+"""
+function coroots(R::RootSystem)
+  return [[r for r in positive_coroots(R)]; [-r for r in positive_coroots(R)]]
+  (dual(R))
```
felix-roehrich:
```suggestion
```

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -313,6 +514,10 @@ function Base.iszero(r::RootSpaceElem)
   return iszero(r.vec)
 end
 
+function Base.length(r::RootSpaceElem)
+  return dot(r, r)
```
felix-roehrich: We need `sqrt` here, but does this work with current types ?
```suggestion
  return sqrt(dot(r, r))
```

lgoettgens: No, that doesn't work

lgoettgens: We could instead name the function something like `length_squared` instead

felix-roehrich: That is neither intuitive nor shorter than `dot(r, r)`, so I don't feel great about this.

lgoettgens: The let's just remove `length`

experimental/LieAlgebras/src/LieAlgebras.jl
```diff
@@ -73,9 +75,10 @@ import ..Oscar:
   ⊕,
   ⊗
 
-import Base: getindex, deepcopy_internal, hash, issubset, iszero, parent, zero
+import Base: getindex, deepcopy_internal, hash, issubset, iszero, length, parent, zero
```
lgoettgens:
```suggestion
import Base: getindex, deepcopy_internal, hash, issubset, iszero, parent, zero
```

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -313,6 +513,10 @@ function Base.iszero(r::RootSpaceElem)
...
+end
+
```
lgoettgens:
```suggestion
```

felix-roehrich (approved):

lgoettgens:
can we get a merge of this? maybe @thofma ?

thofma (approved):

PR #3192, opened 2024-01-16
Project to incorporate Frank Lutz's data

jspreer:
@antonydellavecchia and @micjoswig 

# Idea 

Add functionality to SimplicialComplex so we can avoid creating an extra type.

We need: 
- way to compute automorphism group of a simplicial complex (this should work with minimal effort by computing vertex facet incidence matrix and using function "automorphism_group")
- function to compute orbit representatives from automorphism group, facet list pair
- meta data for SimplicialComplex
- Field for topological type (a string, optional, could be combined with existing functionality)

How would we save data (example):

## From manifold page:

facets=[[1,2,4],[1,2,6],[1,3,4],[1,3,7],
  [1,5,6],[1,5,7],[2,3,5],[2,3,7],[2,4,5],
  [2,6,7],[3,4,6],[3,5,6],[4,5,7],[4,6,7]]
tt = "torus"
  
c = SimplicialComplex(facets)
f_vector(c)
hom = []
for i in 1:dim(c)+1
  push!(hom,homology(c,i-1))
end
automorphism_group(c)
top_type(c) = tt

save(c)


antonydellavecchia:
We should make smaller issues from this one and make a clearer roadmap, but after the release of Oscar 1.0

simpcomp-team:
Summary of our meeting from July 16, 2024:
- We will start with creating .json files for surface triangulations from https://www3.math.tu-berlin.de/IfM/Nachrufe/Frank_Lutz/stellar/surfaces.html
- Next step is to write OSCAR files to access this data
- Roadmap: turn this into a data base, decide where to store this data, larger collections of data should probably directly be stored in a database without a separate file for each simplicial complex.

fingolfin:
How does this relate to the GAP package `simpcomp`, which we already ship with OSCAR -- does it overlap with this in any way?

micjoswig:
@jspreer is behind simpcomp, which is mostly code and little data.

Those data, currently existing on a web page at TU Berlin, are mere data.  The goal is to make that OSCAR accessible.

So; no overlap.

fingolfin:
What is the status of this PR? There has been no activity since last July.

antonydellavecchia:
see #5137

PR #3194, opened 2024-01-17
Fixes length for free resolutions

ederc:
Fixes the length of free resolutions computed via `Singular`.

### DIFF
--- a/src/Modules/UngradedModules/FreeResolutions.jl
+++ b/src/Modules/UngradedModules/FreeResolutions.jl
@@ -356,8 +356,13 @@ function free_resolution(M::SubquoModule{<:MPolyRingElem};
     error("Unsupported algorithm $algorithm")
   end
 
-  if length == 0 || Singular.length(res) < length
+  slen = iszero(res[Singular.length(res)+1]) ? Singular.length(res) : Singular.length(res)+1
+  if length == 0 || slen < length
     cc_complete = true
+  end
+
+  if length != 0
+    slen =  slen > length ? length : slen
   end
 
   br_name = AbstractAlgebra.find_name(base_ring(M))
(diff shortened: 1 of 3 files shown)


thofma:
Do we have an example which is fixed by this? If so, we could add it as a test.

wdecker:
The example we discussed earlier would do

S, (x0, x1, x2, x3, x4) = graded_polynomial_ring(GF(3), ["x0", "x1", "x2", "x3", "x4"]);
m = ideal(S, [x1^2+(-x1+x2+x3-x4)*x0, x1*x2+(x1-x3+x4)*x0,
          x1*x3+(-x1+x4+x0)*x0,
          x1*x4+(-x1+x3+x4-x0)*x0, x2^2+(x1-x2-x4-x0)*x0,
          x2*x3+(x1-x2+x3+x4-x0)*x0, x2*x4+(x1+x2-x3-x4-x0)*x0,
          x3^2+(x3+x4-x0)*x0,x3*x4+(-x3-x4+x0)*x0,
          x4^2+(x1+x3-x4-x0)*x0]);
A, _ = quo(S, m);
FA = free_resolution(A)

ederc:
Working on this, give me a moment.

src/Modules/UngradedModules/FreeResolutions.jl
```diff
@@ -204,28 +204,39 @@ function _extend_free_resolution(cc::Hecke.ComplexOfMorphisms, idx::Int)
   dom = domain(cc.maps[1])
   j   = 2
 
-  while j <= Singular.length(res)
+  #= get correct length of Singular sresolution =#
+  slen = iszero(res[Singular.length(res)+1]) ? Singular.length(res) : Singular.length(res)+1
```
HechtiDerLachs, line 208: Unfortunately, this has not worked for me some days ago. I tried different examples and a fixed offset did not work out. Are you sure this does the job beyond the example provided by @wdecker ?

ederc: According to @hannes14 this should be the way to go, I tried several examples and it worked. If you have some counterexample just let me know.

hannes14 (approved):

src/Modules/UngradedModules/FreeResolutions.jl
HechtiDerLachs: Sorry, I didn't read it correctly. Probably this works.

wdecker (approved):

Issue #2891, opened 2023-10-05
Adjust printing of modules and related stuff to follow OSCAR printing guidelines


fingolfin:
... as described in https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/ 

This includes updating `show` methods in `src/Modules/UngradedModules.jl` for
- [ ] `FreeMod`
- [ ] `FreeModuleHom`
- [ ] `ModuleGens`
- [ ] `SubModuleOfFreeModule`
- [ ] `SubQuoHom`
- [ ] `SubquoModule`

and probably more elsewhere.

fingolfin:
I'd say this is "good enough". Perhaps things can be improved but then dedicated issues should be opened pointing out specific flaws with the current state of things.

PR #3199, opened 2024-01-19
Start revising printing of modules

fingolfin:
First steps towards resolving #2891 which is important for the book.
The code changes are all in the first commit, the second commit just contains the changed produced by:
```
julia> using Revise, Oscar, Documenter ; Oscar.build_doc(; doctest=:fix, open_browser=false)
```

I started this back in October, and just rebased it (as promised on Wednesday, sorry for the delay), and regenerated the docs.

This is definitely not the final version, it has clear issues, but at least it is start. So, what to do with this, some options:
1. discard it and you do your own (but *soon*)
2. someone adopts it and pushes changes to this PR
3. someone adopts it and opens their own PR (we close this)
4. just merge this (perhaps with some tweaks) and build upon it in a new PR

Any is fine by me, it's not as if there is much in it.


src/Modules/ModulesGraded.jl
```diff
@@ -27,13 +27,13 @@ julia> R, (x,y) = graded_polynomial_ring(QQ, ["x", "y"])
 (Graded multivariate polynomial ring in 2 variables over QQ, MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}[x, y])
 
 julia> graded_free_module(R,3)
-Graded free module R^3([0]) of rank 3 over R
+Graded free module Graded multivariate polynomial ring^3([0]) of rank 3 over graded multivariate polynomial ring
```
HechtiDerLachs: I think this will be objected to by @jankoboehm and @wdecker and should be fixed. Probably it has to do with switches from `compact` to `supercompact` somewhere? Probably only the old method's signature (of the one which was looking up the rings to fish out their names) has to be adjusted.

docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/complexes.md
```diff
@@ -58,19 +58,15 @@ by Submodule with 1 generator
...
-by Submodule with 1 generator
-1 -> x^4*e[1]
-Codomain:
-=========
-Subquotient of Submodule with 1 generator
-1 -> e[1]
-by Submodule with 1 generator
-1 -> x^3*e[1]
+Module homomorphism
+  from subquotient of Submodule with 1 generator
```
fingolfin: `Lowercase()` is missing somewhere:
```suggestion
  from subquotient of submodule with 1 generator
```

docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/complexes.md
```diff
@@ -58,19 +58,15 @@ by Submodule with 1 generator
...
+  1 -> e[1]
+  by Submodule with 1 generator
```
fingolfin: `Lowercase()` is missing somewhere:
```suggestion
  by Submodule with 1 generator
```

docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/complexes.md
```diff
@@ -58,19 +58,15 @@ by Submodule with 1 generator
...
+  1 -> x^4*e[1]
+  to subquotient of Submodule with 1 generator
+  1 -> e[1]
```
fingolfin, line 67: Maybe this should use `Indent()` ?
```suggestion
    1 -> e[1]
```

src/Modules/ModulesGraded.jl
fingolfin: This change is bad. I guess the problem is that I changed `compact` to `supercompact` in some places, and `supercompact` does not honor the "names" assigned to things... Perhaps change back to `:compact` in some places; or add/modify some supercompact printing code for these rings to honor the names again?

src/Modules/ModulesGraded.jl
```diff
@@ -1550,7 +1614,14 @@ julia> I = ideal(S, gens(S))
 ideal(x[1], x[2], x[3], x[4])
 
 julia> FI = free_resolution(I)
-Free resolution of I
+Free resolution of Graded submodule of Graded multivariate polynomial ring^1(Element of
+GrpAb: Z
+with components [0])
+1 -> x[1]*e[1]
```
HechtiDerLachs: This is another instance where printing should be repaired, probably similar to the above.

src/Modules/UngradedModules/FreeMod.jl
```diff
@@ -49,23 +49,23 @@ julia> U = complement_of_prime_ideal(P);
 julia> RL, _ = localization(R, U);
 
 julia> FRL = free_module(RL, 2, "f")
-Free module of rank 2 over Localization of multivariate polynomial ring in 3 variables over QQ at complement of prime ideal(x, y, z)
+Free module of rank 2 over localized ring
```
HechtiDerLachs, line 52: maybe `localized polynomial ring` would be better, given that we also spell out `multivariate polynomial ring`? But I have no strong opinion on that .

src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -113,13 +113,9 @@ julia> V = [y*G[1], x*G[1]+y*G[2], z*G[2]]
...
-Map with following data
-Domain:
-=======
-Free module of rank 3 over Multivariate polynomial ring in 3 variables over QQ
-Codomain:
-=========
-Free module of rank 2 over Multivariate polynomial ring in 3 variables over QQ
+Module homomorphism
+  from free module of rank 3 over multivariate polynomial ring
+  to free module of rank 2 over multivariate polynomial ring
```
HechtiDerLachs, line 118: This looks much better now in my opinion.

src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -275,41 +273,37 @@ function morphism_type(
   return FreeModuleHom{DomainType, CodomainType, RingMapType}
 end
 
-function Base.show(io::IO, ::MIME"text/plain", fmh::FreeModuleHom{T1, T2, RingMapType}) where {T1 <: AbstractFreeMod, T2 <: ModuleFP, RingMapType}
-  # HACK
-  show(io, fmh)
+function AbstractAlgebra.show_map_data(io::IO, fmh::FreeModuleHom{T1, T2, RingMapType}) where {T1 <: AbstractFreeMod, T2 <: ModuleFP, RingMapType}
+  if is_graded(fmh)
```
HechtiDerLachs, line 277: I'm not really an insider about this part of the code, but it looks fishy to me that there is no `else` for the ungraded case. Is this correct here?

src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -517,17 +501,11 @@ julia> image(a)
 1 -> y*e[1]
 2 -> x*e[1] + y*e[2]
 3 -> z*e[2]
-represented as subquotient with no relations., Map with following data
-Domain:
-=======
-Submodule with 3 generators
+represented as subquotient with no relations., Hom:
```
HechtiDerLachs:
```suggestion
represented as subquotient with no relations, Hom:
```
We must kill off the full stop somewhere.

jankoboehm:
Thanks a lot for making it available, will make a suggestion based on it, and then we can discuss whether it does the job.

src/Modules/UngradedModules/FreeModuleHom.jl
fingolfin: done

src/Modules/ModulesGraded.jl
fingolfin: Yes I explained this elsewhere. The origin of the issues is that `@show_name` macro from AA only checks for `compact`; this is its code:
```
macro show_name(io, O)
  return :( begin
    local i = $(esc(io))
    local o = $(esc(O))
    s = get_attribute(o, :name)
    if s === nothing
      sy = find_name(o)
      if sy === nothing
        sy = extra_name(o)
      end
      if sy !== nothing
        s = string(sy)
        set_name!(o, s)
      end
    end
    if get(i, :compact, false) && s !== nothing
      if AbstractAlgebra.PrettyPrinting._supports_io_custom(i)
        print(i, LowercaseOff())
      end
      print(i, s)
      return
    end
  end )
end
```

One option would be to set both `compact` and `supercompact`.

@thofma @fieker should `@show_name` be changed to also check `supercompact`? (or perhaps even instead of `compact`?)

fingolfin: For now I made it set both `compact` and `supercompact`, but this will be the last change I'll make to this PR, from here on the algebraic geometry group should take over, as outlined in the description of this PR.

src/Modules/UngradedModules/FreeMod.jl
fingolfin: Maybe, but that's beyond the scope of this PR, the `show` method for `AbsLocalizedRing` would have to be adjusted.

src/Modules/UngradedModules/FreeModuleHom.jl
fingolfin: This matches the old code. `show_map_data` is called as part of a general `show` method in AbstractAlgebra, which delegates to it plus `supercompact` (IIRC) printing -- the "else" part in the old code basically did `print(io, "Module homomorphism")` and this is now handled by invoking the `supercompact` code.

docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/complexes.md
fingolfin: Can't do that right now because it runs into a bug in pretty printing, see https://github.com/Nemocas/AbstractAlgebra.jl/issues/1561

fingolfin:
This has a ton of conflicts now, and has been basically ignored for months, so it seems it is best if someone else starts from scratch with this.

Issue #3184, opened 2024-01-12
Error on 1.10 ubuntu long


thofma:
If one looks at https://github.com/oscar-system/Oscar.jl/commits/master/, one sees that often "Run tests / test (~1.10.0-0, long, ubuntu-latest) (push)" fails. The error looks scary, e.g. in https://github.com/oscar-system/Oscar.jl/actions/runs/7364038493/job/20044077891#step:7:4952 and  https://github.com/oscar-system/Oscar.jl/actions/runs/7364038493/job/20044077891#step:7:26094: 
```
!!! ERROR in jl_ -- ABORTING !!!
```
Does anyone have an idea where that might be coming from? I have not tried to reproduce it locally. It does not look like https://github.com/oscar-system/Oscar.jl/issues/2441.

CC: @lgoettgens @benlorenz

lgoettgens:
No Idea

benlorenz:
Some weird GC corruption that seems to happen when the Serialization/IPC tests happen, it seems related to julia tasks but I haven't been able to reproduce this locally. I have the long testset running in a loop with `rr` to trigger and capture this (currently at about 100 iterations).

So far I got only one other crash but in the test group `elliptic_surfaces.jl` that runs before the IPC stuff:
```
[4832] signal (11.1): Segmentation fault
in expression starting at /home/datastore/lorenz/software/julia/Oscar.jl/test/AlgebraicGeometry/Schemes/elliptic_surface.jl:1
jl_object_id__cold at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/builtins.c:455
type_hash at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/jltypes.c:1575
typekey_hash at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/jltypes.c:1605
jl_precompute_memoized_dt at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/jltypes.c:1685
inst_datatype_inner at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/jltypes.c:2081
jl_inst_arg_tuple_type at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/jltypes.c:2176
arg_type_tuple at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:2232 [inlined]
[...]

PR #3200, opened 2024-01-19
Some more changes for the book

HechtiDerLachs:
This is for cleaning up our joint chapter with @simonbrandhorst .

### DIFF
--- a/experimental/Schemes/MorphismFromRationalFunctions.jl
+++ b/experimental/Schemes/MorphismFromRationalFunctions.jl
@@ -1,3 +1,5 @@
+export morphism_from_rational_functions
+
 @doc raw"""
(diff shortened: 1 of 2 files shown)


simonbrandhorst (approved):

HechtiDerLachs:
Tests fail and look pretty bad. Does anyone know, what's going on? 

@benlorenz

benlorenz:
So far only ubuntu+1.10+long failed and this is the same as reported in #3184.

HechtiDerLachs:
> So far only ubuntu+1.10+long failed and this is the same as reported in #3184.

Alright! Thanks for having a look!

benlorenz:
Regarding the macos failures: we added a workaround in #3203, but this is not yet included in the merge commit for this PR, so we can ignore these for now.

PR #3201, opened 2024-01-19
Make inj/proj of products groups consistent with other products objects

lgoettgens:
Continuation of https://github.com/oscar-system/Oscar.jl/pull/2747 as proposed in https://github.com/oscar-system/Oscar.jl/pull/2747#issuecomment-1700714017.

@fingolfin already approved this in https://github.com/oscar-system/Oscar.jl/pull/2747#issuecomment-1724565377.


fingolfin (approved):
Fine by me, though perhaps we should wait for the actual resolution of the great `direct_product` debate of 2024 ?

fingolfin (approved):
Let's merge it for now so that everything (? at least "more things") use `canonical_injection` and then if need we can change all of those (or not, or whatever is decided)

PR #3203, opened 2024-01-19
fix: workaround for bug in new flint binaries

thofma:


benlorenz (approved):

benlorenz:
I will merge this in a few minutes once I checked that it does install the correct `FLINT_jll` version without waiting for all the tests.
The reason is that for some unclear reasons the new FLINT build 200.900.8 breaks the GModule tests+doctests on macos (maybe only on x86_64?).

PR #3209, opened 2024-01-19
Remove `sylow_subgroup(::GrpAbFinGen)`

StevellM:
After renaming the similar function on Hecke (https://github.com/thofma/Hecke.jl/pull/1355), this won't be necessary anymore. Though one needs to wait for the other PR to be merged and a new version of Hecke.


lgoettgens (approved):
looks good, but needs to wait for a bit

lgoettgens:
Already included in #3231

Issue #2999, opened 2023-11-07
Building morphisms of graded modules takes unexpectedly long


HechtiDerLachs:
Try the following code:
```julia
S, (x, y, z) = graded_polynomial_ring(QQ, [:x, :y, :z])
m = ideal(S, gens(S))
S1 = graded_free_module(S, [0])
g = gens(m^29)
f = [f*S1[1] for f in g]

@time I, inc = sub(S1, f);

@time Oscar.SubModuleOfFreeModule(S1, f);
```
On my machine I get
```
julia> @time I, inc = sub(S1, f);
  0.160066 seconds (2.22 M allocations: 98.650 MiB, 52.27% gc time)

julia> @time Oscar.SubModuleOfFreeModule(S1, f);
  0.000013 seconds (10 allocations: 1.141 KiB)
```
which is a significant differences. In fact, using `sub` in bigger examples killed me. 

Using `ProfileView` revealed that most of the time is spent in inferring the degree of the generators for the domain of the inclusion map `inc`. This looks to me as if there's room for improvement.

@jankoboehm ?

PR #3211, opened 2024-01-19
Fix #3107

### LINKED ISSUE #3107: dualizing does not work for free resolutions
Example due to @wdecker :
```julia
X = veronese();
I = defining_ideal(X);
Pn = base_ring(I)
FI = free_resolution(I)
F = graded_free_module(Pn, 1)
dualFIC = hom(FI.C, F)
```

HechtiDerLachs:
Fix #3107 

CC: @wdecker

### DIFF
--- a/experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
+++ b/experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
@@ -587,3 +587,20 @@ function _has_index(a::SMat, i::Int, j::Int)
   return _has_index(a[i], j)
 end
 
+### Taylormade functionality for modules
+# Provided as a hotfix for issue #3108.
+function _alt_simplify(M::SubquoModule)
+  res, aug = free_resolution(SimpleFreeResolution, M)
+  simp = simplify(res)
+  simp_to_orig = map_to_original_complex(simp)
+  orig_to_simp = map_from_original_complex(simp)
+  result, Z0_to_result = homology(simp, 0)
+  Z0, inc_Z0 = kernel(simp, 0)
+
+  result_to_M = hom(result, M, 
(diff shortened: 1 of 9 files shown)


HechtiDerLachs:
I also added some code that can be used as a hotfix to #3108 . @wdecker : Let me know if you would like this to be made available via the `simplify` command.

HechtiDerLachs:
@jankoboehm : If you run `@profview Oscar._alt_simplify(Omega)` in the added test code, then you will see that about 90% of the computing time is spent on inferring the degrees of various generators. This looks suboptimal. So I think #2999 should be looked into at some point.

HechtiDerLachs:
I added methods to drop the grading of graded modules so that it becomes possible to compare timings between the existing implementation of `simplify` and `Oscar._alt_simplify`. Starting from the snippet in #3108 I get the following results:

 * `Oscar._alt_simplify` returns a module with **two** generators while `simplify` returns one with **32**. So my simplification seems to be much better, at least in this particular example. 
 * My code seems to also be better in terms of speed:
```julia
julia> M, iso = forget_grading(Omega);

julia> @time Oscar._alt_simplify(M);
  1.236710 seconds (4.73 M allocations: 218.460 MiB)

julia> @time simplify(M);
  7.122538 seconds (7.66 M allocations: 184.271 MiB)
```
To me these results suggest that we may just replace the current `simplify` function by what is now `Oscar._alt_simplify` in this branch.

HechtiDerLachs:
I was intrigued by #3239 to compare performance. It turned out that Singular was faster in the beginning.

With the latest commit we're both on the same speed and memory consumption:
```
julia> @time M1, iso1 = prune_with_map(M);
  0.578935 seconds (2.31 M allocations: 115.540 MiB)

julia> @time M2, iso2, iso2_inv = Oscar._alt_simplify(M);
  0.515923 seconds (2.12 M allocations: 113.318 MiB)
```
My offer stands to incorporate `_alt_simplify` as the default simplification routine in case of modules over non-polynomial rings.

jankoboehm (approved):

PR #3214, opened 2024-01-20
docs: fix reference to intersection -> intersect

benlorenz:
should fix failing docs deploy on master after latest AbstractAlgebra (due to https://github.com/Nemocas/AbstractAlgebra.jl/pull/1550)


thofma:
Hm, I think we had this discussion before. Should the downstream tests catch this?

thofma (approved):

benlorenz:
We are running doctests in the downstream tests (via `doctest(...)`), but this doesn't check references in the docstrings. I think this needs `makedocs(...)` which might need the docs-sources of the extra packages?

thofma:
Ah, this is something different. Hm, that sounds more complicated. Let's hope we slow down on breaking releases after 1.0.

PR #3215, opened 2024-01-21
Initial ideals and marked Groebner bases from `TropicalGeometry/groebner_fan.jl/groebner_fan` 

ooinaruhugh:
While working with `groebner_fan`, I found myself computing the corresponding initial ideals very often. That's why I thought it would be convenient to have a parameter `return_initial_ideals` resulting in another dictionary mapping interior points of cones to the initial ideals (like the other two optional parameters).

Also, I added an optional parameter `marked_grobener_bases` that modifies the dictionary of Groebner bases that is already returned to instead give pairs of generators and initial term. In the package `gfanInterface` for Macaulay2, this is called a marked polynomial list, hence the name for this parameter.

### DIFF
--- a/src/TropicalGeometry/groebner_fan.jl
+++ b/src/TropicalGeometry/groebner_fan.jl
@@ -397,7 +397,7 @@ end
 
 
 @doc raw"""
-    groebner_fan(I::MPolyIdeal; return_groebner_bases::Bool=false, return_orderings::Bool=false, verbose_level::Int=0)
+    groebner_fan(I::MPolyIdeal; return_groebner_bases::Bool=false, return_orderings::Bool=false, return_initial_ideals::Bool=false, marked_groebner_bases::Bool=false, verbose_level::Int=0)
 
 Return a `PolyhedralFan` representing the Groebner fan of `I`, where is a multivariate polynomial ideal over a field.
 
@@ -407,7 +407,11 @@ If `I` is not weighted homogeneous with respect to a positive weight vector, the
 
 If `return_groebner_bases==true`, also return a dictionary whose keys are interior points of the maximal cones and whose values are the Groebner bases for those cones.  Their union will be a universal Groebner basis.
 
+If additionally `marked_groebner_bases==true`, the values of above dictionary are the Groebner bases for those cones together with the leading term for each generator.  
+
 If `return_orderings==true`, also return a dictionary whose keys are interior points of the maximal Groebner cones and whose values are monomial orderings for those cones.  These orderings are suboptimal and hence it is generally recommended to create new orderings with the interior points.  However they do contain information on how the fan was traversed.
+
+If `return_initial_ideals==true`, also return a dictionary whose keys are interior points of the maximal cones and whose values are the initial ideals for those cones.
 
 # Examples
 ```jldoctest
--- a/src/TropicalGeometry/groebner_fan.jl
+++ b/src/TropicalGeometry/groebner_fan.jl
@@ -429,7 +429,7 @@ julia> SigmaI,gbs,ords = groebner_fan(I,return_groebner_bases=true,return_orderi
 ```
 """
 
-function groebner_fan(I::MPolyIdeal; return_groebner_bases::Bool=false, return_orderings::Bool=false, verbose_level::Int=0)
+function groebner_fan(I::MPolyIdeal; return_groebner_bases::Bool=false, return_orderings::Bool=false, return_initial_ideals::Bool=false, verbose_level::Int=0)
     ###
     # Preparation:
     #   Test whether the ideal is weighted homogeneous with respect to a positive weight vector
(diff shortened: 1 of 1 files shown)


thofma:
BTW, @YueRen is this a user-facing function?

YueRen:
@thofma Yes, it is a user-facing function.

@ooinaruhugh Thanks for the changes!  One suggestion:  Instead of returning multiple dictionaries (depending on the combination of flags), how about we just return a single list of vectors - one vector per maximal cone - and the entries of that vector are governed by what flags are invoked?

Unless there is a good reason to use dictionaries, I'd rather stick to what gfan and polymake are doing, which is returning simple vectors.  (And yes, I was the person who wrote them to return dictionaries in the first place)

ooinaruhugh:
@YueRen I suppose things are returned as dictionary so you can pair the output with an interior point of each cone. I find that quite useful, so we could also return one big dictionary with with keys the interior points and values the vectors as you said.

YueRen:
@ooinaruhugh I was more thinking of having flags `return_interior_points`, `return_orderings`, `return_groebner_bases`, and `return_initial_ideals`, and - if one of the flags is `true` - return a list of vectors, each vector containing the requested information.  Ideally the list of vectors should be in the same ordering as the list of maximal cones returned by `maximal_cones` (I think we can get the latter for free using the constructor of `PolyhedralFan` that was recently added by Benjamin).

ooinaruhugh:
@YueRen I changed it according to your suggestions. I need to update the docstrings but please tell me if that's what you've had in mind.

thofma:
The return values of this function is highly "un-OSCAR". It is basically of the form
```
function f(args...; return_x = true, return_y = false, return_z = true)
  output = []
  if return_x
    push!(output, x)
  end
  if return_y
    push!(output, y)
  end
  if return_z
    push!(output, z)
  end
  return output
end
```
This is does not really fit with the rest of OSCAR. It is also kind of error prone and ambiguous, if the `x,y,z` have the same type and the order of pushing is changed.

Maybe there could be a better interface? Maybe some dedicated `GroebnerFan` object?

Edit: We can also put the discussion in a different issue, since it is unrelated to the contribution here.

ooinaruhugh:
@thofma I agree with that, I don't personally like either that there is no way to know what each output is except for their order.

I would like to create a dedicated `GroebnerFan` object, but for now I extended the function in the same way as it was, 
so would it be fine to include this for now?

thofma:
Yeah, sure, if Yue is happy with this here, let's get this in.

YueRen (approved):

YueRen:
I am happy to approve the changes, but I also wouldn't mind discussing it a bit further.

@thofma @ooinaruhugh would having two functions `groebner_fan` and `groebner_fan_with_data` be more OSCAR-like?  The first one only returns a polyhedral fan, the second one a polyhedral fan with auxilliary data (specified by the flags).  To avoid the issue with the order in which the auxilliary data is served, we can just make it independent of the flags,  e.g. we have one vector per maximal cone, and:
1. the first entry is either empty or an interior point
2. the second entry is either empty or a monomial ordering
3. ...

ooinaruhugh:
@YueRen I was already thinking about a object `GroebnerFan` that would cache as much from the computation in `groebner_fan` as possible, which right now I think would be the cones (or the `PolyhedralFan` of those), the interior points and the Gröbner bases (possibly, since they've been already computed). The orderings could be recovered from the interior points, as would be the initial ideals.

Then, there would be attributes `maximal_cones` returning the cones, `interior_points` returning those, `initital_ideals` that generates a list of initial ideals etc.

I was talking to @benlorenz right now on how to assure that `maximal_cones` of the `PolyhedralFan` and the other data is ordered the same way. Also, `GroebnerFan` should be added to `_FanLikeType`so one can use the polyhedral fan directly in the corresponding methods.

YueRen:
@ooinaruhugh Sounds good!  If you're already thinking about implementing a `GroebnerFan` object, then I suggest we discuss it in a separate issue and merge this pull-request as is.

src/TropicalGeometry/groebner_fan.jl
```diff
@@ -361,53 +361,26 @@ function groebner_flip_adjacent_ordering(R::MPolyRing,
...
+
+If `return_groebner_bases==true`, above list includes the Groebner bases for those cones.  Their union will be a universal Groebner basis.
+If additionally `marked_groebner_bases==true`, the values the Groebner bases for those cones together with the leading term for each generator.  
+
+If `return_orderings==true`, above list includes the monomial orderings for those cones.  These orderings are suboptimal and hence it is generally recommended to create new orderings with the interior points.  However they do contain information on how the fan was traversed.
+
+If `return_initial_ideals==true`, above list includes the initial ideals for those cones.
 
 # Examples
 ```jldoctest
```
benlorenz, line 386: Right now the doctest does not match the output, can you adjust that?

src/TropicalGeometry/groebner_fan.jl
```diff
@@ -429,7 +402,13 @@ julia> SigmaI,gbs,ords = groebner_fan(I,return_groebner_bases=true,return_orderi
 ```
 """
```
benlorenz:
```suggestion
```
Please delete that empty line, otherwise the docstring is not attached to the function (and the doctest doesn't run).

benlorenz (changes requested):

benlorenz (approved):

PR #3216, opened 2024-01-22
add special objects for subgroup transversals

### LINKED ISSUE #3197: `right_transversal` should return a "smart" transversal object (wrapping the corresponding "smart" object produced by GAP) instead of coercing into a `Vector`

ThomasBreuer:
Use GAP's `IsRightTransversalRep` objects instead of "unpacking" the transversal in Oscar.

Currently the `PositionCanonical` functionality from GAP is not available for the left/right transversals.
This means that Oscar does not support to "act on the transversal" in the sense of GAP.

resolves #3197

### DIFF
--- a/src/Groups/cosets.jl
+++ b/src/Groups/cosets.jl
@@ -271,11 +271,69 @@ function left_cosets(G::T, H::T; check::Bool=true) where T<: GAPGroup
   return [left_coset(H, x) for x in t]
 end
 
+
+# SubgroupTransversal{T<: GAPGroup, S<: GAPGroup, E<: GAPGroupElem}
+#
+# Type of left/right transversals of subgroups in groups.
+# The elements are encoded via a right transversal object in GAP.
+struct SubgroupTransversal{T<: GAPGroup, S<: GAPGroup, E<: GAPGroupElem} <: AbstractVector{E}
+   G::T                    # big group containing the subgroup
+   H::S                    # subgroup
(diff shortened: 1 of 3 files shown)


fingolfin (approved):

fingolfin:
OK great!

Now the next thing on my wish list is that `right_cosets` returns a G-set, so that I can rewrite the following example from the book:
```
G = dihedral_group(6)
U = sub(G, [g for g in gens(G) if order(g) == 2])[1]
r = right_cosets(G, U)
perm([findfirst(isequal(c*G[1]), r) for c in r])  # get action of G[1] on the cosets in r
```
to instead use this in the last line:
```
permutation(r, G[1])
```

I can kind of fudge it right now but only by using the undocumented `Oscar.GSetByElements` directly:
```
x = Oscar.GSetByElements(G, *, r; closed=true)
permutation(x, G[1])
```

ThomasBreuer:
@fingolfin
Currently you could also use the documented `gset(G, *, r, closed = true)`, and then call `permutation` or `action_homomorphism`.

But this is of course not what one wants, the idea is to create a new G-set type that stores a `right_transversal` object instead of the explicit vector of coset objects, and to change `right_cosets` to return such a G-set.
(Then the return value of `right_cosets` will no longer be a vector. Currently `right_cosets` seems to be not used anywhere, this change is perhaps acceptable.)

PR #3217, opened 2024-01-22
Repair flattenings of graded modules

HechtiDerLachs:
Some improvements to handle certain derived pushforwards requested by externals. See the added tests for details.


src/Modules/UngradedModules/Presentation.jl
```diff
@@ -28,7 +28,7 @@ function presentation(SQ::SubquoModule)
     end
   else
     if is_graded(SQ)
-      s, _ = kernel(graded_map(ambient_free_module(SQ), gens(SQ.sum)))
+      s, _ = kernel(graded_map(ambient_free_module(SQ), filter(!iszero, gens(SQ.sum))))
```
HechtiDerLachs, line 31: @jankoboehm : There is an issue when calling `graded_map` with any zero in the second argument. The error could be fixed here by adding the `filter`, but I am not 100% sure this is legitimate. Either way, the rest of the code for `presentation` should eventually be checked as to whether it suffers from similar issues. But I found it too hard to read to actually understand what's going on and simply adding `filter` to the other calls to `graded_map` made things worse. So I only did this one change for the moment.

src/Modules/UngradedModules/SubquoModuleElem.jl
```diff
@@ -1000,7 +1000,8 @@ function is_zero(m::SubquoModuleElem)
   return (ambient_representative(m) in parent(m).quo)
 end
 
-function iszero(m::SubquoModuleElem{<:MPolyRingElem})
+# Method only for the cases which can be handled by singular
+function iszero(m::SubquoModuleElem{<:MPolyRingElem{T}}) where {T<:Union{<:FieldElem, ZZRingElem}}
```
HechtiDerLachs, line 1019: @jankoboehm : This function is not generic, but works only for rings with a singular backend. When writing up your different module layers, this should be corrected, eventually.

PR #3222, opened 2024-01-23
upgrade selected GAP packages at startup

### LINKED ISSUE #3219: Get GAP package repsn 3.1.1 or newer into Oscar
... as it fixes some issues relevant for the G-module code. Some options:
1. get a new GAP package distro snapshot
2. wait for GAP 4.13.0 (soonish)
3. just `GAP.Packages.install` a newer version
4. finally do `GAP_pkg_repsn` etc.

ThomasBreuer:
resolves #3219


lgoettgens:
I'll try to explain the Aqua issue reported here: A packages `__init__()` function gets called during precompilation of its dependents. In OSCAR's case (with this PR), `GAP.Packages.install` is called which creates a `Downloads.Downloader` object. This keeps the connection open for an extended period of time, which hinders the precompilation of OSCAR to finish (but only if OSCAR is a dependency of something).
A fix for something similar in Polymake.jl is https://github.com/oscar-system/Polymake.jl/pull/466 by @benlorenz. There also is a "real" fix in https://github.com/JuliaLang/Downloads.jl/pull/234, but this will only be available in 1.11+.

benlorenz:
That Downloads.jl fix will also be in 1.10.1.

_Edit:_
But I am not totally sure if this will fix this issue on 1.10 as there might be some other changes in the atexit code on nightly.

Also there might be something else involved here, since that Downloads.jl issue should be fixed and included on the latest julia nightly (via https://github.com/JuliaLang/julia/pull/52885).
But it did error even on nightly.

lgoettgens:
Hm, I don't see it in https://github.com/JuliaLang/julia/pull/52755 or am I just blind?

benlorenz:
The commit is there but it doesn't seem to appear in the list: [JuliaLang/julia@`aaaf5de` (#52755)](https://github.com/JuliaLang/julia/pull/52755/commits/aaaf5de5e690f782c22d42109cdf70387c1fa99b)
Maybe because it was added to that branch via a separate PR: https://github.com/JuliaLang/julia/pull/52886

ThomasBreuer:
@lgoettgens @benlorenz Thanks for the information, but I do not understand what the proposal is.
The failing tests with Julia 1.10 are run with Julia 1.10.0, which does not know `Downloads.default_downloader`.
Thus the cited fix from Polymake.jl would not help there. Or what am I missing?
And if I understand @benlorenz' comment right then the failure in Julia nightly (also in the Aqua.jl tests) indicates another (?) problem.

benlorenz:
1.10 should have that function, it has an exclamation mark at the end:
```
julia> Downloads.default_downloader!
default_downloader! (generic function with 2 methods)
```

I am indeed confused why this also happens on nightly which is supposed to clean up these downloaders automatically. Especially since the following seems to work, even though it is not really nice:
```julia
  GAP.Packages.Downloads.default_downloader!(GAP.Packages.Downloads.Downloader(grace=0.01))
  for (pkg, version) in [
     ("repsn", "3.1.1"),
     ]
    GAP.Packages.install(pkg, version, interactive = false, quiet = false)
  end
  GAP.Packages.Downloads.default_downloader!()
```
Note that this would need some extra if statement to skip these calls for 1.9 and older.

Another option would be to add a custom Downloader in the GAP packages.jl file like this:
```julia
   dlhelper = Downloads.Downloader(; grace=0.1)
...
        buffer = Downloads.download(String(url), IOBuffer(); downloader=dlhelper)
```
The dlhelper can probably be created in the `init_packagemanager` function and be used in both lambda functions.

ThomasBreuer:
Thanks @benlorenz.

Logically, the right place to fix the problem is in GAP.jl, as you suggest.
I have created a pull request for that.

This does not solve the problem in Oscar until the next release of GAP.jl. I am going to add the "not really nice" solution sketched above to the current pull request.

src/Oscar.jl
```diff
@@ -80,9 +80,21 @@ function __init__()
   GAP.Globals.BindGlobal(GapObj("Oscar"), Oscar)
   GAP.Globals.SetPackagePath(GAP.Obj("OscarInterface"), GAP.Obj(joinpath(@__DIR__, "..", "gap", "OscarInterface")))
   GAP.Globals.LoadPackage(GAP.Obj("OscarInterface"))
+  # Work around a problem in Julia 1.10 (see https://github.com/oscar-system/Oscar.jl/pull/3222).
+  isdefined(GAP.Packages.Downloads, :default_downloader!) && GAP.Packages.Downloads.default_downloader!(GAP.Packages.Downloads.Downloader(grace=0.01))
```
fingolfin: I have just release GAP.jl 0.10.2 so you could remove this and update `Project.toml` instead.

ThomasBreuer: Great. Thanks.

fingolfin (approved):

PR #3226, opened 2024-01-24
Minor fix for show method of character tables; fix printing of named groups like `SL(2,2)` in some situations

fingolfin:
Don't print the name 'SL(2,2)' as 'sL(2,2)'.


ThomasBreuer (approved):
Nice.
Just the title is not correct, the change is more general, is improves printing of `GL` etc. also in other situations.

PR #3230, opened 2024-01-24
Adjust `eigenvalues` to Nemo changes

joschmitt:
Needs Nemo / Hecke update. Just so we don't forget about it.


joschmitt:
Closing in favour of #3231 .

PR #3232, opened 2024-01-24
Exact arithmetic check for Gröbner related functionality

### LINKED ISSUE #3207: Warning about Gröbner bases with floating point arithmetic
**Problem**
We're currently teaching a course on computational algebraic geometry with Oscar, and a couple of our students have had confusing experiences resulting from trying to compute Gröbner bases over inexact "field-like" number systems like `AcbField(64)` or `ArbField(64)` (which can lead to drastically incorrect results when rounding take place in the Buchberger algorithm). This is especially confusing for those with limited programming experience, who might not be familiar enough with floating point arithmetic to realize the potential danger.

**Potential solution**
In the interest of making Oscar more beginner-friendly, it could be useful if `groebner_basis(I)` threw an error (or at least gave a sharp warning) if `base_ring(base_ring(I))` has floating point arithmetic.

**Example**
A typical problematic attempt to calculate a Gröbner basis of $\langle x^2+y^2+\frac{1}{3},x^2+xy+\frac{1}{3}x\rangle$ in $\mathbb{C}[x,y]$ could look something like this (instead of simply doing the calculation in $\mathbb{Q}[x,y]$):
```
R, (x,y) = polynomial_ring(AcbField(64),["x","y"])
I = ideal([x^2+y^2+1//3,x^2+x*y+1//3*x])
G = groebner_basis(I)
```

ederc:
This PR checks if arithmetic is exact for Gröbner releated functionality. Otherwise an error with a corresponding error message is thrown.

Tries to fix #3207.

### DIFF
--- a/src/Modules/UngradedModules/SubModuleOfFreeModule.jl
+++ b/src/Modules/UngradedModules/SubModuleOfFreeModule.jl
@@ -164,6 +164,7 @@ Compute a standard basis of `submod` with respect to the given `odering``.
 The return type is `ModuleGens`.
 """
 function standard_basis(submod::SubModuleOfFreeModule; ordering::ModuleOrdering = default_ordering(submod))
+  @req is_exact_type(typeof(base_ring(submod)(0))) "This functionality is only supported over exact fields."
   gb = get!(submod.groebner_basis, ordering) do
     return compute_standard_basis(submod, ordering)
   end::ModuleGens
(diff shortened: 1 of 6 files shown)


src/Modules/UngradedModules/SubModuleOfFreeModule.jl
```diff
@@ -164,6 +164,7 @@ Compute a standard basis of `submod` with respect to the given `odering``.
 The return type is `ModuleGens`.
 """
 function standard_basis(submod::SubModuleOfFreeModule; ordering::ModuleOrdering = default_ordering(submod))
+  @req is_exact_type(typeof(base_ring(submod)(0))) "This functionality is only supported over exact fields."
```
lgoettgens:
```suggestion
  @req is_exact_type(elem_type(base_ring(submod))) "This functionality is only supported over exact fields."
```
this avoids constructing an element. can be applied to most other places as well

ederc: Thanks, adjusted correspondingly.

fingolfin (approved):

lgoettgens (approved):

PR #3233, opened 2024-01-24
Fix function name in docstring

StevellM:


experimental/SymmetricIntersections/src/representations.jl
```diff
@@ -617,8 +617,8 @@ function is_faithful(rep::LinRep)
 end
 
 @doc raw"""
-    is_projectively_faithful(rep::LinRep, p::GAPGroupHomomorphism) -> Bool
-    is_projectively_faithful(chi::Oscar.GAPGroupClassFunction, p::GAPGroupHomomorphism) -> Bool
+    is_faithful(rep::LinRep, p::GAPGroupHomomorphism) -> Bool
+    is_faithful(chi::Oscar.GAPGroupClassFunction, p::GAPGroupHomomorphism) -> Bool
 
 Given a linear representation `rep` of the domain `E` of the cover `p` affording the
 character `chi`, return whether `rep` is `p`-faithful, i.e. `rep` reduces to a faithful
```
fingolfin, line 624: Below the fold it says
> This is equivalent to ask that

which should be changed to
> This is equivalent to asking that

fingolfin (approved):

PR #3234, opened 2024-01-24
deprecate SimplicialComplex for simplicial_complex

antonydellavecchia:
Deprecates SimplicialComplex for simplicial_complex, and transpose incidence matrix of a graph.


src/Combinatorics/SimplicialComplexes.jl
```diff
@@ -13,26 +13,26 @@ pm_object(K::SimplicialComplex) = K.pm_simplicialcomplex
 
 
 @doc raw"""
-    SimplicialComplex(generators::Union{Vector{Vector{Int}}, Vector{Set{Int}}})
+   simplical_complex(generators::Union{Vector{Vector{Int}}, Vector{Set{Int}}})
```
fingolfin:
```suggestion
    simplical_complex(generators::Union{Vector{Vector{Int}}, Vector{Set{Int}}})
```
It is important that this first line is indented by (at least) four characters: that's Markdown syntax for a "code block".

antonydellavecchia: Thanks!

src/Combinatorics/Graphs/functions.jl
```diff
@@ -588,7 +588,9 @@ julia> incidence_matrix(g)
 []
 ```
 """
-incidence_matrix(g::Graph{T}) where {T <: Union{Directed, Undirected}} = IncidenceMatrix(Polymake.graph.incidence_matrix(pm_object(g)))
+function incidence_matrix(g::Graph{T}) where {T <: Union{Directed, Undirected}}
+  IncidenceMatrix(transpose(Polymake.graph.incidence_matrix(pm_object(g))))
```
benlorenz: Why is this transposed now? (It does not fit to the doctest anymore)

antonydellavecchia: it can now be passed to simplicial_complex correctly. I just pushed a fix for the docs. It also aligns with signed_incidence_matrix which considers the columns as edges

benlorenz: > it can now be passed to simplicial_complex correctly.

That is not really an argument to change the output here. The julia Graphs.jl package also gives a 5x2 matrix.

> It also aligns with signed_incidence_matrix which considers the columns as edges

But you just changed it to have the edges as rows? (the docs now have a 2x5 matrix for a graph on 5 nodes with 2 edges)

Also currently on master we have:
```
julia> g = Graph{Directed}(5);

julia> add_edge!(g,1,2); add_edge!(g,2,3);;

julia> signed_incidence_matrix(g)
5×2 Matrix{Int64}:
 -1   0
  1  -1
  0   1
  0   0
  0   0

julia> incidence_matrix(g)
5×2 IncidenceMatrix
[1]
[1, 2]
[2]
[]
[]
```
which does fit together.

antonydellavecchia: this was resolved by adding a function that converts a graph to a simplicial complex, and leaving `incidence_matrix(G)` and `simplicial_complex(IM::IncidenceMatrix)` as is

benlorenz (approved):

PR #3235, opened 2024-01-24
change `show` for nat. symmetric/alternating groups

### LINKED ISSUE #3225: Restore printing of symmetric group as `Sym(n)`
It was requested that a group created by `symmetric_group(n)` is again printed as `Sym(n)` (or perhaps some variation of that); and similar for `alternating_group(n)` and perhaps some other "constructors".

This is only for groups created this way; if a group is e.g. an alternating group "by chance" then we don't check for that (well, if the GAP filter `HasIsNaturalSymmetricGroup` is set then we can of course use it, but we shouldn't trigger a computation).

This is consistent with how e.g. `GL(3,3)` is printed.

The idea is *not* to use GAP's `Name` attribute, which has its own issues.

ThomasBreuer:
and adjust the affected doctests

resolves #3225

### DIFF
--- a/experimental/GaloisGrp/src/Solve.jl
+++ b/experimental/GaloisGrp/src/Solve.jl
@@ -290,12 +290,12 @@ one, compute the corresponding subfields as a tower.
 julia> Qx, x = QQ["x"];
 
 julia> G, C = galois_group(x^3-3*x+17)
-(Permutation group of degree 3 and order 6, Galois context for x^3 - 3*x + 17 and prime 7)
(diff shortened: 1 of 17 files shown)


thofma:
There are some printing tests in `test/`, which need to be adjusted.

ThomasBreuer:
Yes, these failing tests are doctests, but they did not get fixed automatically by `build_doc(doctest = :fix)`.
(Thinking about it now, it is clear that this cannot be expected.)

fingolfin (approved):

PR #3240, opened 2024-01-25
Adapt some Group function docstrings

lgoettgens:
As they used to contain `::Group` in the signature even though they actually only work for `GAPGroup`s.

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -177,7 +177,7 @@ julia> exponent(symmetric_group(13))
 Base.exponent(::Type{T}, G::GAPGroup) where T <: IntegerUnion = T(GAP.Globals.Exponent(G.X)::GapInt)
 
 """
-    rand(rng::Random.AbstractRNG = Random.GLOBAL_RNG, G::Group)
+    rand(rng::Random.AbstractRNG = Random.GLOBAL_RNG, G::GAPGroup)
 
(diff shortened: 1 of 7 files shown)


ThomasBreuer (commented):
Is this really desirable?
Currently we have `GAPGroup`s and `GrpAbFinGen`s, and the same docstrings should hold for both of them. Do we want to duplicate all docstrings for `GrpAbFinGen`s?
And if another kind of groups comes up then `gens`, `sylow_subgroup`, `normalizer` etc. should behave for them according to the docstrings.
(Of course this holds only where these functions make sense at all, but this is the case already now, for example many functions are meaningful only for finite groups).

fingolfin (changes requested):
As @ThomasBreuer explained we wrote these docstrings like that quite intentionally.

if there are concrete issued with that, let's discuss them specifically.

lgoettgens:
I am afraid that the implementation of weyl groups and GrpGen does not offer all of that functionality. So users may expect dispatches to be there even though they aren't.
If you want to keep it as is, that can be a deliberate decision and I may live with it.

fingolfin:
We could add a generic fallback method to each of these printing a "not implemented for this type of group" error. We could also add a remark in each docstring that not every group may implement all of these (for now).

Although I hope to change this at some point, by adding generic fallback mechanisms as in GAP.

PR #3244, opened 2024-01-25
Adjust casing in `show(::MatroidRealizationSpace)`

lgoettgens:
While doing other stuff I noticed that there is a `Lowercase` missing before the ring. There are certainly more improvements in the printing possible, but this is already a start.

### DIFF
--- a/experimental/MatroidRealizationSpaces/src/realization_space.jl
+++ b/experimental/MatroidRealizationSpaces/src/realization_space.jl
@@ -33,12 +33,13 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", RS::MatroidRealizationSpace)
   if has_attribute(RS, :is_realizable) && !is_realizable(RS)
-    if RS.char == nothing && RS.q == nothing
+    if RS.char === nothing && RS.q === nothing
       print(io, "The matroid is not realizable.")
     else
       print(io, "The matroid is not realizable over the specified field or characteristic.")
     end
   else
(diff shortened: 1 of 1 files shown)


experimental/MatroidRealizationSpaces/src/realization_space.jl
```diff
@@ -33,25 +33,25 @@ end
...
-
+    print(io, Indent())
     show(io, MIME("text/plain"), RS.realization_matrix)
-    print(io, "\nin the ", RS.ambient_ring)
+    print(io, "\n", Dedent(), "in the ", Lowercase(), RS.ambient_ring)
     I = RS.defining_ideal
-    if (I isa NumFieldOrdIdl && I.gen != ZZ(0)) ||
-      (I isa Ideal && !iszero(I))
+    if !iszero(I)
       print(io, "\nwithin the vanishing set of the ideal\n", I)
```
fingolfin, line 55: probably should have an `Indent` here, too, and also in the final `print` after the fold -- but of course that's beyond the scope of this PR, just wanted to mention it.

fingolfin (approved):

experimental/MatroidRealizationSpaces/src/realization_space.jl
lgoettgens: Yeah, I played around with some things here but wasn't happy with all of them. So that got postponed

benlorenz (approved):

Issue #2166, opened 2023-03-30
Implement the new printing guidelines for more compact printing


simonbrandhorst:
#2148 adds the new printing guidelines to our developer documentation. 
They need to be implemented for our existing Types.  
The first ones to do this seem to be the following:

- [ ] basic rings such as QQ, ZZ, GF
- [ ] polynomial rings
- [ ] number fields
- [ ] homomorphisms of rings
- [ ] groups
- [ ] group homomorphisms

Once these are implemented we can go for the more complicated constructs such as modules, group rings, (toric) varieties etc. 

### TODOs

Implement some some fallback functions

```julia
show_details(x) = show(stdout, "text/plain", x)

show_one_line(x) = show(stdout, x)

show_supercompact(x) = show(IOContext(stdout, :supercompact => true), x)
```

and also provide some macros

```
@show_details x
@show_one_line x
@show_super_compact x
```
 - Helper for `with 1 generator` and not `with 1 generators`
 - make indentation work by using stuff from https://github.com/KristofferC/IOIndents.jl
   the nested call will then be `print(indented(io), R.substruct)`
 - Max wants to add a few more examples so we can discuss them:
   - detailed group homomorphisms given by generator-images, including the mappings (`f1 -> (1,2)(3,4,5)`)
   - permutations
   - matrices (in detailed view they could print their ring?)
   - matrices-as-group-elements (should be distinguishable from plain matrices? at least in detailed mode)
 - Max: Right now all output starts with an uppercase letter. Drawback: this makes it harder to compose outputs.
   Perhaps we need an option to control that? Or we write a helper function
   which changes the case of the first letter of a string --
   but then we need to first print into strings, instead of directly printing to the output io,
   which means overhead and has other concerns

simonbrandhorst:
# Examples for the new 2+1 print modes in Oscar
These examples are intended as guidelines for implementing the new show
functions.

## Examples
For each example we fix a single parent and display its three print modes
consecutively each in its own block
`:details`
`:oneline`
`:supercompact`
### Coefficient rings
```
Integer ring
```
```
Integer ring
```

```
ZZ
```
-----
```
Rational field
```
```
Rational field
```
```
QQ
```
----
##### Prime field
```
Finite field of characteristic 2
```
```
Finite field of characteristic 2
```
```
GF(2)
```
##### Finite field
```
Finite field of degree 7 over GF(29)
```
```
Finite field of degree 7 over GF(29)
```
```
GF(29^7)
[...]

PR #3250, opened 2024-01-26
Implement printing guidelines for multivariate ideals

joschmitt:
Implement the printing guidelines for `MPolyIdeal` and `InvRing`.
For the ideals, I followed the example in #2166, with the exception that so far all generators are printed in detailed mode. I can adjust this, but I wasn't sure what the outcome of this general discussion was (the analogous question was discussed e.g. in #2878).

Note: the printing of `MPolyLocalizedIdeal` should be adjusted consistently (once there is an agreement for `MPolyIdeal`).


src/Serialization/main.jl
```diff
@@ -398,13 +398,44 @@ julia> meta = metadata(author_orcid="0000-0000-0000-0042", name="42", descriptio
 Oscar.MetaData("0000-0000-0000-0042", "42", "The meaning of life, the universe and everything")
 
 julia> save("/tmp/fourtitwo.json", 42; metadata=meta);
+ERROR: ArgumentError: '/tmp/fourtitwo.json' exists. `force=true` is required to remove '/tmp/fourtitwo.json' before copying.
```
fingolfin: Error?

docs/src/CommutativeAlgebra/localizations.md
```diff
@@ -111,11 +111,14 @@ This reflects the way of creating localizations of quotients of multivariate pol
...
+Ideal
+  of multivariate polynomial ring in 3 variables over QQ
+generated by
+  x
 
 julia> U = complement_of_prime_ideal(P)
 Complement
-  of prime ideal(x)
+  of prime ideal with 1 generator in multivariate polynomial ring
   in multivariate polynomial ring in 3 variables over QQ
```
fingolfin: In places like this, we loose information while now printing redundant information. Don't get me wrong, at the same time there are good reasons for the change in this PR (overall I like it!).

But I wonder if that means we perhaps need another printing mode, or just some special cases where we don't just recurse but do something specific when recording? 

Like here, perhaps the complement function should print a list of generators of the ideal? If the printing here was any of these, I think it would be better (but perhaps not in general?)
```
Complement of
  prime ideal
    of multivariate polynomial ring in 3 variables over QQ
  generated by
    x
```
or
```
Complement of
  prime ideal generated by
    x
in multivariate polynomial ring in 3 variables over QQ
```
Of course in this specific case some people might also like
```
Complement of
  prime ideal P
in R
```
or
```
Complement of
  prime ideal generated by [ x ]
  in QQ[x, y, z]
```

but both also have downsides, IMHO...


Perhaps @simonbrandhorst or @fieker or @thofma or others have something to say...

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
```diff
@@ -86,17 +86,17 @@ julia> (x, y) = coordinates(A)
 julia> X = algebraic_set(ideal([y - x^2]))
 Affine algebraic set
   in affine 2-space over QQ with coordinates [x, y]
-defined by ideal(-x^2 + y)
+defined by Ideal with 1 generator in multivariate polynomial ring
```
fingolfin: Another situation where the old printing seems strictly more useful than the new printing :-/.

Of course that might change if there are 50 ideal generators, each spanning three lines...


BTW, there seems to be a `Lowercase()` missing in the show method for `Affine algebraic set`

src/Rings/mpoly.jl
```diff
@@ -718,12 +718,30 @@ Fields:
...
 end
 
+function Base.show(io::IO, ::MIME"text/plain", I::MPolyIdeal)
+  io = pretty(io)
+  println(io, "Ideal")
+  println(io, Indent(), "of ", Lowercase(), base_ring(I))
+  print(io, Dedent(), "generated by")
+  for f in gens(I)
+    print(io, "\n", Indent(), f, Dedent())
+  end
```
fingolfin: It seems for a zero ideal we just print an "empty list". That could be confusing, perhaps better to have a special case: `if ngens(I) == 0 println(io, "Zero ideal")` etc. ?

src/Serialization/main.jl
joschmitt: Thanks. I guess this is what happens if you use `doctest = :fix` without heeding the warnings...

fingolfin (commented):
Overall I agree with the direction, but I wonder if we can't do something to help those cases were IMHO the new printed output is strictly worse than the old. Of course that's just about my subjective *opinion* -- I am not strictly opposed, and if others feel this is the way to go, I am happy enough (also, we can always improve again later).

That said it'd be great if others could chime in, to see if we can get a (partial) consensus

fingolfin:
Oh, and I'd like to be very clear: I am very grateful for @joschmitt to have made this PR! I don't have a better idea than what is in here :-)

src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Constructors.jl
joschmitt: I agree that for a few, small generators the old printing is better. In the end this gets us back at the discussion, whether objects should print all their generators or not or just some.
Here, I followed the example in #2166 by @simonbrandhorst , but we could of course also print something like
```
Ideal generated by x1, x2, ..., xn in multivariate polynomial ring
```
in one-line mode. If one generator is just huge this will of course explode again.

joschmitt:
I changed the one-line mode now, so that the generators are printed if they need not more than 20 characters. See this example from the doc tests:
```
(Ideal with 5 generators in multivariate polynomial ring, Ideal (17, d, c, b, a) in multivariate polynomial ring)
```
We can change the 20 to something else and we can also change the `(x, y, z)` to something else. (Angular brackets are probably forbidden in non-unicode land?)

simonbrandhorst:
My impression is also that it looks worse than before. The one line print mode is useless if we just print the number of generators. 
`ideal(x,y, z)` seems like the perfect way to print it in one line mode. No need to print the parent. 
Edit: In some ways ideals are more like elements and there we do not hide information / and do not print the parent as well.

joschmitt:
I'm not sure I understand. Do you just propose
1) to not print the parent (in one-line mode or maybe at all) or
2) to always print *all* generators in one-line mode? 

(Or both.)
Point 2 would contradict my understanding of "one-line mode".
Don't get me wrong: I'm all in favour of printing all generators in detailed mode because in my opinion the detailed mode should show enough information to define the object (and I voiced this opinion before regarding the question whether groups should print all their generators). But the one-line mode is for example used in a quotient ring (I think?) where I wouldn't want the screen to "explode" because the modulus has a thousand generators. (And if I want to see them, I can do `modulus(R)` to see the detailed mode.)

Here is a new proposal:
#### Detailed mode:
```
Ideal generated by
  first generator
  second generator
  and so on
```
#### One-line mode:
```
Ideal (comma separated list of generators)
```
... assuming this fits a fixed number of characters (80 or 100 or 120 or whatever a "line"is) and
```
Ideal with n generators
```
... if the list of generators would be too long.
#### Supercompact:
```
Ideal
```

And just to say: I feel like the examples we see in the documentation might be misleading since they almost always show ideals with very few and small generators.

simonbrandhorst:
Thanks at lot for the new proposal. I think it is close to optimal! 
Something to think about: Do we want a) `Ideal(...)` or b) `Ideal (...)` ?
I would prefer version a) since it is reminiscent of a function `f(x)` not `f (x)` .

joschmitt:
My thought was that `Ideal (x, y, z)` should be read as "the ideal $\langle x, y, z\rangle$" (maybe we could/should even print angular brackets if the user allows unicode). I considered printing `Ideal generated by x, y, z` but this seems unnecessarily long and considering Murphy's law, there will probably be a situation where something prints as `... Ideal generated by x, y, z and f ...` where the brackets would be important.
I must say that I always found it very weird that `MPolyIdeal` is the only type that "prints like you would type it". I don't think this principle is followed anywhere else in OSCAR. Sure, polynomials print as polynomials, but `one(R)` prints as `1` for almost any ring `R` and if you type `1` you don't even get an OSCAR `RingElem`.
So, basically, this question of space or no space seems to me to be a question about whether we print mathematical information or computer/programming information. I have never heard anybody talk about ideals as "a mapping from the polynomial ring to the set of ideals", so I wouldn't print it like that.

fingolfin:
I agree with @joschmitt . How about just using plain ASCII `<` and `>`, so e.g. `Ideal <x, y, z>`?

simonbrandhorst:
Plain asci `<x, y, z>` just looks horrible.  But in unicode we can use the $\langle \rangle$ if available.

simonbrandhorst:
Oh and i would agree that `ideal generated by (...) ` is too much and should be avoided. 
Note that in commutative algebra it is not uncommon to denote the ideal generated by `x,y` by `(x, y)` ... thus lets please stick to round braces. Maybe even in unicode for consistency?

joschmitt:
I implemented now what I believe is the consensus. I also adjusted the printing of `MPolyLocalizedIdeal` accordingly.

simonbrandhorst:
Is long output in one line mode now truncated .. at how many characters? I could not find an example in the diff.

src/InvariantTheory/affine_algebra.jl
```diff
@@ -61,13 +61,11 @@ Matrix group of degree 3
...
-Invariant ring of
-  Matrix group of degree 3 over cyclotomic field of order 3
-with generators
-  AbstractAlgebra.Generic.MatSpaceElem{nf_elem}[[0 0 1; 1 0 0; 0 1 0], [1 0 0; 0 a 0; 0 0 -a-1]]
+Invariant ring
+  of matrix group of degree 3 over cyclotomic field of order 3
 
 julia> affine_algebra(IR)
-(Quotient of multivariate polynomial ring by ideal(y1^6 - 3*y1^4*y3 - 16*y1^3*y2^3 - 4*y1^3*y4 + 3*y1^2*y3^2 + 24*y1*y2^3*y3 + 4*y1*y3*y4 + 72*y2^6 + 24*y2^3*y4 - y3^3 + 8*y4^2), Hom: quotient of multivariate polynomial ring -> graded multivariate polynomial ring)
+(Quotient of multivariate polynomial ring by ideal with 1 generator, Hom: quotient of multivariate polynomial ring -> graded multivariate polynomial ring)
```
joschmitt, line 68: @simonbrandhorst here is an example where the generator(s) are not printed in one-line mode. I put the limit at 100 characters.

simonbrandhorst (approved):

simonbrandhorst:
This is certainly an improvement. Thank you @joschmitt

Issue #3202, opened 2024-01-19
Schemes Meta Issue


simonbrandhorst:
This is a todo list for the schemes part of OSCAR

- [x] implement `defining_ideal(::AbsSpec)`. In case of a closed subscheme of AA^n it should return an ideal in the corresponding polynomial ring, in case of a closed subscheme of some open subscheme  U of AA^n it should return an ideal in `OO(U)` https://github.com/oscar-system/Oscar.jl/pull/3205
- [x] rename `Spec` to `AffineScheme` https://github.com/oscar-system/Oscar.jl/issues/3345
- [x] change one line printing of closed subschemes of `AA^n` from `V(...)` to `scheme(...)`
- [x] rename `issubset` for (affine & projective) schemes to `is_subscheme` and implement `issubset` 
- [x] migrate ideal sheaves and divisors to src
- [ ] implement weighted projective space including `is_quasi_smooth` 
@jankoboehm @HechtiDerLachs @afkafkafk13 @paemurru 
- [x] implement normalisation for affine schemes
- [x] implement normalisation for covered schemes
- [x] desungularisation for (covered) surfaces via blowup and normalisation
- [ ] check if the singular code for invert_birational_map works more generally and adapt it. Remove the duplicate for plane curves.

HechtiDerLachs:
Let me try to do the `defining_ideal` thing today.

fingolfin:
What's the status of this?

simonbrandhorst:
Some progress, updated.

paemurru:
About weighted projective spaces and `is_quasi_smooth` - weighted projective spaces already exist as toric varieties. Quasismoothness test should then be implemented more generally for subschemes in toric varieties. So I think it is more about toric varieties and can be removed from here.

PR #3252, opened 2024-01-26
Rename issubset to is_subscheme for subschemes (#3202)

paemurru:
The `issubset` functions where the arguments are not schemes were (obviously) not renamed. This change was rather tricky to implement because

* sometimes `U` stands for a scheme and we need `is_subscheme`, sometimes `U` is an `inverted_set` and we need `issubset`,
* the functions do not have the output type written out in the docstring, sometimes being unclear whether the output is a scheme,
* some functions are type unstable, so that trying a simple example did not help either,
* Julia uses dynamic typing so that the errors (which are only type errors in this case) only come out when running the lengthy tests.


simonbrandhorst:
Thanks for the effort. I appreciate it!

lgoettgens:
Can you add deprecations for all renamed functions to src/Deprecations.jl?

simonbrandhorst:
At some point before 1.0 `is_subset` will be implemented again with a different meaning (one of the reasons for renaming is_subset to is_subscheme here).
This will be a breaking change. Would this still go via a deprecation first?

lgoettgens:
Ah ok, in this case I wouldn't do a deprecation.
If you want to use the same name again, you need to make sure that there is a Oscar release somewhere between the deletion of the old function (or rename here) and the introduction of the new function. Otherwise, users of Oscar just get different results without any way of an error

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
```diff
@@ -183,7 +183,7 @@ function preimage(
   )
   X = domain(phi)
   Y = codomain(phi)
-  check && (issubset(Z, Y) || (Z = intersect(Y, Z)))
+  check && (is_subscheme(Z, Y) || (Z = intersect(Y, Z)))
```
fingolfin, line 186: This is not fault of this PR, but I think this code is too clever for its own good and should be changed; don't do an assignment in what looks like a logical expression, one could mistake the `=` for a `==`. Instead do
```suggestion
  if check && !is_subscheme(Z, Y)
    Z = intersect(Y, Z)
  end
```
However, more seriously, I also wonder why it does that -- normally I would expect a failed `check` to trigger an error, not to "correct" the problem. I can't think of any other place we do this, and indeed in this PR multiple other `check`s are modified that all do an error (but I do not claim to have checked all place, neither in this PR nor in all of Oscar)

paemurru: Actually, at least according to [Julia docs](https://docs.julialang.org/en/v1/manual/control-flow/), it is perfectly fine to use assignment together with short-circuit evaluation:
```
julia> true && (x = (1, 2, 3))
(1, 2, 3)
```

About `check`: we can take the preimage of any set by first taking the intersection with the codomain, so it looks OK to me. The user has the option to set `check` to `false` if they know that `Z` is a subscheme and they wish to avoid checking the containment.

thofma: I don't think the question was whether this is valid julia code or not.

It is a question about how much time someone has to spend to understand this, when the code has to be improved/rewritten/fixed in a few years, and the original author is not available.

fingolfin: As Tommy explained, this is not about what is valid syntax, this is about what is maintainable.

Moreover, `check` everywhere else means "raise an error if the input is invalid", and here it now means something different. That's bad. For example, before one could set `check=true` while e.g. debugging to verify one has no bug in the code, and if there is no error, one has evidence for that, while if there is a bug, one may see an error and can debug it.

But here `check` silently papers over mistakes. So presumably no error will be raised, and the behavior may even differ between `check=true` and `check=false`. I don't think that's acceptable.

But of course this is not an issue with this PR, it is an issue with the existing code, so it can be fixed in a separate PR. But fixed it should be.

paemurru: I completely agree. The name of the variable (`check`) should clearly indicate what it is checking (`check_z_contained_in_y`). And there should be consistency whether the function returns an error whenever the check fails. It might be worth elaborating on this in the [OSCAR styleguide](https://docs.oscar-system.org/stable/DeveloperDocumentation/styleguide/) (it only briefly mentions input sanity checks).

lgoettgens (approved):
The only thing blocking this right now for me is the broken CI.

PR #3254, opened 2024-01-29
renaming: `haspreimage/has_preimage` -> `has_preimage_with_preimage`

StevellM:
Should logically follow https://github.com/thofma/Hecke.jl/pull/1367. So for now it stays as a draft.

This decision comes from the HackMD https://hackmd.io/nRnyfrSxTVe5CzXidB1cBQ, where it was suggested to merge both `haspreimage` and `has_preimage` to only use `has_preimage_with_preimage`. As far I have checked, all such functions have two outputs.

Two of them, from the GAPGroup part, have a `check` keyword argument. The reason, as spotted in several gap-related issues and in comment in the code, is that the underlying GAP function may have unexpected behaviour if the said element is not in the image of the input function. Hence, as mentioned there, the `check` argument is kept until some decisions are made on the GAP side (so I did not dare removing it)

### DIFF
--- a/docs/src/Groups/grouphom.md
+++ b/docs/src/Groups/grouphom.md
@@ -80,7 +80,7 @@ julia> x^f
 
 A sort of "inverse" of the evaluation is the following
 ```@docs
-haspreimage(f::GAPGroupHomomorphism, x::GAPGroupElem; check::Bool = true)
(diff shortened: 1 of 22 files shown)


StevellM:
Close this PR since everything has been covered in https://github.com/oscar-system/Oscar.jl/pull/3288.

PR #3260, opened 2024-01-29
Remove obsolete `getindex` methods for accessing generators

fingolfin:
AA defines this generically, with one exception `R[0]` is not defined, yet some code uses it to represent `zero(R)`. Such code should be changed to use `R(0)` or `zero(R)`.

I fixed two of these, but there might be more, we'll see.

### DIFF
--- a/src/Rings/PBWAlgebra.jl
+++ b/src/Rings/PBWAlgebra.jl
@@ -249,10 +249,6 @@ function gen(R::PBWAlgRing, i::Int)
   return PBWAlgElem(R, gen(R.sring, i))
 end
 
-function Base.getindex(R::PBWAlgRing, i::Int)
-  return gen(R, i)
-end
-
 function var_index(a::PBWAlgElem)
   return Singular.var_index(a.sdata)
 end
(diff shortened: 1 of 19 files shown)


lgoettgens (approved):

src/Rings/PBWAlgebra.jl
```diff
@@ -20,7 +20,7 @@ mutable struct PBWAlgElem{T, S} <: NCRingElem
   sdata::Singular.spluralg{S}
 end
 
-mutable struct PBWAlgIdeal{D, T, S}
+mutable struct PBWAlgIdeal{D, T, S} <: Ideal{PBWAlgIdeal{T, S}}
```
lgoettgens: Should this be recursive? I expected something along the lines of `Ideal{PBWAlgebra{...}}`

fingolfin: Actually it should be
```suggestion
mutable struct PBWAlgIdeal{D, T, S} <: Ideal{PBWAlgElem{T, S}}
```

Issue #3257, opened 2024-01-29
Some printing code ignores `is_unicode_allowed()` and prints unicode "wedge" characters


fingolfin:
Specifically these places should be fixed
- [ ] `experimental/LieAlgebras/src/LieAlgebraModule.jl:1047` (by @lgoettgens ?) https://github.com/oscar-system/Oscar.jl/pull/3270
- [ ] `experimental/LieAlgebras/src/LieAlgebraModule.jl:1049` https://github.com/oscar-system/Oscar.jl/pull/3270
- [ ] `src/Modules/ExteriorPowers/FreeModules.jl:76` (by @HechtiDerLachs ?) #3263.
- [ ] `src/Modules/ExteriorPowers/SubQuo.jl:59` #3263

You can just print `^` instead of `∧` if unicode is not allowed.

lgoettgens:
Unfortunately, that is not an easy solution, as the places you pointed out don't print anything. Instead, they just construct the symbol used for a variable/basis element. If I construct an object while allowing unicode and then later disallow it, the previous constructed object will still print using the wedge symbol.

The simplest solution that comes to my mind is to save to vectors of symbols: one using unicode and one without; then the show function can choose with one to use based on `Oscar.is_unicode_allowed()`.

fingolfin:
@lgoettgens OK.

Perhaps it could also use `expressify` / `show_via_expressify` from AA? This helps avoid ugly things like `(1)*(e1^e2) + (-1)*(e1^e3)` which frequently pop up in GAP.

lgoettgens:
> Perhaps it could also use `expressify` / `show_via_expressify` from AA? This helps avoid ugly things like `(1)*(e1^e2) + (-1)*(e1^e3)` which frequently pop up in GAP.

It does, but only for pretty-printing the sum of coefficients and basis elements. However, the wedge is part of the basis vector printing, so it would need to do some iterate expressify or something...

lgoettgens:
@fingolfin Would you be fine to adapt the docstring of `allow_unicode` to state that objects may follow the state of the flag at the point of their creation instead of at the point of the printing? For a user, this means that they should decide at the beginning of the session if they want to have unicode or  not. But for the above cases it would be way more convenient to implement (cf. my previous comments).

cc @HechtiDerLachs

fingolfin:
@lgoettgens fine by me.

[...]

PR #3263, opened 2024-01-29
Fix printing of exterior powers of modules

HechtiDerLachs:
This should address #3257.

CC: @jankoboehm .

### DIFF
--- a/src/Modules/ExteriorPowers/Generic.jl
+++ b/src/Modules/ExteriorPowers/Generic.jl
@@ -21,7 +21,7 @@ function show_exterior_product(io::IO, M::ModuleFP)
   if is_unicode_allowed()
     print(io, "⋀^$p($F)")
   else
-    print(io, "$p-th exterior power of $F")
+    print(io, "$(ordinal_number_string(p)) exterior power of $F")
   end
 end
 
@@ -32,7 +32,7 @@ function show_exterior_product(io::IO, ::MIME"text/html", M::ModuleFP)
   if is_unicode_allowed()
     print(io, "⋀^$p($F)")
   else
-    print(io, "$p-th exterior power of $F")
+    print(io, "$(ordinal_number_string(p)) exterior power of $F")
   end
 end
 
--- a/src/Modules/ExteriorPowers/Generic.jl
+++ b/src/Modules/ExteriorPowers/Generic.jl
@@ -18,14 +18,22 @@ end
 function show_exterior_product(io::IO, M::ModuleFP)
   success, F, p = is_exterior_power(M)
   success || error("module is not an exterior power")
-  print(io, "⋀^$p($F)")
+  if is_unicode_allowed()
+    print(io, "⋀^$p($F)")
+  else
+    print(io, "$p-th exterior power of $F")
+  end
 end
 
 function show_exterior_product(io::IO, ::MIME"text/html", M::ModuleFP)
   success, F, p = is_exterior_power(M)
   success || error("module is not an exterior power")
   io = IOContext(io, :compact => true)
-  print(io, "⋀^$p$F")
+  if is_unicode_allowed()
+    print(io, "⋀^$p($F)")
+  else
+    print(io, "$p-th exterior power of $F")
+  end
 end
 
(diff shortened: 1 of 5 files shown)


test/Modules/ExteriorPowers.jl
```diff
@@ -130,12 +130,14 @@ end
   u = (F[1], F[4], F[3])
 
   @test "$(mm(v))" == "e[1]∧e[3]∧e[4]"
-  @test "$(F3)" == "⋀^3(Free module of rank 5 over Multivariate polynomial ring in 5 variables over QQ)"
+  #@test "$(F3)" == "⋀^3(Free module of rank 5 over Multivariate polynomial ring in 5 variables over QQ)"
+  @test "$(F3)" == "3-th exterior power of Free module of rank 5 over Multivariate polynomial ring in 5 variables over QQ"
```
fingolfin: should of course be `3rd` not `3-th`.

We have `ordinal_number_string` for this.

HechtiDerLachs: Thanks. Didn't know that.

lgoettgens:
This approach has exactly the drawbacks I pointed out in https://github.com/oscar-system/Oscar.jl/issues/3257#issuecomment-1914548021, i.e. the following is not conformant with the unicode preference:
```julia
Oscar.allow_unicode(true)
R, (x,y) = QQ["x","y"]
F = free_module(R, 3)
E = exterior_power(F, 2)
Oscar.allow_unicode(false)
println(gens(E))
```

HechtiDerLachs:
@lgoettgens : You are right. But do we really expect a user to complain about that ever? Who gets the idea to switch on unicode all of a sudden in the middle of some computation?

lgoettgens:
I moved the discussion back to https://github.com/oscar-system/Oscar.jl/issues/3257 with a proposal to adapt the documentation of `allow_unicode`

HechtiDerLachs:
Given @fingolfin 's reply on #3257, this is now ready for merge.

lgoettgens (approved):

test/Modules/ExteriorPowers.jl
```diff
@@ -129,13 +129,15 @@ end
   v = (F[1], F[3], F[4])
   u = (F[1], F[4], F[3])
 
-  @test "$(mm(v))" == "e[1]∧e[3]∧e[4]"
-  @test "$(F3)" == "⋀^3(Free module of rank 5 over Multivariate polynomial ring in 5 variables over QQ)"
+  @test "$(mm(v))" == "e[1]^e[3]^e[4]"
+  #@test "$(F3)" == "⋀^3(Free module of rank 5 over Multivariate polynomial ring in 5 variables over QQ)"
+  @test "$(F3)" == "3rd exterior power of Free module of rank 5 over Multivariate polynomial ring in 5 variables over QQ"
```
simonbrandhorst, line 134: Free and Multivariate should be lowercase. See the comment above on how to achieve this. 
(I suppose the Multivariate still comes from old module printing). 
In any case this need not block merging of this. It is an improvement in any case.

src/Modules/ExteriorPowers/Generic.jl
```diff
@@ -18,14 +18,22 @@ end
 function show_exterior_product(io::IO, M::ModuleFP)
   success, F, p = is_exterior_power(M)
   success || error("module is not an exterior power")
-  print(io, "⋀^$p($F)")
+  if is_unicode_allowed()
+    print(io, "⋀^$p($F)")
+  else
+    print(io, "$(ordinal_number_string(p)) exterior power of $F")
+  end
```
simonbrandhorst, line 25:
```suggestion
  io = pretty(io)
  if is_unicode_allowed()
    print(io, "⋀^$p($F)")
  else
    print(io, "$(ordinal_number_string(p)) exterior power of ", Lowercase(), F)
  end
```

lgoettgens: bump @HechtiDerLachs

simonbrandhorst:
I merge at this point since this is certainly an improvement.

Issue #3224, opened 2024-01-24
Implement new G-module names


fingolfin:
A long time ago we discussed the "right" names for various functions related to G-modules (and perhaps long-term also *beyond* that ...).

The result is a bunch of stub functions in `experimental/GModule/GModule.jl`, e.g.:
```
"""
    extension_of_scalars(M::GModule, phi::Map)

Return the `S`-module obtained by extending the scalars of `M`
from `R` to `S`, where `phi` is an embedding of `R` into `S`.

If `M` has `R`-rank `d` then the returned module has `S`-rank `d`.

The syntax `M ⊗ phi` is supported.

# Examples

(cases that `S` is a finite field or a number field, and `R` is a subfield;
case that `S` is a number field and `R` is the ring of integers in `S`,
for example `R = ZZ` and `S = QQ`)
"""
function extension_of_scalars(M::GModule, phi::Map)
  error("not yet ...")
end
```

These "new" names should finally be filled with life.

Perhaps @ThomasBreuer can look into that?

fingolfin:
Also @mjrodgers can join in

fingolfin:
@mjrodgers @ThomasBreuer could you perhaps summarize the current state of this project in a comment here?

fingolfin:
(So that we can talk about it during the triage meeting next Wednesday -- though hopefully you can both make it there?)

ThomasBreuer:
I have started collecting test examples for constructing GModule objects from scratch, without going through functions like `Oscar.RepPc.reps`.

fingolfin:
some TODOs:
- add G-module morphism type
- import work from https://github.com/oscar-system/Oscar.jl/tree/Book
- add documentation
- ...

PR #3266, opened 2024-01-29
G module missing methods

mjrodgers:
This is some work in progress to address #3224. There are a few methods that need to be implemented, these are

### Missing GModule methods
- [x] `extension_of_scalars` I wrote a simple method for this
- [ ] `can_be_defined_over`
- [ ] `can_be_defined_over_with_data`
- [x]  `descent_to_minimal_degree_field` I borrowed the contents for `descent_to_minimal_degree_field` from the preexisting `gmodule_minimal_field` methods
- [ ] `invariant_lattice_classes`

I believe that the `can_be_defined_over*` methods can be in some cases defined using `gmodule_over`, but the latter seems to only be implemented for fields, while the former seems like it should also work for the ring of integers.

I am not sure of how to implement  `invariant_lattice_classes`. 

I appreciate any feedback on what's here, advice on the remaining methods, and tips for some easy tests to write that won't take too long to run.

### DIFF
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -496,7 +496,7 @@ end
 function irreducible_modules(::QQField, G::Oscar.GAPGroup)
   #if cyclo is not minimal, this is not irreducible
   z = irreducible_modules(CyclotomicField, G)
-  return [gmodule(QQ, m) for m in z]
+  return [gmodule(QQ, descent_to_minimal_degree_field(m)) for m in z]
 end
 
 function irreducible_modules(::ZZRing, G::Oscar.GAPGroup)
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -83,7 +83,7 @@ G-module for G acting on vector space of dimension 8 over rational field
 ```
 """
 function restriction_of_scalars(M::GModule{<:Oscar.GAPGroup, <:AbstractAlgebra.FPModule{<:FieldElem}}, phi::Map)
-  #works iff relative_field above works. At least for AnticNumberField and
+  #works iff relative_field above works. At least for AbsSimpleNumField and
   #finite fields
   @assert codomain(phi) == base_ring(M)
   d = divexact(degree(codomain(phi)), degree(domain(phi)))
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -142,6 +142,25 @@ function can_be_defined_over(M::GModule, phi::Map)
   error("not yet ...")
 end
 
+function can_be_defined_over(M::GModule{<:Any, <:AbstractAlgebra.FPModule{<:FinFieldElem}}, phi::Map)
+  # Only works for irreducible modules
+  k = domain(phi)
+  K = base_ring(M)
+  d = absolute_degree(k)
+  @assert absolute_degree(K) != d
+
+  s = absolute_frobenius(K, d)
+  mkK = embed(k, K)
+  os = divexact(absolute_degree(K), d)
+  hB = hom_base(M, gmodule(M.M, Group(M),
+                      [hom(M.M, M.M, map_entries(s, matrix(x))) for x = M.ac]))
+  if length(hB) != 1
+    length(hB) > 1 && error("Module not irreducible")
+    length(hB) == 0 && return false
+  end
+  return true
+end
+
 
 """
     can_be_defined_over_with_data(M::GModule, phi::Map)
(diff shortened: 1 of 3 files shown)


experimental/GModule/GModule.jl
```diff
@@ -191,10 +196,27 @@ where `M` is a module over `R`.
...
 """
-function descent_to_minimal_degree_field(M::GModule)
-  error("not yet ...")
+function descent_to_minimal_degree_field(C::GModule{<:Any, <:AbstractAlgebra.FPModule{fpFieldElem}})
+  return C
 end
 
+function descent_to_minimal_degree_field(C::GModule{<:Any, <:AbstractAlgebra.FPModule{<:FinFieldElem}})
+  #always over char field
+  K =  base_ring(C)
```
fingolfin:
```suggestion
  K = base_ring(C)
```

experimental/GModule/GModule.jl
```diff
@@ -112,7 +112,12 @@ case that `S` is a number field and `R` is the ring of integers in `S`,
 for example `R = ZZ` and `S = QQ`)
 """
 function extension_of_scalars(M::GModule, phi::Map)
-  error("not yet ...")
+  @assert domain(phi) == base_ring(M)
+
+  d = dim(M)
+  F = free_module(codomain(phi), d)
+
+  return GModule(F, group(M), [hom(F, F, map_entries(phi, matrix(x))) for x = M.ac])
```
fingolfin: nitpick: I personally prefer using `in` instead of `=` for iteration, but that's in the end a matter of taste I guess
```suggestion
  return GModule(F, group(M), [hom(F, F, map_entries(phi, matrix(x))) for x in M.ac])
```

experimental/GModule/GModule.jl
```diff
@@ -191,10 +196,27 @@ where `M` is a module over `R`.
...
+  return C
 end
 
+function descent_to_minimal_degree_field(C::GModule{<:Any, <:AbstractAlgebra.FPModule{<:FinFieldElem}})
+  #always over char field
+  K =  base_ring(C)
+  d = 0
+  while d < degree(K)-1
+    d += 1
+    degree(K) % d == 0 || continue
```
fingolfin: would it make sense to iterate over `sort(divisors(degree(K)))` (and skip the last one) ? Ah well, probably doesn't matter for the degrees that we can actually use here.

experimental/GModule/GModule.jl
```diff
@@ -191,10 +196,27 @@ where `M` is a module over `R`.
...
+    k = GF(Int(characteristic(K)), d)
```
fingolfin: What if the characteristic of `K` does not fit into an `Int`?

Actually with the new finite fields, I think it doesn't matter anyway, you can just do
```suggestion
    k = GF(characteristic(K), d)
```
and it will decide internally which representation to use... Right, @thofma ?

fingolfin (commented):
only very minor nitpicks from me. Maybe @fieker has more suggestions?

experimental/GModule/GModule.jl
mjrodgers: I agree, this is just automatic if we do it this way.

thofma: Yes. We don't need these shenanigans anymore.

experimental/GModule/GModule.jl
fieker: In our range, factoring is MUCH more expensive than trying

fingolfin:
@mjrodgers this has a merge conflict now, please update.

For `can_be_defined_over`: if it only supports fields for now, so be it -- just implement it for that (and add some tests etc.), and throw an error in the case beyond that. A later iteration can perhaps work on more general cases.

As to `invariant_lattice_classes`: have you talked to @fieker about this? In any case, I don't see why this should hold up this PR. Better to have 2-3 of the functions done than none...

experimental/GModule/GModule.jl
```diff
@@ -112,7 +112,12 @@ case that `S` is a number field and `R` is the ring of integers in `S`,
...
 """
 function extension_of_scalars(M::GModule, phi::Map)
-  error("not yet ...")
+  @assert domain(phi) == base_ring(M)
+
+  d = dim(M)
+  F = free_module(codomain(phi), d)
+
+  return GModule(F, group(M), [hom(F, F, map_entries(phi, matrix(x))) for x in M.ac])
 end
```
fingolfin: There should be test cases for all these functions. Ideally also `jldoctest` examples in their docstrings. And the docstrings should be inserted into the manual.

mjrodgers:
> @mjrodgers this has a merge conflict now, please update.
> 
> For `can_be_defined_over`: if it only supports fields for now, so be it -- just implement it for that (and add some tests etc.), and throw an error in the case beyond that. A later iteration can perhaps work on more general cases.
> 
> As to `invariant_lattice_classes`: have you talked to @fieker about this? In any case, I don't see why this should hold up this PR. Better to have 2-3 of the functions done than none...

Sounds good, looks like Lars just updated with master.

I have a pull request just about ready implementing all of the methods (except for `invariant_lattice_classes`) for finite fields, and then I'll get to work doing the same for number fields, and adding tests.

I have talked to Claus about these things, it looks like the main things to be extended later will be
1. Extend the methods `can_be_defined_over`/etc. to the rings of integers
2. Also figure out how to deal with the above methods in the cases where the module is not irreducible
3. The `invariant_lattice_classes`.

I have some reading material to help me work out these details.

experimental/GModule/GModule.jl
```diff
@@ -191,10 +196,27 @@ where `M` is a module over `R`.
...
+    d += 1
+    degree(K) % d == 0 || continue
+    k = GF(characteristic(K), d)
+    D = gmodule_over(k, C, do_error = false)
+    D === nothing || return D
+  end
+  return C
 end
 
+function descent_to_minimal_degree_field(C::GModule{<:Any, <:AbstractAlgebra.FPModule{nf_elem}})
```
lgoettgens:
```suggestion
function descent_to_minimal_degree_field(C::GModule{<:Any, <:AbstractAlgebra.FPModule{AbsSimpleNumField}})
```
`nf_elem` is deprecated

fieker:
can_be_defined_over would also work with ZZ (if the input is over a number field/ QQ)

fingolfin:
@mjrodgers what's the status of this PR? Is there still code in there which is not yet merged and which should be preserved?

mjrodgers:
> @mjrodgers what's the status of this PR? Is there still code in there which is not yet merged and which should be preserved?

No, sorry I will close this; I have my current progress on a separate branch (that has by now superseded this pull request).

PR #3271, opened 2024-01-30
Wrap tests/doctests into `with_unicode(false)`

lgoettgens:
Needs https://github.com/Nemocas/AbstractAlgebra.jl/pull/1586.

If one has unicode enabled and runs tests/doctests/`doctest_fix` locally, this may lead to inconsistent results (as the printing may use unicode).
This PR wraps all of these functions into a block that disables unicode temporarily to get consistent output.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,7 +26,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 cohomCalg_jll = "5558cf25-a90e-53b0-b813-cadaa3ae7ade"
 
 [compat]
-AbstractAlgebra = "0.37.3"
+AbstractAlgebra = "0.37.5"
 AlgebraicSolving = "0.4.6"
 Distributed = "1.6"
 DocStringExtensions = "0.8, 0.9"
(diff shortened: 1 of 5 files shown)


benlorenz:
I think calling it in each `test_module` will lead to race-conditions when running the tests in parallel since each `allow_unicode` will try to write to the same `LocalPreferences.toml` (twice). It would be good to have some way to override the setting without storing it.

lgoettgens:
I didn't realize that `test_module` is called in `runtests.jl`, I thought this was just for interactive use

fingolfin:
Regarding @benlorenz's comment, I've opened https://github.com/Nemocas/AbstractAlgebra.jl/issues/1588

lgoettgens:
Now that https://github.com/Nemocas/AbstractAlgebra.jl/pull/1586 is available, there are no longer any race conditions between different workers as `with_unicode` now uses a global variable instead of the file.
`allow_unicode(something; temporary=false)` leads to weird effects if used in a test file (as it is called inside a `with_unicode` block in `test_module`). Thus, I adapted all of these to only change this temporarily.

benlorenz (approved):
Thanks for taking care of this, seems to work fine.

I did some local tests with several Oscar instances using the same directory, running with many workers and tests + doctests in parallel.

PR #3276, opened 2024-01-30
feat: adjust for upcoming indexing changes

thofma:
No functional changes here, but it will allow us to eventually clean up some of the indexing for matrices.

### DIFF
--- a/experimental/FTheoryTools/src/LiteratureModels/constructors.jl
+++ b/experimental/FTheoryTools/src/LiteratureModels/constructors.jl
@@ -259,7 +259,7 @@ function _construct_literature_model_over_concrete_base(model_dict::Dict{String,
 
   # Find divisor classes of the internal model sections
   auxiliary_base_grading = matrix(ZZ, transpose(hcat([[eval_poly(weight, ZZ) for weight in vec] for vec in model_dict["model_data"]["auxiliary_base_grading"]]...)))
-  auxiliary_base_grading = vcat([[Int(k) for k in auxiliary_base_grading[i,:]] for i in 1:nrows(auxiliary_base_grading)]...)
+  auxiliary_base_grading = vcat([[Int(k) for k in auxiliary_base_grading[i:i,:]] for i in 1:nrows(auxiliary_base_grading)]...)
   internal_model_sections = Dict{String, ToricDivisor}()
   for k in 2+length(model_sections):ngens(auxiliary_base_ring)
     divisor = sum([auxiliary_base_grading[l,k] * model_sections_divisor_list[l] for l in 1:nrows(auxiliary_base_grading)])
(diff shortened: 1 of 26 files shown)


experimental/FTheoryTools/src/LiteratureModels/constructors.jl
```diff
@@ -259,7 +259,7 @@ function _construct_literature_model_over_concrete_base(model_dict::Dict{String,
 
   # Find divisor classes of the internal model sections
   auxiliary_base_grading = matrix(ZZ, transpose(hcat([[eval_poly(weight, ZZ) for weight in vec] for vec in model_dict["model_data"]["auxiliary_base_grading"]]...)))
-  auxiliary_base_grading = vcat([[Int(k) for k in auxiliary_base_grading[i,:]] for i in 1:nrows(auxiliary_base_grading)]...)
+  auxiliary_base_grading = vcat([[Int(k) for k in auxiliary_base_grading[i:i,:]] for i in 1:nrows(auxiliary_base_grading)]...)
```
fingolfin, line 262: Not an objection, just trying to understand: presumably the old and new code will behave differently in the future? What will the difference be?

For Julia matrices, the difference is whether a vector or a matrix is returned. But we don't have dedicated vector objects, have we? Or is the plan to add them? (Ah, perhaps for a sparse matrix we'd be returning `SRow` versus `SMat` ?)

thofma: Yes. It will just be a `Vector` (see the Linear algebra roadmap), so that the dimension of the indices now correspond to the dimension to the output (as for julia arrays).

1. Nowadays we have more support for `Matrix * Vector` and `solve(::Matrix, ::Vector)` and things like that, which we did not have 5 years ago. So back then we avoided `Vector` at all costs.
2. At the moment, it is cumbersome to get rows/columns as `Vector`. With the "new" indexing, it is easy: Just do `[i, :]`. To get it as a matrix, just do `[[i], :]` or `[i:i, :]`.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
```diff
@@ -228,21 +228,21 @@ function normal_toric_variety_from_star_triangulation(P::Polyhedron)
...
-  indices = findall(k -> pts[k,:] == matrix(ZZ, [zero]), 1:nrows(pts))
+  indices = findall(k -> pts[k:k,:] == matrix(ZZ, [zero]), 1:nrows(pts))
   @req length(indices) == 1 "Polyhedron must contain origin (exactly once)"
 
   # Change order of lattice points s.t. zero is the first point
-  tmp = pts[1,:]
-  pts[1,:] = pts[indices[1],:]
-  pts[indices[1],:] = tmp
+  tmp = pts[1:1,:]
+  pts[1:1,:] = pts[indices[1],:]
```
lgoettgens:
```suggestion
  pts[1:1,:] = pts[indices[1]:indices[1],:]
```

lgoettgens (approved):
I noticed one place that probably needs to get adjusted as well. Apart from that, fine with me

joschmitt (approved):

PR #3280, opened 2024-01-31
CI: adjust test subgroups, try macos aarch64 runner for one subgroup and doctests

benlorenz:


benlorenz:
The macos-14 label is the new M1 runner: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
The tests seem to work and run a bit faster, which is good for our macos tests.

fingolfin (approved):

fingolfin:
Great

lgoettgens:
Do we care that we now don't test on maxOS x86?

benlorenz:
> Do we care that we now don't test on maxOS x86?

I kept the testgroup for the larger testsets `long` on the `macos-latest` label which runs on `x86_64`. So we do have both, but not every test on both architectures.
Another reason for this choice is that the new M1 runners only have 7GB of RAM which might become a problem for some of the larger testsets (the `x86_64` runners have 14GB).

PR #3283, opened 2024-01-31
Prepare for `algebraic_closure(::QQField)`

lgoettgens:
Once this is merged, adding the above method to Nemo won't break Oscar anymore


fingolfin (approved):
Looks good to me. Of course we can remove the `if isdefined` bit once a Nemo with this out and we adjusted the compat bound in `Project.toml` to require at least that new version.

lgoettgens:
> Looks good to me. Of course we can remove the `if isdefined` bit once a Nemo with this out and we adjusted the compat bound in `Project.toml` to require at least that new version.

That was my intention

PR #3289, opened 2024-01-31
Say 'Polytope' in show if it is known that a Polyhedron is bounded

micjoswig:
Per request by @JohnAAbbott 

Thanks for the suggestion!


fingolfin (approved):
Nice!

Issue #3265, opened 2024-01-29
Rename `conjugacy_classes_maximal_subgroups ` and `maximal_subgroup_classes ` or even just `*subgroups`


fingolfin:
We discussed it before: `conjugacy_classes_maximal_subgroups ` is hard to find via TAB completion, which we could solve by calling it `maximal_subgroup_classes`.

Likewise for `conjugacy_classes_subgroups` -> `subgroup_classes`.

The old names could be kept but deprecated.

----

If we have this, is there even need for `subgroup_reps`? It would be equivalent to `map(representative, subgroup_classes(g)`

Similar for `maximal_subgroups`.

----

Going one step further (we also discussed this before...), perhaps we should indeed just call these `subgroups` and `maximal_subgroups`? Anyone who *really* wants *all* (maximal) subgroups has to do a  little bit more work then, but otherwise it would steer people into the right direction (in many (most?) applications it is better to work with conjugacy classes). I think this is what Magma does.

IMHO this would also remove the need for `subgroup_reps` / `maximal_subgroup_reps`

----

In addition, I think also all other `*subgroup_reps` functions should be adjusted to provide similar features / variants as `subgroups` and `maximal_subgroups`
- `hall_subgroup_reps`:  either add `hall_subgroup_classes` and `hall_subgroups` (returning a vector of groups), or just add `hall_subgroup` and let it return a vector of conjugacy class
- `low_index_subgroup_reps`: ditto
- `complement_class_reps`: ditto

ThomasBreuer:
The following can be done in a non-breaking way:

- `hall_subgroup_reps`: deprecate, instead provide `hall_subgroups`, a vector of conjugacy classes of subgroups
- `low_index_subgroup_reps`: deprecate, instead provide `low_index_subgroups`, a vector of conjugacy classes of subgroups
- `complement_class_reps`: deprecate, instead provide `complements`, a vector of conjugacy classes of subgroups

The following changes would be breaking:

- concerning maximal subgroups:
  deprecate `maximal_subgroup_reps` and `conjugacy_classes_maximal_subgroups`
  change the meaning of `maximal_subgroups` to return a vector of conjugacy classes of the maximal subgroups,
  change the code where `maximal_subgroups` is used to get a vector of all maximal subgroups

- concerning all subgroups:
[...]

PR #3291, opened 2024-01-31
rename and change some `*_reps` functions

ThomasBreuer:
This is the non-controversial and non-breaking part of #3265:

Provide `complement_classes` instead of `complement_class_reps`,
`hall_subgroups` instead of `hall_subgroup_reps`,
`low_index_subgroups` instead of `low_index_subgroups_reps`.
Each of the new functions returns a vector of subgroup classes.

Also move the outdated `hall_subgroup` to `deprecations.jl`.

### DIFF
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -91,8 +91,8 @@ conjugacy_class(G::GAPGroup, g::GAPGroupElem)
 conjugacy_class(G::T, g::T) where T<:GAPGroup
 conjugacy_classes(G::GAPGroup)
 complement_classes
-hall_subgroups
-maximal_subgroups(G::GAPGroup)
+hall_subgroup_classes
+maximal_subgroup_classes(G::GAPGroup)
 subgroup_classes(G::GAPGroup)
 ```
 
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -53,7 +53,7 @@ derived_series
 sylow_system
 hall_subgroups
 hall_system
-complement_class_reps
+complement_classes
 complement_system
 chief_series
 composition_series
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -43,16 +43,13 @@ intersect(::T, V::T...) where T<:GAPGroup
 The following functions return a vector of subgroups.
 
 ```@docs
-subgroups(G::GAPGroup)
 normal_subgroups
 maximal_normal_subgroups
 minimal_normal_subgroups
 characteristic_subgroups
 derived_series
 sylow_system
-hall_subgroups
 hall_system
-complement_classes
 complement_system
 chief_series
 composition_series
@@ -93,8 +90,10 @@ number_of_conjugacy_classes(G::GAPGroup)
 conjugacy_class(G::GAPGroup, g::GAPGroupElem)
 conjugacy_class(G::T, g::T) where T<:GAPGroup
 conjugacy_classes(G::GAPGroup)
-conjugacy_classes_subgroups(G::GAPGroup)
+complement_classes
+hall_subgroups
 maximal_subgroups(G::GAPGroup)
+subgroup_classes(G::GAPGroup)
 ```
 
 
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -104,6 +104,7 @@ conjugacy_class(G::T, g::T) where T<:GAPGroup
 conjugacy_classes(G::GAPGroup)
 complement_classes
 hall_subgroup_classes
+low_index_subgroup_classes
 maximal_subgroup_classes(G::GAPGroup)
 subgroup_classes(G::GAPGroup)
 ```
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -67,6 +67,18 @@ upper_central_series
     one can type `embedding(G,H)`.
 
 
+The following functions return an iterator of subgroups.
+Usually it is more efficient to work with (representatives of) the
+underlying conjugacy classes of subgroups instead.
+
+```@docs
+complements(G::T, N::T) where T <: GAPGroup
+hall_subgroups
+low_index_subgroups
+maximal_subgroups
+subgroups(G::GAPGroup)
+```
+
 ## Conjugation action of elements and subgroups
 
 ```@docs
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -45,7 +45,6 @@ The following functions return a vector of subgroups.
 ```@docs
 subgroups(G::GAPGroup)
 normal_subgroups
-maximal_subgroups
 maximal_normal_subgroups
 minimal_normal_subgroups
 characteristic_subgroups
@@ -95,7 +94,7 @@ conjugacy_class(G::GAPGroup, g::GAPGroupElem)
 conjugacy_class(G::T, g::T) where T<:GAPGroup
 conjugacy_classes(G::GAPGroup)
 conjugacy_classes_subgroups(G::GAPGroup)
-conjugacy_classes_maximal_subgroups(G::GAPGroup)
+maximal_subgroups(G::GAPGroup)
 ```
 
 
--- a/docs/src/Groups/subgroups.md
+++ b/docs/src/Groups/subgroups.md
@@ -51,7 +51,7 @@ minimal_normal_subgroups
 characteristic_subgroups
 derived_series
 sylow_system
-hall_subgroup_reps
+hall_subgroups
 hall_system
 complement_class_reps
 complement_system
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -789,6 +789,23 @@ function subgroup_classes(G::GAPGroup; order::T = ZZRingElem(-1)) where T <: Int
 end
 
 """
+    subgroups(G::GAPGroup)
+
+Return an iterator of all subgroups in `G`.
+Very likely it is better to use [`subgroup_classes`](@ref) instead.
+
+# Examples
+```jldoctest
+julia> println([order(H) for H in subgroups(symmetric_group(3))])
+ZZRingElem[1, 2, 2, 2, 3, 6]
+
+julia> println([order(H) for H in subgroups(quaternion_group(8))])
+ZZRingElem[1, 2, 4, 4, 4, 8]
+```
+"""
+subgroups(G::GAPGroup) = Iterators.flatten(subgroup_classes(G))
+
+"""
     maximal_subgroup_classes(G::Group)
 
 Return the vector of all conjugacy classes of maximal subgroups of G.
@@ -811,6 +828,23 @@ julia> maximal_subgroup_classes(G)
 end
 
 """
+    maximal_subgroups(G::Group)
+
+Return an iterator of the maximal subgroups in `G`.
+Very likely it is better to use [`maximal_subgroup_classes`](@ref) instead.
+
+# Examples
+```jldoctest
+julia> println([order(H) for H in maximal_subgroups(symmetric_group(3))])
+ZZRingElem[3, 2, 2, 2]
+
+julia> println([order(H) for H in maximal_subgroups(quaternion_group(8))])
+ZZRingElem[4, 4, 4]
+```
+"""
+maximal_subgroups(G::T) where T <: Union{GAPGroup, FinGenAbGroup} = Iterators.flatten(maximal_subgroup_classes(G))
+
+"""
     low_index_subgroup_classes(G::GAPGroup, n::Int)
 
 Return a vector of conjugacy classes of subgroups of index at most `n` in `G`.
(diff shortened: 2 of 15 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -1341,19 +1334,24 @@ together with `N` generates `G`.
...
 
-julia> complement_class_reps(G, center(G)[1])
-PcGroup[]
+julia> complement_classes(G, center(G)[1])
+GAPGroupConjClass{PcGroup, PcGroup}[]
 ```
 """
-function complement_class_reps(G::T, N::T) where T <: GAPGroup
-   return _as_subgroups(G, GAP.Globals.ComplementClassesRepresentatives(G.X, N.X))
+function complement_classes(G::T, N::T) where T <: GAPGroup
```
fingolfin, line 1365: This is one of those signatures that needs to be relaxed when we introduce `SubPcGroupElem`, I guess...

fingolfin:
But if we do this then these functions behave inconsistent with `maximal_subgroups` and `subgroups`. So it is not clear to me that this is a definite "win" -- if we decide not to change those two, then I don't think we should make the changes here?

I'll discuss this some more with @fieker later.

ThomasBreuer:
@fingolfin Thanks for your comment.
From my point of view, a breaking change for `maximal_subgroups*` is not a problem.
For `subgroups`, we could proceed as proposed in #3265 (return conjugacy classes of subgroups) *AND* use a different name for the `subgroups` functionality that comes from Hecke.jl.

ThomasBreuer:
After the discussion of #3265, what about the names `complement_classes`, `hall_subgroup_classes`, `low_index_subgroup_classes`, `maximal_subgroup_classes`, and `subgroup_classes`?
Then the names are self-explanatory (I think), all changes are non-breaking, and also Hecke.jl's `subgroups` can be kept as is.

(If we take this solution then only `subgroups`/`subgroup_reps`/`conjugacy_classes_subgroups` needs further code changes, the rest is just string replacement.)

fingolfin:
That would be fine by me. 

As a variation (possibly for a future PR, `subgroups` and friends could return an iterator instead of a vector.... eg  `subgroups (G::GAPGroups) = Iterators.flatten(subgroup_classes(G))`

src/Groups/GAPGroups.jl
ThomasBreuer: Yes, and here we have also an example which is subtle in the sense that the result may be empty.
In this case, its type is that of an empty array of conjugacy classes of groups, where the type of each conjugacy class is parametrized by the types of the acting group and the representative. In the case of pc groups, the representatives will be of type `SubPcGroup`, and the acting group can be either a `PcGroup` or a `SubPcGroup`.

docs/src/Groups/subgroups.md
```diff
@@ -43,17 +43,13 @@ intersect(::T, V::T...) where T<:GAPGroup
 The following functions return a vector of subgroups.
 
 ```@docs
-subgroups(G::GAPGroup)
```
fingolfin, line 46: (I thought I'd already left this as a comment, but now it's not here -- maybe I once again forgot to press "submit"? Or maybe I left this as a comment elsewhere (and you perhaps even answered it there) -- sorry in that case, and please point me to it.)

Why is this dropping `subgroups` altogether? Isn't the plan to provide both `*subgroups` and `*subgroup_classes` commands in parallel for all four commands?

ThomasBreuer: Currently the proposed changes deprecate `maximal_subgroups`; there were no `complements` and `hall_subgroups` before (just a very old, already deprecated `hall_subgroup`); and `subgroups` will remain for the methods that come from Hecke (not for `GAPGroup`s), with a different behaviour which we did not want to touch.

fingolfin: But won't people still find `subgroups` and try to use it in our groups and be confused about the `method not found` error? 

So either this is renamed in Hecke (CC @thofma @fieker) or we add a `subgroups` method (which then either produces an error and tells the user to use `subgroup_classes`; or, well, it return an iterator or vector over all subgroups).

I do think that for teaching and general "easer of use" the latter (adding a proper `subgroups` method) would be best. The current approach in this PR is good for pushing people to do "the right thing" (but many people react allergic to being pushed to do the right thing...).

My instincts tell me that having something like
```julia
subgroups (G::GAPGroups) = Iterators.flatten(subgroup_classes(G))
```
or
```julia
subgroups (G::GAPGroups) = collect(Iterators.flatten(subgroup_classes(G)))
```
(resp. an optimized version of the latter) would be better. What do @thofma and @fieker think about this?

thofma: Yes, I think `subgroups` should exist for the casual user and it should return subgroups. Not sure if the `Iterators.flatten(subgroup_classes(G))` does this.

ThomasBreuer: O.k., we define and document `subgroups` for a `GAPGroup`, and the result is an iterator of subgroups (without embeddings), analogous to `subgroups(abelian_group([2,3,4]))`, which returns a `Hecke.SubgroupIterator`.
The docstring will contain a hint that probably calling `subgroup_classes` would be better.

Should `complements`, `hall_subgroups`, `low_index_subgroups`, `maximal_subgroups` be treated analogously?

fingolfin: Yes

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
...
+julia> subgroup_classes(G, order = ZZRingElem(2))
+1-element Vector{GAPGroupConjClass{PermGroup, PermGroup}}:
+ Conjugacy class of permutation group in G
 ```
 """
-function conjugacy_classes_subgroups(G::GAPGroup)
+function subgroup_classes(G::GAPGroup; order::T = ZZRingElem(-1)) where T <: IntegerUnion
   L = Vector{GapObj}(GAPWrap.ConjugacyClassesSubgroups(G.X))
+  if order != -1
+    L = [x for x in L if GAPWrap.Order(GAPWrap.Representative(x)) == order]
```
fingolfin: Does the `==` comparison work when `order` is a `ZZRingElem` and the group is a `GapObj` (and not an `Int`)? I would have thought not (it might be hard to add a test case for that, though... perhaps if `G` is a huge group of prime order?)

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
...
+  end
   return [GAPGroupConjClass(G, _as_subgroup_bare(G, GAPWrap.Representative(cc)), cc) for cc in L]
```
fingolfin: Maybe simpler & safer (except that `order` shadows `order`, so this won't work, but you get the idea..):
```suggestion
  res = [GAPGroupConjClass(G, _as_subgroup_bare(G, GAPWrap.Representative(cc)), cc) for cc in L]
  if order != -1
    filter!(x -> order(x) == order, res)
  end
  return res
```

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
...
 end
 
 """
-    subgroup_reps(G::GAPGroup; order::ZZRingElem = ZZRingElem(-1))
+    subgroups(G::GAPGroup)
 
-Return a vector of representatives (under conjugation) for all subgroups of `G`.
-If given, only subgroups of a certain order are returned.
+Return an iterator of all subgroups in `G`.
```
```suggestion
Return an iterator over all subgroups in `G`.
```

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
...
-  end
-  return [Oscar._as_subgroup(G, x)[1] for x = C]
-end
+subgroups(G::GAPGroup) = Iterators.flatten(subgroup_classes(G))
 
 """
-    conjugacy_classes_maximal_subgroups(G::Group)
+    maximal_subgroup_classes(G::Group)
 
 Return the vector of all conjugacy classes of maximal subgroups of G.
```
```suggestion
Return the vector of all conjugacy classes of maximal subgroups of `G`.
```
or
```suggestion
Return a vector of all conjugacy classes of maximal subgroups of `G`.
```

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
 end
 
 """
-    conjugacy_classes_subgroups(G::Group)
+    subgroup_classes(G::GAPGroup; order::T = ZZRingElem(-1)) where T <: IntegerUnion
 
-Return the vector of all conjugacy classes of subgroups of G.
+Return the vector of all conjugacy classes of subgroups of G or,
```
```suggestion
Return the vector of all conjugacy classes of subgroups of `G` or,
```
Should it really be "the vector", not "a vector"? (Of course that's beyond the scope of this PR, but while we are at it...

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
...
+  return Vector{GAPGroupConjClass{T, T}}(LL)
 end
 
 """
-    maximal_subgroup_reps(G::GAPGroup)
+    maximal_subgroups(G::Group)
 
-Return a vector of representatives (under conjugation) for all maximal
-subgroups of `G`.
+Return an iterator of the maximal subgroups in `G`.
```
```suggestion
Return an iterator over the maximal subgroups in `G`.
```

src/Groups/GAPGroups.jl
```diff
@@ -756,123 +758,130 @@ function Base.rand(rng::Random.AbstractRNG, C::GroupConjClass{S,T}) where S wher
...
+function low_index_subgroup_classes(G::GAPGroup, n::Int)
+  @req (n > 0) "index must be positive"
+  ll = GAP.Globals.LowIndexSubgroups(G.X, n)::GapObj
+  return [conjugacy_class(G, H) for H in _as_subgroups(G, ll)]
 end
 
+"""
+    low_index_subgroups(G::Group, n::Int)
+
+Return an iterator of the subgroups of index at most `n` in `G`.
```
```suggestion
Return an iterator over the subgroups of index at most `n` in `G`.
```

src/Groups/GAPGroups.jl
```diff
@@ -1272,37 +1273,53 @@ up to conjugacy.
...
    else
-     return [_as_subgroup_bare(G, res_gap)]
+     return [conjugacy_class(G, _as_subgroup_bare(G, res_gap))]
    end
 end
 
+"""
+    hall_subgroups(G::Group, P::AbstractVector{<:IntegerUnion})
+
+Return an iterator of the Hall `P`-subgroups in `G`.
```
```suggestion
Return an iterator over the Hall `P`-subgroups in `G`.
```

src/Groups/sub.jl
```diff
@@ -295,14 +239,11 @@ i.e., those subgroups that are invariant under all automorphisms of `G`.
 
 # Examples
 ```jldoctest
-julia> subgroups(symmetric_group(3))
-6-element Vector{PermGroup}:
- Permutation group of degree 3 and order 1
- Permutation group of degree 3 and order 2
- Permutation group of degree 3 and order 2
- Permutation group of degree 3 and order 2
+julia> characteristic_subgroups(symmetric_group(3))
```
fingolfin, line 242: Ouch, we had a bad example... good catch!

src/NumberTheory/GaloisGrp/GaloisGrp.jl
```diff
@@ -894,7 +894,7 @@ function set_orbit(G::PermGroup, H::PermGroup)
   #    http://dblp.uni-trier.de/db/journals/jsc/jsc79.html#Elsenhans17
   # https://doi.org/10.1016/j.jsc.2016.02.005
 
-  l = low_index_subgroup_reps(H, 2*degree(G)^2)
+  l = map(representative, low_index_subgroup_classes(H, 2*degree(G)^2))
```
fingolfin: Could also use broadcasting (I am fine with either variant)
```suggestion
  l = representative.(low_index_subgroup_classes(H, 2*degree(G)^2))
```

src/NumberTheory/GaloisGrp/Group.jl
```diff
@@ -7,7 +7,7 @@ minimal supergroups, ie. it fails in ``C_4``.
 function maximal_subgroup_chain(G::PermGroup, U::PermGroup)
   l = [G]
   while order(l[end]) > order(U)
-    m = maximal_subgroups(l[end])
+    m = reduce(vcat, map(collect, maximal_subgroup_classes(l[end])))
```
fingolfin: Should this use `maximal_subgroups` again? Possibly combined with `collect`:
```suggestion
    m = collect(maximal_subgroups(l[end]))
```

src/Groups/GAPGroups.jl
ThomasBreuer: Right, this comparison is not safe, the `GAP.Obj` on the left side should better be converted to the type `T`.
Concerning  a test case: In principle this would work, but there is a bug on the GAP side, see gap-system/gap/issues/5638.

fingolfin:
Added "backport to 1.0" label as I need this for the book.

test/Groups/subgroups_and_cosets.jl
```diff
@@ -388,7 +389,7 @@ end
    @test order(fitting_subgroup(G)[1])==8
    @test fitting_subgroup(S)==sub(S,[S([3,4,1,2]), S([4,3,2,1])])
    @test frattini_subgroup(S)==sub(S,[one(S)])
-   @test frattini_subgroup(G)[1]==intersect(maximal_subgroups(G))[1]
+   @test frattini_subgroup(G)[1]==intersect(collect(Iterators.flatten(maximal_subgroup_classes(G))))[1]
```
fingolfin:
```suggestion
   @test frattini_subgroup(G)[1]==intersect(collect(maximal_subgroups(G)))[1]
```

test/Groups/subgroups_and_cosets.jl
```diff
@@ -35,7 +35,7 @@
 
    G = symmetric_group(4)
    A = alternating_group(4)
-   L = subgroups(G)
+   L = collect(Iterators.flatten(subgroup_classes(G)))
```
fingolfin:
```suggestion
   L = collect(subgroups(G))
```

fingolfin (approved):

PR #3294, opened 2024-01-31
add support to adjoin many elements

fieker:
possibly more draft...

allows to do
rt = [a,b,c] of type Vector{qqbar}
QQ[rt]
should help to transposrt matrices...


experimental/GModule/Misc.jl
```diff
@@ -34,6 +68,7 @@ function Hecke.number_field(::QQField, a::qqbar; cached::Bool = false)
 end
 
 Base.getindex(::QQField, a::qqbar) = number_field(QQ, a)
+Base.getindex(::QQField, a::Vector{qqbar}) = number_field(QQ, a)
```
fingolfin: maybe also
```suggestion
Base.getindex(::QQField, a::Vector{qqbar}) = number_field(QQ, a)
Base.getindex(::QQField, a::qbar, as::qqbar...) = number_field(QQ, [a, as...])
```
for interactive use: `QQ[s,c]`

fingolfin (approved):

PR #3295, opened 2024-01-31
Remove two temporary files

lgoettgens:
They are identical (up to whitespace) to the file with the filename without `~` in the end.


fingolfin (approved):

fingolfin:
We could add `*~` to our `.gitignore` (ideally people already do that in their personal git ignore config, but clearly we can't rely on it, and I see no downside)

antonydellavecchia (approved):
Whoops. Can you also add a *~ to the gitignore to guarantee this doesn't happen again?

PR #3298, opened 2024-02-01
Fixes for the degree function

HechtiDerLachs:
This addresses #2999 .

The PR contains the improved changes from #3238, rebased on #3237. 

It can be rebased again, once #3237 is merged. 

For the last testset in `test/Modules/ModulesGraded.jl` I now get the following timings:
```
julia> @time Oscar._alt_simplify(M); # M is Omega without grading.
  0.490810 seconds (2.12 M allocations: 113.333 MiB)

julia> @time Oscar._alt_simplify(Omega);
  0.667269 seconds (3.38 M allocations: 151.342 MiB)
```
so the gradings still seem to make things slower. But computations now at least finish in comparable time.


HechtiDerLachs:
Tests still fail. In particular in `src/Modules/ModulesGraded.jl` around line 655 (double dual of a resolution). 

@jankoboehm : I do not understand what's going wrong there. Something seems to produce non-homogeneous elements. Could you please have a look?

HechtiDerLachs:
Another issue is that we now make the implementation of `simplify` mandatory for any ring which wants to use graded modules. It must produce a homogeneous representative for a `SubquoModuleElem` of a graded `SubquoModule`. That has been suggested and requested by @jankoboehm in yesterday's discussion. Now the tests fail for graded modules over quotient rings and friends. 

@jankoboehm : Could you also have a look there and let me know how you think this should/can be resolved?

jankoboehm:
As long as you have a ring where you can compute with Gröbner bases, you can use division to get a homogeneous represenative of a homogeneous element. So simplify can and has to be implemented as part of the Gröbner bases abstraction layer. However if you don't have a ring with a Gröbner basis theory, then you could rely on only inputing homogeneous elements and implement simplify as the identity.

docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/free_modules.md
```diff
@@ -174,7 +174,7 @@ is_homogeneous(f::FreeModElem)
 ```
 
 ```@docs
-degree(f::FreeModElem)
+    degree(f::FreeModElem{T}) where {T<:Union{<:MPolyDecRingElem, <:MPolyQuoRingElem{<:MPolyDecRingElem}}}
```
lgoettgens:
```suggestion
degree(f::FreeModElem{T}) where {T<:Union{<:MPolyDecRingElem, <:MPolyQuoRingElem{<:MPolyDecRingElem}}}
```

HechtiDerLachs:
Current timings: 

The whole test suite for modules: 
 `196.094920 seconds (380.70 M allocations: 54.324 GiB, 6.96% gc time, 9.94% compilation time)`
 `177.237958 seconds (380.68 M allocations: 54.324 GiB, 7.13% gc time, 8.88% compilation time)`
Only the graded modules: 
 `70.670991 seconds (214.69 M allocations: 5.748 GiB, 6.36% gc time, 9.00% compilation time)`
 `60.381474 seconds (214.22 M allocations: 5.740 GiB, 6.13% gc time, 9.62% compilation time)`


Current master:

The whole test suite for modules: 
 `211.794955 seconds (814.08 M allocations: 29.130 GiB, 8.75% gc time, 9.24% compilation time)`
Only the graded modules: 
 `162.484008 seconds (701.68 M allocations: 25.221 GiB, 9.07% gc time, 3.93% compilation time)`
 `172.939563 seconds (701.96 M allocations: 25.228 GiB, 8.75% gc time, 3.97% compilation time)`

So for the graded modules time and memory consumption has decreased a lot. But at the same time the rest of the module test code seems to have degraded significantly. At this point I am clueless as to what might be the cause for that.

HechtiDerLachs:
> At this point I am clueless as to what might be the cause for that.

After some more analysis it seems to me now that the tests running longer with the new `hom` is due to some randomization artifact. If I run the whole testsuite I get the following timings:
```
...
Test Summary:       | Pass  Total  Time
tensoring morphisms |   40     40  0.5s
Test Summary:  | Pass  Total     Time
direct product |   50     50  1m13.8s
Test Summary:      | Pass  Total  Time
Coordinates (lift) |   64     64  1.2s
Test Summary:        | Pass  Total  Time
module homomorphisms |   46     46  1.0s
...
```
These timings repeat, i.e. `direct products` consistently runs for more than a minute. But when I run the tests for `direct product` by themselves, I can not reproduce that! It finishes within 4 - 10 seconds! 

If, on the other hand, I run the whole testsuite with the old version of `hom` enabled (which is still buggy in conjunction with the new `degree` function!), then I get these timings: 
```
...
Hom module    |  152    152  3.6s
Test Summary:       | Pass  Total  Time
tensoring morphisms |   39     39  0.5s
Test Summary:  | Pass  Total  Time
direct product |   50     50  0.9s
Test Summary:      | Pass  Total  Time
Coordinates (lift) |   64     64  1.3s
Test Summary:        | Pass  Total  Time
module homomorphisms |   46     46  1.0s
...
```
Even better, but given the randomization, probably not really comparable? 

@benlorenz : You put some effort into reproducibility of random tests. Do you think it's worthwhile to also have some way to call only tests which have no randomization whatsoever? I suspect that the random seed gets shifted here because of some earlier tests triggering a different number of randomization requests, so comparability is probably not really assured. 

@jankoboehm : Did you already follow the current guidelines for randomization of tests? Is there a way to select only non-randomized tests for comparison?

HechtiDerLachs:
My presumption seems to be confirmed by the following findings. I disabled all tests in which I found `randpoly` in `UngradedModules.jl`. For the remaining ones, I get the following timings:

UngradedModules.jl without randomization on new hom:
```
julia> @time Oscar.test_module("test/Modules/UngradedModules.jl", new=false)
Test Summary:         | Pass  Total  Time
Modules: Constructors |   16     16  0.0s
Test Summary:                              | Pass  Total  Time
Modules: Simplify elements of subquotients |    1      1  0.0s
Test Summary:           | Pass  Total  Time
Intersection of modules |   11     11  0.0s
Test Summary: | Pass  Total   Time
Presentation  |  163    163  12.0s
Test Summary:  | Pass  Total  Time
Prune With Map |    7      7  0.0s
Test Summary: | Pass  Total  Time
Ext, Tor      |   13     13  0.1s
Test Summary: | Pass  Total  Time
Gröbner bases |    9      9  0.3s
Test Summary:      | Pass  Total  Time
Coordinates (lift) |   64     64  1.2s
Test Summary:        | Pass  Total  Time
module homomorphisms |   46     46  1.0s
Test Summary:        | Pass  Total  Time
change of base rings |    4      4  0.0s
Test Summary: | Pass  Total  Time
duals         |   10     10  0.2s
Test Summary:                           | Pass  Total  Time
free resolution in case of no relations |    1      1  0.0s
Test Summary:             | Pass  Total  Time
length of free resolution |    1      1  0.1s
Test Summary:                                 | Pass  Total  Time
vector_space_dimension and vector_space_basis |    4      4  0.0s
Test Summary:                         | Pass  Broken  Total  Time
canonical maps and garbage collection |    9       1     10  7.2s
Test Summary: |Time
issue 3107    | None  0.2s
 28.876523 seconds (62.54 M allocations: 1.831 GiB, 24.04% gc time, 26.02% compilation time)
```
UngradedModules.jl without randomization on old hom:
```
julia> @time Oscar.test_module("test/Modules/UngradedModules.jl", new=false)
Test Summary:         | Pass  Total  Time
Modules: Constructors |   16     16  0.0s
Test Summary:                              | Pass  Total  Time
Modules: Simplify elements of subquotients |    1      1  0.0s
Test Summary:           | Pass  Total  Time
Intersection of modules |   11     11  0.0s
Test Summary: | Pass  Total   Time
Presentation  |  189    189  11.1s
Test Summary:  | Pass  Total  Time
Prune With Map |    7      7  0.0s
Test Summary: | Pass  Total  Time
Ext, Tor      |   13     13  0.2s
Test Summary: | Pass  Total  Time
Gröbner bases |    9      9  0.3s
Test Summary:      | Pass  Total  Time
Coordinates (lift) |   64     64  1.1s
Test Summary:        | Pass  Total  Time
module homomorphisms |   46     46  0.9s
Test Summary:        | Pass  Total  Time
change of base rings |    4      4  0.0s
Test Summary: | Pass  Total  Time
duals         |   10     10  1.9s
Test Summary:                           | Pass  Total  Time
free resolution in case of no relations |    1      1  0.0s
Test Summary:             | Pass  Total  Time
length of free resolution |    1      1  0.1s
Test Summary:                                 | Pass  Total  Time
vector_space_dimension and vector_space_basis |    4      4  0.0s
Test Summary:                         | Pass  Broken  Total  Time
canonical maps and garbage collection |    9       1     10  5.9s
Test Summary: |Time
issue 3107    | None  0.2s
 28.178922 seconds (65.98 M allocations: 2.131 GiB, 23.25% gc time, 25.50% compilation time)
```

So regarding performance, I think it is now rather safe to say that my new implementation does not change things for the worse. Plus it works with the new `degree` while the old one doesn't. What might need to be adjusted is the printing of elements in `hom` modules (see the failing doctests). But this can be done. 

@jankoboehm : Could you either 1) debug the old hom-function so that it runs with the new `degree` or 2) document how elements of hom-modules should print so that this can be adjusted for the new one?

HechtiDerLachs:
I have another local commit which removes the deprecated code. But before I push that, please have a look and confirm what's there right now (with old code still present for comparison). 

Tests are passing and this PR replaces #3237 , #3238 , and should also address the issue behind #2234.

Once approved, I will then push the cleanup and we can merge.

fieker:
it will be reviewed and redsvolved by ash-wednesday. (@jankoboehm , @HechtiDerLachs )

HechtiDerLachs:
> it will be reviewed and redsvolved by ash-wednesday.

Just a friendly reminder that if this is supposed to be resolved **by** ash-Wednesday, I need the input about required changes **beforehand**! Wednesday afternoon itself will only provide us with the time to run the tests.

HechtiDerLachs:
The problem with making the old `hom` run was the following. At some point one takes preimages of elements through the augmentation map of a presentation, i.e. a map of type `FreeModuleHom` with a `SubquoModule` in it's codomain. This is done via choice of representatives. This representative might be non-zero, even though the element is. If the map is graded, this leads to faulty degree inference. So the solution was to adjust the preimage function for `FreeModuleHom`s to first `simplify` the element and only afterwards take its representative. 

This made it necessary to introduce dummy `simplify` functions for `FreeModuleElem`s so that the code would run generically. 

We will need to fix the doctests. But other than that the tests for the modules run now.

docs/src/CommutativeAlgebra/ModulesOverMultivariateRings/module_operations.md
```diff
@@ -37,7 +37,7 @@ direct_product(M::ModuleFP{T}...; task::Symbol = :prod) where T
 ## Truncation
 
 ```@docs
-truncate(M::ModuleFP, g::FinGenAbGroupElem, task::Symbol = :with_morphism)
+truncate(M::ModuleFP, g::FinGenAbGroupElem, task::Symbol=:with_morphism)
 ```
```
fieker, line 42: Thr generic idea is/was to abolish `task` everywhere....

HechtiDerLachs: That's what I said. See #3237.

HechtiDerLachs:
Current timings from the example in the beginning:
```
@time Oscar._alt_simplify(M);
  0.146291 seconds (1.53 M allocations: 70.597 MiB)
```
vs.
```
@time Oscar.prune_with_map(M);
  0.114204 seconds (1.03 M allocations: 33.176 MiB)
```
For graded modules we get:
```
julia> @time Oscar._alt_simplify(Omega);
  0.186582 seconds (2.21 M allocations: 91.807 MiB)

julia> @time Oscar.prune_with_map(Omega);
  0.157183 seconds (1.87 M allocations: 57.952 MiB)
```
Memory consumption of my implementation is double and I don't know why. But timings are now comparable.

HechtiDerLachs:
The code seems to still be a bit brittle. My latest tunings revealed that `coordinates` on `SubquoModuleElem`s can produce unreliable results in the sense that if a generator `g` of the `parent` of `v` is zero, then `coordinates(v)` might have a non-zero entry `c` for `g`. But that means that we may create non-homogeneous elements from homogeneous ones when taking preimages along maps using `coordinates`. It is not 100% clear to me what the design should be here. All I know is that the code works for the moment. But maybe this should be given a second thought and cleaned up eventually.

jankoboehm:
Testing the code from the book, I get on my clone the following issue:

```julia
K = GF(3);
S, (x0, x1, x2, x3, x4) = graded_polynomial_ring(K, ["x0", "x1", "x2", "x3", "x4"]);
m = ideal(S, [x1^2+(-x1+x2+x3-x4)*x0, x1*x2+(x1-x3+x4)*x0, x1*x3+(-x1+x4+x0)*x0, x1*x4+(-x1+x3+x4-x0)*x0, x2^2+(x1-x2-x4-x0)*x0, x2*x3+(x1-x2+x3+x4-x0)*x0, x2*x4+(x1+x2-x3-x4-x0)*x0, x3^2+(x3+x4-x0)*x0,x3*x4+(-x3-x4+x0)*x0, x4^2+(x1+x3-x4-x0)*x0]);
R, _ = quo(S, m);
FR = free_resolution(R, algorithm = :mres);
ERROR: UndefVarError: S not defined
Stacktrace:
 [1] free_resolution(Q::MPolyQuoRing{MPolyDecRingElem{FqFieldElem, FqMPolyRingElem}}; length::Int64, algorithm::Symbol)
   @ Oscar ~/oscardegree/Oscar.jl/src/Modules/UngradedModules/FreeResolutions.jl:522
 [2] top-level scope
   @ REPL[7]:1
```
Can you please check whether you also get that?

jankoboehm:
This so far gives a codimension of 14 (which is correct), but now gives 5:

```julia
K = GF(3);

S, (x0, x1, x2, x3, x4) = graded_polynomial_ring(K, ["x0", "x1", "x2", "x3", "x4"]);

m = ideal(S, [x1^2+(-x1+x2+x3-x4)*x0, x1*x2+(x1-x3+x4)*x0, x1*x3+(-x1+x4+x0)*x0, x1*x4+(-x1+x3+x4-x0)*x0, x2^2+(x1-x2-x4-x0)*x0, x2*x3+(x1-x2+x3+x4-x0)*x0, x2*x4+(x1+x2-x3-x4-x0)*x0, x3^2+(x3+x4-x0)*x0,x3*x4+(-x3-x4+x0)*x0, x4^2+(x1+x3-x4-x0)*x0]);

R, _ = quo(S, m);

FR = free_resolution(R, algorithm = :mres);

L = monomial_basis(R, 2)

versal_unfolding = [[i == ((j-1) % 5 + 1) ? S(L[((j-1) % 5) + 1]) : S(0) for i in 1:10] for j in 1:50];

function normal_space_generator(FR, A1t)
    phi1 = map(FR, 1)
    e1 = gen(codomain(phi1),1)
    phi2 = map(FR, 2)
    phi3 = map(FR, 3)
    A1 = hom(domain(phi1), codomain(phi1), [p*e1 for p in A1t])
    A1phi2 = phi2 * A1
    A2 = lift(A1phi2, phi1)
    A2phi3 = phi3 * A2
    A3 = lift(A2phi3, phi2)
    A3m = matrix(A3)
    return A3m[1:2, 16:22]
end;

nlist = [normal_space_generator(FR, v) for v in versal_unfolding];

TS, t = graded_polynomial_ring(K, "t"=>(1:50));

nlist_t = [map_entries(x -> x * t[i], map_entries(constant_coefficient, nlist[i])) for i in 1:50];

B = sum(nlist_t);

tangent_space = ideal(vec(B));

ngens(leading_ideal(tangent_space))
5
```
We probably should follow the computation step-by-step. Don't have a more minimal example yet. Can you please try if you also get 5?

wdecker:
JUST AN ADDENDUM: This is code from the book.

HechtiDerLachs:
I had to change various things about how module elements are lifted through maps. Maybe this affects your code inside `normal_space_generators`. 
However, the changes I made were necessary to guarantee correctness of degrees. But if you say that the result now is mathematically incorrect, that's bad. I will have a look. 

Edit: I get the same error on the latest master.

jankoboehm:
`L` is still right, so the ideal is probably ok.

HechtiDerLachs:
As I said, I also get `5` as the final result on the latest master. Thus, I'm afraid, this is not a problem of this PR in particular.

jankoboehm:
OK found the issue, a character was changed in the jlcon in the book. So this is resolved. Will go on with the code.

jankoboehm (approved):

thofma:
What is the plan for the `:task` keyword, which is still there for the `truncate` function?

thofma (dismissed):
Just adding this here, so that we don't accidentally merge this without resolving the `:task` question.

HechtiDerLachs:
~~It turns out, this PR has another problem~~ (apart from the tests failing again):

**Edit:** This seems to be about the flattenings of modules over rather exotic rings. These can not be expected 
to work in general and so I think it's safe to leave this cleanup to another PR. That this did not blow up on master is only luck. So we should have the problems below in mind, but not have this PR blocked by that. 

```
  kk = GF(40009);
  B, (t,) = polynomial_ring(kk, [:t]);
  IP5 = projective_space(B, [:u, :v, :w, :x, :y, :z]);
  S = homogeneous_coordinate_ring(IP5);
  (u, v, w, x, y, z) = gens(S);
  f = x*y - t^2*v^2 - w^2;
  g = z*w - t*u*v;
  I = ideal(S, [f, g]);
  X, inc_X = sub(IP5, I);
  Om1X = Oscar.relative_cotangent_module(X);
  SX = homogeneous_coordinate_ring(X);
  F1X = graded_free_module(SX,[0]);
  TX,_ = hom(Om1X,F1X);
  TXamb,_ = pushforward(inc_X,TX);
  W1X,_ = hom(TX,F1X)
  M, a, b = Oscar._alt_simplify(W1X)
  is_isomorphism(a) # true
  is_isomorphism(b) #false
  compose(a, b) == identity_map(domain(a)) # true
  compose(b, a) == identity_map(domain(b)) # true
```
If one proceeds by 
```
K, inc = kernel(b)
v = inc(K[1])
is_zero(v) #false 
w = b(v)
is_zero(w) # true
is_zero(a(w)) # false 
```
The only thing I found so far is that the `SubModuleOfFreeModule` layer of the modules is still not sufficiently generic so that in one way or the other Groebner basis reduction via singular is called on modules over towers of polynomial rings. But this will be nasty to debug, I guess. Interestingly, the above code does not fail on the latest master.

jankoboehm:
> Just adding this here, so that we don't accidentally merge this without resolving the `:task` question.

The idea is to do that issue in a separate PR. This one should limit to the degree check question.

jankoboehm:
> Interestingly, the above code does not fail on the latest master.
Before merging, we should find out why it works on master, and now not any more with this modification. Otherwise one surely will be able to produce other examples which fail. Perhaps I have an idea.

Reduction is part of the layer.

HechtiDerLachs:
> Before merging, we should find out why it works on master, and now not any more with this modification. Otherwise one surely will be able to produce other examples which fail. Perhaps I have an idea.

@jankoboehm : This is not a citation, but what you said, correct?

In any case, I have a fix now and I understand why it worked on master but not here. This can be pushed to #3348 . But it will be several hundred lines of code unrelated with this PR's original intention. So I think, it should also go separately.

HechtiDerLachs:
I had no choice but to also include the changes of the flattenings of modules. Otherwise the tests wouldn't run again. Also, they were running for the wrong reasons before, so I think, it's better this way.

HechtiDerLachs:
Good to be merged. Only non-required tests have not run yet and I don't see how to trigger them.

lgoettgens:
Backport?

lgoettgens:
@benlorenz @aaruni96 some heads-up for the backport: This PR introduced some code that is not compatible with AbstractAlgebra v0.40.0. For the master branch, this was replaced by the new interface in https://github.com/oscar-system/Oscar.jl/pull/3374. For the backporting of these two commits, you probably either need to keep the order as on the master branch (first cherry-pick this and #3374 afterwards), or apply [`cb62fb3` (#3374)](https://github.com/oscar-system/Oscar.jl/pull/3374/commits/cb62fb3553b31998f7de251d52b34bfc27e62ee0) and then [`404ac73` (#3374)](https://github.com/oscar-system/Oscar.jl/pull/3374/commits/404ac734df239099dfd3dd3c4f946cb5bb6b6918) after cherry-picking this PR to make it work again.

benlorenz:
Thanks for the info, I did backport #3374 earlier today with some small conflicts that I needed to resolve. (due to not having this PR yet)
But I will now try to redo this in the correct order just to be safe.

PR #3302, opened 2024-02-01
Simplify and unify GAP iteration code

fingolfin:


ThomasBreuer (approved):
Yes, this is an improvement. Thanks.

(Some of the signatures  do in fact not guarantee that the objects are "GAP based", but this can be fixed in other pull requests.)

PR #3288, opened 2024-01-31
Adjust to AA/Nemo/Hecke

joschmitt:
Requires https://github.com/oscar-system/Singular.jl/pull/767

PR #3231, opened 2024-01-24
Make new versions of AA, Nemo, Hecke available

lgoettgens:
This PR collects everything that has to be adapted to keep master green.

GroupsCore: includes and thus closes #3139, closes #3070
Padics: includes and thus closes #3196

[skip ci]

PR #3303, opened 2024-02-01
Add serialization upgrade script for type renaming

lgoettgens:
This approaches to add an upgrade script for serialized files incorporating the renamings from https://github.com/oscar-system/Oscar.jl/pull/3231 and https://github.com/oscar-system/Oscar.jl/pull/3288. A script for applying the renaming can is available in https://github.com/oscar-system/Oscar.jl/pull/3288#issuecomment-1919784205.

I looked through all occurrences of `@register_serialization_type` in the codebase and compared them with the renaming script. If the type got renamed, both old and new names got added to the list in the serialization code.

Everything got a bit finicky, as type strings can be at a lot of different places, and furthermore inside `:_refs` and this needs to get reloaded after the upgrading into the DeserializationState.

I reverted to `data/JohnsonMatrices/` folder to before #3288. They should get either kept at this older state or upgraded once this PR works using 
```julia
for index in keys(Oscar._johnson_names)
       println(index)
       loaded = load(joinpath(Oscar.oscardir, "data", "JohnsonMatrices", string("j", index, ".mat")))
       save(joinpath(Oscar.oscardir, "data", "JohnsonMatrices", string("j", index, ".mat")), loaded)
end
```

cc @thofma @joschmitt @antonydellavecchia


lgoettgens:
Ok, so this works now locally with some examples that I serialized on 56feead522e4f25c3ed68b6dc47238addd7ff990 (i.e. before #3231) and loaded them on this branch.
Furthermore, loading all Johnson matrices in `data/JohnsonMatrices/` works now locally (and should be verified by the CI job testing `johnson_solid`).

@antonydellavecchia can you please have a look if what I've done here looks more or less sensible? You don't have to check the concrete types to update, just the general framework of renaming types. Thanks!

antonydellavecchia (approved):
Things look more or less correct. If tests are passing then great. I was a bit concerned about the fact you need to update s.refs after the upgrade, but this is probably something for me to look into. Thanks for doing this!

PR #3272, opened 2024-01-30
Rename `number_of_*` functions

lgoettgens:
according to [this HackMD](https://hackmd.io/nRnyfrSxTVe5CzXidB1cBQ?both#number-of-%E2%80%A6) as discussed in the coding sprint.

cc @YueRen

PR #3305, opened 2024-02-02
Add paragraph about `number_of_` to the styleguide.

lgoettgens:
As layed out in https://hackmd.io/nRnyfrSxTVe5CzXidB1cBQ?both#number-of-%E2%80%A6 and implemented in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1553, https://github.com/Nemocas/Nemo.jl/pull/1624, https://github.com/thofma/Hecke.jl/pull/1364, https://github.com/oscar-system/Oscar.jl/pull/3272.

Thanks to  @lkastner for pointing out that the style guide change was missed.

### DIFF
--- a/docs/src/DeveloperDocumentation/styleguide.md
+++ b/docs/src/DeveloperDocumentation/styleguide.md
@@ -60,7 +60,7 @@ Here is a summary of the naming convention followed in OSCAR:
   Julia functions, e.g. `is_one` as alias for `isone`.
 - A function returning the number of some things should be named `number_of_things`.
   For some very common things, like the number of generators, we additionally
-  provide an alias `ngens` for `number_of_generators`. These aliases should be
+  provide a shorter alias, e.g. `ngens` for `number_of_generators`. These aliases should be
   short (< 15 chararacters) and without underscores, and one should only use
   them inside of functions, not in the user interface and documentation.
 - For generic concepts choose generic names, based on general algebraic
--- a/docs/src/DeveloperDocumentation/styleguide.md
+++ b/docs/src/DeveloperDocumentation/styleguide.md
@@ -58,11 +58,12 @@ Here is a summary of the naming convention followed in OSCAR:
   For compatibility with standard Julia, while staying consistent internally,
   we also provide aliases (using `AbstractAlgebra.@alias`) for various standard
   Julia functions, e.g. `is_one` as alias for `isone`.
-- A function returning the number of some things should be named `number_of_things`.
+- A function returning the number of some things should be named `number_of_things`,
+  alternatively it can be named `n_things` with an alias to `number_of_things`.
+  The preferred style should be consistent throughout the corresponding part.
   For some very common things, like the number of generators, we additionally
   provide a shorter alias, e.g. `ngens` for `number_of_generators`. These aliases should be
-  short (< 15 chararacters) and without underscores, and one should only use
-  them inside of functions, not in the user interface and documentation.
+  very short and without underscores.
 - For generic concepts choose generic names, based on general algebraic
   concepts, preferably not special names from your area of speciality.
 - **Avoid direct access to members of our objects.** This means, do not use
--- a/docs/src/DeveloperDocumentation/styleguide.md
+++ b/docs/src/DeveloperDocumentation/styleguide.md
@@ -57,7 +57,12 @@ Here is a summary of the naming convention followed in OSCAR:
   also being consistent.
   For compatibility with standard Julia, while staying consistent internally,
   we also provide aliases (using `AbstractAlgebra.@alias`) for various standard
-  Julia functions, e.g. `is_one` as alias for `isone`
+  Julia functions, e.g. `is_one` as alias for `isone`.
+- A function returning the number of some things should be named `number_of_things`.
+  For some very common things, like the number of generators, we additionally
+  provide an alias `ngens` for `number_of_generators`. These aliases should be
+  short (< 15 chararacters) and without underscores, and one should only use
+  them inside of functions, not in the user interface and documentation.
 - For generic concepts choose generic names, based on general algebraic
   concepts, preferably not special names from your area of speciality.
 - **Avoid direct access to members of our objects.** This means, do not use


YueRen (dismissed):

joschmitt (approved):

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -57,7 +57,12 @@ Here is a summary of the naming convention followed in OSCAR:
   also being consistent.
   For compatibility with standard Julia, while staying consistent internally,
   we also provide aliases (using `AbstractAlgebra.@alias`) for various standard
-  Julia functions, e.g. `is_one` as alias for `isone`
+  Julia functions, e.g. `is_one` as alias for `isone`.
+- A function returning the number of some things should be named `number_of_things`.
+  For some very common things, like the number of generators, we additionally
+  provide an alias `ngens` for `number_of_generators`. These aliases should be
```
```suggestion
  provide a shorter alias, e.g. `ngens` for `number_of_generators`. These aliases should be
```

fingolfin (dismissed):

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -57,7 +57,12 @@ Here is a summary of the naming convention followed in OSCAR:
   also being consistent.
   For compatibility with standard Julia, while staying consistent internally,
   we also provide aliases (using `AbstractAlgebra.@alias`) for various standard
-  Julia functions, e.g. `is_one` as alias for `isone`
+  Julia functions, e.g. `is_one` as alias for `isone`.
+- A function returning the number of some things should be named `number_of_things`.
+  For some very common things, like the number of generators, we additionally
+  provide a shorter alias, e.g. `ngens` for `number_of_generators`. These aliases should be
+  short (< 15 chararacters) and without underscores, and one should only use
+  them inside of functions, not in the user interface and documentation.
```
fingolfin: I think this is not correct anymore, but I am not sure what the correct new description would be. Perhaps @lkastner @benlorenz @micjoswig could propose something?

In particular I am not quite sure myself. So the short aliases now start with `n_`, but we keep `number_of_`? Or do we plan to phase out `number_of_` in favor of `n_` (I'd be fine with that, too, just need to know what the plan is).

Thanks!

benlorenz:
```suggestion
- A function returning the number of some things should be named `number_of_things`,
  alternatively it can be named `n_things` with an alias to `number_of_things`.
  The preferred style should be consistent throughout the corresponding part.
  For some very common things, like the number of generators, we additionally
  provide a shorter alias, e.g. `ngens` for `number_of_generators`. These aliases should be
  very short and without underscores.
```
We keep `number_of_` and allow the use of `n_` with a `number_of_` alias as well.

The short and abbreviated aliases `ngens`, `nvars`, `ncols` and `nrows` also stay as they are. But I do not think we can expect developers to use them only inside functions but not for user-facing code so I removed that part.

lgoettgens:
I accepted the suggestion by @benlorenz to be able to add comments to specific things.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -57,7 +57,13 @@ Here is a summary of the naming convention followed in OSCAR:
   also being consistent.
   For compatibility with standard Julia, while staying consistent internally,
   we also provide aliases (using `AbstractAlgebra.@alias`) for various standard
-  Julia functions, e.g. `is_one` as alias for `isone`
+  Julia functions, e.g. `is_one` as alias for `isone`.
+- A function returning the number of some things should be named `number_of_things`,
+  alternatively it can be named `n_things` with an alias to `number_of_things`.
```
lgoettgens, line 62: This reads to me as if everything should have a `number_of_things` function and may additionally have a `n_things` alias. Is this what you intended to say here? I understood it before that every `number_of_things` function should have an `n_things` alias.

docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -57,7 +57,13 @@ Here is a summary of the naming convention followed in OSCAR:
...
+  The preferred style should be consistent throughout the corresponding part.
+  For some very common things, like the number of generators, we additionally
+  provide a shorter alias, e.g. `ngens` for `number_of_generators`. These aliases should be
+  very short and without underscores.
```
lgoettgens, line 66:
```suggestion
  very short and without underscores, and one should only use
  them inside of functions, not in the user interface and documentation.
```
I think this is important to state: For consistency, I think we should attach docstrings not to `ngens` but to one of the other two, and the docstring should contain the signature of the "long form". It may additionally state the existence of the `ngens` alias, but that should not be too dominantly to not confuse users with three names for a single function.

docs/src/DeveloperDocumentation/styleguide.md
benlorenz: Everything should be named either `number_of_things` or `n_things`. If it is named `n_things` (and documented like that) then it should have an alias `number_of_things`. If it is named `number_of_` it may have an alias `n_`.

I did not add `n_` aliases for all `number_of` functions.

docs/src/DeveloperDocumentation/styleguide.md
benlorenz: I don't really like the "developers may use ngens internally but users should not be exposed to it too much" approach. Julia makes it very easy to look at the code with `@less` or similar functions.

There are about 20 `ngens` calls in the book as well.

docs/src/DeveloperDocumentation/styleguide.md
lgoettgens: Ah, I didn't notice you swapped the alias direction of only some functions in #3364. I would find it way more consistent to have all aliases the same there, i.e. `number_of_` is the default, and then `n_` is the alias. But that's a different story. The formulation here now matches the reality of the implementation

lgoettgens:
To settle this discussion once and for all, I would like all of @fieker @fingolfin @micjoswig @thofma to comment on this PR. Once all of them approve, this is good to get merged, and we can point people to the style guide.

docs/src/DeveloperDocumentation/styleguide.md
micjoswig: Since the changes are on a branch of a private repo (which I find unnecessarily tedious), I just write comments.

* `number_of` or `n_` but only one version per section. Example: `n_` is used throughout the polyhedral geometry section.
* `number_of` should be available as alias if the function is written as `n_`
* `n_` may be used as an alias if the function is written as `number_of`
* `ngens` should become `n_gens` then

I agree with @benlorenz that it does not make much sense to distinguish between "internal" and "external".  After all, this is an open source project.

fieker:
wair for @fingolfin

docs/src/DeveloperDocumentation/styleguide.md
fingolfin: I think this paragraph is fine now -- at least it seems to match what @benlorenz described and I am fine with it, too.

docs/src/DeveloperDocumentation/styleguide.md
fingolfin: So it seems we agree at least in so far as: there should always be a `number_of_BLA` and there *may* be also a `n_BLA` (ignoring the technical details about e.g. which one may or may not be an "alias" (as defined via `@alias`) of which one for the moment).


But note that renaming `ngens` to `n_gens` would not quite fit: since we have `number_of_generators` it would have to be called `n_generators`... Moreover we use `ngens` and `gens` in ~1000 spots in Oscar.jl alone, and as Benjamin pointed out, it is used 20 times in book.

This is why in January a dozen of us agreed after multiple discussions over two weeks to allow certain "ultra short" aliases being grandfathered in... besides `ngens` that's also `nrows` and `ncols` and probably more (e.g. perhaps `nv` or `ne` for Graphs.jl compatibility?! (I am not a fan of those names, but if someone thinks this is useful to have, I don't want to block them?).


That said, I have no personal attachment to `ngens` and could easily adapt to using `n_gens` or even `n_generators`, but my point here is that this is a big change requiring a lot of work (with changes over multiple packages, once again carefully coordinated); the `n_gens` variant still is not really consistent; and in the end we can't get rid of `ngens` for now anyway. So is it really worth the trouble at this point?

fingolfin (approved):
I am OK with the text as it is written right now. It may not be perfect and perhaps not yet capture every nuance but it seems like a good foundation, and others can open additional PRs to further tweak it if they think it is helpful.

That said, as @lgoettgens wrote, it would be good to arrive at a form that @fieker @thofma @micjoswig also are fine with. So I hope they'll either approve this, or actively continue the discussion so that we can arrive at a comprosise everyone can work with.

thofma:
Fine with me

docs/src/DeveloperDocumentation/styleguide.md
micjoswig: I agree with @fingolfin and @benlorenz .

That is, we have `n_things` and `number_of_things` for most things, but for historical and workload reasons the odd `ngens` is fine, too.  No `n_gens` then.

Further, as @benlorenz pointed out: no distinction between "interior use" and "interface".

@lgoettgens : Personally, I do not need `number_of_generators`, but if it makes you happy, this could be an alias to `ngens`.

micjoswig (approved):

PR #3306, opened 2024-02-02
change one line printing of affine schemes from V(...) to scheme(...)

simonbrandhorst:
Currently affine algebraic sets and schemes both print in one line mode with `V(...)` but then it is unclear 
whether or not the scheme considered is assumed to be reduced or not. 
After this PR we have `V(...)` for affine algebraic sets and `scheme(...)` for  affine schemes.

### DIFF
--- a/experimental/Schemes/SpaceGerms.jl
+++ b/experimental/Schemes/SpaceGerms.jl
@@ -645,7 +645,7 @@ Spectrum
 julia> phi
 Affine scheme morphism
   from [x1, x2, x3]  Spec of localization of Q at complement of maximal ideal
-  to   [x1, x2, x3]  V(x1^2 + x2^2 + x3^2)
+  to   [x1, x2, x3]  scheme(x1^2 + x2^2 + x3^2)
 given by the pullback function
   x1 -> x1
   x2 -> x2
@@ -723,7 +723,7 @@ Spectrum
 julia> phi
 Affine scheme morphism
   from [x1, x2, x3]  Spec of localization of Q at complement of maximal ideal
-  to   [x1, x2, x3]  V(x1^2 + x2^2 + x3^2)
+  to   [x1, x2, x3]  scheme(x1^2 + x2^2 + x3^2)
 given by the pullback function
   x1 -> x1
   x2 -> x2
@@ -798,7 +798,7 @@ Spectrum
 julia> phi
 Affine scheme morphism
   from [x1, x2, x3]  Spec of localization of Q at complement of maximal ideal
-  to   [x1, x2, x3]  V(x1^2 + x2^2 + x3^2, x1*x2)
+  to   [x1, x2, x3]  scheme(x1^2 + x2^2 + x3^2, x1*x2)
 given by the pullback function
   x1 -> x1
   x2 -> x2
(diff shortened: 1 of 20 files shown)


experimental/Schemes/SpaceGerms.jl
```diff
@@ -645,7 +645,7 @@ Spectrum
 julia> phi
 Affine scheme morphism
   from [x1, x2, x3]  Spec of localization of Q at complement of maximal ideal
-  to   [x1, x2, x3]  V(x1^2 + x2^2 + x3^2)
+  to   [x1, x2, x3]  scheme(x1^2 + x2^2 + x3^2)
```
afkafkafk13, line 648: Here we are disclosing schemes, which are the underlying technology,  to a non-scheme community. For them this represents a morphism of germs of complex spaces. This can certainly be confusing for them.

But as I see no alternative, we leave it as it is now...

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
```diff
@@ -251,15 +251,15 @@ Spectrum
...
+  from [x1, x2, x3]  scheme(x1)
   to   [x1, x2, x3]  affine 3-space over QQ
 given by the pullback function
   x1 -> 0
   x2 -> x2
   x3 -> x3
 
 julia> graph(f)
-(V(x1, -x1, x2 - x2, x3 - x3), Hom: V(x1, -x1, x2 - x2, x3 - x3) -> V(x1), Hom: V(x1, -x1, x2 - x2, x3 - x3) -> affine 3-space over QQ with coordinates [x1, x2, x3])
+(scheme(x1, -x1, x2 - x2, x3 - x3), Hom: scheme(x1, -x1, x2 - x2, x3 - x3) -> scheme(x1), Hom: scheme(x1, -x1, x2 - x2, x3 - x3) -> affine 3-space over QQ with coordinates [x1, x2, x3])
```
afkafkafk13, line 262: 'Hom' feels like the wrong term here. On the algebraic side, 'Hom' is the right word, but usually, this is not referred to as a homomorphism on the geometric side, but as a morphism. Is this printing tied to the printing on the algebraic side? If so, leave it as it is. If not, this needs to be discussed (unless it has already been discussed and decided).

afkafkafk13 (commented):
Overall I agree with the PR and will approve it after an answer to one of my comments, the one concerning 'Hom'.

experimental/Schemes/SpaceGerms.jl
simonbrandhorst: You could implement a custom printing method for space germs?

afkafkafk13: Indeed, this is the way in which I should solve it.

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
simonbrandhorst: It is not tied to the algebraic side. As far as I know there is no executive decision. We should be free to switch to `Mor` but let me check back with someone else.

HechtiDerLachs (approved):

afkafkafk13 (approved):

PR #3310, opened 2024-02-02
Support matrix groups over QQBar

### LINKED ISSUE #3279: Conversion of `(Calcium)QQBar` elements to "plain" number fields; matrix groups over it
I tried producing a rotation matrix in a "natural" fashion by using `CalciumQQBar` (soon to be renamed to `QQBar`). This worked quite well:
```
julia> e = one(CalciumQQBar)
Root 1.00000 of x - 1

julia> s, c = sinpi(2*e/5), cospi(2*e/5);

julia> mat = matrix(CalciumQQBar, [ c -s ; s c ])
[    Root 0.309017 of 4x^2 + 2x - 1   Root -0.951057 of 16x^4 - 20x^2 + 5]
[Root 0.951057 of 16x^4 - 20x^2 + 5        Root 0.309017 of 4x^2 + 2x - 1]

julia> mat^5
[Root 1.00000 of x - 1             Root 0 of x]
[          Root 0 of x   Root 1.00000 of x - 1]
```

Now I wanted to turn this into a group (a simple cyclic group of order 5):
```
julia> G_mat = matrix_group(mat)
Matrix group of degree 2
  over field of algebraic numbers```
```

But this group can't do much, we can't even compute its order. Trying to do so, we first hit a technical error:
```
julia> order(G_mat)
ERROR: attributes storage not supported for type CalciumQQBarField
```
but that can be worked around by adding `@attributes CalciumQQBarField` in the REPL (long term fix at https://github.com/Nemocas/Nemo.jl/pull/1645).

Afterwards we get this:
```
julia> order(G_mat)
ERROR: MethodError: no method matching _iso_oscar_gap(::CalciumQQBarField)

Closest candidates are:
  _iso_oscar_gap(::ZZRing)
   @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/GAP/iso_oscar_gap.jl:215
  _iso_oscar_gap(::QQField)
   @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/GAP/iso_oscar_gap.jl:202
  _iso_oscar_gap(::MPolyRing{T}) where T
   @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/GAP/iso_oscar_gap.jl:538
  ...

Stacktrace:
  [1] __compute_iso_oscar_gap__(F::CalciumQQBarField)
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/GAP/iso_oscar_gap.jl:457
```

Also not surprising, we simply don't support this yet.

I think there are two things of interest here:
- [x] add code which takes a bunch of elements of `QQBar` (or possibly a matrix, or a bunch of matrices...) and tries to produce a number field `K` in which these elements can be represented. I guess this would return a function which converts from `QQBar` elements to `K` (or perhaps cleaner: a morphism `K \to QQBar` which allows computing preimages)
- [ ] code to support matrix groups over this field, at least *finite* matrix groups
   - e.g. by adding `_iso_oscar_gap` method to handle `QQBarField`, at least "when possible"
   - and/or by adapting `_isomorphic_group_over_finite_field` to support this
   - ... or by any other means

CC @ThomasBreuer

fingolfin:
Resolves #3279

Maybe not in the most elegant way but it seems to work sufficiently well for now.


src/Groups/matrices/MatGrp.jl
```diff
@@ -205,12 +205,25 @@ function assign_from_description(G::MatrixGroup)
...
-function Base.getproperty(G::MatrixGroup, sym::Symbol)
+function Base.getproperty(G::MatrixGroup{T}, sym::Symbol) where T
 
    isdefined(G,sym) && return getfield(G,sym)
 
    if sym === :ring_iso
-      G.ring_iso = iso_oscar_gap(G.ring)
+      if T === QQBarFieldElem
+         # get all matrix entries into one vector
+         entries = vcat([reshape(collect(matrix(g)),:) for g in gens(G)]...)
```
thofma:
```suggestion
         entries = reduce(vcat, vec(collect(matrix(g))) for g in gens(G))
```
I hope I did not miscount the `)`.

fingolfin: Ah, `vec` was the function I was looking for and could not remember, thanks.

thofma (approved):

test/Groups/matrixgroups.jl
```diff
@@ -143,6 +143,22 @@ end
    @test order(Oscar.isomorphic_group_over_finite_field(G)[1]) == 1
 end
 
+@testset "matrix group over QQBar" begin
+   #K = algebraic_closure(QQ)
+   K = Nemo.QQBar
```
lgoettgens:
```suggestion
   K = algebraic_closure(QQ)
```
Nemo 0.41.3 already contains this. You just need to bump the compat entry in the `Project.toml`

lgoettgens: You changed the compat but didn't commit this change?

fingolfin: indeed, forgot to `git add test` fixed now

fingolfin:
@ThomasBreuer does this look OK to you (for now, at least, I am happy to look into better solutions longterm)

fingolfin:
I really need this for the book, so I'll merge it now, any improvements can be made atop this.

PR #3311, opened 2024-02-04
Bump Hecke version, fix some regressions

### LINKED ISSUE #3107: dualizing does not work for free resolutions

thofma:
Fixes some woes introduced in some of the new type names.
- Remove redundancy for embedded types by using aliases
- use compact printing for abelian group elements in arrays


lgoettgens:
The 1.10 failure will be fixed by https://github.com/oscar-system/Oscar.jl/pull/3312

fieker (approved):

PR #3315, opened 2024-02-05
Add missing hash methods for maps

fingolfin:
We have this method in Oscar (I think @HechtiDerLachs added it?):
```julia
# Problems arise with comparison for non-trivial coefficient maps 
# in all constellations. This calls for a streamlined set of commands 
# to check whether coefficient maps exist and if so, what they are, 
# so that they can be compared. 
function Base.:(==)(
    f::Map{<:MPolyAnyRing, <:MPolyAnyRing},
    g::Map{<:MPolyAnyRing, <:MPolyAnyRing}
   )
...
```

However, many of our `Map` types don't implement `hash` -- but implementing
`==` for a type requires that also `hash` is implemented for correctness.

As a result, we currently get this:
```
julia> R, (x, y) = QQ[:x,:y]
(Multivariate polynomial ring in 2 variables over QQ, QQMPolyRingElem[x, y])

julia> f1 = hom(R, R, [y,x]);

julia> f2 = hom(R, R, [y,x]);

julia> f1 == f2
true

julia> hash(f1)
0xb7ee188fb7693c7c

julia> hash(f2)
0x683deafb5f3b9ec7
```

This patch proposes a fix. Note that the TODO in the new `hash` function (which points out that the coefficient map could also be taken into account for computing the hash) is just about getting a better hash, it is not needed for correctness.

A similar issue is fixed in `experimental/GModule/Cohomology.jl` ; that code there should perhaps only apply to `FPModuleHomomorphism` but that's a slightly bigger change....

I guess I should add tests ...

### DIFF
--- a/experimental/GModule/Cohomology.jl
+++ b/experimental/GModule/Cohomology.jl
@@ -1713,7 +1713,13 @@ function Oscar.matrix(M::FreeModuleHom{FreeMod{QQAbElem}, FreeMod{QQAbElem}})
 end
 
 function ==(a::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism}, b::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism})
+  domain(a) === domain(b) || return false
+  codomain(a) === codomain(b) || return false
   return matrix(a) == matrix(b)
+end
+
+function Base.hash(a::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism}, h::UInt)
+  return hash(matrix(a), h)
 end
 
 function Oscar.id_hom(A::AbstractAlgebra.FPModule)
--- a/experimental/GModule/Cohomology.jl
+++ b/experimental/GModule/Cohomology.jl
@@ -1719,7 +1719,10 @@ function ==(a::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism}, b::
 end
 
 function Base.hash(a::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism}, h::UInt)
-  return hash(matrix(a), h)
+  h = hash(domain(a), h)
+  h = hash(codomain(a), h)
+  h = hash(matrix(a), h)
+  return h
 end
 
 function Oscar.id_hom(A::AbstractAlgebra.FPModule)
--- a/src/Rings/mpolyquo-localizations.jl
+++ b/src/Rings/mpolyquo-localizations.jl
@@ -2407,6 +2407,14 @@ function Base.:(==)(
   return all(x->f(x) == g(x), gens(domain(f)))
 end
 
+function Base.hash(f::Map{<:MPolyAnyRing, <:MPolyAnyRing}, h::UInt)
+    h = hash(domain(f), h)
+    h = hash(codomain(f), h)
+    # TODO: add in coefficient_map if available
+    h = hash(f.(gens(domain(f))), h)
+    return h
+end
+
 coefficient_map(f::MPolyLocalizedRingHom) = coefficient_map(restricted_map(f))
 coefficient_map(f::MPolyQuoLocalizedRingHom) = coefficient_map(restricted_map(f))
 


HechtiDerLachs:
Thanks for the catch.

I think, I once added something like `_has_coefficient_map` to check whether or not there is one. This can not be deduced from the abstract `Map`-type and the more specialized types differ too much in their signatures. But in principle, the implementation of `_has_coefficient_map` should only depend on the input's type. 

I think with that, one can fill the gaps.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2407,6 +2407,14 @@ function Base.:(==)(
   return all(x->f(x) == g(x), gens(domain(f)))
 end
 
+function Base.hash(f::Map{<:MPolyAnyRing, <:MPolyAnyRing}, h::UInt)
+    h = hash(domain(f), h)
+    h = hash(codomain(f), h)
+    # TODO: add in coefficient_map if available
```
HechtiDerLachs, line 2413:
```suggestion
    _has_coefficient_map(f) && (h = hash(coefficient_map(f), h))
```
Problem is: coefficient maps can be anything. In particular they can be mathematically the same, even though they have different data types in the computer and, hence, most probably also different hashes. So one could also think about restricting the hash and comparison functions to maps without coefficient maps and throw an error.

fingolfin: I saw `_has_coefficient_map` and did not add code like in your suggestion precisely because of the issue you point out. 

I guess could compute the image of the generators under the coefficient map and hash that; but it is not clear to me this will work in general, and also the performance impact (this will cause a bunch of allocations!) is unclear.

Of course we could cache the hash (rhyming not intended ;-) ) to overcome that.

fingolfin: But note that this is strictly optional: if we just don't consider the coefficient maps for the hash, we still have something that is correct.

experimental/GModule/Cohomology.jl
```diff
@@ -1713,9 +1713,15 @@ function Oscar.matrix(M::FreeModuleHom{FreeMod{QQAbElem}, FreeMod{QQAbElem}})
 end
 
 function ==(a::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism}, b::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism})
+  domain(a) === domain(b) || return false
+  codomain(a) === codomain(b) || return false
   return matrix(a) == matrix(b)
 end
 
+function Base.hash(a::Union{Generic.ModuleHomomorphism, Generic.ModuleIsomorphism}, h::UInt)
+  return hash(matrix(a), h)
```
lgoettgens:
```suggestion
  h = hash(domain(a), h)
  h = hash(codomain(a), h)
  h = hash(matrix(a), h)
  return h
```
i think it would make sense to add domain/codomain to the hash as well (as it is the case in `==`)

HechtiDerLachs: Also I would make sure that a `sparse_matrix` method also exists and use that. I know that these modules are dense, but their homomorphisms are probably not.

src/Rings/mpolyquo-localizations.jl
HechtiDerLachs: > but it is not clear to me this will work in general

No, it won't. Take, for instance, the map on `(QQ[i])[x, y]` which is the identity on the variables, but complex conjugation on the coefficients.

HechtiDerLachs: > if we just don't consider the coefficient maps for the hash, we still have something that is correct.

True. So hashing could (and probably should) ignore the coefficient map. The equality check, on the other hand, should throw an error in case there are coefficient maps present which are not `===`. In that case I think, we simply can not write generic code due to the generality allowed for coefficient maps. 

@thofma : Am I seeing this correct?

thofma: Yes, and yes to throwing an error.

(Maybe just do `hash(f::Map, h::UInt) = UInt(0)` and be done with it.)

fingolfin: Error: fine by me.

`hash(f::Map, h::UInt) = UInt(0)`:  this discards any previous hash results, which is bad when chaining hashes; better would be `hash(f::Map, h::UInt) = h` . That said, I personally prefer an `error` in this case, i.e. `hash(f::Map, h::UInt) = error("hashing not implemented for this map type, please report this as bug to the OSCAR team")` :-)

experimental/GModule/Cohomology.jl
fingolfin: Actually I had the code like this initially. But actually `matrix` is not defined for arbitrary `ModuleHomomorphisms`; AFAICT this code is only really used for `FPModuleHomomorphism` and then the matrix uniquely determines `domain` and `codomain`. In practice it is also harmless to have the "weaker" (but faster!) hash which does not take the domain and codomain into account: if we really have two morphisms with equal matrix but differing domain/codomain, *and* we use them, say, as keys in the same dictionary, then so be it.

So right now I'd rather lean in the opposite direction and remove these from the other `hash` method in this PR .... :-).

But in the end I really don't care too much about this either way... more important for me is that we fix the correctness bug, not how *shrug*

fingolfin:
Would be great to get this approved and merged (I applied Lars' suggestion)

lgoettgens (approved):
I think the current state of this is reasonable. We can change it again in the future, if something comes up

Issue #3125, opened 2023-12-20
Add more macros for constructors like `@polynomial_ring`


fingolfin:
- [ ] `@graded_polynomial_ring`
- [x] `@free_group`
...

PR #3317, opened 2024-02-05
Use `@varnames_interface free_group`

fingolfin:
- enables many variants of the `free_group` argument syntax
- adds `@free_group` macro

Implements part of #3125

The main issue is the changed return value. I did not bother to adjust all tests or documentation in this PR to that, the changes I made are enough to illustrate the pros and cons I think: returning the group *and* generators is nice and natural in some cases, and ugly and unnatural in others (just like with `polynomial_ring` I guess ;-) ).

But the major concern I have about it is that it now differs from all other (GAP based?) group constructors. Of course all of these could also be made to return generators, but this change is more and more breaking.

So right now my thought would be to drop this part and let `free_group` only return the group. OK? Thoughts?

To make this possible, we'll have to either tweak `@varnames_interface` a bit, or else use a little trickery (like: rename `free_group` to `_free_group`, then let `free_group` just pass all its arguments to `_free_group`.).

This looses any ability to retain the "grouping" of variables when using `free_group`, i.e., there would be no direct replacement for `F, x, y = free_group(:x => 1:3, :y => 1:3)`.

However, it would be retained for the macro version:
```
julia> F = @free_group(:x => 1:3, :y => 1:3)
Free group of rank 6

julia> x
3-element Vector{FPGroupElem}:
 x[1]
 x[2]
 x[3]

julia> y
3-element Vector{FPGroupElem}:
 y[1]
 y[2]
 y[3]
```

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -2273,7 +2273,7 @@ nonzero then `pairs` is the vector of syllables of `x`, see [`syllables`](@ref).
 julia> G = free_group(2);  pairs = [1 => 3, 2 => -1];
 
 julia> x = G(pairs)
-x1^3*x2^-1
+f1^3*f2^-1
 
 julia> syllables(x) == pairs
 true
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1912,18 +1912,18 @@ end
     relators(G::FPGroup)
 
 Return a vector of relators for the full finitely presented group `G`, i.e.,
-elements $[x_1, x_2, \ldots, x_n]$ in $F =$ `free_group(ngens(G))` such that
-`G` is isomorphic with $F/[x_1, x_2, \ldots, x_n]$.
+elements $[w_1, w_2, \ldots, w_n]$ in $F =$ `free_group(ngens(G))` such that
+`G` is isomorphic with $F/[w_1, w_2, \ldots, w_n]$.
 
 # Examples
 ```jldoctest
-julia> f = free_group(2);  (x, y) = gens(f);
+julia> f = @free_group(:x, :y);
 
 julia> q = quo(f, [x^2, y^2, comm(x, y)])[1];  relators(q)
 3-element Vector{FPGroupElem}:
- f1^2
- f2^2
- f1^-1*f2^-1*f1*f2
+ x^2
+ y^2
+ x^-1*y^-1*x*y
 ```
 """
 function relators(G::FPGroup)
@@ -1973,7 +1973,7 @@ whenever it is possible that the elements in `genimgs` do not support `one`.
 
 # Examples
 ```jldoctest
-julia> F = free_group(2);  F1 = gen(F, 1);  F2 = gen(F, 2);
+julia> F = @free_group(:F1, :F2);
 
 julia> imgs = gens(symmetric_group(4))
 2-element Vector{PermGroupElem}:
@@ -2161,7 +2161,7 @@ Return the syllables of `g` as a list of pairs `gen => exp` where
 
 # Examples
 ```jldoctest
-julia> F = free_group(2);  F1, F2 = gens(F);
+julia> F = @free_group(:F1, :F2);
 
 julia> syllables(F1^5*F2^-3)
 2-element Vector{Pair{Int64, Int64}}:
@@ -2194,7 +2194,7 @@ numbers.
 
 # Examples
 ```jldoctest
-julia> F = free_group(2);  F1, F2 = gens(F);
+julia> F = @free_group(:F1, :F2);
 
 julia> letters(F1^5*F2^-3)
 8-element Vector{Int64}:
@@ -2239,7 +2239,7 @@ otherwise an exception is thrown.
 
 # Examples
 ```jldoctest
-julia> F = free_group(2);  F1 = gen(F, 1);  F2 = gen(F, 2);
+julia> F = @free_group(:F1, :F2);
 
 julia> length(F1*F2^-2)
 3
@@ -2273,7 +2273,7 @@ nonzero then `pairs` is the vector of syllables of `x`, see [`syllables`](@ref).
 julia> G = free_group(2);  pairs = [1 => 3, 2 => -1];
 
 julia> x = G(pairs)
-f1^3*f2^-1
+x1^3*x2^-1
 
 julia> syllables(x) == pairs
 true
--- a/src/Groups/group_constructors.jl
+++ b/src/Groups/group_constructors.jl
@@ -477,8 +477,8 @@ projective_omega_group(n::Int, q::Int) = projective_omega_group(0, n, q)
 
 @doc raw"""
     free_group(n::Int, s::VarName = :f; eltype::Symbol = :letter) -> FPGroup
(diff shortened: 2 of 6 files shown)


thofma:
I find my self always typing `F = free_group(["x", "y"]); x, y = gens(F)`, so I would welcome the change of the return type. But I have no strong opinion and I am also probably not one of the main customer for this functionality.

fingolfin:
I guess `free_group` is some sense much closer to `polynomial_ring` or `free_associative_algebra`: for all of these one often wants to work with combinations of "words" expressed in terms of their generators, so it can be useful to ask for those. Which after all is the very reason why I am considering using `@varnames_interface` on `free_groups` !

The difference with `symmetric_group` and matrix groups is that there I do this much less frequently.


So I could see some justification to handle `free_group` and `symmetric_group` differently in this regard. But it's still not great. Hrm.

But I'd like to decouple that problem from this PR, so I'll look into the `_free_group` approach for now.

fingolfin:
This is now ready for review (regardless of whether it passes the test suite or not, I'll fix that up later if necessary).

In particular I've written a comprehensive docstring for `free_group` and a smaller one for `@free_group`. Feedback highly welcome -- perhaps the text I wrote there could be reused to improve the docstring for `polynomial_ring` which doesn't really explain the available options.

src/Groups/GAPGroups.jl
```diff
@@ -2269,7 +2269,7 @@ nonzero then `pairs` is the vector of syllables of `x`, see [`syllables`](@ref).
 julia> G = free_group(2);  pairs = [1 => 3, 2 => -1];
```
lgoettgens: Is it intentional that you didn't change it here?

src/Groups/group_constructors.jl
```diff
@@ -475,27 +475,47 @@ projective_omega_group(n::Int, q::Int) = projective_omega_group(0, n, q)
...
 
-The third form, if there are `n` arguments `L...`,
-returns the free group of rank `n`,
-where the `i`-th generator is printed as `L[i]`.
+In the final form, the argument list consists of a sequence of one or
+more of the following:
+1. A vector `L` of variable names.
+2. A pair of the form `A => B`, where `A` is a `VarName` (so a string, symbol
+   or character) and `B` is a range or more generally an `AbstractVector`.
+   Then `length(B)` generators are defined whose names derived from a combination
```
lgoettgens:
```suggestion
   Then `length(B)` generators are defined whose names derive from a combination
```
Typo I think?

src/Groups/group_constructors.jl
```diff
@@ -475,27 +475,47 @@ projective_omega_group(n::Int, q::Int) = projective_omega_group(0, n, q)
...
+   a tuple of ranges or v. For example `"a" => (1:2, 1:2)` defines four generators
+   `a[1, 1], a[2, 1], a[1, 2], a[2, 2]`.
 
-!!! warning "Note"
-    Variables named like the group generators are *not* created by this function.
+For the second and third type, optionally the `A` part can contain the
+placeholder `#` to modify where the indices are inserted. For example
+`"a#" => (1:2, 1:2)` defines four generators `a11, a21, a12, a22`.
+
+Also instead of a range, any vector can be used. For example `"#" => ([:x,:y], [:A, :B])`
```
lgoettgens:
```suggestion
Also, instead of a range, any vector can be used. For example `"#" => ([:x,:y], [:A, :B])`
```

src/Groups/group_constructors.jl
```diff
@@ -475,27 +475,47 @@ projective_omega_group(n::Int, q::Int) = projective_omega_group(0, n, q)
...
+defines four generators `xA, yA, xB, yB`.
+
+In all variants, if the optional keyword argument `eltype` is given and
```
lgoettgens, line 510: The kwarg is missing from the second and third form's signatures at the top of the docstring

src/Groups/group_constructors.jl
```diff
@@ -505,34 +525,138 @@ Free group of rank 2
...
+julia> F = @free_group(:a, :b)
+Free group of rank 2
+
+julia> a^2*b*a*b^-2
+a^2*b*a*b^-2
+```
+
+Note that the `varname => vector` syntax for specifying a vector or
+matrix or general array of variables behaves slightly differently
+compared to `free_group`, as the following example demonstrates.
```
lgoettgens, line 618: Why is this different? Is this something you really want to have or is it just very hard to work around?

src/Groups/GAPGroups.jl
```diff
@@ -2269,7 +2269,7 @@ nonzero then `pairs` is the vector of syllables of `x`, see [`syllables`](@ref).
...
 
 julia> x = G(pairs)
-f1^3*f2^-1
+x1^3*x2^-1
```
lgoettgens: With you default char hack below, this shouldn't be changed

lgoettgens (commented):
I think this is going in a great direction. I added some comments to things I find weird. And CI is unhappy that you broke `free_group()` without any arguments

src/Groups/GAPGroups.jl
fingolfin: yes

src/Groups/GAPGroups.jl
fingolfin:
```suggestion
f1^3*f2^-1
```

src/Groups/group_constructors.jl
```diff
@@ -475,27 +475,47 @@ projective_omega_group(n::Int, q::Int) = projective_omega_group(0, n, q)
 #
 ################################################################################
 
-"""
+@doc raw"""
     free_group(n::Int, s::VarName = :f; eltype::Symbol = :letter) -> FPGroup
-    free_group(L::Vector{<:VarName}) -> FPGroup
     free_group(L::VarName...) -> FPGroup
+    free_group(varnames_specifiers...) -> FPGroup
```
fingolfin:
```suggestion
    free_group(L::VarName... ; eltype::Symbol = :letter) -> FPGroup
    free_group(varnames_specifiers... ; eltype::Symbol = :letter) -> FPGroup
```

src/Groups/group_constructors.jl
fingolfin: This was back then agreed upon for the behavior of `@varnames_interface` as users from algebraic geometry wanted e.g. `@polynomial_ring(QQ, :x => (1:2,1:2))` to produce variables `x12` etc. not a matrix `x`.


Honestly I am not that happy with it anymore. I would prefer if we had
- `@polynomial_ring(QQ, "x" => (1:2,1:2))` defines in the current scope a 2x2 matrix `x`
- `@polynomial_ring(QQ, "x#" => (1:2,1:2))` defines the current scope four variables `x11,x12,x21,x22` 

Then hopefully usage would still be convenient enough for users from algebraic geometry? I'll ask @wdecker when I am back from vacation.

fingolfin: (In particular this is not something I can reasonably change in this PR)

ThomasBreuer (approved):
Nice.

There is a cross-reference from `free_group` to `@free_group`, shouldn't `@free_group` be in the manual?

PR #3318, opened 2024-02-05
Prepare for name printing changes in AA

lgoettgens:
Makes https://github.com/Nemocas/AbstractAlgebra.jl/pull/1594 non-breaking by using the interface instead of internals (see https://github.com/Nemocas/AbstractAlgebra.jl/pull/1594#issuecomment-1927623096)

### DIFF
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1290,7 +1290,7 @@ Z^2
 julia> L = homogeneous_component(S, [1, 1]);
 
 julia> L[1]
-homogeneous component of graded multivariate polynomial ring in 5 variables over QQ of degree [1 1]
+S_[1 1] of dim 6
 
 julia> FG = gens(L[1]);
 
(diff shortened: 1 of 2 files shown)


fingolfin (approved):

src/Rings/mpoly-graded.jl
```diff
@@ -1162,7 +1162,7 @@ base_ring(f::MPolyDecRingElem) = base_ring(forget_decoration(f))
 
 function show_homo_comp(io::IO, M)
   (W, d) = get_attribute(M, :data)
-  n = get_attribute(W, :name)
+  n = AbstractAlgebra.find_name(W)
   io = pretty(io)
   if n !== nothing
     print(io, LowercaseOff(), "$(n)_$(d.coeff) of dim $(dim(M))")
```
joschmitt, line 1168:
```suggestion
    print(io, LowercaseOff(), "$(n)_")
    print(IOContext(io, :supercompact => true), d)
    print(io, " of dim $(dim(M))")
```
Should work too and would be more consistent regarding commas in the degree.

lgoettgens: I would like this PR to get going fast. There will be a follow up once https://github.com/Nemocas/AbstractAlgebra.jl/pull/1594 is released anyway, where I will incorporate this change

lgoettgens:
@fingolfin can you merge this so I can get going with https://github.com/Nemocas/AbstractAlgebra.jl/pull/1594?

PR #3321, opened 2024-02-06
Adv/serialize gb

antonydellavecchia:
Adds serialization for IdealGens type, includes a test with a tiny groebner basis calculation.

### DIFF
--- a/src/Serialization/Rings.jl
+++ b/src/Serialization/Rings.jl
@@ -287,9 +287,16 @@ end
 
 @register_serialization_type MPolyIdeal uses_params
 @register_serialization_type LaurentMPolyIdeal uses_params
-const IdealUnionType = Union{MPolyIdeal, LaurentMPolyIdeal, FreeAssAlgIdeal}
 
-function save_type_params(s::SerializerState, x::T) where T <: IdealUnionType
+# we should avoid this list getting too long and find a
+# way to abstract saving params soon
+const IdealOrdUnionType = Union{MPolyIdeal,
+                                LaurentMPolyIdeal,
+                                FreeAssAlgIdeal,
+                                IdealGens,
+                                MonomialOrdering}
+
+function save_type_params(s::SerializerState, x::T) where T <: IdealOrdUnionType
   save_data_dict(s) do
     save_object(s, encode_type(T), :name)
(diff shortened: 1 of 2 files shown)


src/Serialization/Rings.jl
```diff
@@ -642,26 +678,23 @@ function load_object(s::DeserializerState, ::Type{MPolyQuoRing})
...
 
-function load_object(s::DeserializerState, ::Type{MonomialOrdering})
-  the_ring = load_typed_object(s, :ring)
-  the_ordering = load_typed_object(s, :internal_ordering)
-  result = MonomialOrdering(the_ring, the_ordering)
+function load_object(s::DeserializerState, ::Type{MonomialOrdering}, ring::MPolyRing)
+  ord = load_typed_object(s, :internal_ordering)
+  result = MonomialOrdering(ring, ord)
 
-  println("need to come back to this")
```
antonydellavecchia, line 662: @HechtiDerLachs  was this just a reminder to make sure we dealt with :is_total correctly?

HechtiDerLachs: This looks weird. But what I do remember is that we did not make a serious attempt at serializing all monomial orderings. I only did one, the default ordering for `MPolyQuoRing`s. That was all I needed at that point.

antonydellavecchia: true, it seems possible that we left this print in the wrong place. We currently can only do the 
Orderings.SymbOrdering,  I just added a comment on top of orderings serialization that is probablu more useful. Since the SymbOrderings are all I needed for some first examples of Ideal Gens and this loading of Monomial orderings is working for them. I think this is fine, essentially this might break when adding more orderings, but this should be easy to find where the fix needs to happen when faced with a backtrace

HechtiDerLachs: Depends who you have in mind. What I would probably do is to leave an error message along the lines "serialization of monomial orderings of type $(typeof(this_ordering)) not implemented; consult docs/src/how_to_do_my_serialization.md for more information."

antonydellavecchia: This is a good point, and reminds that i need to fix up the error messages for when types aren't implemented. I'll do this in the another pull request.

benlorenz (approved):

PR #3324, opened 2024-02-06
Added function to compute defining ideal of flag variety

danteluber:
This function computes the defining ideal of the flag variety $Fl(w,n)$, where $w = (d_1,...d_k)$, $d_j\leq n-1$, denotes the rank, and $n$ is the dimension of the ambient vector space.


src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+    T = [[(-1)^(_pluecker_sgn(z[1],z[2],t))*xdict[sort(union(z[1],t))]*xdict[setdiff(z[2],t)] 
+            for t in setdiff(z[2],z[1])] for z in X]
+    
+    genz = [sum(t) for t in T]
+    
+    gg = unique(x->x*sign(leading_coefficient(x)),[g for g in genz if !(g == 0)])
+        
+    return ideal(gg)
+    
+end
```
lkastner:
```suggestion
function flag_pluecker_ideal(dimensions::Vector{Int},n::Int)
    N = collect(1:n)
    L = reduce(union,[subsets(N,d) for d in dimensions])
    L1 = reduce(union,[subsets(N,d-1) for d in dimensions])
    L2 = reduce(union,[subsets(N,d+1) for d in dimensions])
    R,x = polynomial_ring(QQ,"x"=>L)
    xdict = Dict{Vector{Int}, MPolyElem}([L[i] => x[i] for i in 1:length(L)])
    X = [[a,b] for a in L1, b in L2 if length(b) - length(a)>=2] 
    T = [[(-1)^(_pluecker_sgn(z[1],z[2],t))*xdict[sort(union(z[1],t))]*xdict[setdiff(z[2],t)] 
            for t in setdiff(z[2],z[1])] for z in X]
    
    gens = [sum(t) for t in T]
    gg = unique(x->x*sign(leading_coefficient(x)),[g for g in gens if !(g == 0)])
    return ideal(gg)
end
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+
+
+
+
+_pluecker_sgn(a::Vector{Int},b::Vector{Int},t::Int) = count(z->z>t,a)+count(z->z<t,b)
+    
+@doc raw"""
+    flag_pluecker_ideal(dimensions::Vector{Int},n::Int)
+
+Returns generators of the defining ideal for the flag variety $\text{Fl}((d_1,\dots,d_k),n)$, where $(d_1,\dots,d_k) =$`dimensions` denotes the rank, with $d_j\leq n-1$. That is, the space of $k$-step flags of linear subspaces $V_1\subset\dots\subset V_k$ in $\mathbb{C}^n$, where $\text{dim}(V_j) = d_{j}$.You can obtain the generators for the \emph{complete flag variety}$ by taking `dimensions` $=(1,\dots,n-1)$ and `n`$=n$.
```
lkastner: Please wrap the line to not be so long.

lkastner (changes requested):

lgoettgens:
I think it would make sense to accept the field as an argument instead of always taking QQ.

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+julia> flag_pluecker_ideal([1,3],4)
+Ideal generated by
+  x[[1]]*x[[2, 3, 4]] - x[[2]]*x[[1, 3, 4]] + x[[3]]*x[[1, 2, 4]] - x[[4]]*x[[1, 2, 3]]
+```
+"""
+function flag_pluecker_ideal(dimensions::Vector{Int},n::Int)
+    
+    N = collect(1:n)
+    
+    L = reduce(union,[subsets(N,d) for d in dimensions])
```
fingolfin:
```suggestion
    L = reduce(union,[subsets(n,d) for d in dimensions])
```
is better (also below, and in the end you can get rid of `N`)

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+    
+    L1 = reduce(union,[subsets(N,d-1) for d in dimensions])
+    
+    L2 = reduce(union,[subsets(N,d+1) for d in dimensions])
+    
+    R,x = polynomial_ring(QQ,"x"=>L)
+    
+    xdict = Dict{Vector{Int}, MPolyElem}([L[i] => x[i] for i in 1:length(L)])
+    
+    pairs = [[a,b] for a in L1, b in L2]
```
fingolfin: Better use tuples instead of vectors here to avoid a ton of allocations
```suggestion
    pairs = [(a,b) for a in L1, b in L2]
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+  x[[1]]*x[[2, 3, 4]] - x[[2]]*x[[1, 3, 4]] + x[[3]]*x[[1, 2, 4]] - x[[4]]*x[[1, 2, 3]]
+```
+"""
+function flag_pluecker_ideal(dimensions::Vector{Int},n::Int)
+    
+    N = collect(1:n)
+    
+    L = reduce(union,[subsets(N,d) for d in dimensions])
+    
+    L1 = reduce(union,[subsets(N,d-1) for d in dimensions])
```
fingolfin: The `reduce(union` will only discard something if dimensions contains a duplicate entry, right?

So perhaps at the start do something like `dimensions = unique(dimensions)` and then here just `reduce(vcat` which avoid comparing a ton of vectors.

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+    
+    L2 = reduce(union,[subsets(N,d+1) for d in dimensions])
+    
+    R,x = polynomial_ring(QQ,"x"=>L)
+    
+    xdict = Dict{Vector{Int}, MPolyElem}([L[i] => x[i] for i in 1:length(L)])
+    
+    pairs = [[a,b] for a in L1, b in L2]
+    
+    X = [t for t in pairs if length(t[2])-length(t[1])>=2]
```
fingolfin: Maybe combine `pairs` and `X` like this (untested, may need some tweaking but I hope the idea is clear)
```suggestion
    X = [(a,b) for a in L1, b in L2 if length(b)-length(a)>=2]
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+    
+    T = [[(-1)^(_pluecker_sgn(z[1],z[2],t))*xdict[sort(union(z[1],t))]*xdict[setdiff(z[2],t)] 
+            for t in setdiff(z[2],z[1])] for z in X]
```
fingolfin:
```suggestion
    T = [[(-1)^(_pluecker_sgn(a,b,t))*xdict[sort(union(a,t))]*xdict[setdiff(b,t)] 
            for t in setdiff(b,a)] for (a,b) in X]
```
Even nicer would be if `_pluecker_sgn` just returned `+1` or `-1` (arguably that fits its name better, too)

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,68 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+    
+    genz = [sum(t) for t in T]
+    
+    gg = unique(x->x*sign(leading_coefficient(x)),[g for g in genz if !(g == 0)])
```
fingolfin:
```suggestion
    filter!(!iszero, genz)
    gg = unique(x->x*sign(leading_coefficient(x)), genz)
```

fingolfin (commented):
Some general Julia hints

lgoettgens:
Please don't re-format the whole file as that may lead to many conflicts with other pull requests. (and it makes it way harder to find the "real" changes)

src/exports.jl
```diff
@@ -55,6 +55,7 @@ export Edge
 export EmptyScheme
 export FPGroup
 export FPGroupElem
+export flag_pluecker_ideal
```
benlorenz: please move this to the lower case functions

lkastner:
> (and it makes it way harder to find the "real" changes)

You can actually hide whitespace changes in the `Files changed` view, the option is in the drop down menu with the settings wheel. Similarly `git blame` has a flag
```
    -w                    ignore whitespace differences
```
It is really quite annoying that we want devs to format their code properly, but refuse to format our code and thereby blocking them from using tools like `JuliaFormatter` that would make this easy. At the same time we block ourselves from having this tested automatically.

lgoettgens:
> It is really quite annoying that we want devs to format their code properly, but refuse to format our code and thereby blocking them from using tools like `JuliaFormatter` that would make this easy. At the same time we block ourselves from having this tested automatically.

I totally agree with you. But I think it should be managed by a central instance and applied at a defined time so that everyone with open changes can prepare for that.
Once we have formatted files, keeping the formatting is way easier than the current state of things.

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,72 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+# Example 2
+
+Flag variety $\text{Fl}((1,3),4,QQ)$.
+```jldoctest
+julia> flag_pluecker_ideal([1,3],4)
+Ideal generated by
+  x[[1]]*x[[2, 3, 4]] - x[[2]]*x[[1, 3, 4]] + x[[3]]*x[[1, 2, 4]] - x[[4]]*x[[1, 2, 3]]
+```
+"""
+function flag_pluecker_ideal(dimensions::Vector{Int}, n::Int, F::Field)
```
```suggestion
function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
```
i think most similar functions take the field as the first argument

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,72 @@ function Base.show(io::IO, I::MPolyIdeal)
     print(io, "Ideal ", _get_generators_string_one_line(I))
   end
 end
+
+
+_pluecker_sgn(a::Vector{Int}, b::Vector{Int}, t::Int)::Int =
+  (-1)^(count(z -> z > t, a) + count(z -> z < t, b))
+
+@doc raw"""
+    flag_pluecker_ideal(dimensions::Vector{Int},n::Int,F)
```
```suggestion
    flag_pluecker_ideal(dimensions::Vector{Int}, n::Int, F::Field)
```
type missing. or with the ordering change
```suggestion
    flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,72 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+  T = [
+    [
+      (_pluecker_sgn(a, b, t)) * xdict[sort(union(a, t))] * xdict[setdiff(b, t)] for
+      t in setdiff(b, a)
+    ] for (a, b) in X
+  ]
+
+  genz = [sum(t) for t in T]
+  filter!(!iszero, genz)
+  gg = unique(x -> x * sign(leading_coefficient(x)), genz)
```
```suggestion
  gg = unique(x -> divexact(x, canonical_unit(x)), genz)
```
I think you need something like this as not all fields define a `sign` function

lgoettgens (commented):
thanks for the formatting. some more comments for consistency with other parts of oscar

src/Rings/mpoly-ideals.jl
fingolfin: Good point!

`canonical_unit` is one of these functions which are in the AA manual but which don't have a docstring, so looking at `?canonical_unit` is not helpful or perhaps anti-helpful)

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,72 @@ function Base.show(io::IO, I::MPolyIdeal)
     print(io, "Ideal ", _get_generators_string_one_line(I))
   end
 end
+
+
+_pluecker_sgn(a::Vector{Int}, b::Vector{Int}, t::Int)::Int =
+  (-1)^(count(z -> z > t, a) + count(z -> z < t, b))
```
fingolfin:
```suggestion
  iseven(count(z -> z > t, a) + count(z -> z < t, b)) ? 1 : -1
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2003,3 +2003,72 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+  x[[1]]*x[[2, 3, 4]] - x[[2]]*x[[1, 3, 4]] + x[[3]]*x[[1, 2, 4]] - x[[4]]*x[[1, 2, 3]]
+  -x[[1, 4]]*x[[1, 2, 3]] + x[[1, 3]]*x[[1, 2, 4]] - x[[1, 2]]*x[[1, 3, 4]]
+  -x[[2, 4]]*x[[1, 2, 3]] + x[[2, 3]]*x[[1, 2, 4]] - x[[1, 2]]*x[[2, 3, 4]]
+  -x[[3, 4]]*x[[1, 2, 3]] - x[[1, 3]]*x[[2, 3, 4]] + x[[2, 3]]*x[[1, 3, 4]]
+  -x[[1, 4]]*x[[2, 3, 4]] + x[[2, 4]]*x[[1, 3, 4]] - x[[3, 4]]*x[[1, 2, 4]]
+
+```
+# Example 2
+
+Flag variety $\text{Fl}((1,3),4,QQ)$.
```
fingolfin: The conventions say that there should be a single section `Examples` with all examples under that. One could use subsections here, but I don't think it is necessary, something like that should do it;

```suggestion
# Examples

Here the result for the
complete flag variety $\text{Fl}((1,2,3),4,\mathbb{Q})$. 
```jldoctest
julia> flag_pluecker_ideal([1,2,3],4,QQ)
Ideal generated by
  x[[1]]*x[[3, 4]] - x[[3]]*x[[1, 4]] + x[[4]]*x[[1, 3]]
  -x[[1, 4]]*x[[2, 3]] + x[[2, 4]]*x[[1, 3]] - x[[3, 4]]*x[[1, 2]]
  x[[2]]*x[[3, 4]] - x[[3]]*x[[2, 4]] + x[[4]]*x[[2, 3]]
  x[[1]]*x[[2, 4]] - x[[2]]*x[[1, 4]] + x[[4]]*x[[1, 2]]
  x[[1]]*x[[2, 3]] - x[[2]]*x[[1, 3]] + x[[3]]*x[[1, 2]]
  x[[1]]*x[[2, 3, 4]] - x[[2]]*x[[1, 3, 4]] + x[[3]]*x[[1, 2, 4]] - x[[4]]*x[[1, 2, 3]]
  -x[[1, 4]]*x[[1, 2, 3]] + x[[1, 3]]*x[[1, 2, 4]] - x[[1, 2]]*x[[1, 3, 4]]
  -x[[2, 4]]*x[[1, 2, 3]] + x[[2, 3]]*x[[1, 2, 4]] - x[[1, 2]]*x[[2, 3, 4]]
  -x[[3, 4]]*x[[1, 2, 3]] - x[[1, 3]]*x[[2, 3, 4]] + x[[2, 3]]*x[[1, 3, 4]]
  -x[[1, 4]]*x[[2, 3, 4]] + x[[2, 4]]*x[[1, 3, 4]] - x[[3, 4]]*x[[1, 2, 4]]

For the flag variety $\text{Fl}((1,3),4,QQ)$ we get a single generator.
```

lgoettgens:
@danteluber Could you click on the "Resolve conversation" button of everything you already took care of? This makes it a bit easier for everyone to follow what's going on in this thread.

src/Rings/mpoly-ideals.jl
lgoettgens: That's true. But honestly I see no way that this will change as long as we have https://github.com/Nemocas/AbstractAlgebra.jl/issues/894, https://github.com/Nemocas/AbstractAlgebra.jl/issues/1166, https://github.com/Nemocas/AbstractAlgebra.jl/issues/1254, ...

Sequenzer (commented):
I have nothing to complain about, except that we might want to change this function so that it computes the minimal set of generators.

lkastner (approved):

PR #3327, opened 2024-02-07
Adapt to name printing changes in AA

lgoettgens:
Refactors everything about name printing to only use the interface discussed in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1594 (and no longer any internals).

This PR does not touch anything around Schemes, this will be postponed to the future since this is a bit more entangled.

No output is changed by this, it is a pure rewrite to use the correct interface without any semantical changes.


fingolfin (approved):

lgoettgens:
Bump. Can we get this merged?

PR #3329, opened 2024-02-07
CI: update to codecov/codecov-action@v4

fingolfin:
Unlike prior versions, this *requires* that CODECOV_TOKEN is provided (see [here](https://github.com/codecov/codecov-action?tab=readme-ov-file#v4-release)). In practice, already v3 has issues without it. I added this as an organization secret to the `oscar-system` and `nemocas` GitHub organizations a few days ago, so it should be safe to use here.

Before merging this, the output of the `codecov` step in the CI logs should be checked to verify it all went well.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -98,7 +98,7 @@ jobs:
       - name: "Upload coverage data to Codecov"
         if: matrix.julia-version == '1.10' && matrix.depwarn != 'depwarn=error'
         continue-on-error: true
-        uses: codecov/codecov-action@v3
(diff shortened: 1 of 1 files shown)


lgoettgens:
We can't tell if this works in general, since this PR is from a fork.
`The PR is happening in a forked repo. Using tokenless upload.` (https://github.com/oscar-system/Oscar.jl/actions/runs/7815127754/job/21317805219?pr=3329#step:10:28).

@fingolfin can you move your branch to the `oscar-system` namespace and create a new PR?

fingolfin:
Ouch, indeed

PR #3332, opened 2024-02-07
support `gen(G::GAPGroup, 0)`

ThomasBreuer:
@fieker had noticed that this is missing.

This was already supported for `FinGenAbGroup`.

(The delegations between methods are not the same: For `FinGenAbGroup`, there is special code for `getindex(A, i)`, and `gen(A, i)` calls `getindex(A, i)`.
For `GAPGroup`, there is special code for `gen(G, i)`, and the generic `getindex` method from AbstractAlgebra calls `gen(G, i)`.)

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -446,18 +446,29 @@ has_gens(G::GAPGroup) = GAP.Globals.HasGeneratorsOfGroup(G.X)::Bool
     gen(G::GAPGroup, i::Int)
 
 Return `one(G)` if `i == 0`,
-otherwise the `i`-th element of the vector `gens(G)`.
-This is equivalent to `G[i]`, and returns `gens(G)[i]`
+the `i`-th element of the vector `gens(G)` if `i` is positive,
+and the inverse of the `i`-th element of `gens(G)` if `i` is negative.
+
+For positive `i`, this is equivalent to `G[i]`, and returns `gens(G)[i]`
 but may be more efficient than the latter.
 
-An exception is thrown if `i` is negative or larger than the length of
-`gens(G)`.
+An exception is thrown if `abs(i)` is larger than the length of `gens(G)`.
+
+# Examples
+```jldoctest
+julia> g = symmetric_group(5);  gen(g, 1)
+(1,2,3,4,5)
+
+julia> g[-1]
+(1,5,4,3,2)
+```
 """
 function gen(G::GAPGroup, i::Int)
    i == 0 && return one(G)
    L = GAPWrap.GeneratorsOfGroup(G.X)::GapObj
-   @req 0 < i && i <= length(L) "i must be in the range 0:$(length(L))"
-   return group_element(G, L[i]::GapObj)
+   0 < i && i <= length(L) && return group_element(G, L[i]::GapObj)
+   i < 0 && -i <= length(L) && return group_element(G, inv(L[-i])::GapObj)
+   @req false "i must be in the range -$(length(L)):$(length(L))"
 end
 
 """
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -445,15 +445,18 @@ has_gens(G::GAPGroup) = GAP.Globals.HasGeneratorsOfGroup(G.X)::Bool
 """
     gen(G::GAPGroup, i::Int)
 
-Return the `i`-th element of the vector `gens(G)`.
+Return `one(G)` if `i == 0`,
+otherwise the `i`-th element of the vector `gens(G)`.
 This is equivalent to `G[i]`, and returns `gens(G)[i]`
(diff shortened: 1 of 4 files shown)


lgoettgens:
@fingolfin took some effort in https://github.com/oscar-system/Oscar.jl/pull/3260 to remove all of these `getindex(..., 0)` calls and methods. Why should we add it here again?

ThomasBreuer:
As I wrote: `FinGenAbGroup` has it, and @fieker wants it.

src/Groups/matrices/MatGrp.jl
```diff
@@ -498,8 +498,6 @@ function gens(G::MatrixGroup)
    return G.gens
 end
 
-gen(G::MatrixGroup, i::Int) = gens(G)[i]
```
fingolfin, line 508: I think we had that because the matrix groups cache the generators, and the generators here are more than just wrappers around the GAP group elements; they can also (optionally) contain Julia matrices (this should have been explained here in a comment...)

fingolfin:
There is a difference between `G[0]` and `gen(G,0)`, of course...

Overall I am not thrilled; I'd love to at least hear a rationale (and have this added to a code comment). After all, the identity element of a group is not a generator? And it seems really strange to write `gen(G,0)` instead of `one(G)`. For modules and rings we tried to get away from `R[0]` in favor of `R(0)` or `zero(R)`.

Maybe the idea is to have a common way to get the neutral element for additive and multiplicative groups, which right now have `zero(G)` and `one(G)` ? I'd be happy to add `neutral(G)` or so (unfortunately `identity` is already taken). 

But if the goal is to allow writing "generic" code, then don't we also need support for
- getting the group operation or alternatively for applying it generically
- computing inverses (`-g` vs. `g^-1`)
- computing powers / multiples
- ... more?

ThomasBreuer:
Concerning `one(G)` vs. `zero(G)` for multiplicative vs. additive groups:
There is (for quite some time) a comment in the code which asks whether we should simply add `one` for `FinGenAgGroup`. I stumbled over this question when I wrote some generic tests for both `GAPGroup`s and `FinGenAbGroup`s, but in the end there was no need for such a generic `one`.

fieker:
There are a couple of different "issues" here:
 - the original "complaint" (from me, privately to Thomas) was that `G[0]` gave a horrible error and backtrace. That should be fixed.
 - support for `G[0]` to give one(G) or not
 - add one(G) for abelian groups
 - add G[-1] as inv(G[1])
 
 Adding one(G) for abelian groups: be my guest or not. No strong opinion.
 adding G[-1] comes from the use of negative indices in the free groups all the time, G[1]^2 as a "word" will be s.th. like [-1, 2]. In order to use those words efficiently, all the code has to check if the index is negative to then use inv(G[-i]). This is annoying.
 Also for some (matrix) groups, the inverse is probably also stored and expensive to compute...
 
 If we add support for G[-1], then G[0] = one(G) would be natural.

fieker:
(I chatted with @fingolfin about this, but @ThomasBreuer  or @lgoettgens would like to comment as well (or anyone else)

fingolfin:
After the explanation by @fieker I am OK with adding `G[0]` *for groups* as well as `G[-i] = G[i]^-1`, these are then indeed convenient.

And yeah we should avoid ugly GAP backtraces as error.

src/Groups/matrices/MatGrp.jl
```diff
@@ -505,7 +505,15 @@ function gens(G::MatrixGroup)
...
+# Note that the `gen(G::GAPGroup, i::Int)` method cannot be used
+# for `MatrixGroup` because of the `:gens` attribute.
+function gen(G::MatrixGroup, i::Int)
+  i == 0 && return one(G)
+  L = gens(G)
+  0 < i && i <= length(L) && return L[i]
+  i < 0 && -i <= length(L) && return inv(L[-i])
```
fingolfin, line 514: I guess on the long run the idea would be to cache this inverse, esp. for matrix groups, where it can be expensive to compute. But that can come in a future PR.

fingolfin (approved):

src/Groups/GAPGroups.jl
```diff
@@ -445,16 +445,30 @@ has_gens(G::GAPGroup) = GAP.Globals.HasGeneratorsOfGroup(G.X)::Bool
...
+julia> g[-1]
+(1,5,4,3,2)
+```
 """
 function gen(G::GAPGroup, i::Int)
+   i == 0 && return one(G)
    L = GAPWrap.GeneratorsOfGroup(G.X)::GapObj
-   @assert length(L) >= i "The number of generators is lower than the given index"
-   return group_element(G, L[i]::GapObj)
+   0 < i && i <= length(L) && return group_element(G, L[i]::GapObj)
```
fieker, line 469: You can do 0<I<length without &&

PR #3333, opened 2024-02-07
Revised interface for homogenization (old UI still there)

JohnAAbbott:
Created new types **homogenizer** and **dehomogenizer;** these give a cleaner interface.


src/Rings/mpoly-graded.jl
```diff
@@ -1562,6 +1562,397 @@ end
...
+# This code delegates to _homogenization_via_saturation for non-positive gradings
+# ============================================
+
+# _sat_poly_by_var and kronecker_delta are internal auxiliary functions
+# This homogenization impl seems to be usefully faster with positive ZZ^m-grading for m > 1
+
+# Saturate a polynomial by a variable -- equiv to  f/gcd(f,h^deg(f))
+function _sat_poly_by_var(f,h)
+  # ASSUMES h is a variable i.e. monic poly of deg 1 with just 1 term
+  # Is there a better way to implement this?
```
joschmitt, line 1812: There is the function `remove`:
```
  remove(z::MPolyRingElem{T}, p::MPolyRingElem{T}) where {T <: RingElement}

  Compute the valuation of z at p, that is, the largest k such that p^k divides z. Additionally, z/p^k
  is returned as the second return. value.

  See also valuation, which only returns the valuation.
```
I don't know whether it is better though :)

JohnAAbbott: Thanks for letting me know.  I'll investigate.  It is probably better to use a pre-existing function.

JohnAAbbott: Ahh.  After a little testing, neither my implementation nor that of `remove` is well-suited to this use-case (though both are mathematically correct).

HechtiDerLachs (approved):

HechtiDerLachs:
checked by @wdecker

src/Rings/mpoly-graded.jl
```diff
@@ -1562,6 +1562,397 @@ end
 ### Homogenization and Dehomogenization
 ############################################################################
 
+## 2024-02-06  New UI to homogenization & dehomogenization
+## used partly the code from 2023-07 (see below)
+
+
+struct homogenizer
```
HechtiDerLachs, line 1569: This should probably be `Homogenizer`, i.e. camel case. Then `homogenizer` should be added later on as the user facing constructor.

src/Rings/mpoly-graded.jl
```diff
@@ -1562,6 +1562,397 @@ end
...
+  P::MPolyRing              # orignal poly ring (?not graded?)
+  P_homog::MPolyDecRing     # graded poly ring: same vars as P, plus g extra homogenizing vars
+  VarMap::Vector{Int}       # var[k] in P embeds to var[VarMap[k]] in P_homog
```
HechtiDerLachs, line 1572: Did you consider to store and use an actual ring map in the internals? Is there a reason not to do so?

src/Rings/mpoly-graded.jl
```diff
@@ -1562,6 +1562,397 @@ end
...
+# Use a homogenizer as a function/operator to actually homogenize
+function (H::homogenizer)(f::MPolyRingElem)
+  @req parent(f) == H.P  "Not in correct polynomial ring"
+  return Oscar._homogenization(f, H.P_homog, H.HVars[1])
+end
+
+function (H::homogenizer)(V::Vector{<:MPolyRingElem})
+  # Call below will check arg compatibility: if V is empty, there is nothing to check!
+  return H.(V)
+end
```
HechtiDerLachs, line 1703: This does not seem to be super necessary since the user can apply `H.(...)` to begin with. Unless (but the current implementation does not do so at the moment) there is some overhead computation which we can avoid by digesting the whole vector at once.

src/Rings/mpoly-graded.jl
```diff
@@ -1562,6 +1562,397 @@ end
...
+  #         return q;
+  #     end
+  #     f = q;
+  # end
+end
+
+# This function should be in a file of utilities (JAA thinks)
+function kronecker_delta(i::Int, j::Int)
+  return (i == j) ? 1 : 0
+end
```
HechtiDerLachs, line 1831: This looks a bit fishy. If this is used to create unit matrices, then it would probably be a better idea to call the constructor for an `n x n`-unit matrix directly. If this is used for selecting `continue` statements in nested loops, then one should probably restructure the loop... But maybe this is useful in the end! Just saying, it makes me suspicious.

src/Rings/mpoly-graded.jl
```diff
@@ -1562,6 +1562,397 @@ end
...
+
+# MAYBE is_positively_graded makes this fn obsolete -- !!NOT SURE!!
+# # Check that W defines a positive grading: namely
+# #     each col contains a non-zero entry, and
+# #     first non-zero going down the col is positive.
+# # Follows defn 4.2.4 from R+K vol 2.
+# function is_positive_grading_matrix(W::Union{ZZMatrix, Matrix{<:IntegerUnion}})
+#   nrows = size(W,1)
```
HechtiDerLachs, line 1839: Is this code left here for a purpose? Do we have another copy of it already in another file? 

If we want to delete it, please do so. Otherwise, it would be good to have a hint why this is kept here so that at a later point one can decide more easily whether this can finally go.

src/Rings/mpoly-graded.jl
```diff
@@ -1764,30 +2155,31 @@ end
...
-    extra_gens = filter(f -> (length(f) < 2*AveNumTerms), elements(GB))
-    G = vcat(G,  extra_gens)
-  end
-  return G
-end
+# 2024-02-06 COMMENTED OUT:  code uplicated above (around line 1700)
+# # Internal function: used only in _homogenization_via_saturation (immediately below)
+# # This function returns gens(I) and possibly some more "small" elements of I
+# # (obtained from 1 or more groebner bases of I).  If extra_gens_flag is true,
+# # we compute one more groebner bases of I in the hope that they contain some small elements.
```
HechtiDerLachs, line 2162: Can this also be deleted, then?

HechtiDerLachs (commented):
Thanks @JohnAAbbott !

We have merged the preliminary version already because of the urgent need for the functionality. 

However, I left some comments about what I think should eventually be cleaned up nevertheless in a follow-up pr.

src/Rings/mpoly-graded.jl
JohnAAbbott: I discussed this briefly yesterday with @wdecker and @jankoboehm .  We concluded that it is nicer to allow also `H(V)` so that Julia newcomers are not obliged to learn the broadcasting syntax (which appears to be a Julia/Matlab foible)

src/Rings/mpoly-graded.jl
JohnAAbbott: From looking at my code, it is used to create something called `Id` (presumably something like an identity matrix), but also to create `RevLexMat` (which is a reversed identity matrix multiplied by -1).  I think it would be nice for Oscar to offer a Kronecker delta function, since it is a well-known and widely-used mathematical concept.  No idea where it should be defined though.

src/Rings/mpoly-graded.jl
JohnAAbbott: There are two definitions of positively graded in the literature.  I must verify that using `is_positively_graded` rather than the commented out code does not adversely affect correctness.  I left the code there since some other urgently needed to use the new UI.

src/Rings/mpoly-graded.jl
JohnAAbbott: Probably.  As I wrote above, the main aim yesterday was to produce a _draft_ PR so that others could check what I've done.  I am well aware that considerable cleaning must still be done.

src/Rings/mpoly-graded.jl
JohnAAbbott: Good point!   Is there a particularly neat way to do this?  I presume the "internal" constructors must have names identical to that of the `struct`.

src/Rings/mpoly-graded.jl
thofma: I personally think this is not a good idea. We do not do this for anything else and adding this randomly does not help the user in the long run. Yes, it is a julia notation, but we are using julia, don't we?

src/Rings/mpoly-graded.jl
thofma: Should be parametrized also?

src/Rings/mpoly-graded.jl
JohnAAbbott: Makes little difference to me.  The Julia/Matlab notation is "cool" but also "alien" to anyone who is new to Julia, so steepens the initial learning curve.  It is also not clear to me how often someone might really want to do `H(V)`.  Of course there is also `[H(f)  for f in V]` which does not explicitly use broadcasting -- I presume the compiler will produce the same code as for broadcasting.  What do @wdecker and @jankoboehm say?

thofma: Edit: My previous choice of words was suboptimal. I think having a uniform experience is better than saving a `.` for some random function. It should not be suprising, that julia has some features that other languages don't have and vice versa. Hiding them behind some inconsistent user interface is in my opinion not a good way.

src/Rings/mpoly-graded.jl
JohnAAbbott: Please clarify about what should be parametrized and why?  I have not so much experience with parametrized `struct` in Julia.  Right the code seems to work OK, but maybe it would be better with a parameter?

src/Rings/mpoly-graded.jl
JohnAAbbott: I have written specific code, which is faster.

src/Rings/mpoly-graded.jl
JohnAAbbott: What is a ring map?  Maybe this can be done after the book deadline (since I don't think it will change the UI).

src/Rings/mpoly-graded.jl
HechtiDerLachs: It is true that the Kronecker delta is widely used in mathematics. However, the only use cases that come to my mind where one would like to use it in programming, are of the kind which I described above and call for different solutions in coding. I'm not sure we want to encourage people using such functions then. And one very practical way to nudge them in this direction is to not supply them with the function in the first place.

src/Rings/mpoly-graded.jl
HechtiDerLachs: @JohnAAbbott : I suggest to meet on gather. Then we can look into the parametrizing business together.

src/Rings/mpoly-graded.jl
JohnAAbbott: The function `is_positively_graded` is not sufficient for my case, so I must use my own "more restrictive" code.

src/Rings/mpoly-graded.jl
JohnAAbbott: I have deleted some cruft.

JohnAAbbott:
Further updates are now in PR #3344

PR #3340, opened 2024-02-08
adds fix for storing strings with characters that need escaping

antonydellavecchia:
Fixes issues storing strings with newline characters and so on

### DIFF
--- a/src/Serialization/serializers.jl
+++ b/src/Serialization/serializers.jl
@@ -112,7 +112,8 @@ function save_data_basic(s::SerializerState, x::Any,
   !isnothing(key) && set_key(s, key)
   begin_node(s)
   str = string(x)
-  write(s.io, "\"$str\"")
+  #write(s.io, "\"$str\"")
+  JSON.print(s.io, str)
 end
 
 function save_data_json(s::SerializerState, jsonstr::Any,
--- a/src/Serialization/serializers.jl
+++ b/src/Serialization/serializers.jl
@@ -112,7 +112,8 @@ function save_data_basic(s::SerializerState, x::Any,
   !isnothing(key) && set_key(s, key)
   begin_node(s)
   str = string(x)
-  JSON.print(s.io, str)
+  JSON.show_string(s.io, str)
+  nothing
 end
 
 function save_data_json(s::SerializerState, jsonstr::Any,
--- a/src/Serialization/serializers.jl
+++ b/src/Serialization/serializers.jl
@@ -112,7 +112,6 @@ function save_data_basic(s::SerializerState, x::Any,
   !isnothing(key) && set_key(s, key)
   begin_node(s)
   str = string(x)
-  #write(s.io, "\"$str\"")
   JSON.print(s.io, str)
 end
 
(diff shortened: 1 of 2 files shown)


src/Serialization/serializers.jl
```diff
@@ -112,7 +112,8 @@ function save_data_basic(s::SerializerState, x::Any,
   !isnothing(key) && set_key(s, key)
   begin_node(s)
   str = string(x)
-  write(s.io, "\"$str\"")
+  #write(s.io, "\"$str\"")
```
benlorenz:
```suggestion
```

src/Serialization/serializers.jl
```diff
@@ -112,7 +112,8 @@ function save_data_basic(s::SerializerState, x::Any,
...
+  JSON.print(s.io, str)
```
benlorenz:
```suggestion
  JSON.show_string(s.io, str)
  nothing
```
This would avoid a few allocations and should be faster.

antonydellavecchia: Why add nothing on 117?

benlorenz: Because `show_string` always returns 1, the return value of `write(io, "\"")`. And we don't want a return value? (Maybe it won't hurt but I think it is better to discard it directly here instead of assuming that this happens further down in the call stack)

benlorenz:
```
julia> str = "sjdlkdsfjslkdjflew kewj dsf sdf\n\nsd lkdsjfldsfsdf\"sdjlkfdsf\ndsfdsf";

julia> @benchmark JSON.show_string(devnull, str)
BenchmarkTools.Trial: 10000 samples with 989 evaluations.
 Range (min … max):  43.068 ns … 74.669 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     51.306 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   51.370 ns ±  1.652 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                   ▁▄▇█▇▇▆▆▇▆▄▃▂▁              
  ▂▁▁▁▁▁▂▁▂▂▂▂▂▂▂▂▂▂▂▃▂▂▂▃▃▃▃▃▃▃▄▅▆███████████████▇▇▆▅▅▄▄▃▃▂▃ ▄
  43.1 ns         Histogram: frequency by time        55.2 ns <

julia> @benchmark JSON.print(devnull, str)
BenchmarkTools.Trial: 10000 samples with 741 evaluations.
 Range (min … max):  168.926 ns …   3.081 μs  ┊ GC (min … max): 0.00% … 92.51%
 Time  (median):     189.308 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   207.036 ns ± 203.506 ns  ┊ GC (mean ± σ):  8.08% ±  7.59%

                ▁▃▇█▅▅▅▃▂                                        
  ▁▁▁▁▁▁▂▁▁▄▇▅▅▆██████████▅▄▃▃▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  169 ns           Histogram: frequency by time          236 ns <

 Memory estimate: 432 bytes, allocs estimate: 5.
```

benlorenz (approved):

PR #3341, opened 2024-02-08
add `map_entries` for matrix groups

ThomasBreuer:


src/Groups/matrices/MatGrp.jl
```diff
@@ -547,6 +547,53 @@ function order(::Type{T}, G::MatrixGroup) where T <: IntegerUnion
...
+  over integer ring
+
+julia> is_finite(G2)
+false
+
+julia> order(map_entries(GF(3), G))
+3
+```
+"""
+function map_entries(R::Ring, G::MatrixGroup)
```
joschmitt, line 580: There is also `change_base_ring` for this case, which does the same. Maybe one should call the other?

thofma: I think one could just have one method of the form
```
map_entries(f, G::MatrixGroup)
```
without specifying the type of the first argument. This is how all other `map_*` methods usually are implemented.

src/Groups/matrices/MatGrp.jl
```diff
@@ -547,6 +547,53 @@ function order(::Type{T}, G::MatrixGroup) where T <: IntegerUnion
...
+
+function map_entries(mp::Map, G::MatrixGroup)
+  imgs = [map_entries(mp, matrix(x)) for x in gens(G)]
+  return matrix_group(codomain(mp), degree(G), imgs)
+end
+
+function map_entries(f::Function, G::MatrixGroup)
+  Ggens = gens(G)
+  if length(Ggens) == 0
+    o = map_entries(f, matrix(one(G)))
```
fingolfin: Wouldn't it suffice to map a single entry, like
```suggestion
    o = f(zero(base_ring(G)))
```

src/Groups/matrices/MatGrp.jl
```diff
@@ -547,6 +547,53 @@ function order(::Type{T}, G::MatrixGroup) where T <: IntegerUnion
...
+  imgs = [map_entries(R, matrix(x)) for x in gens(G)]
+  return matrix_group(R, degree(G), imgs)
+end
+
+function map_entries(mp::Map, G::MatrixGroup)
+  imgs = [map_entries(mp, matrix(x)) for x in gens(G)]
+  return matrix_group(codomain(mp), degree(G), imgs)
+end
+
+function map_entries(f::Function, G::MatrixGroup)
```
fingolfin: Maybe just drop the `::Function` to make it more generic "for free" ? (There are other "callable" objects which are not functions, after all)
```suggestion
function map_entries(f, G::MatrixGroup)
```

src/Groups/matrices/MatGrp.jl
fingolfin: @joschmitt `change_base_ring` seems to only cover the case where the first argument is a ring, right? Of course we can make sure one is implemented in terms of the other. The current method is
```julia
function change_base_ring(R::Ring, G::MatrixGroup)
  g = dense_matrix_type(R)[]
  for h in gens(G)
    push!(g, map_entries(R, h.elm))
  end
  return matrix_group(g)
end
```
which does not deal with empty generator lists. We could just change it to
```julia
change_base_ring(R::Ring, G::MatrixGroup) = map_entries(R, G)
```

@thofma those other `map_entries` methods don't need to deal with an empty list of matrices, though... But yeah one could merge the methods though one could consider the two simple methods as "optimizations" of the generic `Function` one? Is there any harm in having these additional methods?

thofma: No, you are right. Maybe my suggestion is just to remove the restriction on the first argument and shuffle the docstring, so that it is not at the "wrong" method (at the moment the signature of the docstring and the function are not the same).

joschmitt: > We could just change it to
> 
> ```julia
> change_base_ring(R::Ring, G::MatrixGroup) = map_entries(R, G)
> ```

Yes, exactly.

src/Groups/matrices/MatGrp.jl
```diff
@@ -547,6 +547,53 @@ function order(::Type{T}, G::MatrixGroup) where T <: IntegerUnion
...
+  Ggens = gens(G)
+  if length(Ggens) == 0
+    o = map_entries(f, matrix(one(G)))
+    return matrix_group(base_ring(o), degree(G), typeof(o)[])
```
fingolfin: After discussing this: I overlooked the use of `typeof(o)` here. But we can avoid it:
```suggestion
    return matrix_group(base_ring(o), degree(G), MatrixGroupElem[])
```

fingolfin:
@ThomasBreuer it shouldn't be much work to finish this up, I hope? or are there blockers / decisions to be made?

It would be nice to have this merged so I can use it.

fingolfin (approved):
Thanks!

fingolfin:
Huh, doctest failures in `src/Rings/groebner.jl:1022-1042` but I don't understand how this PR could cause them?

joschmitt:
> Huh, doctest failures in `src/Rings/groebner.jl:1022-1042` but I don't understand how this PR could cause them?

See https://github.com/oscar-system/Oscar.jl/pull/3377 .

lgoettgens:
This should get backported, right?

lgoettgens:
> This should get backported, right?

Bump @fingolfin @ThomasBreuer

PR #3270, opened 2024-01-30
Adapt unicode printing of exterior powers of Lie algebra modules

lgoettgens:
Resolves the Lie algebra part of https://github.com/oscar-system/Oscar.jl/issues/3257.

PR #3342, opened 2024-02-09
Experimental: Complex reflection groups

ulthiel:
This is work in progress implementing complex reflection groups. Contributions welcome.


experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,678 @@
...
+        if isa(t, Int)
+            i = t
+            if i == 4
+                K,z = cyclotomic_field(3)
+            elseif i == 5
+                K,z = cyclotomic_field(3)
+            elseif i == 6
+                K,z = cyclotomic_field(12)
+            elseif i == 7
+                K,z = K,z = cyclotomic_field(12)
```
joschmitt:
```suggestion
                K,z = cyclotomic_field(12)
```
Also some times more in the following lines.

experimental/ComplexReflectionGroups/src/unitary_matrices.jl
```diff
@@ -0,0 +1,51 @@
...
+end
+
+function is_unitary(G::MatrixGroup{T}) where T <: Union{Hecke.RelSimpleNumFieldElem, AbsSimpleNumFieldElem, AbsNonSimpleNumFieldElem}
+    for g in gens(G)
+        if is_unitary(g) == false
+            return false
+        end
+    end
+    return true
+end
```
fingolfin:
```suggestion
end

```

experimental/ComplexReflectionGroups/src/unitary_matrices.jl
```diff
@@ -0,0 +1,51 @@
...
+    return is_unitary(matrix(M))
+end
+
+function is_unitary(G::MatrixGroup{T}) where T <: Union{Hecke.RelSimpleNumFieldElem, AbsSimpleNumFieldElem, AbsNonSimpleNumFieldElem}
+    for g in gens(G)
+        if is_unitary(g) == false
+            return false
+        end
+    end
+    return true
```
fingolfin:
```suggestion
    return all(is_unitary, gens(G))
```

experimental/ComplexReflectionGroups/src/unitary_matrices.jl
```diff
@@ -0,0 +1,51 @@
...
+    if !is_square(M)
+        return false
+    end
+
+    # convert base field to absolute and convert matrix as well
+    K = base_ring(M)
+    n = ncols(M)
+    L, f = absolute_simple_field(K)
+    g = inv(f)
+    ML = matrix(L, n, n, [g(x) for x in M])
```
fingolfin: Hm, does this also work?
```suggestion
    ML = map_entries(g, M)
```

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,678 @@
...
+# 
+# * Thiel, U. (2014). On restricted rational Cherednik algebras. TU Kaiserslautern.
+#
+# * Geck, M., & Malle, G. (2006). Reflection groups. In Handbook of algebra. Vol. 4 (Vol. 4, pp. 337–383)
+#
+# Ulrich Thiel, 2023 
+
+export ComplexReflectionGroupType
+
+export complex_reflection_group_type
```
fingolfin: Please sort this list of exports alphabetically

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,678 @@
...
+export is_irreducible
+export is_exceptional
+export order
+export components
+export num_components
+export is_imprimitive
+export is_primitive
+export num_reflections
```
fingolfin: See also the (ongoing...) discussion about unifying the "number" functions. The current two proposals:
1. always have a function `number_of_FOO` available; optionally one can have a short alias of the form `nFOO`, such as `nrefl` or `nreflections` (we mostly implemented this, a few things are not yet renamed)
2. always (?) use `n_` as prefix, so `n_reflections` (presumably there will be either some exceptions or the possibility for alias

Note that for the second proposal, I am guessing a bit, and also may have misunderstood it; I do not mean to mischaracterize it, but I have simply not yet seen a fully fledged out version of it anywhere, presumably we'll be told what it is by @micjoswig, @lkastner et. al. during the meeting this Friday.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,678 @@
...
+function Base.show(io::IO, ::MIME"text/plain", G::ComplexReflectionGroupType)
+    print(io, "Complex reflection group type ")
+    counter = 0
+    N = length(G.type)
+    if N == 0
+        print(io,"trivial")
+    end
+    for t in G.type
+        print(io,"G")
+        if isa(t, Int) == 1
```
fingolfin:
```suggestion
        if isa(t, Int)
```
or
```suggestion
        if t isa Int
```

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,678 @@
...
+
+    if length(fields) == 1
+        return fields[1]
+    else
+        println(fields)
+        # From Lars Göttgens
+        return reduce(compositum, fields; init=rationals_as_number_field()[1])[1]
+    end
+
+end
```
fingolfin: All files should end in a newline?

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,678 @@
...
+            m = ZZ(t[1])
+            p = ZZ(t[2])
+            n = ZZ(t[3])
+        end
+    end
+
+    if length(fields) == 1
+        return fields[1]
+    else
+        println(fields)
```
fingolfin: Debugging leftover?
```suggestion
```

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
```diff
@@ -0,0 +1,1729 @@
...
+
+            # G4
+            if t == 4
+
+                if Cmodel == :LT
+                    # See Lehrer & Taylor (2009), page 85-86
+                    R,x = polynomial_ring(QQ)
+                    K,i = number_field(x^2+1, "i")
+                    R,x = polynomial_ring(K)
+                    K,ω = number_field(x^2+x+1, "ω")
```
fingolfin: We have a general rule against unicode inside code (it is OK in comments, though), see https://docs.oscar-system.org/stable/DeveloperDocumentation/styleguide/#Unicode

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
```diff
@@ -0,0 +1,1729 @@
...
+                    i = K(i)
+                    matspace = matrix_space(K, 2, 2)
+                    gens = elem_type(matspace)[]
+
+                    r1 = ω//2 * matspace(K[-1-i 1-i ; -1-i -1+i])
+                    r1prime = ω//2 * matspace(K[-1-i -1+i ; 1+i -1+i])
+
+                    push!(gens, r1)
+                    push!(gens, r1prime)
```
fingolfin: Alternatively, do this:
```suggestion
                    gens = [
                       ω//2 * matrix(K, [-1-i 1-i ; -1-i -1+i]),  # r1
                       ω//2 * matrix(K, [-1-i -1+i ; 1+i -1+i])   # r1'
                    ]
```

fingolfin (commented):
Thanks that's a great contribution!

A few quick comments

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: Cheers, I'll take care of this in my next coding session.

ulthiel: Note to self: transpose matrices for the unitary models by LT because of x*A action in OSCAR.

Edit: Done.

lgoettgens:
Could you click "resolve conversation" on all comments above once you addressed them? This makes it way easier to keep an overview here.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
micjoswig: Indeed, I am vastly in favor of `n_reflections` etc.  It is concise and systematic.

ulthiel:
> Could you click "resolve conversation" on all comments above once you addressed them? This makes it way easier to keep an overview here.

Yes, once addressed... ;-)

thofma:
If you change the Hecke version in the file `Oscar/src/Project.toml` to `0.30.2`, you should be able to get rid of `complex_conjugation2`.

ulthiel:
> If you change the Hecke version in the file `Oscar/src/Project.toml` to `0.30.2`, you should be able to get rid of `complex_conjugation2`.

Excellent. I will when I'm done (ETA next week).

ulthiel:
> If you change the Hecke version in the file `Oscar/src/Project.toml` to `0.30.2`, you should be able to get rid of `complex_conjugation2`.

Done in [efbb6e5](https://github.com/oscar-system/Oscar.jl/pull/3342/commits/efbb6e539f6bb7adc6f90fd6f035083bfbe4a3cb).

ulthiel:
Can someone (@joschmitt @lgoettgens) help me with the failing tests? This seems to be a problem with experimental/LieAlgebras and the function coroot, which I have implemented (with a different signature) as well. 

The error is "MethodError: no method matching coroot(::RootSystem, ::Int64)"

joschmitt:
> Can someone (@joschmitt @lgoettgens) help me with the failing tests? This seems to be a problem with experimental/LieAlgebras and the function coroot, which I have implemented (with a different signature) as well.
> 
> The error is "MethodError: no method matching coroot(::RootSystem, ::Int64)"

Adding
```
import Oscar.LieAlgebras: coroot
```
in `ComplexReflection.jl` seems to work. (Don't ask me why; this import/export nonsense is always educated guessing at best for me.)

ulthiel:
Thanks, let's see if it works.

joschmitt:
The "long tests" seem to fail because you export `is_exceptional` and `is_shephard_group` without having them?

ulthiel:
Ok, checks passed finally. I will need to document the functions, add further documentation, and address the stuff from the top of this conversation. Then it should be OK for a first set of features for complex reflection groups (this can never be finished...).

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: I really would like to print an ω. No way around this? Do need I need to print it as "omega"? This looks ugly as hell :-(

lgoettgens: There is a user-settable flag for this. In the default case, use `omega` (or `w`), but if a user opts-in to unicode, you can print `ω`. See https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/#Unicode-printing for some documentation and e.g. https://github.com/oscar-system/Oscar.jl/pull/3270 for a real-world example of this

ulthiel: And then how do I enter ω in my code when I am not supposed to use unicode in code?

ulthiel: And how do I do this in this specific case for the generator of my field? (Only for this I used unicode).

lgoettgens: In these branches dependent on `Oscar.is_unicode_allowed()` it is fine to use unicode. See e.g. https://github.com/oscar-system/Oscar.jl/pull/3270/files#diff-11c0a2ef064e801ca05579645558af1993e8decdbbfeb2694996be0c11289fc1R926

experimental/ComplexReflectionGroups/src/complex_reflection_group_LT.jl
```diff
@@ -0,0 +1,501 @@
...
+# Models from Lehrer & Taylor (2009)
+###########################################################################################
+function complex_reflection_group_LT(n::Int)
+
+    if n == 4
+        # See Lehrer & Taylor (2009), page 85-86
+        R,x = polynomial_ring(QQ)
+        K,i = number_field(x^2+1, "i")
+        R,x = polynomial_ring(K)
+        K,ω = number_field(x^2+x+1, "ω") #ω is primitive 3rd root of unity
```
lgoettgens:
```suggestion
        K, omega = number_field(x^2+x+1, Oscar.is_unicode_allowed() ? "ω" : "omega") #ω is primitive 3rd root of unity
```
e.g. like this

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
lgoettgens: https://github.com/oscar-system/Oscar.jl/pull/3342#discussion_r1516157357

docs/oscar_references.bib
```diff
@@ -89,6 +89,20 @@ @Article{BBS02
   doi           = {10.1007/s00454-001-0051-x}
 }
 
+@InCollection{BCP97,
+  author        = {Bosma, Wieb and Cannon, John and Playoust, Catherine},
+  title         = {The {M}agma algebra system. {I}. {T}he user language},
```
lgoettgens:
```suggestion
  title         = {The {M}agma algebra system. {I}. {T}he user language},
```
our bibtex parser doesn't like braces in the title. They can safely be removed as it keeps casing.

docs/oscar_references.bib
```diff
@@ -305,6 +319,21 @@ @Article{BS09
...
+  title         = {Symplectic singularities},
+  mrnumber      = {1738060},
+  journal       = {Invent. Math.},
+  fjournal      = {Inventiones Mathematicae},
+  volume        = {139},
+  number        = {3},
+  pages         = {541--549},
+  year          = {2000},
+  doi           = {10.1007/s002229900043},
+  file          = {:Beauville. Symplectic singularities.pdf:PDF},
```
lgoettgens:
```suggestion
```
this file entry is nonsense

docs/oscar_references.bib
```diff
@@ -687,6 +729,20 @@ @Article{EFS03
   zbmath        = {1963988}
 }
 
+@Article{EG02,
+  author        = {Etingof, Pavel and Ginzburg, Victor},
+  title         = {Symplectic reflection algebras, {C}alogero-{M}oser space, and deformed {H}arish-{C}handra
```
lgoettgens: same as earlier (no braces in title)

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: Thanks! One question: is there anyone on this planet (or universe) who is *not* allowing unicode? 

Let's make this a rhetorical question. I fear there will be (and were) some discussions...

joschmitt: > In these branches dependent on `Oscar.is_unicode_allowed()` it is fine to use unicode.

Didn't know that. I messed around with human unreadable `\u????` codes in the Young tableaux printing.

lgoettgens: > > In these branches dependent on `Oscar.is_unicode_allowed()` it is fine to use unicode.
> 
> Didn't know that. I messed around with human unreadable `\u????` codes in the Young tableaux printing.

don't quote me on that :joy:, but when I did it, nobody objected. I think it depends on how weird the symbols are and if most fonts can display them

docs/oscar_references.bib
```diff
@@ -1802,6 +1892,18 @@ @Article{SS12
   doi           = {10.1016/j.jcta.2011.12.005}
 }
 
+@Article{ST54,
+  author        = {Shephard, G. C. and Todd, J. A.},
+  title         = {Finite unitary reflection groups},
+  mrnumber      = {59914},
+  journal       = {Canad. J. Math.},
+  fjournal      = {Canadian Journal of Mathematics. Journal Canadien de Math\'{e}matiques},
```
lgoettgens: please use unicode instead of escapes. And overall, have a look at https://docs.oscar-system.org/dev/DeveloperDocumentation/documentation/#Updating-the-bibliography please

experimental/ComplexReflectionGroups/README.md
```diff
@@ -0,0 +1,36 @@
+# Complex Reflection Groups
+
+## Aim 
+The aim is to implement [complex reflection groups](https://en.wikipedia.org/wiki/Complex_reflection_group) along with data that is known or can be computed.
+
+By [Ulrich Thiel](https://ulthiel.com/math), 2023.
```
lgoettgens, line 6: There is some common structure to the README.md files of other experimental folders. Maybe make this consistent? but no strong opinion here.

experimental/ComplexReflectionGroups/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Example
+
+```
```
lgoettgens, line 10:
```suggestion
```jldoctest
```
to keep it tested automatically

experimental/ComplexReflectionGroups/docs/doc.main
```diff
@@ -0,0 +1,6 @@
+[
+   "Complex Reflection Groups" => [
+      "introduction.md",
+      "complex_reflection_groups.md",
+   ],
+]
```
lgoettgens:
```suggestion
]

```
ALL files should end with a newline

experimental/ComplexReflectionGroups/docs/src/complex_reflection_groups.md
```diff
@@ -0,0 +1,209 @@
...
+As a demonstration, we create the exceptional complex reflection group $G_{4}$ in the
+notation of Shepard & Todd [ST54](@cite). We choose as *model*—meaning a concrete matrix
+group—the unitary model from [LT09](@cite). The base field of this model is
+$\mathbb{Q}(i,\omega)$ where $i$ is the imaginary unity and $\omega = \exp(2\pi i/3)$. The
+group is generated by two unitary reflections. We display the first generator, verify that
+it is indeed a unitary reflection, and then we compute additional data like a root, the
+hyperplane, and the non-trivial eigenvalue of this reflection. We have implemented an own
+type encapsulating such data of a reflection.
+
+```@juliarepl
```
lgoettgens:
```suggestion
```jldoctest
```
all of these should be doctests (unless they need an eternity to run), so that one gets errors once they get out of date

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
```diff
@@ -0,0 +1,344 @@
+export ComplexReflection
```
lgoettgens: it has been seen in the past that a single list of imports and exports in the `ComplexReflectionGroups.jl` file is the easiest for future reasoning about them. could you do the same here and collect all imports and exports to there?

experimental/ComplexReflectionGroups/src/hermitian_things.jl
```diff
@@ -0,0 +1,81 @@
...
+# Complex conjugation
+###########################################################################################
+# In Hecke v0.30.2 complex_conjugation was extended to more fields following my suggestion.
+# I will only do one minor extension following Tommy Hofmanns suggestion:
+complex_conjugation(K::QQAbField) = conj
+
+###########################################################################################
+# Hermitian scalar product
+###########################################################################################
+function scalar_product(v::AbstractAlgebra.Generic.FreeModuleElem{T}, w::AbstractAlgebra.Generic.FreeModuleElem{T}) where T <: QQAlgFieldElem
```
lgoettgens, line 19: In my opinion, this shouldn't be needed as there already is `dot` for a dot-product. But unfortunately, that does not work for non-real fields currently (see https://github.com/Nemocas/Nemo.jl/issues/1694), so this might only be possible to change in the later future.

experimental/ComplexReflectionGroups/src/hermitian_things.jl
```diff
@@ -0,0 +1,81 @@
...
+
+    V = parent(v)
+    K = base_ring(V)
+    n = dim(V)
+    s = zero(K)
+    conj = complex_conjugation(K)
+    for i=1:n
+        s += v[i]*conj(w[i])
+    end
+    return s
```
lgoettgens:
```suggestion
    @req parent(v) === parent(w) "Incompatible vector spaces"
    conj = complex_conjugation(K)
    return sum(vi * conj(wi) for (vi, wi) in zip(v, w); init=zero(base_ring(v)))
```
parent check missing.
furthermore some suggestion for cleanup

lgoettgens (commented):
some random comments and nitpickts

experimental/ComplexReflectionGroups/docs/src/complex_reflection_groups.md
```diff
@@ -0,0 +1,209 @@
...
+$$\mathrm{Ker}(1-g)^\perp = \mathrm{Im}(1-g)$$
+
+so that we have a $g$-stable orthogonal decomposition
+
+$$V = \mathrm{Ker}(1-g) \oplus \mathrm{Im}(1-g)^\perp \;.$$
+
+Since $\mathrm{Ker}(1-g)$ is a hyperplane, the complement $\mathrm{Im}(1-g)$ is a line. A
+non-zero element $\alpha$ on this line is called a **root** of $g$. A root is unique up to
+multiplication with a non-zero scalar. We then have $g \alpha = \zeta \alpha$ for some
+primitive $m$-throot of unity $\zeta$, where $m$ is the order of $g$.
```
joschmitt:
```suggestion
primitive $m$-th root of unity $\zeta$, where $m$ is the order of $g$.
```

experimental/ComplexReflectionGroups/docs/src/complex_reflection_groups.md
```diff
@@ -0,0 +1,209 @@
...
+
+julia> fundamental_invariants(IR)
+2-element Vector{MPolyDecRingElem{Hecke.RelSimpleNumFieldElem{AbsSimpleNumFieldElem}, AbstractAlgebra.Generic.MPoly{Hecke.RelSimpleNumFieldElem{AbsSimpleNumFieldElem}}}}:
+ x[1]^4 + (-4*ω - 2)*x[1]^2*x[2]^2 + x[2]^4
+ x[1]^5*x[2] - x[1]*x[2]^5
+```
+
+One feature of our implementation is that we explicitly support *different models*. The
+reason is that different models may have different properties. For example, note that in the
+above example the first fundamental invariant has a coefficient in which $\omega$ occurrs,
```
joschmitt:
```suggestion
above example the first fundamental invariant has a coefficient in which $\omega$ occurs,
```

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
```diff
@@ -0,0 +1,344 @@
...
+    K = base_ring(V)
+    n = dim(V)
+    basis = gens(V)
+
+    if root == 0
+        throw(ArgumentError("root needs to be nonzero."))
+    end
+
+    if coroot == 0
+        throw(ArgumentError("root needs to be nonzero."))
```
joschmitt:
```suggestion
        throw(ArgumentError("coroot needs to be nonzero."))
```

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
```diff
@@ -0,0 +1,344 @@
...
+function is_complex_reflection_with_data(w::MatrixGroupElem{T}) where T <: QQAlgFieldElem
+
+    G = parent(w)
+
+    # If G has reflections assigned (see function below), try to find w in the list and 
+    # return the data.
+    if has_attribute(G, :complex_reflections)
+        refls = collect(get_attribute(G, :complex_reflections))
+        i = findfirst(g->matrix(g)==matrix(w), refls)
+        if i != nothing
```
joschmitt: I've been told that one should always compare with nothing using `===` or `!==`.

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
```diff
@@ -0,0 +1,344 @@
...
+    b, w_data = is_complex_reflection_with_data(w)
+    return b
+end
+
+function complex_reflection(w::MatrixGroupElem{T}) where T <: QQAlgFieldElem
+
+    b,w_data = is_complex_reflection_with_data(w)
+
+    if !b
+        throw(ArgumentError("Group element is not a reflection"))
```
joschmitt: I was informed that we still have one foot in the real world and say "pseudo-reflection" in general (like Bourbaki!). But no argument from my side :)

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
```diff
@@ -0,0 +1,344 @@
...
+###########################################################################################
+# Determining all reflections of a matrix group.
+###########################################################################################
+function complex_reflections(G::MatrixGroup{T}) where T <: QQAlgFieldElem
+   
+    if has_attribute(G, :complex_reflections)
+        return get_attribute(G, :complex_reflections)
+    end
+
+    refls = Set(ComplexReflection[])
```
joschmitt: Should this really be a `Set`? Mathematically it is, sure. But I think we usually just use `Vector` in these occasions because they are more convenient. (For example, there is `getindex` for `Vector` but not for `Set`.)

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+
+    # Argument checking Cannot be more specific than Vector for the constructor argument
+    # since [ 4, (4,1,4) ] is of type Vector{Any}
+    function ComplexReflectionGroupType(type::Vector)
+
+        # The normalized type with unique labels (see below)
+        typenorm = Vector{Union{Int, Tuple{Int,Int,Int}}}()
+
+        for t in type
+            @req isa(t, Int) || isa(t, Tuple{Int,Int,Int}) "Type must be a vector of integers (for exceptional groups) or of 3-element vectors (for the infinite series)"
```
joschmitt:
```suggestion
            @req isa(t, Int) || isa(t, Tuple{Int,Int,Int}) "Type must be a vector of integers (for exceptional groups) or of 3-element tuples (for the infinite series)"
```

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+        if counter < N
+            print(io, " x ")
+        end
+    end
+end
+
+# Convenience constructors
+ComplexReflectionGroupType(i::Int) = ComplexReflectionGroupType(convert(Vector{Union{Int, Tuple{Int,Int,Int}}}, [i]))
+ComplexReflectionGroupType(m::Int, p::Int, n::Int) = ComplexReflectionGroupType(convert(Vector{Union{Int, Tuple{Int,Int,Int}}}, [(m,p,n)]))
+ComplexReflectionGroupType(t::Tuple{Int,Int,Int}) = ComplexReflectionGroupType(convert(Vector{Union{Int, Tuple{Int,Int,Int}}}, [t]))
```
joschmitt, line 94: I think those explicit type conversions aren't necessary since the actual constructor is for `<:Vector` anyway?

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+function num_components(G::ComplexReflectionGroupType)
+    return length(G.type)
+end
+
+function is_irreducible(G::ComplexReflectionGroupType)
+    if length(G.type) == 1
+        return true
+    else
+        return false
+    end
```
joschmitt:
```suggestion
    return length(G.type) == 1
```

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+        [ 6, 12, 18, 24, 30, 42 ],
+        [ 2, 5, 6, 8, 9, 12 ],
+        [ 2, 6, 8, 10, 12, 14, 18 ],
+        [ 2, 8, 12, 14, 18, 20, 24, 30 ]
+    ]
+
+    for t in G.type
+        if isa(t, Int)
+            i = t
+            append!(degrees, convert(Vector{ZZRingElem}, excdegrees[i-3]) )
```
joschmitt: This conversion is useless because `degrees` is initialized as a `Vector{Int}`, so everything gets converted back (also below).

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+        [ 0, 12, 18, 24, 30, 36 ], #34
+        [ 0, 3, 4, 6, 7, 10 ], #35
+        [ 0, 4, 6, 8, 10, 12, 16 ], #36
+        [ 0, 6, 10, 12, 16, 18, 22, 28 ] #37
+    ]
+
+    for t in G.type
+        if isa(t, Int)
+            i = t
+            append!(codegrees, convert(Vector{ZZRingElem}, exccodegrees[i-3]) )
```
joschmitt: Same as for `degrees`: the conversion doesn't do anything this way.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+        end
+    end
+
+    return true
+
+end
+
+function is_coxeter_group(G::ComplexReflectionGroupType)
+    return is_real(G)
+end
```
joschmitt: Maybe use `@alias` here, then there needs to be only one docstring and test.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
```diff
@@ -0,0 +1,680 @@
...
+            end
+        end
+    end
+
+    return true
+end
+
+function is_weyl_group(G::ComplexReflectionGroupType)
+    return is_rational(G)
+end
```
joschmitt:
```suggestion
@alias is_weyl_group is_rational
```
(as above for `is_coxeter_group`)

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
fingolfin: Some people will always object, but I at least have done it myself and consciously accepted it in PRs.

@ulthiel &lt;sarcasm>no, nobody has objections to (or experiences issues due to) the use of unicode always and everywhere, we only have this settings because we were bored and wanted to spice up our lives 😈 &lt;/sarcasm> (to be clear, I've been fine using Unicode for everything and everywhere about 2001, thank you Mac OS X)

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
```diff
@@ -0,0 +1,182 @@
...
+
+
+
+function complex_reflection_group_type(G::MatrixGroup)
+    if has_attribute(G, :complex_reflection_group_type)
+        return get_attribute(G, :complex_reflection_group_type)
+    end
+    return nothing
+    # this should be upgraded later to work with a general matrix group (identifying the
+    # type from scratch is not so easy though)
```
joschmitt: Maybe throw an error until then instead of returning `nothing`.

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
```diff
@@ -0,0 +1,182 @@
...
+end
+
+function complex_reflection_group_model(G::MatrixGroup)
+    if has_attribute(G, :complex_reflection_group_model)
+        return get_attribute(G, :complex_reflection_group_model)
+    end
+    return nothing
```
joschmitt: As above: throw an error.

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
```diff
@@ -0,0 +1,182 @@
...
+
+    return WD
+
+end
+
+
+###########################################################################################
+# Checking if a matrix group is a complex reflection group.
+###########################################################################################
+function is_complex_reflection_group(G::MatrixGroup{T}) where T <: QQAlgFieldElem
```
joschmitt, line 132: This functions only checks whether the chosen generators of the group are reflections, but not whether the group is a reflection group. Or am I missing something?

experimental/ComplexReflectionGroups/src/complex_reflection_group_CHEVIE.jl
```diff
@@ -0,0 +1,609 @@
...
+###########################################################################################
+function complex_reflection_group_CHEVIE(n::Int)
+
+    if n == 4
+        K,z_3 = cyclotomic_field(3)
+        matspace = matrix_space(K, 2, 2)
+        gens = elem_type(matspace)[]
+
+        s1 = matspace(K[1 0 ; 0 z_3])
+        s2 = matspace(K[1//3*(2*z_3 + 1) 1//3*(2*z_3 - 2) ; 1//3*(z_3 - 1) 1//3*(z_3 + 2)])
```
joschmitt:
```suggestion
        gens = dense_matrix_type(K)[]

        s1 = K[1 0 ; 0 z_3]
        s2 = K[1//3*(2*z_3 + 1) 1//3*(2*z_3 - 2) ; 1//3*(z_3 - 1) 1//3*(z_3 + 2)]
```
Using the `matrix_space` is redundant here; `K[1 0 ; 0 1]` already gives an OSCAR matrix and the field knows about the dense matrix type itself. More cases below, I'm afraid.

experimental/ComplexReflectionGroups/src/complex_reflection_group_CHEVIE.jl
```diff
@@ -0,0 +1,609 @@
...
+        s3 = matspace(K[-1 0 0 ; 0 1 0 ; 0 0 1])
+                            
+        push!(gens, s1)
+        push!(gens, s2)
+        push!(gens, s3)
+
+        W = matrix_group(gens)
+
+    elseif n == 28
+        K = QQ
```
joschmitt: There is the issue with type stability here (and in some similar functions). The return type can be either `MatrixGroup{QQFieldElem, QQMatrix}` or this but with `AbsSimpleNumFieldElem` etc. I don't see a way to avoid it, but wanted to mention it in case somebody has a bright idea.

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
```diff
@@ -0,0 +1,182 @@
...
+###########################################################################################
+# Important remark: In OSCAR, matrices act by default from the right on vectors, so x*A. 
+# For example the kernel of a matrix A is the space of all vectors such that x*A=0 
+# (confusingly, this is usually referred to a s the *left kernel*....). The same holds for
+# Magma and also CHEVIE.
+# 
+# This means that when we take matrices for the models from the literature in which 
+# matrices act from the left (the usual convention for non-computer stuff), like the book
+# by Lehrer & Taylor, we need to transpose these matrices! This is why down there in the
+# code there are several transpose operation for the final generators.
```
joschmitt: I think this comment is not in the correct file any more.

experimental/ComplexReflectionGroups/src/hermitian_things.jl
```diff
@@ -0,0 +1,81 @@
...
+    return is_unitary(matrix(M))
+end
+
+function is_unitary(G::MatrixGroup{T}) where T <: QQAlgFieldElem
+    for g in gens(G)
+        if is_unitary(g) == false
+            return false
+        end
+    end
+    return true
```
joschmitt:
```suggestion
    return all(is_unitary, gens(G))
```

experimental/ComplexReflectionGroups/src/symplectic_reflection_group.jl
```diff
@@ -0,0 +1,38 @@
...
+    set_attribute!(W_symp, :is_symplectic_reflection_group, true)
+
+    return W_symp
+end
+
+function is_symplectic_reflection_group(G::MatrixGroup)
+    if has_attribute(G, :is_symplectic_reflection_group)
+        return get_attribute(G, :is_symplectic_reflection_group)
+    end
+    return false
```
joschmitt: Throw an error.

experimental/ComplexReflectionGroups/src/symplectic_reflection_group.jl
```diff
@@ -0,0 +1,38 @@
...
+#
+# Ulrich Thiel, 2024
+
+export symplectic_reflection_group
+export is_symplectic_reflection_group
+
+function symplectic_reflection_group(W::MatrixGroup)
+
+    if !is_complex_reflection_group(W)
+        throw(ArgumentError("Group must be a complex reflection group"))
```
joschmitt: For what this function is doing, not really... (somebody should invent a good name for this 'doubling operation').

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
lgoettgens: I would disagree as there are other dispatches for `is_rational` (like `is_rational(x::QQBarFieldElem)`) that would not make sense for `is_weyl_group`. `@alias` should only be used if all methods make sense for both names.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
lgoettgens: No, see https://github.com/oscar-system/Oscar.jl/pull/3342#discussion_r1516419404

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: I can't withstand: I like the is_unicode option. But then there's still unicode in the code, which will still be problematic for people using something like ed on a PDP-7. So, this is somewhat inconsistent, no?!

Anyways, if it's officially okay to put unicode in the code when wrapped around is_unicode as suggested by @lgoettgens above, I'll go with that because I like my ω and τ. 

Stuff like that seems to be more complicated than the mathematics... ;-)

experimental/ComplexReflectionGroups/src/hermitian_things.jl
ulthiel: No, and for exactly the reason you say: I need to involve complex conjugation on the field.

ulthiel:
Okay thanks all. I'll need a bit to address everything...

experimental/ComplexReflectionGroups/src/symplectic_reflection_group.jl
ulthiel: It will make sense in the greater scheme: symplectic_reflection_group(...) will create a symplectic reflection group for a labeling system you will come up with. And then this function will be the convenience constructor for the one cooked up from a complex reflection group...Something like that...

experimental/ComplexReflectionGroups/src/complex_reflection_group_CHEVIE.jl
ulthiel: Doesn't this religious hunt for type stability apply only for performance-critical functions? Who cares for this function?

experimental/ComplexReflectionGroups/src/symplectic_reflection_group.jl
ulthiel: Btw I always called this 'doubling operation' 'symplectic doubling' but it seems no one (in particular you) likes it...

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
ulthiel: I wonder who told you that...

experimental/ComplexReflectionGroups/src/complex_reflection_group_CHEVIE.jl
joschmitt: Yeah, and as I said, I don't see how to avoid it. QQ is a number field mathematically but not by its type.

benlorenz: Maybe use `rationals_as_number_field()` ?
```
julia> K,_ = rationals_as_number_field()
(Number field of degree 1 over QQ, 1)

julia> typeof(K)
AbsSimpleNumField
```
It is also nicer for consistency if someone tries to use the output as input for some other functions.

joschmitt: I just wanted to check how the performance compares, but it turns out that GAP doesn't like `rationals_as_number_field` as base field of a matrix group. So don't use it right now. I'll look into it...

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
ulthiel: There seems to be both `number_of` and `n_ functions` currently in OSCAR, roughly in equal numbers (I would find my `num_` to be a nice intermediate thing...). Which one should I use now?

joschmitt: There is supposed to be an explanation about this in the styleguide which is waiting for ... something, see https://github.com/oscar-system/Oscar.jl/pull/3305 .

ulthiel: What's the tl;dr?

lgoettgens: Current state: Use either `n_` or `number_of_` and create an alias (using `@alias`) for the other ones. (still waiting for the approval of some PIs)

ulthiel: Could you give me the precise code for doing an alias (haven't done this before)? And where should I put the alias? Below the function?

experimental/ComplexReflectionGroups/src/complex_reflection_group_CHEVIE.jl
joschmitt: I opened a pull request #3501 making `rationals_as_number_field` work with GAP. For GAP, there is now no difference between `QQ` or `rationals_as_number_field` on the OSCAR side. Still, there will be some performance penalty, if one works with the matrices themselves. What do you think @thofma ?

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
ulthiel: Function commented out for the moment. Will remove when I finish the function (one day).

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
lgoettgens: Say you rename `num_reflection_classes` to `n_reflection_classes`, then you would additionally add `@alias number_of_reflection_classes n_reflection_classes` to underneath the now accumulated list of exports. 
And the same for the other functions.

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: Let's not discuss further, I'm implementing as suggested by @lgoettgens.

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: I'll leave it as is since this is somehow too much changing with no real effect.

ulthiel:
What is this failing invalidations?

lgoettgens:
> What is this failing invalidations?

If the difference is less than 5 (see e.g. https://github.com/oscar-system/Oscar.jl/actions/runs/8202710398 for the current report), this can be ignored and will usually go away when restarting enough times.

docs/oscar_references.bib
ulthiel: Wait, unicode??? What happened?

lgoettgens: DocumenterCitations.jl (our bibtex Parser) supports Unicode better than the escapes

experimental/ComplexReflectionGroups/README.md
ulthiel: Since I have docs, this README is also somewhat unnecessary I find.

experimental/ComplexReflectionGroups/src/hermitian_things.jl
ulthiel: Ok for the parent check. Otherwise, I find this much more complicated to read. Is this supposed to be faster?

experimental/ComplexReflectionGroups/docs/src/complex_reflection_groups.md
ulthiel: I hate this word!

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
ulthiel: I will stick to complex reflection in the current context. Even though I like Bourbaki, I don't like the word pseudo-reflection.

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
ulthiel: I was not sure really. I later want to define maps on the set of reflections. Can I do this if this is a vector? But yeah, for easy access vectors are better. Opinions?

experimental/ComplexReflectionGroups/src/ComplexReflectionGroupType.jl
ulthiel: I cannot remember. I think it didn't work without this, not sure anymore.

experimental/ComplexReflectionGroups/src/hermitian_things.jl
lgoettgens: It may be, dependent on the underlying implementation. I particular, in your iterated addition, for every addition a new variable containing the result needs to be allocated

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
lgoettgens: In this use-case of maps, vectors are even better, since you have a first map, second map etc. (even though it's only internal). So your maps may just be permutations

experimental/ComplexReflectionGroups/docs/src/complex_reflection_groups.md
```diff
@@ -0,0 +1,209 @@
...
+As a demonstration, we create the exceptional complex reflection group $G_{4}$ in the
+notation of Shepard & Todd [ST54](@cite). We choose as *model*—meaning a concrete matrix
+group—the unitary model from [LT09](@cite). The base field of this model is
+$\mathbb{Q}(i,\omega)$ where $i$ is the imaginary unity and $\omega = \exp(2\pi i/3)$. The
+group is generated by two unitary reflections. We display the first generator, verify that
+it is indeed a unitary reflection, and then we compute additional data like a root, the
+hyperplane, and the non-trivial eigenvalue of this reflection. We have implemented an own
+type encapsulating such data of a reflection.
+
+```@jldoctest
```
lgoettgens: Those should be without the `@` iirc
```suggestion
```jldoctest
```

experimental/ComplexReflectionGroups/src/hermitian_things.jl
ulthiel: Hm, interesting. Is this written/explained somewhere? Given that `scalar_product` may indeed be evaluated many times, I'll do your suggested change.

experimental/ComplexReflectionGroups/src/complex_reflection_group_CHEVIE.jl
ulthiel: I will ignore this for now, you can have a go once it's sorted out.

experimental/ComplexReflectionGroups/src/hermitian_things.jl
ulthiel: Your code does not work:

```
julia> V = vector_space(QQ,2)
v =Vector space of dimension 2 over rational field

julia> v = V([1,1])
(1, 1)

julia> K = base_ring(v)
Rational field

julia> conj = complex_conjugation(K)
Identity map
  of rational field

julia> sum(vi * conj(wi) for (vi, wi) in zip(v, v); init=zero(K))
ERROR: MethodError: no method matching iterate(::AbstractAlgebra.Generic.FreeModuleElem{QQFieldElem})
```

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: Yes, indeed, and this will be remarked in the documentation. For the general thing you'd have to check through all subsets of reflection whether they generate and this can be a lot of work. Also, for computer algebra, it should be the correct notion that the fixed generators are reflections. As I said, this will be remarked in the documentation and if someone wants to implement the general thing it will be another function.

Edit: Changed my mind, see below.

thofma: Hm, I am not sure this is very OSCAR like, where we try to stay as close to the mathematics without too much surprising behaviour for a non-expert. Being a complex reflection group is a mathematical concept independent of the chosen generators. The natural name for checking this of an arbitrary matrix group would be `is_complex_reflection_group`, but this would be already taken by this function here.

A similar situation would be ideals, which for OSCAR always come with a fixed family/set of generators. But `is_principal(I)` for such an ideal `I` will check whether it can be generated by one element and not if `ngens(I) == 1` holds.

Maybe a suggestion would be to rename this into something internal, maybe put an `_` in front of it?

lgoettgens: In similar situations we discussed using `is_created_as_*` for this exact use-case. But we postponed it and made the function internal with a leading underscore for the time being

fieker: Why not name it generated by reflections? And reserve the other one for the mathematical test?

ulthiel: Alright alright alright alright alright alright, you'll get your honest `is_complex_reflection_group`. It's more consistent than what I did but I would never admit this here. Also, it's actually not so bad to check as I thought: just take the set of all reflections and see if the subgroup generated by this is equal to the group. Only finding all reflections could be optimized (as commented in my code) by using conjugacy classes. But I'll leave that to whoever is interested. :-)

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
ulthiel: Will turn into vector.

experimental/ComplexReflectionGroups/src/complex_reflection_group.jl
ulthiel: Done in [d611586](https://github.com/oscar-system/Oscar.jl/pull/3342/commits/d6115864ea13756ba06124f31d8e875eceefb059). I will expand the ComplexReflection structure a bit to save the parent in case of a group element (or just GL if it's a matrix) so that this avoids the G(matrix(w)), which may involve computations which are actually not necessary.

In the tests I've added a fun example: a matrix group generated by two non-reflections. But these generate G4, so this is a complex reflection group.

experimental/ComplexReflectionGroups/src/ComplexReflection.jl
```diff
@@ -180,24 +180,37 @@ function is_complex_reflection_with_data(w::MatElem{T}) where T <: QQAlgFieldEle
   H, Hincl = kernel(f)
 
   if dim(H) != n-1
+    if debug
+      println("Fixed space is not a hyperplane")
```
joschmitt, line 184: Just to say: there is the `@vprintln` macro for optional verbose output. From what I can see, the advantage is that you don't need to hand `debug` statements around if the function is called several levels down, the disadvantage that you need to add a 'verbosity scope' (see the documentation if you are interested).

experimental/ComplexReflectionGroups/src/hermitian_things.jl
ulthiel: @lgoettgens ?

lgoettgens: ok, so apparently vector space elements only support getindex but not iteration. In this case, a workaround here would be `sum(v[i] * conj(w[i]) for i in 1:dim(parent(v)); init=zero(base_ring(v)))`. This is now basically your initial version but using `sum` instead of iterated addition.

ulthiel: Good, thanks.

fingolfin:
This is still marked as "draft". @ulthiel what is missing from your POV before this can be merged?

I worry a bit about drift happening to this PR (already there is a conflict, though it should be trivial to resolve). I would recommend to merge this rather earlier than later; since it is in experimental, it would still be fine to perform major rewrites later. But once merged at least we can take care of adjusting the code added here as needed to changes in AA/Nemo/Hecke/Julia, and can ensure its tests keep passing, etc.


That said, in the end you decided of course.

experimental/Experimental.jl
```diff
@@ -8,6 +8,7 @@ const expdir = joinpath(@__DIR__, "../experimental")
 const orderedpkgs = [
   "LieAlgebras",
   "BasisLieHighestWeight",   # needs code from LieAlgebras
+  "ComplexReflectionGroups"  # Because LieAlgebra implements coroot
```
joschmitt:
```suggestion
  "ComplexReflectionGroups",  # Because LieAlgebra implements coroot
```

ulthiel:
> This is still marked as "draft". @ulthiel what is missing from your POV before this can be merged?
> 
> I worry a bit about drift happening to this PR (already there is a conflict, though it should be trivial to resolve). I would recommend to merge this rather earlier than later; since it is in experimental, it would still be fine to perform major rewrites later. But once merged at least we can take care of adjusting the code added here as needed to changes in AA/Nemo/Hecke/Julia, and can ensure its tests keep passing, etc.
> 
> That said, in the end you decided of course.

Will continue now.

lgoettgens:
This PR now contains a lot of other changes not related to complex reflection groups. Could you try to rebase this onto a current master?

ulthiel:
> This PR now contains a lot of other changes not related to complex reflection groups. Could you try to rebase this onto a current master?

I do not know what this means: 1) changes not related to complex reflection groups? which? 2) rebase? what do I need to do?

lgoettgens:
The commit 6d76ba2ce246d5cafc97a8551ea96f3bedd1fe44 contains other changes.
I think you would need to drop that and instead do a proper merge/rebase. I.e. after dropping that commit do:
```
git checkout yourbranch
git rebase upstream/master
```
And then follow the instructions in the console to resolve conflicts given by typing `git status`

ulthiel:
I have no idea what all these changes are, I didn't change anything. I tried that rebase but it resulted in an error stopping somewhere in the middle of my changes in February. Umpf.

lgoettgens:
I could try to perform the rebase for you right now, if you don't touch the branch/do any commits or pushes/... until I write you that it's finished. Should I give it a try?

ulthiel:
Give me one last try. I didn't know about rebase --continue. Git noob.

ulthiel:
Ok, is it better now? I have no idea what I just did. I had to resolve conflicts in Experimental.jl and oscar_references.bib several times. My tests passed and I could build the documentation without errors.

ulthiel:
What I still want to do before a merge is the following:

- [ ] Finish up documentation
- [ ] Function to determine conjugacy classes of reflections and hyperplane orbits, and extend the ComplexReflection type to carry a reference to the conjugacy class and hyperplane orbit. I need to think about how to do this nicely. I need this for Calogero-Moser spaces and Cherednik algebras. Since this may change ComplexReflection a bit, I first want to figure this out before merge.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroups.jl
```diff
@@ -0,0 +1,79 @@
+
+# A union type for all algebraic extensions of the field of rational numbers
+QQAlgField = Union{NumField, QQField, QQBarField, QQAbField}
+QQAlgFieldElem = Union{NumFieldElem, QQFieldElem, QQBarFieldElem, QQAbElem}
```
fingolfin: These must be `const`, otherwise any code using them is potentially type unstable.
```suggestion
const QQAlgField = Union{NumField, QQField, QQBarField, QQAbField}
const QQAlgFieldElem = Union{NumFieldElem, QQFieldElem, QQBarFieldElem, QQAbElem}
```

experimental/ComplexReflectionGroups/src/hermitian_things.jl
```diff
@@ -0,0 +1,80 @@
...
+
+  if !is_square(M)
+    return false
+  end
+
+  K = base_ring(M)
+  n = ncols(M)
+
+  return M*transpose(M) == identity_matrix(K,n)
+end
```
fingolfin:
```suggestion
is_orthogonal(M::MatElem) = is_square(M) && is_one(M*transpose(M))
```

experimental/ComplexReflectionGroups/src/hermitian_things.jl
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  # create the conjugate transpose of M
+  K = base_ring(M)
+  n = ncols(M)
+
+  conj = complex_conjugation(K)
+  Mct = transpose(map_entries(conj, M))
+
+  return M*Mct == identity_matrix(K,n)
```
fingolfin:
```suggestion
  return is_one(M*Mct)
```

experimental/ComplexReflectionGroups/src/hermitian_things.jl
fingolfin: First off, to avoid misunderstandings, let me emphasize that I am OK with merging this as it is for now.

But I still would like to clear this up on the long run. Perhaps it is already clear for @thofma and @fieker but I am not familiar with the background, so I am asking / pointing out the following:

First off, `dot` as defined in Julia, does involve complex conjugation:

> Compute the dot product between two vectors. For complex vectors, the first vector is conjugated.

However, it doesn't work on `AbstractAlgebra.Generic.FreeModuleElem`. For starters iteration over these is not implemented. Nor is `conj` implemented for most of our types. Which makes me wonder: why not? We could just add
```julia
Base.conj(x::ZZRingElem) = x
Base.conj(x::ZZPolyRing) = x
Base.conj(x::ZZMPolyRing) = x

Base.conj(x::QQFieldElem) = x
Base.conj(x::QQPolyRing) = x
Base.conj(x::QQMPolyRing) = x

...
```

thofma: `dot` implies conjugation on the first vector, where as @ulthiel wants it on the second vector. We use the same convention as @ulthiel for `inner_product`.

I think adding `conj` for ring elements, which are contained in the complex numbers is fine. Although I am not sure I want to extend it automatically to polynomials.

experimental/ComplexReflectionGroups/src/ComplexReflectionGroups.jl
```diff
@@ -0,0 +1,79 @@
+
+# A union type for all algebraic extensions of the field of rational numbers
+const QQAlgField = Union{NumField, QQField, QQBarField, QQAbField}
+const QQAlgFieldElem = Union{NumFieldElem, QQFieldElem, QQBarFieldElem, QQAbElem}
+
+# Imports (for stuff from experimental)
+import Oscar.LieAlgebras: coroot #no conflict, just same function name
```
lgoettgens, line 7:
```suggestion
```
should no longer be needed after https://github.com/oscar-system/Oscar.jl/pull/4399

experimental/Experimental.jl
```diff
@@ -8,6 +8,7 @@ const expdir = joinpath(@__DIR__, "../experimental")
 const orderedpkgs = [
   "LieAlgebras",
   "BasisLieHighestWeight",   # needs code from LieAlgebras
+  "ComplexReflectionGroups",  # Because LieAlgebra implements coroot
```
lgoettgens, line 11:
```suggestion
```
should no longer be needed after https://github.com/oscar-system/Oscar.jl/pull/4399

experimental/ComplexReflectionGroups/src/ComplexReflectionGroups.jl
```diff
@@ -0,0 +1,79 @@
+
+# A union type for all algebraic extensions of the field of rational numbers
+const QQAlgField = Union{NumField, QQField, QQBarField, QQAbField}
+const QQAlgFieldElem = Union{NumFieldElem, QQFieldElem, QQBarFieldElem, QQAbElem}
```
lgoettgens, line 4:
```suggestion
const QQAlgFieldElem = Union{NumFieldElem, QQFieldElem, QQBarFieldElem, QQAbFieldElem}
```
has been renamed at some point

experimental/ComplexReflectionGroups/src/ComplexReflectionGroups.jl
```diff
@@ -0,0 +1,79 @@
...
+include("ComplexReflectionGroupType.jl")
+
+include("complex_reflection_group.jl")
+include("complex_reflection_group_LT.jl")
+include("complex_reflection_group_Magma.jl")
+include("complex_reflection_group_CHEVIE.jl")
+
+include("symplectic_reflection_group.jl")
+
+# Exports
```
lgoettgens, line 25: one generally wants to put the exports above the include statements. We had some very weird issues regarding that a few weeks ago (just with julia nightly, but anyway)

lgoettgens:
This PR contains a lot of great code and IMO it would be great to have it available (even though there may be large changes again in the future). Furthermore, once it is merged, it will be kept up to date by all the maintainers; instead of bitrotting (there are conflicts again, and there are some obsolete things, mentions of renamed types etc.).

@ulthiel wrote in https://github.com/oscar-system/Oscar.jl/pull/3342#issuecomment-2225155481 some bullet-points they want to work on before merging, but that was more than a year ago without any change.

If you (@ulthiel) agree, I could hijack your branch and attempt to do the few maintanence tasks that are required to have this running again, and then we could proceed with merging (if nobody else voices objections). The changes you mentioned you wanted to work on can just go into a follow-up PR; code in `experimental` does not have to be perfect in the first PR, and you are free to still do major changes later. However, this is ultimately your decision.

ulthiel:
I need someone to discuss (make great suggestions, or simply do it) on how to properly handle conjugacy classes of reflections of a matrix group and constructing some sort of "reflection library" for a matrix group: each reflection should know its conjugacy class (of reflections) and its reflection hyperplane orbit etc. I had something like this in [CHAMP](https://github.com/ulthiel/Champ/blob/6a4d0e60fc5f5157f4851ed6dbdcb3e22d1b4d04/Source/ReflectionGroups/Reflections.i.m#L414) but I was thinking I could do better. This stuff is really important to work with Cherednik algebras etc. Well and then I wasn't sure how to do it properly---and here we are. I really want to have this merged as well but somehow I got stuck with this.

Edit: And I wasn't sure whether the final approach would change the Reflection type, so I did not want to have this merged before figuring this out. Ugh.

fingolfin:
Well, since this code would be placed into "experimental", it would be OK to change the `Reflection` type later on. I mean, realistically, it could become a problem if someone started to heavily use that code. But if that happens, I'll thank Murphy's law for the exciting news of this heavy user being around, and we'd work with them to make the necessary changes. So to me this would be lucky.

More likely is that nobody will care either way.

Hence I'd be tempted to chance Murphy's law striking, as it's a win-win either way.

Regarding

> how to properly handle conjugacy classes of reflections of a matrix group and constructing some sort of "reflection library" for a matrix group:

Perhaps we can discuss that with @ulthiel @lgoettgens myself and others at the Jahrestagung?

PR #3343, opened 2024-02-09
More updates for FTheoryTools

HereAround:
* Put model, defined over arbitrary base, over a concrete base. For simplicity, this currently only works for Tate and Weierstrass models. (Potentially my code can also be optimized - it appears rather lengthy, but I could not yet see a good way to make it more concise.)
* Remove unused variable "Kbar".
* Construct polynomial rings with `cached = false`.

cc @apturner


HechtiDerLachs (approved):
This looks very promising! Do you plan to add a more detailed documentation on the interfaces that this is using? For example, it seems to rely on a command `generic_section` which takes a line bundle as input. For me it would be interesting to know what functionality I need to implement so that this works over a (non-toric) scheme.

Concerning the code: It looks like one could make good use of method dispatch in several places with `if ... isa ...`. But either way, this looks like it could work for the moment. 

Tests are still missing at this point.

HereAround:
> This looks very promising! Do you plan to add a more detailed documentation on the interfaces that this is using? For example, it seems to rely on a command `generic_section` which takes a line bundle as input. For me it would be interesting to know what functionality I need to implement so that this works over a (non-toric) scheme.
> 
> Concerning the code: It looks like one could make good use of method dispatch in several places with `if ... isa ...`. But either way, this looks like it could work for the moment.
> 
> Tests are still missing at this point.

Thank you @HechtiDerLachs , only just saw your comment after I pushed a few more changes here.

Regarding your questions:
* The documentation can of course always be extended. Still, I do not have specific plans regarding the documentation of the `put_over_concrete_base` method. What is missing from your perspective?
* The generic_section docu is here: https://docs.oscar-system.org/dev/AlgebraicGeometry/ToricVarieties/ToricLineBundles/#generic_section-Tuple%7BToricLineBundle%7D
* Regarding pushing "this" to non-toric schemes. The question is what "this" refers to exactly - generic sections of a line bundle or elliptic fibrations? Or yet something else? Either way, you must be able to get a basis of the sections of said line bundle. And this may get tricky quickly. (Aside: Similar discussions for what @apturner and I will have to do with our family of basis. Maybe turns into an abstract variety/scheme. But not sure at this point...)
* I do not know how the dispatch can help to clean up the code. At this point, I was mostly thinking of removing unnecessary loops, variables, etc. Or generally speaking - as little code as possible that produces the desired functionality. Certainly and as always, more can be done. But one thing at a time. :)
* I conduct a doc test for the `put_over_concrete_base`. I admit that I have not yet added a CI test, but there is a test. (Aside: I recently pushed the test coverage of the FTheoryTools up a lot.)

Does that answer your questions (and raise enough new questions ;) )?

PR #3344, opened 2024-02-09
JA homogenizer

JohnAAbbott:
First round of improvements after feedback via GitHub


JohnAAbbott:
**QUICK QUESTION**  should I remove the old interface (namely the function `homogenization`)?

wdecker:
Please remove the old interface.

JohnAAbbott:
I have just pushed the revised implementation.
Updated: doc, tests, calls to homogenization in other people's code.  Removed old implementation.
I hope this is practically final now.

fieker:
@HechtiDerLachs  will review (and APPROVE) it

HechtiDerLachs (approved):
If it works, it's good enough for me. Code seems to follow most conventions now. Type parametrization is still missing, but also not super important. 

It looks like more of the old tests were deleted than new ones added. If this is the case: Is it possible to convert more of the old tests into tests for the new implementation?

PR #3349, opened 2024-02-09
Add `Lowercase()` and two exports

joschmitt:
Makes the code in the book look prettier (and is hopefully non-controversial).

### DIFF
--- a/experimental/LinearQuotients/src/linear_quotients.jl
+++ b/experimental/LinearQuotients/src/linear_quotients.jl
@@ -16,7 +16,8 @@ function linear_quotient(G::MatrixGroup)
 end
 
 function show(io::IO, L::LinearQuotient)
-  print(io, "Linear quotient by $(group(L))")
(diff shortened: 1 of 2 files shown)


joschmitt:
@fingolfin you seemed to have an opinion on the reflection vs pseudo-reflection terminology. Is it fine to export these like this?

fingolfin:
I did? All I recall is my comment in https://github.com/oscar-system/Oscar.jl/pull/2669 where I was against calling this a "reflection". I think I am happy to call it a pseudo reflection (and export that name). If I voiced something different elsewhere, I forgot (quite possible!)

fingolfin (approved):

joschmitt:
> I did? All I recall is my comment in #2669 where I was against calling this a "reflection". I think I am happy to call it a pseudo reflection (and export that name). If I voiced something different elsewhere, I forgot (quite possible!)

No, that's all I remember too. Thanks!

PR #3350, opened 2024-02-10
New linear solving interface + more upstream changes

### LINKED ISSUE #1062: Inconsistent behaviour of kernel
As already discussed in Slack, there is some inconsistency with what `kernel` returns (and in particular with `kernel` and `nullspace`).
Sometimes the output just contains the basis, sometimes there are additional zero columns.
```
julia> M = matrix(GF(2), [ 1 0 1; 1 0 1]);

julia> kernel(M)[2] # calling Hecke
[0   1   0]
[1   0   0]
[0   1   0]

julia> nullspace(M)[2] # calling AbstractAlgebra
[0   1]
[1   0]
[0   1]
```
Also, the echelon structure is sometimes different:
```
julia> M = matrix(ZZ, [ 1 0 1; 1 0 1]);

julia> kernel(M)[2] # calling Hecke
[ 1   0]
[ 0   1]
[-1   0]

julia> nullspace(M)[2] # calling AbstractAlgebra
[0    1]
[1    0]
[0   -1]
```
Regarding the second point: I have the impression that the basis of the kernel is always in a reduced column echelon form, but this is not documented, so one cannot rely on it. Would it be possible to guarantee something here, like that the basis is always in lower-left reduced column echelon form?

joschmitt:
The functions `solve`, `can_solve`, `can_solve_with_solution`, `can_solve_with_solution_and_kernel` and `kernel` (for matrices) now use the "new" interface from AbstractAlgebra.
This is a breaking change ~and needs a Hecke update~.

~I put everything on top of #3366 so that OSCAR works with the newest Nemo/AbstractAlgebra. Will rebase once #3366 is merged.~

Closes #1062 :partying_face: :tada:


experimental/LieAlgebras/src/Util.jl
```diff
@@ -24,7 +24,7 @@ function coefficient_vector(M::MatElem{T}, basis::Vector{<:MatElem{T}}) where {T
   for i in 1:nr, j in 1:nc
     rhs[(i - 1) * nc + j, 1] = M[i, j]
   end
-  fl, sol = can_solve_with_solution(lgs, rhs)
+  fl, sol = can_solve_with_solution(lgs, rhs; side = :right)
   @assert fl
```
fieker: replace by plain solve?

joschmitt: Right. There are probably more such instances.

experimental/Schemes/DerivedPushforward.jl
```diff
@@ -75,7 +75,7 @@ end
 =#
 
 function rank(phi::FreeModuleHom{FreeMod{T}, FreeMod{T}, Nothing}) where {T<:FieldElem}
-  return ngens(domain(phi)) - left_kernel(sparse_matrix(phi))[1]
+  return ngens(domain(phi)) - nrows(kernel(sparse_matrix(phi), side = :left))
```
fieker, line 78: I think we need a function nullity - to return the rank of the kernel only...

joschmitt: I add it to the list of future improvements.

joschmitt:
There is a type inference problem on nightly. It already happens only with Hecke (but not only with Nemo):
```
julia> function foo(R::MPolyRing)
           return [ x for x in gens(R) ]
       end
foo (generic function with 1 method)

julia> K, a = cyclotomic_field(3, "a");

julia> Kx, (x, y) = K["x", "y"];

julia> @code_warntype foo(Kx)
MethodInstance for foo(::AbstractAlgebra.Generic.MPolyRing{AbsSimpleNumFieldElem})
  from foo(R::MPolyRing) @ Main REPL[2]:1
Arguments
  #self#::Core.Const(foo)
  R::AbstractAlgebra.Generic.MPolyRing{AbsSimpleNumFieldElem}
Body::Any
1 ─ %1 = Main.gens::Core.Const(AbstractAlgebra.gens)
│   %2 = (%1)(R)::AbstractVector{<:AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}}
│   %3 = Base.Generator(Base.identity, %2)::Base.Generator{I, typeof(identity)} where I<:(AbstractVector{<:AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}})
│   %4 = Base.collect(%3)::Any
└──      return %4
```
@thofma can/should we do something about this?

fingolfin (approved):

fingolfin:
Regarding the new invalidation: open an issue with the details (we can look at it with help of JET at some point), but otherwise move on.

thofma:
Not clear if this is about invalidations.
```julia
julia> @inferred gens(Kx)
ERROR: return type Vector{AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}} does not match inferred return type AbstractVector{<:AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}}
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[11]:1

julia> @code_warntype gens(Kx)
[...]
Body::Vector{AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}}
[...]
```
This is inconsistent and one of them is lying.

thofma:
There is something dodgy in AA, where we rely on some compiler heuristic. Not sure what exactly broke it. I think I have a fix in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1611. Can you confirm that this fixes the issue @joschmitt?

joschmitt:
> There is something dodgy in AA, where we rely on some compiler heuristic. Not sure what exactly broke it. I think I have a fix in [Nemocas/AbstractAlgebra.jl#1611](https://github.com/Nemocas/AbstractAlgebra.jl/pull/1611). Can you confirm that this fixes the issue @joschmitt?

Yes, this also solves the original failure in the nightly tests here.

lgoettgens:
This should get backported, right?

thofma:
This should be backported. It includes two breaking changes to the linear algebra, which are long overdue. Clean up `kernel` and clean up indexing.

I adjusted the book two weeks ago for the indexing changes (https://github.com/oscar-system/Oscar.jl/pull/3276) is the corresponding OSCAR PR) and the book does not use `kernel` of matrices.

CC: @fieker

PR #3351, opened 2024-02-10
Overhaul root system type detection

lgoettgens:
The function `cartan_type_with_ordering` errored and gave wrong results on some inputs. This PR completely rewrites the function and adds exhaustive tests.
Furthermore, all of the constructors of root system related types now take the same types of arguments.
And there are now more ways to print a Dynkin diagram. These were added here to be able to provide some visual tests for `cartan_type_with_ordering`.

### DIFF
--- a/experimental/LieAlgebras/src/RootSystem.jl
+++ b/experimental/LieAlgebras/src/RootSystem.jl
@@ -39,7 +39,7 @@ Passing `detect_type=false` will skip the detection of the root system type.
 function root_system(cartan_matrix::ZZMatrix; check::Bool=true, detect_type::Bool=true)
   @req !check || is_cartan_matrix(cartan_matrix) "Requires a generalized Cartan matrix"
   R = RootSystem(cartan_matrix)
-  detect_type && set_root_system_type(R, cartan_type_with_ordering(cartan_matrix)...)
+  detect_type && is_finite(weyl_group(R)) && set_root_system_type(R, cartan_type_with_ordering(cartan_matrix)...)
   return R
 end
 
--- a/experimental/LieAlgebras/src/RootSystem.jl
+++ b/experimental/LieAlgebras/src/RootSystem.jl
@@ -340,10 +340,6 @@ function simple_roots(R::RootSystem)
   return positive_roots(R)[1:rank(R)]
 end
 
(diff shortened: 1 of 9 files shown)


experimental/LieAlgebras/test/WeylGroup-test.jl
```diff
@@ -55,6 +55,7 @@ include(
     ("B4", weyl_group(root_system(:B, 4))),
     ("D5", weyl_group(cartan_matrix(:D, 5))),
     ("F4+G2", weyl_group((:F, 4), (:G, 2))),
+    ("E6+C3", weyl_group([(:E, 6), (:C, 2)])),
```
felix-roehrich:
```suggestion
    ("E6+C3", weyl_group([(:E, 6), (:C, 3)])),
```

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -280,10 +284,29 @@ function rank(R::RootSystem)
 end
 
 function root_system_type(R::RootSystem)
-  @req has_root_system_type(R) "root system type not defined"
```
felix-roehrich, line 283: What happens when the root system type is not known ?

lgoettgens: The next line errors

experimental/LieAlgebras/test/CartanMatrix-test.jl
felix-roehrich, line 1: For the tests there is a lot of repetition. Maybe put the common tests into a test function and test special cases like B2/C2 extra. I don't know how I feel about the tests for D4 and E6: The returned ordering could be said to be an implementation detail, in the sense that if ordering is changed by changes to the algorithm this not a breaking change; in this case that should be clearly documented.

experimental/LieAlgebras/src/RootSystem.jl
felix-roehrich: Maybe a more explicit error would be good for the user to know why this fails. Getting a "not defined" (?) error is not really helpful.

lgoettgens:
Everything should be addressed now

felix-roehrich (approved):
Thanks. Lgtm.

PR #3357, opened 2024-02-12
tests: adjust the project and load path to add test dependencies in test_module

### LINKED ISSUE #3354: `test_module` not working for some files/directories
Here is one example, but there are others, e.g. `experimental/LieAlgebras`:
```
julia> Oscar.test_module("Groups")
┌ Info: spawning
[...]
Test Summary:                                         | Pass  Total  Time
abelian_invariants_schur_multiplier for FinGenAbGroup |  173    173  0.4s
ERROR: LoadError: ArgumentError: Package Documenter not found in current path.
- Run `import Pkg; Pkg.add("Documenter")` to install the Documenter package.
Stacktrace:
  [1] macro expansion
    @ Base ./loading.jl:1766 [inlined]
```

benlorenz:
To have test dependencies (e.g. Documenter or Aqua) available when running tests via `test_module`.
Achieved by creating a temporary project using the `test/Project.toml` file and stacking this with the currently active project.
An alternative would be to try to hook into some of the internals of Pkg.jl and reuse that but it seemed rather fragile.

Seems to work fine in local tests, both with `new=false` and `new=true`, even for the `Aqua.jl` tests which use several temporary environments itself. Tested with the parallel test jobs as well.

This feature is always active unless one specifies `tempproject=false`, which is used in the main `runtests.jl` to avoid creating new projects for every testfile.

fixes #3354

### DIFF
--- a/src/utils/tests.jl
+++ b/src/utils/tests.jl
@@ -87,6 +87,12 @@ function _gather_tests(path::AbstractString; ignore=[])
   return tests
 end
 
+_setactiveproject(s::String) = @static if VERSION >= v"1.8"
+                                 Base.set_active_project(s)
+                               else
+                                 Base.ACTIVE_PROJECT[] = s
+                               end
+
 @doc raw"""
     test_module(path::AbstractString; new::Bool = true, timed::Bool=false, ignore=[])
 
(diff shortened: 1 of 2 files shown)


thofma (approved):
I tested it with `Groups` and some other failing examples in the experimental folder. Seems to work, thanks!

src/utils/tests.jl
```diff
@@ -140,18 +151,39 @@ function test_module(path::AbstractString; new::Bool=true, timed::Bool=false, ig
...
+        tmpproj = joinpath(mktempdir(), "Project.toml")
+        cp(joinpath(Oscar.oscardir, "test", "Project.toml"), tmpproj)
+        # activate the temporary project
+        _setactiveproject(tmpproj)
+        # and make sure the current project is still available to allow e.g. `using Oscar`
+        pushfirst!(LOAD_PATH, dirname(project_path))
+        Pkg.resolve()
+      end
+
+      try
```
thofma, line 168: Out of curiosity, why add the `try` here?

benlorenz: For the finally block that restores the load path and active project (when exiting the block with an exception).

PR #3360, opened 2024-02-13
added line about function argument order

antonydellavecchia:
adds line to styleguide


docs/src/DeveloperDocumentation/styleguide.md
```diff
@@ -29,6 +29,8 @@ deviate from them in some cases; in that case just do so.
   different, even if the result describes the same mathematical object, it
   should be indicated in the function name, for example `automorphism_group` vs
   `automorphism_group_generators` vs `automorphism_list`.
+- Whenever functions expect a Ring, Field, Algebra, etc. as input they should
+  be passed as the first argurment. For example `polynomial_ring(QQ, "x")`
```
lgoettgens:
```suggestion
- Whenever functions expect a ring, field, algebra, etc. as input they should
  be passed as the first argument, for example, `polynomial_ring(QQ, "x")`.
```

lkastner (approved):

Issue #3345, opened 2024-02-09
Rename Spec to AffineScheme


paemurru:
We have the types `Spec` and `ProjectiveScheme`. To make the names consistent, we will change `Spec` into `AffineScheme`. The functions to construct affine and projective schemes will still be called `spec` and `proj`. This completes one bullet point in [Schemes Meta Issue #3202](https://github.com/oscar-system/Oscar.jl/issues/3202).

For consistency, we need to also change many related types and functions like `AbsSpec` and `SpecMor`. See https://github.com/oscar-system/Oscar.jl/issues/3345#issuecomment-1961313435 for the list of planned changes.

fieker:
I don't have any (strong) opinion. Wha I would like you (all) to think about, is how someone who does not know what a scheme is can create and use the P^2(QQ), the projective space as introduced in many (classical) (basic) texts. How is this interaction going to be? Those will be the vast majority of users....

simonbrandhorst:
The names of the underscore constructors won't change. So your question is tangential.
`P = projective_space(QQ,2)` will create projective space.
`rational_points(P)` should return the set of its rational points and `P([0,0,1])` create the rational point `[0:0:1]` just like in a text book. We could consider `P(QQ)` to return the set of QQ rational points. 
Further constructors could be 
`projective_space(ToricVariety, QQ, 2)`
`projective_space(Scheme, QQ, 2)`
`projective_space(Set, QQ, 2)` 
you name it.

paemurru:
In yesterday's OSCAR meeting it was decided with @micjoswig, @wdecker, @jankoboehm and others to:

* keep the type `ProjectiveScheme`,
* rename the type `Spec` to `AffineScheme`,
* rename the function `Spec` to `spec`, and
* introduce an alias `affine_scheme` of `spec`.

wdecker:
Similarly, we should have ProjectiveScheme, proj, and projective_scheme.

wdecker:
Yesterday, I was not aware of the original issue which suggests moving ProjectiveScheme to Proj etc. So this contradicts to what we decided yesterday. So why moving rojectiveScheme to Proj etc? Isn’t it fine, as it is?

paemurru:
@wdecker The original issue was the inconsistency that currently we have the types `Spec` and `ProjectiveScheme`. For consistency, we would like to have either `AffineScheme` and `ProjectiveScheme` or `Spec` and `Proj`.

[...]

PR #3364, opened 2024-02-14
refactor `number of` aliases

benlorenz:
Implementation of some followup discussions to #3272.

Change short form to use `n_*` with snake case and no abbreviations, following the general oscar style guide.

Switch polyhedral and related code to use `n_` variants by default, but keep `number_of` aliases.

Add more `n_` aliases for consistency (`n_gens`, `n_cols`, `n_rows`, `n_vars`, `n_digits`).

cc: @lgoettgens @micjoswig @fingolfin @fieker @lkastner @antonydellavecchia


lkastner (approved):
Thank you.

lgoettgens:
Thank you for working on this!

I have no strong opinion neither to this suggestion nor to the current state. But IF we decide to go down this road, I would like it to be consistent throughout, in particular the following would need to be thought of / adapted, at least in my opinion:

1. The `n_*` aliases shouldn't contain abbreviations. But that's not true for `n_gens`, `n_vars`, `n_cols`, that really would need to be named `n_generators`, `n_variables`, `n_columns`.
2. All aliases should be moved as far upstream as possible, to make code look more similar between our projects. E.g. the `n_rows` thing should be moved to AA. 
3. The "old" shorthand without underscore (e.g. `ngens`) should be removed. Otherwise we have three ways to write the same thing, which just seeds future confusion. 
4. Due to the previous points, code will already become a lot more lengthy, as I now need to type `n_generators` instead of `ngens` (increase by 140%).
5. The change should be documented in the style guide (basically https://github.com/oscar-system/Oscar.jl/pull/3305 but for this proposal).

Pinging @thofma @YueRen as they took part in the discussion for the `number_of_*` version in the coding sprints to let them know.

micjoswig:
Let me comment on this.

At some point we had agreed to prefer long names over short ones.  The reasoning was that OSCAR covers many different fields of math, with many different users.  Long names are easier to understand.

Sometimes, maybe there are reasons for different decisions; e.g., see #3345.

What bugs me about "number_of_" is something else.  The decision in favor of long names generally eats up a lot of space, literally.  I mean: space on the screen when you hit <tab> for instance or space on a slide when you give a presentation showing OSCAR code.  As a consequence, we need to be careful with that space.  And here the argument in favor of "number_of" (systematic) is too weak for my taste, because it does not really add to the readability of the code.

A second reason for touching these functions probably was that it was inconsistent before, e.g., "nfacets" was maybe not too readable.  But that had never been my choice.  I had always voted for "n_" (as it is implemented now) because I think it is a good compromise between readability, systematic and space; but my opinion was overruled by some "Julia systematics" argument.  Since that had been revoked apparently, it is now time for "n_".

I hope this does not make too many people too unhappy.

micjoswig:
> I have no strong opinion neither to this suggestion nor to the current state. But IF we decide to go down this road, I would like it to be consistent throughout, in particular the following would need to be thought of / adapted, at least in my opinion:

We had discussed this within the Berlin team.  I think the "ngens and friends" thingies are deeply rooted in the minds of many users and developers.  Taking that away will inflict pain.  Therefore, we stayed away from touching these.

I am absolutely open to "n_gens" etc. But I am shying away from forcing this.  I am also OK with some exceptions.  After all it is almost impossible to stay globally consistent.  For instance, for graphs somebody made an early executive decision (not me) to make the OSCAR graphs compatible with the Julia graphs.  This gave us "nv" and "ne" that I do not like particularly.  But there is an advantage from being compatible with the Julia graphs.

It is just difficult.

benlorenz:
> I have no strong opinion neither to this suggestion nor to the current state. But IF we decide to go down this road, I would like it to be consistent throughout, in particular the following would need to be thought of / adapted, at least in my opinion:
> 
> 1. The `n_*` aliases shouldn't contain abbreviations. But that's not true for `n_gens`, `n_vars`, `n_cols`, that really would need to be named `n_generators`, `n_variables`, `n_columns`.

Indeed, done.

> 2. All aliases should be moved as far upstream as possible, to make code look more similar between our projects. E.g. the `n_rows` thing should be moved to AA.

I don't think this is critical, we just want to provide a consistent interface for Oscar, especially now for the book.
For example Polymake.jl has lots of names that do not fit the Oscar naming scheme.

> 3. The "old" shorthand without underscore (e.g. `ngens`) should be removed. Otherwise we have three ways to write the same thing, which just seeds future confusion.

For every rule that we have there will be exceptions, I would be fine with keeping a few shortened versions in whatever form. (`ngens`, `_ngens`, exported or non-exported)
I think removing `ngens` now throughout all packages (and the book) is too disruptive right now.

> 4. Due to the previous points, code will already become a lot more lengthy, as I now need to type `n_generators` instead of `ngens` (increase by 140%).

See above.
For internal use we can always have short non-exported aliases. And it is still way better than having to write `number_of_generators`.

> 5. The change should be documented in the style guide (basically [Add paragraph about `number_of_` to the styleguide. #3305](https://github.com/oscar-system/Oscar.jl/pull/3305) but for this proposal).

I will add some suggestions to that PR soon.

micjoswig (approved):
Looks good to me.

fingolfin:
Ugh, I thought the plan was to discuss this tomorrow, why was it merged already now?

micjoswig:
> Ugh, I thought the plan was to discuss this tomorrow, why was it merged already now?

At the meeting last Friday we agreed that potentially controversial issues should be debated in issues.  And there was no complaint here.  The goal is to keep today's meeting as light as possible.  After all, we finally need to meet our combined three deadlines.

fingolfin:
It is unfortunate this was not communicated beyond that meeting then. It seem ms good communication is indeed hard for all of us.

PR #3365, opened 2024-02-14
Make `partitions` and `(semi)standard_tableaux` return iterators

joschmitt:
... the quick and dirty way to have a stable API.

There is no real change here, but it was agreed in #3159 that these functions should return iterators and not arrays. Back then I assumed, there would be some months before 1.0 to make this fundamental change. As this is apparently not the case, I just aim to get the API right for now.

Note: if we ever decide afterwards to return to arrays (and introduce e.g. `iterate_partitions`), we can still do that WITHOUT breaking API because arrays are iterable.
The other direction doesn't work because an iterator does e.g. not support `getindex`.


lgoettgens (approved):
Lgtm

joschmitt:
Merge?

Issue #3330, opened 2024-02-07
`normalize!` vs `normalise`


fingolfin:
AA defines `normalise` (which has no docstring but is documented as part of the polynomial element interface).  Oscar defines `normalize!`.

The two really have nothing to do with each other, but this is till confusing. Perhaps we can rename one (which? did we agree whether we prefer US, UK or some other english "default" spelling?).

Of course the old name can be kept as an alias ...

lgoettgens:
For `centralizer`, `normalizer`, `localized`, `finalize!` we already chose the US spelling.

Concerning aliases, we already have https://github.com/oscar-system/Oscar.jl/blob/757489f253692ef05392deb6eb4e1e5fdf231236/src/Groups/sub.jl#L357 but I am not a fan of having such aliases between US and UK spelling.

Another note on the `AA.normalise` for polynomials. I expected it to normalize the polynomial, so something along `normalise(f::PolyRingElem) = divexact(f, leading_coefficient(f))`, but instead it seems to just be about the internal representation of the polynomial in memory(?)? I would want a function for the former functionality as well (but maybe non-mutating) and would expect it to be called `normalization`. So I don't know if changing `s`->`z` in the AA function is the way to go here

micjoswig:
I agree with @lgoettgens Let's stick to US spelling, without BE aliases.

lgoettgens:
https://github.com/oscar-system/Oscar.jl/pull/3367 deals with some British spellings I found in the source code, but explicitly not with `normalise`: First, it is introduced in AA and thus needs some work to propagate changes to Oscar. Second, I don't really understand what it is used for and if another name would be better (see https://github.com/oscar-system/Oscar.jl/issues/3330#issuecomment-1932023841). Can somebody give an insight? Maybe @thofma @fingolfin or @fieker (all familiar with AA and Nemo)?

PR #3367, opened 2024-02-14
Change some things from British to American spelling

lgoettgens:
As brought up in https://github.com/oscar-system/Oscar.jl/issues/3330#issuecomment-1943638467.
This explicitely does *not* deal with `normalise`.

Can somebody with access to the book please check if the two new deprecations are an issue?


micjoswig (approved):
Looks good to me.

joschmitt:
> Can somebody with access to the book please check if the two new deprecations are an issue?

Neither `factorisations` nor `centraliser` are used in code in the book as of now.

PR #3369, opened 2024-02-15
Performance improvements for PointVector, RayVector, SubObjectIterator

fingolfin:
Add type assertions to their `getindex` methods to improve type stability.
Similarly ensure that the return value of `size` is recognized as `Tuple{Int}`.

Silly microbenchmark, before:

    julia> v = PointVector{ZZRingElem}(matrix(ZZ, [1 2 3 ; ]));

    julia> @btime prod(v)
      1.217 μs (21 allocations: 336 bytes)

After:

    julia> @btime prod(v);
      288.449 ns (5 allocations: 80 bytes)

### DIFF
--- a/src/PolyhedralGeometry/iterators.jl
+++ b/src/PolyhedralGeometry/iterators.jl
@@ -13,7 +13,7 @@ for (T, _t) in ((:PointVector, :point_vector), (:RayVector, :ray_vector))
     end
 
     Base.IndexStyle(::Type{<:$T}) = IndexLinear()
-    Base.getindex(po::$T, i::Base.Integer) = po.p[1, i]
+    Base.getindex(po::$T{U}, i::Base.Integer) where U = po.p[1, i]::U
 
     function Base.setindex!(po::$T, val, i::Base.Integer)
       @boundscheck 1 <= length(po) <= i
@@ -23,7 +23,7 @@ for (T, _t) in ((:PointVector, :point_vector), (:RayVector, :ray_vector))
 
     Base.firstindex(::$T) = 1
     Base.lastindex(iter::$T) = length(iter)
-    Base.size(po::$T) = (size(po.p, 2),)
+    Base.size(po::$T) = (size(po.p, 2)::Int,)
 
     coefficient_field(po::$T) = base_ring(po.p)
 
@@ -237,7 +237,7 @@ Base.IndexStyle(::Type{<:SubObjectIterator}) = IndexLinear()
 
 function Base.getindex(iter::SubObjectIterator{T}, i::Base.Integer) where T
     @boundscheck 1 <= i && i <= iter.n
-    return iter.Acc(T, iter.Obj, i; iter.options...)
+    return iter.Acc(T, iter.Obj, i; iter.options...)::T
 end
 
 Base.firstindex(::SubObjectIterator) = 1


benlorenz:
Thanks for working on this, this does seem to introduce a few new invalidations, can we do something about these?

I am a bit unsure which of these are really needed and why, `eltype` for example is inherited from `AbstractVector{U}` and works fine.

I tried this in the shell on master without the changes here and it seems that mostly the hint for size is important for this speed improvement:
```julia
julia> v = PointVector{ZZRingElem}(matrix(ZZ, [1 2 3 ; ]));

julia> @btime prod(v)
  687.593 ns (21 allocations: 336 bytes)
6

julia> Base.size(po::PointVector) = (size(po.p, 2)::Int,)

julia> @btime prod(v)
  118.364 ns (5 allocations: 80 bytes)
6
```
A bit confused why this type cannot be inferred automatically, we just use the size for MatrixElem.

PS: What julia version / what machine is that? It seems a lot slower than on my laptop.

benlorenz (approved):
thanks!

fingolfin:
Thank you for the feedback, Benjamin!

> Thanks for working on this, this does seem to introduce a few new invalidations, can we do something about these?

I think they were a fluke (?) or caused by an update of a dependency; at least now this seems to be gone? (I pushed an update to this branch earlier)

> 
> I am a bit unsure which of these are really needed and why, `eltype` for example is inherited from `AbstractVector{U}` and works fine.

Indeed, my mistake -- I overlooked the inheritance and just added it -- should have checked whether it really has an effect! Anyway I removed those now.

> 
> I tried this in the shell on master without the changes here and it seems that mostly the hint for size is important for this speed improvement:

Yes, that's the main part; but together they still are better than just the `size` change.

The code where this came up initially used a `for` loop and I first noticed that it was type unstable and then changed `getindex`. Then I thought "let's rewrite this using `prod` and discovered it was much slower -- hence the `size` and `eltype` changes.

Anyway, here is a new micro benchmark using a `for` loop.:
```julia
function f(v)
  x = v[1]
  for i in 2:length(v)
    x *= v[i]
  end
  x
end
```

And now some measurements:

Before this PR:
```
julia> @btime f(v);
  648.773 ns (11 allocations: 224 bytes)

julia> @btime prod(v);
  1.267 μs (21 allocations: 336 bytes)
```
With just the `size` change:
```
julia> @btime f(v);
  364.333 ns (5 allocations: 80 bytes)

julia> @btime prod(v);
  410.000 ns (5 allocations: 80 bytes)
```
With just the `getindex` change:
```
julia> @btime f(v);
  647.610 ns (11 allocations: 224 bytes)

julia> @btime prod(v);
  1.262 μs (21 allocations: 336 bytes)
```
With the full PR:
```
julia> @btime f(v);
  351.807 ns (5 allocations: 80 bytes)

julia> @btime prod(v);
  369.335 ns (5 allocations: 80 bytes)
```

So `getindex` alone is useless, but the combo still is slightly ahead.

> A bit confused why this type cannot be inferred automatically, we just use the size for MatrixElem.

Because the member `p` has type `MatElem{U}` which is not concrete; and the generic method is implemented as
```
size(x::MatrixElem{T}) where T <: NCRingElement = (nrows(x), ncols(x))
```

but it can't infer the return type of `nrows(x)`...

That suggests two things to me:
1. we probably have some type unstable `nrows` methods (a quick check with `Base.return_types` suggests 2 out of 17)
2. even if we fix those, perhaps we can add `::Int` type annotation to that generic `size` method just to compensate for similar issues in the future.

So hopefully the `::Int` annotation in this PR here can be made redundant; but in the meantime it doesn't hurt...

> 
> PS: What julia version / what machine is that? It seems a lot slower than on my laptop.

Julia 1.10.1 on my MacBook Pro M1 Max.

fingolfin:
Timings with Julia 1.10.1 on an AMD EPYC 9554 64-Core Processor (which, despite the high core count, is currently the fastest machine I have access to, in terms of single-core performance):

Before this PR:
```
julia> @btime f(v);
  293.838 ns (11 allocations: 224 bytes)

julia> @btime prod(v);
  713.264 ns (21 allocations: 336 bytes)
```

With this PR:
```
julia> @btime f(v);
  139.726 ns (5 allocations: 80 bytes)

julia> @btime prod(v);
  139.512 ns (5 allocations: 80 bytes)
```

benlorenz:
Thanks for the explanations!

It makes sense to do this here and it doesn't hurt and we can try to fix nrows later.
The invalidation might have been from the `eltype`.

The `i7-1165G7` in my laptop boosts to 4.7GHz and with the full PR I get:
```julia
julia> @btime f(v);
  101.805 ns (5 allocations: 80 bytes)

julia> @btime prod(v);
  102.918 ns (5 allocations: 80 bytes)

```

This microbenchmark seems surprisingly slow on the M1, maybe julia is missing some optimization there.

fingolfin:
Yeah I agree - usually my MacBook is twice as fast as the older servers, and only the AMD beat it, but not by much (it doesn't clock as high as yours, of course). But in this case the old servers are as fast or faster.

For fun, here are 8 machines (including the M1 munk in the lower right) running on master (with this PR here merged) with Julia 1.10.1:

Screen Shot 2024-02-16 at 14 29 24

The top left is one at home and it is on an ancient i7-3820 and it still blows away the M1. Wow.

PR #3372, opened 2024-02-15
adds fix to return non reduced (co)homology

antonydellavecchia:
Return non reduced (co)homology

### DIFF
--- a/src/Combinatorics/SimplicialComplexes.jl
+++ b/src/Combinatorics/SimplicialComplexes.jl
@@ -90,13 +90,15 @@ function vertexindices(K::SimplicialComplex)
     end
 end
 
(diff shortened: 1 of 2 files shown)


micjoswig (approved):
Thanks for taking care of this.  Let me explain why this change is necessary for v1.0.

In terms of information content there is no difference between reduced and non-reduced homology.  Yet, the two different conventions both have their minor (cosmetic) advantages and disadvantages.  Roughly speaking, reduced is nicer for combinatorics (hence this is polymake's default), non-reduced is nicer algebraically: The algebra structure of the cohomology looks weird in the reduced setting.

Because OSCAR has a much stronger algebraic focus than polymake, the preferences should be different.

PR #3373, opened 2024-02-15
Fix show for RelativeBrauerGroupElem

fingolfin:
- fix error because `Indent` was undefined
- remove extra newline at end
- add a missing `Lowercase`
- simplify the code

### DIFF
--- a/experimental/GModule/GaloisCohomology.jl
+++ b/experimental/GModule/GaloisCohomology.jl
@@ -1401,7 +1401,7 @@ function Base.show(io::IO, m::MIME"text/plain", a::RelativeBrauerGroupElem)
   print(io, "Element of relative Brauer group of ", Lowercase(), parent(a).k)
   io = IOContext(io, :supercompact => true, :compact => true)
   print(io, Indent())
-  data = sort(a.data; by = x -> x isa AbsSimpleNumFieldEmbedding ? Inf : minimum(x))
+  data = sort(collect(a.data); by =(x -> first(x) isa AbsSimpleNumFieldEmbedding ? Inf : minimum(first(x))))
   for (p,v) in data
     print(io, "\n", p, " -> ", v)
   end
--- a/experimental/GModule/GaloisCohomology.jl
+++ b/experimental/GModule/GaloisCohomology.jl
@@ -1401,7 +1401,8 @@ function Base.show(io::IO, m::MIME"text/plain", a::RelativeBrauerGroupElem)
   print(io, "Element of relative Brauer group of ", Lowercase(), parent(a).k)
   io = IOContext(io, :supercompact => true, :compact => true)
   print(io, Indent())
-  for (p,v) in a.data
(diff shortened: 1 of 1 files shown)


lgoettgens (approved):

lgoettgens (commented):
Can you maybe add a doctest to just any function exercising this show? This would help to find future similar things and increases coverage.

fingolfin:
I'll see what I can do. 

Added "backport to 1.0 label" as I I need this for the book (where it is "tested")

experimental/GModule/GaloisCohomology.jl
```diff
@@ -1514,6 +1511,30 @@ an infinite direct sum of the local Brauer groups.
...
+julia> B, mB = relative_brauer_group(base_ring(S), character_field(S));
+
+julia> B
+Relative Brauer group for cyclotomic field of order 5 over number field of degree 1 over QQ
+
+julia> b = B(S)
+Element of relative Brauer group of number field of degree 1 over QQ
+  <2, 2> -> 1//2 + Z
+  Complex embedding of number field -> 1//2 + Z
+  <5, 5> -> 0 + Z
```
fingolfin, line 1536: Seems the order in which these three lines are printed is not stable, hence some of the doctest CI jobs fail :/

lgoettgens: Is there a canonical way to sort them (for a consistent printing)? If not, you might use a doctest-filter

fingolfin: I don't see a good way to sort the keys *in general*, they are ideals in a number field, or field embeddings. But we could make sure that the ideals come first. And perhaps the ideals could be sorted by minimum or norm. That won't give a unique sorting in all cases, but it is good enough for this example and it probably is nice for the user overall. Will do that for now.


@fieker I also wonder if we should print something nicer instead of "Complex embedding of number field"?

lgoettgens (approved):
I have no idea of the mathematics behind this, but the printing change by itself look good

experimental/GModule/GaloisCohomology.jl
```diff
@@ -1398,14 +1398,12 @@ end
...
+  io = IOContext(io, :supercompact => true, :compact => true)
   print(io, Indent())
-  for (p,v) = a.data
-    show(ioC, p)
-    print(io, " -> ")
-    show(ioC, v)
-    print(io, "\n")
+  data = sort(a.data; by = x -> x isa AbsSimpleNumFieldEmbedding ? Inf : minimum(x))
+  for (p,v) in data
+    print(io, "\n", p, " -> ", v)
```
lgoettgens: Dictionaries are not sortable, you need to collect first
```suggestion
  data = sort(collect(a.data); by =(x -> first(x) isa AbsSimpleNumFieldEmbedding ? Inf : minimum(first(x))))
  for (p,v) in data
    print(io, "\n", p, " -> ", v)
```

PR #3374, opened 2024-02-15
Hide AbstractAlgebra `ordering` of a polynomial ring (a bit)

### LINKED ISSUE #3040: Clarify how monomial orderings are supposed to be specified
There is a lot of potential for confusion in how one can specify monomial orderings in OSCAR (see e.g. a question by @gfourier in Slack some time ago and #3039).
* One can either give an ordering with the ring ("Singular style"):
```
julia> R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"], ordering = :degrevlex)
```
There are only three different orderings available and this appears to be used for the functionality from `AbstractAlgebra.jl`, most prominently `divrem` and printing.
* Secondly, there are the fancy orderings of type `MonomialOrdering` which one specifies for e.g. `leading_term`, `groebner_basis` and so on.

There should be a consistent way of using monomial orderings in OSCAR.

**Additional context**
I think we should get rid of the orderings from `AbstractAlgebra.jl`. I lost track of what the state of `divrem` vs `normal_form` is though.

joschmitt:
The `ordering` keyword of `polynomial_ring` is renamed to `internal_ordering` to notify that this has nothing to do with the `ordering` keyword of many other functions in OSCAR and to discourage non-expert users from using it.

I removed all mentions of this keyword from the native OSCAR documentation. It is still mentioned/explained in the doc string coming from AbstractAlgebra which I adjusted a bit in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1597. I think lengthy explanations "why one shouldn't use it" are only confusing.

The keyword is used in more places than I expected. I ping the respective authors (according to `git blame`) in the comments with questions.

Closes #3040.

~Requires https://github.com/Nemocas/AbstractAlgebra.jl/pull/1597, https://github.com/Nemocas/Nemo.jl/pull/1678, https://github.com/thofma/Hecke.jl/pull/1402, https://github.com/oscar-system/Singular.jl/pull/783, https://github.com/algebraic-solving/AlgebraicSolving.jl/pull/43 and~ is a breaking change.


docs/src/CommutativeAlgebra/rings.md
```diff
@@ -25,15 +25,13 @@ of the coefficient ring of the polynomial ring.
 The basic constructor below allows one to build multivariate polynomial rings:
 
 ```@julia
-polynomial_ring(C::Ring, V::Vector{String}; ordering=:lex, cached::Bool = true)
+polynomial_ring(C::Ring, V::Vector{String}; cached::Bool = true)
```
joschmitt, line 28: @wdecker I removed mentioning this keyword here and below in this file. I think this way, it is less confusing for the non-expert user who does not need to know about such implementation details (it is still explained in the documentation on rings which we copy from AbstractAlgebra). Is this fine?

src/Rings/groebner.jl
```diff
@@ -318,13 +318,13 @@ See [Fau99](@cite) for more information.
 
 # Examples
 ```jldoctest
-julia> R,(x,y,z) = polynomial_ring(GF(101), ["x","y","z"], ordering=:degrevlex)
+julia> R,(x,y,z) = polynomial_ring(GF(101), ["x","y","z"])
```
joschmitt, line 321: @ederc I would like to remove the keyword here because it is part of the doc string which makes it look as if one needs to set the ordering like this. Can you confirm that it is not mathematically necessary to set the ordering?
Notice also the small change when printing the ideal below.

src/Rings/mpoly-ideals.jl
```diff
@@ -1944,14 +1944,13 @@ julia> grassmann_pluecker_ideal(2, 4)
 Ideal generated by
   x[1]*x[6] - x[2]*x[5] + x[3]*x[4]
 
-julia> R, x = polynomial_ring(residue_ring(ZZ, 7)[1], "x" => (1:2, 1:3), ordering=:degrevlex)
+julia> R, x = polynomial_ring(residue_ring(ZZ, 7)[1], "x" => (1:2, 1:3))
```
joschmitt: @antonydellavecchia Similar to the above: I would like to remove the keyword here because it is part of the doc string which makes it look as if one needs to set the ordering like this. Can you confirm that it is not mathematically necessary to set the ordering?
Notice also the small change when printing the ideal below.

test/Rings/groebner.jl
```diff
@@ -147,7 +147,7 @@ end
 end
 
 @testset "f4" begin
-  R, (x1,x2,x3,x4) = polynomial_ring(GF(next_prime(2^28)), ["x1", "x2", "x3", "x4"], ordering=:degrevlex)
+  R, (x1,x2,x3,x4) = polynomial_ring(GF(next_prime(2^28)), ["x1", "x2", "x3", "x4"], internal_ordering=:degrevlex)
```
joschmitt: I leave the keyword here although I assume/hope that it is not mathematically necessary to do the computations with `:degrevlex`. If I remove the keyword, I get slightly different polynomials which still give a Gröbner basis from what I can see, but are not the exact basis that is tested for.
CC @ederc

EDIT: Of course, the algorithm does computations w.r.t. the ordering degrevlex, but I hope that it is not necessary that the polynomials are stored in the corresponding order.

src/Rings/groebner.jl
ederc: Yes, you can remove the keyword here.

test/Rings/groebner.jl
ederc: I do not understand what you mean: The result should be a degrevlex GB for the input ideal.

ederc: Or is the difference only the order in which the terms are printed?

joschmitt: No, I actually get different polynomials in the Gröbner basis with `internal_ordering = :lex` (the default). They still form a degrevlex GB as far as I can tell, but the test checks for the exact same polynomials and hence fails.

src/Rings/mpoly-ideals.jl
antonydellavecchia: it is mathematically necessary, i.e. if you want to remove this well need to force the ordering in the grassman pluecker ideal when we override the ring

joschmitt: Okay, then we should really do this (override the ring). I don't think we can or should expect from the user to supply a ring with the correct ordering. I'll look into it.

wdecker:
I agree, thanks for doing this, WolframVon meinem iPhone gesendetAm 15.02.2024 um 15:53 schrieb Johannes Schmitt ***@***.***>:﻿
@joschmitt commented on this pull request.



In src/Rings/mpoly-ideals.jl:
> @@ -1944,14 +1944,13 @@ julia> grassmann_pluecker_ideal(2, 4)
 Ideal generated by
   x[1]*x[6] - x[2]*x[5] + x[3]*x[4]
 
-julia> R, x = polynomial_ring(residue_ring(ZZ, 7)[1], "x" => (1:2, 1:3), ordering=:degrevlex)
+julia> R, x = polynomial_ring(residue_ring(ZZ, 7)[1], "x" => (1:2, 1:3))

Okay, then we should really do this (override the ring). I don't think we can or should expect from the user to supply a ring with the correct ordering. I'll look into it.

src/Rings/mpoly-ideals.jl
antonydellavecchia: i agree, it shouldn't be too bad to fix

joschmitt: I don't understand why `:degrevlex` is necessary. The functions calls `grassmann_pluecker_ideal(::Int, ::Int)` which calls Polymake and then converts the result into an ideal in an OSCAR polynomial ring. The ordering is not specified in the conversion, so the newly created polynomial ring stores the polynomials according to `:lex`. And then all *this* functions does is building polynomials in the supplied polynomial ring out of the ones coming from Polymake by converting the coefficients. Where (on the OSCAR side) is the ordering coming in?

antonydellavecchia: polymake computes a groebner basis for the degrevlex ordering, so maybe we should fix `grassmann_pluecker_ideal(::Int, ::Int)` to return an IdealGens type now that these exist, since i don't believe they did when I wrote this?

joschmitt: I assume that we should return an ideal with the Gröbner basis cached as `IdealGens` in the corresponding field. But then it sounds like this is independent of this pull request?

antonydellavecchia: sure, I can deal with this in  new pull request

joschmitt: Great, thanks!

test/Rings/groebner.jl
ederc: Can I somehow see what you get as a result. If I use the current master of Oscar (w/o the required PRs for AA, Hecke, etc. mentioned above, I get the same reduced degrevlex GB computed by `groebner_basis_f4()` in the corresponding test case for any of these ring settings:

```
R, (x1,x2,x3,x4) = polynomial_ring(GF(next_prime(2^28)), ["x1", "x2", "x3", "x4"], ordering=:degrevlex)
R, (x1,x2,x3,x4) = polynomial_ring(GF(next_prime(2^28)), ["x1", "x2", "x3", "x4"], ordering=:lex)
R, (x1,x2,x3,x4) = polynomial_ring(GF(next_prime(2^28)), ["x1", "x2", "x3", "x4"])
```

Could you please paste the resulting GB on your side, so that I might be able to understand the problem.

joschmitt: This is very strange. I get the following on OSCAR#master, so independent of this pull request:
```
julia> R, (x1,x2,x3,x4) = polynomial_ring(GF(next_prime(2^28)), ["x1", "x2", "x3", "x4"]);

julia> I = ideal(R,[x1+2*x2+2*x3+2*x4-1,
               x1^2+2*x2^2+2*x3^2+2*x4^2-x1,
               2*x1*x2+2*x2*x3+2*x3*x4-x2,
               x2^2+2*x1*x3+2*x2*x4-x3]);

julia> H = groebner_basis_f4(I);

julia> G = [x1 + 2*x2 + 2*x3 + 2*x4 + 268435458
                     x3^2 + 2*x2*x4 + 76695850*x3*x4 + 115043772*x4^2 + 115043768*x2 + 191739613*x3 + 230087535*x4
                     x2*x3 + 268435457*x2*x4 + 230087533*x3*x4 + 76695842*x4^2 + 210913575*x2 + 38347923*x3 + 153391692*x4
                     x2^2 + 2*x2*x4 + 153391692*x3*x4 + 230087538*x4^2 + 230087536*x2 + 115043768*x3 + 191739613*x4
                     x3*x4^2 + 238609298*x4^3 + 14913081*x2*x4 + 56338306*x3*x4 + 129246702*x4^2 + 263464432*x2 + 260150414*x3 + 258493405*x4
                     x2*x4^2 + 89478486*x4^3 + 29826162*x2*x4 + 263464432*x3*x4 + 238609297*x4^2 + 141674270*x2 + 9942054*x3
                     x4^4 + 35851649*x4^3 + 232885084*x2*x4 + 251179133*x3*x4 + 231479137*x4^2 + 191484965*x2 + 85955250*x3 + 167408122*x4];

julia> @test elements(H) == G
Test Failed at REPL[19]:1
  Expression: elements(H) == G
  ...
```
With `ordering = :degrevlex` the test passes.
The computed GB above is:
```
julia> H
Gröbner basis with elements
1 -> x1 + 2*x2 + 2*x3 + 2*x4 + 268435458
2 -> x3^2 + 2*x2*x4 + 76695850*x3*x4 + 115043772*x4^2 + 115043768*x2 + 191739613*x3 + 230087535*x4
3 -> x2*x3 + 268435457*x2*x4 + 230087533*x3*x4 + 76695842*x4^2 + 210913575*x2 + 38347923*x3 + 153391692*x4
4 -> x2^2 + 2*x2*x4 + 153391692*x3*x4 + 230087538*x4^2 + 230087536*x2 + 115043768*x3 + 191739613*x4
5 -> x3*x4^2 + 238609298*x4^3 + 14913081*x2*x4 + 56338306*x3*x4 + 129246702*x4^2 + 263464432*x2 + 260150414*x3 + 258493405*x4
6 -> x2*x4^2 + 89478486*x4^3 + 29826162*x2*x4 + 263464432*x3*x4 + 238609297*x4^2 + 141674270*x2 + 9942054*x3
7 -> x4^4 + 35851649*x4^3 + 232885084*x2*x4 + 251179133*x3*x4 + 231479137*x4^2 + 191484965*x2 + 85955250*x3 + 167408122*x4
with respect to the ordering
degrevlex([x1, x2, x3, x4])
```
And you are right: `H` contains exactly the same elements as `G` if one looks closely.
For example
```
julia> H[2]
x3^2 + 2*x2*x4 + 76695850*x3*x4 + 115043772*x4^2 + 115043768*x2 + 191739613*x3 + 230087535*x4

julia> G[2]
2*x2*x4 + 115043768*x2 + x3^2 + 76695850*x3*x4 + 191739613*x3 + 115043772*x4^2 + 230087535*x4
```
are indeed the same polynomials, the terms are just printed in a different order.
But now comes the really strange part:
```
julia> parent(G[2]) === parent(H[2])
true

julia> G[2] == H[2]
false

julia> G[2] - H[2]
2*x2*x4 + 115043768*x2 + 268435457*x2*x4 + 191739613*x3 + 153391691*x2 + 76695846*x3
```
Can you confirm this?

Versioninfo:
```
julia> Oscar.versioninfo(full = true)
OSCAR version 1.0.0-DEV - #master, 871f2244a5 -- 2024-02-15 13:44:43 +0100
  combining:
    AbstractAlgebra.jl   v0.37.6
    GAP.jl               v0.10.2
    Hecke.jl             v0.27.0
    Nemo.jl              v0.41.5
    Polymake.jl          v0.11.13
    Singular.jl          v0.22.2
  building on:
    Antic_jll               v0.201.500+0
    Arb_jll                 v200.2300.0+0
    Calcium_jll             v0.401.100+0
    FLINT_jll               v200.900.9+0
    GAP_jll                 v400.1200.200+7
    Singular_jll            v403.214.1400+0
    libpolymake_julia_jll   v0.11.2+0
    libsingular_julia_jll   v0.41.0+0
    polymake_jll            v400.1100.1+0

Julia Version 1.10.1
Commit 7790d6f0641 (2024-02-13 20:41 UTC)
Build Info:
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 20 × 13th Gen Intel(R) Core(TM) i5-13600
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, goldmont)
Threads: 1 default, 0 interactive, 1 GC (on 20 virtual cores)
Environment:
  JULIA_EDITOR = vim
```

ederc: Yes, this I can confirm. The computed reduced GBs are always the same, but the ordering of the printing is different. So, for example, this is always `true`:
```
[g for g in terms(G[2], ordering=degrevlex(R))] == [h for h in terms(H[2], ordering=degrevlex(R))]
```
For `G[2]-H[2]` it is pretty clear what happens: The subtraction looks for the highest term in H[2] and keeps the first two terms of G[2] and then things mess up. I do not know how this subtraction is internally handled, but I assume it depends on the `internal_ordering` to be efficient.

joschmitt: There is a small bug in AlgebraicSolving.jl, see https://github.com/algebraic-solving/AlgebraicSolving.jl/pull/42.

ederc: Ah, I understand. We did the polynomial construction that way because it is faster when the examples get bigger. I will make a new release for `AlgebraicSolving` with your fix.

joschmitt: Thank you!

lgoettgens:
I can provide you a patch to fix the `find_name` issues. It seems some new occurrences have been introduced after https://github.com/oscar-system/Oscar.jl/pull/3327.

joschmitt:
> I can provide you a patch to fix the `find_name` issues. It seems some new occurrences have been introduced after #3327.

I assume that I just need to do `PrettyPrinting.find_name`?

joschmitt:
Otherwise be my guest.

lgoettgens:
No. The semantics have changed

joschmitt:
Please backport this. The change to `ordering` does not interfere with the book. I cannot vouch for the Hecke functions that are not exported any more unfortunately.

thofma (approved):

src/Rings/mpoly.jl
```diff
@@ -311,7 +311,7 @@ function singular_generators(B::IdealGens, monorder::MonomialOrdering=default_or
   singular_assure(B)
   # in case of quotient rings, monomial ordering is ignored so far in singular_poly_ring
   isa(B.gens.Ox, MPolyQuoRing) && return B.gens.S
-  isdefined(B, :ord) && B.ord == monorder && monomial_ordering(B.Ox, ordering(base_ring(B.S))) == B.ord && return B.gens.S
+  isdefined(B, :ord) && B.ord == monorder && monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord && return B.gens.S
```
HechtiDerLachs, line 314: @joschmitt : This change leads to a use of 30% of ~4.3s computation time in an example of @simonbrandhorst and mine. The issue seems to be that a new monomial ordering is created and then comparison can not rely on caching. Can we find a workaround for this?

joschmitt: I don't see how this change can have a performance impact? The only thing happening in this line should be that `ordering` is not exported from Singular anymore. One could also do 
```suggestion
  isdefined(B, :ord) && B.ord == monorder && monomial_ordering(B.Ox, internal_ordering(base_ring(B.S))) == B.ord && return B.gens.S
```
I think. I'm fairly certain that all I did was renaming a function.

fingolfin: @HechtiDerLachs just to clarify: when you say this change causes the slow down, how did you determine this, resp. check this? By reverting just that one change and then benchmarking again?

If so, knowing the example (or a minimized version) would be essential to figure out what's going on.

HechtiDerLachs: I have to admit that I did not revert this particular change to see whether it solves the problem. Sorry about this premature notification, @joschmitt . Either way: There seems to be an issue with the comparison of orderings here again. 

A small example will be hard to construct. But I have one on my machine right now and I'm in my office.

PR #3379, opened 2024-02-16
CI: switch to 1.11-nightly for now

benlorenz:
Nightly was changed to 1.12 and we do not have any compatible jlls yet. To keep testing with 1.11 use those nightlies for now.


lgoettgens:
We should do the same for the release branch, right?

lgoettgens (approved):

PR #3381, opened 2024-02-16
Enable edit and show source buttons in docs

### LINKED ISSUE #588: links to source code in the Oscar documentation

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/588.

In the upgrade to Documenter.jl version 1.0 in https://github.com/oscar-system/Oscar.jl/pull/2927, the Documenter cross-referencing features were disabled due to issues with the imported doc pages from AA/Nemo/Hecke.

This PR enables the "show source" buttons in all docstring blocks and resolves the URL for non-oscar docstrings based on some heuristic with `Pkg.jl` internals (if a `git_revision` is known take that, otherwise a version is known and we can take the corresponding git tag).

Furthermore, it enables the "edit page on github" buttons in the top navigation bar, which now point to the markdown file in the correct project (again based on the same heuristic). (For reviewers: Using a `symlink` instead `cp` to for the import makes it possible for Documenter to guess where a file comes from as the symlink gets resolved using `realpath`.)

Note that the links behind these buttons may be broken/useless in a local build with non-released (dev'ed) versions of dependencies. Anyway, I think this is a great addition.

A preview is available here: https://docs.oscar-system.org/previews/PR3381/

### DIFF
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -1,6 +1,6 @@
 #
 # This file is included by docs/make.jl and by a helper function
-# in src/Oscar.jl
+# in src/utils/docs.jl
 #
 module BuildDoc
 
@@ -139,7 +139,7 @@ function doit(
     joinpath(Oscar.oscardir, "docs", "oscar_references.bib"); style=oscar_style
   )
 
-  # Copy documentation from Hecke, Nemo, AnstratAlgebra
+  # Copy documentation from Hecke, Nemo, AbstractAlgebra
   other_packages = [
     (Oscar.Hecke, Oscar.heckedir),
     (Oscar.Nemo, Oscar.nemodir),
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -172,6 +172,20 @@ function doit(
     end
   end
 
+  function get_rev(uuid::Base.UUID)
+    deps = Documenter.Pkg.dependencies()
+    @assert haskey(deps, uuid)
+    if !isnothing(deps[uuid].git_revision)
+      return deps[uuid].git_revision
+    else
+      return "v$(deps[uuid].version)"
+    end
+  end
+  aarev = get_rev(Base.UUID("c3fe647b-3220-5bb0-a1ea-a7954cac585d"))
+  nemorev = get_rev(Base.UUID("2edaba10-b0f1-5616-af89-8c11ac63239a"))
+  heckerev = get_rev(Base.UUID("3e1990a7-5d81-5526-99ce-9ba3ff248f21"))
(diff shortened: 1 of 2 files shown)


benlorenz (approved):
Thanks! I checked a few pages and it seems to work very well.

docs/make_work.jl
```diff
@@ -166,12 +166,30 @@ function doit(
         end
         src = normpath(root, file)
         dst = normpath(dstbase, relpath(root, srcbase), file)
-        cp(src, dst; force=true)
+        if endswith(file, ".md")
```
fingolfin, line 169: Not an objection, just wondering: why not just symlink all the files?

docs/make_work.jl
```diff
@@ -166,12 +166,30 @@ function doit(
...
+  function get_rev(uuid::Base.UUID)
+    deps = Documenter.Pkg.dependencies()
+    @assert haskey(deps, uuid)
+    if !isnothing(deps[uuid].git_revision)
+      return deps[uuid].git_revision
+    else
+      return "v$(deps[uuid].version)"
+    end
+  end
+  aarev = get_rev(Base.UUID("c3fe647b-3220-5bb0-a1ea-a7954cac585d"))
```
fingolfin: I guess it's a matter of taste, but instead of using the UUID here, one could also query it:
```suggestion
  aarev = get_rev(Base.PkgId(AbstractAlgebra).uuid)
```
or maybe this is needed
```suggestion
  aarev = get_rev(Base.PkgId(Oscar.AbstractAlgebra).uuid)
```

fingolfin (approved):

docs/make_work.jl
lgoettgens: I didn't know this worked. Will change this in a few

docs/make_work.jl
lgoettgens: I had it first with only symlinks, but that leaded to a red wall on https://github.com/oscar-system/Oscar.jl/actions/workflows/pages/pages-build-deployment on the weekend, so it basically blocked all updates to the docs from being published.
How I try to explain it: In the building stage (converting md to HTML), Documenter is fine with files not in the project folder it is called on. However, in later stages (where e.g. `*.js` are still there), it does not like symlinks outside of the project folder anymore. And thinking again about it, that would be kind of hard to check-in into git anyway :joy:

PR #3382, opened 2024-02-16
Fix Grassmann Pluecker 

antonydellavecchia:
grassmann_pluecker_ideal returns ideal with knowledge of the the computed groebner basis

related to #3374


thofma:
I think it would be unfortunate to call the function `blabla_ideal` and not return an ideal. What about returning an ideal, which knows this Gröbner basis computed by polymake? If you agree, I can give it a try.

antonydellavecchia:
> I think it would be unfortunate to call the function `blabla_ideal` and not return an ideal. What about returning an ideal, which knows this Gröbner basis computed by polymake? If you agree, I can give it a try.

Ya that would be ideal.

thofma:
OK, this was easier than expected. I also took the liberty to clean up one of the conversions.

src/Rings/mpoly-ideals.jl
```diff
@@ -1944,33 +1952,32 @@ julia> grassmann_pluecker_ideal(2, 4)
...
+  I = grassmann_pluecker_ideal(subspace_dimension, ambient_dimension)
+  converted_generators = elem_type(ring)[]
+  coeff_ring = base_ring(ring)
+  o = degrevlex(ring)
+  coeffmap = c -> begin
+    @assert is_integral(c)
+    return coeff_ring(numerator(c))
+  end
+  h = hom(base_ring(I), ring, coeffmap, gens(ring))
+  converted_generators = [h(g) for g in groebner_basis(I; ordering = degrevlex(base_ring(I)))]
```
benlorenz:
```suggestion
  converted_generators = elem_type(ring)[h(g) for g in groebner_basis(I; ordering = degrevlex(base_ring(I)))]
```

src/Rings/mpoly-ideals.jl
```diff
@@ -1944,33 +1952,32 @@ julia> grassmann_pluecker_ideal(2, 4)
...
-        for (i, exponent) in enumerate(AbstractAlgebra.exponent_vectors(g))
-            c = converted_coeffs[i]
-            push_term!(polynomial, c, exponent)
-        end
-        converted_g = finish(polynomial)
-        push!(converted_generators, converted_g)
-    end
-    return ideal(ring, converted_generators)
+  I = grassmann_pluecker_ideal(subspace_dimension, ambient_dimension)
+  converted_generators = elem_type(ring)[]
```
benlorenz:
```suggestion
```
No need to assign this here when it is re-assigned further down.

benlorenz (approved):

PR #3384, opened 2024-02-19
Improve docstrings for `is_conjugate`/`is_conjugate_with_data`.

### LINKED ISSUE #3264: Improve `is_conjugate` and `is_conjugate_with_data` docstrings
They should ref each other. And e.g. say "to also get an element realizing the conjugation, use `is_conjugate_with_data`"   or  "if you are only interested in whether x and y are conjugate use `is_conjugate`".

mjrodgers:
Improve docstrings for `is_conjugate` and `is_conjugate_with_data` methods, in particular making sure they reference each other where relevant.

Closes #3264 .

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -710,7 +710,7 @@ end
     is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
 
 Return whether `x` and `y` are conjugate elements in `G`,
-i.e., there is an element `z in `G` such that `x^z` equals `y`.
+i.e., there is an element `z` in `G` such that `x^z` equals `y`.
 To also return the element `z`, use [`is_conjugate_with_data`](@ref).
 """
 function is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -914,7 +914,7 @@ Base.:^(H::GAPGroup, y::GAPGroupElem) = conjugate_group(H, y)
 # (The name is confusing because it is not clear *of which group* the result
 # shall be a subgroup.)
 
-"""
+@doc raw"""
     is_conjugate(G::GAPGroup, H::GAPGroup, K::GAPGroup)
 
 Return whether `H` and `K` are conjugate subgroups in `G`,
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -710,8 +710,8 @@ end
     is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
 
 Return whether `x` and `y` are conjugate elements in `G`,
-i.e., there is an element $z$ in `G` such that `x^`$z$ equals `y`.
-To also return the element $z$, use [`is_conjugate_with_data`](@ref).
+i.e., there is an element `z in `G` such that `x^z` equals `y`.
+To also return the element `z`, use [`is_conjugate_with_data`](@ref).
 """
 function is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
    if isdefined(G,:descr) && (G.descr == :GL || G.descr == :SL)
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -711,6 +711,7 @@ end
 
 Return whether `x` and `y` are conjugate elements in `G`,
 i.e., there is an element $z$ in `G` such that `x^`$z$ equals `y`.
+To also return the element $z$, use [`is_conjugate_with_data`](@ref).
 """
 function is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
    if isdefined(G,:descr) && (G.descr == :GL || G.descr == :SL)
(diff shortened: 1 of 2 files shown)


src/Groups/gsets.jl
```diff
@@ -709,9 +710,10 @@ is_conjugate(Omega::GSet, omega1, omega2) = omega2 in orbit(Omega, omega1)
     is_conjugate_with_data(Omega::GSet, omega1, omega2)
 
 Determine whether `omega1`, `omega2` are in the same orbit of `Omega`.
-If yes, return `true, g` where `g` is an element in the group `G` of
+If yes, return `(true, g)` where `g` is an element in the group `G` of
 `Omega` that maps `omega1` to `omega2`.
-If not, return `false, nothing`.
+If not, return `(false, nothing)`.
```
joschmitt, line 715: This function apparently does not return `nothing` (in the Julia sense), but a group element.

Aside: I'm wondering whether the other two should do the same in terms of type stability?

lgoettgens: I would say yes, maybe just `one(G)`

src/Groups/GAPGroups.jl
```diff
@@ -942,8 +948,10 @@ is_conjugate(G::GAPGroup, H::GAPGroup, K::GAPGroup) = GAPWrap.IsConjugate(G.X,H.
 """
     is_conjugate_with_data(G::Group, H::Group, K::Group)
 
-If `H` and `K` are conjugate subgroups in `G`, return `true, z`
-where `H^z = K`; otherwise, return `false, nothing`.
+If `H` and `K` are conjugate subgroups in `G`, return `(true, z)`
+where `H^z = K`; otherwise, return `(false, nothing)`.
```
lgoettgens, line 952: here similar

src/Groups/gsets.jl
mjrodgers: I don't like this, I think it should return `nothing`; I guess it's returning `false` anyway, but returning the identity element to me should only happen if `omega1 == omega2`.

joschmitt: Well, it's common practice in OSCAR, to return a second "dummy" element for type stability in case the first argument is `false`. But no strong argument from my side.

thofma:
If there is an issue with the return values of these methods, it should be a separate issue. This is just a quick update to the docstring here, which is independent of this discussion.

joschmitt:
> If there is an issue with the return values of these methods, it should be a separate issue. This is just a quick update to the docstring here, which is independent of this discussion.

Yes, sorry.

mjrodgers:
I see I messed up typesetting some LaTeX things so all tests are failing (although there had already been a `$z$` before I made my edits that didn't seem to mess things up?). I will fix this real quick.

src/Groups/GAPGroups.jl
```diff
@@ -914,7 +917,10 @@ Base.:^(H::GAPGroup, y::GAPGroupElem) = conjugate_group(H, y)
 """
```
thofma:
```suggestion
@doc raw"""
```
If you want to use `$` as you did.

mjrodgers: If the LaTeX is better I'm open to it, but I think it maybe didn't fit well mixed in with everything else; I ended up taking the LaTeX out (I do know how to typeset it now in the normal way though, and good to know this lets the `$` work).

src/Groups/GAPGroups.jl
```diff
@@ -710,7 +710,8 @@ end
     is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
 
 Return whether `x` and `y` are conjugate elements in `G`,
-i.e., there is an element $z$ in `G` such that `x^`$z$ equals `y`.
+i.e., there is an element `z in `G` such that `x^z` equals `y`.
```
```suggestion
i.e., there is an element `z` in `G` such that `x^z` equals `y`.
```

fingolfin:
@mjrodgers if you write `fixes ISSUENUMBER` or `resolves ISSUENUMBER` in the description of the issue, then merging the pull request will automatically close the referenced issue ([see here for details](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests))

src/Groups/GAPGroups.jl
```diff
@@ -710,7 +710,8 @@ end
     is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
 
 Return whether `x` and `y` are conjugate elements in `G`,
-i.e., there is an element $z$ in `G` such that `x^`$z$ equals `y`.
+i.e., there is an element `z` in `G` such that `x^z` equals `y`.
```
fingolfin, line 713: Here you replaced  `` `x^`$z$ `` by `` `x^z` `` (I think @ThomasBreuer used the former deliberately, though one may argue about that).

Below you introduced a new occurrence of `` `H^`$z$ `` in the `is_conjugate(G::GAPGroup, H::GAPGroup, K::GAPGroup)` docstring.

Finally in the `is_conjugate_with_data(G::Group, H::Group, K::Group)` docstring very similar new text was added by here with `` `H^z = K` ``

I have no strong opinion which to use, as long as their is some consistency or rule

mjrodgers: Yeah, I guess I felt like it looked strange when the the z needed to appear as a literal (ex "returns `(true, z)`") to have it typeset in LaTeX in another part of the docstring; I'll make these consistent (no LaTeX).

mjrodgers:
I believe this should be ready to merge

fingolfin:
Fine by me, but it would be good if @ThomasBreuer could have a look and approve (or request changes)

ThomasBreuer (approved):
Looks good, thanks.

PR #3387, opened 2024-02-19
Rename `vec` for AbstractAlgebra matrices to `_vec`

lgoettgens:
As suggested by @thofma in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1614#issuecomment-1952429263.

In the initial commit, I only touched the function itself and everything related in `experimental/GModule`. All failures besides that are due to code relying on functionality in `src` and have to be dealt with on a case-by-case basis.

The easiest would be to move `_vec` out of experimental and adapt the call-sites to use `_vec` instead of `vec`. What do you think?

cc @thofma @fieker


fingolfin (approved):
Also fine by me.

lgoettgens:
This now seems to work.

thofma (approved):

Issue #697, opened 2021-09-24
oscar needs speed regression tests


tthsqe12:
candlesticks. Changes to one part of the system can have huge unintended performance consequences for other parts of the system. We need a set S of examples to run to check for regressions every so often. If we get a situation of the form "I added feature X to part Y, but this slowed down S by more than 2x", then we can say the performance regression needs to be addressed before X is added.

fingolfin:
@aaruni96 we've discuss this before, it's yet another thing I'd like us to take care of at some point...

Some thoughts:
- we need a machine on which we can run these tests which is otherwise not under heavy load... That makes it a bit problematic to use our regular CI servers. Perhaps we could use `munk` (an Apple Mac Mini M1 we sometimes use for debugging, but probably not at, say, 3:00 in the night)
- https://github.com/MilesCranmer/AirspeedVelocity.jl sounds useful
- https://github.com/python/codespeed/ might be useful as a web frontend (its master is from 5 years ago but it has active branches driving e.g. https://speed.pypy.org
- but of course one probably also use Grafana or other tools
- - we need to select a list of "benchmarks". While we could use the entire test suite, that might not be ideal, and those two have somewhat different goals... That said, we gotta start somewhere, and at the beginning one could just use a few hand curated benchmarks
- similar setups for our other packages might be useful to have as well!

fingolfin:
Julia now has [a page tracking timings](https://juliaci.github.io/julia-ci-timing/?s=lb.lt.lc.mb.mt.mc.wb.wt.wc.fb.ft&x=1764703618406.1767903618406&y=-3245.39965&l=7), see also <https://github.com/JuliaCI/julia-ci-timing>, perhaps we can borrow from that?

aaruni96:
Looking into julia-ci-timing, it looks to be something like an overly complicated HTML file reading data from a yaml file.

The yaml file roughly needs to  look like the following, and then their data can be swapped out for data we want to display:

```yml
{
    "generated_at": "2026-01-21T14:22:35Z",
    "jobs": {
        ":linux: test 1.12 short": {
            "recent": [
                    {
                        "agent": "blah",
                        "author": "blah",
                        "build": 55567,
                        "commit": "54f3afd",
                        "duration": 500,
                        "message": "stuff",
                        "pipeline": "master",
                        "retry": 0,
                        "state": "passed",
                        "date": "2026-01-21 05:44"

                    },
                    {
                        "agent": "blah",
                        "author": "blah",
[...]

PR #3389, opened 2024-02-19
Convert partition methods to return iterators

mjrodgers:
With the recent migration of combinatorics functions from `experimental` to `src` in #3159 , there is some work to be done converting the methods to return iterators instead of very large collections of combinatorial objects. This is work to address #3147 

Here we are converting the methods that focus on returning integer partitions.

There are 3 different "batches" of partition methods that need to be converted in this way.

1. the basic `partitions(n)` for computing all integer partitions of n - this has been converted
2. the methods `partitions(n,m)` (also optionally with upper/lower bounds) for partitioning into a fixed number of parts - this has been converted.
3. the methods `partitions(n,v)` where v is a vector of allowed values to use in the partition (with options to specify the number of parts, or to restrict the multiplicities of the allowed values) - this one still needs to be converted.

### DIFF
--- a/src/Combinatorics/EnumerativeCombinatorics/compositions.jl
+++ b/src/Combinatorics/EnumerativeCombinatorics/compositions.jl
@@ -152,11 +152,11 @@ parts(C::CompositionsFixedNumParts) = C.k
 Base.eltype(C::CompositionsFixedNumParts{T}) where T = Composition{T}
 
 function Base.show(io::IO, C::CompositionsFixedNumParts)
-  if get(io, :supercompact, false)
-    print(io, "Iterator")
+  if is_terse(io)
+    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "parts"))
   else
     io = pretty(io)
-    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "part"))
+    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "parts"))
   end
 end
 
@@ -218,8 +218,8 @@ base(C::Compositions) = C.n
 Base.eltype(C::Compositions{T}) where T = Composition{T}
 
 function Base.show(io::IO, C::Compositions)
-  if get(io, :supercompact, false)
-    print(io, "Iterator")
+  if is_terse(io)
+    print(io, "Iterator over the compositions of $(base(C))")
   else
     io = pretty(io)
     print(io, "Iterator over the compositions of $(base(C))")
@@ -304,8 +304,8 @@ base(C::AscendingCompositions) = C.n
 Base.eltype(C::AscendingCompositions{T}) where T = Composition{T}
 
 function Base.show(io::IO, C::AscendingCompositions)
-  if get(io, :supercompact, false)
-    print(io, "Iterator")
+  if is_terse(io)
+    print(io, "Iterator over the ascending compositions of $(base(C))")
   else
     io = pretty(io)
     print(io, "Iterator over the ascending compositions of $(base(C))")
--- a/src/Combinatorics/EnumerativeCombinatorics/compositions.jl
+++ b/src/Combinatorics/EnumerativeCombinatorics/compositions.jl
@@ -152,12 +152,7 @@ parts(C::CompositionsFixedNumParts) = C.k
 Base.eltype(C::CompositionsFixedNumParts{T}) where T = Composition{T}
 
 function Base.show(io::IO, C::CompositionsFixedNumParts)
-  if is_terse(io)
-    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "parts"))
-  else
-    io = pretty(io)
-    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "parts"))
-  end
+  print(pretty(io), "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "part"))
 end
 
 Base.length(C::CompositionsFixedNumParts) = BigInt(number_of_compositions(base(C), parts(C)))
@@ -218,12 +213,7 @@ base(C::Compositions) = C.n
 Base.eltype(C::Compositions{T}) where T = Composition{T}
 
 function Base.show(io::IO, C::Compositions)
-  if is_terse(io)
-    print(io, "Iterator over the compositions of $(base(C))")
-  else
-    io = pretty(io)
-    print(io, "Iterator over the compositions of $(base(C))")
-  end
+  print(pretty(io), "Iterator over the compositions of $(base(C))")
 end
 
 Base.length(C::Compositions) = BigInt(number_of_compositions(base(C)))
@@ -304,12 +294,7 @@ base(C::AscendingCompositions) = C.n
 Base.eltype(C::AscendingCompositions{T}) where T = Composition{T}
 
 function Base.show(io::IO, C::AscendingCompositions)
-  if is_terse(io)
-    print(io, "Iterator over the ascending compositions of $(base(C))")
-  else
-    io = pretty(io)
-    print(io, "Iterator over the ascending compositions of $(base(C))")
-  end
+  print(pretty(io), "Iterator over the ascending compositions of $(base(C))")
 end
 
 # Ascending compositions are basically partitions turned around
--- a/src/Combinatorics/EnumerativeCombinatorics/partitions.jl
+++ b/src/Combinatorics/EnumerativeCombinatorics/partitions.jl
@@ -373,6 +373,11 @@ function partitions(n::IntegerUnion, k::IntegerUnion; only_distinct_parts::Bool 
 end
 
 
+
+# Algorithm "parta" in [RJ76](@cite), de-gotoed from old ALGOL 60 code by E. Thiel.
+
+# Note that the algorithm is given as partitioning m into n parts,
+# but we have refactored to align more closely with standard terminology.
 function Base.iterate(P::PartitionsFixedNumParts{T}) where T
   n = P.n
   k = P.k
@@ -380,7 +385,6 @@ function Base.iterate(P::PartitionsFixedNumParts{T}) where T
   ub = P.ub
   only_distinct_parts = P.distinct_parts
 
-  # TODO : fix the states returned once we know what those will look like
   if n == 0 && k == 0
     return partition(T[], check=false), (T[], T[], 0, 0, 1, false)
   end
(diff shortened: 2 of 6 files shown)


src/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -182,54 +188,55 @@ julia> collect(partitions(Int8(4))) # using less memory
...
-  if n == 0
-    return (p for p in Partition{T}[ partition(T[], check = false) ])
-  elseif n == 1
-    return (p for p in Partition{T}[ partition(T[1], check = false) ])
+
+function Base.iterate(P::PartitionSet{T}) where T
+  if P.n == 0
+    return Partition{T}[ partition(T[]) ], (T[], 0, 0)
+  elseif P.n == 1
+    return Partition{T}[ partition(T[1]) ], (T[1], 1, 0)
```
joschmitt:
```suggestion
    return partition(T[]), (T[], 0, 0)
  elseif P.n == 1
    return partition(T[1]), (T[1], 1, 0)
```

src/Combinatorics/EnumerativeCombinatorics/types.jl
```diff
@@ -8,6 +8,10 @@ struct Partition{T<:IntegerUnion} <: AbstractVector{T}
   p::Vector{T}
 end
 
+struct PartitionSet{T<:IntegerUnion}
```
joschmitt: Maybe this should be called `PartitionIterator`? But not really important. We have to see whether one can use the same type for the restricted partitions (with potentially unused fields) or whether there need to be several types.

joschmitt (dismissed):
Looks good to me so far, but I'm not really an expert on iterators either. Thank you for working on this!

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
mjrodgers: thanks, I fixed this the same way before seeing your remark once I saw the failing tests

src/Combinatorics/EnumerativeCombinatorics/types.jl
mjrodgers: I don't know that it matters to me either what it is called; I am hopeful that I can use the same type for all of them, we will see I guess.

mjrodgers: I renamed to `Partitions` because I guess it is not really a set; I don't really like including `Iterator` in the type name, but if this is the OSCAR standard for iterator types I am okay with it and can make that change.

joschmitt: Sounds good to me. I'm not aware of any OSCAR standard regarding names for iterators.

src/Combinatorics/EnumerativeCombinatorics/types.jl
```diff
@@ -134,6 +134,51 @@ struct Partition{T<:IntegerUnion} <: AbstractVector{T}
   p::Vector{T}
 end
 
+# Iterator type
```
fingolfin: Maybe expand the comment a little bit to say what this represents?

src/Combinatorics/EnumerativeCombinatorics/types.jl
```diff
@@ -134,6 +134,51 @@ struct Partition{T<:IntegerUnion} <: AbstractVector{T}
...
+  distinct_parts::Bool
+
+  function PartitionsFixedNumParts(n::T, m::IntegerUnion, l1::IntegerUnion, l2::IntegerUnion, only_distinct_parts::Bool) where T<:IntegerUnion
+    @req n >= 0 "n >= 0 required"
+    @req m >= 0 "m >= 0 required"
+    @req l1 >= 0 "l1 >=0 required"
+    if l1 == 0
+      l1 = 1
+    end
+    return new{T}(n, convert(T, m), T(l1), T(l2), only_distinct_parts)
```
fingolfin: So l1 and l2 are used to initialize lb and ub? Why not call the arguments lb and ub, too?

Should lb <= ub hold?

src/Combinatorics/EnumerativeCombinatorics/types.jl
```diff
@@ -134,6 +134,51 @@ struct Partition{T<:IntegerUnion} <: AbstractVector{T}
...
+# TODO: This should also have:
+#  - a flag for `only_distinct_parts`
+#  - support for upper/lower bounds
+struct PartitionsFixedNumParts{T<:IntegerUnion}
+  n::T
+  m::Int
+
+  lb::T
+  ub::T
+  distinct_parts::Bool
```
fingolfin: Here, too, a few comments that explain what everything is (about) would be nice

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -123,6 +125,39 @@ function getindex_safe(P::Partition{T}, i::IntegerUnion) where T
   return (i > length(data(P)) ? zero(T) : getindex(data(P), Int(i)))
 end
 
+base(P::Partitions) = P.n
+
+Base.eltype(::Partitions{T}) where T = Partition{T}
+
+function Base.show(io::IO, ::MIME"text/plain", P::Partitions)
+  if is_terse(io)
+    print(io, "Iterator")
```
fingolfin:
```suggestion
    print(io, "Iterator over partitions")
```
Actually this function is a case where I am tempted to say the terse version makes no sense, the regular oneline version is "terse enough" in all real world examples.


Wonder what @simonbrandhorst @thofma @fieker think

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -123,6 +125,39 @@ function getindex_safe(P::Partition{T}, i::IntegerUnion) where T
...
+  else
+    print(io, "Iterator over the partitions of $(base(P)) into $(P.m) parts")
+  end
+end
+
+# NOTE this will not be accurate in many cases,
+# in particular if upper/lower bounds are given,
+# or if `only_distinct_parts == true`.
+# Base.length(P::PartitionsFixedNumParts) = BigInt(number_of_partitions(P.n, P.m))
+Base.IteratorSize(::Type{PartitionsFixedNumParts{T}}) where T = Base.SizeUnknown()
```
fingolfin, line 154:
```suggestion

Base.IteratorSize(::Type{PartitionsFixedNumParts{T}}) where T = Base.SizeUnknown()
```

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -185,91 +221,93 @@ julia> collect(partitions(Int8(4))) # using less memory
  Int8[1, 1, 1, 1]
 ```
 """
-function partitions(n::T) where T <: IntegerUnion
+function partitions(n::IntegerUnion)
```
fingolfin, line 218: Off-topic, but: I just noticed a function `partitions(n::Int, k::Int=n, m::Int=-1)` in `experimental/IntersectionTheory/src/Misc.jl` which of course clashes with the code here.

Which makes me wonder whether that function has any business sitting in `experimental/IntersectionTheory/src/Misc.jl`, and the whole file has a bunch of "interesting" content... huhrm

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
lgoettgens: This should be kept consistent with the iterators for Compostions and WeakCompositions (in any case od decision)

joschmitt: Yeah, I didn't know how short/terse a `terse` version is supposed to be in the other iterators.

src/Combinatorics/EnumerativeCombinatorics/types.jl
mjrodgers: The existing code used l1 and l2, though I think I prefer to use lb and ub - I can refactor to do this (while I'm at it, I will make it partitions of n into k parts instead of m parts).

Again following the preexisting code, we do not require that lb <= ub, but if this is not the case we return an empty iterator.

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
mjrodgers: yeah, I copied this from (Weak)Compositions for consistency, but I do think we should make them both just a little less terse.

test/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -52,118 +55,105 @@
...
 
-          # Check that P and Q coincide (up to reordering)
-          @test length(P) == length(Q)
-          @test Set(P) == Set(Q)
-        end
-      end
-    end
+    # Check that P and Q coincide (up to reordering)
+    @test length(P) == length(Q)
+    @test Set(P) == Set(Q)
```
lkastner: This loop (and probably the ones above too) creates a ton of output, see e.g.
https://github.com/oscar-system/Oscar.jl/actions/runs/9093356754/job/24992087132?pr=3389
Can this maybe be suppressed by wrapping it in another `@testset` or something similar? GitHub even refuses to give the full log initially since it is so large.

I also find this kind of a testing against itself a bit dubious if it happens at this scale.

lkastner:
Another question I have: You seem to rely on this being fast. Are there some timing tests to make sure it stays fast?

test/Combinatorics/EnumerativeCombinatorics/partitions.jl
lgoettgens: The printing of the test set names here doesn't bring much benefit in any case. If something fails, you get thousands of lines with the list of testsets that worked/failed. Maybe just disentangle the test set and the for

fingolfin (approved):

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
lgoettgens: Triage agrees with @fingolfin

test/Combinatorics/EnumerativeCombinatorics/partitions.jl
mjrodgers: I had to wrap this because I was getting failing tests and it was impossible to see which ones were actually failing. I feel like I do need to see the test set names for the failing tests. But I agree with lkastner, I will split this into a nested `@testset`

mjrodgers: @lkastner  Okay, now this testset is split, and I also reduced the scale since I also think it's silly to test quite this many cases. Let me know if you think this is better now.

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
lgoettgens: @mjrodgers can you apply the same printing change as in [`376e9ec` (#3389)](https://github.com/oscar-system/Oscar.jl/pull/3389/commits/376e9ec1b506e6466834866d68a5a6dbd68b658a) for Compositions and WeakCompositions as well?

mjrodgers: sure, done

joschmitt:
> Another question I have: You seem to rely on this being fast. Are there some timing tests to make sure it stays fast?

This is an ancient issue #697...

Also, in this case the comparison is not completely fair. From my experience with the other combinatoric's things, calling `collect` on the iterator is a bit slower than just generating the whole array in one go. Without being the target audience for these functions, I would argue that one usually doesn't call `collect`, though, and the iterator is much more versatile.

src/Combinatorics/EnumerativeCombinatorics/compositions.jl
```diff
@@ -152,11 +152,11 @@ parts(C::CompositionsFixedNumParts) = C.k
...
 function Base.show(io::IO, C::CompositionsFixedNumParts)
-  if get(io, :supercompact, false)
-    print(io, "Iterator")
+  if is_terse(io)
+    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "parts"))
   else
     io = pretty(io)
-    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "part"))
+    print(io, "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "parts"))
   end
```
joschmitt: This should just be
```
print(pretty(io), "Iterator over the compositions of $(base(C)) into ", ItemQuantity(parts(C), "part"))
```
without the `if`/`else`. And the `ItemQuantity` takes care of the plural s, so `part` is correct here :)

Same for the weak compositions.

lkastner:
> > Another question I have: You seem to rely on this being fast. Are there some timing tests to make sure it stays fast?
> 
> This is an ancient issue #697...
> 
> Also, in this case the comparison is not completely fair. From my experience with the other combinatoric's things, calling `collect` on the iterator is a bit slower than just generating the whole array in one go. Without being the target audience for these functions, I would argue that one usually doesn't call `collect`, though, and the iterator is much more versatile.

I didn't mean to compare to the older version, I simply wanted to recommend adding some performance tests now. They could just be for the iterator without `collect`. Nevertheless, I'll leave that decision to you.

src/Combinatorics/EnumerativeCombinatorics/compositions.jl
mjrodgers: ah okay, my mistake. fixed now

joschmitt: Thank you!

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -123,6 +125,33 @@ function getindex_safe(P::Partition{T}, i::IntegerUnion) where T
...
+Base.length(P::Partitions) = BigInt(number_of_partitions(P.n))
+
+
+
+base(P::PartitionsFixedNumParts) = P.n
+
+Base.eltype(::PartitionsFixedNumParts{T}) where T = Partition{T}
+
+function Base.show(io::IO, ::MIME"text/plain", P::PartitionsFixedNumParts)
+  print(pretty(io), "Iterator over the partitions of $(base(P)) into",
```
lgoettgens:
```suggestion
  print(pretty(io), "Iterator over the partitions of $(base(P)) into ",
```

fingolfin:
@lkastner how do you envision performance tests would work? I.e. who would run them to test for performance regressions? We can't do this in CI, as timings fluctuate wildly.

lkastner:
> @lkastner how do you envision performance tests would work? I.e. who would run them to test for performance regressions? We can't do this in CI, as timings fluctuate wildly.

But we already do run timing tests in CI? https://github.com/oscar-system/Oscar.jl/blob/master/test/PolyhedralGeometry/timing.jl

src/Combinatorics/EnumerativeCombinatorics/partitions.jl
```diff
@@ -332,121 +364,356 @@ julia> collect(partitions(7, 3, 1, 4; only_distinct_parts = true))
...
+    i += 1
+  end
+  x[i] = y[i] + N
+  return partition(x[1:k], check = false), (x,y,N,L2,i,true)
 end
 
-function partitions(m::T, n::IntegerUnion; only_distinct_parts::Bool = false) where T <: IntegerUnion
-  @req m >= 0 "m >= 0 required"
+
+# function partitions(n::T, k::IntegerUnion, lb::IntegerUnion, ub::IntegerUnion; only_distinct_parts::Bool = false) where T <: IntegerUnion
```
fingolfin: Any convincing reason to keep this commented out code? Otherwise just delete it...?

mjrodgers: I migrated some of the comments, and deleted the code.

lgoettgens:
I think https://github.com/oscar-system/Oscar.jl/pull/3389#discussion_r1600763120 is the only unresolved comment above.

joschmitt:
> I think [#3389 (comment)](https://github.com/oscar-system/Oscar.jl/pull/3389#discussion_r1600763120) is the only unresolved comment above.

I'll assume that one can replace the `partitions` method form intersection theory with one of the established ones. But this is unrelated to this pull request; I will open a new one. Let's merge this.

PR #3392, opened 2024-02-19
Eliminate more spelling differences between American and British English

lgoettgens:
Follow-up to https://github.com/oscar-system/Oscar.jl/pull/3367 with even more things I found.

This introduces two new deprecations, namely:

- `labelled_matrix_formatted` -> `labeled_matrix_formatted`
- `recognise` -> `recognize` (in the `SolveRadical` submodule)

Can someone with access to the book please check, if these deprecations are an issue?


joschmitt:
The functions are not used in the book. But a general question: I understand that we want to have consistency with function names, but is the policy now to enforce American spelling in the documentation too?

fingolfin (approved):
I am fine with that, but if this is now the standard, we should document it at such, too, if we don't already (not a blocker for this PR, though)

lgoettgens:
If we want to have these two function renamings in 1.0, please go ahead with merging.

Issue #3390, opened 2024-02-19
Support `character(group, values)`?


joschmitt:
I would like to have an exported version of the function `Oscar.class_function(::Oscar.GAPGroup, ::Vector{<:QQAbElem})` without the need to work in the abelian closure on the user side. So something like `character(::GAPGroup, ::Vector{<:FieldElem})`. Is there any reason against this?

CC @ThomasBreuer

ThomasBreuer:
Defining a class functions over a field F requires to define how F embeds into a suitable cyclotomic field.
Several such embeddings are possible, we have to make sure that the choices are compatible.

fingolfin:
@joschmitt Would you expect this function to perform input validation? I.e. should it check that the given vector actually defines a character? This problem does not exist for `class_function`, of course.

joschmitt:
I think it should check the input, yes. We can also call it `class_function`. I guess, I haven't really thought about this very well.

thofma:
The interface will be awkward, since one has to construct the character table first to get the right order for the conjugacy classes. But then I could have called `class_funtion(::GAPGroupCharacterTable, values)` directly.

joschmitt:
Yeah, so all of this answers my question why the function does not exist.

PR #3393, opened 2024-02-20
feat: add `regular_character`

### LINKED ISSUE #3107: dualizing does not work for free resolutions

thofma:
CC: @joschmitt

### DIFF
--- a/docs/src/Groups/group_characters.md
+++ b/docs/src/Groups/group_characters.md
@@ -151,7 +151,7 @@ natural_character(G::Union{MatrixGroup{QQFieldElem}, MatrixGroup{AbsSimpleNumFie
 natural_character(G::MatrixGroup{T, MT}) where T <: FinFieldElem where MT
 natural_character(rho::GAPGroupHomomorphism)
 trivial_character(G::GAPGroup)
-regular_character(G::GAPGRoup)
+regular_character(G::GAPGroup)
 ```
 
 ## Operations for group characters
--- a/docs/src/Groups/group_characters.md
+++ b/docs/src/Groups/group_characters.md
@@ -140,6 +140,7 @@ orders_centralizers
 orders_class_representatives
 ordinary_table(tbl::GAPGroupCharacterTable)
 trivial_character(tbl::GAPGroupCharacterTable)
+regular_character(tbl::GAPGroupCharacterTable)
 ```
 
 ## Construct group characters from groups
@@ -150,6 +151,7 @@ natural_character(G::Union{MatrixGroup{QQFieldElem}, MatrixGroup{AbsSimpleNumFie
 natural_character(G::MatrixGroup{T, MT}) where T <: FinFieldElem where MT
 natural_character(rho::GAPGroupHomomorphism)
 trivial_character(G::GAPGroup)
+regular_character(G::GAPGRoup)
 ```
 
 ## Operations for group characters
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -1741,7 +1741,7 @@ julia> values(regular_character(G))
 ```
 """
 function regular_character(G::GAPGroup)
-  return regular_chacter(character_table(G))
+  return regular_character(character_table(G))
 end
 
 @doc raw"""
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -1751,7 +1751,7 @@ Return the regular character of `tbl`.
 
 # Examples
 ```jldoctest
-julia> tbl = character_table(symmetric_group(3))
+julia> tbl = character_table(symmetric_group(3));
 
 julia> values(regular_character(tbl))
 3-element Vector{QQAbElem{AbsSimpleNumFieldElem}}:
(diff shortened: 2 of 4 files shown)


src/Groups/group_characters.jl
```diff
@@ -1724,6 +1724,28 @@ function trivial_character(G::GAPGroup)
...
+
+julia> values(regular_character(G))
+3-element Vector{QQAbElem{AbsSimpleNumFieldElem}}:
+ 6
+ 0
+ 0
+```
+"""
+function regular_character(G::GAPGroup)
+  ccl = conjugacy_classes(G)
```
thofma: @ThomasBreuer is this correct? I saw that the other functions used `conjugacy_classes(character_table(G))`.

thofma:
@ThomasBreuer What do you think of exporting `class_function`?

src/Groups/group_characters.jl
thofma: OK, I read some docstring and understand why I need the character table

joschmitt:
This would resolve #3390 assuming we export `class_function`, right?

thofma:
Exporting `class_function` would be more complicated, so let's keep this PR here focused on adding `regular_character`.

ThomasBreuer (commented):
Actually `regular_character` can assume that the class of the identity element is the first one, thus the function need not do computations with the group.
(I should state this fact in the docstring for `conjugacy_classes` for character tables.)

What about providing `regular_character(tbl::GAPGroupCharacterTable)`,
and defining `regular_character(G::GAPGroup) = regular_character(character_table(G))`?

fingolfin:
I think what @ThomasBreuer suggests would be the way to go. Of course we can also do that later if you need this in immediately.

thofma:
Adjusted following your comments. Please have a look again @ThomasBreuer

src/Groups/group_characters.jl
```diff
@@ -1724,6 +1724,48 @@ function trivial_character(G::GAPGroup)
...
+```
+"""
+function regular_character(G::GAPGroup)
+  return regular_chacter(character_table(G))
+end
+
+@doc raw"""
+    regular_character(tbl::GAPGroupCharacterTable)
+
+Return the regular character of `G`.
```
ThomasBreuer: "of `tbl`"

ThomasBreuer (approved):
Looks good. Thanks. Just one minor comment on a docstring.

src/Groups/group_characters.jl
thofma:
```suggestion
Return the regular character of `tbl`.
```

src/Groups/group_characters.jl
```diff
@@ -1724,6 +1724,48 @@ function trivial_character(G::GAPGroup)
...
+
+julia> values(regular_character(G))
+3-element Vector{QQAbElem{AbsSimpleNumFieldElem}}:
+ 6
+ 0
+ 0
+```
+"""
+function regular_character(G::GAPGroup)
+  return regular_chacter(character_table(G))
```
thofma:
```suggestion
  return regular_character(character_table(G))
```

docs/src/Groups/group_characters.md
```diff
@@ -150,6 +151,7 @@ natural_character(G::Union{MatrixGroup{QQFieldElem}, MatrixGroup{AbsSimpleNumFie
 natural_character(G::MatrixGroup{T, MT}) where T <: FinFieldElem where MT
 natural_character(rho::GAPGroupHomomorphism)
 trivial_character(G::GAPGroup)
+regular_character(G::GAPGRoup)
```
thofma:
```suggestion
regular_character(G::GAPGroup)
```

src/Groups/group_characters.jl
```diff
@@ -1724,6 +1724,48 @@ function trivial_character(G::GAPGroup)
...
+julia> values(regular_character(tbl))
+3-element Vector{QQAbElem{AbsSimpleNumFieldElem}}:
+ 6
+ 0
+ 0
+```
+"""
+function regular_character(tbl::GAPGroupCharacterTable)
+  val = ZZRingElem[0 for i in 1:length(tbl)]
+  val[1] = order(G)
```
thofma:
```suggestion
  val[1] = order(group(tbl))
```

fingolfin (approved):

src/Groups/group_characters.jl
```diff
@@ -1724,6 +1724,48 @@ function trivial_character(G::GAPGroup)
...
+end
+
+@doc raw"""
+    regular_character(tbl::GAPGroupCharacterTable)
+
+Return the regular character of `tbl`.
+
+# Examples
+```jldoctest
+julia> tbl = character_table(symmetric_group(3))
```
fingolfin:
```suggestion
julia> tbl = character_table(symmetric_group(3));
```

PR #3397, opened 2024-02-20
Make `tempproject` available in `test_experimental_module` and add docs

lgoettgens:
Follow-up to https://github.com/oscar-system/Oscar.jl/pull/3357 that makes the new kwarg available in `test_experimental_module` as well, and adds it to both docstrings.

### DIFF
--- a/src/utils/tests.jl
+++ b/src/utils/tests.jl
@@ -112,6 +112,10 @@ With the optional parameter `timed` the function will return a dict mapping file
 names to a named tuple with compilation times and allocations.
 This only works for `new=false`.
 
+The optional parameter `tempproject` decides whether a temporary project with the
+test environment is used. Setting this to `false` requires all test dependencies
+to be installed in the current environment stack.
+
 The parameter `ignore` can be used to pass a list of `String` or `Regex` patterns.
 Test files or folders matching these will be skipped. Strings will be compared as
 suffixes.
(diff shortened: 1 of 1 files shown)


benlorenz:
It was not really meant to be user-facing, the default (true) should be preferred in general, except when running from the `runtests.jl` where everything is run via `test_module` anyway. Why do you need this?

lgoettgens:
> Why do you need this?

to easily `]dev` test dependencies

and the default `true` makes a lot of output, but that's a different issue. (can be resolved by setting `io = devnull` in the `Pkg` commands, but only since julia 1.9)

benlorenz:
> > Why do you need this?
> 
> to easily `]dev` test dependencies

We could copy a `Manifest.toml` (if it exists) together with the `Project.toml`, that would preserve the `dev` dependencies from the `test` folder. I was reluctant to do that because some people tend to have old / broken manifests lying around and getting confused when stuff doesn't work as expected.
(Or we could instantiate the test folder instead but that will create a new manifest there that might cause problems later ...)

If you specify an absolute path then `test_module` can test experimental stuff as well, e.g:
```julia
Oscar.test_module(Oscar.oscardir*"/experimental/OrthogonalDiscriminants/test"; new=false, tempproject=false)
```

> and the default `true` makes a lot of output, but that's a different issue. (can be resolved by setting `io = devnull` in the `Pkg` commands, but only since julia 1.9)

That is indeed annoying, we could also redirect the output, e.g.
```julia
redirect_stderr(devnull) do 
  Pkg.resolve()
end
```
but that would hide errors as well.

lgoettgens:
modifying the path to be able to use `test_module` seems like a fine solution for now, if the kwarg shouldn't be user facing.

Issue #3395, opened 2024-02-20
Provide suitable credits for group databases (and for other GAP packages we use -- and Singular, Polymake, ... ???)


fingolfin:
We provide easy access to various group databases provided by GAP and various GAP packages.

I think we should provide some more credit to the authors of these databases in the manual. Possibly also in a docstring assigned to a "database object" if we introduce those (see #1168).

Writing this now, actually we probably also should explicitly list and credit the GAP packages we use directly, which includes these plus `ferret`, `PackageManger` and possibly some more we use indirectly (e.g. `polycyclic`?)
```
  # We need some GAP packages.
  for pkg in [
     "atlasrep",
     "ctbllib",  # character tables
     "crisp",    # faster normal subgroups, socles, p-socles for finite solvable groups
     "fga",      # dealing with free groups
     "forms",    # bilinear/sesquilinear/quadratic forms
     "packagemanager", # has been loaded already by GAP.jl
     "polycyclic", # needed for Oscar's pc groups
     "primgrp",  # primitive groups library
     "repsn",    # constructing representations of finite groups
     "smallgrp", # small groups library
     "transgrp", # transitive groups library
     "wedderga", # provides a function to compute Schur indices
     ]
    GAP.Packages.load(pkg) || error("cannot load the GAP package $pkg")
  end
  # We want some GAP packages. (It is no error if they cannot be loaded.)
  for pkg in [
     "ferret",   # backtrack in permutation groups
     ]
    GAP.Packages.load(pkg)
  end
```

CC @ThomasBreuer 

Edit: Updated list after merging #3398, #3421

ThomasBreuer:
> and possibly some more we use indirectly (e.g. `polycyclic`?)

Good point.
The GAP operation `UpdatePolycyclicCollector` is used in Oscar, it belongs to the `polycyclic` package, and there is no explicit `GAP.Packages.load("polycyclic")` in Oscar.

On the one hand, this is currently not a problem because `"polycyclic"` is in the default list of package names used by GAP's `AutoloadPackages`. On the other hand, it would be safer to get an error message in Oscar if `polycyclic` cannot be loaded.

PR #2689, opened 2023-08-17
Add Coxeter Groups and Root Systems

felix-roehrich:
I am currently working on implementing https://github.com/ulthiel/CoxeterGroups.jl (see https://github.com/oscar-system/Oscar.jl/pull/2217). To have compatibility with root systems and since many efficient algorithms for Weyl groups rely on geometric properties, I have also written a first draft for root systems in OSCAR.
Currently this is still under heavy development, especially distributing logic between root systems and Weyl groups.
At the moment Cartan matrices are implemented as ZZMatrix, but it might make sense to implement it as a sparse matrix instead.
Documentation and tests are also missing.

For simplicity everything lives in experimental/LieAlgebras but everything from Coxeter groups that does not depend on root systems should probably moved to the Groups folder at some point.

PR #3398, opened 2024-02-20
Remove unneeded GAP package `sla`

lgoettgens:
When reading through https://github.com/oscar-system/Oscar.jl/issues/3395, I noticed that `sla` has been added in https://github.com/oscar-system/Oscar.jl/pull/2572, but I don't think it is used anywhere anymore (e.g. as https://github.com/oscar-system/Oscar.jl/pull/2572 has been majorly refactored in https://github.com/oscar-system/Oscar.jl/pull/2689).


ThomasBreuer (approved):
I see.
The relevant GAP code (for example in `cartan_matrix`) has been reimplemented in Julia.
Good.

PR #3399, opened 2024-02-20
Fix an equality check

joschmitt:
Replace an `==` by an `===`. The `==` actually tries to do math and is super slow.


thofma (approved):

thofma:
Adding backport label, since we need this for some upcoming OSCAR article.

joschmitt:
I have no idea why this one doctest fails...

PR #3400, opened 2024-02-20
Adresses issue with equality/hashing on module orderings

### LINKED ISSUE #3347: Equality check and hashing for module orderings is broken
I just witnessed this on the latest master (commit a0913ed719ece74a890e19120670cd79dbffc3aa):
```
julia> R, (x,y,z,w) = polynomial_ring(GF(32003), ["x", "y", "z", "w"]);

julia> F = FreeMod(R, 2)
Free module of rank 2 over Multivariate polynomial ring in 4 variables over GF(32003)

julia> o = lex(gens(F))*lex(gens(R))
lex([gen(1), gen(2)])*lex([x, y, z, w])

julia> oo = lex(gens(F))*lex(gens(R))
lex([gen(1), gen(2)])*lex([x, y, z, w])

julia> o == oo
false

julia> hash(o) == hash(oo)
false
```
It is not clear to me whether this is a recent regression or whether this has always been an issue. In any case, this is a serious problem, because caching of Groebner bases does not work. We should probably fix this soon.

CC: @jankoboehm , @wdecker

RafaelDavidMohr:
Adds `==` and hashing for `ModuleOrdering`s.

I don't know if this is the best solution, I essentially just circumvent what julia does by default when calling `==` on structs (which is to call `===` on every field) by instead calling `==` recursively on every field.

If this is potentially to dangerous, let me know. It does fix https://github.com/oscar-system/Oscar.jl/issues/3347 on my machine.

Tag: @ederc

### DIFF
--- a/src/Rings/orderings.jl
+++ b/src/Rings/orderings.jl
@@ -1783,7 +1783,11 @@ end
 
 function canonical_matrix(o::ModuleOrdering)
   if !isdefined(o, :canonical_matrix)
-    o.canonical_matrix = _canonical_matrix_intern(o)
+    if isone(ngens(o.M))
+      o.canonical_matrix = canonical_matrix(induced_ring_ordering(o))
+    else
+      o.canonical_matrix = _canonical_matrix_intern(o)
+    end
   end
   return o.canonical_matrix
 end
--- a/src/Rings/orderings.jl
+++ b/src/Rings/orderings.jl
@@ -1728,6 +1728,10 @@ mutable struct ModOrdering{T} <: AbsModOrdering
    end
 end
 
+function ==(o1::ModOrdering, o2::ModOrdering)
+  return o1.gens == o2.gens && o1.ord == o2.ord
+end
+
 mutable struct ModuleOrdering{S}
    M::S
    o::AbsModOrdering # must allow gen*mon or mon*gen product ordering
@@ -1739,6 +1743,29 @@ mutable struct ModProdOrdering <: AbsModOrdering
    a::AbsOrdering
    b::AbsOrdering
 end
+
+# Equality checking and hashing
+# Helper for checking equality by checking equality for each entry in a struct
+@generated function _fieldsequal(x, y)
+  if !isempty(fieldnames(x))
+    mapreduce(n -> :(_fieldsequal(x.$n, y.$n)), (a,b)->:($a && $b), fieldnames(x))
+  else
+    :(x == y)
+  end
+end
+
+@generated function _hashbyfields(x, h)
+  if !isempty(fieldnames(x))
+    mapreduce(n -> :(_hashbyfields(x.$n, h)), (a,b) -> :(hash(($a, $b), h)), fieldnames(x))
+  else
+    :(hash(x, h))
+  end
+end 
+
+Base.hash(o1::AbsModOrdering, h::UInt) = _hashbyfields(o1, h)
+Base.hash(o1::ModuleOrdering, h::UInt) = hash((hash(o1.o), hash(o1.M)), h)
+==(o1::AbsModOrdering, o2::AbsModOrdering) = _fieldsequal(o1, o2)
+==(o1::ModuleOrdering, o2::ModuleOrdering) = o1.M == o2.M && o1.o == o2.o
 
 Base.:*(a::AbsGenOrdering, b::AbsModOrdering) = ModProdOrdering(a, b)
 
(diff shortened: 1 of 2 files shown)


fieker:
Can't this be handled by matrices or similar like in the monomial Case?
Is there a canonical normal form for those orderings?

thofma:
Hm, yes, I wonder whether this is correct:
```
mutable struct ModOrdering{T} <: AbsModOrdering
   gens::T
   ord::Symbol
```
If one only checks if `gens` and `ord` are equal, then this does not check equality of orderings.

RafaelDavidMohr:
> Can't this be handled by matrices or similar like in the monomial Case? Is there a canonical normal form for those orderings?

It can but it is more involved, see theorem 2 [here](https://www.sciencedirect.com/science/article/pii/S0747717104000999). I decided against this because it would go substantially beyond the addressed issue to implement this and because, as far as I can tell, we don't support any functionality for such general matrix orderings.

If required, I am however happy to take a look at a canonical matrix form for the orderings we do support, maybe it's easier in this case.

RafaelDavidMohr:
> Hm, yes, I wonder whether this is correct:
> 
> ```
> mutable struct ModOrdering{T} <: AbsModOrdering
>    gens::T
>    ord::Symbol
> ```
> 
> If one only checks if `gens` and `ord` are equal, then this does not check equality of orderings.

I apologize I don't quite understand what you mean. Do you have an example?

thofma:
I was thinking about the following situation: The polynomial ring has one variable and the module is free of rank one. Then `:lex` and `:degrevlex` and whatnot define all the same orderings, but the equality check proposed here should yield `false`?

RafaelDavidMohr:
> I was thinking about the following situation: The polynomial ring has one variable and the module is free of rank one. Then `:lex` and `:degrevlex` and whatnot define all the same orderings, but the equality check proposed here should yield `false`?

Yes, good point. Ok, I will think of something else.

RafaelDavidMohr:
Brief update: I am working on this but this will take a while. We will need canonical representations of module orderings and they will take a while to implement.

RafaelDavidMohr:
I've added equality and hashing by embedding the underlying module of the ordering in question into a new polynomial ring and computing the canonical matrix there, which is then used for equality checking and hashing.

While this is not a unique representation of the ordering, it catches at least all reasonable cases that I could come up with.

src/Rings/orderings.jl
```diff
@@ -1717,33 +1717,89 @@ end
...
+      o.canonical_matrix = canonical_matrix(induced_ring_ordering(o))
+    else
+      o.canonical_matrix = _canonical_matrix_intern(o)
+    end
+  end
+  return o.canonical_matrix
+end
+
+function Base.:(==)(o1::ModuleOrdering, o2::ModuleOrdering)
+  return canonical_matrix(o1) == canonical_matrix(o2)
```
lgoettgens, line 1796: I think that == should compare the underlying modules of the orderings as well. And then same for the hash: the field M should be used there as well

src/Rings/orderings.jl
```diff
@@ -1717,33 +1717,91 @@ end
...
+end
+
+function Base.:(==)(o1::ModuleOrdering, o2::ModuleOrdering)
+  return o1.M == o2.M && canonical_matrix(o1) == canonical_matrix(o2)
+end
+
+function Base.hash(o::ModuleOrdering, h::UInt)
+  h1 = hash(o.M, h)
+  h2 = hash(canonical_matrix(o), h)
+  return hash(h2, h)
```
lgoettgens:
```suggestion
  h = hash(o.M, h)
  h = hash(canonical_matrix(o), h)
  return h
```
`h1` is not used anywhere. I think you wanted to do something like this

RafaelDavidMohr: that makes more sense, yes. Fixed.

src/Rings/orderings.jl
ederc: I'm not sure I completely understand why. If I have two different modules and I want to find out if both share the same ordering, how shall I do that? Note that also for `MonomialOrdering` we do not compare the corresponding rings.

lgoettgens: In that case, we should definitely keep it consistent and revert the change here. Sorry for the double work @RafaelDavidMohr

RafaelDavidMohr: No worries, it's reverted.

ederc (approved):

fingolfin (approved):

Issue #3408, opened 2024-02-21
Make sure that code in `src/` does not use code in `experimental/`


lgoettgens:
This is stated in a big red box in https://docs.oscar-system.org/dev/Experimental/intro/#Purpose but not enforced.

For stability guarantees of Oscar 1.x related to the `experimental` folder (see https://github.com/oscar-system/Oscar.jl/issues/2011), we should try to enforce the above rule, or at least create a list of issues and decide to not deal with them for now.

Potential roadmap:
- [x] #3409: Adapt loading of experimental packages to be more self-contained. This makes it then possible to delete a module in experimental without the need to adapt anything outside of that.
- [x] #3525: Add a CI job that deletes everything in the experimental folder (but a whitelist) and tries to run the Oscar tests. In the beginning, we would put most of the current experimental folders into the whitelist. A similar job should check if the documentation can still be build without experimental projects (aka experimental's docstrings are not contained in non-experimental doc pages).
- [ ] Long-term: Try to remove things from the whitelist. Adapt them or the calling code to keep everything working. Eventually, the whitelist should be empty.
  - [ ] DoubleAndHyperComplexes: `MethodError: no method matching simplify(::SubquoModule{MPolyQuoRingElem{MPolyDecRingElem{FqMPolyRingElem, AbstractAlgebra.Generic.MPoly{FqMPolyRingElem}}}})`
  - [ ] ExteriorAlgebra ``undefined binding 'exterior_algebra' in `@docs` block in src/NoncommutativeAlgebra/PBWAlgebras/quotients.md:40-42`` and `Error During Test at /home/runner/work/Oscar.jl/Oscar.jl/test/Rings/PBWAlgebraQuo.jl:41`
  - [ ] GaloisGrp: ``no docs found for 'fixed_field(C::Oscar.GaloisGrp.GaloisCtx, s::Vector{PermGroup})' in `@docs` block in src/NumberTheory/galois.md:275-278``
  - [ ] GModule: many doctest failures and `MethodError: no method matching numerator(::QQPolyRingElem)`
  - [ ] InvariantTheory: ``undefined binding 'linearly_reductive_group' in `@docs` block in src/InvariantTheory/reductive_groups.md:53-55` and more docs errors``
  - [ ] ModStd: `MethodError: no method matching monomial(::QQMPolyRing, ::Vector{Int64})` and many similar errors
  - [ ] Schemes

@thofma and @joschmitt liked the idea.

[...]

PR #3409, opened 2024-02-21
Make experimental packages more self-contained

lgoettgens:
Resolves the first bullet point in the list in https://github.com/oscar-system/Oscar.jl/issues/3408.

With this PR, removing an experimental project is now as easy as deleting the corresponding folder in `experimental/` (and maybe removing it from `oldexppkgs` or `orderedpkgs` in `experimental/Experimental.jl`), in particular `using Oscar` and `Aqua.test_undefined_exports(Oscar)` succeed. The exemption is `experimental/Rings/` which, currently, needs to be available for Oscar to load.
Tests in general, will not work yet, see https://github.com/oscar-system/Oscar.jl/issues/3408 for a roadmap on how to achieve that.

This PR does not change any current behaviour/exports/deprecations/etc., it just moves them to a more suited places.


benlorenz:
@lgoettgens I am not really sure why this should be in 1.0, it is mostly meant to help keep the code-bases separate but there will not be much further development on the release branch. Moving lots of stuff around now seems not helpful for the release.

lgoettgens:
Yes, you are right. I think my intention was to reduce the number of potential conflicts occurring while backporting

benlorenz:
Maybe it would make sense to wait with merging this PR until we are done with most of the backporting, to avoid any conflicts, both for master and for the backports.

lgoettgens:
Yeah, waiting a few more days until all of the feature backports are done is more than fine. The minor doc changes later shouldn't become an issue with conflicts

fingolfin (approved):
Looks good to me

benlorenz:
I changed it to draft so that we don't accidentally merge this right now, we can change it back some time next week I guess.

fieker:
What is the motivatio to further tinker with the experimental stuff? As far as I can see: it works.
Is there any real, known, problem with the current situation?

lgoettgens:
> What is the motivatio to further tinker with the experimental stuff? As far as I can see: it works. Is there any real, known, problem with the current situation?

Please see https://github.com/oscar-system/Oscar.jl/issues/3408 for the motivation. And let's keep the fundamental discussion there.

lkastner (approved):
Thanks for taking care of this.

lgoettgens:
> Maybe it would make sense to wait with merging this PR until we are done with most of the backporting, to avoid any conflicts, both for master and for the backports.

Since 1.0 is now released and the bulk of backports is done, I think we could continue with this.

benlorenz (approved):

lgoettgens:
Can we get this merged before any more conflicts come up?

Issue #3304, opened 2024-02-02
Tropical groebner bases - minor overhaul


YueRen:
(This is related to discussions with @oskarhenriksson and a bug communicated by @VictoriaSchleis)

Currently, tropical Groebner bases (and consequently tropical initial ideals) can only be computed if the ideal is homogeneous, but that is unnecessary restrictive.  There are many special cases for which they can be computed even if the ideal is inhomogeneous:

- ideal is principal
- ideal is linear, i.e., generated by polynomials with monomials of degree 1 OR 0
- ~~ideal is binomial~~ (this is harder than anticipated, in case the weight vector does not belong in the tropicalization)
- weight vector is strictly negative (min-convention) or strictly positive (max-convention), and the valuation is trivial

Instead of raising an error, the tropical Groebner basis function should recognize the special cases and use a specialised algorithm wherever appropriate.

YueRen:
Closed with pull request #3411

PR #3411, opened 2024-02-21
Yr/tropical bugs

YueRen:
Just minor bugfixes, no breaking changes.  

The main difference in feature is that `groebner_basis` (the tropical version, called by specifying a `TropicalSemiringMap`) now checks whether the generators are principal or (affine) linear and, if yes, returns the Groebner basis.  The previous version merely returned an error.

Draft only for know, missing some tests to be added tomorrow.


YueRen:
This pull request should also resolve https://github.com/oscar-system/Oscar.jl/issues/3304

YueRen:
@lkastner Would you mind reviewing this PR?  Sorry, I know tropical Groebner bases aren't in your area, but it really isn't in any OSCAR maintainer's area except myself, and even I wouldn't count myself an expert since I don't work with them regularly.

lkastner (approved):

fingolfin:
Should this be backported?

PR #3412, opened 2024-02-22
Corrected defn of deg(ideal); correction from W Decker

JohnAAbbott:
Two corrections:
* return type is now `ZZRingElem` (previously `Int`)
* defn is now correct
* no longer triggers SEGV when given zero ideal


wdecker (approved):

thofma:
Some tests for all these three issues would have been nice.

wdecker:
I will add a relevant test example tomorrow.

PR #3421, opened 2024-02-23
list the needed GAP packages

ThomasBreuer:
We want to have the needed GAP packages in the list in `src/Oscar,jl`.

- `polycyclic` gets loaded automatically by GAP, but we want to get an error if it is for some reason not loadable.
- `packagemanager` gets loaded by GAP.jl.
- `ferret` gets loaded in the Oscar module `GaloisGrp`, in fact there we even try to install this package if it is not yet there. However, it is not an error if the package cannot be loaded.

addresses #3395


fingolfin (approved):

ThomasBreuer:
`Run tests / test (1.11-nightly, long, ubuntu-latest)` fails.
I do not think that this is caused by the proposed changes.

PR #3424, opened 2024-02-23
[FTheoryTools] Rewrite _kodaira_type to work with any locus

apturner:
Rewrite the function `_kodaira_type` so that it should now work with any locus, rather than just loci of the form `w = 0` for `w` a coordinate.

### DIFF
--- a/experimental/FTheoryTools/src/auxiliary.jl
+++ b/experimental/FTheoryTools/src/auxiliary.jl
@@ -176,24 +176,45 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
     poly_g = ring_map(g)
     poly_d = ring_map(d)
     locus = ring_map(gens(id)[1])
+    locus_id = ideal([locus])
     
     if f_ord == 0 && g_ord == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(9 * poly_g, [locus], [0]), evaluate(2 * poly_f, [locus], [0]))
+      quotient_gens = gens(groebner_basis(quotient(ideal([9 * poly_g]), ideal([2 * poly_f])) + locus_id))
+      deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      quotient_val = quotient_gens[1]
+
+      monodromy_poly = _psi^2 + quotient_val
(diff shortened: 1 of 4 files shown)


experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,45 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
     poly_g = ring_map(g)
     poly_d = ring_map(d)
     locus = ring_map(gens(id)[1])
+    locus_id = ideal([locus])
     
     if f_ord == 0 && g_ord == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(9 * poly_g, [locus], [0]), evaluate(2 * poly_f, [locus], [0]))
+      quotient_gens = gens(groebner_basis(quotient(ideal([9 * poly_g]), ideal([2 * poly_f])) + locus_id))
+      deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      quotient_val = quotient_gens[1]
```
HereAround: I am wondering if the check `g == locus` will always work. Generators of the quotient are not unique. Will the Groebner basis computation completely eliminate this redundancy? Also, could additional signs/numeric factors appear that would force `g == locus` to return  `false`?

Less critical, but relevant: Unless I am mistaken this pattern of lines appears four times in this PR. Does it make sense to add an auxiliary function and call it instead to make the code shorter/cleaner?

HereAround (approved):
Thank you @apturner . Generally speaking, looks good. I like the new tests.

I left a few nitpicks above, in case you want to address that. Otherwise, this can be merged from my point of view.

HereAround:
Thank you @apturner!

HereAround:
Sorry @apturner - I just wanted to rebase this PR to the latest master (was very outdated - apparently a lot of last minute changes for the OSCAR 1.0 release over the last two weeks), so that I could convince myself that this PR works with all those changes. Usually, the rebase is pretty easy (for me, even more so if it is my own PR), but in this particular case something malfunctioned... Sorry about that.

In any case: I just opened a new PR with your changes here: https://github.com/oscar-system/Oscar.jl/pull/3506.

apturner:
I rolled back the two additional changes mentioned in #3506, as they are causing tests to fail for an unrelated reason. I will add these changes to the TODO.

PR #3426, opened 2024-02-26
TropicalGeometry: add and improve polymake conversions

benlorenz:
cc: @lenweis @YueRen


YueRen (approved):
Thanks, this will make my life much easier!  Oh, I didn't know `[td-sum(alpha); alpha] for alpha in exponents(f))` works as it does.

PR #3432, opened 2024-02-26
add docstrings for `acting_group` and `action_function`

### LINKED ISSUE #3370: G-sets: define and document an interface for getting group,domain,action
$G$-sets in OSCAR represent group actions. That is, given an action $f:\Omega\times G\to \Omega$, they capture the acting group $G$, the domain $\Omega$ and the action "function" $f$.

I think there should thus be a documented way to get to these components that works for all $G$-sets. It should be explained on <https://docs.oscar-system.org/dev/Groups/action/#G-Sets>.

Current status:

- `acting_group`
  - [x] exists
  - [x] exported
  - [ ] docstring
- `action_function`
  - [x] exists
  - [ ] exported   
  - [ ] docstring
- `domain` (or `action_domain` or ...?)
  - [ ] exists
  - [ ] exported
  - [ ] docstring

(We can of course also change some or all of these names).

BTW @ThomasBreuer recently added `action_range` on my suggestion (not exported) but that's something else (and perhaps should be renamed `_action_range` as it is somewhat special and internal?)

@ThomasBreuer @mjrodgers thoughts? can you please coordinate and look into it (it is of course also fine if one you handles it alone, just coordinate it)

mjrodgers:
Closes #3370

### DIFF
--- a/src/Groups/gsets.jl
+++ b/src/Groups/gsets.jl
@@ -76,7 +76,7 @@ Return the group `G` acting on `Omega`.
 """
 acting_group(Omega::GSetByElements) = Omega.group
 
-"""
+@doc raw"""
     action_function(Omega::GSetByElements)
 
 Return the function ``f: \Omega \times G \to \Omega`` that defines the G-set.
--- a/src/Groups/gsets.jl
+++ b/src/Groups/gsets.jl
@@ -73,6 +73,14 @@ end
     acting_group(Omega::GSetByElements)
 
 Return the group `G` acting on `Omega`.
+
+# Examples
+```jldoctest
+julia> G = symmetric_group(4);
+
+julia> acting_group(gset(G, [1])) == G
+true
+```
 """
 acting_group(Omega::GSetByElements) = Omega.group
 
@@ -80,6 +88,20 @@ acting_group(Omega::GSetByElements) = Omega.group
     action_function(Omega::GSetByElements)
 
 Return the function $f: \Omega \times G \to \Omega$ that defines the G-set.
+
+# Examples
+```jldoctest
+julia> G = symmetric_group(4);
+
+julia> action_function(gset(G, [1])) == ^
+true
+
+julia> action_function(gset(G, [[1, 2]])) == on_tuples
+true
+
+julia> action_function(gset(G, on_sets, [[1, 2]])) == on_sets
+true
+```
 """
 action_function(Omega::GSetByElements) = Omega.action_function
 
--- a/src/Groups/gsets.jl
+++ b/src/Groups/gsets.jl
@@ -70,7 +70,18 @@ function Base.show(io::IO, x::GSetByElements)
 end
 
 # TODO: document `acting_group`, `action_function`
+"""
+    acting_group(Omega::GSetByElements)
+
+Return the group `G` acting on `Omega`.
+"""
 acting_group(Omega::GSetByElements) = Omega.group
+
+"""
+    action_function(Omega::GSetByElements)
+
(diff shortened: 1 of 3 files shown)


src/Groups/gsets.jl
```diff
@@ -70,7 +70,18 @@ function Base.show(io::IO, x::GSetByElements)
 end
 
 # TODO: document `acting_group`, `action_function`
```
lgoettgens:
```suggestion
```

src/Groups/gsets.jl
```diff
@@ -69,8 +69,18 @@ function Base.show(io::IO, x::GSetByElements)
   end
 end
 
-# TODO: document `acting_group`, `action_function`
+"""
+    acting_group(Omega::GSetByElements)
+
+Return the group `G` acting on `Omega`.
```
ThomasBreuer, line 75: Perhaps add a doctest:
```
# Examples
```jldoctest
julia> G = symmetric_group(4);

julia> acting_group(gset(G, [1])) == G
true
```

src/Groups/gsets.jl
```diff
@@ -69,8 +69,18 @@ function Base.show(io::IO, x::GSetByElements)
...
+"""
 acting_group(Omega::GSetByElements) = Omega.group
+
+@doc raw"""
+    action_function(Omega::GSetByElements)
+
+Return the function $f: \Omega \times G \to \Omega$ that defines the G-set.
```
ThomasBreuer, line 90: Perhaps add a doctest:
````
# Examples
```jldoctest
julia> G = symmetric_group(4);

julia> action_function(gset(G, [1])) == ^
true

julia> action_function(gset(G, [[1, 2]])) == on_tuples
true

julia> action_function(gset(G, on_sets, [[1, 2]])) == on_sets
true
```
````

**Review by ThomasBreuer — CHANGES_REQUESTED** [review]
Thanks. Just a few useful additions:
- The documentation for the two functions should be mentioned in `docs/src/Groups/action.md`.
- `action_function` should be added to `src/exports.jl`.
- Some doctests could be added.

**Comment by fingolfin** [comment]
Thanks that's a good start, but we should also verify that those functions are actually exported - I don't think they are (at least not both).

Perhaps you could add a simple `jldoctest` to both docstrings showing these functions, I think this then also automatically checks that they are exported.

**Comment by fingolfin** [comment]
It seems @ThomasBreuer and me had the same thoughts at about the same moment, so our comments crossed :-)

File: src/Groups/gsets.jl (hunk 3 above)
Comment (fingolfin) [reply src/Groups/gsets.jl:90]: BTW to add a code block which contains triple backticks... you can simply use quadruple backticks (and so on). E.g.
`````
````
This is triple-quoted doctest code block:
```jldoctest
julia> 1+1
2
```
````
`````
results in this:
````
This is triple-quoted doctest code block:
```jldoctest
julia> 1+1
2
```
````

fingolfin (approved):

lgoettgens:
> The documentation for the two functions should be mentioned in docs/src/Groups/action.md.

This request of @ThomasBreuer is still open

mjrodgers:
> This request of @ThomasBreuer is still open

Thanks, I overlooked this one - I will update now

ThomasBreuer (approved):
Thanks, looks good.

mjrodgers:
Okay, this should have everything addressed now. I also mentioned using `collect(Omega)` to obtain the underlying G-set in `action.md`, so this should suffice to address everything now from #3370

lgoettgens:
Backport? Most of this is just documentation enhancements

benlorenz:
The new exports are used in the groups chapter, adding it to the backports list.

PR #3404, opened 2024-02-21
Add `all_primitive_groups` and `all_transitive_groups` variants taking a single int or int range

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/3403.

PR #3433, opened 2024-02-26
Align `all_*_groups` methods some more

fingolfin:
For `all_small_groups` we allow specifying the order as a single
integer or abstract vector, followed by any number of additional
filters. For transitive and primitive groups, we also allowed
specifying the degree as a single integer or abstract vector, but
this then could not be followed by filters. With this patch the
additional filters work in all cases.

Depends on PR #3404 (which has already been merged), so this PR here can only be backported after PR #3404.

### DIFF
--- a/src/Groups/libraries/primitivegroups.jl
+++ b/src/Groups/libraries/primitivegroups.jl
@@ -163,6 +163,10 @@ may be of one of the following forms:
 - `func` selects groups for which the function `func` returns `true`
 - `!func` selects groups for which the function `func` returns `false`
 
+As a special case, the first argument may also be one of the following:
+- `intval` selects groups whose degree equals `intval`; this is equivalent to `degree => intval`
+- `intlist` selects groups whose degree is in `intlist`; this is equivalent to `degree => intlist`
+
 The following functions are currently supported as values for `func`:
 - `degree`
 - `is_abelian`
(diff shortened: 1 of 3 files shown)


lgoettgens:
primitive and transitive groups were both changed similarly in https://github.com/oscar-system/Oscar.jl/pull/3404, so it would make sense to add a similar change to primitive groups here as well.

Regarding the backport: https://github.com/oscar-system/Oscar.jl/pull/3404 is not yet backported. So backporting this here will create ugly conflicts.

lgoettgens (approved):

fingolfin:
@lgoettgens thank you! I stumbled over this for transitive groups today while working on the book, and instead of checking, blindly assumed that primitive groups followed the small groups... Argh.

Anyway, fixed now, also adjusted the commit message / PR description.


As for the backporting: yes, this should only be backported if and after PR #3404 is backported.

lgoettgens:
When I worked on https://github.com/oscar-system/Oscar.jl/pull/3404, I should have looked at all the `all_*_groups` implementations (including the small groups) to make it completely consistent. I cannot remember how I overlooked them.

fingolfin:
@lgoettgens well, when I reviewed PR #3404 I should have thought "wait a moment, ....", looked at those other methods, and told you about them. :-) but I didn't, I thought "this is fine". If there is any "blame" (if one wants to use that word, I don't really) it is on me, not you :-)

PR #3434, opened 2024-02-26
Add `all_perfect_groups`

### LINKED ISSUE #3376: Add `all_perfect_groups`
We have small groups, primitive groups, transitive groups, but somehow we forgot the perfect groups?

fingolfin:
Resolves #3376

### DIFF
--- a/src/Groups/libraries/perfectgroups.jl
+++ b/src/Groups/libraries/perfectgroups.jl
@@ -178,7 +178,75 @@ function number_of_perfect_groups(n::IntegerUnion)
    return res::Int
 end
 
-# TODO: add all_perfect_groups() iterator
+
+"""
+    all_perfect_groups(L...)
+
+Return the list of all perfect groups (up to permutation isomorphism)
+satisfying the conditions described by the arguments. These conditions
+may be of one of the following forms:
+
+- `func => intval` selects groups for which the function `func` returns `intval`
+- `func => list` selects groups for which the function `func` returns any element inside `list`
+- `func` selects groups for which the function `func` returns `true`
+- `!func` selects groups for which the function `func` returns `false`
+
+As a special case, the first argument may also be one of the following:
+- `intval` selects groups whose order equals `intval`; this is equivalent to `order => intval`
+- `intlist` selects groups whose order is in `intlist`; this is equivalent to `order => intlist`
+
+The following functions are currently supported as values for `func`:
+- `is_quasisimple`
+- `is_simple`
+- `is_sporadic_simple`
+- `number_of_conjugacy_classes`
+- `order`
+
+The type of the returned groups is `PermGroup`.
+
+# Examples
+```jldoctest
+julia> all_perfect_groups(7200)
+2-element Vector{PermGroup}:
+ Permutation group of degree 29 and order 7200
+ Permutation group of degree 288 and order 7200
+
+julia> all_perfect_groups(order => 1:200, !is_simple)
+2-element Vector{PermGroup}:
+ Permutation group of degree 1 and order 1
+ Permutation group of degree 24 and order 120
+```
+"""
+function all_perfect_groups(L...)
+   @req !isempty(L) "must specify at least one filter"
+   if L[1] isa IntegerUnion || L[1] isa AbstractVector{<:IntegerUnion}
+      L = (order => L[1], L[2:end]...)
+   end
+   # first get all order restrictions
+   ordsL = [x for x in L if x isa Pair && x[1] == order]
+   @req !isempty(ordsL) "must restrict the order"
+   conds = [x for x in L if !(x isa Pair && x[1] == order)]
+   orders = intersect([x[2] for x in ordsL]...)
+   @req has_perfect_groups(maximum(orders)) "only orders up to 2 million are supported"
+   res = PermGroup[]
+   for n in orders, i in 1:number_of_perfect_groups(n)
+      G = perfect_group(n, i)
+      ok = true
+      for c in conds
+         if c isa Pair
+            val = c[1](G)
+            ok = (val == c[2] || val in c[2])
+         elseif c isa Function
+            ok = c(G)
+         else
+            throw(ArgumentError("expected a function or a pair, got $arg"))
+         end
+         ok || break
+      end
+      ok && push!(res, G)
+   end
+   return res
+end
 
 function __init_extraperfect()
   for i in [27, 33]
--- a/test/Groups/libraries.jl
+++ b/test/Groups/libraries.jl
@@ -118,7 +118,7 @@ end
    Gs = all_perfect_groups(order => 1:200)
    @test length(Gs) == sum(number_of_perfect_groups, 1:200)
    @test Gs == all_perfect_groups(1:200)
-   @test length(all_perfect_groups(7200)) = number_of_perfect_groups(7200)
+   @test length(all_perfect_groups(7200)) == number_of_perfect_groups(7200)
 
    # all_perfect_groups with additional attributse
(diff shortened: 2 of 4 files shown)


src/Groups/libraries/perfectgroups.jl
```diff
@@ -178,7 +178,78 @@ function number_of_perfect_groups(n::IntegerUnion)
...
+- `func => list` selects groups for which the function `func` returns any element inside `list`
+- `func` selects groups for which the function `func` returns `true`
+- `!func` selects groups for which the function `func` returns `false`
+
+As a special case, the first argument may also be one of the following:
+- `intval` selects groups whose order equals `intval`; this is equivalent to `order => intval`
+- `intlist` selects groups whose order is in `intlist`; this is equivalent to `order => intlist`
+
+The following functions are currently supported as values for `func`:
+- `is_almost_simple`
```
ThomasBreuer: This is not really useful (as long as `is_simple` is supported).

src/Groups/libraries/perfectgroups.jl
```diff
@@ -178,7 +178,78 @@ function number_of_perfect_groups(n::IntegerUnion)
...
+- `is_quasisimple`
+- `is_simple`
+- `is_sporadic_simple`
+- `is_supersolvable`
```
ThomasBreuer: This is not really useful.

ThomasBreuer (approved):
Thanks, good new feature.
(Perhaps remove two of the supported functions, see the comments.)

src/Groups/libraries/perfectgroups.jl
```diff
@@ -178,7 +178,78 @@ function number_of_perfect_groups(n::IntegerUnion)
...
+2-element Vector{PermGroup}:
+ Permutation group of degree 29 and order 7200
+ Permutation group of degree 288 and order 7200
+
+julia> all_perfect_groups(order => 1:200, !is_simple)
+2-element Vector{PermGroup}:
+ Permutation group of degree 1 and order 1
+ Permutation group of degree 24 and order 120
+```
+returns the list of all abelian transitive groups acting on 3, 4 or 5 points.
```
lgoettgens:
```suggestion
returns the list of all non-simple perfect groups of order at most 200.
```

test/Groups/libraries.jl
```diff
@@ -115,6 +115,15 @@ end
    @test_throws ArgumentError number_of_perfect_groups(0) # invalid argument
    @test_throws ArgumentError number_of_perfect_groups(ZZRingElem(60)^10)  # result not known
 
+   Gs = all_perfect_groups(order => 1:200)
+   @test length(Gs) == sum(number_of_perfect_groups, 1:200)
+   @test Gs == all_perfect_groups(1:200)
+   @test filter(G -> number_conjugacy_classes(G) in 5:8, Gs) == all_perfect_groups(1:200, number_conjugacy_classes => 5:8)
```
lgoettgens:
```suggestion
   @test filter(G -> number_of_conjugacy_classes(G) in 5:8, Gs) == all_perfect_groups(1:200, number_of_conjugacy_classes => 5:8)
```

src/Groups/libraries/perfectgroups.jl
```diff
@@ -178,7 +178,78 @@ function number_of_perfect_groups(n::IntegerUnion)
...
+- `func => intval` selects groups for which the function `func` returns `intval`
+- `func => list` selects groups for which the function `func` returns any element inside `list`
+- `func` selects groups for which the function `func` returns `true`
+- `!func` selects groups for which the function `func` returns `false`
+
+As a special case, the first argument may also be one of the following:
+- `intval` selects groups whose order equals `intval`; this is equivalent to `order => intval`
+- `intlist` selects groups whose order is in `intlist`; this is equivalent to `order => intlist`
+
+The following functions are currently supported as values for `func`:
```
fingolfin, line 198: Note that I do not actually restrict which functions are used in the code; my reading of "supported" here is "we promise these will work, anything else can lead to errors. I could explicitly reject other functions, but there seems no point in this right now.

src/Groups/libraries/perfectgroups.jl
```diff
@@ -178,7 +178,78 @@ function number_of_perfect_groups(n::IntegerUnion)
...
+function all_perfect_groups(L...)
+   @req !isempty(L) "must specify at least one filter"
+   if L[1] isa IntegerUnion || L[1] isa AbstractVector{<:IntegerUnion}
+      L = (order => L[1], L[2:end]...)
+   end
+   # first get all order restrictions
+   ordsL = [x for x in L if x isa Pair && x[1] == order]
+   @req !isempty(ordsL) "must restrict the order"
+   conds = [x for x in L if !(x isa Pair && x[1] == order)]
+   orders = intersect([x[2] for x in ordsL]...)
```
lgoettgens, line 229: out of simplicity, I would throw an error if multiple `order` restrictions are there. This right now could work from the first glance, if you want to keep it, please add a test with multiple `order` restrictions, e.g.
```julia
	@test all_perfect_groups(order => 1:5:200, order => 25:50) == all_perfect_groups(order => intersect(1:5:200, 25:50))
```

src/Groups/libraries/perfectgroups.jl
fingolfin: Thanks, another copy&paste mistake -- but it also indicates the source has an issue, two, will fix that as well in my other PR

src/Groups/libraries/perfectgroups.jl
ThomasBreuer: No problem with the code; if someone asks for supersolvable perfect groups then this is fine.
But it is irritating to read `is_supersolvable` here.

lgoettgens: This was exactly what I expected from the wording.
Is there a reason why this list is so much shorter than e.g. for the transitive groups?
https://github.com/oscar-system/Oscar.jl/blob/83f05750497b654dfabafc210919f88dfbea534c/src/Groups/libraries/transitivegroups.jl#L162-L179

I suppose this is too computationally expensive?

fingolfin: No, just most of these filters make no sense: the only perfect group which is abelian/cyclic/nilpotent/solvable/super-solvable is the trivial group.

And primitive/transitive/degree etc. don't really make sense here -- while we return permutation groups, (a) in a future version we may give the user a choice to also get fp groups, (b) which permutation group representations we use is arbitrary. I.e.: the "degree" of a transitive group is well-defined; the "degree" of a perfect group is not.

lgoettgens: Ah okay, thanks for clarifying this.

lgoettgens (approved):

fingolfin:
@ThomasBreuer @lgoettgens thank you both for your great feedback. I hope I've addressed it all for now, please let me know if there is anything else.

lgoettgens:
No more objections from me (assuming CI succeeds)

test/Groups/libraries.jl
```diff
@@ -115,6 +115,21 @@ end
    @test_throws ArgumentError number_of_perfect_groups(0) # invalid argument
    @test_throws ArgumentError number_of_perfect_groups(ZZRingElem(60)^10)  # result not known
 
+   Gs = all_perfect_groups(order => 1:200)
+   @test length(Gs) == sum(number_of_perfect_groups, 1:200)
+   @test Gs == all_perfect_groups(1:200)
+   @test length(all_perfect_groups(7200)) = number_of_perfect_groups(7200)
```
benlorenz:
```suggestion
   @test length(all_perfect_groups(7200)) == number_of_perfect_groups(7200)
```

PR #3440, opened 2024-02-27
TropicalGeometry: Fix dual subdivision

lkastner:


YueRen:
Mind if I add some minor changes on top of yours?  I agree that returning the polymake object was a bad idea.  

In it's current form, - if the tropical hypersurface was constructed using a polynomials - users will get back the subdivision on the homogenized exponent vectors.  I'd like it to be the original exponent vectors.

lkastner:
> Mind if I add some minor changes on top of yours? I agree that returning the polymake object was a bad idea.
> 
Sure, go ahead.

src/TropicalGeometry/hypersurface.jl
```diff
@@ -210,15 +210,41 @@ end
 @doc raw"""
     dual_subdivision(TropH::TropicalHypersurface)
 
-Return the dual subdivision of `TropH`.  Raises an error, if neither it nor the the internal polymake object is cached.
+Return the dual subdivision of `TropH`.  Raises an error, if neither it nor the
+the internal polymake object is cached.
```
fingolfin: Nitpick from an outsider (so this is by no means a blocker, double so since this is just reformatting existing text): I find the second sentence rather technical and I am not sure what it even is telling me. My uninformed guess would be: this function can not actually compute a dual subdivision; it can only return one if it is already set somehow. If that's what it means, then perhaps it could be stated as such? If it means something else, well, perhaps that could be explained better?

src/TropicalGeometry/hypersurface.jl
```diff
@@ -210,15 +210,41 @@ end
...
+```
 """
-function dual_subdivision(TropH::TropicalHypersurface{minOrMax,true}) where minOrMax
-    if has_attribute(TropH,:dual_subdivision)
-        return get_attribute(TropH,:dual_subdivision)
-    elseif has_attribute(TropH,:polymake_bigobject)
-        return get_attribute(TropH,:polymake_bigobject)
+@attr function dual_subdivision(TropH::TropicalHypersurface{minOrMax,true}) where minOrMax
+    if has_attribute(TropH,:polymake_bigobject)
+        result = get_attribute(TropH,:polymake_bigobject)
```
fingolfin: Just to say, you could also do this (and avoid scanning the attributes dict twice)
```suggestion
    result = get_attribute(TropH, :polymake_bigobject)
    if result !== nothing
```
But of course the code is also fine as it is

lkastner: Thank you.

src/TropicalGeometry/hypersurface.jl
```diff
@@ -210,15 +211,11 @@ end
 @doc raw"""
     dual_subdivision(TropH::TropicalHypersurface)
 
-Return the dual subdivision of `TropH`.  Raises an error, if neither it nor the the internal polymake object is cached.
+Return the dual subdivision used to construct `TropH`.  Raises an error, if it is not cached.
```
fingolfin:
```suggestion
Return the dual subdivision used to construct `TropH`. Raises an error if it is not cached.
```
Ahh, thank you, that makes it clear!

YueRen:
Just a quick list of changes that I have pushed:
- created a new function `newton_subdivision`, since it is good to have (input is a tropical polynomial or a non-tropical polynomial + tropical semiring map). This may be of interest to @FiroozehDastur 
- fixed the `TropicalHypersurface` constructor that takes a subdivision as input
- made `dual_subdivision(::TropicalHypersurface)` return an error if the dual subdivision is not cached (and changed the documentation accordingly).  this is in line with all other attributes of all other tropical objects

It would be more intuitive and user-friendly if - instead of returning an error - `dual_subdivision` constructs the subdivision from other cached data if possible.  The on-demand computation of desired attributes from existing attributes is a more general issue that I may implement if there is some demand.  I'm hesitant to do so, because - unless there is an elegant way of implementing it (like rules in polymake) - it'll boil down to a wall of if-statements.

YueRen (approved):

lkastner:
Ok, I eliminated the double caching of the `dual_subdivision` and added a jldoctest so we can avoid repeating this exercise. ;)

In the long run it would be nice to have some more mathematical explanation describing whether a dual subdivision can be returned instead of "it is cached or not". ;)

YueRen (approved):
Looks good, thanks for the docstring example!

PR #3442, opened 2024-02-27
Rename `InvRing` + small changes to invariant theory

joschmitt:
Rename:
* `InvRing` to `FinGroupInvarRing`
* `RedGrpInvRing` to `RedGroupInvarRing`
* `TorGrpInvRing` to `TorGroupInvarRing`

The renamings do not interfere with the book and are no breaking changes since they are either internal or in experimental.

I made further small internal changes to the experimental invariant theory (which do not interfere with the book).
This contains the changes from #3469, I will rebase once that is merged.


thofma:
Just mentioning here, that `Group` in most type names is often not abbreviated to `Grp`: `PermGroup`, `PcGroup`, `MatrixGroup`, `AbGroup`, but it seems invariant theory settled on `Grp`.

joschmitt:
I make this a draft as it creates conflicts with #3443. Also, we should settle the `Grp` vs `Group` question first. (And maybe `Inv` vs `Invar` while we are at it.)

wdecker:
@joschmitt Sorry, I have seen your PR only now.

joschmitt:
> @joschmitt Sorry, I have seen your PR only now.

Don't worry. This is only internal changes and can wait.

fingolfin:
I am neutral on the `Grp` vs `Group` thing here, and if @thofma prefers to have `Group`, I am happy with that, too.

Regarding `Inv` vs `Invar` I'd prefer `Invar`, as `InvRing` could also stand for "involutory ring". But I am happy either way, for me the most important part is that we achieve "local" consistency, that is: the types for invariant rings for finite groups and for reductive groups are getting aligned.

joschmitt:
Okay, I will change everything to `*GroupInvarRing` once #3443 is through (and I checked that it does not interfere with the book...)

joschmitt:
This is good to go from my side, but waiting for #3469.

fingolfin (approved):
#3469 is in, so now a conflict needs to be resolved. Otherwise seems fine to me, thanks

joschmitt:
This is good to go now (once CI approves).

Issue #2011, opened 2023-03-03
Make the user aware of that they are using code from `experimental`


lkastner:
I decided to merge #1999 and relocate the discussion on the separation of `experimental` from the rest of the source here.
- We all seemed to agree that the user should in some way be made aware of that they are using code from `experimental`. Just for the sake of completeness: Please disagree here if you disagree.
- Should code from experimental be available by default? As already stated, if not its discoverability goes down, but that is not necessarily a bad thing. In my pov code should not stay "experimental" for too long.

These were the ideas of @fingolfin :
> Indeed. Some options come to mind:
> 
>     * maybe accessing code in experimental should be "guarded" and require `using Oscar.Experimental`, or require writing `Experimantal.FOO` (so `Experimental.galois_group(...)`
>       
>       * pro: obviously experimental
>       * con: less discoverable
> 
>     * or maybe the primary functions should have a `@warn_experimental()` call in them, which prints a warning about them being experimental (possibly only the first time they are called)
> 
>     * ... other ideas?

So I'll cc some people who might be interested: @micjoswig @fieker @benlorenz . Maybe we can discuss a bit here and then decide during one of the next Friday meetings, preferably when we are all there, unless we immediately find a consensus here.

fieker:
On Fri, Mar 03, 2023 at 08:36:09AM -0800, Lars Kastner wrote:
> I decided to merge #1999 and relocate the discussion on the separation of `experimental` from the rest of the source here.
> - We all seemed to agree that the user should in some way be made aware of that they are using code from `experimental`. Just for the sake of completeness: Please disagree here if you disagree.
I do not agree here - but do not strongly disagree
> - Should code from experimental be available by default? As already stated, if not its discoverability goes down, but that is not necessarily a bad thing. In my pov code should not stay "experimental" for too long.
> 
yep

Question:
  What do we want to achieve/ avoid?

  - lower the barrier to share code
  - make it available immediately
  - do not promise no further (interface) changes
  - ???
  - s.o. getting wrong results (yes, we want to avoid that!!)

So if I write a function CoolStuff in experimental (OK, I will write
cool_stuff), and expoert it, then
 - anyone can play with it
 - there is a slight danger that the code will immediately be used
   programmatically which might break later
   This would also indicate a more experienced user who might be able
   to discover the experimental nature alone
 - there is a slightly higher chance that s.o. will just interactively
   play with it - in which the discoverability is important and the
   breaking of th API irrelevant
[...]

PR #3443, opened 2024-02-27
Docu invariants lin. red. groups (Lakshmi Ramesh and Wolfram Decker)

wdecker:
…d Wolfram Decker)


wdecker:
What is the status quo here?
1) This documents functionality used in the book.
2) There is additional functionality to be documented later.
3) There is additional functionality under development.

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -54,8 +54,46 @@ function Base.show(io::IO, G::LinearlyReductiveGroup)
 end
 
 #getter functions
-linearly_reductive_group(sym::Symbol, m::Int, R::MPolyRing) = LinearlyReductiveGroup(sym,m,R)
+
+@doc raw"""
+    linearly_reductive_group(sym::Symbol, m::Int, K::Field)
+
+If `sym` stands for implementing the special linear group, return SL$(m, K)$.
```
joschmitt: This sentence is a bit hard to understand, it is not clear what "stands for implementing" means. Maybe
```
Return the linearly reductive group defined by `sym` over the field $K$ of degree $m$.

Currently, the supported options for `sym` are:
* `:SL` corresponding to the special linear group.
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -54,8 +54,46 @@ function Base.show(io::IO, G::LinearlyReductiveGroup)
...
+Ideal generated by
+  z[1, 1]*z[2, 2] - z[2, 1]*z[1, 2] - 1
+```
+"""
 linearly_reductive_group(sym::Symbol, m::Int, F::Field) =  LinearlyReductiveGroup(sym,m,F)
+
+@doc raw"""
+    linearly_reductive_group(sym::Symbol, m::Int, R::MPolyRing)
+
+If `sym` stands for implementing the special linear group, return SL$(m, K)$, where $K$ is the base field
```
joschmitt: Similar to above.

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -489,9 +604,51 @@ function reynolds_operator(X::RepresentationLinearlyReductiveGroup, elem::MPolyR
...
+
+Return the invariant ring `RG` as an affine algebra (this amounts to compute the algebra syzygies among the fundamental invariants of `RG`).
+
+In addition, if `A` is this algebra, and `R` is the polynomial ring of which `RG` is a subalgebra,
+return the inclusion homomorphism  `A` $\hookrightarrow$ `R` whose image is `RG`.
+"""
+function affine_algebra(R::RedGrpInvRing)
+   V = fundamental_invariants(R)
+   s = length(V)
+   S,t = polynomial_ring(field(group(representation(R))), "t"=>1:s)
```
joschmitt: Assuming the fundamental invariants are homogeneous, we could/should grade this polynomial ring appropriately.
Also, the presentation as affine algebra should be cached in the invariant ring.

This can be filed under "future improvements".

docs/src/InvariantTheory/reductive_groups.md
```diff
@@ -4,18 +4,23 @@ CurrentModule = Oscar
 
 # Invariants of Linearly Reductive Groups
```
joschmitt, line 5: I think we should make it clear somewhere that this part of the documentation is about experimental functionality. Either there could be some "warning" here or the files could be moved to the experimental directory (and we just link at them here).

In the end, this is all part of the discussion in #2011.

docs/src/InvariantTheory/reductive_groups.md
```diff
@@ -24,32 +29,57 @@ See also [DK15](@cite) and [DJ98](@cite).
 
 ## Creating Invariant Rings
 
-There are no exact means to handle algebraically closed fields on the computer. For the computation of invariant rings in the above setting, on the other hand, there is no need to deal with explicit elements of ``G`` or with its group structure. The implementation of Derksen's algorithm in OSCAR can  handle situations where both $G$ and the representation $\rho$ are defined over an exact subfield $k$ of $K$ which is supported by OSCAR: 
+### How Linearly Reductive Groups and Their Representations are Given
+
+There are no exact means to handle algebraically closed fields on the computer. For the computation of invariant rings in the above setting,
+on the other hand, there is no need to deal with explicit elements of ``G`` or with its group structure. The implementation of Derksen's
```
fingolfin: That's not quite true... We have `algebraic_closure(QQ)` and the `QQBarField` it returns works quite well, and we are using it in parts of the book.

But of course we can't just use "the complex numbers" or something like that.

```suggestion
For the computation of invariant rings in the above setting, there is no need
to deal with explicit elements of ``G`` or with its group structure. The
implementation of Derksen's
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -21,22 +21,22 @@ mutable struct LinearlyReductiveGroup
...
         R, _ = polynomial_ring(fld, :z => (1:m,1:m))
         return LinearlyReductiveGroup(sym, m, R)
     end
 
     function LinearlyReductiveGroup(sym::Symbol, m::Int, pring::MPolyRing) #the ring input is the group ring
         #check char(field)
         G = new()
-        @assert sym == :SL 
-        fld = base_ring(pring)
+	fld = base_ring(pring)
```
fingolfin: Please no tabs
```suggestion
        fld = base_ring(pring)
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -21,22 +21,22 @@ mutable struct LinearlyReductiveGroup
...
+        @assert sym == :SL  && characteristic(fld) == 0 
         characteristic(fld) == 0 || error("Characteristic should be 0 for linearly reductive groups")
         G.field = fld
         @req m^2  == ngens(pring) "ring not compatible"
         G.group = (sym,m)
         G.reynolds_operator = reynolds_slm
-        M = matrix(pring, m, m, gens(pring))
+        M = transpose(matrix(pring, m, m, gens(pring)))
```
fingolfin, line 39: That's not a docu change but a change in behavior. Is this covered by a test? What is it fixing?

(Not arguing against this, I have no idea, but I also don't want this to accidentally slip through)

experimental/InvariantTheory/src/InvariantTheory.jl
fingolfin: Instead of "defined" I would go for something like "indicated" or so. Otherwise I agree with Johannes' suggestion.

docs/src/InvariantTheory/reductive_groups.md
wdecker: @joschmitt, @fingolfin How about moving all this this from experimental to non-experimental part?

experimental/InvariantTheory/src/InvariantTheory.jl
wdecker: This is correct now and fits with what we write in the book.

wdecker: It also fits with Sturmfels' book.

wdecker: And, it fits with the Disquisitiones Arithmetica.

experimental/InvariantTheory/src/InvariantTheory.jl
wdecker: @joschmitt As far as I can see, the ring is NOT graded for finite groups. Or?

joschmitt: It is graded:
```
julia> RG = invariant_ring(matrix_group(QQ[-1 0 ; 0 -1]));

julia> A, AtoR = affine_algebra(RG);

julia> A
Quotient
  of multivariate polynomial ring in 3 variables over QQ graded by
    y1 -> [2]
    y2 -> [2]
    y3 -> [2]
  by ideal (-y1*y3 + y2^2)
```

docs/src/InvariantTheory/reductive_groups.md
joschmitt: The idea of experimental is (to my understanding) that people can get stuff in easy and that the interface is still allowed to change. I do not know what the status of the code is. I assume that names like `fundamental_invariants` etc. are set in stone but maybe other things will change?
The other question (to me) is, who will maintain and extend this code in the future, for example, to include more groups than SL_n and tori. I'm happy to do minor maintenance work, but I don't see myself working with linearly reductive groups.

fingolfin: I just talked to @wdecker about this and I think we agree on the following: we don't add a remark about "experimental" here right now. Instead, one of the following will happen:
1. in a few weeks we have had enough time to review the code once more, clean it up, and promote it to non-experimental status
2. for some reason we decide against this, then we can still mark this as "experimental" somehow (however I think this scenario is unlikely)


Regarding the question
> who will maintain and extend this code in the future

I don't think it is fair to require a promise that code will be *extended* in the future in order to promote it (otherwise we should possibly *demote* some code right now). Sure, of course that's a goal, but I don't see how we can guarantee this, it depends on funding and finding the right people. (That said, @Lax202 is already working on implementing additional cases and algorithms)

However, I am happy to act as a maintainer down the road, in the sense of taking care of fixing bugs, and also as acting as advisor in the code, i.e. helping anyone who would like to extend it (and I'll try to find students to do so as well, but as I said I don't see how I could make any guarantees about that)

joschmitt: Okay, that's fine by me.

experimental/InvariantTheory/src/InvariantTheory.jl
fingolfin: As discussed: I am happy to trust you this is correct as in: fits what we document and fits our book.

That it fits other books or papers only is relevant if we establish that they use exact same conventions as we and hence are supposed to give identical results. I had a look at that a few months ago, but I didn't keep my notes, and don't have time to repeat this right now. Anyway, as I already wrote, I am happy to trust you, I just want it to be explicit and not "by oversight" ;-). 

BTW, if we verified it fits with the convention in those publications, but does not fit with the conventions in Derksen and Kemper; then I think we should actually point that out in the documentation, perhaps even with an example, as this is of course highly relevant for users of our code. And since we refer to e.g. Derksen/Derksen-Kemper in there, this is doubly important as a user might expect the reverse, or not even be aware of the differences in conventions.

(Ideally we would also state how it compares to Singular and Magma conventions, I guess?)

fingolfin: (That's not a blocker for this PR, of course, just something "nice to have")

wdecker:
@fingolfin, @joschmitt: We believe that we adressed all your remarks. Please check.

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -489,9 +609,55 @@ function reynolds_operator(X::RepresentationLinearlyReductiveGroup, elem::MPolyR
...
+"""
+@attr function affine_algebra(R::RedGrpInvRing)
+   V = fundamental_invariants(R)
+   s = length(V)
+   weights_ = zeros(Int, s)
+   for i in 1:s
+       weights_[i] = total_degree(V[i])
+   end
+   S,_ = graded_polynomial_ring(field(group(representation(R))), "t"=>1:s; weights = weights_)
+   R_ = poly_ring(R)
```
joschmitt, line 657:
```suggestion
   R_ = polynomial_ring(R)
```

joschmitt:
```suggestion
   R_ = polynomial_ring(R)
```

joschmitt (approved):
I added one more doctest.

fingolfin (approved):

PR #3444, opened 2024-02-27
add `od_from_atlas_group`, `od_from_p_subgroup`, and helpers

ThomasBreuer:
- `od_from_p_subgroup(chi, p)` tries to use the restriction of chi to the Sylow `p`-subgroup
- add `known_class_fusions`
- add optional argument to `induced_cyclic`
- add helper `possible_permutation_characters_from_sylow_subgroup`
- the new optional `min_char` argument in `isomorphic_group_over_finite_field` is used to guarantee that the reduction of an irreducible module is irreducibe

@fieker These changes are essentially the ones from #3005, which got closed for technical reasons.


experimental/OrthogonalDiscriminants/test/direct.jl
```diff
@@ -0,0 +1,17 @@
+@testset "compute via the Atlas of Group Representations" begin
+  # Check equality of positive results in small cases.
+  l = all_od_infos(comment_matches => "AGR", dim => 1:8);
+  for entry in l
+    chi = Oscar.OrthogonalDiscriminants.character_of_entry(entry)
```
fingolfin: Tests fail in this line:
```
  MethodError: no method matching getindex(::Nothing, ::Int64)
  Stacktrace:
    [1] character_of_entry(d::Dict{Symbol, Any})
      @ Oscar.OrthogonalDiscriminants ~/work/Oscar.jl/Oscar.jl/experimental/OrthogonalDiscriminants/src/data.jl:234
    [2] macro expansion
      @ ~/work/Oscar.jl/Oscar.jl/experimental/OrthogonalDiscriminants/test/direct.jl:5 [inlined]
```

benlorenz:
We need `od_from_p_subgroup` for the book.

fingolfin (approved):

experimental/OrthogonalDiscriminants/src/utils.jl
```diff
@@ -98,6 +100,303 @@ function reduce_mod_squares(val::AbsSimpleNumFieldElem)
...
+  for omega in column_orbits
+    for i in 1:n
+      push!(MM[i], sum(M[i][omega]))
+    end
+  end
+
+  # Compute the orbits on rows.
+  row_orbits = Vector{Int}[]
+  reps = []
+  poss = []
```
fingolfin, line 277: whenever I see `[]` I am think "not type stable, beware". We see a bunch of these in this PR. But perhaps it is fine for other reasons -- just wanted to draw your attention to it.

benlorenz:
@ThomasBreuer I just tried to backport this to finally get that part working but the return value for `od_from_p_subgroup` is very different from the jlcon in the book.

In the book (`expl_syl.jlcon`):
```julia
julia> ch = character_table("R(27)")[16];

julia> degree(ch)
18278

julia> Oscar.OrthogonalDiscriminants.od_from_p_subgroup(ch, 3)
(true, "-3")
```

Running this on the current Oscar master gives:
```julia
julia> ch = character_table("R(27)")[16];

julia> degree(ch)
18278

julia> Oscar.OrthogonalDiscriminants.od_from_p_subgroup(ch, 3)
(false, "")
```

ThomasBreuer:
@benlorenz I cannot reproduce the problem.
In the branch `TB_OD_pgroups` as well as in the current master, I get the same output as in the jlcon file of the book.

ThomasBreuer:
@benlorenz Ah, now I understand what happens. In my situation, a newer version of the GAP package CTblLib is used.

benlorenz:
> @benlorenz Ah, now I understand what happens. In my situation, a newer version of the GAP package CTblLib is used.

@ThomasBreuer Does that mean that at some point there will be a new CTbLib package that will be automatically picked up by the current Oscar (without any code changes here) and corrects this output?

ThomasBreuer:
@benlorenz Sorry for not being precise.
The example works with CTblLib version at least 1.3.5, which has been released in March 2023.
The last released version of GAP (4.12.2) contains only CTblLib 1.3.4, and this is the GAP version that comes with GAP.jl and thus with Oscar.jl.
I had not been aware of this problem. (I had expected that Oscar 1.0 would contain a newer GAP version, which will come with CTblLib at least 1.3.7.)

Forcing a newer version of CTblLib in Oscar does not work because it gets loaded already at startup of GAP.
We could only try to install a new version, and then the next Oscar session will load this one, but that does not help with the CI tests.

(Installing something at runtime is not a good idea. Nevertheless, currently we do this for the "repsn" package. Here the situation is different, this package does not get loaded automatically by GAP, thus Oscar can first try to upgrade and then load the package.)

benlorenz:
Maybe we could do something like https://github.com/oscar-system/GAP.jl/pull/855 and make a new package tarball that contains a newer CTblLib (but preferrably not many other changes) for GAP.jl. And then release this in a GAP.jl patch release 0.10.3. This would get picked up by the Oscar 1.0 release automatically and could fix the code that we want to have in the book? (Needs to be tested properly to make sure we don't break anything ...)

fingolfin:
We could do that, sure.

Or we wait 1-2 weeks, then GAP 4.13.0 will be out, and will be used in GAP.jl 0.11, and we can use that in OSCAR 1.0.1 or 1.1?

benlorenz:
Can't really say how much the GAP minor version bump might affect Oscar and whether we can put this in a patch release but otherwise yes that would work as well.

PR #3445, opened 2024-02-27
Bugfix for bugfix for printing of affine schemes

thofma:
Turns
```
julia> [Spec(L)]
1-element Vector{AffineScheme{QQField, Oscar.MPolyLocRing{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem, Oscar.MPolyPowersOfElement{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem}}}}:
AA^2 \ scheme((x)*(y)*(x*y))
```
into
```
julia> [Spec(L)]
1-element Vector{AffineScheme{QQField, Oscar.MPolyLocRing{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem, Oscar.MPolyPowersOfElement{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem}}}}:
AA^2 \ scheme(x*y*x*y)
```

I dedicate this to Dan S., who taught us to avoid parentheses like never before.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
@@ -81,7 +81,7 @@ function _show(io::IO, X::AbsAffineScheme{<:Any, <:MPolyQuoLocRing{<:Any, <:Any,
   else
     AA = AbstractAlgebra
     PP = AbstractAlgebra.PrettyPrinting
-    print(io, raw" \ ")
+    print(io, raw") \ ")
     AA.show_obj(io,
                 MIME("text/plain"),
                 PP.canonicalize(Expr(:call,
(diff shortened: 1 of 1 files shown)


HechtiDerLachs (approved):
Thanks! 

I have little idea of what's really happening in the background, but I am leaning towards trusting you and @tthsqe12 on their good work. With the missing bracket added, maybe we do not even need to fix the doctests.

Edit: It seems that I don't have the right to push to this pr. I will leave a suggestion, then.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
```diff
@@ -78,12 +78,18 @@ function _show(io::IO, X::AbsAffineScheme{<:Any, <:MPolyQuoLocRing{<:Any, <:Any,
...
     print(io, raw")")
-  elseif isone(length(denominators(S)))
-    print(io, raw") \ scheme(", first(denominators(S)), ")")
   else
-    print(io, raw") \ scheme((")
-    join(io, denominators(S), ")*(")
-    print(io, "))")
+    AA = AbstractAlgebra
+    PP = AbstractAlgebra.PrettyPrinting
+    print(io, raw" \ ")
```
HechtiDerLachs:
```suggestion
    print(io, raw") \ ")
```

fingolfin (approved):
Very nice. Can/should this be backported? (I don't know if it affects the book)

HechtiDerLachs:
Yes, I think it should be backported. At least #3437 must be backported, since the output is just wrong. Then this one doesn't seem to hurt on top of it.

PR #3451, opened 2024-02-28
Simplify for modules

### LINKED ISSUE #3108: `simplify` does not respect grading
```
X = rational_d10_pi9_quart_1();
I = defining_ideal(X);
Pn = base_ring(I)
n = ngens(Pn)-1
c = codim(I)
FI = free_resolution(I)
FIC = FI.C;
r = range(FIC)
C = shift(FIC[first(r):-1:1], -c)
F = free_module(Pn, 1)
OmegaPn = grade(F, [n+1])
D = hom(C, OmegaPn)
Omega = homology(D, 0);
is_graded(Omega)
SOmega, _ = simplify(Omega);
is_graded(SOmega)
```

HechtiDerLachs:
Makes `_alt_simplify` (which is also the workhorse behind `prune_with_map` in the generic case) the new `simplify` and resolves #3108.

### DIFF
--- a/experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
+++ b/experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
@@ -589,7 +589,17 @@ end
 
 ### Taylormade functionality for modules
 # Provided as a hotfix for issue #3108.
-function _alt_simplify(M::SubquoModule)
+@doc raw"""
+    simplify(M::SubquoModule)
+
+Simplify the given subquotient `M` and return the simplified subquotient `N` along
+with the injection map $N \to M$ and the projection map $M \to N$. These maps are
+isomorphisms.
+The simplifcation is heuristical and includes steps like for example removing
+zero-generators or removing the i-th component of all vectors if those are
+reduced by a relation.
+"""
+function simplify(M::SubquoModule)
   res, aug = free_resolution(SimpleFreeResolution, M)
   simp = simplify(res)
   simp_to_orig = map_to_original_complex(simp)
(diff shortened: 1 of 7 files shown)


thofma:
Since `_old_simplify` is buggy and not used, is there a reason to not remove it?

HechtiDerLachs:
We should ask @jankoboehm. It's his code and I don't know what it was introduced for originally and what its future purpose is.

HechtiDerLachs:
It is my impression that the function which is now `_old_simplify` is still somewhat experimental. It has another approach to the simplification which is different from the one used in the current `simplify` routine. So we might want to keep it at least for comparison.

wdecker:
1. @HechtiDerLachs Thx for doing this. @jankoboehm, @RafaelDavidMohr  As I realized now for the first time, there are TWO functions called `prune_with_map`, one based on `simplify` and one based on the corresponding Singular function (I used the latter one lately). I guess, calling the function will take the second implementation as this comes also second in the `presentation.jl` file. We must compare the behaviour of the two functions with respect to speed and correctness of results.

2. The docu text is very vague: `The simplifcation is heuristical and includes steps like for example removing
zero-generators or removing the i-th component of all vectors if those are
reduced by a relation.` Can we give a mathematical definition what exactly simplify is doing?

HechtiDerLachs:
> 1. We must compare the behaviour of the two functions with respect to speed and correctness of results.

This has been done multiple times in various PRs. The results are correct. My method (the one not using singular) finishes in comparable time, but still needs more memory. But it works generically over any ring, not just polynomial rings over fields.

jankoboehm:
_old_simplify is essentially equivalent to prune_with_map, but uses a different algorithm, which has the potential to be more efficient. Since it worked at some point, it should be not too difficult to fix. Addition of gradings is also something which is not too difficult. The code is actually predating OSCAR, so some improvements and optimizations are certainly possible. I will have a look. Bottom line is that we probably want the algorithm as an algorithmic option, in addition to the two implementations utilized now in prune_with_map (the one wrapped from Singular in cases where it applies, and the generic one taken from the generic code for the complexes). An of course we want to call the whole bunch simplify (or at least have such an allias), as we agreed on already some time ago, and is implemented here.

HechtiDerLachs:
> 2\. Can we give a mathematical definition what exactly simplify is doing?

I understand you are talking about what is now `_old_simplify` in this PR? That question goes to @jankoboehm . But since the function is now internal and experimental, I would say, it can stay as is, once this PR is merged. 

The real `simplify` does exactly the same thing as `prune_with_map` now: It considers a presentation of the module of a complex of free modules and looks for units in the matrices of the homomorphisms in order to find a homotopy equivalent complex with fewer generators in its terms.

wdecker (approved):

PR #3452, opened 2024-02-28
uses the schema during serialization testing

antonydellavecchia:
adds a schema to the data folder, adds the dependancy JSONSchema
and uses the scheam to validate during serialization test

and fixes spacing in the polyhedral geometry serialization test file


test/Serialization/setup_tests.jl
```diff
@@ -1,41 +1,47 @@
+using JSONSchema, JSON
```
lgoettgens:
```suggestion
using JSONSchema, Oscar.JSON
```
this way you could get rid of the new test dependency. The `test/Project.toml` is intended to be used for test-only dependencies.

benlorenz (approved):

PR #1224, opened 2022-03-31
Mention community page + various

lkastner:

PR #3457, opened 2024-02-28
Adjust to renaming of `rank(A::FinGenAbGroup)` to `torsion_free_rank(A::FinGenAbGroup)`

fingolfin:
The renaming would be done on the Hecke side <https://github.com/thofma/Hecke.jl/pull/1224>.

This is a breaking change, as code that currently uses `rank(A::FinGenAbGroup)` would either run into an error (current state of that PR), or even get "wrong" (well, "unexpected" perhaps) results. 

This PR tries to get ahead of it by changing all uses of `rank` that really want a "torsion free rank" to use that. (Actually: in all those places the groups in question are free, and then both "ranks" coincide; so this change here is not so much about making sure the code stays correct as about allowing us to ensure via OscarCI test that we caught all places that call `rank(A::FinGenAbGroup)` (or at least all that are actually being tested...).

The big question then is this: I think we have more or less consensus that we want this change, at least eventually. The big question is: can we still get it into 1.0?

I am not aware of any consequences of this PR for the book (which of course is not saying there are none...), but since it is a really very breaking change, if we don't have it in 1.0 we'd have to wait for 2.0. And if we do that, then perhaps we shouldn't merge this here either before we have a clear time line for 2.0, otherwise the difference between master and 1.x is too painful.

So while I'd love this PR and the one in Hecke to land and be backported to 1.0, ultimately the release managers (@benlorenz @aaruni96) should decide. (I've added the backports label as a kind of request, of course it can and should be removed if it is decided that it is not suitable)


fingolfin:
(I am also going to run the test suite on my computer to find more issues and will push updates here as I encounter and fix them)

fingolfin:
As it turns out, the toric varieties code uses `rank` on abelian groups extensively, e.g. `rank(picard_group(x))` and `rank(torusinvariant_cartier_divisor_group)` and more.

@HereAround are these groups generally torsion free (so the two kinds of rank coincide), or not?

benlorenz:
> So while I'd love this PR and the one in Hecke to land and be backported to 1.0, ultimately the release managers (@benlorenz @aaruni96) should decide. (I've added the backports label as a kind of request, of course it can and should be removed if it is decided that it is not suitable)

I think it makes sense to backport this and we will try to do so but we need it really soon, what is the current approach (and timeline) for this? Check that this runs in the OscarCI for Hecke, do a breaking Hecke release, adapt the version and merge this PR?

I will try to run this (together with the Hecke PR) for the book examples now.

thofma:
> > So while I'd love this PR and the one in Hecke to land and be backported to 1.0, ultimately the release managers (@benlorenz @aaruni96) should decide. (I've added the backports label as a kind of request, of course it can and should be removed if it is decided that it is not suitable)
> 
> I think it makes sense to backport this and we will try to do so but we need it really soon, what is the current approach (and timeline) for this? Check that this runs in the OscarCI for Hecke, do a breaking Hecke release, adapt the version and merge this PR?

Yes, that would be the plan. I could do a Hecke breaking release as soon as the OSCAR CI (with the matching branches) in https://github.com/thofma/Hecke.jl/pull/1224 shows green.

Can you report back about the book examples as soon as you have the result?

benlorenz:
> Can you report back about the book examples as soon as you have the result?

No new errors in the booktests and no unexpected output changes, running with [thofma/Hecke.jl@`bf86fb2` (#1224)](https://github.com/thofma/Hecke.jl/pull/1224/commits/bf86fb2ff3ef1347ede1779dc5f5f77694e10cb0), and on a branch (based on the backports branch) where I backported the first three commits on this PR.

experimental/GModule/Cohomology.jl
```diff
@@ -474,7 +474,7 @@ export action, cohomology_group, extension, pc_group_with_isomorphism
 export induce, is_consistent, istwo_cocycle, all_extensions
 export split_extension, extension_with_abelian_kernel
 
-Oscar.dim(C::GModule) = rank(C.M)
+Oscar.dim(C::GModule) = torsion_free_rank(C.M)
```
lgoettgens:
```suggestion
Oscar.dim(C::GModule) = rank(C.M)
```
This delegates to the module and not the group. So the change here is wrong

joschmitt: I'm not sure this is true. `C.M` is "the module", but there are some signatures with `C::GModule{..., FinGenAbGroup}` and the second type parameter is for `C.M` as far as I can see. Somebody who knows about the code should have a look.

lgoettgens: The CI in Hecke agrees that here we have a "module". See https://github.com/thofma/Hecke.jl/actions/runs/8094631614/job/22124040318#step:7:1375

lgoettgens: Oh, I think it can be both here. So we would need something like
```julia
Oscar.dim(C::GModule{..., FinGenAbGroup}) = torsion_free_rank(C.M)
Oscar.dim(C:GModule) = rank(C.M)
```

joschmitt: Yes, exactly. This is an example from the tests:
```
julia> k, a = cyclotomic_field(5)
(Cyclotomic field of order 5, z_5)

julia> zk = maximal_order(k)
Maximal order of Cyclotomic field of order 5 
with basis AbsSimpleNumFieldElem[1, z_5, z_5^2, z_5^3]

julia> U, mU = unit_group_fac_elem(zk)
(Z/10 x Z, UnitGroup map of Factored elements over Cyclotomic field of order 5
)

julia> A, mA = automorphism_group(PermGroup, k)
(Permutation group of degree 4, Map: permutation group of degree 4 -> generic group of order 4 with multiplication table -> set of automorphisms of Cyclotomic field of order 5)

julia> C = gmodule(A, mU)
G-module for A acting on U

julia> typeof(C.M)
FinGenAbGroup

```

experimental/GModule/GaloisCohomology.jl
```diff
@@ -1912,7 +1912,7 @@ function shrink(C::GModule{PermGroup, FinGenAbGroup}, attempts::Int = 10)
       o = Oscar.orbit(q, rand(gens(q.M)))
       if length(o) == order(group(q))
         s, ms = sub(q.M, collect(o), false)
-        if rank(s) == length(o)
+        if torsion_free_rank(s) == length(o)
```
joschmitt, line 1915: This is another place where we need to be careful about the type of `s` I think.

thofma: If I remember the idea correctly, the change should be correct. But @fieker should confirm.

experimental/GModule/Cohomology.jl
```diff
@@ -474,7 +474,7 @@ export action, cohomology_group, extension, pc_group_with_isomorphism
 export induce, is_consistent, istwo_cocycle, all_extensions
 export split_extension, extension_with_abelian_kernel
 
-Oscar.dim(C::GModule) = torsion_free_rank(C.M)
+Oscar.dim(C::GModule) = rank(C.M)
```
thofma: There is no `rank(::FinGenAbGroup` anymore in conjunction with https://github.com/thofma/Hecke.jl/pull/1224.

joschmitt: See the discussion above. `C.M` can also be a `FreeModule` (or something).

fingolfin: I find this `dim` function rather problematic to start with, there is a reason we use `rank` for modules in general... but that's besides the point right now

thofma: I'll push a fix

thofma:
OK https://github.com/thofma/Hecke.jl/pull/1224 is looking good. I'll tag 0.30.0 later today.

thofma:
OK, later today means now.

thofma:
Someone needs to approve and squash this.

experimental/GModule/GaloisCohomology.jl
fieker: Looks good

fieker:
As a comment: there are almost no requirements for the 2nd type. It needs to be a mathematical module, not a module type

HereAround:
> As it turns out, the toric varieties code uses `rank` on abelian groups extensively, e.g. `rank(picard_group(x))` and `rank(torusinvariant_cartier_divisor_group)` and more.
> 
> @HereAround are these groups generally torsion free (so the two kinds of rank coincide), or not?

In general they are not torsion free, but for my applications they always are. In any case, torsion-free rank should be the correct replacement/extension.

PR #3464, opened 2024-02-28
Ensure `fp_group(G)` transfers group attributes

fingolfin:
Minor niceness for the book


ThomasBreuer (approved):
Good.

Are we sure that the code line can be safely removed once the changes from pull request 5660 for GAP (which deals only with fp groups) are available?

PR #3465, opened 2024-02-28
Some more uses of GAPWrap

fingolfin:


ThomasBreuer (approved):
Looks o.k.

It is a little bit irritating that now `_get_iso_function` returns sometimes a GAP function and sometimes a Julia function. There is no problem as long as the result of `_get_iso_function` is just called (and not passed to GAP, for being stored somewhere).

fingolfin:
`_get_iso_function` should ultimately do `GAPWrap` for all the functions. Next time...

PR #3466, opened 2024-02-29
Fix a doc signature

joschmitt:
Documenter didn't find the correct doc string and used `gen(::SubquoModule)` instead.


thofma:
Did you test it locally? We unfortunately don't have a preview.

joschmitt:
Yes, this works locally.

thofma (approved):

PR #3467, opened 2024-02-29
Added comment on convention

wdecker:


experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -151,7 +151,11 @@ vector_space_dimension(R::RepresentationLinearlyReductiveGroup) = ncols(R.rep_ma
 @doc raw"""
     representation_on_forms(G::LinearlyReductiveGroup, d::Int)
 
-If `G` is the special linear group acting by linear substitution on forms of degree `d`, return the corresponding representation.
+If `G` is the special linear group acting by linear substitution on, say, `m`-ary forms of degree `d`, return the corresponding representation.
```
joschmitt:
```suggestion
If `G` is the special linear group acting by linear substitution on, say, `n`-ary forms of degree `d`, return the corresponding representation.
```
It's `n` everywhere else.

joschmitt (approved):

PR #3469, opened 2024-02-29
Grading + caching for affine algebra of torus invariants

Lax202:


experimental/InvariantTheory/src/TorusInvariantsFast.jl
```diff
@@ -264,14 +262,9 @@ julia> fundamental_invariants(RT)
...
 """
-function fundamental_invariants(z::TorGrpInvRing)
-    if isdefined(z, :fundamental)
-        return z.fundamental
-    else
-        R = z.representation
-        z.fundamental = torus_invariants_fast(weights(R), polynomial_ring(z))
-        return z.fundamental
-    end
+@attr function fundamental_invariants(z::TorGrpInvRing)
```
lgoettgens, line 265:
```suggestion
@attr SomeType function fundamental_invariants(z::TorGrpInvRing)
```
Please add the return type here to make the attribute lookup type stable

experimental/InvariantTheory/src/TorusInvariantsFast.jl
```diff
@@ -399,10 +392,14 @@ julia> affine_algebra(RT)
 (Quotient of multivariate polynomial ring by ideal (-t[1]*t[3] + t[2]^2), Hom: quotient of multivariate polynomial ring -> graded multivariate polynomial ring)
 ```
 """
-function affine_algebra(R::TorGrpInvRing)
+@attr function affine_algebra(R::TorGrpInvRing)
```
lgoettgens, line 395:
```suggestion
@attr SomeType function affine_algebra(R::TorGrpInvRing)
```
Here as well

experimental/InvariantTheory/src/TorusInvariantsFast.jl
Lax202: I don't know what type to use here other than Vector{MPolyRingElem}  or Vector{MPolyDecRingElem} - which are not concrete types.

Lax202: Faced the same issue with the other file - InvariantTheory.jl

lgoettgens: That's true. In most other places, we would add the field type as a type parameter to the invariant ring, which would then allow something smarter here.
But I think this is out of scope for this PR. 

@fingolfin can you keep this in mind for some future refactoring?

joschmitt: Yes, one would have to add parameters to the type `TorGrpInvRing` to make this more precise. I can have a look; I wanted to rename the types anyway in #3442.

joschmitt: Sorry, I typed to slow. We can keep this here as is and I look into it in #3442.

joschmitt (approved):
Looks good, I try to handle the return type business later.

joschmitt:
The doctest needs to be adjusted: The return type of `fundamental_invariants` is now more precise.

experimental/InvariantTheory/src/TorusInvariantsFast.jl
fingolfin: @lgoettgens sure, I actually had pointed out the same issue on a prior PR. We'll get there eventually ;-)

PR #3470, opened 2024-02-29
Speed up fundamental invariants of finite groups

joschmitt:
Use linear algebra to produce new invariants instead of the Reynolds operator in King's algorithm if the group is big.

Without this, computing fundamental invariants of
```
julia> K, i = cyclotomic_field(4, "i");

julia> q1 = diagonal_matrix(K[i 0; 0 i^-1], identity_matrix(K, 4));

julia> q2 = diagonal_matrix(K[0 -1; 1 0], identity_matrix(K, 4));

julia> q3 = K[0 0 1 0 0 0; 0 0 0 1 0 0; 0 0 0 0 1 0; 0 0 0 0 0 1; 1 0 0 0 0 0; 0 1 0 0 0 0];

julia> q4 = K[0 0 1 0 0 0; 0 0 0 1 0 0; 1 0 0 0 0 0; 0 1 0 0 0 0; 0 0 0 0 1 0; 0 0 0 0 0 1];

julia> G = matrix_group(q1, q2, q3, q4);

julia> RG = invariant_ring(G);
```
would take 20 to 30 minutes, now we are down to ~20 seconds.

I can't guarantee that there is absolutely no regression. In particular, the heuristic I implemented is a bit "magical" (see the comments). But I would hope that this is only slower in small examples, where a difference of a few milliseconds shouldn't matter. I did not notice any regression in the examples I tried though!

### DIFF
--- a/src/InvariantTheory/affine_algebra.jl
+++ b/src/InvariantTheory/affine_algebra.jl
@@ -65,7 +65,7 @@ Invariant ring
   of matrix group of degree 3 over K
 
 julia> affine_algebra(IR)
-(Quotient of multivariate polynomial ring by ideal with 1 generator, Hom: quotient of multivariate polynomial ring -> graded multivariate polynomial ring)
+(Quotient of multivariate polynomial ring by ideal (9*y1^6 + y1^3*y2^3 - 6*y1^3*y2*y3 + 3*y1^3*y4 - y2*y3*y4 + y3^3 + y4^2), Hom: quotient of multivariate polynomial ring -> graded multivariate polynomial ring)
 ```
 """
 function affine_algebra(IR::FinGroupInvarRing; algo_gens::Symbol = :default, algo_rels::Symbol = :groebner_basis)
--- a/src/InvariantTheory/fundamental_invariants.jl
+++ b/src/InvariantTheory/fundamental_invariants.jl
@@ -239,9 +239,9 @@ Invariant ring
 
 julia> fundamental_invariants(IR)
 4-element Vector{MPolyDecRingElem{AbsSimpleNumFieldElem, AbstractAlgebra.Generic.MPoly{AbsSimpleNumFieldElem}}}:
- x[1]^3 + x[2]^3 + x[3]^3
  x[1]*x[2]*x[3]
- x[1]^6 + x[2]^6 + x[3]^6
+ x[1]^3 + x[2]^3 + x[3]^3
+ x[1]^3*x[2]^3 + x[1]^3*x[3]^3 + x[2]^3*x[3]^3
  x[1]^3*x[2]^6 + x[1]^6*x[3]^3 + x[2]^3*x[3]^6
 ```
 """
--- a/src/InvariantTheory/fundamental_invariants.jl
+++ b/src/InvariantTheory/fundamental_invariants.jl
@@ -65,14 +65,44 @@ function fundamental_invariants_via_king(RG::FinGroupInvarRing, beta::Int = 0)
       end
     end
 
+    if is_zero(dimension_via_molien_series(Int, RG, d))
+      d += 1
+      continue
+    end
+
+    # There are two possible strategies to find new candidates in degree d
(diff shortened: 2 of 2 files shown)


joschmitt:
This does change the output of some examples in the invariant theory chapter in the book; not mathematically, but there are different polynomials returned because "fundamental invariants" are not unique.
So, I would not aim to have this pull request in 1.0. Still, it would be great if it would be in a released version in the not too distant future; it actually makes one of the examples in the book much faster.
How do we handle this? Is it allowed to change the output in this way in comparison to the printed version of the book in, say, version 1.1.0?

(See the fixed doctests for the kind of changes I'm talking about.)

wdecker (approved):
Excellent, thx!

wdecker:
With respect to the book and version 1.0: @ederc, @benlorenz, @lkastner: What do you think?

benlorenz:
Version 1.1 is definitely allowed to change the output. So once this is merged to master this will end up in 1.1.0 once that is released.

For 1.0 I would prefer not to include a new heuristic that is not that well tested and changes the output.

joschmitt:
That sounds reasonable. Thank you!

joschmitt:
I did some more serious benchmarking now which confirms the implemented heuristic. There were almost no cases where this pull request is slower and even in these only slightly. For some other cases, a different factor `X` in the heuristic would be somewhat better, but also there this is still better than what we have in `master` (which corresponds to `X = 0`).

For future reference, I tested with:
* In characteristic 0: complex reflection groups of type G(m, p, n) acting on V \oplus V^\ast, where V is the reflection representation. I did this with n = 2 and m <= 16. Would have done more, but I ran into #3482 .
* In positive characteristic: wreath products C_m \wr S_n where C_m is generated by a 1x1 matrix with a primitive m-th root of unity as entry again acting on V \oplus V^\ast and over a field which contains the root of unity and whose characteristic does not divide the group order. I did this up to group order 1000.

These groups are of course "very special", but I think they have interesting enough invariant theory.
In all cases I tested the values 0, 1, 10, 100, 1000, 1/2|G|, |G| for `X`.

wdecker:
@joschmitt Out of couriosity: What happens if you take Heisenburg groups H_p in the sense of our book article, say for p = 3, 5, 7.

joschmitt:
The first one is always (equivalent to) the current master, the second one this pull request.
```
julia> function heisenberg_group(p::Int)
           K, a = cyclotomic_field(p, "a")
           s = permutation_matrix(K, inv(cperm(1:p)))
           t = diagonal_matrix([a^i for i in 0:p - 1])
           return matrix_group(s, t)
       end
heisenberg_group (generic function with 1 method)

julia> RG = invariant_ring(heisenberg_group(3));

julia> @btime Oscar.fundamental_invariants_via_king(RG, X = QQ(0));
  2.822 ms (96901 allocations: 5.58 MiB)

julia> @btime Oscar.fundamental_invariants_via_king(RG, X = QQ(1)//2*order(group(RG)));
  1.568 ms (42631 allocations: 1.97 MiB)

julia> RG = invariant_ring(heisenberg_group(5));

julia> @btime Oscar.fundamental_invariants_via_king(RG, X = QQ(0));
  541.666 ms (12529280 allocations: 821.70 MiB)

julia> @btime Oscar.fundamental_invariants_via_king(RG, X = QQ(1)//2*order(group(RG)));
  101.082 ms (2446770 allocations: 128.11 MiB)

julia> RG = invariant_ring(heisenberg_group(7));

julia> @btime Oscar.fundamental_invariants_via_king(RG, X = QQ(0));
  149.433 s (1769841231 allocations: 78.18 GiB)

julia> @btime Oscar.fundamental_invariants_via_king(RG, X = QQ(1)//2*order(group(RG)));
  137.769 s (1624763537 allocations: 67.52 GiB)
```
So, still not really killing it, but well.

wdecker:
Comparing with the numbers we had in the early times, the progress  until now is really impressing!

PR #3473, opened 2024-03-01
Set version to 1.0.0

benlorenz:


fingolfin:
Oh, so are we doing this now? I thought the plan was to wait for this a bit longer to give us time past-book to improve the manual etc., so that we avoid potentially leaving a bad impression by having a 1.0 release with some major gaps (and things like "TODO: add this in") in the manual.

Not that I feel strong either way, I just don't see the point of having this now?

fingolfin:
(I don't mean to sound too negative, sorry -- I have no serious objections to this, I just feel confused, seems I misunderstood something)

benlorenz:
> Oh, so are we doing this now? I thought the plan was to wait for this a bit longer to give us time past-book to improve the manual etc., so that we avoid potentially leaving a bad impression by having a 1.0 release with some major gaps (and things like "TODO: add this in") in the manual.
> 
> Not that I feel strong either way, I just don't see the point of having this now?

We discussed this a bit in the last meetings and the consensus was that we want to have a released version now (strictly speaking it should have been two days ago) that corresponds exactly to the code examples from the book.

But after that we can keep backporting changes for the manual and maybe tiny bugfixes to do a 1.0.1 or 1.0.2 release before May. That version should then show up for the docs of 1.0. (For julia you can also go to the 1.9 docs and get the 1.9.4 version)

PR #3474, opened 2024-03-01
Fix `free_group` for `Char` arguments

fingolfin:
It used to do this:

    julia> F = free_group('x','y');

    julia> gens(f)
    1-element Vector{FPGroupElem}:
     xy

because for GAP lists of chars are strings and vice versa. The fix is to make sure all variable names get converted into GAP strings, no matter whether they are strings, symbols or chars on the Julia side.


ThomasBreuer (approved):
Nice.
(If there would have been a test for this situation, the problem would have been observed earlier ...)

fingolfin:
@ThomasBreuer yeah there is a lot of our code that is not being tested systematically

PR #3476, opened 2024-03-01
Rewrite Gap(::GModule) to use @attr macro

fingolfin:
Long term I think this should be turned into a GAP.Obj method, but I would like to avoid interfering too much with other ongoing work.

### DIFF
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -1032,14 +1032,10 @@ function Hecke.modular_proj(C::GModule{T, <:AbstractAlgebra.FPModule{AbsSimpleNu
   return R
 end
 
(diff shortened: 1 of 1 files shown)


mjrodgers (approved):
This looks good to me, and shouldn't interfere with anything.

What do you think about running `JuliaFormatter` on `experimental/GModule` and merging it along with this? is this a bad idea? I am hesitant to do it along with a bigger pull request, since it will make it look like everything has changed and could make it hard to review. But it would fix all of these spacing problems and pesky instances of `for i=X`. :D

fingolfin:
I am happy to run JuliaFormatter on that file -- but it should be in a commit of its own (or ideally a PR of its own). And ideally only if we have no major changes for the same file in flight, else resolving the conflicts may be a PITA.

mjrodgers:
Ah okay, maybe I will check with Claus and put that off until I'm sure it is a good time to do so.

PR #3478, opened 2024-03-01
add documentation about group libraries

### LINKED ISSUE #3427: Complete documentation for Group Libraries
There are TODO comments on https://docs.oscar-system.org/dev/Groups/grouplib -- clearly those most go in time for 1.0 resp. in time for the Begehung.

ThomasBreuer:
The additions were guided by the corresponding descriptions in the Groups chapter of the Oscar book.

(Note that currently the GAP packages `SOTGrps` and `SglPPow` that are mentioned in the book as extensions of the Small Groups Library are actually not loaded into Oscar.)

### DIFF
--- a/docs/oscar_references.bib
+++ b/docs/oscar_references.bib
@@ -133,6 +133,15 @@
   doi           = {10.1016/j.jsc.2012.07.002}
 }
 
+@Misc{BEO23,
+  author        = {Besche, Hans Ulrich and Eick, Bettina and O'Brien, Eamonn},
+  title         = {SmallGrp, The GAP Small Groups Library, Version 1.5.3},
+  note          = {GAP package},
+  year          = {2023},
+  month         = {5},
+  url           = {https://gap-packages.github.io/smallgrp/}
+}
+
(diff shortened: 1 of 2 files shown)


fingolfin:
We should load those extensions, but SOTGrps will only be available with GAP 4.13.0, so in 1-2 weeks.

ThomasBreuer:
O.k., then I add the two packages to the list in `src/Oscar.jl` (try to load but do not complain if not loadable), and extend the documentation of the small groups library accordingly.

ThomasBreuer:
I forgot to say that this pull request resolves #3427.

fingolfin (approved):

ThomasBreuer:
This got merged too quickly ...
The addition of the two packages to the list of GAP packages to load is missing.

I noticed some strange behaviour after that addition: Leaving Julia immediately after `using Oscar` then results in
```
fgets() failed with errno 5
Input/output error
```
which did not happen before the change.

(I had still pushed the change in order to see how the CI tests behave; but apparently the pull request got merged before this new commit arrived.)

PR #3479, opened 2024-03-01
reworked FreeAssAlgIdeals

Sequenzer:
I have changed the way the ideal_membership function works for `FreeAssAlgIdeals`. My changes make it faster and easier to use. To do this I wrote a `groebner_basis` function. This returns the Groebner basis of an ideal using letterplace or AbstractAlgebra.

The `ideal_membership` function now doesn't store the groebner basis it computes for an ideal `I`, which I think is fine, since if we were to set `I.gb` as the return of the `groebner_basis` function, this might lead a user to believe that this is a full groebner basis instead of a truncated one. Also, if you now set the `deg_bound` argument to `-1`, it will use the non-commutative Buchberger algorithm implemented in AbbstractAlgebra, which will return a full groebner basis if it terminates.


src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -115,32 +130,70 @@ function (R::Singular.LPRing)(a::FreeAssAlgElem)
...
+
+function _to_FreeAssAlgElem(A::FreeAssAlgebra, a::NCRingElem)
+    B = 0
+    for (c,e) in zip(Oscar.coefficients(a), Singular.exponent_words(a))
+        x = base_ring(A)(c)
+        if e == [] 
+            B+= x
+            continue
+        end
+        B+= x * (prod(FreeAssAlgElem[A[i] for i in e]))
```
antonydellavecchia: You should be using a build context here to build your elems,

```
B = MPolyBuildCtx(A)
push_term!(B, x, e)
finish(B)
```

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -45,17 +35,17 @@ end
 
 Return the two-sided ideal of $R$ generated by $g$.
 """
-function ideal(R::FreeAssAlgebra, g::Vector{T}) where T <: FreeAssAlgElem
+function ideal(R::FreeAssAlgebra, g::Vector{<:T}) where T <: FreeAssAlgElem
```
antonydellavecchia: you shouldn't need the T here
```suggestion
function ideal(R::FreeAssAlgebra, g::Vector{FreeAssAlgElem}) 
```

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -92,15 +82,40 @@ end
...
+I = ideal([f1])
+ideal_membership(f1, I, 4)
+
+# output
+true
+```
 """
-function ideal_membership(a::FreeAssAlgElem, I::FreeAssAlgIdeal, deg_bound::Int)
+function ideal_membership(a::FreeAssAlgElem, I::FreeAssAlgIdeal, deg_bound::Int=-1)
+  return ideal_membership(a, I.gens, deg_bound)
```
antonydellavecchia: try and use functions insteads of accessing properties
```suggestion
  return ideal_membership(a, gens(I), deg_bound)
```

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -92,15 +82,40 @@ end
...
+end
+function ideal_membership(a::FreeAssAlgElem, I::IdealGens{T}, deg_bound::Int=-1) where T <: FreeAssAlgElem
+  return ideal_membership(a, collect(I), deg_bound)
+end
+function ideal_membership(a::FreeAssAlgElem, I::Vector{T}, deg_bound::Int=-1) where T <: FreeAssAlgElem
```
antonydellavecchia: T isn't use here so you can just do
```suggestion
function ideal_membership(a::FreeAssAlgElem, I::Vector{FreeAssAlgElem}, deg_bound::Int=-1) 
```

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -115,32 +130,70 @@ function (R::Singular.LPRing)(a::FreeAssAlgElem)
...
+4 -> y^2*x + y*z*y
+```
+"""
+function groebner_basis(I::FreeAssAlgIdeal, deg_bound::Int=-1; protocol::Bool=false)
+  isdefined(I, :gb) && return I.gb
+  gb = groebner_basis(I.gens, deg_bound, protocol=protocol)
+  I.gb = gb
+  return gb
 end
+function groebner_basis(g::IdealGens{T}, deg_bound::Int=-1; protocol::Bool=false) where T <: FreeAssAlgElem
```
antonydellavecchia: same as above

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -115,32 +130,70 @@ function (R::Singular.LPRing)(a::FreeAssAlgElem)
...
+    gb = groebner_basis(collect(g), deg_bound, protocol=protocol)
+    return IdealGens(gb)
+end
+function groebner_basis(g::Vector{<:T}, deg_bound::Int=-1; protocol::Bool=false) where T <: FreeAssAlgElem
```
antonydellavecchia: same as above

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -92,15 +81,35 @@ end
 Return `true` if calculations with intermediate degrees bounded by `deg_bound`
 prove that $a$ is in $I$. Otherwise, a return of `false` indicates an
 inconclusive answer, but larger `deg_bound`s give more confidence in a negative
-answer.
+answer. If `deg_bound` is not provided, the default value is `-1`, which means 
+that no degree bound is imposed, which leads to a computation that uses a much
+slower algorithm, that may not terminate, but returns a full groebner basis if
```
lgoettgens: part of the sentence is missing

src/Rings/binomial_ideals.jl
```diff
@@ -672,7 +672,7 @@ Given a cellular binomial ideal `I`, return the associated primes of `I`.
 The result is defined over the abelian closure of $\mathbb Q$. In the output, if needed, the generator for roots of unities is denoted by `zeta`. So `zeta(3)`, for example, stands for a primitive third root of unity.
 
 # Examples
-```jldoctest
+```jldoctest; filter = r"(coeff \d+ already initialized\n|)"
```
lgoettgens, line 675: If there is non-deterministic verbose Singular.jl output, shouldn't this be an issue there to get fixed?

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -115,32 +124,63 @@ function (R::Singular.LPRing)(a::FreeAssAlgElem)
...
+
+julia> gb = groebner_basis(I, 3; protocol=false)
+Ideal generating system with elements
+1 -> x*y + y*z
+2 -> x^2 + y^2
+3 -> y^3 + y*z^2
+4 -> y^2*x + y*z*y
+```
+"""
+function groebner_basis(I::FreeAssAlgIdeal, deg_bound::Int=-1; protocol::Bool=false)
```
lgoettgens, line 156: Other parts of Oscar would rather use the printing mechanisms in https://docs.oscar-system.org/dev/Hecke/features/macros/#Printings instead of a kwarg. I don't want to say that this is the superior choice here, but wanted to mention it in case you were not aware of if.

src/Rings/binomial_ideals.jl
benlorenz: This seems to be fixed already via https://github.com/Singular/Singular/commit/aabd0c258f09541f944044c02f87388aee115919
but hasn't made it's way into the jlls. To avoid delaying this PR we added the filter.

julien-schanz (approved):

PR #3480, opened 2024-03-01
Resolution structure (and curve case)

afkafkafk13:
previously: Only for collaboration on this task! 

Now: structure of desingularization framework plus curve case with naive choice of center as the testing ground  for the structure


HechtiDerLachs:
Some tests still need to be adjusted. But other than that, I think this is rather fine as a working base. The adjustments can be made in the end, or at any time where we would like to have a good picture of the test suite.

afkafkafk13:
Thank you for the help and the preparation for monday.

afkafkafk13:
Currently untested: Framework for handling embedded setting.

HechtiDerLachs:
Today I contributed:

 - the method `find_refinement_with_local_system_of_params` which, given an affine scheme `W` returns a pair `U, d` where `U` is a `Covering` of `W` and `d` a dictionary with `d[U_i]` a tuple `(I, J, f)` with `I` a subset of the rows and `J` a subset of the columns of the jacobian matrix `A` for `W` such that the `(I, J)`-th minor of `A` is a unit `d` on `U_i`. 
 - the very similar method `find_refinement_with_local_system_of_params_rec` which does almost the same thing, but recursively and with a different output at the moment. 
 - `non_snc_locus(list)` which digests a list of `EffectiveCartierDivisor`s on a smooth scheme and returns an `IdealSheaf` indicating where these divisors are not snc. 

In particular, this required the construction of common refinements of coverings. These are not yet fully optimized, but at least functional. 

What did the others do? Let's wrap it up here a bit so that next week we know where to take it from. 
(@afkafkafk13, @jankoboehm, @simonbrandhorst, @wdecker)

simonbrandhorst:
@paemurru  worked on implementing normalizations of affine schemes. 
We started to work on lifting normalizations to gluings then hit a missing corner case for normalizations.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,620 @@
...
+  end
+
+  return Delta_list
+end
+
+
+
+function _delta_ideal_for_order(inc::CoveredClosedEmbedding, Cov::Covering, 
+       ambient_param_data::IdDict{<:AbsAffineScheme,
+                                  Tuple{Vector{Int64},Vector{Int64},<:QQMPolyRingElem}};
```
HechtiDerLachs:
```suggestion
                                  <:Tuple{Vector{Int64},Vector{Int64},<:RingElem}};
```

HechtiDerLachs: That would be my guess. But I can't confirm this at the moment since I'm working on something else.

afkafkafk13: Yes, this does the job. Thank you.

afkafkafk13:
I finished yesterdays work and implemented _delta_list and the locus of order at least b, based on yesterday's work of @HechtiDerLachs.

afkafkafk13:
> @paemurru worked on implementing localizations of affine schemes. We started to work on lifting normalizations to gluings then hit a missing corner case for normalizations.

@simonbrandhorst Did you want to say 'normalization of affine schemes'? What is the missing corner case? 

I looked up, how to specify the testideal in Singular: It is done by setting InputJ. This needs to be made accessible through Singular.jl.  
@wdecker : The warning message about this option is precisely, what is to be expected: If one ignores components of the singular locus by setting the testideal manually, the result will not be a normalization (because of the ignored points). This is precisely, what is needed in the setting we discussed yesterday.

wdecker:
To make the option accessible just check the `normalization` function in mpoly-affine-algebras.jl and add this possibility.
Here is a code sample from @fieker:

`julia> function bla(a::Int; one::Symbol = :f, kw...)`
`@show a, one`
`@show kw, typeof(kw)`
`end`
`bla (generic function with 1 method)`

`julia> bla(1, one=:b, two = 4)`
`(a, one) = (1, :b)`
`(kw, typeof(kw)) = (Base.Pairs(:two => 4),`
`Base.Pairs{Symbol, Int64, Tuple{Symbol},`
`@NamedTuple{two::Int64}})`
`(Base.Pairs(:two => 4), Base.Pairs{Symbol, Int64,`
`Tuple{Symbol}, @NamedTuple{two::Int64}})`

`julia> bla(1, one=:b, two = 4, go=5.5)`
`(a, one) = (1, :b)`
`(kw, typeof(kw)) = (Base.Pairs{Symbol, Real,`
`Tuple{Symbol, Symbol}, @NamedTuple{two::Int64,`
`go::Float64}}(:two => 4, :go => 5.5),`
`Base.Pairs{Symbol, Real, Tuple{Symbol, Symbol},`
`@NamedTuple{two::Int64, go::Float64}})`
`(Base.Pairs{Symbol, Real, Tuple{Symbol, Symbol},`
`@NamedTuple{two::Int64, go::Float64}}(:two => 4,`
`:go => 5.5), Base.Pairs{Symbol, Real, Tuple{Symbol,`
`Symbol}, @NamedTuple{two::Int64, go::Float64}})`

HechtiDerLachs:
This PR has already become quite big. So I suggest we bring this to a mergable state and then continue in a next step? 

@afkafkafk13 : Could you add some tests from the code snippets you're currently working with?

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,743 @@
...
+##################################################################################################
+# setting values in DesingMors -- Watch out: only place with direct access to fields!!!
+##################################################################################################
+function add_map!(f::AbsDesingMor, phi::BlowupMorphism)
+  push!(f.maps, phi)
+  ex_div = [strict_transform(phi,E) for E in f.ex_div[1:end]]
+  push!(ex_div, Oscar.exceptional_divisor(phi))
+  f.ex_div = ex_div
+  return f
+end
```
HechtiDerLachs: The signature of the method has an abstract type for `f`. Yet, in the internals we access the fields of a struct. That seems brittle and should in general be forbidden within the schemes framework. (I think some `Hecke.Map`s assume that a field `header` exists in concrete instances, but we don't follow this design here) 

Long story short: Can we use getters? Otherwise, please restrict the signature so that the fields accessed are sure to be there.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,743 @@
...
+  else
+    f.is_trivial = true
+  end
+  f.resolves_sing = false                                # we have no information, wether we are done
+                                                         # without further computation
+  f.is_embedded = false
+  return f
+end
+
+function add_map_embedded!(f::AbsDesingMor, phi::BlowupMorphism)
```
HechtiDerLachs: Same here.

afkafkafk13:
Now everything that needed to be in here is in and I will not add any more features in this PR. 

As @HechtiDerLachs suggested, the remaining goal is mergable state from here on.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,743 @@
...
+    return_value.resolves_sing = true
+    return_value.is_trivial = true
+    return return_value
+  end
+
+  ## I_sl non-empty, we need to do something 
+# here the keyword algorithm ensures that the desired method is called
+  dimX = dim(X)
+  if dimX == 1
+@show "overriding specified method for curves: use naive method"
```
HechtiDerLachs: TODO/memo: Remove before merge.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,743 @@
...
+  phi.resolves_sing = true
+  return(phi)
+end
+
+function _desing_emb_curve(f::CoveredClosedEmbedding, I_sl::IdealSheaf)
+  ## note: I_sl not unit_ideal_sheaf, because this has been caught before in embedded_desingularization(f)
+  decomp = Oscar.maximal_associated_points(pushforward(f)(I_sl))
+  I = small_generating_set(pop!(decomp))
+  current_blow_up = blow_up(I)
+  phi = initialize_embedded_blowup_sequence(current_blow_up,f)
```
HechtiDerLachs: Just a suggestion: One could add a type assertion here, i.e. finish the line with `...,f )::SomeConcreteTypeWeExpect`. 
That is not necessary for functionality, but since below we again have fields which are accessed directly, it might make bugs easier to fix (if return values change in the future?) and it helps other readers of the code to understand how the local variables are developed and what kind of objects we hold in our hands.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,743 @@
...
+function find_refinement_with_local_system_of_params_rec(
+    W::AbsAffineScheme, 
+    mod_gens::Vector{PolyType} = lifted_numerator.(gens(modulus(OO(W)))),
+    row_ind::Vector{Int} = Int[],
+    col_ind::Vector{Int} = Int[],
+    trans_mat::MatrixElem{RingElemType} = change_base_ring(OO(W), jacobi_matrix(mod_gens));
+    check::Bool=true
+  ) where {PolyType <: MPolyRingElem, RingElemType <: RingElem}
+  @show row_ind
+  @show col_ind
```
HechtiDerLachs: TODO/memo: Remove before merge.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: indeed: I wrote the getters after the functions and did not clean up.

HechtiDerLachs:
@afkafkafk13 : It looks like the tests are failing now because of some anticipated error messages. Could you have a look?

HechtiDerLachs:
@afkafkafk13 : I squashed all changes for the rebase. Be careful when you pull this now! Better do a backup of your local branch before that.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -149,43 +149,33 @@ ideal_sheaf(X::AbsCoveredScheme) = IdealSheaf(X)
...
     X::AbsCoveredScheme, U::AbsAffineScheme, 
-    g::Vector{RET}; check::Bool=false
-  ) where {RET<:RingElem}
-  I = ideal(OO(U), g)
+    I::Ideal; check::Bool=false
+  )
+  @assert base_ring(I) === OO(U) || error("ideal not defined in the correct ring")
   @check is_prime(I) "ideal must be prime"
   return PrimeIdealSheafFromChart(X, U, I)
+end
```
HechtiDerLachs, line 168:
```suggestion
end

function IdealSheaf(X::AbsAffineScheme, I::Ideal; covered_scheme::AbsCoveredScheme=CoveredScheme(X))
  @assert base_ring(I) === OO(X)
  @assert length(affine_charts(covered_scheme)) == 1 && X === first(affine_charts(covered_scheme))
  return IdealSheaf(covered_scheme, IdDict{AbsAffineScheme, Ideal}([X=>I]); check=false)
end
```

@afkafkafk13 : I propose to add and use this convenience method for your purposes. Also with a view towards the failing tests.

afkafkafk13: Thank you, This is a very convenient compatibility function.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,754 @@
...
+    I_sl_temp = pushforward(last_emb, Oscar.ideal_sheaf_of_singular_locus(domain(last_emb)))
+    decomp = Oscar.maximal_associated_points(I_sl_temp)
+  end
+
+  phi = _ensure_ncr(phi)
+  phi.resolves_sing = true
+  return phi
+end
+
+function _ensure_ncr(f::AbsDesingMor)
```
HechtiDerLachs:
```suggestion
function _ensure_ncr!(f::AbsDesingMor)
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,754 @@
...
+    f =_do_blow_up_embedded(f,cent)
+    if length(inter_div)>0
+      inter_div = [strict_transform(last_map(f),J) for J in inter_div]
+    end
+  end
+  return f
+end
+
+
+function _do_blow_up(f::AbsDesingMor, cent::AbsIdealSheaf)
```
HechtiDerLachs:
```suggestion
function _do_blow_up!(f::AbsDesingMor, cent::AbsIdealSheaf)
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,754 @@
...
+  old_sequence = maps(f)
+  X = domain(old_sequence[end])
+  X === scheme(cent) || error("center needs to be defined on same scheme")
+  current_blow_up = blow_up(cent,var_name=string("v", length(old_sequence), "_"))
+  add_map!(f, current_blow_up)
+  return(f)
+end
+
+function _do_blow_up_embedded(phi::AbsDesingMor,cent::AbsIdealSheaf)
```
HechtiDerLachs:
```suggestion
function _do_blow_up_embedded!(phi::AbsDesingMor,cent::AbsIdealSheaf)
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,754 @@
...
+    I = ordered_multi_index(i, codim, n)
+    J = ordered_multi_index(j, codim, r)
+    push!(ref_patches, U_ij)
+    minor_dict[U_ij] = (indices(I), indices(J), M_ext[i, j])
+  end
+  res_cov = Covering(ref_patches)
+  inherit_glueings!(res_cov, Covering(W))
+  return res_cov, minor_dict
+ #=
+  all_entries = Vector{Int}[[i, j] for i in 1:n for j in 1:r]
```
HechtiDerLachs: These lines can probably go now?

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,754 @@
...
+end
+
+function _delta_ideal_for_order(inc::CoveredClosedEmbedding, Cov::Covering, 
+       ambient_param_data::IdDict{<:AbsAffineScheme,
+                                 <: Tuple{Vector{Int64},Vector{Int64},<:RingElem}};
+       check::Bool=true)
+
+  W = codomain(inc)                                
+  @check is_smooth(W) "codomain of embedding needs to be smooth"
+#  @check is_equidimensional(W) "codomain of embedding needs to be equidimensional"
```
HechtiDerLachs, line 528: Is this still disabled intentionally?

HechtiDerLachs (approved):

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Yes, because this is not systematically set right now at other places in the CommAlg and AlgGeom context. So we cannot rely on it being set and it is expensive to compute (using equidimensionality check in each chart).
I will make a tour through different functions making sure that is_irreducible, is_reduced, is_equidimensional, is_integral etc. are set and used systematically.  But this is for another PR.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Feel free to throw them away. You were the one who wanted to keep them in the first place.

HechtiDerLachs: Yes, I remember. But I wasn't sure whether this internal helper function had already grown to everyone's full satisfaction, or whether we still wanted to sample from that snippet. Thanks for the cleanup.

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -3,6 +3,36 @@ export center
 export exceptional_divisor
 export projection
 
+@doc raw"""
+    AbsDesingMor{
```
simonbrandhorst, line 7:
```suggestion
    AbsDesingMor{
```
Please consider renaming to `AbsDesingularizationMor`.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+  push!(f.maps, phi)
+  ex_div = [strict_transform(phi,E) for E in f.ex_div[1:end]]
+  push!(ex_div, Oscar.exceptional_divisor(phi))
+  f.ex_div = ex_div
+  return f
+end
+
+function initialize_blow_up_sequence(phi::BlowupMorphism)
+  f = BlowUpSequence([phi])
+  f.ex_div = [Oscar.exceptional_divisor(phi)]
```
simonbrandhorst:
```suggestion
  f.ex_div = [exceptional_divisor(phi)]
```
and elsewhere

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+exceptional_divisor_list(phi::BlowUpSequence) = phi.ex_div  ## derzeit Liste von Eff. Cartier Div.
+embeddings(phi::BlowUpSequence) = phi.embeddings
+
+## do not use!!! (for forwarding and certain emergenies)
+function underlying_morphism(phi::AbsDesingMor)
+  if !isdefined(phi, :composed_map)
+    len=length(maps(phi))
+    result=underlying_morphism(maps(phi)[1])
+    for i in 2:len
+      result = compose(underlying_morphism(maps(phi)[i]), result)
```
fingolfin, line 56: If I understand your code right, then `maps(phi)` each time returns a fresh copy of the vector maps. So better to call it only once. (This might affect other places, too, I didn't check)
```suggestion
    m = maps(phi)
    len = length(,)
    result = underlying_morphism(m[1])
    for i in 2:len
      result = compose(underlying_morphism(m[i]), result)
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+    end
```
fingolfin: If I understand your code right, then `maps(phi)` each time returns a fresh copy of the vector maps. So better to call it only once. (This might affect other places, too, I didn't check)
```suggestion
    m = maps(phi)
    len = length(m)
    result = underlying_morphism(m[1])
    for i in 2:len
      result = compose(underlying_morphism(m[i]), result)
    end
```

Alrernatively you could also use `mapreduce` here:
```suggestion
    result = mapreduce(underlying_morphism, compose, maps(phi))
```
That should be equivalent... maybe the loop is clearer, though. Just wanted to mention it as an option

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+  return phi.composed_map
+end
+
+##################################################################################################
+# setting values in DesingMors -- Watch out: only place with direct access to fields!!!
+##################################################################################################
+function add_map!(f::BlowUpSequence, phi::BlowupMorphism)
+  push!(f.maps, phi)
+  ex_div = [strict_transform(phi,E) for E in f.ex_div[1:end]]
+  push!(ex_div, Oscar.exceptional_divisor(phi))
```
fingolfin: Is the `Oscar.` necessary? If yes, should `exceptional_divisor` be exported (resp. imported here)?
```suggestion
  push!(ex_div, exceptional_divisor(phi))
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+end
+
+function initialize_blow_up_sequence(phi::BlowupMorphism)
+  f = BlowUpSequence([phi])
+  f.ex_div = [Oscar.exceptional_divisor(phi)]
+  if !is_one(center(phi))
+    f.is_trivial = false
+  else
+    f.is_trivial = true
+  end
```
fingolfin:
```suggestion
  f.is_trivial = is_one(center(phi))
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+  f.resolves_sing = false                                # we have no information, wether we are done
+                                                         # without further computation
+  f.is_embedded = false
+  return f
+end
+
+function add_map_embedded!(f::BlowUpSequence, phi::BlowupMorphism)
+  push!(f.maps, phi)
+  ex_div = typeof(f.ex_div[1])[strict_transform(phi, E) for E in f.ex_div[1:end]]
```
fingolfin, line 82: Is the `[1:end]` really necessary here?
```suggestion
  ex_div = typeof(f.ex_div[1])[strict_transform(phi, E) for E in f.ex_div]
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+      if length(decomp)>0 
+        decomp = [strict_transform(last_map(phi),J) for J in decomp]
+      end
+    end
+    I_sl_temp = Oscar.ideal_sheaf_of_singular_locus(domain(last_map(phi)))
+    decomp = Oscar.maximal_associated_points(I_sl_temp)
+  end
+
+  phi.resolves_sing = true
+  return(phi)
```
fingolfin:
```suggestion
  return phi
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+    end
+  end
+  new_keys = copy(empty_keys)
+
+# add intersections
+  while !is_empty(old_keys)
+    old_keyvec = popfirst!(old_keys)         # this is the intersection to which we add a new divisor
+    for i in (old_keyvec[end]+1):length(current_div)
+      copykey = copy(old_keyvec)
+      if haskey(inter_div_dict, [i])
```
fingolfin: No need to make a copy if you then don't need it
```suggestion
      if haskey(inter_div_dict, [i])
        copykey = copy(old_keyvec)
```

Personally I also prefer not to nest logic to deeply, as it makes it harder to read, so I probably would do
```suggestion
      haskey(inter_div_dict, [i]) || continue
      copykey = copy(old_keyvec)
```
and then adjust the indentation of the following code.

And then do something similar for `if !is_one(Idiv)` below, changing it to `!is_one(Idiv) || continue` or `is_one(Idiv) && continue` (which one is a matter of taste)

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+          elseif dim(Idiv) == n_max - length(copykey)
+                                             # non-offending intersection
+            inter_div_dict[copykey] = (Idiv,0)
+            push!(new_keys,copykey)
+          else
+                                             # offending intersection, new
+            inter_div_dict[copykey] = (Idiv,1)
+            push!(new_keys,copykey)
+            push!(essential_inter, Idiv)
+          end
```
fingolfin: Extract common code?
```suggestion
          if (sum([inter_div_dict[a][2] for a in subsetlist])> 0)
                                             # offending intersection, known before
            inter_div_dict[copykey] = (Idiv,2)
          elseif dim(Idiv) == n_max - length(copykey)
                                             # non-offending intersection
            inter_div_dict[copykey] = (Idiv,0)
          else
                                             # offending intersection, new
            inter_div_dict[copykey] = (Idiv,1)
            push!(essential_inter, Idiv)
          end
          push!(new_keys, copykey)
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+  com_ref, incs = common_refinement(triv_cov, default_covering(X))
+
+  ideal_dict = IdDict{AbsAffineScheme, Ideal}() # ideal sheaf of the non_snc_locus
+  for U in patches(com_ref)
+    loc_eqns = elem_type(OO(U))[]
+    for k in 1:length(incs)
+      I = ideal_sheaf(divs[k])
+      inc = incs[k][U]
+      V = codomain(inc)
+      h = first(gens(I(V)))
```
fingolfin: This should avoid an allocation?
```suggestion
      h = gen(I(V), 1)
```

fingolfin (approved):
@afkafkafk13 asked me to have a look at the technical side of this PR (I can't comment on the math). Overall it seems fine to me. I left a few suggestions, but they are mostly optional (maybe handle the one with the many calls to `maps` which each allocate, but even that can be done later if you prefer)

HechtiDerLachs:
@fingolfin : Thanks for having a look. I agree with your suggestions. 

Concerning the maps: I propose to introduce a method `map(f, i)` here as a direct getter in accordance with what we already have for `ComplexOfMorphism`s. I know that this is debated as `map` also stands for something more general. But it is what we have in Oscar in many other cases at the time and it would only be coherent if we introduced it also here. If at some later point we decide to clean this up, then its usage will also be cleaned up here.

fingolfin:
If `Oscar.map` is a distinct (and then necessarily non-exported) function from `Base.map`, I guess we could live with it... But overall this I find this very problematic, and it runs completely counter to the definition of `Base.map`, and even has the same arity as it has.

IMHO at the very least this needs to be run by @fieker and @thofma.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -0,0 +1,738 @@
...
+#    end
+#    return new{typeof(domain(maps[end])),typeof(codomain(first(maps)))}(maps)
+#  end
+#end
+
+
+##################################################################################################
+# getters
+##################################################################################################
+maps(phi::AbsDesingMor) = copy(phi.maps)
```
simonbrandhorst: Why copy? Typically our getters do not copy stuff. The user should assume that anything returned is unsafe to mutate?
... but it is a common pitfall

simonbrandhorst:
How about replacing `maps` by `morphisms`? Then there is no clash.

experimental/Schemes/Resolution_structure.jl
HechtiDerLachs: I suggest not to return the fully composed morphism here, but an instance of `CompositeCoveredSchemeMorphism`. Everything is already implemented for that one and like this all the special functionality for the latter will automatically be harvested.

experimental/Schemes/Resolution_structure.jl
HechtiDerLachs: @afkafkafk13 and I have talked about this. I suggested to return a copy. Otherwise, the chances are high that unsafe code leads to a mess with the internals of the structure. And this tends to be almost impossible to debug. The way to solve this, is to provide the getter with the index and encourage the user to use that instead of creating copies all over the place. 

And I could live with `morphisms` and `morphism` below.

afkafkafk13: taking your other suggestions for `morphisms` instead of  `maps` into account, I would go for

> morphisms(phi::AbsDesingMor) = phi.maps
> morphisms_copy(phi::AbsDesingMor) = copy(phi.maps)
> morphism(phi::AbsDesingMor, i::Int) = phi.maps[i]

Do you agree? Note that morphisms_copy is necessary at certain places, because the desingularization has to modify it by appending. I will carefully go through the existing code and change to morphisms_copy precisely, where it is necessary.

HechtiDerLachs: No `morphisms_copy`, please! `morphisms` should create a copy each time and not give access to the internals. Such things have proved to be devastating in many instances. IMHO saving memory allocations has to be achieved by other means and this is also possible, e.g. by encouraging the use of `morphism(phi, i)`.

HechtiDerLachs: > Typically our getters do not copy stuff.

I do not think that this is true. Even `gens` for polynomial rings usually creates a copy. And it does so for good reasons.

afkafkafk13: ok then just morphisms(phi) and morphism(phi,i) giving back copies. I will make the appropriate changes.

simonbrandhorst: best rename `.maps` to `.morphisms` as well for consistency.

fingolfin: Regarding `copy` in `morphisms` (and similar elsewhere in `gens` etc.): for OSCAR 2.0 we could consider using a type like the one defined in https://github.com/bkamins/ReadOnlyArrays.jl which implements the `AbstractArray` / `AbstractVector` interface by wrapping an existing array/vector and passing all operations through -- except anything that *writes* to the vector. The whole thing is basically "free" in terms of overhead.

But the downside is that the result is not a `Vector` anymore -- code which accepts `AbstractVector` will keep working, but surely a lot of code will need (trivial) adjustments -- hence this can't go into Oscar 1.x


In this regard, it would be much better if `Array` just had a bit flag that turned any attempt to write to it into an error

afkafkafk13: > best rename `.maps` to `.morphisms` as well for consistency.

That was part of the content of my last statement.

HechtiDerLachs:
@afkafkafk13 : I took the liberty to implement the `AbsBlowdownMorphism` interface we once worked out. Simply because I was curious whether this was doable and we were recently working on this topic with @HereAround for the toric blowups. 

In particular, I resolved the suggestion about the underlying morphism which I pointed out above.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -736,3 +731,58 @@ end
...
+  end
+  return phi.exceptional_divisor
+end
+
+function exceptional_locus(phi::BlowUpSequence)
+  if !isdefined(phi, :exceptional_locus)
+    phi.exceptional_locus = weil_divisor(exceptional_divisor(phi))
+  end
+  return phi.exceptional_locus
+end
```
afkafkafk13, line 745: This is the point where we need to discuss. For non-embedded desingularization this is what we need. But in the embedded case and for principalization of Ideal Sheaves we do weak transforms and need to store multiplicities along with the reduced structure.
I could envision that we additionally add a function which returns a Weil divisor with the appropriate multiplicities -- from the tradition of the resolution of singularities community, this would be called `monomial_part`. But this name is absolutely impossible in the computeralgebra context, because it is taken otherwise. So I would envision `exceptional_factors_with_multiplicities`.

HechtiDerLachs: I am not sure that I get 100% what you mean here. Let's discuss this in person at some point. 

Until then: Does my implementation give wrong results? If not, it sounds to me as if we could leave it as is for the moment and later add further functions or modify the implementation.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Ups, leftovers from debugging...... Thanks for pointing this out.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: It is fine for now and I am thankful that you included these functions. It was just the explanation of the additional function which needs to come later on (e.g. for BEV).

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

PR #3483, opened 2024-03-01
Add `is_conjugate*` to `docs/src/Groups/action.md`

### LINKED ISSUE #3454: `is_conjugate_with_data(Omega::GSet, omega1, omega2)` missing from "Group Actions" section in the manual
We already have a docstring for it, but there is no hint about this function on https://docs.oscar-system.org/dev/Groups/action

It should be added.

Going beyond this, we should look through src/Groups/gsets.jl to see if any other things in there are missing from that manual section and should be mentioned there one way or another.

mjrodgers:
Closes #3454 

Simple oversight, these docstrings belong here.

@benlorenz I think should be backported also.

### DIFF
--- a/docs/src/Groups/action.md
+++ b/docs/src/Groups/action.md
@@ -61,6 +61,8 @@ permutation
 acting_group(Omega::GSetByElements)
 action_function(Omega::GSetByElements)
 action_homomorphism(Omega::GSetByElements{T}) where T<:GAPGroup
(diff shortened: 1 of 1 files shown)


benlorenz:
I added the label to backport this for 1.0.1.

ThomasBreuer (approved):
Thanks.

PR #3490, opened 2024-03-05
Lazy ideal sheaves

HechtiDerLachs:
Makes ideal sheaves more lazy. 

This turned out to be necessary to progress with the quest for Oguiso's automorphism.

CC: @simonbrandhorst

### DIFF
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -7,6 +7,7 @@ export subscheme
 
 ### Forwarding the presheaf functionality
 underlying_presheaf(I::IdealSheaf) = I.I
+underlying_presheaf(I::PrimeIdealSheafFromChart) = I.F
 
 # an alias for the user's convenience
 scheme(I::IdealSheaf) = space(I)
@@ -247,7 +248,7 @@ end
 #  return IdealSheaf(X, C, new_dict)
 #end
 
-function +(I::IdealSheaf, J::IdealSheaf)
+function +(I::AbsIdealSheaf, J::AbsIdealSheaf)
   X = space(I)
   X == space(J) || error("ideal sheaves are not defined over the same scheme")
   new_dict = IdDict{AbsAffineScheme, Ideal}()
@@ -258,7 +259,7 @@ function +(I::IdealSheaf, J::IdealSheaf)
   return IdealSheaf(X, new_dict, check=false)
 end
 
-function *(I::IdealSheaf, J::IdealSheaf)
+function *(I::AbsIdealSheaf, J::AbsIdealSheaf)
   X = space(I)
   X == space(J) || error("ideal sheaves are not defined over the same scheme")
   new_dict = IdDict{AbsAffineScheme, Ideal}()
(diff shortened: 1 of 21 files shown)


HechtiDerLachs:
This is now ready for merge. 

The PR contains several more changes towards a more general overhaul of coherent sheaves. I learned some more things during the preparation of the hypercomplexes and now I will, step by step, adapt the coherent sheaves accordingly. 

At the moment this is in a somewhat intermediate state. But that should be OK, because doing all at once would be too big a pull request anyway. I already put the structure sheaf and ideal sheaves in the new setup, but sheaves of modules yet have to wait a bit. They still work just as fine, though. 

I'm afraid there are few people who are really qualified and willing to look into all this in detail and I might be the only one who has a sufficiently good overview over this part of the code at this point. But let me just say: All what is substantially affected is in `experimental`. And: I did not change anything about the tests. They still run and the code behaves correctly. 

So all in all I would be happy if anyone volunteered to superficially approve the changes so that I can continue with the actual work on the Oguiso project. 

CC: @afkafkafk13 @simonbrandhorst .

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2267,10 +2267,10 @@ If the localization is at a point, a minimal set of generators is returned.
   L = localized_ring(Q)
 
   J = pre_image_ideal(I)
-  return filter(!iszero, Q.(small_generating_set(J)))
+  return unique!(filter(!iszero, Q.(small_generating_set(J))))
```
afkafkafk13, line 2270: Just out of curiosity: Do you have an small example, where the `unique!` is indeed necessary?

HechtiDerLachs: For quotient rings the small generating set is computed over the polynomial ring. Then generators might fall together under the reduction map.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -999,13 +1038,21 @@ end
   return IdealSheaf(X, ID, check=false)
 end
 
-function small_generating_set(II::IdealSheaf)
+radical(I::PrimeIdealSheafFromChart) = I
+
+# TODO: This function should be removed for ideal sheaves! 
+# Reason: There is no "generating set" for ideal sheaves and even if there was, 
+# this is not what the function returns. Whatever it is doing should probably 
+# be subsumed under `simplify` and `simlpify!`.
```
fingolfin, line 1046:
```suggestion
# be subsumed under `simplify` and `simplify!`.
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -247,35 +258,49 @@ end
...
+  return SumIdealSheaf(vcat(summands(I), [J]))
+end
+
+function +(I::AbsIdealSheaf, J::SumIdealSheaf)
+  return SumIdealSheaf(vcat([I], summands(J)))
+end
+
+function +(I::SumIdealSheaf, J::SumIdealSheaf)
+  return SumIdealSheaf(vcat(summands(I), summands(J)))
+end
```
fingolfin, line 275: You could simplify this to
```suggestion
function +(I::AbsIdealSheaf, J::AbsIdealSheaf)
  return SumIdealSheaf(vcat(summands(I), summands(J)))
end
```
and similar for the `*` methods coming next if you simply add
```julia
summands(I::AbsIdealSheaf) = [I]
```

afkafkafk13 (approved):
Looks fine overall. 

I have a few minor questions (not blocking the PR, just for understanding the inner workings better), which I will ask in person today.

HechtiDerLachs:
@aaruni96 , @benlorenz : Can I merge this? Or will this produce any conflicts with the book or something alike?

aaruni96:
If I understand correctly, you can merge anything you want into master, without affecting 1.0 or the book (and as long as it doesn't introduce any backwards incompatibilities, it can then eventually be ported to Oscar 1.x). Right, @benlorenz  ?

benlorenz:
Everything that is merged into master should maintain compatibility with existing code that works with 1.0.0 (for example in the book, but also anywhere else). We will only backport small bugfixes or documentation improvements for 1.0.1.
The output is allowed to change for master (1.1).

HechtiDerLachs:
Just to be clear: I do not expect these changes here to be backported. They affect printing and some internals, but not the user-facing calls.

PR #3494, opened 2024-03-06
Normalization of covered schemes

paemurru:
Work in-progress.

At the moment only computes normalizations of integral schemes.


afkafkafk13:
@paemurru : I resolved (among others) the following minor problems of this PR
- scheme(I) provides the underlying scheme of an ideal sheaf I, but you needed subscheme(I), which is the subscheme of the underlying scheme defined by I   
- handling of empty affine charts (V(1)) was still missing
- property caching (in the background) was still somewhat incomplete 

@HechtiDerLachs @simonbrandhorst : We should rethink the naming of scheme(I); if Erik is already confused by it, the users will be as well. Maybe the name scheme(I) should not exist at all or only print the information to use underlying_scheme or subscheme depending on the desired scheme. In any case, it needs to replaced by something that clarifies in the name that it refers to the underlying scheme.

HechtiDerLachs:
> We should rethink the naming of scheme(I)

I do not agree. A sheaf of ideals is first of all a sheaf. A sheaf is defined on a scheme. `scheme(I)` is the straightforward way to ask for that information. `subscheme(I)` is the straightforward way to obtain a subscheme. The users can read the manual and everything will be clear. I'm sure that @paemurru would eventually have found and corrected this mistake himself and so will other users.

HechtiDerLachs:
> > We should rethink the naming of scheme(I)
> 
> I do not agree. A sheaf of ideals is first of all a sheaf. A sheaf is defined on a scheme. `scheme(I)` is the straightforward way to ask for that information. `subscheme(I)` is the straightforward way to obtain a subscheme. The users can read the manual and everything will be clear. I'm sure that @paemurru would eventually have found and corrected this mistake himself and so will other users.

Edit: `underlying_scheme` is internal and should not be exported. @HereAround once suggested to have `underlying_scheme` as a user-facing forgetful functor for toric varieties. But we decided to give `forget_toric_structure` to the user instead. In this spirit I am also against bringing `underlying_scheme` in the game as a candidate for the above getter.

afkafkafk13:
> > > We should rethink the naming of scheme(I)
> > 
> > 
> > I do not agree. A sheaf of ideals is first of all a sheaf. A sheaf is defined on a scheme. `scheme(I)` is the straightforward way to ask for that information. `subscheme(I)` is the straightforward way to obtain a subscheme. The users can read the manual and everything will be clear. I'm sure that @paemurru would eventually have found and corrected this mistake himself and so will other users.
> 
> Edit: `underlying_scheme` is internal and should not be exported. @HereAround once suggested to have `underlying_scheme` as a user-facing forgetful functor for toric varieties. But we decided to give `forget_toric_structure` to the user instead. In this spirit I am also against bringing `underlying_scheme` in the game as a candidate for the above getter.

I see the point. Then we should just keep in mind that we add a note in the .md file on Ideal Sheaves, when it is time to write this documentation. Without a prominent explicit note stating the naming choice, this can become a never ending source of user questions (by occasional users) -- I have seen such things in Singular in the past.

simonbrandhorst:
This should be good to go now. I added a bunch of tests, documentation and fixed some bugs.

simonbrandhorst:
this implements a brute force approach where each patch is normalized individually and no information is passed around

normalization for localized quotient rings can be done later in another pr

test/AlgebraicGeometry/Schemes/CoveredScheme.jl
afkafkafk13, line 1: Just for my understanding: Why do you nest the testsets? Is my following guess right?
- In the case of success, the user should only see a single success
- In the case of failure, the user should still gets a more detailed feedback as what innermost testset failed

afkafkafk13 (approved):
The code is fine and should be included quickly.

test/AlgebraicGeometry/Schemes/CoveredScheme.jl
simonbrandhorst: This is my understanding. But to be honest I did not check.

PR #3495, opened 2024-03-06
Tjurina Algebra/Number and finite determinancy for Hypersurfaces

Syz-MS:
Adds 
-Tjurina algebra and number for  Hypersurfaces
-upper determinancy bounds w.r.t. right resp. contact equivalence for IHS
-contact equivalence via Mather-Yau for IHS

(developed at the request of @afkafkafk13)


experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+julia> tjurina_number(X)
+2
+```
+"""
+function tjurina_number(X::HypersurfaceGerm, k::Integer = 0)                                    #TODO: simplify after vec_dim is improved
+  # Fix for zero dimensional vector space
+  I = modulus(tjurina_algebra(X, k))
+  if base_ring(I)(1) in I
+    return 0
+  end
```
afkafkafk13: Maybe you would like to say:

R = tjurina_algebra(X,k)
!is_one(modulus(R)) || return 0

and then use R at later occurrences of the Tjurina algebra

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  # Fix for infinite dimensional vector space
+  try
+    return vector_space_dimension(tjurina_algebra(X, k))
+  catch
+    return -1
+  end
```
afkafkafk13: You could test for the dimension of the Tjurina algebra and return -1, if it exceeds zero.

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  return MPolyQuoRing(R, ideal(R, f) + jacobian_ideal(f))
+end
+
+
+
+@doc raw"""
+    tjurina_algebra(X::AffineScheme{<:Field,<:MPolyQuoRing})
+
+Return the global Tjurina algebra of the affine scheme `X`, if `X` is a hypersurface.
+Return an error otherwise.
```
fingolfin: Does it really "return" an error (and what does that mean)? Or does it raise an error / throw an exception?

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+"""
+function tjurina_algebra(X::AffineScheme{<:Field,<:MPolyQuoRing})
+  ngens(modulus(OO(X))) == 1 || error("not a hypersurface (or unnecessary generators in specified generating set)")
+  return tjurina_algebra(gen(modulus(OO(X)),1))
+end
+
+
+
+@doc raw"""
+    tjurina_algebra(X::HypersurfaceGerm)
```
fingolfin: Redundant with the next line
```suggestion
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+###############################################################################
+
+#####                           Tjurina number                            #####
+
+###############################################################################
+
+@doc raw"""
+    tjurina_number(f::MPolyRingElem)
+
+Return the global Tjurina number of a polynomial `f`. Return -1 if infinite.
```
fingolfin: I really don't like this convention of returning `-1` to indicate infinity. That's gone cause all kinds of bugs down the line.

Alternatives we use elsewhere include:
- return an actual infinite (via `PosInf()`); 
- raise a (dedicated) exception indicating infinite order
- add a separate test function like `has_finite_tjurina_number`

experimental/Schemes/Tjurina.jl
fingolfin: This kind of test is problematic because it will intercept *any* error, not just one that indicates something about the dimension.

In general, it would be preferable to not handle this kind of thing via exceptions anyway.

BTW, why does the method just before it not need this "fix" (even though its docstring also says it can return `-1` to indicate infinity)?

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+
+################################################################################
+
+#####                                 Order                                #####
+
+################################################################################
+
+
+function _order(f::MPolyRingElem)                                                                                 #TODO:  export???
+  f != 0 || return -1
```
fingolfin: Better:
```suggestion
  !is_zero(f) || return -1
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  x = gens(parent(f))
+  n = length(x)
```
fingolfin: No need to allocate `x` if you then don't use it...
```suggestion
  n = ngens(parent(f))
```
or equivalently
```suggestion
  n = nvars(parent(f))
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  degrees = [sum([degree(m, j) for j in 1:n]) for m in monomials(f)]
+  return minimum(degrees)
```
fingolfin: ... Isn't this just computing the minimal total degrees of all terms? If so something like this should be much more efficient
```suggestion
  return minimum(sum(e) for e in AbstractAlgebra.exponent_vectors(f))
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+julia> L,_  = localization(R, complement_of_point_ideal(R, [0, 0]));
+
+julia> is_finitelydetermined(L(x^3-y^2))
+true
+
+julia> is_finitelydetermined(L(x^3-y^2), :right)
+true
+```
+"""
+function is_finitelydetermined(f::MPolyLocRingElem{<:Field, <:Any, <:Any, <:Any, <:MPolyComplementOfKPointIdeal}, equivalence::Symbol = :contact)
```
fingolfin: Should be like this following our conventions
```suggestion
function is_finitely_determined(f::MPolyLocRingElem{<:Field, <:Any, <:Any, <:Any, <:MPolyComplementOfKPointIdeal}, equivalence::Symbol = :contact)
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  end
+end
+
+
+
+
+@doc raw"""
+    is_finitelydetermined(X::HypersurfaceGerm, equivalence::Symbol = :contact)
+
+Return if the hypersurface germ 'X' is finitely determined w.r.t. ':right' or ':contact' equivalence.
```
fingolfin: I would avoid using abbreviations like `w.r.t`, as not everyone knows w.t.m.

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+function is_finitelydetermined(X::HypersurfaceGerm, equivalence::Symbol = :contact)
+  f = defining_ideal(X)[1]
+  return is_finitelydetermined(f, equivalence)
+end
+
+
+
+###############################################################################
+@doc raw"""
+    determinacy_bound(f::MPolyLocRingElem)
```
fingolfin: Redundant with next line (the same in a bunch of other docstrings)
```suggestion
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+      return determinacy_bound(f_new, equivalence)
+    end
+    X = HypersurfaceGerm(quo(parent(f), ideal(parent(f), f))[1])
+    k = milnor_number(X)
+  else  ## equivalence == :contact
+    ord_f != 0 || return 0
+    k = tjurina_number(f)
+  end
+  characteristic(parent(f)) != 0 || return k + 1
+  characteristic(parent(f)) == 0 || return 2*k - ord_f + 2
```
fingolfin: This seems unnecessarily clever to me. I find this much easier to understand (hopefully I got the order right ... if not I think it kinda underlines my point ;-) )
```suggestion
  return characteristic(parent(f)) == 0 ? k + 1 : 2*k - ord_f + 2
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  end
+  G = standard_basis(I, ordering = negdeglex(parent(a)))
+  h = Singular.highcorner(G.gens.S)
+  l = total_degree(R(h))  
+  ## m^(l+1) \subseteq I
+  ## char. 0:                                         
+  characteristic(R) != 0 || (k = l)
+  ## pos. char.: 2*(l-1) - order_as_series(f) + 2
+  characteristic(R) == 0 || (k = 2*l - order_as_series(f))
+  return k
```
fingolfin:
```suggestion
  return characteristic(R) == 0 ? l : 2*l - order_as_series(f)
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  mA_basis = minimal_generating_set(mA)
+  mB_basis = minimal_generating_set(mB)
+  length(mA_basis) == length(mB_basis) || return false
+    ## edim(A) == 1 
+  length(mA_basis) != 1 || return true
+  ## can't do more for infinite dimensional vectorspaces
+  n != -1 || error("infinite dimensional vectorspaces") 
+  ## Save minimalngens of m^k during calculation
+  ngens_m_k = [length(mA_basis)]  
+  while(length(mA_basis) != n-1)
```
fingolfin:
```suggestion
  while length(mA_basis) != n-1
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  length(mA_basis) != 1 || return true
+  ## can't do more for infinite dimensional vectorspaces
+  n != -1 || error("infinite dimensional vectorspaces") 
+  ## Save minimalngens of m^k during calculation
+  ngens_m_k = [length(mA_basis)]  
+  while(length(mA_basis) != n-1) 
+    k += 1
+    gens_mA_k = minimal_generating_set(mA^k)
+    ngens_m_k = vcat(ngens_m_k, length(gens_mA_k))
+    mA_basis = vcat(mA_basis, gens_mA_k)
```
fingolfin: Can't you use `append!` here and above/below?

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+    mB_basis = vcat(mB_basis, minimal_generating_set(mB^k))
+    length(mA_basis) == length(mB_basis) || return false
+  end
+  ## lift bases
+  mA_basis = [lifted_numerator(mA_basis[j]) for j in 1:length(mA_basis)]  
+  mB_basis = [lifted_numerator(mB_basis[j]) for j in 1:length(mB_basis)]
```
fingolfin:
```suggestion
  mA_basis = lifted_numerator.(mA_basis)
  mB_basis = lifted_numerator.(mB_basis)
```

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,770 @@
...
+  I_A = Oscar.shifted_ideal(ideal(L, minimal_generating_set(modulus(A)))) 
+  I_B = ideal(standard_basis(Oscar.shifted_ideal(modulus(B)), ordering = negdeglex(R))) 
+  ## construct homomorphism with parameters
+  phi = elem_type(P)[]  
+  for i in 0:ngens_m_k[1]-1
+    img = P()
+    for j in 1:length(mB_basis)
+      img += t[(i*length(mB_basis)+j)]*iota(mB_basis[j])
+    end
+    phi = vcat(phi, img)
```
fingolfin: Consider using `append!` or `push! (as appropriate)

experimental/Schemes/Tjurina.jl
afkafkafk13: An exception would raise all kinds of complications down the road, I would rather use infinity, as this is the mathematically correct statement.
has_finite_tjurina_number has no added value here, because this is synonymous with is_finitely_determined by the theorem of Mather-Yau and the test for this is available.

experimental/Schemes/Tjurina.jl
afkafkafk13: Is this a typo? it should read ...` ? l :` .... shouldn't it?

fingolfin:
(Just to be clear, overall this PR is fine, the things I pointed out are all minor details and should hopefully to be easy to resolve)

experimental/Schemes/Tjurina.jl
Syz-MS: Removed due to zero dimensional vectorspace case being fixed in #3497

experimental/Schemes/Tjurina.jl
Syz-MS: I totally agree with using PosInf() instead of an exception. But I think the problem lies in vector_space_dimension(A::MPolyQuo) throwing an exception instead of returning infinity and therefore it should be fixed at that level. 
Should I open a feature request or fix that issue in this PR?

Syz-MS:
@fingolfin Thank you for your suggestions. They have been quite helpful.

experimental/Schemes/Tjurina.jl
joschmitt: I wouldn't do it in this pull request.
Tinkering with the output of `vector_space_dimension` would be a breaking change to my understanding and we are not allowed to do these any more since we released 1.0. It would break the code of people who did a similar try-catch work around to catch an infinite dimensional ring if there is no error.
But that's just my two cents.

afkafkafk13: I would prefer that you open a feature request issue for vectore_space_dimension and use positive infinity here in the meantime, to make sure that everything building on top of it does not break in the future, when the feature is changed.

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,751 @@
...
+  return characteristic(parent(f)) == 0 ? k + 1 : 2*k - ord_f + 2
+end
+
+
+
+@doc raw"""
+    determinacy_bound(X::HypersurfaceGerm, equivalence::Symbol = :contact)
+
+Compute some determinacy bound of the hypersurface germ 'X' with respect to ':right' or ':contact' equivalence.
+Return infinity if not finitely determined.
```
afkafkafk13, line 425: Please state explicitly somewhere that PosInf() does not prove the absence of finite determinacy in prime characteristic. In positive characteristic infinity is also returned, if no determinacy bound can be obtained by this approach.

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,751 @@
...
+  return characteristic(R) == 0 ? l : 2*l - order_as_series(f)
+end
+
+
+
+@doc raw"""
+    sharper_determinacy_bound(X::HypersurfaceGerm, equivalence::Symbol = :contact)
+
+Compute some determinacy bound of the hypersurface germ 'X' with respect to ':right' or ':contact' equivalence.
+Return infinity if not finitely determined.
```
afkafkafk13, line 518: dito

afkafkafk13 (commented):
add the two small remarks to the documentation, then this is good to go

afkafkafk13 (approved):

experimental/Schemes/Tjurina.jl
```diff
@@ -0,0 +1,751 @@
...
+export sharper_determinacy_bound
+
+export is_contact_equivalent
+
+## literature:
+##        Greuel, Lossen, Shustin: 'Introduction to Singularities and Deformations'
+
+##    for positive characteristic the following papers
+##        Greuel, Pham: 'Mather-Yau Theorem in Positive Characteristic'
+##        Boubakri, Greuel, Markwig: 'Invariants of Hypersurface Singularities in Positive Characteristic'
```
fingolfin, line 17:
```suggestion
##        Bourbaki, Greuel, Markwig: 'Invariants of Hypersurface Singularities in Positive Characteristic'
```

experimental/Schemes/Tjurina.jl
Syz-MS: In my understanding an upper determinancy bound (strictly less than infinity) can always be computed this way in prime characteristic acording to the paper ''Invariants of Hypersurface Singularities in Positive Characteristic'' by Boubakri, Greuel, Markwig (arXiv:1005.4503). 
The upper bounds provided by Theorem 2.1 respectivly Corollary 2.4 are computated here and as a consequence of Theorem 2.5 these bounds are always finite in the finite determined case. See also Theorem 2.8. Therefore PosInf() is only returned in the non finite determined case.

afkafkafk13:
Concerning the tests: The structure of including a file in experimental has changed. You need to add your file in `experimental/Schemes/Schemes.jl`

experimental/Schemes/Tjurina.jl
afkafkafk13: Yes indeed. I take my previous comment back (I had some other invariant in mind).

PR #3500, opened 2024-03-07
Betti tables over quotient rings

HechtiDerLachs:
This was asked for by some externals from the group of @micjoswig . 

Have a look at the newly added tests and let me know whether this suits your needs.

### DIFF
--- a/experimental/DoubleAndHyperComplexes/src/Morphisms/free_resolutions.jl
+++ b/experimental/DoubleAndHyperComplexes/src/Morphisms/free_resolutions.jl
@@ -136,13 +136,18 @@ augmentation_map(c::SimpleFreeResolution) = c.augmentation_map
 function betti(b::SimpleFreeResolution; project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool = false)
   return betti_table(b; project, reverse_direction)
 end
-function betti_table(C::AbsHyperComplex; project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool=false)
+function betti_table(C::AbsHyperComplex; 
+    lower_bound::Union{Int, Nothing}=(has_lower_bound(C) ? Oscar.lower_bound(C) : nothing),
+    upper_bound::Union{Int, Nothing}=(has_upper_bound(C) ? Oscar.upper_bound(C) : nothing),
+    project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool=false
+  )
+  @show upper_bound
+  @show lower_bound
   @assert dim(C) == 1 "complex must be one-dimensional"
-  @assert has_upper_bound(C) "no upper bound known for this resolution"
+  @assert lower_bound !== nothing && upper_bound !== nothing "explicit bounds must be known"
+  @assert direction(C, 1) == :chain "only implemented for chain complexes"
   generator_count = Dict{Tuple{Int, Any}, Int}()
-  rng = upper_bound(C):-1:lower_bound(C)
-  n = first(rng)
-  for i in 0:upper_bound(C)
+  for i in lower_bound:upper_bound
     @assert is_graded(C[i]) "one of the modules in the graded free resolution is not graded"
     module_degrees = degree.(gens(C[i]))
     for degree in module_degrees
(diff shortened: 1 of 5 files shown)


experimental/DoubleAndHyperComplexes/src/Morphisms/free_resolutions.jl
```diff
@@ -136,13 +136,18 @@ augmentation_map(c::SimpleFreeResolution) = c.augmentation_map
...
   return betti_table(b; project, reverse_direction)
 end
-function betti_table(C::AbsHyperComplex; project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool=false)
+function betti_table(C::AbsHyperComplex; 
+    lower_bound::Union{Int, Nothing}=(has_lower_bound(C) ? Oscar.lower_bound(C) : nothing),
+    upper_bound::Union{Int, Nothing}=(has_upper_bound(C) ? Oscar.upper_bound(C) : nothing),
+    project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool=false
+  )
+  @show upper_bound
+  @show lower_bound
```
lgoettgens:
```suggestion
```
these seem to be an artifact from debugging

experimental/DoubleAndHyperComplexes/test/free_resolutions.jl
```diff
@@ -19,3 +19,21 @@
...
+  F = graded_free_module(A, 1)
+
+  J, inc = sub(F, [x*F[1], y*F[1]])
+
+  M = cokernel(inc)
+
+  res, aug = free_resolution(Oscar.SimpleFreeResolution, M)
+
+  betti_table(res; upper_bound=3)
+  minimal_betti_table(res; upper_bound=3)
```
lgoettgens, line 38: maybe add a "real" test to catch regressions/changes instead of only testing that it doesn't throw any errors.

experimental/DoubleAndHyperComplexes/src/Morphisms/free_resolutions.jl
```diff
@@ -136,13 +136,18 @@ augmentation_map(c::SimpleFreeResolution) = c.augmentation_map
 function betti(b::SimpleFreeResolution; project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool = false)
   return betti_table(b; project, reverse_direction)
 end
-function betti_table(C::AbsHyperComplex; project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool=false)
+function betti_table(C::AbsHyperComplex; 
+    lower_bound::Union{Int, Nothing}=(has_lower_bound(C) ? Oscar.lower_bound(C) : nothing),
+    upper_bound::Union{Int, Nothing}=(has_upper_bound(C) ? Oscar.upper_bound(C) : nothing),
```
lgoettgens, line 141:
```suggestion
    lower_bound::Union{Int, Nothing}=(has_lower_bound(C) ? lower_bound(C) : nothing),
    upper_bound::Union{Int, Nothing}=(has_upper_bound(C) ? upper_bound(C) : nothing),
```
you don't need the qualification here as these functions are exported

lgoettgens (commented):
some random comments. I have no idea of the mathematics involved

experimental/DoubleAndHyperComplexes/src/Morphisms/free_resolutions.jl
HechtiDerLachs: I wasn't 100% sure about the scope here since the keyword argument has the same name.

experimental/DoubleAndHyperComplexes/test/free_resolutions.jl
HechtiDerLachs: I suppose the Berlin crew has some use cases from which we can cook up real tests. This was only to assure it doesn't throw errors and to give them an idea for how to use this stuff. We can change all this later in case people are really happy with this. If not: no need to put in all this extra work at this point.

experimental/DoubleAndHyperComplexes/src/Morphisms/free_resolutions.jl
lgoettgens: True. Good catch

HechtiDerLachs:
Ready for merge. If there are use-cases, we can add them as tests in a new PR.

src/Modules/ModulesGraded.jl
```diff
@@ -2657,12 +2662,15 @@ function _constant_sub_matrix(
...
-      result[i, j] = first(coefficients(c))
+      result[i, j] = _constant_coeff(c)
     end
   end
   return ind_dom, ind_cod, result
 end
 
+_constant_coeff(f::MPolyDecRingElem) = first(coefficients(f))
+_constant_coeff(f::MPolyQuoRingElem) = first(coefficients(lift(f)))
+
```
joschmitt, line 2673: Hmm, I don't really like the name `_constant_coeff` because it doesn't return the coefficient of the constant term in general.
Maybe you could overload `coeff` instead?
```suggestion
      result[i, j] = _coeff(c, 1)
    end
  end
  return ind_dom, ind_cod, result
end

_coeff(f::MPolyDecRingElem, i::Int) = coeff(f, i)
_coeff(f::MPolyQuoRingElem, i::Int) = coeff(lift(f), i)

```
We could also have `coeff(::MPolyQuoRingElem, ::Int)`, but I'm not user I like it because it doesn't make sense mathematically to take the coefficients of a residue class.

HechtiDerLachs: Thanks for having a look. We can call it whatever we like. The situation was the following: I needed to make the code run for graded polynomial rings and quotients of these at the same time. The expression I used for polynomials did not work for quotients. So instead  of rewriting the whole methods for quotient rings, I just introduced this dummy internal function. If there was a coherent way to get the constant term in both settings, I am happy to use that! But otherwise this seems to be the cheapest option. 

Edit: The integer argument does not make sense in the multivariate case.

joschmitt: I understand! I was just wondering about the name `constant_coeff` as it suggests "coefficient of the constant term" to me (which it will not return for e.g. `x + y`). But feel free to ignore the suggestion.

(I think `coeff(..., 1)` should actually be the cheapest because it doesn't need to compare according to a monomial ordering.)

joschmitt (approved):
Looks good, let see what people in Berlin do with it.

jankoboehm (approved):

jankoboehm:
Needed in any case, so let's merge and see if more is neccessary.

PR #3503, opened 2024-03-08
Provide links for some documentation references

### LINKED ISSUE #2483: Change some occurrence of `see section XYZ` to use actual references
There are multiple places in the documentation saying something like "See the section on Gröbner bases" or "... on modules".

Ideally these would be changed into hyperlinks, by using the `[TITLE](@ref)` syntax, e.g. `[Permutation groups](@ref)` produces a link to the section on permutation groups.

Here are some of them (I do not claim this is complete)
```
docs/src/CommutativeAlgebra/GroebnerBases/orderings.md:40:  ordering. Independently of the attached orderings, Gröbner bases can be computed with respect to any monomial ordering. See the section on Gröbner bases.
docs/src/CommutativeAlgebra/affine_algebras.md:17:    $R$-module means to implement it as the quotient of a free $R$-module of rank 1. See the section on modules.
docs/src/CommutativeAlgebra/affine_algebras.md:23:  to the `default_ordering` on the underlying polynomial ring; see the section on *Gröbner/Standard Bases* for default orderings in
docs/src/CommutativeAlgebra/affine_algebras.md:33:  with respect to the default ordering on $R$ (see the section on *Gröbner/Standard Bases* for normal forms). Operations involving
docs/src/CommutativeAlgebra/localizations.md:26:called the *localization  of ``M`` at ``U``*. See the section on modules.
docs/src/CommutativeAlgebra/rings.md:33:order). The other possible choices are `:deglex` and `:degrevlex`. Gröbner bases, however, can be computed with respect to any monomial ordering. See the section on Gröbner bases.
docs/src/NoncommutativeAlgebra/PBWAlgebras/intro.md:31:Working with Gröbner bases requires that we take monomial orderings into account (see the section
src/Rings/groebner.jl:1255:    See the corresponding section of the OSCAR documentation for some details.
```

joschmitt:
Closes #2483.

I learnt about `@id`.
I grepped for 'section', 'see' and 'chapter' to find these references without link, if you feel this does not close #2483, unlink the issue.
I did not manage to do a link to the number field section in `NumberTheory/galois.md` because that is imported from the Hecke documentation.


simonbrandhorst (approved):
Thanks for looking into this.

PR #3439, opened 2024-02-27
Experimental: Partitioned Permutations

BjSchaefer:
This is a new module for working with partitioned permutations which are used in Free Probablility Theory. For more information see the Readme file.

The module uses the module SetPartitions which is currently in the folder experimentals. To compile properly, we needed to rename the project folder to ZPartitionedPermutations so that this folder is below the folder SetPartitions within the folder experimentals.

PR #3505, opened 2024-03-11
Add `cycles` for `PermGroupElem`

joschmitt:
A function `cycles` is requested for `PermGroupElem` in https://github.com/oscar-system/Oscar.jl/pull/3439#discussion_r1515993975 . Here is an easy implementation of this by calling the AbstractAlgebra function `cycledec`. I find this a bit weird since the element must know its cycles on the GAP side; after all it prints as a cycle decomposition. I didn't find out how to access this piece of information directly though.

AbstractAlgebra also has a fancy iterator type for the cycles. I don't see a reason for this (other than nicer printing maybe?).

### DIFF
--- a/src/Groups/perm.jl
+++ b/src/Groups/perm.jl
@@ -642,7 +642,8 @@ end
 @doc raw"""
     cycles(g::PermGroupElem)
 
-Return the cycles of the permutation `g` as a list of integer vectors.
+Return all cycles (including trivial ones) of the permutation `g` as
+a sorted list of integer vectors.
 
 # Examples
 ```jldoctest
--- a/src/Groups/perm.jl
+++ b/src/Groups/perm.jl
@@ -639,6 +639,42 @@ function cycle_structures(G::PermGroup)
   return Set(cycle_structure(x) for x in r)
 end
 
+@doc raw"""
+    cycles(g::PermGroupElem)
+
+Return the cycles of the permutation `g` as a list of integer vectors.
+
+# Examples
+```jldoctest
+julia> g = cperm(1:3, 6:7, 8:10, 11:15)
+(1,2,3)(6,7)(8,9,10)(11,12,13,14,15)
+
+julia> cycles(g)
+6-element Vector{Vector{Int64}}:
+ [1, 2, 3]
+ [4]
+ [5]
+ [6, 7]
+ [8, 9, 10]
+ [11, 12, 13, 14, 15]
+
+julia> g = cperm()
+()
+
+julia> cycles(g)
+1-element Vector{Vector{Int64}}:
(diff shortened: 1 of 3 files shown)


joschmitt:
Please review this so that the people in #3439 can use it.

thofma:
@fingolfin or @ThomasBreuer can you have a look?

ThomasBreuer (approved):
This looks fine. (In particular, the underlying set of points is the range from 1 to the degree of the element.)

Concerning the question whether GAP knows the cycles internally: No, the cycles of a GAP permutation have to be computed in order to show the permutation as a product of cycles, and the cycles are not cached afterwards. Furthermore, there is apparently no fast GAP function to compute the cycles, the method proposed here is (in some experiments) faster than a simpleminded call of `GAP.Globals.Cycles` followed by a conversion to Julia.

src/Groups/perm.jl
```diff
@@ -639,6 +639,41 @@ function cycle_structures(G::PermGroup)
   return Set(cycle_structure(x) for x in r)
 end
 
+@doc raw"""
+    cycles(g::PermGroupElem)
+
+Return the cycles of the permutation `g` as a list of integer vectors.
```
fingolfin: Does it return *all* cycles or only non-trivial ones? Maybe would be a good idea to pick an example with a trivial cycle (i.e. a fixed point) to show case that.

fingolfin (approved):
That's a useful addition, thanks. One minor comment.

src/Groups/perm.jl
joschmitt: It returns all cycles, I adjusted the example. Should one also write something about this in the doc string?

fingolfin:
As to "faster" etc.: I am working on a PR which provides low-level access to GAP permutations which then can be used to e.g. implement a faster version of this. But that should not delay this PR, to the contrary, I only mention it to say we can worry about performance tuning later.

src/Groups/perm.jl
fingolfin: For me it is fine if the example shows it. But I've noticed that since the convention is for hand-written representation to skip trivial cycles, students tend to forget about them. Perhaps this extends to more people and they'd appreciate the hint? 

Of course it might also cause confusion for other people ("why do they emphasize this? Am I missing something")... Dunno
```suggestion
Return all cycles (including trivial ones) of the permutation `g` as a list of integer vectors.
```

Independent of this, one might also wish to add some guarantees to the docstring:
```suggestion
Return all cycles (including trivial ones) of the permutation `g` as
a sorted list of integer vectors.
```

fingolfin:
Thank you, @joschmitt

Issue #2699, opened 2023-08-17
FTheoryTools - Goals for 2024 and beyond


HereAround:
# To be discussed

# Current WIP

- [ ] Make sure there is documentation for every rng kwarg. Make this uniform in the form of a note?
- [ ] `special_flux_family` has optional kwarg to select the employed algorithm. I suppose it makes sense to propagate this choice to any function that calls `special_flux_family`? How about the random seed? This is only used in case the special algorithm is used. So I suppose this should also be propagated...
- [ ] As an example, look at the following:
```julia
@attr FamilyOfG4Fluxes function g4_flux_family(gf::G4Flux; completeness_check::Bool = true)
  nb = breaks_non_abelian_gauge_group(gf)
  gfs = special_flux_family(model(gf), not_breaking = !nb, completeness_check = completeness_check)
  return gfs
end
```

# Important admin tasks

- [ ] Follow-up submission at https://onlinelibrary.wiley.com/journal/15213978.
- [ ] JHEP or https://www.sigsam.org/cca/index.html?

# Important coding tasks

- [ ] Algorithm selection via symbols, not strings.

### Code optimization

- [ ] The data type of the attributes `weierstrass_model` and `global_tate_model` for a given `hypersurface_model`...
        1. changes its type upon calling said attributes. Before the call, the value is likely (?) a string, afterwards likely (?) a Weierstrass/global Tate model. Is this truly desired? Type stable?
        2. In theory, both the 'Weierstrass and global Tate model should exist for a given hypersurface model. However, in practice this is very different from knowing an exact representation/computing such a representation. As such, we also need to carefully gauge the words to be used in the error message, like "Sorry - we are currently too stupid to compute the Weierstrass/global Tate model, and thus cannot provide details. But in theory, it exists."
- [ ] Kwargs need not be used in doc string refs. Can be removed.
- [ ] Datatypes in FTheoryTools https://github.com/oscar-system/Oscar.jl/issues/4941.
- [ ] `Hypersurface_equation_parametrization` is only supported for hypersurface models. It is a missing feature for Weierstrass and global Tate models. Also, when over a concrete base and the genesis is favorable, it should have a meaningful hypersurface equation parametrization. Is this true? If not, fix it.
- [ ] Increase code coverage of tests.
- [ ] Use macro to generate attribute getters. This should reduce code a lot and be easier to maintain.
- [ ] There will be a framework on parallel computing in OSCAR. Try to make use of it.
[...]

PR #3506, opened 2024-03-11
[FTheoryTools] Rewrite _kodaira_type to work with any locus (Part 2)

HereAround:
cc @apturner (cf. https://github.com/oscar-system/Oscar.jl/pull/3424)

### DIFF
--- a/experimental/FTheoryTools/src/auxiliary.jl
+++ b/experimental/FTheoryTools/src/auxiliary.jl
@@ -176,24 +176,45 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
     poly_g = ring_map(g)
     poly_d = ring_map(d)
     locus = ring_map(gens(id)[1])
+    locus_id = ideal([locus])
     
     if f_ord == 0 && g_ord == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(9 * poly_g, [locus], [0]), evaluate(2 * poly_f, [locus], [0]))
+      quotient_gens = gens(groebner_basis(quotient(ideal([9 * poly_g]), ideal([2 * poly_f])) + locus_id))
+      deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      quotient_val = quotient_gens[1]
+
+      monodromy_poly = _psi^2 + quotient_val
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I_$d_ord", "Split I_$d_ord"])
     elseif d_ord == 4 && g_ord == 2 && f_ord >= 2
-      monodromy_poly = _psi^2 - evaluate(divexact(poly_g, locus^2), [locus], [0])
+      quotient_gens = gens(groebner_basis(ideal([div(poly_g, locus^2)]) + locus_id))
+      deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      quotient_val = quotient_gens[1]
+
+      monodromy_poly = _psi^2 - quotient_val
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split IV", "Split IV"])
     elseif d_ord == 6 && f_ord >= 2 && g_ord >= 3
-      monodromy_poly =  _psi^3 + _psi * evaluate(divexact(poly_f, locus^2), [locus], [0]) + evaluate(divexact(poly_g, locus^3), [locus], [0])
+      quotient_gens = gens(groebner_basis(ideal([_psi^3 + _psi * div(poly_f, locus^2) + div(poly_g, locus^3)]) + locus_id))
+      deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      
+      monodromy_poly = quotient_gens[1]
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I^*_0", "Semi-split I^*_0", "Split I^*_0"])
-    elseif f_ord == 2 && g_ord == 3 && d_ord >= 7 && d_ord % 2 == 1
-      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * poly_f, locus^2)^3, [locus], [0]), 4 * evaluate(divexact(9 * poly_g, locus^3), [locus], [0])^3)
-      kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I^*_$(d_ord - 6)", "Split I^*_$(d_ord - 6)"])
-    elseif f_ord == 2 && g_ord == 3 && d_ord >= 8 && d_ord % 2 == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * poly_f, locus^2)^2, [locus], [0]), evaluate(divexact(9 * poly_g, locus^3), [locus], [0])^2)
+    elseif f_ord == 2 && g_ord == 3 && d_ord >= 7
+      quotient_1 = div(poly_d, locus^d_ord)
+      quotient_2 = div(2 * poly_f, locus^2)^(2 + d_ord % 2)
+      quotient_3 = 4 * div(9 * poly_g, locus^3)^(2 + d_ord % 2)
+
+      quotient_4_gens = gens(groebner_basis(quotient(ideal([quotient_1 * quotient_2]) + locus_id, ideal([quotient_3]))))
+      deleteat!(quotient_4_gens, findall(g -> g == locus, quotient_4_gens))
+      quotient_4_val = quotient_4_gens[1]
+
+      monodromy_poly = _psi^2 + quotient_4_val
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I^*_$(d_ord - 6)", "Split I^*_$(d_ord - 6)"])
     elseif d_ord == 8 && g_ord == 4 && f_ord >= 3
-      monodromy_poly = _psi^2 - evaluate(divexact(poly_g, locus^4), [locus], [0])
+      quotient_gens = gens(groebner_basis(ideal([div(poly_g, locus^4)]) + locus_id))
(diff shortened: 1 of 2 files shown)


experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,59 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
     
     if f_ord == 0 && g_ord == 0
-      monodromy_poly = _psi^2 + divexact(evaluate(9 * poly_g, [locus], [0]), evaluate(2 * poly_f, [locus], [0]))
+      g_quotient = divrem(9 * poly_g, locus)[2]
+      f_quotient = divrem(2 * poly_f, locus)[2]
+      quotient_val = div(g_quotient, f_quotient)
+
+      # quotient_gens = gens(groebner_basis(quotient(ideal([9 * poly_g]), ideal([2 * poly_f])) + locus_id))
+      # deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      # quotient_val = quotient_gens[1]
```
HereAround, line 188: Is this still needed? If not, maybe just remove it?

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,59 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
+
+      monodromy_poly = _psi^2 + quotient_val
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I_$d_ord", "Split I_$d_ord"])
     elseif d_ord == 4 && g_ord == 2 && f_ord >= 2
-      monodromy_poly = _psi^2 - evaluate(divexact(poly_g, locus^2), [locus], [0])
+      g_quotient = divrem(div(poly_g, locus^2), locus)[2]
+
+      # quotient_gens = gens(groebner_basis(ideal([div(poly_g, locus^2)]) + locus_id))
+      # deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      # quotient_val = quotient_gens[1]
```
HereAround, line 197: Likewise.

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,59 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
+
+      monodromy_poly = _psi^2 - g_quotient
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split IV", "Split IV"])
     elseif d_ord == 6 && f_ord >= 2 && g_ord >= 3
-      monodromy_poly =  _psi^3 + _psi * evaluate(divexact(poly_f, locus^2), [locus], [0]) + evaluate(divexact(poly_g, locus^3), [locus], [0])
+      f_quotient = divrem(div(poly_f, locus^2), locus)[2]
+      g_quotient = divrem(div(poly_g, locus^3), locus)[2]
+
+      # quotient_gens = gens(groebner_basis(ideal([_psi^3 + _psi * div(poly_f, locus^2) + div(poly_g, locus^3)]) + locus_id))
+      # deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
```
HereAround, line 206: As above.

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,59 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
+      d_quotient = div(poly_d, locus^d_ord)
+      f_quotient = div(2 * poly_f, locus^2)
+      g_quotient = div(9 * poly_g, locus^3)
+      num_quotient = divrem(d_quotient * f_quotient^(2 + d_ord % 2), locus)[2]
+      den_quotient = divrem(4 * g_quotient^(2 + d_ord % 2), locus)[2]
+      quotient_val = div(num_quotient, den_quotient)
+
+      # quotient_4_gens = gens(groebner_basis(quotient(ideal([quotient_1 * quotient_2]) + locus_id, ideal([quotient_3]))))
+      # deleteat!(quotient_4_gens, findall(g -> g == locus, quotient_4_gens))
+      # quotient_4_val = quotient_4_gens[1]
```
HereAround, line 220: Same here too.

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,59 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
+
+      monodromy_poly = _psi^2 + quotient_val
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I^*_$(d_ord - 6)", "Split I^*_$(d_ord - 6)"])
     elseif d_ord == 8 && g_ord == 4 && f_ord >= 3
-      monodromy_poly = _psi^2 - evaluate(divexact(poly_g, locus^4), [locus], [0])
+      g_quotient = divrem(div(poly_g, locus^4), locus)[2]
+
+      # quotient_gens = gens(groebner_basis(ideal([div(poly_g, locus^4)]) + locus_id))
+      # deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      # quotient_val = quotient_gens[1]
```
HereAround, line 229: And here.

experimental/FTheoryTools/test/tate_models.jl
```diff
@@ -233,7 +233,7 @@ end
   irr_i5_s = irrelevant_ideal(tas);
   sri_i5_s = stanley_reisner_ideal(tas);
   lin_i5_s = ideal_of_linear_relations(tas);
-  id_fin,  = _blowup_global_sequence(id_i5_s, [[7, 8, 6], [2, 3, 1], [3, 4], [2, 4]], irr_i5_s, sri_i5_s, lin_i5_s)
+  id_fin,  = _blowup_global_sequence(id_i5_s, [[8, 9, 6], [2, 3, 1], [3, 4], [2, 4]], irr_i5_s, sri_i5_s, lin_i5_s)
```
HereAround, line 236: Why is this change needed? I was under the impression that this change would only affect the behavior "under the hood", but this change indicates differently. Or was there a mistake in the tests to begin with?

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -176,24 +176,59 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
+      f_quotient = divrem(div(poly_f, locus^2), locus)[2]
+      g_quotient = divrem(div(poly_g, locus^3), locus)[2]
+
+      # quotient_gens = gens(groebner_basis(ideal([_psi^3 + _psi * div(poly_f, locus^2) + div(poly_g, locus^3)]) + locus_id))
+      # deleteat!(quotient_gens, findall(g -> g == locus, quotient_gens))
+      
+      monodromy_poly = _psi^3 + _psi * f_quotient + g_quotient
       kod_type = _string_from_factor_count(monodromy_poly, ["Non-split I^*_0", "Semi-split I^*_0", "Split I^*_0"])
-    elseif f_ord == 2 && g_ord == 3 && d_ord >= 7 && d_ord % 2 == 1
-      monodromy_poly = _psi^2 + divexact(evaluate(divexact(poly_d, locus^d_ord) * divexact(2 * poly_f, locus^2)^3, [locus], [0]), 4 * evaluate(divexact(9 * poly_g, locus^3), [locus], [0])^3)
```
HereAround, line 190: With an eye towards migrating eventually into `source` and satisfying all `OSCAR` code standards - this line is pretty hard to read because it is so long. And having several combined calls to `divexact` and `evaluate` does not exactly make the reading any easier (at least for me).

Suggestions:
Since the computation for the monodromy polynomial happens multiple times, maybe we could add an auxiliary function that completes this task? And this auxiliary function has strictly more than one line of code, maybe also a few comments on what exactly it does?

HereAround:
Our to-do list (https://github.com/oscar-system/Oscar.jl/issues/2699) just reminded me of the following:

"Note: Quite a few places in the code use this assumption, and should be modified once the limitation is removed."

This applies to the literature models: When we build a model (presumably over a concrete base, but also needs to be checked about the arbitrary base) by specifying a line bundle as "gauge divisor", then it seems natural that this method picks the generic section of said line bundle and then declares the vanishing locus of this generic section as the gauge divisor. Currently, this does not happen.

This change should be easy and - as I see it - should happen in this PR.

I am not sure if there are other places that make a similar assumption - we should check.

experimental/FTheoryTools/src/auxiliary.jl
apturner: Sorry, I intended to remove all of these commented blocks that were in place during testing, but apparently neglected to, I will remove them

experimental/FTheoryTools/test/tate_models.jl
apturner: This change is needed because an additional coordinate was added to the base space for the tests, in order to make the gauge locus more complicated

experimental/FTheoryTools/src/auxiliary.jl
apturner: The line you're commenting on here is removed in this PR, and I think the current version of the function doesn't have any excessively long lines in it. The reason I didn't add a helper function is because the monodromy polynomial is different for every singularity type. The bulk of the `_kodaira_type` function is this case-by-case computation of the monodromy polynomial anyway.

I do, however, agree that more comments would be helpful, I will add some in!

HereAround:
Ok. Back to https://github.com/oscar-system/Oscar.jl/pull/3424... sorry for the commotion.

experimental/FTheoryTools/src/auxiliary.jl
HereAround: Pardon. My bad.

PR #1538, opened 2022-09-13
More compact printing

joschmitt:
I'm supposed to give a short talk in the "OSCAR case studies" at the SFB meeting next week and used this opportunity to clean up some of the printing methods.
I'm aware that all the doc-tests need to be changed now, but before I do this, I would like to agree on what we want to have.
I changed the `show` methods for the types `InvRing`, `MPolyQuo`, `MPolyRing_dec` and `MPolyIdeal`. For the latter one I removed this `expressify` functionality - mostly because I could not get what I want with it. Is there documentation for this somewhere?
I'm of course happy to change things to conform with other aesthetic requirements.
Here are some impressions of the differences
```
# master
julia> K, a = CyclotomicField(3);

julia> r = matrix(K, [a 0 0 0; 0 a^-1 0 0; 0 0 a^-1 0; 0 0 0 a]);

julia> s = matrix(K, [ 0 1 0 0; 1 0 0 0; 0 0 0 1; 0 0 1 0 ]);

julia> G = matrix_group(r, s);

julia> RG = invariant_ring(G)
Invariant ring of
Matrix group of degree 4 over K
with generators
AbstractAlgebra.Generic.MatSpaceElem{nf_elem}[[z_3 0 0 0; 0 -z_3-1 0 0; 0 0 -z_3-1 0; 0 0 0 z_3], [0 1 0 0; 1 0 0 0; 0 0 0 1; 0 0 1 0]]

julia> polynomial_ring(RG)
Multivariate Polynomial Ring in x[1], x[2], x[3], x[4] over K graded by 
  x[1] -> [1]
  x[2] -> [1]
  x[3] -> [1]
  x[4] -> [1]

julia> affine_algebra(RG)
(Quotient of Multivariate Polynomial Ring in 7 variables y[1], y[2], y[3], y[4], ..., y[7] over K graded by 
  y[1] -> [2]
  y[2] -> [2]
  y[3] -> [2]
  y[4] -> [3]
  y[5] -> [3]
  y[6] -> [3]
[...]

PR #3507, opened 2024-03-11
Unify printing of all `MPoly*Ideal` types

joschmitt:
Adjust the printing of multivariate ideals in (localized) quotient rings to what was introduced in #3250.
I overlooked these special types back then, sorry.


src/Rings/MPolyQuo.jl
```diff
@@ -266,13 +266,11 @@ HasGroebnerAlgorithmTrait(::Type{Nemo.zzModRing}) = HasSingularGroebnerAlgorithm
     return MPolyQuoIdeal(Ox, ideal(R, V))
   end
 end
-@enable_all_show_via_expressify MPolyQuoIdeal
 
 function AbstractAlgebra.expressify(a::MPolyQuoIdeal; context = nothing)
   return Expr(:call, :ideal, [expressify(g, context = context) for g in gens(a)]...)
 end
```
lgoettgens, line 272:
```suggestion
```
Line 269 calling this gets removed, so this is no longer needed

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1780,28 +1780,58 @@ function ideal(
...
+Ideal generated by
+  x
+  y
+
+julia> (I,)
+(Ideal (x, y),)
+
+```
+"""
+function Base.show(io::IO, ::MIME"text/plain", I::MPolyAnyIdeal)
```
lgoettgens, line 1798: I would look for this "generic mpoly ideal" function in `mpoly-ideals.jl` rather than `mpolyquo-localizations.jl`. Any reason why you chose this file?

src/Rings/MPolyQuo.jl
joschmitt: I was once told (https://github.com/oscar-system/Oscar.jl/pull/1538#discussion_r969748418) that one shouldn't remove the `expressify` function even if one removes the `enable...` macro. I do know what this does or why we have it to be honest. (I vaguely recall it is used in Jupyter notebooks.)

src/Rings/mpolyquo-localizations.jl
joschmitt: Inclusion order. There is no centralized `types.jl` (or similar) in this directory and every file contains some `struct`s. This is the first file in which all ideal types are available.

simonbrandhorst (approved):

joschmitt:
I rebased and resolved conflicts. Merge?

PR #3509, opened 2024-03-12
changed the function flag_pluecker_ideal

Sequenzer:
Added the possibility to input a Ring as an argument. As well as an optional argument minimal which returns a reduced Groebner basis which should be a minimal generating set based on theorem 14.6 in the book "combinatorial commutative algebra" by Miller and Sturmfels.


src/Rings/mpoly-ideals.jl
```diff
@@ -2072,8 +2085,36 @@ function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
   genz = [sum(t) for t in T]
   filter!(!iszero, genz)
   gg = unique(x ->  divexact(x, canonical_unit(x)), genz)
+  I  = ideal(gg)
 
-  return ideal(gg)
+  !minimal && return I 
+  # Computes the groebner_basis of the ideal
+  converted_generators = collect(groebner_basis(I; ordering = degrevlex(base_ring(I))))
+  o = degrevlex(R)
```
antonydellavecchia: you should define the ordering `o` before you defined `converted_generators`

test/Rings/mpoly.jl
```diff
@@ -546,4 +546,8 @@ end
   @test [f1] == gens(I)
   I2 = flag_pluecker_ideal(GF(3),[1,2,3], 4);
   @test length(gens(I2)) == 10
+
+  R, _ = polynomial_ring(QQ, 6)
+  I = flag_pluecker_ideal(R,dimension_vector, ambient_dimension, minimal=false)
```
antonydellavecchia, line 551: is there a very small example where the GB computation is quick? since you'RE not actually testing the minimal flag here

src/Rings/mpoly-ideals.jl
```diff
@@ -2072,8 +2085,36 @@ function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
...
+  return ideal(IdealGens(R, converted_generators, o;
+                   keep_ordering=true,
+                   isReduced=true,
+                   isGB=true))
 end
 
-flag_pluecker_ideal(dimensions::Vector{Int}, n::Int) = flag_pluecker_ideal(QQ,dimensions,n)
+flag_pluecker_ideal(dimensions::Vector{Int}, n::Int; minimal::Bool=true) = flag_pluecker_ideal(QQ,dimensions,n, minimal=minimal)
+
+function flag_pluecker_ideal(ring::MPolyRing, dimensions::Vector{Int}, n::Int; minimal::Bool=true)
```
antonydellavecchia: since you are expecting the coefficients to belong to a field you should add that to the type signature
```suggestion
function flag_pluecker_ideal(ring::MPolyRing{<: FieldElem}, dimensions::Vector{Int}, n::Int; minimal::Bool=true)
```

test/Rings/mpoly.jl
antonydellavecchia: ah sorry, I see your testing it on the line above.

src/Rings/mpoly-ideals.jl
Sequenzer: It's defined on line 2073 I believe.

src/Rings/mpoly-ideals.jl
```diff
@@ -2072,8 +2085,36 @@ function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
...
+  return ideal(IdealGens(R, converted_generators, o;
+                   keep_ordering=true,
+                   isReduced=true,
+                   isGB=true))
 end
 
-flag_pluecker_ideal(dimensions::Vector{Int}, n::Int) = flag_pluecker_ideal(QQ,dimensions,n)
+flag_pluecker_ideal(dimensions::Vector{Int}, n::Int; minimal::Bool=true) = flag_pluecker_ideal(QQ,dimensions,n, minimal=minimal)
+
+function flag_pluecker_ideal(ring::MPolyRing{<: FieldElem}, dimensions::Vector{Int}, n::Int; minimal::Bool=true)
```
lgoettgens: I think you need to check that the poly ring has the right number of variables for everything below to work

test/Rings/mpoly.jl
```diff
@@ -546,4 +546,8 @@ end
...
+  @test dim(R) == 6
```
lgoettgens: This doesn't test any properties of the ideal, but only something that is already known by construction of the poly ring

Sequenzer: fixed it. Thank you.

src/Rings/mpoly-ideals.jl
```diff
@@ -2072,8 +2085,40 @@ function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
...
+
+  L = reduce(vcat, [subsets(n, d) for d in dimensions])
+  @req length(L) == ngens(ring) "The ring does not have the correct amount of variables"
```
```suggestion
  L = sum(binomial(n, d) for d in dimensions)
  @req L == ngens(ring) "The ring does not have the correct amount of variables"
```
if you don't need the actual subsets, it is way more efficient to just count them without construction

src/Rings/mpoly-ideals.jl
```diff
@@ -2072,8 +2085,40 @@ function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
...
+  @req L == ngens(ring) "The ring does not have the correct amount of variables"
+  
+  base_field = base_ring(ring)  
+  I = flag_pluecker_ideal(base_field, dimensions, n; minimal = false) 
+  
+  coeffmap = c -> begin
+    @assert is_integral(c)
+    return base_field(numerator(c))
+  end
+  h = hom(base_ring(I), ring, coeffmap, gens(ring))
```
joschmitt: Do you need this `coeffmap`? `base_ring(I)` and `ring` have the same coefficient ring, so this should just work.
```suggestion
  h = hom(base_ring(I), ring, gens(ring))
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2009,48 +2009,61 @@ function Base.show(io::IO, I::MPolyIdeal)
...
 =$`dimensions` denotes the rank, with $d_j\leq n-1$.  That is, the vanishing
 set of this ideal corresponds to the space of $k$-step flags of linear
 subspaces $V_1\subset\dots\subset V_k$ in $\mathbb{F}^n$, where
 $\text{dim}(V_j) = d_{j}$.  You can obtain the generators for the
-\emph{complete flag variety}$ of $\mathbb{F}^{n}$ by taking `dimensions`
+$\emph{complete flag variety}$ of $\mathbb{F}^{n}$ by taking `dimensions`
 $=(1,\dots,n-1)$ and `n`$=n$.  We remark that evaluating for `F = QQ` yields
 the same set of generators as any field of characteristic $0$.
+Evaluating this function with the parameter `minimal = true` returns the reduced Gröbner basis for
+the flag Plücker ideal with respect to the reverse lexographical order. The first parameter can
```
joschmitt:
```suggestion
the flag Plücker ideal with respect to the degree reverse lexographical order. The first parameter can
```
At least that's what happens in the code.

gfourier:
Are you implicitly assuming that the user input satisfies d_i \leq d_{i+1} ? Because if d_i > d_{i+1} at some position, you don't have to start any computation.

Returns generators of the defining ideal for the flag variety
$\text{Fl}(\mathbb{F}, (d_1,\dots,d_k), n)$, where $(d_1,\dots,d_k)
=$`dimensions` denotes the rank, with $d_j\leq n-1$.

src/Rings/mpoly-ideals.jl
```diff
@@ -2014,20 +2014,20 @@ _pluecker_sgn(a::Vector{Int}, b::Vector{Int}, t::Int)::Int =
...
 the same set of generators as any field of characteristic $0$.
+
+The first parameter can either be $\mathbb{F}$, or a polynomial ring over $\mathbb{F}$, with $\Sum^{k}_{j=1}{n\choose d_{j}$ variables.  
+The parameter `dimensions` needs to be a vector of distinct increasing entries.
 Evaluating this function with the parameter `minimal = true` returns the reduced Gröbner basis for
-the flag Plücker ideal with respect to the reverse lexographical order. The first parameter can
-either be $\mathbb{F}$ or a polynomial ring R over $\mathbb{F}$. If the input is $\mathbb{F}$, the
-function returns the ideal in a polynomial ring indexed by elements of $2^{[n]}$. 
-Otherwise, the function returns the ideal in R.
+the flag Plücker ideal with respect to the degree reverse lexicographical order.
```
antonydellavecchia: can you add a reference to where the algorithm can be found?

src/Rings/mpoly-ideals.jl
```diff
@@ -2009,57 +2009,85 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+end
+
+flag_pluecker_ideal(dimensions::Vector{Int}, n::Int; minimal::Bool=true) = flag_pluecker_ideal(QQ,dimensions,n, minimal=minimal)
+
+function flag_pluecker_ideal(ring::MPolyRing{<: FieldElem}, dimensions::Vector{Int}, n::Int; minimal::Bool=true)
   dimensions = unique(dimensions)
+   
+  @req maximum(dimensions) <= n-1 "The dimensions must be at most n-1"
+  @req sum(binomial(n, d) for d in dimensions) == ngens(ring) "The ring doesn't have the right number of variables"
+  @req dimensions == sort(dimensions) "The dimensions must be increasing"
```
```suggestion
  @req issorted(dimensions) "The dimensions must be increasing"
```
Checking is O(n)  while actually sorting O(n*log(n))

src/Rings/mpoly-ideals.jl
```diff
@@ -2009,57 +2009,85 @@ function Base.show(io::IO, I::MPolyIdeal)
...
+
+julia> flag_pluecker_ideal(R, [1,3], 4; minimal=false)
+Ideal generated by
+  -x1*x6 + x2*x5 - x3*x7 + x4*x8
 ```
 """
-function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int)
+function flag_pluecker_ideal(F::Field, dimensions::Vector{Int}, n::Int; minimal::Bool=true)
+  L = sum(binomial(n, d) for d in dimensions)
+  ring, _ = polynomial_ring(F, L; cached=false)
```
lgoettgens: I think you now want to change this back to a reduce-vcat-subsets thing, as you want to use the actual subsets for the variable naming in the poly ring

antonydellavecchia (approved):

PR #3456, opened 2024-02-28
Add (truncated) basis computation of the coordinate ring of a fixed Kodaira embedding

lgoettgens:
Some follow-up features to the book chapter by G. Fourier and X. Fang.
I paid attention to not changing any printing of existing functionality, to not creating any issues with the book.

This PR is NOT needed for the book.

@gfourier Please have a look at the new docstrings and the tests. The rest is unchanged compared to our discussion earlier today.

PR #3511, opened 2024-03-13
Adapt `basis_coordinate_ring_kodaira`

lgoettgens:
Some follow-up to https://github.com/oscar-system/Oscar.jl/pull/3456.
While writing something about this function, we became aware that we should return the list of monomials here instead of just the number.

The first commit just renames some local variables and can thus be excluded from review.


gfourier (approved):

gfourier (approved):

gfourier (approved):

lgoettgens:
Can we get this merged?

fingolfin (approved):

PR #3514, opened 2024-03-14
Lazy toric ideals

HechtiDerLachs:
@HereAround : 

This is how I want to make toric ideals lazy. Have a look whether you approve this. If you like it, I would replace the old constructor by `Oscar.ToricIdealSheafFromCoxRingIdeal`. 

Timings are the same, vaguely in favor of the new implementation. 

Edit: But note that the construction of an `Oscar.ToricIdealSheafFromCoxRingIdeal` is instant.


HechtiDerLachs:
@afkafkafk13 : We should talk about making strict and controlled transforms lazy in a similar way. The current implementation is due to you and it looks like there might be more things to be taken into account.

HereAround:
Thank you @HechtiDerLachs. I definitely like those changes. Will leave a few nitpicks now...

experimental/Schemes/Auxiliary.jl
```diff
@@ -85,6 +85,7 @@ function pullback(f::AbsCoveredSchemeMorphism, II::AbsIdealSheaf)
   X = domain(f)
   Y = codomain(f)
   scheme(II) === Y || error("ideal sheaf is not defined on the codomain of the function")
+  return PullbackIdealSheaf(f, II)
   phi = covering_morphism(f)
   ID = IdDict{AbsAffineScheme, Ideal}()
   for U in patches(domain(phi))
```
HereAround: Unless I am mistaken, the lines below the newly added line (return `PullbackIdealSheaf(f, II)`) will not be executed. Can those line be removed?

HereAround: Minor nitpick: Support for `pullback_ideal_sheaf` existing?

experimental/Schemes/ToricIdealSheaves/constructors.jl
```diff
@@ -105,7 +105,7 @@ function IdealSheaf(X::NormalToricVariety, I::MPolyIdeal)
     end
   end
 
-  return IdealSheaf(X, ideal_dict, check=true) # TODO: Set the check to false eventually.
+  return IdealSheaf(X, ideal_dict, check=false) # TODO: Set the check to false eventually.
```
HereAround: Since you set the check to false, maybe remove the TODO comment?

HereAround: You mentioned this yesterday on the other PR that we worked on yesterday - are there other checks that we want to set to `false`? If so, maybe include those changes here?

experimental/Schemes/Types.jl
```diff
@@ -422,4 +422,26 @@ end
...
+    @assert cox_ring(X) === base_ring(I) "incompatible rings"
+    Ipre = PreSheafOnScheme(X,
+                      OpenType=AbsAffineScheme, OutputType=Ideal,
+                      RestrictionType=Map,
+                      is_open_func=_is_open_func_for_schemes_without_affine_scheme_open_subscheme(X)
+                     )
+    I = new{typeof(X), AbsAffineScheme, Ideal, Map}(X, I, Ipre)
+    return I
+  end
+end
```
HereAround, line 446: Nitpick (ignore if you don't like it): Implement also the function `toric_ideal_sheaf_from_cox_ring_ideal(X::NormalToricVariety, I::MPolyIdeal)`, as we eventually hope to have user-facing `snake_case` constructors for all types (and the `CamelCase` constructors are only used internally/by developers).

test/AlgebraicGeometry/ToricVarieties/toric_schemes.jl
```diff
@@ -167,6 +167,8 @@ end
   x1, x2, x3 = gens(cox_ring(p231)) 
   my_ideal = ideal([x1*x2]) 
   II = ideal_sheaf(p231, my_ideal) 
+  JJ = Oscar.ToricIdealSheafFromCoxRingIdeal(p231, my_ideal)
+  @test II == JJ
```
HereAround, line 171: Do we want to export this function (or its `snake_case` equivalent, see above)? If so, the `Oscar.` can be removed.

HereAround: (This also applies to the 2nd test that you are adding.)

HereAround (approved):
Approved modulo the above nitpicks. Thank you @HechtiDerLachs!

HereAround:
Actually, I just noticed that some of the tests are failing with `type PullbackIdealSheaf has no field I`. I trust you know how to fix this @HechtiDerLachs .

HechtiDerLachs:
I don't plan on exporting `Oscar.ToricIdealSheafFromCoxRingIdeal` or introduce a snake_case constructor. 

The user facing functionality should be restricted to `ideal_sheaf`. What you get is some concrete instance of `AbsIdealSheaf` which implements the interface specified for this abstract type. That's all the user should need to know and hence also all that we export.

PR #3516, opened 2024-03-14
more functionality for dict serialization

antonydellavecchia:
Adds more functionality for serializing dicts. 

Specific use case for when the values are the same type and all  share the same parameters

### DIFF
--- a/test/Serialization/containers.jl
+++ b/test/Serialization/containers.jl
@@ -131,6 +131,16 @@
       end
     end
 
+    @testset "(de)serialization Dict{Symbol, T}" begin
+      Qx, x = QQ[:x]
+      for (T, values) in ((Int, [1, 2]), (PolyRingElem, [x^2, x - 1]))
+        original = Dict{Symbol, T}(:a => values[1], :b => values[2])
+        test_save_load_roundtrip(path, original) do loaded
+          @test original == loaded
+        end
+      end
+    end
+
     @testset "Testing (de)serialization of Set" begin
       original = Set([Set([1, 2])])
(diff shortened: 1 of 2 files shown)


test/Serialization/containers.jl
```diff
@@ -139,6 +139,11 @@
           @test original == loaded
         end
       end
+
+      original = Dict{Symbol, Int}()
+      test_save_load_roundtrip(path, original) do loaded
+        @test original == loaded
```
lgoettgens, line 145: If you want to test that loading correctly infers the type parameters, this test is not semantically enough since 
```julia
julia> Dict{Symbol,Int}() == Dict()
true
```

benlorenz: `test_save_load_roundtrip` will always do a type check:
https://github.com/oscar-system/Oscar.jl/blob/49cf335f10a24d78fd333e0308c33a4eae35505e/test/Serialization/setup_tests.jl#L16

Without the fixes in this PR I get:
```julia
julia> test_save_load_roundtrip(path, original) do loaded
         @test original == loaded
         end
Test Failed at /home/lorenz/software/polymake/julia/Oscar.jl/test/Serialization/setup_tests.jl:14
  Expression: loaded isa T
   Evaluated: Dict{Symbol, Any}() isa Dict{Symbol, Int64}
```

benlorenz (approved):

test/Serialization/containers.jl
lgoettgens: Great to know. Thanks!

PR #3517, opened 2024-03-14
load the GAP packages sotgrps and sglppow (extensions of the library of small groups)

ThomasBreuer:
This was intended to belong to pull request #3478, which got merged before commit 9cec252 "arrived".

As soon as the GAP fix from gap-system/gap/pull/5666 becomes available, loading the two GAP packages will not cause strange error messages when leaving Julia.

### DIFF
--- a/src/Oscar.jl
+++ b/src/Oscar.jl
@@ -107,6 +107,8 @@ function __init__()
   # We want some GAP packages. (It is no error if they cannot be loaded.)
   for pkg in [
      "ferret",   # backtrack in permutation groups
+     "sotgrps",  # extend the small groups library
+     "sglppow",  # extend the small groups library
      ]
     GAP.Packages.load(pkg)
   end


ThomasBreuer:
O.k., we have again problems on the GAP side:

The `LiePRing` package is a suggested package of `sglppow`. Fine.
The `Singular` package is a needed package of `LiePRing`. This would also be fine if `Singular` would have a valid availability test in its `PackageInfo.g`.

However, currently `Singular` just claims that it is available, thus `LiePRing` is regarded as available, and GAP runs into an error when it tries to load it if there is no Singular executable.

I think that the problem must be solved in the Singular package, by turning parts of `CheckSingularExecutableAndTempDir` (in `gap/singular.g`) into the `TestPackageAvailability` function.

(I will provide a pull request for the Singular package.)

ThomasBreuer:
Thinking more about the situation, I decided that the right place for adding a `TestAvailability` function is the LiePRing package, see gap-packages/liepring/pull/27.

fingolfin:
I've added a comment to https://github.com/gap-packages/liepring/pull/27 explaining that I think it would be better for us to patch the LiePRing package to call into OSCAR instead of using the GAP package `singular`.

In addition, we could proactively set `sing_exec` to point to the Singular executable shipped with OSCAR (or perhaps rather a wrapper script for it made using https://github.com/oscar-system/BinaryWrappers.jl  e.g. inside the Oscar sources, one could write
```
import Singular.Singular_jll
const singular_path = Singular.Setup.BinaryWrappers.@generate_wrappers(Singular_jll)
GAP.Globals.sing_exec = GAP.Obj(joinpath(singular_path, "Singular"))
```
(As an aside, I just submitted https://github.com/oscar-system/BinaryWrappers.jl/pull/5 -- with that one could also do
```
const singular_path = Singular.Setup.BinaryWrappers.@generate_wrappers(Singular.Singular_jll)
GAP.Globals.sing_exec = GAP.Obj(joinpath(singular_path, "Singular"))
```
but either version is fine)

ThomasBreuer:
@fingolfin Once we have patches for LiePRing in Oscar such that `MyFactors` and `CallGroebner` get replaced after loading LiePRing, what would be a good solution to express that GAP's Singular package should *not* be regarded as a suggested package if we are in the context of Oscar? Shall Oscar try to modify the dependencies of LiePRing, or do we need something like conditional dependencies for GAP packages?

fingolfin (approved):

PR #3518, opened 2024-03-14
Manually store gbs for mirrors of modules over quotient rings

HechtiDerLachs:
This should speed up some computations from #3500.

@jankoboehm : This is what we briefly discussed today. Maybe you can have a look whether the manual setting of the Groebner basis is fine?


HechtiDerLachs:
@jankoboehm : Do you want this in the end?

fingolfin:
I'll merge this now, if @jankoboehm is unhappy with it we can still revert it.

PR #3519, opened 2024-03-14
[FTheoryTools] Unify blowup of singularities among toric varieties and schemes

HereAround:
cc @HechtiDerLachs


src/Rings/MPolyMap/MPolyQuo.jl
```diff
@@ -110,6 +110,15 @@ function _evaluate_plain(F::MPolyAnyMap{<: MPolyQuoRing}, u)
...
 end
 
+function _evaluate_plain(F::MPolyAnyMap{<:MPolyQuoRing, <:MPolyQuoRing}, u)
+  # This workaround deals with the fact that arithmetic in quotient rings is TERRIBLY slow.
+  # All the simplify calls make it unusable in this case, probably due to the fact that 
+  # setting `is_reduced` flags does not pay off in such iterative procedures. 
+  A = codomain(F)
+  v = evaluate(lift(u), lift.(_images(F)))
+  return A(v)
+end
```
HechtiDerLachs: @thofma : Would this change be OK with you? We have an example which goes from ~14 seconds down to ~1.3 seconds only because of this.

thofma: Makes sense

HechtiDerLachs: @HereAround : Your blowups should run 10 times as fast now.

HereAround: Wonderful! Thank you!

HechtiDerLachs:
I improvised a fix for the inclusion order similar to #3480 . Now the actual test failures are exposed. 

@HereAround : I hope this was useful and you can take it from here?

HereAround:
Thank you @HechtiDerLachs!

At the top of my head, I am not sure what caused the error `no method matching keys(::Base.Generator{UnitRange{Int64}, Oscar.var"#7896#7897"{ZZMatrix}})`. It strikes me as odd, but I will of course debug this and shall hope to eventually fix this.

src/Rings/MPolyMap/MPolyQuo.jl
HereAround: I am cleaning up this PR (hence, holding back a fair amount of changes locally right now, so that I can check each commit individually for mistakes/errors). My finding is that this change in the quotient ring evaluation leads to changed printing of quotient rings and thus failures in the doctests.

@HechtiDerLachs and @thofma, can you take a look and let me know what to do about them?

thofma: Seems that the images of the generators are not reduced anymore. But I am not sure how this is effected by this change here. Would be good if @HechtiDerLachs could investigate.

HereAround: Shame on me - Matthias is on well-deserved vacation this week and the next(and so should not have pinged him).

HereAround: @thofma Maybe we can think of a fix? Otherwise I will move this change to a separate PR, so that I can finalise the changes here (even though they will be slower without the suggested quotient gymnastics) by the end of this week.

Thoughts?

fingolfin:
What is the problem with the printing changes? Having a quick look, it seems that e.g.
```
│    Covering morphism
...
│    given by the pullback function
│      1a -> 3b
│        (x//z) -> 0
```
is changed to
```
│    Covering morphism
...
│    given by the pullback function
│      1a -> 3b
│        (x//z)
```
So the ` -> 0` is dropped? Maybe that's undesirable?

fingolfin:
According to @afkafkafk13 this printing change is not desirable, so perhaps this part of the printing can be restored. Are there other issues? if so perhaps list them explicitly so we can discuss them as needed?

src/Rings/MPolyMap/MPolyQuo.jl
thofma: I don't have too much time right now, but maybe `return simplify(A(v))` does the trick? Or whatever is the function to simplify in the quotient.

HereAround: Ok. Let me try that.

HereAround:
> According to @afkafkafk13 this printing change is not desirable, so perhaps this part of the printing can be restored. Are there other issues? if so perhaps list them explicitly so we can discuss them as needed?

Yes, I indeed feel so too.

This comes from an improvement that @HechtiDerLachs suggested. And as mentioned above by @thofma , it seems that the images of the generators are no longer reduced, and also made a suggestion for an attempt to fix it. I just pushed such a change - let us now wait to see if this reverts the printing of quotient rings to its original form.

... And this fix works. Great!

src/Rings/MPolyMap/MPolyQuo.jl
HereAround: Great. This works. Thank you!

src/Rings/MPolyMap/MPolyQuo.jl
```diff
@@ -110,6 +110,15 @@ function _evaluate_plain(F::MPolyAnyMap{<: MPolyQuoRing}, u)
...
 
+function _evaluate_plain(F::MPolyAnyMap{<:MPolyQuoRing, <:MPolyQuoRing}, u)
+  # This workaround deals with the fact that arithmetic in quotient rings is TERRIBLY slow.
+  # All the simplify calls make it unusable in this case, probably due to the fact that
+  # setting `is_reduced` flags does not pay off in such iterative procedures.
+  A = codomain(F)
+  v = evaluate(lift(u), lift.(_images(F)))
+  return simplify(A(v))
+end
+
```
afkafkafk13, line 121: No need to apologize for this in the comment! Arithmetic in quotients of **polynomial** rings is dead slow, when you reduce after each step. The art is to be rather lazy, not care about the choice of representative in complicated sequences of basic operations for a while and only reduce occasionally to normal form -- and of course whenever it is necessary to have a normal form as representative. This is precisely the philosophy you are pursuing here.

fingolfin:
@lgoettgens the tests here fail in the new `Run CI without experimental / doctest` job, inside the FTheoryTools tests, because `combinations` is not defined.

What confuses us is: why does it even run the doctests for FTheoryTools here?

HereAround:
@apturner Modulo the above failure in the `Run CI without experimental / doctest` job, I think this PR is ready for review. Can you please let me know your thoughts?

Thank you!

experimental/Experimental.jl
```diff
@@ -18,9 +18,9 @@ const oldexppkgs = [
 # For more background, see https://github.com/oscar-system/Oscar.jl/issues/2300.
 const orderedpkgs = [
   "LieAlgebras",
-  "BasisLieHighestWeight",   # nees code from LieAlgebras
+  "BasisLieHighestWeight",   # needs code from LieAlgebras
 ]
-exppkgs = filter(x->isdir(joinpath(expdir, x)) && !(x in oldexppkgs) && !(x in orderedpkgs), readdir(expdir))
+exppkgs = filter(x->isdir(joinpath(expdir, x)) && !(x in oldexppkgs) && !(x in orderedpkgs) && (x != "FTheoryTools"), readdir(expdir))
```
benlorenz: This looks suspicious, why is this necessary? If this needs a specific ordering it can be added to the list above?

```suggestion
exppkgs = filter(x->isdir(joinpath(expdir, x)) && !(x in oldexppkgs) && !(x in orderedpkgs), readdir(expdir))
```

lgoettgens:
> @lgoettgens the tests here fail in the new `Run CI without experimental / doctest` job, inside the FTheoryTools tests, because `combinations` is not defined.
> 
> What confuses us is: why does it even run the doctests for FTheoryTools here?

I am on vacation this week, but will have a look at this once I again have access to a computer.

benlorenz:
> @lgoettgens the tests here fail in the new `Run CI without experimental / doctest` job, inside the FTheoryTools tests, because `combinations` is not defined.
> 
> What confuses us is: why does it even run the doctests for FTheoryTools here?

Probably because of this line:
https://github.com/oscar-system/Oscar.jl/blob/890af7c741161c1b0a0a5212ca2460c660dae6b0/experimental/Experimental.jl#L62
This, in combination with what I asked [here](https://github.com/oscar-system/Oscar.jl/pull/3519#discussion_r1540938549) breaks the no-experimental setup.

experimental/Experimental.jl
HereAround: I need `FTheoryTools` to be loaded after the experimental `schemes` code... and found the above solution.

Key: The `schemes` code has not been formatted according to the "new experimental packages style" (i.e. with its docs, tests etc. in the experimental folder), whereas `FTheoryTools` complies with this new standard.

Hence, we are mixing new-style and old-style experimental packages and expect them to be loaded in a certain order (here: Schemes first, then FTheoryTools).

However, as I see it, it is currently hard coded that the old-style experimental packages are loaded after the new-style experimental packages. And so, it seemed that my code could only work with some additional tweaking: 
* The line that you highlight excludes `FTheoryTools` to be considered as an old-style experimental package. (Otherwise, there is an error, as the code cannot find the include files for `FTheoryTools` under the assumption that it is an old-style package).
* `FTheoryTools` has been removed from the list of the new-style experimental packages. (Otherwise, it would be loaded before schemes, which leads to `undefvar errors`, as we need schemes functionality now.)
* Finally, the very last line of the `Experimental.jl` file includes/loads the `FTheoryTools`.

Not pretty, and am happy to improve. For now, it seemed to work though.

experimental/Experimental.jl
```diff
@@ -58,3 +58,5 @@ for pkg in oldexppkgs
   # Load the package
   include(joinpath(expdir, pkg, "$pkg.jl"))
 end
+
+include("FTheoryTools/src/FTheoryTools.jl")
```
benlorenz:
```suggestion
```
We don't have any explicit includes anymore in this file.

lgoettgens:
> > @lgoettgens the tests here fail in the new `Run CI without experimental / doctest` job, inside the FTheoryTools tests, because `combinations` is not defined.
> > What confuses us is: why does it even run the doctests for FTheoryTools here?
> 
> Probably because of this line:
> 
> https://github.com/oscar-system/Oscar.jl/blob/890af7c741161c1b0a0a5212ca2460c660dae6b0/experimental/Experimental.jl#L62
> 
> 
> This, in combination with what I asked [here](https://github.com/oscar-system/Oscar.jl/pull/3519#discussion_r1540938549) breaks the no-experimental setup.

Yes, indeed.
We should get rid of the old-style experimental things at all (I.e. convert them). I think most should be just moving around some files. I can try to do this In a little disruptive way next week.
Mixing old and new styles as this is done in this PR will bite us in the long term and makes this already very finicky automatic inclusion even harder to maintain. (And it is completely orthogonal to the means of the new CI job)

HereAround:
> > > @lgoettgens the tests here fail in the new `Run CI without experimental / doctest` job, inside the FTheoryTools tests, because `combinations` is not defined.
> > > What confuses us is: why does it even run the doctests for FTheoryTools here?
> > 
> > 
> > Probably because of this line:
> > https://github.com/oscar-system/Oscar.jl/blob/890af7c741161c1b0a0a5212ca2460c660dae6b0/experimental/Experimental.jl#L62
> > 
> > This, in combination with what I asked [here](https://github.com/oscar-system/Oscar.jl/pull/3519#discussion_r1540938549) breaks the no-experimental setup.
> 
> Yes, indeed. We should get rid of the old-style experimental things at all (I.e. convert them). I think most should be just moving around some files. I can try to do this In a little disruptive way next week. Mixing old and new styles as this is done in this PR will bite us in the long term and makes this already very finicky automatic inclusion even harder to maintain. (And it is completely orthogonal to the means of the new CI job)

Most likely, it will be a lot of work to turn the schemes experimental folder into the new-style (if at all possible, with potential other PRs hanging.) Certainly would be great to have this, but maybe worth pinging @afkafkafk13 , @HechtiDerLachs and @simonbrandhorst, to hear if this is realistic.

afkafkafk13:
> > Yes, indeed. We should get rid of the old-style experimental things at all (I.e. convert them). I think most should be just moving around some files. I can try to do this In a little disruptive way next week. Mixing old and new styles as this is done in this PR will bite us in the long term and makes this already very finicky automatic inclusion even harder to maintain. (And it is completely orthogonal to the means of the new CI job)
> 
> Most likely, it will be a lot of work to turn the schemes experimental folder into the new-style (if at all possible, with potential other PRs hanging.) Certainly would be great to have this, but maybe worth pinging @afkafkafk13 , @HechtiDerLachs and @simonbrandhorst, to hear if this is realistic.

I agree that it is necessary and urgent to clean up experimental/Schemes, but from my point of view this should happen in a 2 step process:
a) Check which parts are sufficiently mature to go to src and move them
b) adjust the rest to the new style
Doing step b before step a will be extremely expensive, as a lot of files (and parts of files) will be moved around twice. this would also double the risks of this error-prone moving of stuff.

HereAround:
> > > Yes, indeed. We should get rid of the old-style experimental things at all (I.e. convert them). I think most should be just moving around some files. I can try to do this In a little disruptive way next week. Mixing old and new styles as this is done in this PR will bite us in the long term and makes this already very finicky automatic inclusion even harder to maintain. (And it is completely orthogonal to the means of the new CI job)
> > 
> > 
> > Most likely, it will be a lot of work to turn the schemes experimental folder into the new-style (if at all possible, with potential other PRs hanging.) Certainly would be great to have this, but maybe worth pinging @afkafkafk13 , @HechtiDerLachs and @simonbrandhorst, to hear if this is realistic.
> 
> I agree that it is necessary and urgent to clean up experimental/Schemes, but from my point of view this should happen in a 2 step process: a) Check which parts are sufficiently mature to go to src and move them b) adjust the rest to the new style Doing step b before step a will be extremely expensive, as a lot of files (and parts of files) will be moved around twice. this would also double the risks of this error-prone moving of stuff.

Just to clarify - this was about the style of the experimental code, not moving it into the OSCAR source.

From the above discussion (@lgoettgens and @benlorenz please correct if I am mistaken), I understood that this PR cannot be merged before `experimental/Schemes` complies with the new style of the experimental code, that @lkastner introduced a while back. This would mean to have folders:
* `experimental/Schemes/src` , containing all the files currently in `experimenal/Schemes`,
* `experimental/Schemes/docs`, with the corresponding documentation (currently presumably sprinkled into `Oscar.jl/docs`
* `experimental/Schemes/test`, with the corresponding tests (also, unless I am mistaken, currently mixed in with `Oscar.jl/test`)
* and a file `experimental/Schemes/Schemes.jl`, with the relevant `include` commands.

Copying the existing code from `experimental/Schemes` to a newly created `experimental/Schemes/src` is trivial, but disentangling the docs and tests probably a nightmare (please correct @afkafkafk13 if I am mistaken).

So I wonder what could be done to move this PR forward with the least effort on all parties involved.

afkafkafk13:
Of course, I am aware that moving stuff from Schemes to Schemes/src is not the problem ! This is the least of my worries.

But the docs and even more so some of the tests will not be a pleasant and safe task at all. Some other tests, on the other hand, are in separate files, which are again easy to move.  But we might loose some intentional implicit testing of underlying functionality from src by means of these tests.

Luckily, quite a few recent PRs in the Schemes part have been merged over the last week, but @HechtiDerLachs has several older ones sitting around and I am not sure, what the status is for these and how badly each of these would be affected. I am not even sure, whether he might have already prepared to move some things to the src-directory of Oscar (this was, what I had alluded at). This needs to be discussed with @HechtiDerLachs as soon as he is back next week.

lgoettgens:
> > > > Yes, indeed. We should get rid of the old-style experimental things at all (I.e. convert them). I think most should be just moving around some files. I can try to do this In a little disruptive way next week. Mixing old and new styles as this is done in this PR will bite us in the long term and makes this already very finicky automatic inclusion even harder to maintain. (And it is completely orthogonal to the means of the new CI job)
> > > 
> > > 
> > > Most likely, it will be a lot of work to turn the schemes experimental folder into the new-style (if at all possible, with potential other PRs hanging.) Certainly would be great to have this, but maybe worth pinging @afkafkafk13 , @HechtiDerLachs and @simonbrandhorst, to hear if this is realistic.
> > 
> > 
> > I agree that it is necessary and urgent to clean up experimental/Schemes, but from my point of view this should happen in a 2 step process: a) Check which parts are sufficiently mature to go to src and move them b) adjust the rest to the new style Doing step b before step a will be extremely expensive, as a lot of files (and parts of files) will be moved around twice. this would also double the risks of this error-prone moving of stuff.
> 
> Just to clarify - this was about the style of the experimental code, not moving it into the OSCAR source.
> 
> From the above discussion (@lgoettgens and @benlorenz please correct if I am mistaken), I understood that this PR cannot be merged before `experimental/Schemes` complies with the new style of the experimental code, that @lkastner introduced a while back. This would mean to have folders:
> 
> * `experimental/Schemes/src` , containing all the files currently in `experimenal/Schemes`,
> * `experimental/Schemes/docs`, with the corresponding documentation (currently presumably sprinkled into `Oscar.jl/docs`
> * `experimental/Schemes/test`, with the corresponding tests (also, unless I am mistaken, currently mixed in with `Oscar.jl/test`)
> * and a file `experimental/Schemes/Schemes.jl`, with the relevant `include` commands.
> 
> Copying the existing code from `experimental/Schemes` to a newly created `experimental/Schemes/src` is trivial, but disentangling the docs and tests probably a nightmare (please correct @afkafkafk13 if I am mistaken).
> 
> So I wonder what could be done to move this PR forward with the least effort on all parties involved.

I think it might be possible to just move the source for now, and postpone tests and docs for now. That's what I meant with little disruptive.
I can coordinate this next week during my visit in KL with @HechtiDerLachs and @HereAround.

HereAround:
> > > > > Yes, indeed. We should get rid of the old-style experimental things at all (I.e. convert them). I think most should be just moving around some files. I can try to do this In a little disruptive way next week. Mixing old and new styles as this is done in this PR will bite us in the long term and makes this already very finicky automatic inclusion even harder to maintain. (And it is completely orthogonal to the means of the new CI job)
> > > > 
> > > > 
> > > > Most likely, it will be a lot of work to turn the schemes experimental folder into the new-style (if at all possible, with potential other PRs hanging.) Certainly would be great to have this, but maybe worth pinging @afkafkafk13 , @HechtiDerLachs and @simonbrandhorst, to hear if this is realistic.
> > > 
> > > 
> > > I agree that it is necessary and urgent to clean up experimental/Schemes, but from my point of view this should happen in a 2 step process: a) Check which parts are sufficiently mature to go to src and move them b) adjust the rest to the new style Doing step b before step a will be extremely expensive, as a lot of files (and parts of files) will be moved around twice. this would also double the risks of this error-prone moving of stuff.
> > 
> > 
> > Just to clarify - this was about the style of the experimental code, not moving it into the OSCAR source.
> > From the above discussion (@lgoettgens and @benlorenz please correct if I am mistaken), I understood that this PR cannot be merged before `experimental/Schemes` complies with the new style of the experimental code, that @lkastner introduced a while back. This would mean to have folders:
> > 
> > * `experimental/Schemes/src` , containing all the files currently in `experimenal/Schemes`,
> > * `experimental/Schemes/docs`, with the corresponding documentation (currently presumably sprinkled into `Oscar.jl/docs`
> > * `experimental/Schemes/test`, with the corresponding tests (also, unless I am mistaken, currently mixed in with `Oscar.jl/test`)
> > * and a file `experimental/Schemes/Schemes.jl`, with the relevant `include` commands.
> > 
> > Copying the existing code from `experimental/Schemes` to a newly created `experimental/Schemes/src` is trivial, but disentangling the docs and tests probably a nightmare (please correct @afkafkafk13 if I am mistaken).
> > So I wonder what could be done to move this PR forward with the least effort on all parties involved.
> 
> I think it might be possible to just move the source for now, and postpone tests and docs for now. That's what I meant with little disruptive. I can coordinate this next week during my visit in KL with @HechtiDerLachs and @HereAround.

Yeah, I was also starting to hope that we could get away by just taking care of the source. If successful, this is probably the way to go here.

(I was admittedly hoping to have this PR merged today, or at least this week. But then, I guess it can wait until next week. @apturner please let me know your feedback on the FTheoryTools changes, so I can try to work them in before we face up to the challenges with `experimenal/Schemes` next week).

Indeed, let us discuss this next week in KL.

HereAround:
(Reverted to draft for the time being...)

HereAround:
@HechtiDerLachs suggested an easier workaround to make us all happy, namely we "downgrade" (even though we keep the structure) `FTheoryTools`, in the sense that it will be included as an old experimental package. Thereby it can be loaded after the schemes, without the heroic effort of aligning the experimental schemes code to the new experimental standard.

HechtiDerLachs (approved):

lgoettgens:
> @HechtiDerLachs suggested an easier workaround to make us all happy, namely we "downgrade" (even though we keep the structure) `FTheoryTools`, in the sense that it will be included as an old experimental package. Thereby it can be loaded after the schemes, without the heroic effort of aligning the experimental schemes code to the new experimental standard.

This looks like a good idea for the meantime, in particular, to not block this PR.

experimental/Experimental.jl
```diff
@@ -10,6 +10,7 @@ const oldexppkgs = [
   "ModStd",
   "Rings",
   "Schemes",
+  "FTheoryTools"
```
lgoettgens:
```suggestion
  "FTheoryTools", # needs to be loaded after Schemes
```
adding a comment to not remember this for the future.

HereAround: Added to my last commit.

experimental/Schemes/CoveredProjectiveSchemes.jl
```diff
@@ -238,7 +238,7 @@ gluing_morphisms(PG::ProjectiveGluing) = (PG.f, PG.g)
     )
     C in coverings(Y) || error("covering not listed")
     for P in values(projective_patches)
-      base_scheme(P) in patches(C) || error("base scheme not found in covering")
+      any(x->x===base_scheme(P), patches(C)) || error("base scheme not found in covering")
```
simonbrandhorst, line 241: is this a `==` vs `===` issue? Looks like an easy mistake.

PR #3525, opened 2024-03-18
Add CI job that runs tests without the `experimental` folder

lgoettgens:
Addresses the second bullet point of https://github.com/oscar-system/Oscar.jl/issues/3408.

The new CI job runs test-short, test-long, doctest, and documentation building after making everything in `experimental/` (but a whitelist) available. I started with the whitelist containing everything and then tried to remove them one by one and either keep them removed or add a comment with an error or similar.
The new script in `etc/` takes `init` or `cleanup` as an argument to remove or revert removal of the stuff in experimental. This can be used for local testing.


benlorenz:
Do we really need to do all this moving around, this makes it somewhat complicated to test locally?

I would try this with a `filter!` on `exppkgs` and `oldexppkgs`, something like this:
```
if haskey(ENV, "OSCAR_TEST_NOEXPERIMENTAL")
  include("expignore.jl")
  filter!(x -> !(x in expignore), exppkgs)
  filter!(x -> !(x in expignore), oldexppkgs)
end
```
The file can then just contain one variable definition with the folder names.

We could also keep the file in the repo under a different name and then link it to `expignore.jl` in the CI job, then we could get rid of the ENV variable and instead just rely on the existence of that symlink. (And creating / deleting that symlink would also automatically trigger a new precompilation)
Or use a preference to choose whether that file is loaded, then one can set this directly from the Oscar prompt? (Makes it slightly less convenient for the CI, but preferences should also trigger precompilation)

_Edit:_ sorry about the close/reopen, that was just a misclick.

lgoettgens:
Thanks for your good ideas. Let me change this back to a draft for the time being. I'll try to make this a bit more convenient to use. (The latest in the week after Easter due to vacation in the meantime)

benlorenz (approved):
Thanks!

lgoettgens:
Can someone with appropriate rights can add the new 4 jobs as "required"? Since there is an easy workaround to circumvent, I see no real reason to not force this to keep working.

benlorenz:
> Can someone with appropriate rights can add the new 4 jobs as "required"? Since there is an easy workaround to circumvent, I see no real reason to not force this to keep working.

I would prefer to do that tomorrow, then we don't have to re-run as many CI jobs due to missing these checks.

PR #3526, opened 2024-03-19
Handle empty charts for subschemes of ideal sheaves

HechtiDerLachs:
As was pointed out by @afkafkafk13 in #3494, the constructor for subschemes from ideal sheaves had some deficiencies. This should be ruled out with this fix.

### DIFF
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -333,33 +333,28 @@ end
 For an ideal sheaf ``ℐ`` on an `AbsCoveredScheme` ``X`` return
 the subscheme ``Y ⊂ X`` given by the zero locus of ``ℐ``.
 """
-function subscheme(I::AbsIdealSheaf)
+function subscheme(I::AbsIdealSheaf; covering::Covering=default_covering(scheme(I)))
   X = space(I)
-  C = default_covering(X)
-  new_patches = [subscheme(U, I(U)) for U in basic_patches(C)]
+  C = covering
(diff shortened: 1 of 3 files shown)


experimental/Schemes/IdealSheaves.jl
```diff
@@ -333,33 +333,28 @@ end
...
-    Unew = new_patches[i]
-    Vnew = new_patches[j]
-    G = C[U, V]
-    #new_gluings[(Unew, Vnew)] = restrict(C[U, V], Unew, Vnew, check=false)
-    new_gluings[(Unew, Vnew)] = LazyGluing(Unew, Vnew, _compute_restriction,
+  for (U, Unew) in new_patches
+    for (V, Vnew) in new_patches
+      (U, V) in keys(gluings(C)) || continue # No gluing before, no gluing after.
+      old_glue = C[U, V]
+      new_gluings[(Unew, Vnew)] = LazyGluing(Unew, Vnew, _compute_restriction,
```
simonbrandhorst, line 346: it seems to me that the variable `_compute_restriction` is undefined?

HechtiDerLachs: It is a function which is passed on as an argument. 

Maybe this just reads a little odd?

afkafkafk13: Yes, it is the function name to be used in the lazy gluing. A function of this name is available in LazyGluing.jl.
(Sorry this was typed at the same time as Hechti's answer, but took longer to upload (slow line).)

afkafkafk13 (approved):
Looks good to me.

PR #3532, opened 2024-03-20
Add `is_finite_dimensional_vector_space`

### LINKED ISSUE #3510: vector_space_dimension returning infinity instead of throwing an error
Working with vector_space_dimension can be annoing since it throws an error exception in the infinite dimensional case. 
You always have to treat the infinite dimensional case separately, even if it would not be nessesary like for example if you want to check if two MPolyQuoRings have the same vector space dimension.  When doing simple things like that you could easily forget to treat the infinite dimensional case at all.

 vector_space_dimension(A::MPolyQuoRing)  should in my opinion return infinity instead of throwing an error exception if A is a infinite dimensional vectorspace.

**Describe alternatives you've considered**
A minimal solution would be to provide a function is_finite_dimensional_vector_space().

joschmitt:
Resolves #3510.

Implements what I believe to be the result of the discussion in #3510: A function `is_finite_dimensional_vector_space(::MPolyQuoRing)` is added and `vector_space_dimension` now throws an error of type `InfiniteDimensionError` if the dimension is infinite.

For completeness: `vector_space_dimension` also exists for certain localized rings and modules. In theory, we could/should add `is_finite_dimensional_vector_space` also for those, but it seems to me that in those cases `is_finite_dimensional_vector_space` would do some non-trivial computations which have to be repeated by `vector_space_dimension`.

### DIFF
--- a/src/Rings/mpoly-affine-algebras.jl
+++ b/src/Rings/mpoly-affine-algebras.jl
@@ -19,11 +19,45 @@ julia> dim(A)
 1
 ```
 """
-function dim(A::MPolyQuoRing) 
+function dim(A::MPolyQuoRing)
   I = A.I
   return dim(I)
 end
 
+@doc raw"""
+    is_finite_dimensional_vector_space(A::MPolyQuoRing)
+
+If, say, `A = R/I`, where `R` is a multivariate polynomial ring over a field
+`K`, and `I` is an ideal of `R`, return `true` if `A` is finite-dimensional
+as a `K`-vector space, `false` otherwise.
+
+# Examples
+```jldoctest
+julia> R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"]);
+
+julia> A, _ = quo(R, ideal(R, [x^3+y^3+z^3-1, x^2+y^2+z^2-1, x+y+z-1]));
+
+julia> is_finite_dimensional_vector_space(A)
+true
+
+julia> A, _ = quo(R, ideal(R, [x]));
+
+julia> is_finite_dimensional_vector_space(A)
+false
+```
+"""
+function is_finite_dimensional_vector_space(A::MPolyQuoRing)
+  # We check '<=' because A might be the zero ring, so dim(A) == -1
+  return dim(A) <= 0
+end
+
+struct InfiniteDimensionError <: Exception
+end
+
+function Base.showerror(io::IO, err::InfiniteDimensionError)
+  println(io, "Infinite-dimensional vector space")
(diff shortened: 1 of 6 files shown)


afkafkafk13:
> Resolves #3510.
> 
> Implements what I believe to be the result of the discussion in #3510: A function `is_finite_dimensional_vector_space(::MPolyQuoRing)` is added and `vector_space_dimension` now throws an error of type `InfiniteDimensionError` if the dimension is infinite.
> 
> For completeness: `vector_space_dimension` also exists for certain localized rings and modules. In theory, we could/should add `is_finite_dimensional_vector_space` also for those, but it seems to me that in those cases `is_finite_dimensional_vector_space` whould do some non-trivial computations which have to be repeated by `vector_space_dimension`.

Indeed. Your fix now raises an appropriate error also in the local case, as this is referred to the leading ideal of the  "shifted ideal" and the global case.

For 'is_finite_dimensional_vector_space' in the local setting, we would need to cache the leading ideal of the shifted ideal -- which makes a lot of sense anyway. But I would prefer to discuss this also with @HechtiDerLachs , when he is back from his holidays. He designed the localizations and know the pitfalls best, which might appear, if we take this route.

afkafkafk13:
P.S.: Please add is_finite_dimensional_vector_space returning false for MPolyAnyNonQuoRing and one which says 'not implemented yet' for the localized quotient ring case.
Edit: This is to make it look agnostic for the enduser.

src/Rings/mpoly-affine-algebras.jl
```diff
@@ -19,11 +19,46 @@ julia> dim(A)
...
+false
+```
+"""
+function is_finite_dimensional_vector_space(A::MPolyQuoRing)
+  # We check '<=' because A might be the zero ring, so dim(A) == -1
+  return dim(A) <= 0
+end
+
+struct InfiniteDimensionError <: Exception
+    InfiniteDimensionError() = new()
```
thofma:
```suggestion
```
Should not be necessary.

joschmitt:
> P.S.: Please add is_finite_dimensional_vector_space returning false for MPolyAnyNonQuoRing and one which says 'not implemented yet' for the localized quotient ring case. Edit: This is to make it look agnostic for the enduser.

Good idea, I added those functions.

afkafkafk13:
Thank you. Looks good now. I will approve it, as soon as I have seen the output of the tests.
By the way: why are they stalled right now?

afkafkafk13 (approved):

fingolfin (approved):
Thank you

PR #3535, opened 2024-03-20
Bugfixes for normalization of covered schemes

simonbrandhorst:
and the corresponding test. 
Also removes some very expensive and unnecessary checks. 
Please merge asap. 

(it turned out the "denominators" were not invertible as we previously thought)
cc. @paemurru


afkafkafk13 (approved):
Yes, indeed. divexact does the job.

PR #3540, opened 2024-03-22
Upgrade serialization of `ZZLatWithIsom`

StevellM:
@antonydellavecchia I have tried to create an upgrade file, though I am not sure it respects the wanted format. I have added a part of summary, and the only thing I wanted to upgrade. Please let me know if I should modify it, and how (in particular to be more flexible with possible other upgrade scripts)

### DIFF
--- a/src/Serialization/Upgrades/1.0.0.jl
+++ b/src/Serialization/Upgrades/1.0.0.jl
@@ -1,0 +1,20 @@
+################################################################################
+# Upgrade Summary
+# 
+# - `ZZLatWithIsom`: we now only serialized the ambient `QuadSpaceWithIsom` and
+# a basis matrix of the underlying lattice; the upgrade recovers the basis matrix
+# in the old tree.
+
+push!(upgrade_scripts_set, UpgradeScript(
+  v"1.0.0",
+  function upgrade_1_0_0(s::UpgradeState, dict::Dict)
+
+    upgraded_dict = dict
+
+    if haskey(dict, :_type) && dict[:_type] == "ZZLatWithIsom"
+      haskey(dict[:data], :basis) || (upgraded_dict[:data][:basis] = dict[:data][:lattice][:data][:basis])
+    end
+
+    return upgraded_dict
+  end
+))
--- a/experimental/QuadFormAndIsom/src/serialization.jl
+++ b/experimental/QuadFormAndIsom/src/serialization.jl
@@ -30,16 +30,15 @@ end
 
 @register_serialization_type ZZLatWithIsom
 
-# This should be changed by saving only the basis matrix of `L`
 function save_object(s::SerializerState, L::ZZLatWithIsom)
   save_data_dict(s) do
     save_typed_object(s, ambient_space(L), :ambient_space)
-    save_typed_object(s, lattice(L), :lattice)
+    save_typed_object(s, basis_matrix(L), :basis)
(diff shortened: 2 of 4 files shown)


src/Serialization/Upgrades/1.0.0.jl
```diff
@@ -0,0 +1,20 @@
...
+# in the old tree.
+
+push!(upgrade_scripts_set, UpgradeScript(
+  v"1.0.0",
+  function upgrade_1_0_0(s::UpgradeState, dict::Dict)
+
+    upgraded_dict = dict
+
+    if haskey(dict, :_type) && dict[:_type] == "ZZLatWithIsom"
+      haskey(dict[:data], :basis) || (upgraded_dict[:data][:basis] = dict[:data][:lattice][:data][:basis])
```
antonydellavecchia:
```suggestion
      haskey(dict[:data], :basis) && (upgraded_dict[:data][:basis] = dict[:data][:lattice][:data][:basis])
```
I imagine you meant to apply the code on the right?

src/Serialization/Upgrades/1.0.0.jl
```diff
@@ -0,0 +1,20 @@
+################################################################################
+# Upgrade Summary
+# 
+# - `ZZLatWithIsom`: we now only serialized the ambient `QuadSpaceWithIsom` and
+# a basis matrix of the underlying lattice; the upgrade recovers the basis matrix
+# in the old tree.
+
+push!(upgrade_scripts_set, UpgradeScript(
+  v"1.0.0",
```
antonydellavecchia: technically you should put the version of oscar that this will be released with, 
not exactly sure what this will be, but probably 1.1.0?

antonydellavecchia:
@StevellM so you shouldn't have to worry about other upgrades if there were no adjustments in the format for the types you are using, otherwise, the upgrades from earlier versions will be run first, and you should only care about your specific types upgraded from the latest upgrade to the current. It's possible some types you need were missed, and if this is the case, I'll fix the earlier upgrade scripts to make sure that you only have to worry about your upgrades

src/Serialization/Upgrades/1.0.0.jl
StevellM: No, because if there is already `:basis` known we do not have anything to do. In the previous serialization, we had `:ambient_space` and `:lattice` and I replaced the `:lattice` by `:basis`: so for the upgrade, I go through the old tree to recover `:basis` inside `:lattice`. But the new file already know `:basis`, so I do not have to modify anything in that case. Hope it is clear.

src/Serialization/Upgrades/1.0.0.jl
StevellM: Ah yes indeed, I misunderstood it and put the current version. I will update this, thank you!

src/Serialization/Upgrades/1.0.0.jl
antonydellavecchia: ah ok, but then `=` should be  `==`?

antonydellavecchia: ah ok, i see

fingolfin:
So is this OK to merge or not? Maybe @antonydellavecchia can comment?

antonydellavecchia (approved):
Looks good

PR #3541, opened 2024-03-25
Hide Oscar.jl banner, when used in other module

lkastner:
Uses https://github.com/Nemocas/AbstractAlgebra.jl/pull/1643


benlorenz (approved):

lgoettgens (approved):

lgoettgens:
Is this suitable for backporting?

benlorenz:
> Is this suitable for backporting?

Probably yes, could be considered as a bug fix...? And we do have the corresponding AbstractAlgebra in the deps anyway.

PR #3544, opened 2024-03-27
[FTheoryTools] Remove stanley reisner ideal for family of spaces

HereAround:
~~This is based on https://github.com/oscar-system/Oscar.jl/pull/3519 for my convenience and needs to be rebased once said PR is merged.~~

cc @apturner

(Seems good to go once the above PR has been fixed. This PR removed the function `stanley_reisner_ideal` for `FamilyOfSpaces`, as we discussed quite some time ago.)


HereAround:
@apturner What is your take on this PR?

apturner (approved):
Thanks for this @HereAround, this looks good, sorry for the delay.

We could probably also remove the irrelevant ideal and ideal of linear relations, along the same lines.

HereAround:
> Thanks for this @HereAround, this looks good, sorry for the delay.
> 
> We could probably also remove the irrelevant ideal and ideal of linear relations, along the same lines.

Maybe I do recall this incorrectly - I was under the impression that a while back we seemed to have reached the conclusion that the irrelevant ideal and the ideal of linear relations do make sense for family of spaces? If not, then of course we should remove them (and adjust all functions that make use of this information for family of spaces).

HereAround:
@apturner I will merge this change here. Let us discuss the irrelevant ideal and ideal of linear relations separately.

PR #3546, opened 2024-03-27
feat: add resultant for list of multivariate polynomials

thofma:
I need it for an application and I think it would be nice to have. There are more algorithms that one could implement, but this one works fine for my application.

@wdecker maybe you could have a look?


fieker:
Can someone checke the math? The wiki page wants homogenous polynomials, but this is not checked (by the looks: deliberatey not)
Should this apply to graded polynomials (only)? Standart grading? Any grading?

thofma:
It must be homogeneous with respect to the standard grading. It is not checked, because it cannot be checked with the current state of OSCAR:
```
julia> K = algebraic_closure(QQ); Kxy, (x, y) = K["x", "y"]; is_homogeneous(x)
true

julia> K = QQ; Kxy, (x, y) = K["x", "y"]; is_homogeneous(x)
ERROR: MethodError: no method matching is_homogeneous(::QQMPolyRingElem)
```

fingolfin:
https://github.com/Nemocas/AbstractAlgebra.jl/pull/1649 adds `is_homogeneous` for all `MPolyRingElem`

thofma:
I remembered that we have an `_is_homogeneous` for this purpose. I am using this and added the test. This should be good to go now.

ederc (approved):

PR #3551, opened 2024-04-02
Move partial characters from `experimental/Rings/` to `src/`

lgoettgens:
If somebody currently works on these files, please object so that we can postpone this change here. Otherwise, I think this is a no-brainer.

Edit: On request by @fieker, this moves the files to src instead (and some obvious cleanup).


fieker:
On Tue, Apr 02, 2024 at 04:46:08AM -0700, Lars Göttgens wrote:

Maybe generically, sort this. There are no rings defined in here
anymore, QQab is long mainstram.
The partial characters are used (exclusively) for the binomial stuff,
the binomial stuff is used in
 examples/BinomialIdeal and BinomialIdeal2
 src/Rings/binomial_ideals.jl
So I think this is a prime candidate for not being experimental any
more...

fingolfin (approved):

PR #3557, opened 2024-04-03
Delete unused old code and tests for graded modules

fingolfin:
My understanding is that this code is currently unused, yet this is not immediately obvious and in the past lots of people have spent time trying to adapt this code to changes elsewhere, without realizing it is never used.

Unless we have plans to reactivate some or all of this, I propose we just outright delete this.

Perhaps @fieker or @jankoboehm have something to say about this?


fingolfin:
@fieker is OK with this, @jankoboehm did not respond, so I'll just merge this now.

PR #3562, opened 2024-04-04
Normalization output is now a triple

paemurru:
Everything done.


thofma:
Is `normalization` an exported function that exists in version 1.0?

simonbrandhorst:
It is an exported function and existed in 1.0 but with a different signature (e.g. only for affine algebras).
This one taking schemes as input was added about 2 weeks ago in #3494
If there was no Oscar release in between, a deprecation should not be necessary?

thofma:
Yes, in this case everything is fine. Just wanted to double check.

simonbrandhorst:
> Yes, in this case everything is fine. Just wanted to double check.

Thanks for double checking :-)

afkafkafk13:
@paemurru : from my perspective this looks good!
What do you plan to add further to this PR? As soon as the docstrings are available and tests for the new functionality are running, this seems ready for approval/merging.

paemurru:
I have now added everything that I planned.

I would have preferred if in my `disjoint_union` example one of the schemes is an affine scheme, but apparently the function `covered_scheme` cannot be applied to affine schemes.

simonbrandhorst (approved):

Issue #3547, opened 2024-03-27
inconsistent directions of isomorphisms


ThomasBreuer:
(The following arose from a discussion with @fieker.)

There are quite a few Oscar functions that take a group and return an isomorphism from or to an isomorphic group, typically one of a prescribed different type.
The problem is whether "from" or "to" shall be used.

The `isomorphism(T, G)` methods for a type `T` and a group `G` return an isomorphism from the given `G` to a new group of type `T`, whereas `Oscar.fp_group_with_isomorphism` and `Oscar.pc_group_with_isomorphism` return an isomorphism from the newly created group to the given one.
(One argument for the latter choice could be that `isomorphism(H, G)`, for two groups, returns an isomorphism from `H` to `G`.)

Technically, the direction of the map does not really matter; one can ask for images or preimages, depending on which direction one gets. Thus there is no "natural" direction, contrary to the situation with embeddings (where the map naturally goes from the small to the large object) or quotients (from the large object to the factor) or maps describing actions (from the given group to something that contains the image, without the need to actually compute this image).

One non-breaking solution would be to keep and deprecate the current `Oscar.fp_group_with_isomorphism` (`Oscar.pc_group_with_isomorphism`), and to introduce recommended `isomorphism_from_fp_group` or `isomorphism_to_fp_group` (`isomorphism_from_pc_group` or `isomorphism_to_pc_group`), such that the function name explains what the result is. (Claus does not like this idea.)

Magma seems to live with the same inconsistencies: `PCGroup(G)` returns a map from `G` to a pc group, whereas `FpGroup(G)` returns a map from an fp group to `G`.

If we keep the current "mixed" situation then the documentation of each function that returns an isomorphism must state which direction the result has.

fieker:
On Wed, Mar 27, 2024 at 07:51:01AM -0700, Thomas Breuer wrote:
> (The following arose from a discussion with @fieker.)
> 
> There are quite a few Oscar functions that take a group and return an isomorphism from or to an isomorphic group, typically one of a prescribed different type.
> The problem is whether "from" or "to" shall be used.
> 
> The `isomorphism(T, G)` methods for a type `T` and a group `G` return an isomorphism from the given `G` to a new group of type `T`, whereas `Oscar.fp_group_with_isomorphism` and `Oscar.pc_group_with_isomorphism` return an isomorphism from the newly created group to the given one.
[...]

PR #3563, opened 2024-04-05
change `fp_group_with_isomorphism`, `pc_group_with_isomorphism`

ThomasBreuer:
- add optional boolean `on_gens` for `isomorphism(FPGroup, G)` (changes the internal structure of the `:isomorphisms` attribute)
- reverse the direction of the maps returned by `fp_group_with_isomorphism` and `pc_group_with_isomorphism`
- do not export `fp_group_with_isomorphism`, `pc_group_with_isomorphism` (experimental functions)
- create a collector for a pc group in Oscar not in GAP
- fix code using sparse matrices (add a test for the case that did not work before)

(This is the first step towards a solution for #3547.)


fieker (approved):

lgoettgens:
This should fix the Aqua failures here.

PR #3569, opened 2024-04-09
Visualization in Polyhedral Geometry: Allow options and multiple objects

alexej-jordan:
The `visualize` function now allows to redirect options given as keyword arguments to polymake's visualization.

We introduce a `composed_visualization` function to visualize multiple objects in the same sketch. This could also b renamed to a `visualize` method that just takes several objects. Any feelings about this?

Options for multiple objects can be given as a `Dict` which tries to preserve as much natural syntax as possible. Here an example:

```
C = [cube(3, 0, i) for i in 1:10]
10-element Vector{Polyhedron{QQFieldElem}}:
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3
 Polytope in ambient dimension 3

julia> x = (2 => (PointColor = "blue", EdgeColor = "red"), 4 => (PointColor = "green", EdgeStyle = "hidden"))
(2 => (PointColor = "blue", EdgeColor = "red"), 4 => (PointColor = "green", EdgeStyle = "hidden"))

julia> y = Dict(x)
Dict{Int64, NamedTuple{names, Tuple{String, String}} where names} with 2 entries:
  4 => (PointColor = "green", EdgeStyle = "hidden")
  2 => (PointColor = "blue", EdgeColor = "red")

julia> Oscar.compose_visualization(C...; kwargs = y)
```


src/PolyhedralGeometry/visualization.jl
```diff
@@ -1,13 +1,79 @@
...
+- `LabelAlignment`: Defines the alignment of the vertex labels: `"left"`, `"right"` or `"center"`.
 """
-function visualize(P::Union{Polyhedron{T}, Cone{T}, PolyhedralFan{T}, PolyhedralComplex{T}}) where T<:Union{Float64, FieldElem}
+function visualize(P::Union{Polyhedron{T}, Cone{T}, PolyhedralFan{T}, PolyhedralComplex{T}, SubdivisionOfPoints{T}}; kwargs...) where T<:Union{Float64, FieldElem}
+  _prepare_visualization(P)
+  pmo = pm_object(P)
+  Polymake.visual(pmo; kwargs...)
+end
+
+function compose_visualization(P::Vararg{Union{Polyhedron{T}, Cone{T}, PolyhedralFan{T}, PolyhedralComplex{T}, SubdivisionOfPoints{T}}}; kwargs::Dict = Dict{Int, Nothing}()) where T<:Union{Float64, FieldElem}
```
lgoettgens: This vararg here is suboptimal as julia needs to recompile the function for each number of inputs and combination of types. I think a `Vector{<:Union{Polyhedron{T}, Cone{T}, PolyhedralFan{T}, PolyhedralComplex{T}, SubdivisionOfPoints{T}}` would perform better here.

Furthermore, the failing Aqua test is due to this signature. In the zero-argument case, the type parameter `T` has no occurrences in the signature.

lkastner (approved):

PR #3572, opened 2024-04-09
improve `isomorphism(FPGroup, abelian_group(...))`

ThomasBreuer:
Use known information instead of checking it.


lgoettgens (approved):

fieker (approved):
looks like it makes sense

PR #3573, opened 2024-04-09
Weak compositions

joschmitt:
As remarked in the context of #3147 we have three (known) implementations of weak compositions of `n` into `d` parts. (To my knowledge, a weak composition is a sequence of `d` non-negative integers that sum up to `n`.)

This pull request introduces a type `WeakComposition` analogously to `Partition` and an iterator over all weak compositions of `n` into `d` parts. There is not much more functionality for weak compositions so far because I personally am happy if I can iterate them.

This unifies the mentioned three implementations:
* The former non-exported `weak_compositions` is removed/replaced
* The new iterator is used in `monomials_of_degree` -- the user shouldn't notice anything
* The non-exported `MultiIndicesOfDegree` is removed. Warning: `MultiIndicesOfDegree(d, n)` has to be replaced by `weak_compositions(n, d)`!


src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
```diff
@@ -0,0 +1,182 @@
...
+```
+"""
+function weak_compositions(n::IntegerUnion, k::IntegerUnion)
+  kk = Int(k)
+  return WeakCompositions(n, kk)
+end
+
+# I have no idea what to call these getter functions
+base(W::WeakCompositions) = W.n
+parts(W::WeakCompositions) = W.k
```
lgoettgens, line 103: Do we even need these at all?

src/aliases.jl
```diff
@@ -38,6 +38,7 @@ function n_rays end
 function n_vertices end
 
 function number_of_partitions end
+function number_of_weak_compositions end
 function number_of_patches end
```
lgoettgens:
```suggestion
function number_of_patches end
function number_of_weak_compositions end
```
keep the list sorted

src/aliases.jl
```diff
@@ -58,6 +59,7 @@ function number_of_patches end
 @alias number_of_vertices n_vertices
 
 @alias n_partitions number_of_partitions
+@alias n_weak_compositions number_of_weak_compositions
 @alias n_patches number_of_patches
```
lgoettgens:
```suggestion
@alias n_patches number_of_patches
@alias n_weak_compositions number_of_weak_compositions
```
keep the list sorted

src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
```diff
@@ -0,0 +1,182 @@
...
+  return getindex(data(C), Int(i))
+end
+
+function Base.setindex!(C::WeakComposition, x::IntegerUnion, i::IntegerUnion)
+  return setindex!(data(C), x, Int(i))
+end
+
+function Base.copy(C::WeakComposition)
+  return weak_composition(copy(data(C)))
+end
```
lgoettgens, line 62: Does this provide a suitable fallback for `deepcopy`? If not, this (or rather `deepcopy_internal`) should be added as well.

src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
```diff
@@ -0,0 +1,182 @@
...
+function Base.getindex(C::WeakComposition, i::IntegerUnion)
+  return getindex(data(C), Int(i))
+end
+
+function Base.setindex!(C::WeakComposition, x::IntegerUnion, i::IntegerUnion)
+  return setindex!(data(C), x, Int(i))
+end
+
+function Base.copy(C::WeakComposition)
+  return weak_composition(copy(data(C)))
```
lgoettgens:
```suggestion
  return weak_composition(copy(data(C)); check=false)
```

src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
```diff
@@ -0,0 +1,182 @@
...
+#
+#  Generating weak compositions
+#
+################################################################################
+
+@doc raw"""
+    weak_compositions(n::IntegerUnion, k::IntegerUnion)
+
+Return an iterator over all weak compositions of a non-negative integer `n` into
+`k` parts, produced in lexicographically *descending* order.
```
lgoettgens, line 74: Is there a particular reason why you chose the descending order? I am not sure if I find asc or desc more intuitive to think about.

lgoettgens (commented):
Thanks! I have some minor comments, but overall it looks great.

src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
joschmitt: Since the fields are called `.n` and `.k`, I prefer getter functions. I could of course also give the fields more meaningful names, but then I still would not know what to call them :)

src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
joschmitt: The reason is that this the order the algorithm I implemented years ago produces. Also the `partitions` produce lexicographically descending order, so there would be some 'consistency'.
As long as nobody has an application where the order is crucial, I would say it is good enough that we guarantee anything at all.

joschmitt: That said, I have the suspicion the order might be relevant for the `DoubledAndHyperComplex` tests... I have to investigate more.

src/Combinatorics/EnumerativeCombinatorics/weak_compositions.jl
joschmitt: The `copy` is here because otherwise `copy(::WeakComposition)` would produce a `Vector`. `deepcopy` seems to work fine presumably because the type `WeakComposition` has just one field.

lgoettgens: Yeah, for basic recursive deepcopying there is a fallback. For most of our types we have to add something on our own to not create copies of the parents.

joschmitt: Sure. These don't have parents.

experimental/DoubleAndHyperComplexes/test/ext.jl
```diff
@@ -193,25 +193,25 @@ end
...
-  @test !iszero(phi[1])
+  @test !iszero(phi[0])
   @test compose(map(dom_tot, 1), phi[0]) == compose(phi[1], map(cod_tot, 1))
 
   phi = interp(H0[3])
   dom_tot = domain(phi)
   cod_tot = codomain(phi)
-  @test !iszero(phi[0]) 
+  @test !iszero(phi[0])
+  @test !iszero(phi[1])
```
joschmitt, line 214: @HechtiDerLachs I made these adjustments to the tests: The order produced by `weak_compositions` is the opposite of what `MultiIndicesOfDegree` did and apparently this swaps the 2nd and 3rd generator of `H0` in this test. I assume that this is still correct and the comment in line 195-196 suggests it's expected, but could you have a look whether this makes sense?

lgoettgens (approved):
Once @HechtiDerLachs has approved the changes to the DoubleAndHyperComplexes, I am happy with this change. Thanks @joschmitt for working on this.

joschmitt:
Hechti is on holidays. We can wait with merging until he is back as far as I'm concerned.

joschmitt:
Design question: What should a function like `number_of_partitions` or `number_of_weak_compositions` do for negative input: error or return 0?

I would have said 'error' because asking for the number of partitions of -1 does not make sense, but apparently so far we returned 0 and this is even tested. What do you think? @ulthiel @mjrodgers (or anybody else interested in combinatorics)

mjrodgers:
> Design question: What should a function like `number_of_partitions` or `number_of_weak_compositions` do for negative input: error or return 0?
> 
> I would have said 'error' because asking for the number of partitions of -1 does not make sense, but apparently so far we returned 0 and this is even tested. What do you think? @ulthiel @mjrodgers (or anybody else interested in combinatorics)

0 is almost always better in my experience. There are often applications where we want to compute a sum of such numbers and take all of the ones that “don’t make sense” as 0. It’s a pain to have to catch errors to deal with this.

ulthiel:
0 probably: number_of_partitions(-1) is the cardinality of the set of partitions of -1, this set is empty, so its cardinality is 0.

ulthiel:
Question: What is the weak in weak compositions? Isn't this just compositions? Also I have https://github.com/ulthiel/Oscar.jl/blob/ut/compositions/experimental/JuLie/compositions.jl in which I have put quite some energy (but somehow lost track of opening a PR).

lgoettgens:
> Question: What is the weak in weak compositions? Isn't this just compositions?

0 is an allowed summand in weak compositions.

joschmitt:
> 0 is almost always better in my experience. There are often applications where we want to compute a sum of such numbers and take all of the ones that “don’t make sense” as 0. It’s a pain to have to catch errors to deal with this.

> 0 probably: number_of_partitions(-1) is the cardinality of the set of partitions of -1, this set is empty, so its cardinality is 0.

Sounds reasonable, thank you.

joschmitt:
> Question: What is the weak in weak compositions? Isn't this just compositions? Also I have https://github.com/ulthiel/Oscar.jl/blob/ut/compositions/experimental/JuLie/compositions.jl in which I have put quite some energy (but somehow lost track of opening a PR).

The parts in a weak composition are allowed to be 0, in a composition they have to be positive.
I'm aware of the JuLie functionality for compositions and started working locally on moving it to OSCAR. This would then also make ascending compositions out of the `ascending_partitions` that still float around in experimental as discussed at some point. (No promise on when this will be done.)

ulthiel:
Excellent.

experimental/DoubleAndHyperComplexes/test/ext.jl
HechtiDerLachs: I think it's fine. Thanks for the work!

HechtiDerLachs (approved):

lgoettgens:
The booktests failed. I think this is already fixed by https://github.com/oscar-system/Oscar.jl/pull/3604, but restarting CI here to make sure. Once it is finished, I will go ahead and merge (unless there are any objections).

joschmitt:
> The booktests failed. I think this is already fixed by #3604, but restarting CI here to make sure. Once it is finished, I will go ahead and merge (unless there are any objections).

Thanks.

PR #3575, opened 2024-04-09
`QuadFormAndIsom`: small optimizations after users' feedbacks

StevellM:
Small obvious optimizations that were missing in the code.


simonbrandhorst (approved):

experimental/QuadFormAndIsom/src/embeddings.jl
```diff
@@ -705,7 +716,7 @@ function _primitive_extensions_generic(
         stabHMphi, _ = sub(OHN, _stabHMphi)
         SM, _ = intersect(C, stabHMphi)
 
-        if length(elementary_divisors(HN)) == 1
+        if allequal(elementary_divisors(HN))
```
simonbrandhorst: This requires at least julia 1.8. But we support also lower versions.

PR #3577, opened 2024-04-10
Standard basis computations using highest corner

ederc:
This implements a variant of standard basis computations from Singular for local monomial orderings: If computing over QQ, do a first computation over a finite field in order to get the highest corner. Now use this highest corner to speed up the computation over QQ. If the highest corner approach does not work (e.g. dimension  > 0) Singular's usual standard basis implementation is used.

### DIFF
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -308,7 +308,7 @@ function groebner_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_orde
 end
 
 @doc raw"""
-    standard_basis_with_highest_corner(I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I))) 
+    standard_basis_highest_corner(I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I))) 
 
 Return a standard basis of `I` with respect to `ordering`. `ordering` needs to be local, the coefficient ring needs to be `QQ`.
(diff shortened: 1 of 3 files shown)


src/Rings/groebner.jl
```diff
@@ -304,6 +307,30 @@ function groebner_basis(I::MPolyIdeal; ordering::MonomialOrdering = default_orde
     return standard_basis(I, ordering=ordering, complete_reduction=complete_reduction, algorithm=algorithm)
 end
 
+@doc raw"""
+    standard_basis_highest_corner(I::MPolyIdeal; ordering::MonomialOrdering = default_ordering(base_ring(I))) 
+
+Return a standard basis of `I` with respect to `ordering`. `ordering` needs to be local, the coefficient ring needs to be `QQ`.
+The algorithm first computes a standard basis over a finite field in order to get a highest corner fast. Then this highest corner
```
afkafkafk13: This is acutally only an upper bound for the highest corner w.r.t the monomial ordering, but this suffices to speed up the computation anyway. Maybe this should be stated in that way.

afkafkafk13 (approved):
The PR is fine.
Feel free to ignore my minor comment on the docstring.

src/Rings/groebner.jl
ederc: Yes, you are right. I updated the docstring correspondingly.

PR #3578, opened 2024-04-10
[FTheoryTools] Introduce optional argument generic for literature model constructor

HereAround:
@apturner A proposal to fix the book example by introducing an optional argument to the literature model constructor.

### DIFF
--- a/experimental/FTheoryTools/src/LiteratureModels/constructors.jl
+++ b/experimental/FTheoryTools/src/LiteratureModels/constructors.jl
@@ -32,6 +32,16 @@ Some literature models require additional parameters to specified to single out
 a model from a family of models. Such models can be provided using the optional
 argument `model_parameters`, which should be a dictionary such as `Dict("k" => 5)`.
 
+Notice that most F-theory models require the specification of sections of line
+bundles. Oftentimes, these sections appear in the singular loci of the models.
+Hence, generic sections also mean complicated singular loci. In case those loci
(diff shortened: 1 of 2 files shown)


experimental/FTheoryTools/src/LiteratureModels/constructors.jl
```diff
@@ -269,9 +269,11 @@ function _construct_literature_model_over_concrete_base(model_dict::Dict{String,
...
   for (key, value) in model_sections
     @req is_effective(toric_divisor_class(value)) "Encountered a non-effective model section"
-    explicit_model_sections[key] = generic_section(toric_line_bundle(value))
-    # Lead to error when computing singular loci - currently only monomials allowed...
-    #explicit_model_sections[key] = basis_of_global_sections(toric_line_bundle(value))[end]
+    if generic
+      explicit_model_sections[key] = generic_section(toric_line_bundle(value))
+    else
+      explicit_model_sections[key] = basis_of_global_sections(toric_line_bundle(value))[end]
+    end
```
HereAround, line 286: This is the key change/proposal to make the book example function again. The remainder is cosmetics, including a doc test that will fail if there is a new breaking change to said book example.

apturner (approved):
Thank you @HereAround, I think this makes sense and is the simplest fix for the problem we're in.

HereAround:
@lkastner This is the fix that is needed to make the book work again.

I believe this also needs a backport to 1.0.x.

PR #3580, opened 2024-04-10
Fix a `vcat`

joschmitt:
`vcat` shouldn't work for matrices over different base rings and here it does so just by accident.

(Also replaces some glueings by gluings.)


thofma:
This exploits a bug in the current implementation, which we want to fix.

lgoettgens:
Note for backporting: The relevant part is the `vcat`. The gluing may be dropped if creating conflicts or it doesn't exist on the release branch yet.

lgoettgens (approved):

PR #3584, opened 2024-04-10
`experimental/GModule`: collect AA/Nemo/Hecke stuff in `Misc.jl`

ThomasBreuer:
In `experimental/GModule`, move code to `Misc.jl` that belongs to AbstractAlgebra.jl, Nemo.jl, or Hecke.jl.

Once we agree that this code should be moved to these packages, the code should be added there,
and as soon as it becomes available in Oscar, it can be removed from `Misc.jl`.

### DIFF
--- a/experimental/GModule/Brueckner.jl
+++ b/experimental/GModule/Brueckner.jl
@@ -1,8 +1,6 @@
 module RepPc
 using Oscar
 
-export coimage
-
 #=TODO
  - construct characters along the way as well?
  - compare characters rather than the hom_base
@@ -265,10 +263,6 @@ function brueckner(mQ::Map{<:Oscar.GAPGroup, PcGroup}; primes::Vector=[])
   return allR
 end
 
-function coimage(h::Map)
-  return quo(domain(h), kernel(h)[1])
-end
-
 """
   mp: G ->> Q
   C a F_p[Q]-module
@@ -427,5 +421,3 @@ orbits(G)
 end #module RepPc
 
 using .RepPc
-
-export coimage
--- a/experimental/GModule/Brueckner.jl
+++ b/experimental/GModule/Brueckner.jl
@@ -3,17 +3,6 @@ using Oscar
 
 export coimage
 
-Base.pairs(M::MatElem) = Base.pairs(IndexCartesian(), M)
-Base.pairs(::IndexCartesian, M::MatElem) = Base.Iterators.Pairs(M, CartesianIndices(axes(M)))
-
-function Hecke.roots(a::FinFieldElem, i::Int)
-  kx, x = polynomial_ring(parent(a), cached = false)
-  return roots(x^i-a)
-end
-
-Oscar.matrix(phi::Generic.IdentityMap{<:AbstractAlgebra.FPModule}) = identity_matrix(base_ring(domain(phi)), dim(domain(phi)))
-
-
 #=TODO
  - construct characters along the way as well?
  - compare characters rather than the hom_base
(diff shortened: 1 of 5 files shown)


experimental/GModule/Brueckner.jl
```diff
@@ -276,69 +265,10 @@ function brueckner(mQ::Map{<:Oscar.GAPGroup, PcGroup}; primes::Vector=[])
...
-end
-
-function Oscar.dual(h::Map{<:AbstractAlgebra.FPModule{ZZRingElem}, <:AbstractAlgebra.FPModule{ZZRingElem}})
-  A = domain(h)
-  B = codomain(h)
-  @assert is_free(A) && is_free(B)
-  return hom(B, A, transpose(matrix(h)))
-end
-
 function coimage(h::Map)
```
fieker: Wh not coimage?

experimental/GModule/GModule.jl
```diff
@@ -299,10 +263,6 @@ function invariant_lattice_classes(M::GModule{<:Oscar.GAPGroup, <:AbstractAlgebr
   return invariant_lattice_classes(MZ)
 end
 
-function Oscar.pseudo_inv(h::Generic.ModuleHomomorphism)
```
fieker, line 302: For now, but eventually, this will have to be done differently as I'd like to have inv(inv()) to be the identity

experimental/GModule/GModule.jl
```diff
@@ -299,10 +263,6 @@ function invariant_lattice_classes(M::GModule{<:Oscar.GAPGroup, <:AbstractAlgebr
...
-  return MapFromFunc(codomain(h), domain(h), x->preimage(h, x))
-end
-
 function _hom(f::Map{<:AbstractAlgebra.FPModule{T}, <:AbstractAlgebra.FPModule{T}}) where T
```
fieker, line 266: is this still used? It feels like either it should move or be deleted

fieker (approved):
it adds up, doesn't it? No idea if this is now complete...

experimental/GModule/Brueckner.jl
ThomasBreuer: Yes, good idea.

experimental/GModule/GModule.jl
ThomasBreuer: `_hom` is used just once, a few lines more down, in order to abbreviate the code.
If we move it, we should call it `hom` (?), and document what it is expected to do:
Take a map, assume that this map is linear, and return the corresponding "linear-map-by-images".
(In the current situation, I think it would be better to do this directly, by mapping generators under the various maps, instead of first creating several intermediate maps; but this would look less elegant.)
In this sense, deleting `_hom` is perhaps a good idea.

ThomasBreuer:
> it adds up, doesn't it? No idea if this is now complete...

The problem is that the `is_complete` function is not applicable here.

fieker:
On Thu, Apr 11, 2024 at 07:50:24AM -0700, Thomas Breuer wrote:
> @ThomasBreuer commented on this pull request.
> 
> 
> 
> > @@ -299,10 +263,6 @@ function invariant_lattice_classes(M::GModule{<:Oscar.GAPGroup, <:AbstractAlgebr
>    return invariant_lattice_classes(MZ)
>  end
>  
> -function Oscar.pseudo_inv(h::Generic.ModuleHomomorphism)
> -  return MapFromFunc(codomain(h), domain(h), x->preimage(h, x))
> -end
> -
>  function _hom(f::Map{<:AbstractAlgebra.FPModule{T}, <:AbstractAlgebra.FPModule{T}}) where T
> 
> `_hom` is used just once, a few lines more down, in order to abbreviate the code.
> If we move it, we should call it `hom` (?), and document what it is expected to do:
> Take a map, assume that this map is linear, and return the corresponding "linear-map-by-images".
> (In the current situation, I think it would be better to do this directly, by mapping generators under the various maps, instead of first creating several intermediate maps; but this would look less elegant.)
> In this sense, deleting `_hom` is perhaps a good idea.
Then go ahead!

fingolfin:
Further changes e.g. to `_hom` can come in a follow-up PR

PR #3585, opened 2024-04-10
Temp fix: Show FTheoryTools docs again

lgoettgens:
This should get reverted once FTheoryTools gets upgraded to the experimental structure again.

A preview should be available soon-ish at https://docs.oscar-system.org/previews/PR3585

### DIFF
--- a/docs/make_work.jl
+++ b/docs/make_work.jl
@@ -132,6 +132,12 @@ function doit(
       append!(collected, pkgdocs)
     end
   end
+  if "FTheoryTools" in oldexppkgs # remove this block once FTheoryTools is upgraded to new experimental structure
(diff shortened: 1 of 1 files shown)


docs/make_work.jl
```diff
@@ -132,6 +132,12 @@ function doit(
       append!(collected, pkgdocs)
     end
   end
+  if "FTheoryTools" in oldexppkgs # remove this block once FTheoryTools is upgraded to new experimental structure
```
lgoettgens:
```suggestion
  if "FTheoryTools" in Oscar.oldexppkgs # remove this block once FTheoryTools is upgraded to new experimental structure
```

HereAround (approved):
Thank you very much @lgoettgens . Great!

PR #3589, opened 2024-04-11
internal consistency of `GModule` objects

ThomasBreuer:
- add an assertion in the constructor,
- fix the test where this assertion did not hold

### DIFF
--- a/experimental/GModule/Brueckner.jl
+++ b/experimental/GModule/Brueckner.jl
@@ -29,6 +29,10 @@ Note: the reps are NOT necessarily over the smallest field.
 
 Note: the field is NOT extended - but it throws an error if it was too small.
 
+Note: `group(M)` for the returned gmodules `M` will have a pcgs of `G` as
+      its `gens` value, thus these generators will in general differ from
+      the generators of `G`.
+
 Implements: Brueckner, Chap 1.2.3
 """
 function reps(K, G::Oscar.GAPGroup)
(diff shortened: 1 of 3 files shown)


fieker (commented):
here, for the code to work, G has to be a (mathematical) PC group. So once we have pc groups (as discussed elsewhere), the signature should change to reflect this?

fieker (approved):

ThomasBreuer:
@fieker 
> here, for the code to work, G has to be a (mathematical) PC group. So once we have pc groups (as discussed elsewhere), the signature should change to reflect this?

In this particular example, the group SL(2, 3) is a matrix group. For the purpose of computing representations, a polycyclic generating sequence of it is used. But still it is not a pc group but a (finite and solvable) matrix group.

The same holds for a solvable permutations group: Some algorithms compute a pcgs for it and then use this structure.
Julia's type system does not support this kind of information.

fieker:
On Thu, Apr 11, 2024 at 05:14:27AM -0700, Thomas Breuer wrote:
> @fieker 
> > here, for the code to work, G has to be a (mathematical) PC group. So once we have pc groups (as discussed elsewhere), the signature should change to reflect this?
> 
> In this particular example, the group SL(2, 3) is a matrix group. For the purpose of computing representations, a polycyclic generating sequence of it is used. But still it is not a pc group but a (finite and solvable) matrix group.
> 
> The same holds for a solvable permutations group: Some algorithms compute a pcgs for it and then use this structure.
> Julia's type system does not support this kind of information.
Indeed - but then for pcgroups, most of the time I cannot use gens...
This is part of a design decision that you and me and Max need to do
sometime

PR #3590, opened 2024-04-11
[ToricSchemes] Fixes to blow up of toric variety from new ray

HereAround:
As pointed out by @lkastner, we currently have the following bug:
```julia
julia> rs = [1 1; -1 1];

julia> max_cones = IncidenceMatrix([[1,2]]);

julia> v = normal_toric_variety(max_cones, rs);

julia> bu = blow_up(v, [0,1])
ERROR: ArgumentError: List of elements must be non-empty
Stacktrace:
 [1] macro expansion
   @ ~/.julia/packages/Hecke/AP42u/src/Assertions.jl:561 [inlined]
 [2] ideal(g::Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
   @ Oscar ~/oscars/master/src/Rings/mpoly-ideals.jl:56
 [3] blow_up(v::NormalToricVariety, new_ray::Vector{Int64}; coordinate_name::String)
   @ Oscar ~/oscars/master/experimental/Schemes/ToricBlowups/constructors.jl:101
 [4] blow_up(v::NormalToricVariety, new_ray::Vector{Int64})
   @ Oscar ~/oscars/master/experimental/Schemes/ToricBlowups/constructors.jl:92
 [5] top-level scope
   @ REPL[6]:1
```

This PR aims to fix this bug and make the corresponding blow up function more versatile.

cc @HechtiDerLachs


HereAround:
Summary of my proposal to fix this bug:
* If we cannot easily read-off the blow up center from the new ray, then we default to the generic schemes method.
* This methods return an `IdealSheaf` (rather than a `ToricIdealSheafFromCoxRingIdeal`). Hence, I had to extend the signature of (blow up) center for `ToricBlowdownMorphism`.
* I have tried to make the distinction clear by extending the doc string. The latter now also contains the example that triggered the bug.
* Minor related cleanups (remove functions that were commented out, consistency test).

lkastner (approved):
Thank you.

PR #3512, opened 2024-03-13
Test functions for regular solids

zkgeiselmann:
There are three types of regular 3-dimensional polytopes, that is, polytopes with regular polygons as faces. 
Platonic solids are vertex transitive and all faces have the same number of vertices. 
Archimedean solids are vertex transitive but can have different faces. 
Johnson solids are the remaining ones. 

In the main helper function we check for regularity by 
1.) checking that all edges have the same length
2.) checking that all vertices of a face lie on a circle. 

CC @alexej-jordan

PR #3591, opened 2024-04-11
PolyhedralGeometry: Format most of the files

lkastner:
I am only aware of #3512 and excluded the two files being worked on there.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Ran JuliaFormatter on most of src/PolyhedralGeometry
+ab444bbc46b5493588b2c3217ed35d86396d1260
--- a/etc/test_formatting.jl
+++ b/etc/test_formatting.jl
@@ -1,0 +1,104 @@
+using Pkg
+Pkg.activate(temp=true)
+Pkg.add("Test")
+Pkg.add("JuliaFormatter")
+using Test
+using JuliaFormatter
+
+# Disclaimer:
+# The goal should be to work with 
+# https://github.com/julia-actions/julia-format
+# . However, currently too many files have broken formatting and since there
+# are many ongoing pull requests, we will need to extend proper formatting to
+# the entire codebase in a step by step fashion.
+
+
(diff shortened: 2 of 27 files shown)


lgoettgens:
Thanks, @lkastner, for taking care of this!

Wouldn't it be nicer for the history if the new action and formatting of the PolyhedralGeometry folder were in different PRs? If you need some formatted files to test the action with, there is already the LieAlgebras stuff you found.

etc/test_formatting.jl
```diff
@@ -0,0 +1,104 @@
...
+          "experimental/LieAlgebras/src/LieSubalgebra.jl",
+          "experimental/LieAlgebras/src/LinearLieAlgebra.jl",
+          "experimental/LieAlgebras/src/RootSystem.jl",
+          "experimental/LieAlgebras/src/Util.jl",
+          "experimental/BasisLieHighestWeight/src/MainAlgorithm.jl",
+         ]
+
+  # Collect all code files.
+  entire = [
+            _gather_source_files(joinpath(oscardir, "src/"));
```
lgoettgens, line 67:
```suggestion
            _gather_source_files(joinpath(oscardir, "src/")); 
            _gather_source_files(joinpath(oscardir, "test/")); 
```

benlorenz:
> Wouldn't it be nicer for the history if the new action and formatting of the PolyhedralGeometry folder were in different PRs? If you need some formatted files to test the action with, there is already the LieAlgebras stuff you found.

If we make sure not to squash this PR then the commits stay separate, also with a proper merge instead of a rebase to make sure the commit for the blame-ignore-rev stays correct.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Ran JuliaFormatter on most of src/PolyhedralGeometry
+ab444bbc46b5493588b2c3217ed35d86396d1260
```
lgoettgens, line 2: This is now wrong due to the rebase

fingolfin (approved):

PR #3596, opened 2024-04-12
Use a new OSCAR email address on our domain

fingolfin:
The old address was that of the OSCAR mailing list, but IMHO the mathematik.uni-kl.de address had at least two issues:
1. its legacy these days as we were renamed; so at some point everything will migrate to a new address...
2. it is not reflective of the fact that OSCAR is developed by many people in many places.

So I've now set up a new address oscar@oscar-system.org -- for now this just forwards mails to @fieker and myself, though we can change that in the future as needed (up to having a "proper" mailing list backing it should we want to). But I don't think we'll get much traffic there.

CC @micjoswig @thofma @simonbrandhorst @wdecker

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Oscar"
 uuid = "f1435218-dba5-11e9-1e4d-f1a5fab5fc13"
-authors = ["The OSCAR Team <oscar@mathematik.uni-kl.de>"]
+authors = ["The OSCAR Team <oscar@oscar-system.org>"]
 version = "1.1.0-DEV"
 
 [deps]


benlorenz (approved):
I think this makes sense, maybe include more people in the alias but that is not really relevant for the PR.

I would prefer to have this merged soon, then I can still add this to 1.0.1. It doesn't seem anyone from the pinged people has an opinion on this?

thofma:
Fine with me.

micjoswig (approved):

lkastner (approved):

simonbrandhorst (approved):

fingolfin:
I am happy to add more people to which the email address forwards, just let me know. I did not want to put somebody on there without asking, as potentially there might be spam etc. (in reality, so far there is basically nothing arriving this route, but who knows, it might change in the future; at some point we could then also point it at a dedicated "proper" mailing list)

Issue #3450, opened 2024-02-28
the constructors which allow for caching need to be documented


micjoswig:
The caching mechanism is used extensively for polynomial rings and, more generally, objects/types from the Hecke/Nemo universe. 

I think it would be useful to have one extra page of documentation which just lists all constructors which support this feature.

The reason is that the behavior is sometimes unexpected to the uninitiated user, as it is not uniform across all types in OSCAR.

@fieker

thofma:
We can try to cook something up (after certain deadlines have passed), but I am not sure what to address exactly. Is the existence of caching surprising or the non-existence? I think it would be helpful if there is a victim of caching that I could talk to.

thofma:
Conclusion of the meeting: The request is to have a table of functions which support caching and some explanation.

JohnAAbbott:
I agree with @fieker that caching is a *convenience for interactive use,* but any code *we* write as part of OSCAR should be independent of caching (as regards *correctness*, and hopefully also *execution efficiency*).  A suggestion of @benlorenz to have a (developer only) global flag which disables all caching would facilitate verification that our code is caching-independent.
Might it be helpful to designate/document 3 categories of function:
* functions where caching will "always" be available
* functions where caching will "never" be available
* the rest

Is caching relevant only for constructors?  Maple's `option remember` can be applied to (almost?) any function.

fingolfin:
Perhaps to get started we can write code which iterates over all functions exported (?) by Oscar and check if they have a `cache` keyword argument (we kinda need this list for #3491, too). That may contain some false positives (and/or miss some functions) but at least it would be a start... 

Also this list is susceptible to becoming outdated over time, so if we could generate it automatically (perhaps by combining the above trick with a list of additional inclusions/exclusion) that would be best...?

Now we just need "someone"(TM) to work on it. Perhaps @aaruni96 can look into it (next week, after ICMS deadline)

aaruni96:
I've come up with a pipe of greps and seds from the Oscar.jl directory which produces 28 functions.

```bash
grep -d recurse "^export" src | sed "s/^.*export /\^/" | sed "s/, /\n/g" | sed "s/\(\^[0-9a-zA-Z_\!]*\).*/\1/"  > exports.txt
grep -n -d recurse -w "^\s*function" src | grep "cached::Bool" | sed "s/^.*function //" | sed "s/).*$/)/" | grep -f exports.txt -w
```


```
[...]

PR #3599, opened 2024-04-12
Add list of cached functions

aaruni96:
Addresses #3450


docs/src/DeveloperDocumentation/CachedFunctions.md
```diff
@@ -0,0 +1,323 @@
+# Functions Which Cache Things
+
+Many functions take in a keyword argument `cached::Bool`, which then cache or do not cache the results of computations to make things faster. This can sometimes lead to strange behaviour. So we document all the functions which do this, and at a future date, maybe do something with this list.
```
joschmitt: "cache or do not cache the results of computations to make things faster"
I don't think this is what is happening here.

To my understanding there are two reasons for caching in OSCAR:
* Functions which do complex computations and cache the result (like `groebner_basis` or `maximal_order`), so it does not need to be computed again. These functions usually do not have (maybe never have) a `cached` keyword.
* Constructors which add the constructed parent object to a global dictionary (= cache), so that if people construct the same polynomial ring or whatever twice, they get the identical ring and everything will *just work*. This is what the `cached` keyword is for. (So it is not about speed but convenience.)

Please correct me, if I am wrong here.

lgoettgens:
Can you adapt it so that the functions get put into code blocks (I.e. surrounded by backticks)? That should nice up the printing

docs/src/DeveloperDocumentation/CachedFunctions.md
aaruni96: You might be completely right. For now, I've removed that, and just added a "FIXME". Someone more qualified than me will have to figure out how to best phrase that paragraph.

joschmitt:
I added some description of the different ways 'caching' occurs in OSCAR (at least the ones I'm aware of), I hope you don't mind @aaruni96.
This is just reflecting my understanding of things; somebody should check this for correctness and also decide how much we want to document these things at all.

docs/src/DeveloperDocumentation/CachedFunctions.md
```diff
@@ -0,0 +1,367 @@
...
+(Univariate polynomial ring in x over QQ, x)
+
+julia> Qx === QQx
+true
+
+julia> x == xx
+true
+```
+This is mainly intended for interactive use where this behaviour may be convenient.
+However, the parent objects constructed in this way cannot be deleted (that is, garbage collected) as they are stored in the dictionary.
```
lgoettgens, line 43: I believe this is wrong, at least this is what I understand from the WeakValueDict docstring in AbstractAlgebra

> `WeakValueDict()` constructs a hash table where the values are weak
references to objects which may be garbage collected even when
used as values in a hash table.

joschmitt: This is how I understood it from the OSCAR book.
But yes, apparently garbage collection works:
```
julia> Qx, x = QQ[:x]

julia> Oscar.Nemo.FmpqPolyID
AbstractAlgebra.WeakValueDict{Symbol, QQPolyRing}(...):
  :x => Univariate polynomial ring in x over QQ

julia> Qx = 0
0

julia> x = 0
0

julia> GC.gc()

julia> Oscar.Nemo.FmpqPolyID
AbstractAlgebra.WeakValueDict{Symbol, QQPolyRing}()
```
Conclusion: I'm not qualified either to write this documentation.

docs/src/DeveloperDocumentation/CachedFunctions.md
```diff
@@ -0,0 +1,367 @@
...
+> - `matrix_space(R::ZZRing, r::Int, c::Int; cached::Bool = true)`
+> - `matrix_space(R::ZZModRing, r::Int, c::Int; cached::Bool = true)`
+> - `similar(f::PolyRingElem, R::ZZModRing, s::Symbol=var(parent(f)`
+> - `similar(f::PolyRingElem, R::ZZRing, s::Symbol=var(parent(f)`
+> - `matrix_space(R::FqField, r::Int, c::Int; cached::Bool = true)`
+> - `similar(f::PolyRingElem, R::FqField, s::Symbol=var(parent(f)`
+> - `matrix_space(R::FqPolyRepField, r::Int, c::Int; cached::Bool = true)`
+> - `matrix_space(R::fqPolyRepField, r::Int, c::Int; cached::Bool = true)`
+> - `similar(f::PolyRingElem, R::fqPolyRepField, s::Symbol=var(parent(f)`
+> - `similar(f::PolyRingElem, R::FqPolyRepField, s::Symbol=var(parent(f)`
```
joschmitt, line 250: @aaruni96 your magical pipe seems to cut off some function signatures like here. (Just saying. As long as we don't know what we *do* with this list, there is probably no point in debugging this.)

aaruni96: I see, it cuts off at the first closing parenthesis. Its probably worth fixing, as it might be confusing later as to why something without `cached` seems to appear in the list.

I'll change the `sed` pattern.

fingolfin:
So what should we do with this list? Put it somewhere? But it will get outdated quickly. Or put a program to generate the list somewhere?

Or add another section to the "General" section of the OSCAR manual? Maybe in https://docs.oscar-system.org/dev/General/faq/ or a new section "Things to be aware of when using OSCAR" (which then could also mention we have our own matrix and integer types, which is right now in the FAQ)

@micjoswig since you asked for this, would be good to hear from you what you'd prefer

fingolfin:
I don't know how to move forward with this, if at all, so closing this for now.

@micjoswig if you disagree, of course feel free to re-open, but then please indicate what it is you wish us to achieve here...

PR #3600, opened 2024-04-12
README.md: update citation info once more

fingolfin:
... with info we just got from the publisher

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -31,6 +31,7 @@ preferred-citation:
     given-names: "Max"
   - family-names: "Joswig"
     given-names: "Michael"
-  title: The Computer Algebra System OSCAR. Algorithms and Examples
+  title: "The Computer Algebra System OSCAR: Algorithms and Examples"
   year: 2024
-
+  publisher:
+    name: "Springer"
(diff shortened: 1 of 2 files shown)


joschmitt:
There is also this `CITATION.CFF` file which produces its own bibtex entry. I have a commit https://github.com/joschmitt/Oscar.jl/commit/9ad1f982ed020ef49c06c113b1d67fc12d6ef310 that updates that, but there is no field `series` so I put the series name into `collection-title`. In the automatically produced bibtex, the series name is not shown (as you can see if you click "Cite this repository" on the `js/cite` branch of my fork). So I am not sure whether this (i.e. two different bibtex blocks) is what we want.

fingolfin:
Thanks I'll update the CFF. If CFF does not support features like series, I don't think that should block us from using the correct bibtex.

fingolfin:
See https://github.com/citation-file-format/citation-file-format/issues/394

fingolfin:
Update the CFF file (and also the exact text in the README.md). I decided against using `collection-title` in the CFF as it has different semantics.

joschmitt (approved):

fingolfin:
Thanks for carefully checking this, @joschmitt . I hope I'll remember the CFF file next time, but in the meantime I am very grateful you are paying close attention :-)

PR #3601, opened 2024-04-12
Implement normalization for non-reduced schemes

paemurru:
This is just removing the check for non-reducedness. Also added a test for normalization of a non-reduced scheme.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Methods.jl
@@ -157,13 +157,11 @@ function base_change(phi::Any, X::AbsCoveredScheme)
 end
 
 @doc raw"""
-    normalization(X::AbsCoveredScheme; check::Bool=true) -> (AbsCoveredScheme, AbsCoveredSchemeMor, Vector{<:AbsCoveredSchemeMor})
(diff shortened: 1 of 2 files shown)


simonbrandhorst:
Are you sure that things will work? The singular manual says:

```
The ideal must be radical, for non-radical ideals the output may be wrong (id=radical(id); makes id radical). However, when using the "prim" option the minimal associated primes of id are computed first and hence normal computes the normalization of the radical of id.
```
https://www.singular.uni-kl.de/Manual/latest/sing_1464.htm#SEC1545

wdecker:
As already pointed out by @simonbrandhorst, this change does not make sense. Of course, the normalization of a non-reduced scheme can be defined under some assumptions and is, in fact, nothing else but the normalization of the reduction of the given scheme. See https://stacks.math.columbia.edu/tag/035E for a precise statement. But why do we need this? In any case, just dropping the assumption as you do may lead to errors if the default option `equidimDec` is used.

paemurru:
@simonbrandhorst OK, you are correct.

I had assumed that `Oscar.maximal_associated_points` gives the maximal associated points and taking `scheme` gives their natural reduced structure. But actually `Oscar.maximal_associated_points` returns an ideal, remembering the non-reduced structure.

In this case, I agree that it is not worth adding normalization for non-reduced schemes.

PR #3603, opened 2024-04-12
`GModule`: replace `_hom`

ThomasBreuer:
Create the images of generators directly, instead of first creating pseudo-inverses and compositions of maps.
(Eventually, restricting the group action to a submodule should become a function of its own.)

(also fixed a call of `maximal_submodule_bases`)

### DIFF
--- a/experimental/GModule/GModule.jl
+++ b/experimental/GModule/GModule.jl
@@ -263,11 +263,6 @@ function invariant_lattice_classes(M::GModule{<:Oscar.GAPGroup, <:AbstractAlgebr
   return invariant_lattice_classes(MZ)
 end
 
-function _hom(f::Map{<:AbstractAlgebra.FPModule{T}, <:AbstractAlgebra.FPModule{T}}) where T
-  @assert base_ring(domain(f)) == base_ring(codomain(f))
-  return hom(domain(f), codomain(f), f.(gens(domain(f))))
-end
-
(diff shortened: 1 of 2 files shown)


fieker:
On Fri, Apr 12, 2024 at 05:22:21AM -0700, Thomas Breuer wrote:
> Create the images of generators directly, instead of first creating pseudo-inverses and compositions of maps.
> (Eventually, restricting the group action to a submodule should become a function of its own.)
> 
I think sub(GModule, module or similar) already exists...

ThomasBreuer:
O.k., `sub(M::GModule, phi::Map)` shall take an embedding with codomain `M` and shall return `S, emb` such that `S` is a `GModule` for the same group as `M` but with module `S`.
(I have not found this, will add it.)

fieker:
On Fri, Apr 12, 2024 at 05:47:06AM -0700, Thomas Breuer wrote:
> O.k., `sub(M::GModule, phi::Map)` shall take an embedding with codomain `M` and shall return `S, emb` such that `S` is a `GModule` for the same group as `M` but with module `S`.
> (I have not found this, will add it.)
> 
I thought you even fixed a bug in it...

ThomasBreuer:
There is `Oscar.sub(C::GModule{<:Any, <:AbstractAlgebra.FPModule{T}}, m::MatElem{T}) where {T <: FinFieldElem}` but this is a different situation.

fingolfin (approved):

PR #3604, opened 2024-04-12
booktests: move one currently broken tests to skipped, remove another

benlorenz:
to avoid failures when these are fixed upstream

### DIFF
--- a/test/book/test.jl
+++ b/test/book/test.jl
@@ -21,9 +21,6 @@ isdefined(Main, :FakeTerminals) || include(joinpath(pkgdir(REPL),"test","FakeTer
              # Something broken, probably needs some updated GAP packages
              "specialized/breuer-nebe-parker-orthogonal-discriminants/expl_syl.jlcon",
 
-             # needs new hecke release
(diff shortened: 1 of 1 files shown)


test/book/test.jl
```diff
@@ -9,21 +9,23 @@ isdefined(Main, :FakeTerminals) || include(joinpath(pkgdir(REPL),"test","FakeTer
...
-
-            #  broken with recent changes ... to be fixed on master soon
-            "cornerstones/number-theory/galoismod.jlcon",
            ]
   skipped = [
+             # Something broken, probably needs some updated GAP packages
+             "specialized/breuer-nebe-parker-orthogonal-discriminants/expl_syl.jlcon",
+
+             # needs new hecke release
+             "cornerstones/number-theory/galoismod.jlcon",
```
thofma: I just tagged a new Hecke version (https://github.com/JuliaRegistries/General/pull/104799), which should make this unncessary.

benlorenz:
```suggestion
```

PR #1270, opened 2022-04-20
small adjustment docu `isnormal`

wdecker:

PR #3605, opened 2024-04-12
Function is_normal for covered and affine schemes

paemurru:
Also fixed a bug in `is_normal` for quotient rings: now it checks that the ring is reduced. This bug is still present in Singular ( https://www.singular.uni-kl.de/hannes/sing_1485.htm#SEC1566 ).

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
@@ -175,6 +175,31 @@ end
 # Normalization                                                        #
 ########################################################################
 
+@doc raw"""
+    is_normal(X::AbsAffineScheme; check::Bool=true) -> Bool
+
+# Input:
+- a reduced scheme ``X``,
+- if `check` is `true`, then confirm that ``X`` is reduced; this is expensive.
+
+# Output:
+Returns whether the scheme ``X`` is normal.
+
+# Examples
+```jldoctest
+julia> R, (x, y, z) = rational_field()["x", "y", "z"];
+
+julia> X = spec(R);
+
+julia> is_normal(X)
+```
+"""
+function is_normal(X::AbsAffineScheme; check::Bool=true)
+  !check || is_reduced(X) || return false
+  R = coordinate_ring(X)
+  return is_normal(R)
+end
+
 # Todo: Normalizations of localizations
 #function _normalization(X::AbsAffineScheme{<:Field, <:MPolyLocRing}; algorithm=:equidimDec)
 #  L = OO(X)
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
@@ -198,7 +198,7 @@ true
 function is_normal(X::AbsAffineScheme; check::Bool=true)
   !check || is_reduced(X) || return false
   R = coordinate_ring(X)
-  return is_normal(R)
+  return is_normal(R; check=check)
 end
 
 # Todo: Normalizations of localizations
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
@@ -192,6 +192,7 @@ julia> R, (x, y, z) = rational_field()["x", "y", "z"];
 julia> X = spec(R);
 
 julia> is_normal(X)
+true
 ```
 """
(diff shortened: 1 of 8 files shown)


src/Rings/mpoly-affine-algebras.jl
```diff
@@ -931,25 +941,29 @@ julia> is_normal(A)
...
+function is_normal(A::MPolyQuoRing; check::Bool=true)
+  return get_attribute!(A, :is_normal) do
+    @req coefficient_ring(A) isa AbstractAlgebra.Field "Only implemented if coefficient ring is a field"
+    @req is_perfect(coefficient_ring(A)) "Only implemented if coefficient ring is a perfect field"
+    @req !(base_ring(A) isa MPolyDecRing) "Not implemented for quotients of decorated rings"
+    !check || is_reduced(A) || return false
+
+    I = A.I
+    # TODO remove old1 & old2 once new Singular jll is out
+    # QUESTION: How old is this? Has this been fixed yet?
```
afkafkafk13, line 953: @hannes14 : Do you have an idea whether this has already propagated up or not?
                         This would be the moment to get rid of the workaround, if it is not needed any more.

test/Rings/mpoly_affine_algebras.jl
```diff
@@ -37,8 +37,9 @@
     end
   end
 
-  @test !is_normal(quo(R, ideal(R, [x^2 - y^3]))[1])
-  @test is_normal(quo(R, ideal(R, [x - y^3]))[1])
+  @test !is_normal(quo(R, ideal(R, [x^2 - y^3]))[1]; check=false)
+  @test is_normal(quo(R, ideal(R, [x - y^3]))[1]; check=false)
+  @test !is_normal(quo(R, ideal(R, [x^2]))[1])
```
afkafkafk13, line 42: Please add [z^2-x*y] to the test cases.

afkafkafk13:
This looks fine. The only reason, why I have not approved yet, is the question about the old bugfix.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
```diff
@@ -175,6 +175,31 @@ end
...
+Returns whether the scheme ``X`` is normal.
+
+# Examples
+```jldoctest
+julia> R, (x, y, z) = rational_field()["x", "y", "z"];
+
+julia> X = spec(R);
+
+julia> is_normal(X)
+```
```
afkafkafk13: A line of output seems to be missing here.

test/Rings/mpoly_affine_algebras.jl
paemurru: Added.

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Methods.jl
paemurru: Fixed.

afkafkafk13 (approved):
Thank you. The code looks good.

hannes14:
Last important change top normal.lib was  Wed Nov 8 2023 (and has nothing to do with options).
What was the bug this work-around was for?
There were fixes for handling computation in quotient rings, but this seems to be something different.

afkafkafk13:
> Last important change top normal.lib was Wed Nov 8 2023 (and has nothing to do with options). What was the bug this work-around was for? There were fixes for handling computation in quotient rings, but this seems to be something different.

I am not even sure that it was a bug in the first place. It looks rather like a performance hack to make the Singular computation faster by not requiring reduction. What struck me as odd were the comments. Leaving this bit of code in the function is not a problem per se, but leaving it hanging around with a TODO-comment and a question was rather unclear. Either the hack can be removed or the comments should be removed to make this part of the code look less experimental (in the stable part of the tree!).

benlorenz:
I think this was done here [`f290fec` (#1270)](https://github.com/oscar-system/Oscar.jl/pull/1270/commits/f290feca4473bb30a8f09e475dbf3beea1c06905) to make sure that calling this singular function doesn't change the output for code that runs later. But it seems that was fixed in Singular in https://github.com/Singular/Singular/commit/74622283ad46295f0602f84dc752ccf62f699461.

afkafkafk13:
> I think this was done here [`f290fec` (#1270)](https://github.com/oscar-system/Oscar.jl/pull/1270/commits/f290feca4473bb30a8f09e475dbf3beea1c06905) to make sure that calling this singular function doesn't change the output for code that runs later. But it seems that was fixed in Singular in [Singular/Singular@7462228](https://github.com/Singular/Singular/commit/74622283ad46295f0602f84dc752ccf62f699461).

Indeed! This is the history which explains the mysterious comments -- and obsoletes them (along with the option-setting code).  Thank you.

PR #3606, opened 2024-04-12
Matroids: add constructors for algebraic matroids

taboege:
Add `matroid_from_prime_ideal` and `algebraic_matroid` which create a `Matroid` from a prime ideal in a multivariate polynomial ring or a list of elements in an affine algebra (or a fraction field of one), respectively.

*Note:* that most of the added code in `algebraic_matroid` is used to find the algebraic relations among elements in a fraction field of (a quotient of) a polynomial ring. Neither `is_algebraically_dependent_with_relations` nor `kernel` of a homomorphism currently seem to support such fraction fields.

### DIFF
--- a/src/Combinatorics/Matroids/matroids.jl
+++ b/src/Combinatorics/Matroids/matroids.jl
@@ -387,6 +387,118 @@ See `bond_matroid`.
 """
 cocycle_matroid(g::Graph) = bond_matroid(g::Graph)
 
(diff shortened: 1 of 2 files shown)


lkastner (approved):
Two comments:
- I think it would make sense to have this functionality in a separate file `Combinatorics/Matroids/algebraic_matroids.jl`
- Please also add some tests in `test/Combinatorics/Matroids`. Not every testset runs the doctests.

PR #3610, opened 2024-04-15
Posets in OSCAR

felix-roehrich:
This is an initial implementation for posets in OSCAR, feedback is welcome.

### DIFF
--- a/experimental/Posets/src/Poset.jl
+++ b/experimental/Posets/src/Poset.jl
@@ -12,14 +12,14 @@ function Base.length(P::Poset)
 end
 
 function Base.show(io::IO, P::Poset)
-  print(io, "Poset with $(ItemQuantity(length(P), "element")")
+  print(io, "Poset with $(ItemQuantity(length(P), "element"))")
 end
 
 # PosetElem
 
 struct PosetElem
-  i::Int
   parent::Poset
+  i::Int
 end
 
 function index(x::PosetElem)
@@ -58,6 +58,7 @@ function poset(cov::Matrix{Int}, elems::Vector{<:VarName}=["x_$i" for i in 1:nco
   @req is_upper_triangular(cov, 1) "matrix must be strictly upper triangular"
   @req nrows(cov) == ncols(cov) "must be a square matrix"
   @req ncols(cov) == length(elems) "size of matrix must match number of elements"
+  @req allunique(elems) "element names must be unique"
 
   d = nrows(cov)
   rel = BitMatrix(!iszero(cov[i, j]) for i in 1:d, j in 1:d)
@@ -68,7 +69,7 @@ end
 
 function (P::Poset)(i::Int)
   @req 1 <= i <= length(P.elems) "index out of range"
-  return PosetElem(i, P)
+  return PosetElem(P, i)
 end
 
 function (P::Poset)(elem::VarName)
@@ -77,7 +78,7 @@ function (P::Poset)(elem::VarName)
     error("unknown element")
   end
 
-  return PosetElem(i, P)
+  return PosetElem(P, i)
 end
 
 # MaximalChainsIterator constructors
--- a/experimental/Posets/src/Poset.jl
+++ b/experimental/Posets/src/Poset.jl
@@ -12,7 +12,7 @@ function Base.length(P::Poset)
 end
 
 function Base.show(io::IO, P::Poset)
-  print(io, "poset with $(length(P)) elements")
+  print(io, "Poset with $(ItemQuantity(length(P), "element")")
 end
 
 # PosetElem
@@ -77,7 +77,7 @@ function (P::Poset)(elem::VarName)
     error("unknown element")
   end
 
-  return Poset(i, P)
+  return PosetElem(i, P)
 end
 
 # MaximalChainsIterator constructors
--- a/experimental/Posets/src/Poset.jl
+++ b/experimental/Posets/src/Poset.jl
@@ -47,6 +47,13 @@ end
 
 # Poset constructors
 
+@doc raw"""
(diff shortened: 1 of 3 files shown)


experimental/Posets/src/Poset.jl
```diff
@@ -0,0 +1,199 @@
...
+  set::BitMatrix # indicates whether a general relation was computed
+  elems::Vector{Symbol} # symbols to use for the elements
+end
+
+function Base.length(P::Poset)
+  return length(P.elems)
+end
+
+function Base.show(io::IO, P::Poset)
+  print(io, "poset with $(length(P)) elements")
```
lgoettgens:
```suggestion
  print(io, "Poset with $(ItemQuantity(length(P), "element")")
```

experimental/Posets/src/Poset.jl
```diff
@@ -0,0 +1,199 @@
...
+end
+
+# PosetElem
+
+struct PosetElem
+  i::Int
+  parent::Poset
+end
+
+function index(x::PosetElem)
```
lgoettgens, line 25: The already existing `index` methods do something completely different. Does somebody know of other occurrences in the Oscar world of something like this and how we call it there?

experimental/Posets/src/Poset.jl
```diff
@@ -0,0 +1,199 @@
...
+  return PosetElem(i, P)
+end
+
+function (P::Poset)(elem::VarName)
+  i = findfirst(==(Symbol(elem)), P.elems)
+  if isnothing(i)
+    error("unknown element")
+  end
+
+  return Poset(i, P)
```
lgoettgens:
```suggestion
  return PosetElem(i, P)
```
typo?

experimental/Posets/src/Poset.jl
```diff
@@ -0,0 +1,199 @@
...
+
+function Base.show(io::IO, P::Poset)
+  print(io, "poset with $(length(P)) elements")
+end
+
+# PosetElem
+
+struct PosetElem
+  i::Int
+  parent::Poset
```
lgoettgens:
```suggestion
  parent::Poset
  i::Int
```
I think in most structs we have the parent as the first field. But no strong preference here.

experimental/Posets/src/Poset.jl
```diff
@@ -0,0 +1,199 @@
...
+            return ps.rel[x.i, y.i]
+          else
+            # k is not comparable to y
+            continue
+          end
+        end
+
+        # add k to the stack and continue from k
+        push!(q, k)
+        @goto outer
```
lgoettgens, line 152: I think you can just replace this goto by a `break` and then get rid of the `@label` as well.

lgoettgens (commented):
I have some minor remarks. But what I think is way more important is to get feedback from people that actually want to use Posets in Oscar.

experimental/Posets/src/Poset.jl
felix-roehrich: If I replace this with break, I will have to introduce a variable and an if statement after for loop. In the end I felt that goto is easier and the flow is clear too.

fieker:
This is not an initial implementation of posets in OSCAR, there is one in
src/NumberTheory/GaloisGrp/POSet.jl 
and has been there for a long time. Is the new one supposed to replace the existing one? Can it cover everything (which is not a lot) that the old one does?
The Posets were supposed to be updated/ replaced eventually, so feel free to do so, but please have a look and see what is required in this particular application.

experimental/Posets/src/Poset.jl
fieker: its mostly called `data` to access the defining datum...

felix-roehrich:
I didn't know there was an existing implementation, since it isn't exported; thanks for letting me know. The goals pursued in GaloisGrp are different from my implementation. For my application I am mostly interested in maximal chains, so I need covering relations and the edges have bonds/weights. Currently the constructor uses (and only needs) the adjacency matrix of the Hasse diagram.

@lgoettgens suggested to bring this up in the next OSCAR meeting, so other people (who need posets) can voice their needs, and then plan from there.

micjoswig:
There is a new polymake release in the making.  One of its main features is substantial new functionality concerning posets. For instance, a fairly fast implementation of all maximal chains etc.

We need to align our efforts.

fingolfin:
@micjoswig is there already some kind of preliminary documentation on this somewhere, or a person to talk to about it, so that @felix-roehrich can coordinate with them?

micjoswig:
Let's discuss posets on Friday (26 April) in the regular zoom.

micjoswig:
We will see the release of poylmake 4.12 soon (say, by May 3).  That features some additions to (and renaming of)  polymake's posets (which are still called lattice in 4.11 and before).  It will trickle down through Polymake.jl to OSCAR.

In case you want to see this in action, just as an example, here is the [polymake 4.11 code](https://github.com/polymake/polymake/blob/Releases/apps/polytope/src/order_polytope.cc) for computing an arbitrary order polytope. One nontrivial ingredient is how the maximal anti-chains are computed; this goes through a [max clique computation](https://github.com/polymake/polymake/blob/Releases/apps/graph/include/max_cliques.h) in the comparability graph.

See also Section 5 of [our preprint](https://arxiv.org/abs/2309.01626).

fingolfin:
Polymake 4.12 has been released now; but integration into OSCAR is still work in progress, see PR #3819.

fingolfin:
Status seems to be: There is still ongoing work regarding polymake integration in PR #3928

micjoswig:
> Status seems to be: There is still ongoing work regarding polymake integration in PR #3928

Indeed; thanks for the reminder.  This required a small discussion within the polymake team.  Now completed; design decisions have been made.

@benlorenz will implement the (few) missing bits soon.

PR #3617, opened 2024-04-17
improve `isomorphism(FPGroup, G, on_gens = true)`

ThomasBreuer:
in the situation that the generators of G form a pcgs, for example if `gens(G)` is the tail of the defining pcgs of the full pc group that contains `G`.
@fieker had noticed that one runs into big unnecessary computations in this situation.

(Perhaps this situation should be supported also in GAP.)

### DIFF
--- a/src/Groups/homomorphisms.jl
+++ b/src/Groups/homomorphisms.jl
@@ -547,6 +547,11 @@ function isomorphism(::Type{FPGroup}, G::GAPGroup; on_gens::Bool=false)
          if GAP.Globals.HasFamilyPcgs(G.X)
            pcgs = GAP.Globals.InducedPcgsWrtFamilyPcgs(G.X)
            if pcgs == Ggens
+             # `pcgs` fits *and* is an object in `GAP.Globals.IsPcgs`,
+             # for which a special `GAPWrap.IsomorphismFpGroupByGenerators`
+             # method is applicable.
+             # (Currently the alternative is a cokernel computation.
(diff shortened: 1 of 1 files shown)


src/Groups/homomorphisms.jl
```diff
@@ -542,6 +542,14 @@ function isomorphism(::Type{FPGroup}, G::GAPGroup; on_gens::Bool=false)
          f = GAP.Globals.GroupHomomorphismByImages(G.X, GAP.Globals.FreeGroup(0), GAP.Obj([]), GAP.Obj([]))
          GAP.Globals.SetIsBijective(f, true)
        else
+         # The computations are easy if `Ggens` is a pcgs,
+         # otherwise GAP will call `CoKernel`.
+         if GAP.Globals.HasFamilyPcgs(G.X)
+           pcgs = GAP.Globals.InducedPcgsWrtFamilyPcgs(G.X)
+           if pcgs == Ggens
+             Ggens = pcgs
```
fingolfin: I don't understand: if they are equal, why does this have an effect?


Or is one a pcgs object, the other a plain list, and they happen to compare as equal? If so, this is a subtle point and I think a comment should explain it?

ThomasBreuer: Yes, `pcgs` is an object in the filter `IsPcgs`, whereas `Ggens` is a plain list.

fingolfin (approved):

PR #3619, opened 2024-04-17
Add dev docs on caching parent objects

### LINKED ISSUE #3616: Add a "caching policy" section to OSCAR dev docs
... based on what we just discussed... 

to be written this afternoon

fingolfin:
Following the discussions last Friday and again this Wednesday, where Claus explained the caching rules as they were so far; I've now tried to codify them here, for future reference or revision. In particular, what I describe here is not meant to be something new, but reflect what was the policy *so far*.

I am sure I forgot stuff or maybe even misrepresented something, so please let the feedback flow in.

Resolves #3616

### DIFF
--- a/docs/src/DeveloperDocumentation/caching.md
+++ b/docs/src/DeveloperDocumentation/caching.md
@@ -1,0 +1,105 @@
+```@meta
+CurrentModule = Oscar
+```
+
+# Caching parent objects in OSCAR
+
+Many functions in OSCAR that construct parent objects (such as rings, modules,
+groups, etc.) have an optional keyword argument `cached::Bool`. If set to
+`true` then the object is put into a cache, and when the construction function
+is later called again with identical inputs, then the cached object is
+returned instead of creating a new object. In contrast when `cached` is set to
+`false` then each time a new object is returned.
+
+Example:
+```jldoctest
+julia> R1, = polynomial_ring(QQ, :x; cached = true);
+
+julia> R2, = polynomial_ring(QQ, :x; cached = true);
+
+julia> R1 === R2  # identical as both were created with `cached = true`
+true
+
+julia> R3, = polynomial_ring(QQ, :x; cached = false);
+
+julia> R1 === R3  # not identical as R3 was created with `cached = false`
+false
+
+julia> R4, = polynomial_ring(QQ, :y; cached = true);
+
+julia> R1 === R4  # not identical despite `cached = true` due to differing variable names
+false
+```
+
+## Why cache parent objects?
+
+The main reason for supporting caching of parent objects is **user convenience**:
+experience shows that most mathematicians (espescially those who are not also
+programmers; but it really affects all) are surprised if, say, `QQ[:x] == Q[:x]`
+produces `false`.
+
+For interactive use, it is often simply convenient: e.g. in the following example,
+we use `map_coefficients` to map polynomials over the integers to polynomials
+over a finite field, and the results can be added -- this is only possible because
+the new polynomials have the same parent, thanks to caching.
+```jldoctest
+julia> Zx, x = ZZ["x"]
+(Univariate polynomial ring in x over ZZ, x)
+
+julia> F = GF(2);
+
+julia> map_coefficients(F, x^2) + map_coefficients(F, x)
+x^2 + x
+```
+
+Caching parents also has downsides. E.g. all those cached objects take up memory which
+in some cases can add up to significant amounts.
+
+
+## Rules for implementations
+
+In the following we describe some rules related to caching for people implementing
+parent constructor functions
+
+1. Don't use caching in code inside OSCAR (caching is for end users!)
+   - i.e., code inside OSCAR by default should always construct rings with `cached = false`.
+   - In other words: internal code should not rely on caching being active.
+     Usually the need for using cached parents can be overcome by allowing callers to
+     pass in a parent object as an additional function argument. One may still provide a
+     default value for that as a user convenience, but these default parents then should
+     be created with `cached=false`.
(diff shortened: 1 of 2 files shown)


fingolfin:
I just discovered some notes on caching at https://nemocas.github.io/AbstractAlgebra.jl/dev/ring_interface/#Parent-object-caches -- perhaps some of that should also be included or referenced here (and/or perhaps the text there should be updated). I'll wait to hear what @fieker thinks.

docs/src/DeveloperDocumentation/caching.md
```diff
@@ -0,0 +1,95 @@
...
+## Why cache parent objects?
+
+The main reason for supporting caching of parent objects is **user convenience**:
+experience shows that most mathematicians (espescially those who are not also
+programmers; but it really affects all) are surprised if, say, `QQ[:x] == Q[:x]`
+produces `false`.
+
+For interactive use, it is often simply convenient: e.g. taking the characteristic
+polynomials of two matrices and multiplying them, without worrying about specifying
+the parent rings of the polynomials.
```
joschmitt: Funnily enough, exactly this example doesn't work because caching is deliberately turned off:
```
julia> M = QQ[1 0; 0 1];

julia> charpoly(M)*charpoly(M)
ERROR: Incompatible polynomial rings in polynomial operation
[...]
```

docs/src/DeveloperDocumentation/caching.md
```diff
@@ -0,0 +1,95 @@
...
+
+Caching parents also has downsides. E.g. all those cached objects take up memory which
+in some cases can add up to significant amounts.
```
joschmitt, line 56: Tommy also mentioned issues with parallelization/thread safety as a (possible) downside to me.

(And I don't really understand how the memory could fill up if we use these `WeakValueDict`s? But that's more for my curiosity.)

docs/src/DeveloperDocumentation/caching.md
```diff
@@ -0,0 +1,95 @@
...
+   with a default value, regardless of whether caching is actually supported or not.
+   - if caching is supported, then `cached` should default to *true*
+   - if caching is not supported, then `cached` should default to *false*
+   - Rationale: this allows us to comply pro-actively with the first rule: when creating
+     a parent object, you always pass in `cached = false`. If not all constructors
+     support this, we can't comply with it. Even if a constructor does not support
+     caching right now: this might change in the future. So by allowing the `cached`
+     argument in all cases, we can write future-proof code.
+3. Caches must not overflow
+   - the simples solution to achieve this is to use an `AbstractAlgebra.CacheDictType`
```
joschmitt:
```suggestion
   - the simplest solution to achieve this is to use an `AbstractAlgebra.CacheDictType`
```

docs/src/DeveloperDocumentation/caching.md
```diff
@@ -0,0 +1,95 @@
...
+     instances (which really is an alias for `WeakValueDict`) together with `get_cached!`
+     which automatically removes objects from caches if nothing outside the cache references
+     it anymore
+   - Alternatively one may offer a manual way for users to "flush" caches, But beware
```
lgoettgens:
```suggestion
   - Alternatively one may offer a manual way for users to "flush" caches, but beware
```

joschmitt (approved):
This matches what I was told over the past years. It's good to have a 'reference' now.

fieker:
Maybe add `Hecke.clear_cache()` which tries its best to empty all(?) caches in AA, Nemo and Hecke?

docs/src/DeveloperDocumentation/caching.md
fingolfin: Ouch! Thank you very much for fact checking this.

That makes me question the point of entire caching thing even more *sigh*.

So, does anyone then have a not completely unconvincing example for a situation where a cached parent helps a "naive" user?

docs/src/DeveloperDocumentation/caching.md
fingolfin: Unfortunately not all our caches actually use `WeakValueDict`. We introduced it much later and e.g. Hecke still uses Dict and IdDict caches quite a bit (I just grepped for `get_cached!` to find examples for caches and then checked their definitions). Might be a good idea for somebody to clean that up.

thofma:
```
julia> Zx, x = ZZ["x"]
(Univariate polynomial ring in x over ZZ, x)

julia> F = GF(2);

julia> map_coefficients(F, x^2) + map_coefficients(F, x)
x^2 + x
```

docs/src/DeveloperDocumentation/caching.md
joschmitt: I hope that I got all caching dictionaries which weren't `WeakValueDict`s in https://github.com/Nemocas/Nemo.jl/pull/1724 and https://github.com/thofma/Hecke.jl/pull/1460.

fieker (approved):
it moves into the right direction.

fingolfin:
Thanks @thofma I've added your nice example.

BTW I find it quite disheartening that I get this on my M1 MacBook the first time I execute `map_entries` (that's not direct against Tommy or against anyone, really, just something that makes me sad / where we badly need to improve, somehow)
```
julia> Zx, x = ZZ["x"]
(Univariate polynomial ring in x over ZZ, x)

julia> F = GF(2);

julia> @time map_coefficients(F, x^2);
 11.979656 seconds (60.42 M allocations: 3.764 GiB, 7.85% gc time, 100.00% compilation time)

julia> @time map_coefficients(F, x^2);
  0.000021 seconds (18 allocations: 976 bytes)
```

PR #3624, opened 2024-04-18
Workaround for https://github.com/oscar-system/Singular.jl/issues/796

RafaelDavidMohr:
This is a workaround if we pass the zero module to Singular in `prune_with_map`. In this case the Singular function `prune_with_map_projection` has inconsistent output with other modules, causing a bug.

### DIFF
--- a/src/Modules/UngradedModules/Presentation.jl
+++ b/src/Modules/UngradedModules/Presentation.jl
@@ -522,7 +522,8 @@ function prune_with_map(M::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}} # 
   pres_map = map(pm, 1)
   krnel = SubquoModule(pm[0], pres_map.(gens(pm[1]))) # Create the image of pres_map without the inclusion map
 
-  # work around if we're passing the zero module to singular
+  # work around if we're passing the zero module to singular, see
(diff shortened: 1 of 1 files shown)


src/Modules/UngradedModules/Presentation.jl
```diff
@@ -521,6 +521,12 @@ function prune_with_map(M::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}} #
   pm = presentation(M)
   pres_map = map(pm, 1)
   krnel = SubquoModule(pm[0], pres_map.(gens(pm[1]))) # Create the image of pres_map without the inclusion map
+
+  # work around if we're passing the zero module to singular
```
joschmitt: Maybe add the link to the Singular issue in the comment for future reference.
```suggestion
  # work around if we're passing the zero module to singular, see
  # https://github.com/oscar-system/Singular.jl/issues/796
```

wdecker (approved):

hannes14:
fixed with https://github.com/Singular/Singular/commit/5704a442d5557973dbede7f6f4f317bf9402aeff in Singular,
need new Singular_jll, and Singular.jl release

wdecker (approved):

PR #3625, opened 2024-04-18
Some clean-up on monomial orderings of ideal gens

HechtiDerLachs:
After talking with @jankoboehm today about the problem mentioned on the other (merged) PR, we would like to try this new solution to hopefully avoid unnecessary comparison of monomial orderings in the call to `singular_generators`. 

The current version is to check whether the CI passes. If this is the case, then the assertion will be turned into an `hassert` so that one can still test this, but the check is not run all the time and the unnecessary comparison of monomial orderings does not take place. 

CC: @joschmitt , @fingolfin

### DIFF
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -315,8 +315,8 @@ function singular_generators(B::IdealGens, monorder::MonomialOrdering=default_or
     # Whenever both are set, the monomial orderings on the Oscar side and the Singular side should agree.
     # However, it might be the case that some (newly added) code is breaking this. In order to snoop 
     # for that, we allow for the assertion level to be put up and then this will throw. 
-    @hassert :IdealGens 1 monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord "orderings did not agree on the Oscar and the Singular side"
-    #@assert monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord "orderings did not agree on the Oscar and the Singular side"
+    @hassert :IdealGens 1 monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord
+    @assert monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord "orderings did not agree on the Oscar and the Singular side"
     return B.gens.S
   end
   SR = singular_poly_ring(B.Ox, monorder)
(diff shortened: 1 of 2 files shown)


HechtiDerLachs:
@jankoboehm : Tests fail. 

Hence, I propose to just pull back this PR. If anyone has ideas how to implement a working shortcut, we can dig it out again.

src/Rings/mpoly.jl
```diff
@@ -311,7 +311,14 @@ function singular_generators(B::IdealGens, monorder::MonomialOrdering=default_or
...
   isa(B.gens.Ox, MPolyQuoRing) && return B.gens.S
-  isdefined(B, :ord) && B.ord == monorder && monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord && return B.gens.S
+  if isdefined(B, :ord) && (B.ord === monorder || B.ord == monorder)
+    # Whenever both are set, the monomial orderings on the Oscar side and the Singular side should agree.
+    # However, it might be the case that some (newly added) code is breaking this. In order to snoop 
+    # for that, we allow for the assertion level to be put up and then this will throw. 
+    @hassert :IdealGens 1 monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord
+    @assert monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord "orderings did not agree on the Oscar and the Singular side"
+    return B.gens.S
+  end
```
joschmitt, line 321: My (unqualified) take on this would be to remove this whole `if` block. The `singular_assure` creates the ring on the Singular side with the same ordering as the internal ordering of the base ring on the OSCAR side. As long as one doesn't do anything fancy, this will be lex. (`B` has a field `keep_ordering` which appears to be `true` by default, maybe it should be `false`, no idea. And there are supposedly situations where the Singular ring is set from the beginning, no idea again.)
Either way, there appears to be no relationship between `B.ord` and `Singular.ordering(base_ring(B.S))` other than coincidence if I start with some ideal created in OSCAR. So, if the `if` condition is rarely `true` (somebody should confirm this) and testing it is expensive (as you say), I'd suggest to remove it.

jankoboehm:
For Gröbner bases the ordering on the Oscar side should agree with that on the Singular side, so one should avoid pulling everything through a morphism. If it is not a Gröbner basis, there is no reason why the orderings should not agree. The ordering field is there to avoid digging out the ordering from the Singular side all the time, and allow for Gröbner bases which do not even have a Singular side. If this is done consistently, there is no need to check. We will look into it.

PR #3626, opened 2024-04-18
Finish formatting of PolyhedralGeometry

lkastner:


.git-blame-ignore-revs
```diff
@@ -1,5 +1,7 @@
-# Ran JuliaFormatter on most of src/PolyhedralGeometry
+# Ran JuliaFormatter on most of {src,test}/PolyhedralGeometry
```
joschmitt:
```suggestion
# Ran JuliaFormatter on {src,test}/PolyhedralGeometry
```
I assume

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -242,13 +242,71 @@ Construct the regular icosahedron, one out of two exceptional Platonic solids.
...
+const _johnson_indexes_from_oscar = Set{Int}([
+  9,
+  10,
+  13,
+  16,
+  17,
+  18,
+  20,
+  21,
+  22,
```
joschmitt: Hmm, I'm not sure I like this change. If one scrolls down to the end of the list, one can't see the name `_johnson_indexes_from_oscar` anymore, at least on my screen. Apparently, one could leave it as it were and add comments for the formatter to ignore it (https://domluna.github.io/JuliaFormatter.jl/stable/#Turn-off/on-formatting) so that the tests pass, but I did not test this.

But you can also leave it like this; long lists of constants never look great I guess.

joschmitt (approved):

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
lgoettgens: I like the idea of putting *some few* noformat tags around things like this to keep files readable. That's why we even format them in the first place imo.

lkastner: I know that I can make the formatter ignore this. But maybe there is something to make it little more forgiving in general?

.git-blame-ignore-revs
lkastner: Thanks.

test/PolyhedralGeometry/cone.jl
```diff
@@ -136,30 +136,33 @@
...
   # We also put inverse here, since inversion in Julia produces float errors
   # and we don't want to do complicated matrix conversions here.
-  invA = [1   2   3   -4;
-          1   0   1   -1;
-          1   9   0   -6;
-          1   1   5   -5]
+  invA = [1 2 3 -4;
+    1 0 1 -1;
+    1 9 0 -6;
+    1 1 5 -5]
```
lgoettgens, line 161: I am not a fan of how this new option formats matrices like this. The offset of the first row is weird. I think without this option this was more aligned.
But let's hear other peoples opinions as well.

lkastner:
In general I would rather avoid anything were we manually interfere with the formatter since this always entails mentioning this in PRs and explaining it to devs. One main goal of the formatting tests is to reduce our work when reviewing PRs. It has the added benefit that there is an automated way for devs to fix their formatting. The many comments on formatting can seem a bit like micromanagement and be quite discouraging.

joschmitt:
What is the plan now? I would think that with every set of formatting rules, there will be some situation where the result is not optimal. And there will always be somebody who doesn't like something about the current style (I personally would put more white space than the formatter for example).
In case you are waiting for opinions: I can live with it as it is now.
If there is an option that magically aligns the matrices as we want, great. `align_matrix` seems not to be it; it even adds white space with every run for me, so the tests won't work.

PR #3628, opened 2024-04-18
booktest: fix projects when called from test_module 

benlorenz:
- use current load path instead of default
- make sure oscar is available
- restore correct load path

`test_module` also modifies the load path and thus broke the temp projects for the booktests.

cc: @lgoettgens

### DIFF
--- a/test/book/test.jl
+++ b/test/book/test.jl
@@ -181,7 +181,9 @@ isdefined(Main, :FakeTerminals) || include(joinpath(pkgdir(REPL),"test","FakeTer
   function test_chapter(chapter::String="")
     # add overlay project for plots
     custom_load_path = []
-    copy!(custom_load_path, Base.DEFAULT_LOAD_PATH)
+    old_load_path = []
+    copy!(custom_load_path, LOAD_PATH)
+    copy!(old_load_path, LOAD_PATH)
     curdir = pwd()
     act_proj = dirname(Base.active_project())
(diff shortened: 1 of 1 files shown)


lgoettgens (approved):
I can verify that both `Oscar.test_module("book"; new=false)` and `Oscar.test_module("book")` succeed locally. Thanks for the quick fix!

lgoettgens:
There seems to be an issue with loading `generic_root_count` (ref https://github.com/oscar-system/Oscar.jl/actions/runs/8742200378/job/23989900314?pr=3628#step:8:370).

Off-topic: Shouldn't the book tests be "required"?

benlorenz:
> There seems to be an issue with loading `generic_root_count` (ref [oscar-system/Oscar.jl/actions/runs/8742200378/job/23989900314?pr=3628#step:8:370](https://github.com/oscar-system/Oscar.jl/actions/runs/8742200378/job/23989900314?pr=3628#step:8:370)).

Will hopefully run through now. `Pkg.test` removes `@stdlib` from the load path which we previously avoided by switching back to the default load path. But now we reuse the current load path to make sure we can find Oscar even with the extra environments from `test_module` so we need to re-add this temporarily.

> Off-topic: Shouldn't the book tests be "required"?

Yes, I guess we should change that soon.

lkastner (approved):

PR #3630, opened 2024-04-19
Implement morphisms of elliptic surfaces given by translations

HechtiDerLachs:
CC: @simonbrandhorst 

Sections seem to be nasty. In the example added for the tests I couldn't get all charts to produce the local ideals for 2*P.


experimental/Schemes/elliptic_surface.jl
```diff
@@ -2119,7 +2136,71 @@ function morphism_from_section(X::EllipticSurface, i::Int)
...
+  set_attribute!(result, :is_isomorphism=>true)
+  return result
+end
+
+function _pushforward_section(
+    phi::MorphismFromRationalFunctions{<:EllipticSurface, <:EllipticSurface}, 
+    P::EllipticCurvePoint;
+    divisor::AbsWeilDivisor=_section(domain(phi), P),
+    codomain_charts::Vector{<:AbsAffineScheme} = affine_charts(codomain(phi))
+  )
```
HechtiDerLachs, line 2209: We should use this procedure here in general for sections. Today it seemed that `simplify` can not really take care of finding the parametrization.

experimental/Schemes/elliptic_surface.jl
```diff
@@ -2103,13 +2106,27 @@ end
...
+function morphism_from_section(
+    X::EllipticSurface, P::EllipticCurvePoint;
+    divisor::AbsWeilDivisor=_section(X, P)
+  )
   U = weierstrass_chart_on_minimal_model(X)
+  II = first(components(divisor))
+
+  # For the zero section we can not use the Weierstrass chart
+  if P.is_infinite
+    error("not implemented for the zero section")
```
simonbrandhorst: return the identity?

experimental/Schemes/elliptic_surface.jl
```diff
@@ -2103,13 +2106,27 @@ end
...
 # We keep the code for the moment as it will probably help us to reconstruct this map as a 
 # proper CoveredSchemeMorphism, once this is needed. 
-function morphism_from_section(X::EllipticSurface, i::Int)
-  P = X.MWL[i]
-  @assert !P.is_infinite
+=#
+function morphism_from_section(
+    X::EllipticSurface, P::EllipticCurvePoint;
+    divisor::AbsWeilDivisor=_section(X, P)
+  )
```
HechtiDerLachs, line 2139: This works now, but returns two maps: The identification IP^1 with the section and the inclusion of the latter.

HechtiDerLachs:
@afkafkafk13 : There's a lazy type for strict transforms of ideal sheaves here now. Maybe you would like to have a look.

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
```diff
@@ -146,3 +146,42 @@ end
...
+  # But it suffices to compare the result on the weierstrass charts.
+  weier = weierstrass_chart_on_minimal_model(X)
+  @test IIX(weier) == JJ(weier)
+
+  lat = algebraic_lattice(X)
+  A = [intersect(a, b) for a in lat, b in lat]
+
+  ll = Oscar._pushforward_lattice_along_isomorphism(trans)
+  B = [intersect(a, b) for a in ll, b in ll]
+  @test A == B
```
HechtiDerLachs: I added some more expensive tests here. Maybe we should disable them again, but leave them here as comment for later reference.

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -892,3 +893,71 @@ function blow_up(m::AbsCoveredScheme, I::AbsIdealSheaf; coordinate_name = "e")
...
+  f_loc = covering_morphism(f)[U]
+  V = codomain(f_loc)
+  IE_loc = IE(U)
+  tot = pullback(f_loc)(J(V))
+  result, _ = saturation_with_index(tot, IE_loc)
+  return result
+end
+
+function is_prime(I::StrictTransformIdealSheaf)
+  is_subset(original_ideal_sheaf(I), radical(center(morphism(I)))) && return false # It's the unit ideal sheaf in this case
```
simonbrandhorst, line 950: This could be replaced by radical membership?

afkafkafk13: This should indeed be replaced by a radical membership test (implemented in charts).

simonbrandhorst:
I just tried to compute the algebraic lattice with this branch and after 12 hours the computation did not complete. It was stuck in the computation of some gluings.

simonbrandhorst:
For the record on master I got
`392.821593 seconds (774.37 M allocations: 667.636 GiB, 41.50% gc time, 0.08% compilation time)` 
(the amount of allocations is scary)

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: I was thinking the same. I suggest to introduce a new type `RadicalOfIdealSheaf <: AbsIdealSheaf` and overwrite the methods for `in` and `is_subset` for these.

HechtiDerLachs:
> I just tried to compute the algebraic lattice with this branch and after 12 hours the computation did not complete. It was stuck in the computation of some gluings.

I can confirm these findings. Interestingly, it did not get stuck on any Groebner basis computation, but on `divexact` for two multivariate polynomials over a finite field. @fieker informed me the other day that these can be unexpectedly expensive. But these divisions are hardcoded into the fraction fields for addition and multiplication of elements and these are, in turn, used by the localizations. So we can not switch them off. It seems that there are just some gluings which are too hard to even be spelled out. 

This problem does not occur on the second last commit https://github.com/oscar-system/Oscar.jl/pull/3630/commits/54288dfaaa806c2c9029cd351a7ff55aeade8465 and this suggests to me that we are only stepping on a mine which has been there all the time. This unfortunate step is probably taken due to new heuristics implemented with `cheap_sub_ideal` for `StrictTransformIdealSheaf`. 

All this is pretty worrying to me since the desingularizations we are considering here should be regarded as rather simple, but yet they can not be fully computed. @afkafkafk13 : Does this seem familiar from your earlier experiences in Singular's desingularization?

HechtiDerLachs:
> the amount of allocations is scary

It is. Looking at some intermediate computations with `ProfileView`, it seems that a lot of time is spent on conversions from and to the Singular side. That might also be a reason for the massive number of allocations. But I'm not sure.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: I implemented this now. One can overwrite the `is_subset` method, but not what happens on the charts. This would require a similar procedure for the `MPolyIdeal`s themselves which is probably stepping outside my boundaries.

afkafkafk13:
> > the amount of allocations is scary
> 
> It is. Looking at some intermediate computations with `ProfileView`, it seems that a lot of time is spent on conversions from and to the Singular side. That might also be a reason for the massive number of allocations. But I'm not sure.

Yes, this observation makes a lot of sense to me: the overhead for translation back and forth does not make a **large** difference for each single instance of non-trivial computations, but working on covered schemes we have a huge amount of  these (often smaller) non-trivial computations and this just adds up.

afkafkafk13:
> > I just tried to compute the algebraic lattice with this branch and after 12 hours the computation did not complete. It was stuck in the computation of some gluings.
> 
> I can confirm these findings. Interestingly, it did not get stuck on any Groebner basis computation, but on `divexact` for two multivariate polynomials over a finite field. @fieker informed me the other day that these can be unexpectedly expensive. But these divisions are hardcoded into the fraction fields for addition and multiplication of elements and these are, in turn, used by the localizations. So we can not switch them off. It seems that there are just some gluings which are too hard to even be spelled out.
> 
> This problem does not occur on the second last commit [54288df](https://github.com/oscar-system/Oscar.jl/commit/54288dfaaa806c2c9029cd351a7ff55aeade8465) and this suggests to me that we are only stepping on a mine which has been there all the time. This unfortunate step is probably taken due to new heuristics implemented with `cheap_sub_ideal` for `StrictTransformIdealSheaf`.
> 
> All this is pretty worrying to me since the desingularizations we are considering here should be regarded as rather simple, but yet they can not be fully computed. @afkafkafk13 : Does this seem familiar from your earlier experiences in Singular's desingularization?

Here I have two bits of input:
A) Avoid computing over fraction fields of multivariate polynomial rings as much as possible. Emulate whatever is possible in a polynomial ring setting (I am not saying quotient ring!). If you are dealing with results of sequences of blow-ups and you are not facing subsets of (rather young) exceptional divisors, you might consider passing via a common ancester instead of glueing.
B) Often you do not need all charts to have the full information, but you need to either choose carefully, what you need, or start a parallel race. Try a heuristic (taking into accout the number of monomials of the data and distance in the tree of blow-up charts) to find a smaller set of gluings to actually be computed. Note that passing through 2 or 3 rather easy gluings of nearby charts in the tree of charts can be a lot less expensive than computing the direct gluing map.

HechtiDerLachs:
> Here I have two bits of input:

But if I understand this correctly, this does not save us from such gluing-mines in total, correct? It's just a guessing game to avoid stepping on them. That's rather unsatisfactory in a linear environment where we can not perform "race-to-win" tasks in parallel on different workers.

afkafkafk13:
> > Here I have two bits of input:
> 
> But if I understand this correctly, this does not save us from such gluing-mines in total, correct? It's just a guessing game to avoid stepping on them. That's rather unsatisfactory in a linear environment where we can not perform "race-to-win" tasks in parallel on different workers.

Indeed A is just a guessing game.
For B: The passage through the tree of charts is fine for nice centers (which you should be seeing in your easy example after simplification), but becomes nasty, whenever your center is described using complicated polynomials (explosion of the number of terms after just a few steps!).  This is more than just guesswork. Without this and a game of ignoring charts without new information (X does not meet chart outside the locus covered by other charts), you are quickly 'dead'.

This is the background, why I was so interested to have all this in parallel in the first place. In a non-sequential setting, you are allowed to simply let thoses mines explode on  some of the nodes, clean up the debris in due time (after heureka) and then reuse the nodes for new tasks.
Other lesson learned in this context are: forget (free!!!) all refinements after use, as soon as your data has been expressed in terms of  a coarser covering. By using an evitable finer covering, you nearly always loose more than you gain.

HechtiDerLachs:
So: It seems to be essential to simplify the multiplicative set as much as possible by factoring all the admissible denominators. But at the same time, factoring denominators of degree ~100 also killed the process. Hence, I suggest to choose a heuristic based on the total degree for `simplify_light` and use that.

simonbrandhorst:
With the newest patch I could compute each of the 325 gluings in < than a second. Great!
The computation of the blowups and all gluings:
`75.179361 seconds (789.43 M allocations: 29.539 GiB, 18.79% gc time, 1.26% compilation time: 15% of which was recompilation)`
I suppose that if we can stomp on some of these allocations we could get down by another factor of 10.

afkafkafk13:
> With the newest patch I could compute each of the 325 gluings in < than a second. Great! The computation of the blowups and all gluings: `75.179361 seconds (789.43 M allocations: 29.539 GiB, 18.79% gc time, 1.26% compilation time: 15% of which was recompilation)` I suppose that if we can stomp on some of these allocations we could get down by another factor of 10.

The number of allocations actually went up, but the size of used memory dropped by more than a factor of 20. This is really a huge step.

HechtiDerLachs:
The bibtool tests fail due to external reasons. Let's see whether they get back to normal.

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
```diff
@@ -146,3 +146,85 @@ end
...
+             [ 9,  4,  -1,  -2,  -3,  -2,  -2,  -2,  -2,  -2,  0,   0,  0,  -1,  -1,  -1,   0,   0],
+             [ 4,  2,   0,   0,  -1,  -1,  -1,   0,   0,  -1,  1,   0,  0,   0,  -1,  -1,   0,   0],
+             [ 6,  2,  -1,  -1,  -1,  -1,  -1,   0,  -1,  -2,  1,  -1,  1,   0,  -1,  -1,   0,   0],
+             [ 6,  2,   0,  -1,  -2,  -1,  -1,  -1,  -1,  -1,  1,   1,  0,  -1,  -1,  -1,   0,   0],
+             [11,  4,  -1,  -2,  -3,  -2,  -2,  -1,  -1,  -2,  1,   0,  0,  -1,  -1,  -2,   0,   0],
+             [11,  4,  -1,  -2,  -3,  -2,  -2,  -1,  -1,  -2,  1,   0,  0,  -1,  -2,  -1,   0,   0],
+             [11,  4,  -1,  -1,  -2,  -2,  -2,  -2,  -2,  -2,  1,   0,  0,  -1,  -1,  -1,  -1,   0],
+             [10,  4,  -1,  -2,  -3,  -2,  -2,  -1,  -1,  -2,  1,   0,  0,  -1,  -1,  -1,   0,  -1]
+            ]
+  @test_broken res_mat == ref_mat # TODO: Simon, are you sure about the form of ref_mat?
```
HechtiDerLachs: @simonbrandhorst : The good news: All computations go through now rather quickly. The bad news: Comparing with the matrix you sent me on Slack the other day, we still get different results. Can you have a look whether I interpreted everything correctly?

HechtiDerLachs: I checked that the resulting matrix `res_mat` is indeed orthogonal and gives the correct intersection matrix. One could also add this as a (working) test here. But the translation by this particular section does not seem to produce the base change you anticipated.

experimental/Schemes/BlowupMorphism.jl
```diff
@@ -892,3 +893,71 @@ function blow_up(m::AbsCoveredScheme, I::AbsIdealSheaf; coordinate_name = "e")
...
+
+function produce_object_on_affine_chart(I::StrictTransformIdealSheaf, U::AbsAffineScheme)
+  f = morphism(I)
+  X = domain(f)
+  Y = codomain(f)
+  J = original_ideal_sheaf(I)
+  @assert any(x->x===U, affine_charts(X))
+  E = exceptional_divisor(f)
+  IE = ideal_sheaf(E)
+  # We assume that the covering morphism has the default_covering of X as its domain.
```
simonbrandhorst, line 939: how restrictive is this assumption? will it explode otherwise?
In this please raise a useful error message.

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: here it still says radical, is the code up to date?

simonbrandhorst: further this should be an attribute.

experimental/Schemes/IdealSheaves.jl
```diff
@@ -555,6 +555,10 @@ function is_one(I::PrimeIdealSheafFromChart; covering::Covering=default_covering
   return false
 end
 
+function dim(I::PrimeIdealSheafFromChart)
+  return dim(I(original_chart(I)))
+end
+
 function is_one(I::SumIdealSheaf; covering::Covering=default_covering(scheme(I)))
```
simonbrandhorst, line 562: make this an attribute?

experimental/Schemes/IdealSheaves.jl
```diff
@@ -555,6 +555,10 @@ function is_one(I::PrimeIdealSheafFromChart; covering::Covering=default_covering
   return false
 end
 
+function dim(I::PrimeIdealSheafFromChart)
```
simonbrandhorst, line 558:
```suggestion
@attr function dim(I::PrimeIdealSheafFromChart)
```

experimental/Schemes/IdealSheaves.jl
```diff
@@ -865,14 +854,66 @@ function maximal_associated_points(I::AbsIdealSheaf; covering=default_covering(s
...
 
+function radical_membership(x::RingElem, I::MPolyLocalizedIdeal)
+  return radical_membership(lifted_numerator(x), saturated_ideal(I))
+end
+
+function radical_membership(x::RingElem, I::MPolyQuoLocalizedIdeal)
+  return radical_membership(lifted_numerator(x), saturated_ideal(I))
+end
+
+# Proof of concept method but seemingly slower than the above
```
simonbrandhorst: seems to contradict you comment above?

experimental/Schemes/IdealSheaves.jl
```diff
@@ -865,14 +854,66 @@ function maximal_associated_points(I::AbsIdealSheaf; covering=default_covering(s
...
+  # Manually fill up the cache
+  if !use_decomposition_info # We can only do this if all we treated all components on all charts.
+    for U in patches(covering)
+      I_one = ideal(OO(U), one(OO(U)))
+      for P in comps
+        haskey(object_cache(P), U) && continue
+        object_cache(P)[U] = I_one
+      end
+    end
+  end
```
simonbrandhorst, line 914: This bit of code looks rather dangerous. what if garbage collection decided to clear 
the weak references in between?

experimental/Schemes/IdealSheaves.jl
```diff
@@ -1842,4 +1882,82 @@ function sub(I::AbsIdealSheaf)
...
+function is_subset(I::PrimeIdealSheafFromChart, rad::RadicalOfIdealSheaf)
+  X = scheme(rad)
+  @assert X === scheme(I) "ideal sheaves do not live on the same scheme"
+  U = original_chart(I)
+  return all(g->radical_membership(g, rad(U)), gens(I(U)))
+end
+
+function radical(I::AbsIdealSheaf)
+  result = RadicalOfIdealSheaf(I)
+  set_attribute!(result, :is_radical=>true) # Some methods might be blind to is_radical and only want to check `is_known_to_be_radical` via attributes. Setting this makes sure they get it.
```
simonbrandhorst: the safer way is probably to do this in the constructor of `RadicalOfIdealSheaf` ?

experimental/Schemes/elliptic_surface.jl
```diff
@@ -1021,21 +1021,25 @@ function section(X::EllipticSurface, P::EllipticCurvePoint)
...
-    PX = ideal_sheaf(X0,U,[OO(U)(i) for i in [x*denominator(b[1])(t)-numerator(b[1])(t),y*denominator(b[2])(t)-numerator(b[2])(t)]])
+    return IdealSheaf(X0, V, [x,z])
   end
+  U = X0[1][1]
+  (x,y,t) = coordinates(U)
+  b = P
+  return ideal_sheaf(X0,U,[OO(U)(i) for i in [x*denominator(b[1])(t)-numerator(b[1])(t),y*denominator(b[2])(t)-numerator(b[2])(t)]])
+end
+
+function _section(X::EllipticSurface, P::EllipticCurvePoint)
```
simonbrandhorst, line 1039: We could use the strict transform ideal sheaf or the prime ideal sheaf here instead. 
... in the past this somehow killed performance but with the current changes it is maybe okay again?

src/Rings/mpoly-localizations.jl
```diff
@@ -126,7 +126,23 @@ function simplify_light(S::MPolyPowersOfElement)
   end
 
   new_denom = elem_type(R)[a for a in unique!(new_denom) if !is_unit(a)]
+  fac_denom = elem_type(R)[]
 
+  # Try to factor the denominators if this is deemed possible
```
simonbrandhorst: would it be worth to make `simplify` and `simplify_light` idempotent?

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
```diff
@@ -146,3 +146,89 @@ end
...
+  #set_verbose_level(:EllipticSurface, 5)
+
+  # The following should not take more than at most two minutes.
+  # But it broke the tests at some point leading to timeout, 
+  # so we put it here to indicate regression.
+  for (i, g) in enumerate(values(gluings(default_covering(X))))
+    gluing_domains(g) # Trigger the computation once
+  end
+
+  D_P = Oscar._section(X, P)
```
simonbrandhorst:
```suggestion
  D_P = section(X, P)
```

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
```diff
@@ -146,3 +146,89 @@ end
...
+  weier = weierstrass_chart_on_minimal_model(X)
+  @test IIX(weier) == JJ(weier)
+
+  
+  # pushforward of the whole algebraic lattice with verification of the result.
+  # This test is hardcoded, but should remain stable throughout future changes.
+  # Details can be found in a forthcoming paper on Oguiso's automorphism 
+  # by Simon Brandhorst and Matthias Zach
+  lat = algebraic_lattice(X)[1]
+  A = [intersect(a, b) for a in lat, b in lat]
```
simonbrandhorst:
```suggestion
```
There are enough intersection computations done in `algebraic_lattice`

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
simonbrandhorst: I can confirm your matrix is definitely fine. The reference matrix describes something which is not quite a translation.

simonbrandhorst:
In conclusion I think that this can be merged once the tests are adjusted and some parts of the comments dealt with. 

It is a big step forward in performance. Thank you for your work @HechtiDerLachs !

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: No. It's already an attribute of the `original_ideal`.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: Yes. It's deprecated now.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: Garbage collection is not the issue since all these patches are stored in the covering of the scheme. But people might mess with the implementations of `==`, so I leave a comment here.

experimental/Schemes/elliptic_surface.jl
HechtiDerLachs: At least the creation of the object should run more or less instantly now, yes. And it is using the two already now.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: Are they not? If so, then only up to the order of the elements.

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
HechtiDerLachs: Didn't know this existed. Thx.

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
HechtiDerLachs: But we need the matrix further down.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: But this is the constructor, isn't it?

HechtiDerLachs:
About the `radical` questions: Yes, the code is up to date. The method for `is_subset` has been overwritten so that only `radical_membership` is used on the affine charts.

experimental/Schemes/IdealSheaves.jl
simonbrandhorst: it is the outer constructor. I would put it in the inner constructor `RadicalOfIdealSheaf`

src/Rings/mpoly-localizations.jl
simonbrandhorst: I think it tries to factor each time again? What I mean is to give them an attribute like 
`is_simplified`

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
simonbrandhorst: `gram_matrix(ambient_space(algebraic_lattice(x)[3]))` is what you need.
Perhaps I should document this.

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: Ah now this works as intended, because radical is lazy and just build a `RadicalIdealSheaf`?

experimental/Schemes/BlowupMorphism.jl
simonbrandhorst: I don't see how this is resolved :P. But I can live with it.

experimental/Schemes/IdealSheaves.jl
simonbrandhorst: This is a `SumIdealSheaf`, what is the `original_ideal` ? There should be several.

experimental/Schemes/elliptic_surface.jl
simonbrandhorst: Because `strict_transform` returns a `StrictTransformIdealSheaf`?

test/AlgebraicGeometry/Schemes/elliptic_surface.jl
```diff
@@ -146,3 +146,65 @@ end
...
+  IIX = first(components(section(X, 2*P)));
+  # We have little chance to get through with the computations on all charts. 
+  # But it suffices to compare the result on the weierstrass charts.
+  weier = weierstrass_chart_on_minimal_model(X)
+  @test IIX(weier) == JJ(weier)
+
+  
+  # pushforward of the whole algebraic lattice with verification of the result.
+  lat, _, A = algebraic_lattice(X)
+  A = gram_matrix(A)
```
simonbrandhorst:
```suggestion
  A = gram_matrix(ambient_space(A))
```

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: I think I added an assertion just the line below. Didn't I?

simonbrandhorst: right. I overlooked this. Sorry.

experimental/Schemes/BlowupMorphism.jl
HechtiDerLachs: It is an attribute already. I don't know why it does not show the updated version of the code here.

experimental/Schemes/IdealSheaves.jl
HechtiDerLachs: I thought you were referring to line 558.

HechtiDerLachs: Line 562 is already an attribute. Look at the code below.

simonbrandhorst (approved):

experimental/Schemes/IdealSheaves.jl
simonbrandhorst: Right :-) I was just looking for the `@attr`. But this methods takes several arguments and hence `@attr` does not work.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: That could make sense, yes.

simonbrandhorst: I can also be delayed to a later PR. Your choice.

simonbrandhorst:
Did you check that the test runs reasonably fast?

HechtiDerLachs:
> Did you check that the test runs reasonably fast?

Roughly 12 minutes for the newly added test. Still a bit much, so we might want to leave a comment that this test is a candidate for being disabled? Any suggestions/opinions, @fingolfin ?

simonbrandhorst:
11 Minutes is far too much. I guess we could live with something that takes 2 minutes.

simonbrandhorst:
The good news is that the old tests which were previously too slow now all run in less than a minute. I think we can enable some of them now.

simonbrandhorst:
My suggestion would be to disable the  test for now, get it merged and then do the testing in a different PR.

src/Rings/mpoly-localizations.jl
HechtiDerLachs: Done.

PR #3631, opened 2024-04-19
Refactor `AuxDocTests`

### LINKED ISSUE #35: OscarCI runs fake doctests with `depwarn=yes`
Running normal tests with `depwarn=yes` is fine as they still succeed and only some additional output is printed. Normal doctests are run directly using `doctest(Oscar)` (and not via `Pkg.test`), so the default of `depwarn=false` is applied here as well.

However, in Oscar there are some "fake doctests", e.g. https://github.com/oscar-system/Oscar.jl/blob/556be5ae27a4fe3d45dfd26e536101e535085d0b/test/Groups/group_characters.jl#L755. These are called via a `doctest` command from the test job, and thus `depwarn=yes` is set there. As the doctest compares output, the presence of deprecation warnings lets the tests fail, even though deprecation warnings shouldn't do that.

I really don't know if this can be avoided at all, and if the place to fix it is here in the way the tests are run, or in Oscar itself at the place where the fake doctests are called.

lgoettgens:
This resolves some issues with the `AuxDocTests` (doctests used for testing some fancy printing, but run as part of tests, e.g. for character tables) I experienced for a while. The changes are as follows:

- All code regarding the AuxDocTests (except the test cases of course) is extracted into a macro in `utils/test.jl` to ease future changes and keep all of them consistent.
- When running a file with AuxDocTests with e.g. `test_module`, afterwards GC logging is always enabled. This gets a bit annoying. I resolve this as follows: For julia 1.11+, get the previous GC logging flag using `GC.logging_enabled` (thanks @fingolfin for adding that to Julia), disable for the doctests, and enable afterwards using the saved flag. For older julia versions, the GC logging lines gets disabled for the doctests and enabled after disregarding the state of the flag (as it is currently the case).
- Deprecation warnings during an AuxDocTest makes it fail. To circumvent this, depwarns now get removed using a doctestfilter. (resolves https://github.com/oscar-system/OscarDevTools.jl/issues/35)


lgoettgens:
@fingolfin volunteered to have fun reviewing macros and regexes

fingolfin (approved):

PR #3636, opened 2024-04-22
Consolidate and optimize `minimal_subalgebra_generators_with_relations`

joschmitt:
Documents and tests the (internal) function `minimal_subalgebra_generators_with_relations` and fixes some bugs found by testing...

Also renames the file `test/Rings/mpoly_affine_algebras.jl` to `test/Rings/mpoly-affine-algebra.jl`.

EDIT: This became a bit more now: I split up the function `subalgebra_membership_homogeneous` in two internal functions, so that one can reuse hard-to-compute data. This makes the logic of the code more complicated, but it reduced the runtime of `minimal_subalgebra_generators_with_relations` from ~4 seconds to ~1.2 seconds in a 'real life' example.


fingolfin (dismissed):

joschmitt:
To record the outcome of the discussion on last Wednesday:
* There are objections whether the function should be called `..._with_relations` because it does not return *all* relations between the polynomials.
* Moving around lists of polynomials which are 'interpreted' as a subalgebra should arguably not be the user-facing interface (I don't have an idea how it could be done better though).

src/Rings/mpoly-affine-algebras.jl
```diff
@@ -1214,11 +1231,50 @@ julia> minimal_subalgebra_generators(V)
...
-  I = ideal(parent(V[1]), [ zero(parent(V[1])) ])
-  return _minimal_subalgebra_generators_with_relations(V, I, check = check)[1]
+  return minimal_subalgebra_generators_with_relations(V; check=check)[1]
 end
 
+@doc raw"""
+    minimal_subalgebra_generators_with_relations(V::Vector{T}; check::Bool = true) where T <: Union{MPolyRingElem, MPolyQuoRingElem}
+
+Given a vector `V` of homogeneous elements of a positively graded multivariate
+polynomial ring, or of a quotient of such a ring, return a minimal subset `W` of
```
fingolfin: Minimal in which sense? By inclusion (i.e. no further element can be dropped from `W`), or by size (which imply the former but be stronger in general)?

src/exports.jl
```diff
@@ -992,6 +992,7 @@ export minimal_nonfaces
 export minimal_normal_subgroups, has_minimal_normal_subgroups, set_minimal_normal_subgroups
 export minimal_primes
 export minimal_subalgebra_generators
+export minimal_subalgebra_generators_with_relations
```
fingolfin: I am (still) opposed to exporting this function under that name.

fingolfin (commented):
I still think `minimal_subalgebra_generators_with_relations` is a horrible name, but this PR does not introduce it. We should of course allow to improve the existing code.

But I a concerned about exporting the name as we are then bound to provide it in all further OSCAR 1.x releases.

I'd be OK with merging this if that `export` goes away (and of course that would not preclude a future PR from exporting it, if circumstances change).

Perhaps I can discuss this with @fieker tomorrow.

src/Rings/mpoly-affine-algebras.jl
joschmitt: It's of minimal cardinality, I clarified the doc string.

src/exports.jl
joschmitt: I removed the export.

joschmitt:
Please review this. It contains bugfixes and performance improvements.

wdecker (commented):
Why `Ideal(0)` with capital `I`? Also, should we replace `0` by the zero of the corresponding ring?

return `(true, ideal(0))` if the elements of `V` are algebraically independent over `K`.
return `(true, Ideal (0))` if the elements of `V` are algebraically independent over `K`.

wdecker (commented):
We alread export `is_algebraically_independent_with_relations`. We must also export `minimal_subalgebra_generators_with_relations`.

joschmitt:
> Why `Ideal(0)` with capital `I`? Also, should we replace `0` by the zero of the corresponding ring?

`Ideal (0)` is how the zero ideal would be printed. If we want to write the code to produce the zero ideal there, it would be something like `ideal([zero(parent(V[1]))])`.

joschmitt:
> We alread export `is_algebraically_independent_with_relations`. We must also export `minimal_subalgebra_generators_with_relations`.

I think the function `is_algebraically_independent_with_relations` is actually a good argument to *not* export `minimal_subalgebra_generators_with_relations` under this name. The 'relations' returned by the latter function are not the relations of all generators, but only describe how the discarded generators can be represented in the chosen minimal set. I probably introduced this misleading name, but I can also not think of anything better apart from generic/vague things like `minimal_subalgebra_generators_with_data`.

wdecker:
O.k., then leave it as it is.

wdecker:
O.k., I misunderstood what the function is doing.

wdecker (approved):

PR #3637, opened 2024-04-22
Add `_ring_iso` helper to help avoid some `Base.getproperty` methods

fingolfin:
I plan to get rid of the rest, too, but that'll take some more time.

### DIFF
--- a/src/Groups/matrices/MatGrp.jl
+++ b/src/Groups/matrices/MatGrp.jl
@@ -205,11 +205,10 @@ function _ring_iso(G::MatrixGroup{T}) where T
       entries = reduce(vcat, vec(collect(matrix(g))) for g in gens(G))
       # construct a number field over which all matrices are already defined
       nf, nf_to_QQBar = number_field(QQ, entries)
-      psi = pseudo_inv(nf_to_QQBar)
       iso = iso_oscar_gap(nf)
       G.ring_iso = MapFromFunc(G.ring, codomain(iso),
                                x -> iso(preimage(nf_to_QQBar, x)),
-                               y -> preimage(iso,nf_to_QQBar(y))
+                               y -> nf_to_QQBar(preimage(iso, y))
                                )
     else
       G.ring_iso = iso_oscar_gap(G.ring)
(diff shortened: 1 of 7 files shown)


src/Groups/matrices/MatGrp.jl
```diff
@@ -198,28 +198,32 @@ function assign_from_description(G::MatrixGroup)
...
+    if T === QQBarFieldElem
+      # get all matrix entries into one vector
+      entries = reduce(vcat, vec(collect(matrix(g))) for g in gens(G))
+      # construct a number field over which all matrices are already defined
+      nf, nf_to_QQBar = number_field(QQ, entries)
+      psi = pseudo_inv(nf_to_QQBar)
+      iso = iso_oscar_gap(nf)
+      G.ring_iso = MapFromFunc(G.ring, codomain(iso),
+                               x -> iso(preimage(nf_to_QQBar, x)),
+                               y -> preimage(iso,nf_to_QQBar(y))
```
joschmitt: I know that this was there before, but shouldn't it be
```suggestion
                               y -> nf_to_QQBar(preimage(iso, y))
```
Also the map `psi` seems not to be used in this function.

ThomasBreuer (approved):
Yes, thanks, good idea.

src/Groups/matrices/MatGrp.jl
```diff
@@ -198,28 +198,32 @@ function assign_from_description(G::MatrixGroup)
...
 end
 
+function _ring_iso(G::MatrixGroup{T}) where T
+  if !isdefined(G, :ring_iso)
+    if T === QQBarFieldElem
+      # get all matrix entries into one vector
+      entries = reduce(vcat, vec(collect(matrix(g))) for g in gens(G))
+      # construct a number field over which all matrices are already defined
+      nf, nf_to_QQBar = number_field(QQ, entries)
+      psi = pseudo_inv(nf_to_QQBar)
```
fingolfin:
```suggestion
```

src/Groups/matrices/MatGrp.jl
fingolfin: Ah indeed!

`psi` was from an earlier iteration of this code and I forgot to delete it.

I guess we should have a test case for this problem, too...

PR #3639, opened 2024-04-23
Zariski/Lipman approach to resolution of singularities for 2-dim. schemes

afkafkafk13:
- introduces mixed sequences of Blow-Ups and Normalization 
- resolution of singularities for 2-dim schemes via these mixed sequences

@HechtiDerLachs : I had problems getting the EffectiveCartierDivisors from the blow-up steps and the IdealSheafs from the normalization steps into one Vector.  

edit:
now  done:
- tweak strict_transform to do something sensible (pullback) for normalization steps
- allow phi.ex_div to swallow both EffectiveCartierDivisors and IdealSheafs
-tests

still missing:
- documentation


experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+   } <:AbsCoveredSchemeMorphism{
+                                 DomainType,
+                                 CodomainType,
+                                 Nothing,
+                                 NormalizationMorphism,
+                                }
+
+  X::DomainType
+  Y::CodomainType
+  f::CoveringMorphism
```
simonbrandhorst: Here you are essentially duplicating the implementation of `CoveredSchemeMorphism`.
This is maybe possible, but not necessary. After all you want a `CoveredSchemeMorphism` which knows a bit more. You can do as follows:
```suggestion
  f::CoveredSchemeMorphism{DomainType, CodomainType, BaseMorphismType}
```
then implement 
`underlying_scheme(f::NormalizationMorphism) = f.X`
and you are done.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+                                                                MixedBlowUpSequence{DomainType, CodomainType}
+                                              }
+  maps::Vector{Union{<:BlowupMorphism,<:NormalizationMorphism}}       # count right to left:
+                                                 # original scheme is codomain of map 1
+  # boolean flags
+  resolves_sing::Bool                            # domain not smooth yet?
+  is_trivial::Bool                               # codomain already smooth?
+
+  # fields for caching, to be filled during desingularization
+  ex_div::Vector{AbsIdealSheaf}                  # list of exc. divisors arising from individual steps
```
simonbrandhorst:
```suggestion
  ex_div::Vector{<:AbsIdealSheaf}                  # list of exc. divisors arising from individual steps
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+
+end
 
 
 ##################################################################################################
 # getters
 ##################################################################################################
+underlying_morphism(phi::NormalizationMorphism) = CoveredSchemeMorphism(phi.X,phi.Y,phi.f)
+domain(phi::NormalizationMorphism) = phi.X
+codomain(phi::NormalizationMorphism) = phi.Y
```
simonbrandhorst: these getters will work automatically once `underlying_morphism` is implemented as sketched above.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
 morphisms(phi::AbsDesingMor) = copy(phi.maps)
 morphism(phi::AbsDesingMor,i::Int) = copy(phi.maps[i])
 last_map(phi::AbsDesingMor) = phi.maps[end]
-exceptional_divisor_list(phi::BlowUpSequence) = phi.ex_div  ## derzeit Liste von Eff. Cartier Div.
+
+exceptional_divisor_list(phi::BlowUpSequence) = phi.ex_div
+
+## entries of ex_div corresponding to normalization steps are only exceptional divisors at the very end
+## so only return them at the very end or on specific demand
```
simonbrandhorst, line 136: would it be feasible to have a function for the exceptional locus, i.e. the locus of points where our morphism is not a local isomorphism?

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+exceptional_divisor_list(phi::BlowUpSequence) = phi.ex_div
+
+## entries of ex_div corresponding to normalization steps are only exceptional divisors at the very end
+## so only return them at the very end or on specific demand 
+function exceptional_divisor_list(phi::MixedBlowUpSequence, seq_unclean::Bool=false)
+  phi.resolves_sing || seq_unclean || error("only available for completed resolution")
+  return phi.ex_div
+end
+
+normal_steps(phi::MixedBlowUpSequence) = phi.normal_steps
```
simonbrandhorst: how about `nomalization_steps`?

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+ 
 embeddings(phi::BlowUpSequence) = phi.embeddings
 
-## do not use!!! (for forwarding and certain emergenies)
+## do not use!!! (for forwarding and certain emergencies)
```
simonbrandhorst: I wonder why one should not use this?

simonbrandhorst (changes requested):

experimental/Schemes/Resolution_structure.jl
afkafkafk13: This is a leftover from before Matthias brushed it up a bit. I will delete it.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Yes, I already planned having this.

experimental/Schemes/CoveredScheme.jl
```diff
@@ -633,13 +633,13 @@ codomain(f::CompositeCoveredSchemeMorphism) = codomain(f.maps[end])
...
+    result = underlying_morphism(first(maps(f)))::AbsCoveredSchemeMorphism
     for i in 2:length(maps(f))
-      result = compose(result, underlying_morphism(maps(f)[i]))::CoveredSchemeMorphism
+      result = compose(result, underlying_morphism(maps(f)[i]))::AbsCoveredSchemeMorphism
     end
     f.composed_map = result
   end
-  return f.composed_map::CoveredSchemeMorphism
+  return f.composed_map::AbsCoveredSchemeMorphism
 end
```
HechtiDerLachs: Please do not change this. The `CompositeCoveredSchemeMorphism` should not be touched and stay independent of your changes. It can be used, but I think it is essential that `underlying_morphism` returns an actual `CoveredSchemeMorphism` in this case. At least that was the intention. 

Now that I think about it: Maybe we have to introduce and use something like `as_covered_scheme_morphism` here instead of `underlying_morphism` as the latter is supposed to also work recursively.

experimental/Schemes/Resolution_structure.jl
HechtiDerLachs:
```suggestion
  underlying_morphism::CoveredSchemeMorphism
```
It is just as @simonbrandhorst says. I will also suggest accompanying changes below.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+             BaseMorType
+            }
+    @check is_refinement(domain(f), default_covering(X)) "covering not found in domain"
+    @check is_refinement(codomain(f), default_covering(Y)) "covering not found in codomain"
+    @check is_normal(X) "not a normalization morphism"
+    @check all(Xi->is_subset(codomain(Xi), X), inclusions) "decomposition does not match"
+    ret_value = new{DomainType,CodomainType, BaseMorType}(X,Y,f)
+    ret_value.inclusions = inclusions
+    return ret_value    
+  end
```
HechtiDerLachs:
```suggestion
  function NormalizationMorphism(
      f::CoveredSchemeMorphism{DomainType, CodomainType, Nothing},
      inclusions::Vector{<:AbsCoveredSchemeMorphism};
      check::Bool=true
    ) where {
             DomainType<:AbsCoveredScheme,
             CodomainType<:AbsCoveredScheme
            }
    @check is_normal(X) "not a normalization morphism"
    @assert all(inc->codomain(inc) === domain(f), inclusions) "domains and codomains do not match"
    ret_value = new{DomainType,CodomainType}(f, inclusions)
    return ret_value    
  end
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+@attributes mutable struct NormalizationMorphism{
+    DomainType<:AbsCoveredScheme,
+    CodomainType<:AbsCoveredScheme,
+    BaseMorphsimType
+   } <:AbsCoveredSchemeMorphism{
+                                 DomainType,
+                                 CodomainType,
+                                 Nothing,
+                                 NormalizationMorphism,
+                                }
```
HechtiDerLachs:
```suggestion
@attributes mutable struct NormalizationMorphism{
    DomainType<:AbsCoveredScheme,
    CodomainType<:AbsCoveredScheme
   } <:AbsCoveredSchemeMorphism{
                                 DomainType,
                                 CodomainType,
                                 Nothing,
                                 NormalizationMorphism,
                                }
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+
+  X::DomainType
+  Y::CodomainType
+  f::CoveringMorphism
+  inclusions::Vector{CoveredSchemeMorphism}
```
HechtiDerLachs:
```suggestion
  inclusions::Vector{<:AbsCoveredSchemeMorphism}
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,138 @@
...
+             CodomainType<:AbsCoveredScheme,
+             BaseMorType
+            }
+    @check is_refinement(domain(f), default_covering(X)) "covering not found in domain"
+    @check is_refinement(codomain(f), default_covering(Y)) "covering not found in codomain"
+    @check is_normal(X) "not a normalization morphism"
+    ret_value = new{DomainType,CodomainType, BaseMorType}(X,Y,f)
+    ret_value.inclusions = [identity_map(X)]
+    return ret_value    
+  end
```
HechtiDerLachs:
```suggestion
  function NormalizationMorphism(
      f::CoveredSchemeMorphism{DomainType, CodomainType, Nothing};
      check::Bool=true
    ) where {
             DomainType<:AbsCoveredScheme,
             CodomainType<:AbsCoveredScheme
            }
    @check is_normal(X) "not a normalization morphism"
    ret_value = new{DomainType,CodomainType}(f, [identity_map(X)])
    return ret_value    
  end
```

experimental/Schemes/Resolution_structure.jl
HechtiDerLachs:
```suggestion
underlying_morphism(phi::NormalizationMorphism) = phi.underlying_morphism
```

experimental/Schemes/Resolution_structure.jl
HechtiDerLachs: We have also already discussed this months ago when working on the general interface for blowups and desingularizations. It would be nice if this was streamlined with what we agreed upon back then.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -241,6 +380,45 @@ function _desing_curve(X::AbsCoveredScheme, I_sl::AbsIdealSheaf)
...
+    f = initialize_mixed_blow_up_sequence(phi,I_sl)
+    f_initialized = true
+  end    
+  
+  I_sl_temp = ideal_sheaf_of_singular_locus(Xnorm)
+  decomp = maximal_associated_points(I_sl_temp)
+
+  while !is_one(I_sl_temp)
+    while length(decomp) > 0
+      I = small_generating_set(pop!(decomp))
```
HechtiDerLachs, line 614: In my opinion `small_generating_set` does not make sense for ideal sheaves. We should take care that `simplify` does what is needed here and use that.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -241,6 +380,45 @@ function _desing_curve(X::AbsCoveredScheme, I_sl::AbsIdealSheaf)
...
+        f = initialize_mixed_blow_up_sequence(phi)::MixedBlowUpSequence
+        f_initialized = true
+      else
+        f = _do_blow_up!(f,I)
+        if length(decomp)>0 
+          decomp = [strict_transform(last_map(f),J) for J in decomp]
+        end
+      end
+    end
+    I_sl_temp = ideal_sheaf_of_singular_locus(domain(last_map(f)))
```
HechtiDerLachs, line 620: You can add the exceptional locus here to make sure you don't get duplicate components which are already coming up from `strict_transform` above. Eventually, we should think about providing some extra argument to specify that `ideal_sheaf_of_singular_locus` is supposed to only look for singularities along the support of another ideal sheaf.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Methods.jl
```diff
@@ -250,8 +250,8 @@ function normalization(X::AbsCoveredScheme; check::Bool=true)
   pr_dict = empty(first(pr_dicts))
   merge!(pr_dict, pr_dicts...)
   pr_covering_mor = CoveringMorphism(default_covering(Y), default_covering(X), pr_dict; check=false)
-  pr_mor = CoveredSchemeMorphism(Y, X, pr_covering_mor; check=false)
-  return Y, pr_mor, injs
+  pr_mor = Oscar.NormalizationMorphism(Y, X, pr_covering_mor,injs; check=false)
+  return Y, pr_mor
```
HechtiDerLachs:
```suggestion
  pr_mor = CoveredSchemeMorphism(Y, X, pr_covering_mor; check=false)
  return Y, NormalizationMorphism(pr_mor, injs; check=false)
```

HechtiDerLachs (changes requested):
@simonbrandhorst was right. I left some suggestions which hopefully lead to running code which makes the inheritance work in the anticipated way.

experimental/Schemes/Resolution_structure.jl
HechtiDerLachs: The idea is that the constructor does nothing but rewrapping an existing morphism and storing extra data.

experimental/Schemes/CoveredScheme.jl
afkafkafk13: I reverted this change.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

HechtiDerLachs:
One thing I forgot to explain: There was a `BaseMorType` floating around. This is different from `Nothing` only if there is a base change (as for passing from `QQ` to a field extension). This will never be the case for the morphisms considered here, so it can be dropped. This dropping is done by setting the corresponding type parameter in `AbsCoveredSchemeMorphism` to `Nothing` when deriving the concrete type from it.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Currently (while WIP), I use small_generating_set locally as a testing ground -- without messing up simplify and all the instances where it is applied. In the end, this should, of course, be a simplify (possibly with something changed in a suitable way).

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

afkafkafk13:
> One thing I forgot to explain: There was a `BaseMorType` floating around. This is different from `Nothing` only if there is a base change (as for passing from `QQ` to a field extension). This will never be the case for the morphisms considered here, so it can be dropped. This dropping is done by setting the corresponding type parameter in `AbsCoveredSchemeMorphism` to `Nothing` when deriving the concrete type from it.

Ah. This was für BaseChange. That had not occurred to me. Thank you for the explanation.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Methods.jl
```diff
@@ -251,7 +251,8 @@ function normalization(X::AbsCoveredScheme; check::Bool=true)
   merge!(pr_dict, pr_dicts...)
   pr_covering_mor = CoveringMorphism(default_covering(Y), default_covering(X), pr_dict; check=false)
   pr_mor = CoveredSchemeMorphism(Y, X, pr_covering_mor; check=false)
-  return Y, pr_mor, injs
+  pr_mor = Oscar.NormalizationMorphism(pr_mor,injs; check=false)
```
simonbrandhorst, line 283:
```suggestion
  pr_mor = NormalizationMorphism(pr_mor,injs; check=false)
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,54 +1,114 @@
...
+## of disjoint components
+##############################################################################
+@attributes mutable struct NormalizationMorphism{
+    DomainType<:AbsCoveredScheme,
+    CodomainType<:AbsCoveredScheme
+   } <:AbsCoveredSchemeMorphism{
+                                 DomainType,
+                                 CodomainType,
+                                 Nothing,
+                                 NormalizationMorphism,
```
simonbrandhorst, line 38: This looks a bit circular. What is going on?

afkafkafk13: AbsCoveredSchemeMorphism needs 4 arguments: domain, codomain, base change (here nothing) and self.

HechtiDerLachs: The requirements for implementing a `Hecke.Map` is what's going on here. One argument is the type itself.

afkafkafk13:
@HechtiDerLachs : On my local copy the tests run just fine, but I am behind master. In particular, your changes to make the IdealSheaves lazy are not present on my local copy. I suppose that I need to (A) rebase and (B) switch to the lazy IdealSheaves. But I will need help for (B) in any case.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: This has been streamlined now.

HechtiDerLachs:
> (B) switch to the lazy IdealSheaves

As a general rule of thumb: Replace any occurence of `IdealSheaf` by `AbsIdealSheaf` (except for constructor calls) and make sure that `Vector`s of ideal sheaves are initialized correctly as vectors of type `Vector{AbsIdealSheaf}` (e.g. by writing `my_vector = AbsIdealSheaf[]` instead of `my_vec = typeof(I)[]`). Then everything should work.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: The preceding loop exhausted all components of the singular locus as centers of blow-ups, so there is no risk of getting duplicates.

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,61 +1,201 @@
...
+  # always carried along to domain(maps[end])) using strict_transform
+  ex_div::Vector{AbsIdealSheaf}      # list of exc. divisors arising from individual steps
+
+  # keep track of the normalization steps
+  normalization_steps::Vector{Int}
+
+  # fields for caching to be filled a posteriori (on demand, only if partial_res==false)
+  underlying_morphism::CompositeCoveredSchemeMorphism{DomainType, CodomainType}
+  exceptional_divisor::WeilDivisor
+  exceptional_locus::AlgebraicCycle
```
HechtiDerLachs:
```suggestion
  exceptional_divisor::AbsWeilDivisor
  exceptional_locus::AbsAlgebraicCycle
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,61 +1,201 @@
...
+  underlying_morphism::CoveredSchemeMorphism
+  inclusions::Vector{<:AbsCoveredSchemeMorphism}
+
+  function NormalizationMorphism(
+      f::CoveredSchemeMorphism,
+      inclusions::Vector{<:AbsCoveredSchemeMorphism};
+      check::Bool=true
+    ) 
+    @check is_normal(X) "not a normalization morphism"
+    @check all(inc->codomain(inc) === domain(f), inclusions) "domains and codomains do not match"
```
HechtiDerLachs:
```suggestion
    @assert all(inc->codomain(inc) === domain(f), inclusions) "domains and codomains do not match"
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,61 +1,201 @@
...
+
+@attr CartierDivisor function exceptional_divisor_with_multiplicities(f::BlowUpSequence)
+  is_embedded || error("only available for embedded desingularization")
+  phi.transform_type != :strict || error("only available for weak and controlled transforms")
+
+  ex_div_list = exceptional_divisor_list(f)
+  C = f.ex_mult[1] * cartier_divisor(ex_div_list[1])
+  for i in 2:length(ex_div_list)
+    C = C + f.ex_mult[i] * cartier_divisor(ex_div_list[i])
+  end
```
HechtiDerLachs:
```suggestion
  C = sum(f.ex_mult[i] * cartier_divisor(ex_div_list[i]) for i in 1:length(ex_div_list))
```

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,61 +1,201 @@
...
+  !isdefined(f,:exceptional_divisor_on_X) || return f.exceptional_divisor_on_X
+
+  ex_div_list = exceptional_divisor_list(f)
+  C = WeilDivisor(scheme(ex_div_list[1]),ZZ)
+  for i in 2:length(ex_div_list)
+    dim(ex_div_list[i])== -1 && continue            # kick out empty ones
+    C = C + weil_divisor(ex_div_list[i])
+  end
+
+  f.exceptional_divisor = C
```
HechtiDerLachs, line 258:
```suggestion
  f.exceptional_divisor_on_X = C
```
At least this is suggested by the first line of the method?

experimental/Schemes/Resolution_structure.jl
```diff
@@ -67,36 +207,72 @@ function add_map!(f::BlowUpSequence, phi::BlowupMorphism)
...
 end
 
-function initialize_blow_up_sequence(phi::BlowupMorphism)
-  f = BlowUpSequence([phi])
-  f.ex_div = [exceptional_divisor(phi)]
-  f.is_trivial = is_one(center(phi))
-  f.resolves_sing = false                                # we have no information, wether we are done
-                                                         # without further computation
-  f.is_embedded = false
+function add_map!(f::MixedBlowUpSequence, phi::BlowupMorphism)
```
HechtiDerLachs, line 315: These should also update the `underlying_morphism` if it is defined, correct?

HechtiDerLachs (approved):
Looks good already. I marked some small parts which made me skeptic. But I would like to see this in action.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Indeed.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: Yes, after removing the debug output, I can now contract this to a single line.

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: done

afkafkafk13:
@HechtiDerLachs , @simonbrandhorst : There is still documentation to come, but otherwise I put in, what I wanted to have in this PR. 
(aside: The `Oscar.` will be removed with the documentation commit some time tomorrow.)

experimental/Schemes/Resolution_structure.jl
```diff
@@ -1,61 +1,209 @@
-## Warnung: show auf desingMor geht noch nicht!!!
 export _desing_curve
```
lgoettgens:
```suggestion
```
Internal functions (as indicated by the leading underscore) should not get exported

afkafkafk13: done

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Methods.jl
afkafkafk13: done

experimental/Schemes/Resolution_structure.jl
afkafkafk13: I will leave this in for now. It will be changed in a future PR, as soon as the desingularizations and principalizations have made more progress.

afkafkafk13:
Further documentation will be in a separate PR. 
I would like to have this here finished soon so that the files can be moved to the correct structure of experimental.

afkafkafk13:
@simonbrandhorst : I incorporated all changes you requested. Could you have another look?

simonbrandhorst (approved):

PR #3640, opened 2024-04-23
Add `reduce(::MPolyRingPoly, ::IdealGens)` etc.

joschmitt:
Adds the variants `reduce(::T, ::IdealGens{T}; ...)` and `reduce(::Vector{T}, ::IdealGens{T})` and similar for `reduce_with_quotients` and `reduce_with_quotients_and_unit`.

Without this the following would not work:
```
julia> R, (x, y) = QQ["x", "y"];

julia> I = ideal(R, [y^2 - x, x^3 - 2*y^2]);

julia> G = groebner_basis(I);

julia> reduce(x, G)
x
```
(It works by calling `reduce` with `collect(G)`, but then the information on the Singular side is forgotten and has to be recreated.)

### DIFF
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -646,14 +646,14 @@ function reduce(I::IdealGens, J::IdealGens; ordering::MonomialOrdering = default
 end
 
 @doc raw"""
-    reduce(g::T, F::Vector{T}; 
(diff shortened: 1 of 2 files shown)


simonbrandhorst (approved):

jankoboehm:
Looks good!

PR #3649, opened 2024-04-26
call GAP's `ImagesRepresentative` instead of `Image`

ThomasBreuer:
The point is that the mapping on the GAP side may be not single-valued. (For example, it can be an "inverse" of an epimorphism.)

Oscar uses `preimage` with the meaning of GAP's `PreImagesRepresentative`, thus `ImagesRepresentative` should be used in the other direction.

### DIFF
--- a/src/GAP/wrappers.jl
+++ b/src/GAP/wrappers.jl
@@ -108,6 +108,7 @@ GAP.@wrap HasSize(x::Any)::Bool
 GAP.@wrap Identity(x::GapObj)::GapObj
 GAP.@wrap Image(x::Any)::GapObj
 GAP.@wrap Image(x::Any, y::Any)::GapObj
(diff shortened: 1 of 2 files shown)


fingolfin:
Your description ends with "In the same way as", is there something missing?

fingolfin (approved):

fingolfin:
Can we perhaps add a test which tests this, i.e. which fails before this PR?

ThomasBreuer:
> Can we perhaps add a test which tests this, i.e. which fails before this PR?

I got a piece of an Oscar session from @fieker which shows that the problem can occur, but I have not been able to construct input for this situation.
(Yes, it is likely that something else went wrong in such a situation, but I think that conceptually the proposed change is correct.)

PR #3656, opened 2024-04-29
Some convenience methods to push rational points around

HechtiDerLachs:
@simonbrandhorst : As suggested on Slack. I will write another follow-up method for extraction of a Mordell-Weil-basis from a pushforward.


HechtiDerLachs:
This seems to work rather fine now and I think it can be very useful for us. Let me know what you think.

simonbrandhorst:
Certainly very useful. Please add 
https://arxiv.org/pdf/2103.15101
Algorithm 1 and Lemma 2.20. 
Then you can get rid of the restriction that the divisor must be a section.

simonbrandhorst (approved):

simonbrandhorst:
It it works, it works

PR #3657, opened 2024-04-29
Use more `fat_scheme` for `AbsProjectiveAlgebraicSet`

### LINKED ISSUE #3653: Printing changes attributes
```
R = graded_polynomial_ring(QQ, [:x])
quo(R[1], R[2][1]^2)
x = projective_scheme(ans[1])
xx = ProjectiveAlgebraicSet(x);
```
then `X.Xred` is undefined as it should be, but doing
```
julia> xx
```
ie. printing it, will set `X.Xred` after an expensive computation. I suspect this is due to `Oscar.underlying_scheme` being called - which triggers the computation.
The severe bug is that printing should never (substantially) change any structure...
(this confused @wdecker  - as clearly it was set, despite the constructor not doing it)

StevellM:
Should fix #3653

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Types.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineAlgebraicSet/Objects/Types.jl
@@ -42,14 +42,11 @@ defined by ideal(y, x)
   function AffineAlgebraicSet(X::AffineScheme; is_reduced::Bool=false, check::Bool=true)
     A = new{typeof(base_ring(X)), typeof(OO(X))}()
     A.X = X
-    if check && is_reduced
-      is_geometrically_reduced(X) || error("Algebraic sets must be geometrically reduced")
-    end
     if is_reduced
+      @check is_geometrically_reduced(X) "Algebraic sets must be geometrically reduced"
       A.Xred = X
       set_attribute!(A.Xred, :is_geometrically_reduced, true)
       set_attribute!(A.Xred, :is_reduced, true)
-      # unlock the scheme methods for geometrically reduced schemes.
     end
     return A
   end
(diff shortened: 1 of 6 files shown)


simonbrandhorst (approved):

simonbrandhorst:
I suppose some of the newly added tests for `elliptic_surface` oomed the worker. I can disable them

simonbrandhorst:
@benlorenz  
Can we see in the log which test exactly it is? 
And would you know how much memory our long test MacOS worker has?  Apparently the long tests run fine with Linux.

thofma:
@benlorenz with `macOS-latest` being equal to `macOS-14` now, I think we have less resources for the long tests. I also noticed this for Hecke.

benlorenz:
The arm64 macos runners only have 7GB memory.

But the same error happened in the `without experimental` job which runs on Ubuntu with 16GB memory.
This also happened on master and a few other PRs:

1.11 ubuntu
https://github.com/oscar-system/Oscar.jl/actions/runs/8878829055/job/24375248112#step:8:2329

1.10 ubuntu
https://github.com/oscar-system/Oscar.jl/actions/runs/8876295842/job/24367499457#step:8:412
https://github.com/oscar-system/Oscar.jl/actions/runs/8874218248/job/24361288406#step:8:408
https://github.com/oscar-system/Oscar.jl/actions/runs/8868048924/job/24347226536#step:8:424

But it does not really look like running out of memory, for julia 1.11 the log ends with:
```
Mon, 29 Apr 2024 13:20:11 GMT
GC: pause 316.79ms. collected 662.935200MB. incr 
Mon, 29 Apr 2024 13:20:11 GMT
Heap stats: bytes_mapped 3072.75 MB, bytes_resident 2359.02 MB,
Mon, 29 Apr 2024 13:20:11 GMT
heap_size 2943.99 MB, heap_target 3091.19 MB, Fragmentation 0.519
Mon, 29 Apr 2024 15:18:46 GMT
Error: The operation was canceled.
```
The heap size was 3GB for the last GC pause and then 2 hours later the job was killed.

HechtiDerLachs:
The tests failing here were probably recently added by me. But I'm wondering why they fail only on MacOS. I had them run also on my local machine with only 8GB of RAM, so that should not really be an issue. But we could disable them again if that makes things easier. 

Another thing: Since this is not merged yet, could we maybe also take care of the regression discovered by @fingolfin on #3653 ?

PR #3659, opened 2024-04-29
Improve _normal_form_singular: accept more inputs, allocate less

### LINKED ISSUE #3658: Reducing a non-homogenous polynomial in a graded polynomial ring
It seems I am not allowed to reduce an arbitrary (graded) polynomial by an ideal of a graded polynomial ring; it gives an error as in the example below:
```
julia> R,(a,b,c) = graded_polynomial_ring(QQ,["a","b","c"]);

julia> J = ideal(R,[a^2 + b^2, c^2]);

julia> normal_form(a^2 + c^2, J)   # works
-b^2

julia> normal_form(a^2 + c, J)     # doesn't work
ERROR: ArgumentError: The generators of the ideal must be homogeneous
Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/Hecke/jLTFH/src/Assertions.jl:561 [inlined]
  [2] MPolyIdeal(B::Oscar.IdealGens{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/Rings/mpoly.jl:629
  [3] MPolyIdeal
    @ ~/Projekte/OSCAR/Oscar.jl/src/Rings/mpoly.jl:614 [inlined]
  [4] ideal(Rx::MPolyDecRing{QQFieldElem, QQMPolyRing}, s::Singular.sideal{Singular.spoly{Singular.n_Q}})
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/Rings/mpoly.jl:648
  [5] _normal_form_singular(A::Vector{…}, J::MPolyIdeal{…}, ordering::MonomialOrdering{…})
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/Rings/groebner.jl:1163
  [6] normal_form(A::Vector{…}, J::MPolyIdeal{…}; ordering::MonomialOrdering{…})
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/Rings/groebner.jl:1066
  [7] normal_form
    @ ~/Projekte/OSCAR/Oscar.jl/src/Rings/groebner.jl:1061 [inlined]
  [8] normal_form(f::MPolyDecRingElem{…}, J::MPolyIdeal{…}; ordering::MonomialOrdering{…})
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/Rings/groebner.jl:1056
  [9] normal_form(f::MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}, J::MPolyIdeal{MPolyDecRingElem{…}})
    @ Oscar ~/Projekte/OSCAR/Oscar.jl/src/Rings/groebner.jl:1055
 [10] top-level scope
    @ REPL[20]:1
```

But why is that? Mathematically it seems absolutely fine to me?

fingolfin:
Instead of creating a fresh Singular ring each time, use the one we already have.

Fixes #3658 for me.

No tests yet. I am happy to add some, but first I want to know that this PR is OK, i.e., there is no deeper reason why normal forms of non-homogeneous polynomials can't be computed right now.

It also reveals a deeper issue: this function got faster with my change,  but it is still makes a *ton* of computations. E.g. to convert a single polynomial:
```
julia> @time Oscar._normal_form_singular([a^2+b^2+c^2], J, ordering);
  0.000100 seconds (136 allocations: 3.469 KiB)
```
So there are 136 allocations. Of these 125 are for converting that polynomial from Oscar to Singular, and the result back. That's IMHO way too much. I will open a separate issue for that.

Ping @wdecker @simonbrandhorst @afkafkafk13 @jankoboehm @HechtiDerLachs @ederc

### DIFF
--- a/src/Rings/groebner.jl
+++ b/src/Rings/groebner.jl
@@ -1157,11 +1157,13 @@ julia> Oscar._normal_form_singular(A, J, default_ordering(base_ring(J)))
 ```
 """
 function _normal_form_singular(A::Vector{T}, J::MPolyIdeal, ordering::MonomialOrdering) where { T <: MPolyRingElem }
-  SR = singular_poly_ring(base_ring(J), ordering)
-  IS = Singular.Ideal(SR, [SR(x) for x in A])
   GS = singular_groebner_generators(J, ordering)
-  K  = ideal(base_ring(J), reduce(IS, GS))
-  return [J.gens.Ox(x) for x = gens(K.gens.S)]
+  SR = base_ring(GS)
+  tmp = map(SR, A)
+  IS = Singular.Ideal(SR, tmp)
+  K = reduce(IS, GS)
+  OR = base_ring(J)
+  return map(OR, gens(K))
 end
 
 @doc raw"""
--- a/test/Rings/groebner.jl
+++ b/test/Rings/groebner.jl
@@ -122,6 +122,13 @@
     @test_throws ArgumentError standard_basis_highest_corner(I)
 end
 
+@testset "normal form graded" begin
+    R, (a, b, c) = graded_polynomial_ring(QQ, ["a", "b", "c"])
+    I = ideal([a^2+b^2, c^3])
+    # verify normal form of inhomogeneous polynomial work
+    @test normal_form(a^2+b, I) == -b^2+b
+end
+
 @testset "groebner leading ideal" begin
    R, (t, x, y, z) = polynomial_ring(QQ, ["t", "x", "y", "z"])
 


src/Rings/groebner.jl
```diff
@@ -1157,11 +1157,13 @@ julia> Oscar._normal_form_singular(A, J, default_ordering(base_ring(J)))
 ```
 """
 function _normal_form_singular(A::Vector{T}, J::MPolyIdeal, ordering::MonomialOrdering) where { T <: MPolyRingElem }
-  SR = singular_poly_ring(base_ring(J), ordering)
-  IS = Singular.Ideal(SR, [SR(x) for x in A])
-  GS = singular_groebner_generators(J, ordering)
-  K  = ideal(base_ring(J), reduce(IS, GS))
-  return [J.gens.Ox(x) for x = gens(K.gens.S)]
+  GS = Oscar.singular_groebner_generators(J, ordering)
```
joschmitt:
```suggestion
  GS = singular_groebner_generators(J, ordering)
```
I assume?

ederc (approved):

afkafkafk13:
@ederc : Just out of curiosity -- Things like redTail etc. are explicitly set in the reduce call or are they inherited from the Singular ring, if we reuse it?

ederc:
`complete_reduction` is set to be `true` in the implementation in Singular.jl by default. This is also documented: ... _the fully reduced remainder_ ...  in the documentation of `nomal_form` in Oscar. This behaviour was asked for when I implemented `normal_form`.

afkafkafk13:
> `complete_reduction` is set to be `true` in the implementation in Singular.jl by default. This is also documented: ... _the fully reduced remainder_ ... in the documentation of `nomal_form` in Oscar. This behaviour was asked for when I implemented `normal_form`.

Thanks for the explanation.

fingolfin:
I've now added an (admittedly very basic) test, so from my POV this is good to go (assuming it passes CI)

PR #3660, opened 2024-04-29
Temporarily change the default ordering

### LINKED ISSUE #1487: Preferred ordering for ideal computations
We have an example from @AlheydisGeiger where computing a saturation with Singular(.jl) ideals is much faster than with Oscar ideals, because the provided ordering made everything faster. At the moment we have no way to have this as fast in Oscar, which is a shame. I think it would make sense to have a way to specify a possible ordering when doing ideal computations. This is for experienced users, who needs more control over these things.

So more concretely, I would propose to have a method
```julia
saturation(I::MPolyIdeal, J::MPolyIdeal; ordering = nothing)
```
and the docstring explains that by default the system will choose a default ordering and otherwise use the ordering provided.

Any other ideas or suggestions on how to tackle this problem?

CC: @tthsqe12 @wdecker @ederc

joschmitt:
Introduces a function `with_ordering(f, ::MPolyRing, ::MonomialOrdering)` that allows to set the default ordering of a ring for the execution of a function (or `do` block).
For example, in #1487 one could now do:
```
n = 6
R, c = polynomial_ring(QQ,vcat(["c$i" for i in 1:n+7]))
x, y, z, u, v, w, t = c[n + 1:n + 7];
p = c[1] + c[2]*y + c[3]*y*z + c[4]*y^2 + c[5]*x + c[6]*x*y^4*z
q = evaluate(p,[x,y,z],[u,v,w])
I = ideal(R,[p,q,derivative(p,x),derivative(p,y),derivative(p,z),derivative(q,u),derivative(q,v),derivative(q,w),1-t*x*y*z*u*v*w])
with_ordering(R, degrevlex(gens(R)[1:6])*degrevlex(gens(R)[7:13])) do
  saturation(I, ideal(R,[x-u,y-v,z-w]))
end
```

This is certainly only intended for expert users who know what they do. I'm open to other solutions, but I really want to have some functionality like this.

This closes #1487 from my point of view. It still doesn't help if one doesn't know the ring because it is created several levels down, but I don't see what we can do in this case anyway.

### DIFF
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -120,8 +120,11 @@ z^2 + z
 function with_ordering(f, R::MPolyRing, o::MonomialOrdering)
   old = default_ordering(R)
   set_default_ordering!(R, o)
-  x = f()
-  set_default_ordering!(R, old)
+  x = try
+    f()
+  finally
+    set_default_ordering!(R, old)
+  end
   return x
 end
 
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -63,10 +63,67 @@ using .Orderings
 #type for orderings, use this...
 #in general: all algos here needs revision: do they benefit from gb or not?
 
+@doc raw"""
+    default_ordering(R::MPolyRing)
+
+Return the monomial ordering that is used for computations with ideals in `R`
+if no other ordering is specified -- either directly by the user or by
+requirements of a specific algorithm.
+"""
 @attr MonomialOrdering{T} function default_ordering(R::T) where {T<:MPolyRing}
   return degrevlex(R)
 end
 
+# Only for internal use
+function set_default_ordering!(R::MPolyRing, o::MonomialOrdering)
+  @assert R === base_ring(o)
+  set_attribute!(R, :default_ordering, o)
+  return nothing
+end
+
+@doc raw"""
+    with_ordering(f, R::MPolyRing, o::MonomialOrdering)
+
+Use the monomial ordering `o` for computations in `R` during the execution of
+`f`.
+This may be used with `do` block syntax, see the example.
+
+This functionality is meant for advanced users. In general it should not be
+necessary to explicitly set a monomial ordering.
+Further, there is no guarantee that `o` is actually used. For example, if an
+algorithm requires an elimination ordering, `o` might be ignored.
+
+# Example
+```jldoctest withordering
+julia> R, (x, y, z) = QQ["x", "y", "z"];
+
+julia> f = x + y^2;
+
+julia> I = ideal(R, [y^2 - z, x - z^2]);
+
+julia> normal_form(f, I) # this uses degrevlex
+x + z
+
+julia> with_ordering(R, lex(R)) do
+           # this uses lex
+           normal_form(f, I)
+       end
+z^2 + z
+```
+Notice that in this small example we could have achieved the same by using the
+keyword argument `ordering`:
+```jldoctest withordering
+julia> normal_form(f, I, ordering = lex(R))
+z^2 + z
+```
+"""
+function with_ordering(f, R::MPolyRing, o::MonomialOrdering)
+  old = default_ordering(R)
+  set_default_ordering!(R, o)
+  x = f()
+  set_default_ordering!(R, old)
+  return x
+end
 
 mutable struct BiPolyArray{S}
   Ox::NCRing #Oscar Poly Ring or Algebra
@@ -301,7 +358,7 @@ function singular_generators(B::IdealGens, monorder::MonomialOrdering=default_or
 end
 
 @doc raw"""
-set_ordering(I::IdealGens, monord::MonomialOrdering)
+    set_ordering(I::IdealGens, monord::MonomialOrdering)
 
 Return an ideal generating system with an associated monomial ordering.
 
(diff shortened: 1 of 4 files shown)


fingolfin (dismissed):
This seems sensible to me.

We *definitely* want the change making `default_ordering` type stable.

thofma:
We went through great pains to not make orderings part of the polynomial ring and now we reintroduce it? We could have saved a lot of time and discussions.

fieker:
It's a bad name, what about `with_ordering(Ring, ordering) do` this would semantically not imply that we change any ring at all (that would be magic behind the scene) and i captures more what the original intend is: do this one operation using this ordering.

wdecker:
I agree with @fieker.Von meinem iPhone gesendetAm 30.04.2024 um 08:21 schrieb Claus Fieker ***@***.***>:﻿
It's a bad name, what about with_ordering(Ring, ordering) do this would semantically not imply that we change any ring at all (that would be magic behind the scene) and i captures more what the original intend is: do this one operation using this ordering.

joschmitt:
> We went through great pains to not make orderings part of the polynomial ring and now we reintroduce it? We could have saved a lot of time and discussions.

I'm aware that the first rule about Gröbner bases in OSCAR is that nobody wants to see Gröbner bases.
The current situation is that (almost) every operation is done with respect to degrevlex or an appropriate weight ordering if the ring is positively graded. This was already an improvement to the situation before where it was just always degrevlex.

Here are three real-life examples where the current situation is sub-optimal in my opinion:
a) The saturation example in #1487.
b) My personal motivation: I want to compute preimages under a morphism of polynomial rings. Almost all of the images are homogeneous, so using a weight ordering seems to be (much) faster.
c) I was asked by Tobias Metzlaff whether there is a way to set the monomial ordering used in `fundamental_invariants`. (In this case there is also the `reduce`-`divrem` schism but that's another problem.)

In https://github.com/oscar-system/Oscar.jl/issues/1487#issuecomment-1199288100 there were 4 possible solutions presented:
> 1. Have a preferred ordering attached to the polynomial ring. Probably a non-starter because of == issues for polynomial rings.

More or less what is proposed here. We can discuss names and other details. I don't know what these `==` issues are here.

> 2. Attach a preferred ordering to ideals. In the case of a clash, choose one deterministically.

This wouldn't work in cases b) and c) because the user never gets in contact with an ideal. Also in case a) it is not clear which ideal I need to attach the ordering to or both or what.

> 3. Have a keyword ordering argument to each function that deals with a GB. This seems like it will become cumbersome for all parties involved.

Maybe we could do this, but well. It will become cumbersome.

> 4. Have the system scan the input and automatically choose The Best(TM) ordering.

I'm not sure how serious this was, but let's just pretend it is. Of course, we should always do the best we can. In cases a) and b) one could *maybe* scan the input to recognize block patterns or homogeneity etc. But in the end people want to do research and play around like in c).

My personal opinion is that OSCAR is software (also) for professionals who sometimes *do* want to see Gröbner bases or know better than any heuristic that we can think of right now. And these people should get the tools to do what they want.
If these points were discussed and decided against: I would like to hear the arguments, but fine by me.

I mark this as draft until this fundamental discussion is finished.

joschmitt:
> It's a bad name, what about `with_ordering(Ring, ordering) do` this would semantically not imply that we change any ring at all (that would be magic behind the scene) and i captures more what the original intend is: do this one operation using this ordering.

I was trying to find a better name, `with_ordering` sounds good. The caveat is that the ordering will not necessarily be used, it is just the default fall-back. The go-to example would be elimination, I guess. Probably we can assume that people who specify monomial orderings know about this stuff though.

fieker:
Oscar was always also meant to allow experts to do whatever they want - but also to make stuff available for non-experts, hence this fight against Groebner bases. It forced also the experts to think more about math (and mathematical interfaces) and less about specific algorithms. At the end, at bleeding edge, we need to and want to provide the expert with all tools. The question how, in particular, how without starting down the slippery slope of having to understand orderings before I can use commutative algebra...

Difficult, and of course, not restricted to comm. algebra. We're having the same conflict elsewhere...

joschmitt:
What is the status? Does this need to be discussed further (if so, by whom)?
Or is there a decision in either direction?

src/Rings/mpoly.jl
```diff
@@ -78,10 +78,67 @@ using .Orderings
...
+julia> normal_form(f, I, ordering = lex(R))
+z^2 + z
+```
+"""
+function with_ordering(f, R::MPolyRing, o::MonomialOrdering)
+  old = default_ordering(R)
+  set_default_ordering!(R, o)
+  x = f()
+  set_default_ordering!(R, old)
+  return x
```
fingolfin: This should use `try/finally` to ensure the ordering is really reset, even if `f` throws an error. Something like this
```suggestion
  old = default_ordering(R)
  try
    set_default_ordering!(R, o)
    return f()
  finally
    set_default_ordering!(R, old)
  end
```

fingolfin:
I still think this is fine (modulo the purely technically remark about try/finally I just added). Note that all those default orderings on the polynomial ring etc. etc. are already there, this PR is just about adding a *better* API which is focused on *temporarily* changing the default ordering (the existing API, `set_default_ordering!`, only allows for permanent changes, which is precisely what we don't really want users to do in general). 

But let's discuss it in triage.

A remark beyond this PR: `set_default_ordering!` also has methods for `SubquoModule` and `SubModuleOfFreeModule` -- perhaps those also should get `with_ordering` counterparts? But I am happy to wait with that until someone needs them.

src/Rings/mpoly.jl
joschmitt: Good point!

fingolfin (approved):

PR #3663, opened 2024-04-30
Refactor deprecations

lgoettgens:
For types, use `Base.@deprecate_binding`, while for functions and methods use `@deprecate`.

Furthermore, added a comment between deprecations in 1.0 and accumulated ones for 1.1

### DIFF
--- a/docs/src/DeveloperDocumentation/styleguide.md
+++ b/docs/src/DeveloperDocumentation/styleguide.md
@@ -338,12 +338,12 @@ printing in Oscar](@ref).
 
 Sometimes it is necessary to rename a function or otherwise change it. To allow
 for backwards compatibility, please then introduce a new line in the file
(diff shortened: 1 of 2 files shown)


joschmitt:
Could this knowledge about `@deprecate` and `@deprecate_binding` be written down somewhere?

lgoettgens:
> Could this knowledge about `@deprecate` and `@deprecate_binding` be written down somewhere?

This should be part of https://github.com/JuliaLang/julia/issues/51360. But I can add two sentences about it to the top of the deprecations.jl file

joschmitt:
Maybe here https://docs.oscar-system.org/stable/DeveloperDocumentation/styleguide/#Deprecating-functions would be good. But if this will be documented by Julia that's also fine.

joschmitt:
Great, thank you for the documentation!

fingolfin (approved):

PR #3664, opened 2024-04-30
add `linear_characters`, extend syntax of `induce`, `restrict`

ThomasBreuer:
items from @fieker's wishlist


fieker (approved):
You know what you're doing. The interface looks good to me. Are you computing all characters to return the linear ones?

fingolfin (approved):

ThomasBreuer:
@fieker 
> Are you computing all characters to return the linear ones?

The method delegates to GAP.
GAP checks whether the irreducibles are known.
If yes then it returns the linear ones.
If not then it delegates to the commutator factor group.

PR #3667, opened 2024-05-01
Bump dependencies

### LINKED ISSUE #1722: Ambiguity in QQBarField conversion
Observed in https://github.com/Nemocas/AbstractAlgebra.jl/actions/runs/8740738993/job/23985171446?pr=1677#step:7:934
```julia
MethodError: (::QQBarField)(::Polymake.LibOscarNumber.OscarNumberAllocated) is ambiguous.
  
  Candidates:
    (a::QQBarField)(b)
      @ Nemo ~/work/AbstractAlgebra.jl/AbstractAlgebra.jl/oscar-dev/Nemo/src/calcium/qqbar.jl:1523
    (F::Field)(x::Polymake.LibOscarNumber.OscarNumber)
      @ Oscar ~/work/AbstractAlgebra.jl/AbstractAlgebra.jl/oscar-dev/Oscar/src/PolyhedralGeometry/helpers.jl:217
  
  Possible fix, define
    (::QQBarField)(::Polymake.LibOscarNumber.OscarNumber)
```
This error will occur in every OscarCI run until fixed.

I think this was introduced in https://github.com/Nemocas/Nemo.jl/pull/1702, but only exercised once https://github.com/oscar-system/Oscar.jl/pull/3512 was merged.

@fingolfin @thofma should we fix this here or in Oscar? In Nemo, this would be a revert of some parts of #1702. In Oscar, we could add a disambiguation function like proposed in the error message. Which of the two do you prefer?

lgoettgens:
cc @thofma @fingolfin 

Closes https://github.com/Nemocas/Nemo.jl/issues/1722.

Waiting for:
- https://github.com/Nemocas/AbstractAlgebra.jl/pull/1699
- https://github.com/Nemocas/AbstractAlgebra.jl/pull/1700
- https://github.com/thofma/Hecke.jl/pull/1488
- https://github.com/thofma/Hecke.jl/pull/1491


lgoettgens:
I now added `base_ring_type` methods for many of the already existing `base_ring` methods. In algebraic geometry, I got a bit lost and skipped some. So let's hope that CI passes anyway.

lgoettgens:
I know of some other stuff that needs to be adjusted here. Possibly getting to it tomorrow

fingolfin:
Needs a rebase anyway

fingolfin:
It seems this does not yet have the `QQBarField` disambiguation method for Polymake code

fingolfin:
I merged master (hopefully resolved the conflicts in `src/aliases.jl` correctly) and added a QQBarFieldElem disambiguation.

However there is still a precompilation "warning" / error:

```
WARNING: Method definition hom(Hecke.FinGenAbGroup, Hecke.FinGenAbGroup, Array{var"#s685", 1} where var"#s685"<:(AbstractAlgebra.Map{Hecke.FinGenAbGroup, Hecke.FinGenAbGroup, S, T} where T where S)) in module Hecke at /Users/mhorn/.julia/packages/Hecke/56bDu/src/GrpAb/GrpAbFinGen.jl:990 overwritten in module Misc at /Users/mhorn/Projekte/OSCAR/Oscar.jl/experimental/GModule/Misc.jl:257.
```

Looking at the two methods, they indeed have identical signature but one implements a tensor produce and one a direct product?!

fingolfin:
(CC @ThomasBreuer and @fieker as they may know best what to do about the `hom` method in `experimental/GModule/Misc.jl` vs. that in Hecke)

lgoettgens:
> (CC @ThomasBreuer and @fieker as they may know best what to do about the `hom` method in `experimental/GModule/Misc.jl` vs. that in Hecke)

This error is due to me renaming this from `direct_sum` to `hom` in spite of https://github.com/Nemocas/AbstractAlgebra.jl/pull/1675#issuecomment-2066060234. We could just merge both of these functions to a single one that decides which construction to use depending on the if `G` is a direct sum or tensor product.

See https://github.com/thofma/Hecke.jl/pull/1483 for the proposal.

thofma:
I think it is problematic that we now have to write methods for objects of some `T` and then we have to distinguish between how they were constructed. It is rather error prone.

This will only get worse, since there are also `hom(G, H, vector)`, where `H` is a direct product and `vector` homomorphisms of `G` into the single components of `H` (aka universal property).

Clearly if I intend to do `hom(G, H, vector_of_morphisms)`, I just constructed `G` and `H` as say tensor products, and then `hom_tensor_product(G, H, vector_of_morphisms)` makes more sense I would say. (Or some name containing the word "tensor").

lgoettgens:
> I think it is problematic that we now have to write methods for objects of some `T` and then we have to distinguish between how they were constructed. It is rather error prone.
> 
> This will only get worse, since there are also `hom(G, H, vector)`, where `H` is a direct product and `vector` homomorphisms of `G` into the single components of `H` (aka universal property).
> 
> Clearly if I intend to do `hom(G, H, vector_of_morphisms)`, I just constructed `G` and `H` as say tensor products, and then `hom_tensor_product(G, H, vector_of_morphisms)` makes more sense I would say. (Or some name containing the word "tensor").

That was exactly my point in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1675#discussion_r1570864990. Let's wait for @ThomasBreuer and @fieker to comment on this. In any way, the functions for FinGenAbGrp and FPModule should be named the same, so in the worst case we need to rename the thing in AA again.

src/Rings/MPolyMap/MPolyRing.jl
```diff
@@ -53,8 +53,8 @@ julia> R, (x, y) = polynomial_ring(K, ["x", "y"]);
 
 julia> F = hom(R, R, z -> z^2, [y, x])
 Ring homomorphism
-  from multivariate polynomial ring in 2 variables over GF(2, 2)
-  to multivariate polynomial ring in 2 variables over GF(2, 2)
+  from multivariate polynomial ring in 2 variables over K
+  to multivariate polynomial ring in 2 variables over K
```
lgoettgens, line 57: This is one case where I don't like the name printing (aka it drastically reduces information). Could/should we make an exception for finite fields (as we already do for the singletons like `QQ`)?

fieker:
On Thu, May 02, 2024 at 06:29:12AM -0700, Lars Göttgens wrote:
> @lgoettgens commented on this pull request.
> 
> 
> 
> > @@ -53,8 +53,8 @@ julia> R, (x, y) = polynomial_ring(K, ["x", "y"]);
>  
>  julia> F = hom(R, R, z -> z^2, [y, x])
>  Ring homomorphism
> -  from multivariate polynomial ring in 2 variables over GF(2, 2)
> -  to multivariate polynomial ring in 2 variables over GF(2, 2)
> +  from multivariate polynomial ring in 2 variables over K
> +  to multivariate polynomial ring in 2 variables over K
> 
> This is one case where I don't like the name printing. Could/should we make an exception for finite fields (as we already do for the singletons like `QQ`)?
> 
Why? You named the ring K?
Or is it about not just printing

Ring homomorphism from R to R?

Which is what I'd like to see here...

Mind you, GF(2,3) is not neccessarily the same as GF(2, 3), they can
come with different polynomials, and cached or not.
You're taste (not you as in you, but in general) is always very local.
For this output on this day, I really like (or not) this printing.
If we want to be consistent, then either all parent need to support
named printing - or none. Here it looks like: (some) people really like
ti (mostly when dealing with complicated rings), while some have always
hated it...
We can also choose to be not consistent...

(I was bitten today by groups: the output basically was I am a group of
type bla - very informative. To see more I need to ask for what I want
to see....)

thofma:
@lgoettgens did you disable the tests with the problematic `hom` things?

lgoettgens:
> @lgoettgens did you disable the tests with the problematic `hom` things?

Yes, the GModule tests (exercising the "wrong" `hom` function) and the Aqua tests (failing because of method overwriting) are currently disabled to find everything else that needs to be adapted

lgoettgens:
I now expect all of CI to be green, as I pushed the fixed doctests and booktests.

However, [`85c84e0` (#3667)](https://github.com/oscar-system/Oscar.jl/pull/3667/commits/85c84e0f92900959e6a1220880e42494309708c5) disabled some tests and must thus be reverted before merging.

The only issue there is the existence of two methods `hom(G::FinGenAbGroup, H::FinGenAbGroup, V::Vector{<:Map{FinGenAbGroup, FinGenAbGroup}})`, one in Oscar for direct products and one in Hecke for tensor products. One way to solve this would be to merge https://github.com/thofma/Hecke.jl/pull/1483 and remove the method from Oscar. Other ways would be to rename the functions `hom_tensor` and `hom_direct_product`/`hom_direct_sum` etc. (and https://github.com/Nemocas/AbstractAlgebra.jl/blob/1f2ae64a68bda72ff68e65687e74818323f58e87/src/generic/DirectSum.jl#L255 and https://github.com/Nemocas/AbstractAlgebra.jl/blob/1f2ae64a68bda72ff68e65687e74818323f58e87/src/generic/DirectSum.jl#L260 consistently as well).

What do you guys think? @fingolfin @thofma @fieker @ThomasBreuer

thofma:
I am strongly in favour of disentangling this by using `hom_tensor` etc.

src/Rings/MPolyMap/MPolyRing.jl
fingolfin: I don't see how this "drastically reduces information" -- K was defined a few lines above so this seems quite clear to me?

thofma (approved):

fingolfin (approved):

PR #3671, opened 2024-05-02
Replace many uses of `G.X` by `GapObj(X)`

fingolfin:
... so that ultimately we can refactor the underpinnings of GAPGroup etc. -- initially by dropping those slightly evil `getproperty` methods that produce `G.X` on demand.

In the future this enables other refactoring. E.g. we could contemplate adding a mechanism where one could replace `GAPWrap.Foo(GapObj(x))` by `GAPWrap.Foo(x)` (namely: whenever the wrapper specifies that an argument *must* be a `GapObj`, as in e.g. `GAP.@wrap ClassNames(x::GapObj)::GapObj` -- then the generated wrapper may as well assume it has to convert anything else to `GapObj`. 

Anyway: this PR does not strive to eliminate all accesses to `.X` -- first off, some of them are of course legit. Secondly, this is already very big and I'd rather resolve a large chunk of the issue *now*, before delaying this further and having to deal with merge conflicts...


ThomasBreuer (approved):
Good.

Tests fail because `GapObj` is not known in the tests; the (few) tests that use `GapObj` already now call it as `GAP.GapObj`. Shall we better export `GapObj`, since now its use as a constructor in tests seems to be natural?

(It is interesting that `GAPGroupConjClass` has a field `X` whose value is the Oscar object given by `acting_group`, and that the underlying GAP object is stored in `CC`.)

fingolfin:
> Tests fail because `GapObj` is not known in the tests; the (few) tests that use `GapObj` already now call it as `GAP.GapObj`. Shall we better export `GapObj`, since now its use as a constructor in tests seems to be natural?

Seems reasonably, so I've now exported `GapObj`.

> (It is interesting that `GAPGroupConjClass` has a field `X` whose value is the Oscar object given by `acting_group`, and that the underlying GAP object is stored in `CC`.)

"Interesting" is putting it nicely -- I found it very confusing and surprising, and actually additional motivation from moving away from access to `FOO.X` to use of accessor functions with more self-explanatory names :-)

PR #3672, opened 2024-05-02
Some bugfixes and improvements

HechtiDerLachs:
This is work in progress for all the small fixes necessary to make some old scripts of @simonbrandhorst and myself run again.

### DIFF
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -1979,4 +1979,38 @@ function _pullback_along_base_change(f::AbsCoveredSchemeMorphism, D::AbsWeilDivi
   return WeilDivisor(AlgebraicCycle(domain(f), coefficient_ring(D), comp_dict; check=false); check=false)
 end
 
+########################################################################
+# Singular locus ideal sheaf
+########################################################################
+
+function produce_object_on_affine_chart(II::SingularLocusIdealSheaf, U::AbsAffineScheme)
+  return radical(produce_non_radical_ideal_of_singular_locus(II, U))
+end
+
+function produce_non_radical_ideal_of_singular_locus(II::SingularLocusIdealSheaf, U::AbsAffineScheme)
+  if !haskey(II.non_radical_ideals, U)
+    if is_one(focus(II)(U))
+      II.non_radical_ideals[U] = focus(II)(U)
+      return focus(II)(U)
+    end
+    X = scheme(II)
+    @assert any(V===U for V in affine_charts(X))
+    UU = simplify(U)
+    id, id_inv = identification_maps(UU)
+
+    # Prepare for shortcuts in the computation
+    @show has_attribute(X, :is_equidimensional)
+    has_attribute(X, :is_equidimensional) && get_attribute(X, :is_equidimensional)===true && set_attribute!(UU, :is_equidimensional=>true)
+    has_attribute(X, :is_irreducible) && get_attribute(X, :is_irreducible)===true && set_attribute!(UU, :is_irreducible=>true)
+    has_attribute(X, :is_reduced) && get_attribute(X, :is_reduced)===true && set_attribute!(UU, :is_reduced=>true)
+    S, inc_S = singular_locus(UU; compute_radical=false)
+    res = ideal(OO(U), pullback(id_inv).(gens(image_ideal(inc_S)))) + focus(II)(U)
+    II.non_radical_ideals[U] = res
+  end
+  return II.non_radical_ideals[U]::Ideal
+end
+
+is_one(II::SingularLocusIdealSheaf) = all(is_one(produce_non_radical_ideal_of_singular_locus(II, U)) for U in affine_charts(scheme(II)))
+
+in_radical(J::AbsIdealSheaf, II::SingularLocusIdealSheaf) = all(all(radical_membership(g, produce_non_radical_ideal_of_singular_locus(II, U)) for g in gens(J(U))) for U in affine_charts(scheme(J)))
 
--- a/experimental/Schemes/IdealSheaves.jl
+++ b/experimental/Schemes/IdealSheaves.jl
@@ -802,7 +802,10 @@ Background:
 More generally, a point ``x`` on a scheme ``X`` associated to a quasi-coherent sheaf ``F`` is embedded, if it is the specialization of another associated point of ``F``.
 Note that maximal associated points of an ideal sheaf on an affine scheme ``Spec(A)`` correspond to the minimal associated primes of the corresponding ideal in ``A``.
 """
-function maximal_associated_points(I::AbsIdealSheaf; covering=default_covering(scheme(I)))
+function maximal_associated_points(I::AbsIdealSheaf; 
+    covering=default_covering(scheme(I)),
+    algorithm::Symbol=:GTZ
+  )
   # The following would reroute to a more high-brow method with cleaner code.
   # It performs equally well in terms of time and memory consumption, but is disabled for the moment. 
   # return _maximal_associated_points(I; covering, use_decomposition_info=false)
@@ -822,7 +825,7 @@ function maximal_associated_points(I::AbsIdealSheaf; covering=default_covering(s
     @vprint :MaximalAssociatedPoints 2 "$(length(charts_todo)) remaining charts to go through\n"
     U = pop!(charts_todo)
     !is_one(I(U)) || continue                        ## supp(I) might not meet all components
-    components_here = minimal_primes(I(U))
+    components_here = minimal_primes(I(U); algorithm)
     
     ## run through all primes in MinAss(I(U)) and try to match them with previously found ones
     for comp in components_here
@@ -873,7 +876,12 @@ function radical_membership(x::RingElem, I::MPolyQuoLocalizedIdeal)
 end
 
 # Proof of concept method
-function _maximal_associated_points(I::AbsIdealSheaf; covering=default_covering(scheme(I)), use_decomposition_info::Bool=true)
+function _maximal_associated_points(
+    I::AbsIdealSheaf; 
+    covering=default_covering(scheme(I)), 
+    use_decomposition_info::Bool=true,
+    algorithm::Symbol=:GTZ
+  )
   X = scheme(I)
   comps = AbsIdealSheaf[]
   dec_inf = decomposition_info(covering)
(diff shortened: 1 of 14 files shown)


src/Rings/mpoly-ideals.jl
```diff
@@ -1103,13 +1107,56 @@ end
...
+    final_list = typeof(I)[]
+    for p in result
+      any((q !== p && is_subset(q, p)) for q in result) && continue
+      push!(final_list, p)
+    end
+    for p in final_list
+      set_attribute!(p, :is_prime=>true)
+    end
+    return final_list
+  end
```
HechtiDerLachs: @afkafkafk13 , @hannes14 : Could you let me knew what you think of these? I tried to run a script again from about 5 months ago and it broke down on some primary decompositions (which used to go though at some point). Now I'm trying to get back there with these tweaks. And I thought that they might be useful in general. 

@hannes14 : This was also used to prepare the example I sent you. Interestingly it is solved rather quickly with `algorithm=:charSets`, but never terminated for me with `:GTZ`.

src/Rings/mpoly-ideals.jl
```diff
@@ -1103,13 +1107,56 @@ end
...
+  # Try to eliminate variables first. This will often speed up computations significantly.
+  if simplify_ring
+    Q, pr = quo(R, I)
+    W, id, id_inv = simplify(Q)
+    @assert domain(id) === codomain(id_inv) === Q
+    @assert codomain(id) === domain(id_inv) === W
+    res_simp = minimal_primes(modulus(W); algorithm, factor_generators, simplify_ring=false)
+    result = [I + ideal(R, lift.(id_inv.(W.(gens(j))))) for j in res_simp]
+    for p in result
+    end
```
afkafkafk13: This looks a bit weird: either some line was lost or some leftovers were not deleted

HechtiDerLachs: Indeed. I will clean that up.

afkafkafk13:
General question: Do you have any idea, what increased the cost of these older scripts? I find this a bit scary!
To me this sounds like some optimization (at some other and maybe completely unrelated) point having very negative (and up to now undetected) side effects. 

The improvements make sense in any case. I have used such tweaks many times for large settings in Singular, where primary decomposition as a black box did not go through. But you need to be aware that you are deliberately circumventing the heuristics of primdec.lib and this should only be necessary in nasty examples. Is your offending example really sufficiently nasty?

HechtiDerLachs:
Thanks for your comments. 

No, in my opinion the offending example is not nasty. I sent it to @hannes14 via email, but I can put it here again:
```
julia> minimal_primes(list[2])
R = Multivariate polynomial ring in 4 variables over QQ
gens(R) = QQMPolyRingElem[θ_2, θ_1, (s0//s1), (w//x)]
g = -θ_1^2*(s0//s1) + θ_1^2 + (s0//s1)*(w//x)^4
g = θ_2^8 + 1
g = -θ_2^6 + θ_2^4 - θ_2^2 + θ_1^2
```
Be aware that `s0//s1` etc. are not fractions, but only symbols for variable names. So this is really a polynomial computation and the three `g`s are the generators of the ideal in question (printed in a `for`-loop with `g` as local variable). 

To me this did not look scary at all. But `minimal_primes` gives up on this with `GTZ`.

HechtiDerLachs:
Does anyone have an explanation for the failing tests? I suppose the same things are tested in other threads, aren't they? Why do they fail on ubuntu 1.10 long? 

@lgoettgens ?

lgoettgens:
> Does anyone have an explanation for the failing tests? I suppose the same things are tested in other threads, aren't they? Why do they fail on ubuntu 1.10 long?
> 
> @lgoettgens ?

ubuntu-1.10-long spent 2.5 hours in `/home/runner/work/Oscar.jl/Oscar.jl/test/AlgebraicGeometry/Schemes/elliptic_surface.jl` and timed out

withoutexperimental spent about 2 hours in `/home/runner/work/Oscar.jl/Oscar.jl/test/AlgebraicGeometry/Schemes/elliptic_surface.jl` and timed out

ubuntu-1.10-book failed due to some change in output in decker-schmitt-invariant-theory: https://github.com/oscar-system/Oscar.jl/actions/runs/8927137202/job/24519841523?pr=3672#step:8:3632

afkafkafk13:
> Thanks for your comments.
> 
> No, in my opinion the offending example is not nasty. I sent it to @hannes14 via email, but I can put it here again:
> 
> ```
> julia> minimal_primes(list[2])
> R = Multivariate polynomial ring in 4 variables over QQ
> gens(R) = QQMPolyRingElem[θ_2, θ_1, (s0//s1), (w//x)]
> g = -θ_1^2*(s0//s1) + θ_1^2 + (s0//s1)*(w//x)^4
> g = θ_2^8 + 1
> g = -θ_2^6 + θ_2^4 - θ_2^2 + θ_1^2
> ```
> 
> Be aware that `s0//s1` etc. are not fractions, but only symbols for variable names. So this is really a polynomial computation and the three `g`s are the generators of the ideal in question (printed in a `for`-loop with `g` as local variable).
> 
> To me this did not look scary at all. But `minimal_primes` gives up on this with `GTZ`.

I entered the computation in Singular and found the following observations:
`ring r=0,(a,b,c,d),dp;
ideal I=b^2*c+b^2+c*d^4, a^8+1, -a^6+a^4-a^2+b^2;`
internally uses a ring (0,c),(a,b,d),(lp(3),C) and stalls

`ring r=0,(d,c,b,a),dp;
ideal I=b^2*c+b^2+c*d^4, a^8+1, -a^6+a^4-a^2+b^2;`
internally uses a ring (0,d),(c,b,a),(lp(3),C) and returns the result in a fraction of a second.

So the difference is in the internal choice of the "maximal independent set" in this primary decomposition algorithm and of course in the difficulty of the lp-groebner basis to be computed internally. This choice is not done by an elaborate heuristic; hence the situation is brittle, whenever the different choices have significantly different runtimes. So probalby some change further up in the hierarchy changed the sequence of the variables, which in turn triggers a different choice of a maximal independent set.
Actually, the most solid (but currently rather unrealistic) variant would be to run the different choices in parallel and let the first one win. (Maybe you want to at least make sure that the variables from the tower of algebraic field extensions are added in a suitable order to avoid unnecessary spoly among the minimal polynomials w.r.t. lex ordering.)

HechtiDerLachs:
I changed the default algorithm for the place where primary decomposition is called in the elliptic surfaces. Maybe that already helps to have the tests run more reliably (#3676).

test/Rings/mpoly-localizations.jl
```diff
@@ -466,3 +466,17 @@ end
...
+@testset "radical computation" begin
+  R, (x, y) = QQ["x", "y"]
+  #U = complement_of_prime_ideal(ideal(R, y))
+  U = powers_of_element(x)
+  J = ideal(R, x-y^2+ 1)
+  W, _ = quo(R, J)
+
+  L, loc_map = localization(W, U)
+
+  I = ideal(L, [y^2, 0]) # Dealing with zero generators lead to a bug once, so this is a real test.
```
afkafkafk13, line 478:
```suggestion
  I = ideal(L, [y^2, 0]) # Dealing with zero generators led to a bug once, so this is a real test.
```

afkafkafk13 (approved):
Thank you.

Concerning the long-running tests: It makes sense to keep them, but disable them. We should, however, remember to run them locally for all PR concerning optimizations in the basic functionality for  IdealSheaf, ..Divisor etc to detect regression.

joschmitt:
The change to `minimal_primes` changes the ordering in one of the book tests. How do we handle this again?
Otherwise I would like to have this in, so that the CI stops running into #3676 (I hope).

benlorenz:
> The change to `minimal_primes` changes the ordering in one of the book tests. How do we handle this again? Otherwise I would like to have this in, so that the CI stops running into #3676 (I hope).

You can change the output line in the jlcon file (`test/book/specialized/decker-schmitt-invariant-theory/H3.jlcon`).

experimental/Schemes/IdealSheaves.jl
```diff
@@ -873,7 +876,12 @@ function radical_membership(x::RingElem, I::MPolyQuoLocalizedIdeal)
 end
 
 # Proof of concept method
-function _maximal_associated_points(I::AbsIdealSheaf; covering=default_covering(scheme(I)), use_decomposition_info::Bool=true)
+function _maximal_associated_points(
+    I::AbsIdealSheaf; 
+    covering=default_covering(scheme(I)), 
+    use_decomposition_info::Bool=true,
+    algorithm::Symbol=:GTZ
+  )
```
simonbrandhorst, line 884: Quoting the singular manual:

https://www.singular.uni-kl.de/Manual/latest/sing_1426.htm#SEC1507
"In small characteristic and for algebraic extensions, the procedures via Gianni, Trager, Zacharias may not terminate. "

simonbrandhorst: Better link:
https://www.singular.uni-kl.de/Manual/latest/sing_1511.htm#SEC1592

afkafkafk13: The algebraic extensions only refer to those implemented as extended basefield in Singular. If you encode the algebraic extension by hand, the problem does not occur (hence the encoding by hand).  This is the reason why these particular calls to Singular should not use the algebraic extensions in Singular, but do the encoding themselves (this was subject of discussion in another PR). 
The restriction on the characteristic results from the sufficiently general linear coordinate change, which is used (the open set needs to be non-empty, which might be a problem in very small characteristics). This should be stated as a warning somewhere in the docstring, but allowing characteristics like 101 or 32003 is vital for many practical applications where coefficients are known to explode in characteristic zero.

joschmitt:
I adjusted the output of the failing book test.

experimental/Schemes/SimplifiedAffineScheme.jl
```diff
@@ -21,6 +21,9 @@ a `SimplifiedAffineScheme` ``Y`` with ``X`` as its `original`.
 from the one of `X` and hence the two schemes will not compare using `==`.
 """
 function simplify(X::AbsAffineScheme{<:Field})
+  # We need to avoid creating a polynomial ring with zero variables
+  dim(X) == 0 && vector_space_dimension(OO(X)) == 0 && return SimplifiedAffineScheme(X, X, identity_map(X), identity_map(X), check=false)
```
simonbrandhorst, line 25: May I suggest moving the point to `0`? There is the function `rational_point_coordinates`. 
It will push some of the complexity into the gluings ... but things like saturation should be easier. 
Not sure though how exactly `simplify` is hooked into everything.

simonbrandhorst:
I would suggest to merge this and relegate further improvements to a new PR.

experimental/Schemes/SimplifiedAffineScheme.jl
HechtiDerLachs: That's actually not that easy to program. I don't think it pays off since these schemes are already very simple and stuff like saturation will not be applied here.

HechtiDerLachs:
Good to go from my side.

Issue #3635, opened 2024-04-22
Short term TODOs for FTheoryTools


apturner:
- The parameters `D1` and `D2` parameters for hypersurface models need to be set properly in the literature model database
    - The current values of `D1` and `D2` are incorrect for the toric hypersurface models
    - The coordinates of the fiber ambient space need to be reordered so that `D1` and `D2` correspond to the first two given coordinates, OR we need a new piece of information that indicates how `D1` and `D2` correspond to the given coordinates and update the constructor appropriately
        - If the coordinates are reordered, the ambient space rays need to be correspondingly reordered, and the maximal cones need to be rewritten to match this ordering
- The model F4 cannot currently be handled, which requires updating the constructor
- The name `model_sections` should be updated to clarify that these are the parameters whose divisor *class* needs to be given by the user; the actual sections themself do not need to be specified a priori
- The constructor for literature hypersurface models should handle any grading, without assuming that the first two base coordinates are elements of the basis for the grading 
- Extend the index for the literature model database to allow for hypersurface model and Weierstrass model versions of the same model
- Add CI tests for Weierstrass model version of toric hypersurface models
- Add new field to literature models for torsion sections of Mordell–Weil group

@emikelsons

HereAround:
Based on our discussion this afternoon @apturner and @emikelsons , I believe this issue can be closed.

PR #3675, opened 2024-05-02
FTheoryTools: Clarify and improve mechanics of literature model constructor

HereAround:
This PR aims to complete https://github.com/oscar-system/Oscar.jl/issues/3635. Changes in this PR include:
* Optimizing the literature model constructor by removing obsolete information.
* Renaming fields in the JSON database and the literature model constructor to clarify the mechanics.
* Support the fiber_twist_matrix for hypersurface models.
* Update the JSON database accordingly and parametrize the F4-model (F-theory on all toric hypersurfaces) in tandem with the paper version.

As I see it, this complete all of https://github.com/oscar-system/Oscar.jl/issues/3635 but
> Add new field to literature models for torsion sections of Mordell–Weil group
Maybe this should be relegated to a separate PR in which also this information could be filled in? Or do you we want to add this on top of this very PR?

@apturner and @emikelsons, please let me know your thoughts.


experimental/FTheoryTools/docs/src/hypersurface.md
```diff
@@ -76,7 +76,7 @@ Completeness is an expensive check. Therefore, we provide an optional argument w
  demonstrate this:
 ```@docs
 hypersurface_model(base::NormalToricVariety; completeness_check::Bool = true)
-hypersurface_model(base::NormalToricVariety, fiber_ambient_space::NormalToricVariety, D1::ToricDivisorClass, D2::ToricDivisorClass; completeness_check::Bool = true)
+hypersurface_model(base::NormalToricVariety, fiber_ambient_space::NormalToricVariety, fiber_twist_divisor_classes::Vector{ToricDivisorClass}; completeness_check::Bool = true)
```
apturner, line 79: Is this not what we were planning to call `fiber_twist_matrix`, at least in the JSON files? I imagine this name could match that

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
```diff
@@ -23,12 +23,13 @@ function hypersurface_model(base::NormalToricVariety; completeness_check::Bool =
...
+  return hypersurface_model(base, fiber_ambient_space, [D1, D2, D3]; completeness_check = completeness_check)
 end
 
 
 @doc raw"""
-    hypersurface_model(base::NormalToricVariety, fiber_ambient_space::NormalToricVariety, D1::ToricDivisorClass, D2::ToricDivisorClass; completeness_check::Bool = true)
+    hypersurface_model(base::NormalToricVariety, fiber_ambient_space::NormalToricVariety, fiber_twist_divisor_classes::Vector{ToricDivisorClass}; completeness_check::Bool = true)
 
 Construct a hypersurface model, for which the user can specify a fiber ambient space
 as well as divisor classes of the toric base space, in which the first two homogeneous
```
apturner, line 35: This is no longer "the first two", right?

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
```diff
@@ -66,8 +70,8 @@ function hypersurface_model(base::NormalToricVariety, fiber_ambient_space::Norma
   end
   
   # Compute an ambient space
-  ambient_space = _ambient_space(base, fiber_ambient_space, D1, D2)
-  
+  ambient_space = _ambient_space(base, fiber_ambient_space, fiber_twist_divisor_classes)
+
   # Construct the model
   hypersurface_equation = generic_section(anticanonical_bundle(ambient_space))
```
apturner, line 76: This generates a hypersurface equation with random coefficients in front of every monomial, right? Why do we want this? It doesn't seem like this would ever be used, unless I'm misunderstanding something

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
```diff
@@ -23,12 +23,13 @@ function hypersurface_model(base::NormalToricVariety; completeness_check::Bool =
   set_coordinate_names(fiber_ambient_space, ["x", "y", "z"])
   D1 = 2 * anticanonical_divisor_class(base)
   D2 = 3 * anticanonical_divisor_class(base)
-  return hypersurface_model(base, fiber_ambient_space, D1, D2; completeness_check = completeness_check)
+  D3 = trivial_divisor_class(base)
+  return hypersurface_model(base, fiber_ambient_space, [D1, D2, D3]; completeness_check = completeness_check)
```
apturner, line 27: As noted in a later comment on the function that this calls, I'm worried about what this constructor does. It seems that the intention is for this to return a hypersurface model object that is actually a Weierstrass/Tate model, since the fiber is P^{2, 3, 1}. But I don't think it will return the hypersurface equation of a Weierstrass/Tate model, but instead a similar equation with randomized coefficients in front of the y^2 and x^3 terms. Also, why do we want a constructor that makes hypersurface models that are Weierstrass/Tate models?

experimental/FTheoryTools/src/LiteratureModels/Models/model1109_3454.json
```diff
@@ -38,16 +38,16 @@
     },
     "model_data": {
         "base_dim": 3,
-        "model_sections": ["w"],
-        "base_coordinates": ["w", "a1", "a21", "a32", "a43"],
+        "divisors_user_must_specify": ["w"],
+        "divisors_parametrized_by_user_specified_divisors": ["a1", "a21", "a32", "a43"],
```
apturner, line 42: The word "parametrized" here is one that I want to think about.

There are two separate things going on that we need to track:
1. There is a basis of divisor classes, defined in terms of some user-given classes, in terms of which the divisor classes of all other line bundles are determined.
2. There are sections of line bundles, some of which may be determined by others, such as a2 = a21 * w in this example.

Previously, we have used the word "parametrized" in the second case, where we say that the section a2 is parameterized by the independent sections a21 and w. If we still want to use that language there, I think we should avoid it when talking about the divisor classes, so as to avoid confusion (which we've already run into in the past). This is also nice to avoid because it feels unusual to use the word "parametrized" for the relationship between a vector and the basis it's written down in.

So, I propose `user_specified_classes` and `model_parameters` (could be another name there) instead of `divisors_user_must_specify` and `divisors_parametrized_by_user_specified_divisors`. Similarly below I would replace `parametrization_of_parametrized_divisors_by_Kbar_and_user_specified_divisors` with `parameter_classes_in_basis_of_Kbar_and_user_specified_classes`.

My choice of names here is related to another change I think would be beneficial, which I discuss at length in a later comment: I think all parameters of the model should be listed in the field that is currently called `divisors_parametrized_by_user_specified_divisors`. This would tell the constructor which parameters actually appear, and thus which parameters it needs to construct sections for. We could even go so far as to distinguish the names of the classes versus the parameters (maybe with capitalization), so in this case we would have something like:
```
        "defining_classes": ["W"],
        "model_sections": ["w", "a1", "a21", "a32", "a43"]
        ...
        "classes_of_model_sections_in_basis_of_Kbar_and_defining_classes": [
            [0, 1, 2, 3, 4],
            [1, 0, -1, -2, -3]
        ]
```

This applies to all other JSON files as well, of course.

EDIT: Changed the code sample above to match the names we discussed on Zoom.

experimental/FTheoryTools/src/LiteratureModels/Models/model1208_2695-1.json
apturner, line 1: It has just occurred to me that our file naming convention is no longer consistent. This file, "model1208_2695-1.json", is a hypersurface model, and the file "model1208_2695-2.json" is the Weierstrass version of the same model. But as of the inclusion of the toric hypersurfaces paper, the final number is only being used to distinguish entirely distinct models, and the additional decoration "-WSF" is being added for the Weierstrass forms. This is not really a huge deal, as the file names are all internal, but I thought I'd mention it.

experimental/FTheoryTools/src/LiteratureModels/Models/model1408_4808-11-WSF.json
```diff
@@ -46,14 +46,14 @@
     },
     "model_data": {
         "base_dim": 3,
-        "model_sections": ["s5", "s9"],
-        "base_coordinates": ["s5", "s9", "s1", "s2", "s3", "s6"],
+        "divisors_user_must_specify": ["s5", "s9"],
+        "divisors_parametrized_by_user_specified_divisors": ["s1", "s2", "s3", "s6"],
```
apturner, line 50: This is a point we have discussed some elsewhere, but I'll raise it again here for visibility. In a future PR, this file and other will be updated so that the user-given divisor classes are always S7 and S9. This model does not have an s7 parameter, but does have an s9 parameter, but since `divisors_parametrized_by_user_specified_divisors` only contains parameters that don't appear in `divisors_user_must_specify`, the constructor has no way of telling that it *should* choose a section for s9 but *should not* choose a section for s7. I think this is not necessarily harmful, as the chosen section for s7 will just never be used, but it is slightly wasteful.

Listing all divisor classes in this second field would provide the necessary information to only create the desired section. See earlier related comment on "model1109_3454.json"

experimental/FTheoryTools/src/WeierstrassModels/constructors.jl
```diff
@@ -69,7 +69,8 @@ function weierstrass_model(base::NormalToricVariety,
   set_coordinate_names(fiber_ambient_space, ["x", "y", "z"])
   D1 = 2 * anticanonical_divisor_class(base)
   D2 = 3 * anticanonical_divisor_class(base)
-  ambient_space = _ambient_space(base, fiber_ambient_space, D1, D2)
+  D3 = trivial_divisor_class(base)
+  ambient_space = _ambient_space(base, fiber_ambient_space, [D1, D2, D3])
   
   # construct the model
   pw = _weierstrass_polynomial(explicit_model_sections["f"], explicit_model_sections["g"], cox_ring(ambient_space))
```
apturner, line 76: This is an aside, but the fact that the WeierstrassModel type takes in the entire hypersurface polynomial as input, rather than f and g, means that in principle the user could make a WeierstrassModel object that has a completely different hypersurface equation, and then presumably all of our functions related to Weierstrass models would fail. This is maybe something we should think about in the future.

experimental/FTheoryTools/src/LiteratureModels/constructors.jl
```diff
@@ -257,58 +257,54 @@ end
...
+function _construct_literature_model_over_concrete_base(model_dict::Dict{String,Any}, base_space::FTheorySpace, user_spec_divs::Dict{String,ToricDivisor}, completeness_check::Bool, generic::Bool)
 
   # We first create a polynomial ring in which we can read all model sections as polynomials of the defining sections
-  @req ((model_dict["model_descriptors"]["type"] == "tate") || (model_dict["model_descriptors"]["type"] == "weierstrass") || (model_dict["model_descriptors"]["type"] == "hypersurface")) "Model is not a Tate or Weierstrass model"
-  @req haskey(model_dict["model_data"], "base_coordinates") "No base coordinates specified for model"
-  auxiliary_base_ring, _ = polynomial_ring(QQ, string.(model_dict["model_data"]["base_coordinates"]), cached=false)
-  vars = [string(g) for g in gens(auxiliary_base_ring)]
+  @req haskey(model_dict["model_data"], "divisors_parametrized_by_user_specified_divisors") "No base coordinates specified for model"
+  vars = vcat([string.(model_dict["model_data"]["divisors_user_must_specify"]), string.(model_dict["model_data"]["divisors_parametrized_by_user_specified_divisors"])]...)
```
apturner, line 264: This is the reason that we are currently choosing sections for parameters that don't necessarily exist in the actual model. If the second field, which I'm proposing to be called `model_parameters`, just contains all the parameters, you could just use that directly here instead of combining two fields, and it would only contain the parameters we need.

experimental/FTheoryTools/src/LiteratureModels/constructors.jl
```diff
@@ -257,58 +257,54 @@ end
...
+  auxiliary_ring, _ = polynomial_ring(QQ, vars, cached=false)
 
   # Make list of divisor classes which express the internal model sections.
-  model_sections_divisor_list = vcat([anticanonical_divisor(base_space)], [model_sections[vars[k]] for k in 1:length(model_sections)])
+  lusd = length(user_spec_divs)
+  user_specified_divisors = vcat([anticanonical_divisor(base_space)], [user_spec_divs[vars[k]] for k in 1:lusd])
```
apturner, line 269: Should this be called `basis_of_divisors` or `basis_of_classes` or something? Since it includes Kbar now

apturner (approved):
Thanks for all this @HereAround. I think the changes I propose here would improve clarity and get rid of some waste in producing unnecessary sections, but this could be merged now and those changes made in a future PR if you prefer, since this code does *work*. For that reason, I'm approving now.

HereAround:
~~@apturner Since there is a lot to be discussed, back to draft.~~

As agreed on slack: I will merge this PR now and we can discuss the many open ends next Monday to find an effective strategy for addressing them.

experimental/FTheoryTools/docs/src/hypersurface.md
HereAround: As discussed in zoom, we are now fine with the name.

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
HereAround: I will chase that in the documentation in a separate PR.

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
HereAround: All correct. The function in lines 21 to 28 of HypersurfaceModels/constructors.jl should just be gone. I will take care of that.

experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
HereAround: Your observation is yet again correct. This needs fixing in a separate PR. I will look into this.

experimental/FTheoryTools/src/LiteratureModels/Models/model1109_3454.json
HereAround: As just discussed at length in zoom, those proposed changes are good. I will take care of them in a separate PR.

experimental/FTheoryTools/src/LiteratureModels/Models/model1408_4808-11-WSF.json
HereAround: Will be included in an upcoming PR. Thank you!

experimental/FTheoryTools/src/WeierstrassModels/constructors.jl
HereAround: We agreed to reserve this for a future discussion.

experimental/FTheoryTools/src/LiteratureModels/constructors.jl
HereAround: As discussed, will be fixed in a future PR.

experimental/FTheoryTools/src/LiteratureModels/constructors.jl
HereAround: As discussed, this will be fixed in a future PR.

PR #3654, opened 2024-04-29
tests and cleanup for elliptic surfaces

simonbrandhorst:

Issue #3676, opened 2024-05-02
elliptic surface tests get stuck (and cancelled after 2.5 hours)


benlorenz:
e.g. here with julia 1.10 on ubuntu:
https://github.com/oscar-system/Oscar.jl/actions/runs/8927137202/job/24519839319?pr=3672#step:8:474

But this seems to happen on other julia versions / operating systems as well:

Downstream tests for Polymake ran for almost 5 hours with julia 1.6:
https://github.com/oscar-system/Polymake.jl/actions/runs/8921273104/job/24501206313?pr=485

Or on macos here:
https://github.com/oscar-system/Oscar.jl/actions/runs/8868048924/job/24347226536#step:8:424

julia 1.11:
https://github.com/oscar-system/Oscar.jl/actions/runs/8920051759/job/24497347178

See also https://github.com/oscar-system/Oscar.jl/pull/3672#issuecomment-2091305190

HechtiDerLachs:
But it does not always happen, does it? I mean: Usually the test seems to go through.

benlorenz:
> But it does not always happen, does it? I mean: Usually the test seems to go through.

Correct, it does not happen every time but quite often. (So much that it is annoying in the CI)

HechtiDerLachs:
As already mentioned on #3677 : I guess that randomness in primary decomposition might be the reason for the occasional failure. If we can pass on some random seed that should solve the problem. Otherwise I would disable the tests for elliptic surfaces.

benlorenz:
I did get a stuck process last night, but it doesn't seem to reproduce with the same seed, maybe there is something involved that uses a different RNG. (The CI should also use the same seed every time)

But I did get a backtrace of the process (at about 6 hours of runtime):
```
__libc_malloc at /lib64/libc.so.6 (unknown line)
omEmulateValloc at /workspace/srcdir/Singular/omalloc/omAllocSystem.c:166
_omVallocFromSystem at /workspace/srcdir/Singular/omalloc/omAllocSystem.c:320
omAllocNewBinPagesRegion at /workspace/srcdir/Singular/omalloc/omBinPage.c:302
omAllocBinPage at /workspace/srcdir/Singular/omalloc/omBinPage.c:129
omAllocNewBinPage at /workspace/srcdir/Singular/omalloc/om_Alloc.c:38
omAllocBinFromFullPage at /workspace/srcdir/Singular/omalloc/om_Alloc.c:145
[...]

PR #3677, opened 2024-05-02
Revert "tests and cleanup for elliptic surfaces (#3654)"

fingolfin:
This reverts PR #3654 which was merged despite some CI runs failing with a timeout in `test/AlgebraicGeometry/Schemes/elliptic_surface.jl`. And now master and other PRs are running into these timeouts.

### DIFF
--- a/experimental/Schemes/elliptic_surface.jl
+++ b/experimental/Schemes/elliptic_surface.jl
@@ -1323,7 +1323,7 @@ function horizontal_decomposition(X::EllipticSurface, F::Vector{QQFieldElem})
   else
     found = false
     for (i,(T, tor)) in enumerate(tors)
-      d = F1 - _vec(tor)
+      d = F2 - _vec(tor)
       if all(isone(denominator(i)) for i in d)
         found = true
         T0 = mordell_weil_torsion(X)[i]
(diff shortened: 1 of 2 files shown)


benlorenz:
The timeouts in the elliptic surface tests started a few days earlier, e.g. https://github.com/oscar-system/Oscar.jl/actions/runs/8868048924/job/24347226536#step:8:424 (4 days ago).
See also https://github.com/oscar-system/Oscar.jl/issues/3676

HechtiDerLachs:
I suggest to keep the changes of this PR, but just disable the tests for the moment. 

Unfortunately I can't reproduce the problem; the tests just ran fine again on my machine. But I'm afraid that the timeout might be due to randomness in primary decomposition. @hannes14 : Do we have any chance to set a random seed for this? Otherwise the tests should probably be disabled for good for the CI, but left as comments.

fingolfin:
OK, if this started already a few days ago then reverting won't be helpful :-(.

@hechtiderlachs please have a look at the `Oscar.randseed!` docstring -- perhaps you can find a seed that reproduces the timeout with that? I.e.
```julia
    seed=0
    while true
      println("Now trying seed $seed")
      Oscar.randseed!(seed)
      run_tricky_computation()
      seed += 1
    end
```

hannes14:
Oscar.randseed!(seed) set (besides others) also the random generator of Singular

PR #3679, opened 2024-05-03
Improve performance of `permutation_of_terms`, `monomials`

### LINKED ISSUE #3055: Iteration over monomials in modules is slow
I came across this issue in #3002 and also left a comment in the source. But it is probably better to also have this documented here. Try the following:
```julia
R, (x, y, z) = QQ[:x, :y, :z]

m = ideal(R, gens(R))
pow_m = m^200;

F = FreeMod(R, 5)

M, _ = pow_m*F;

w = sum(gens(M));
v = ambient_representative(w);
@time collect(monomials(v));

g = sum(gens(m));
@time collect(monomials(g));
```
What you would expect is that the second timing is roughly five times the first, because `F` is free of rank 5. However, I get 
```
julia> @time collect(monomials(v));
  0.447076 seconds (7.59 M allocations: 301.277 MiB, 2.17% gc time)
```
vs.
```
julia> @time collect(monomials(g));
  0.000033 seconds (18 allocations: 840 bytes)
```
The use of a profiler reveals that apparently there is some "ordering up to permutation" working on putting the monomials in modules in correct order with respect to some monomial ordering of the module.

fingolfin:
Resolves #3055 (CC @HechtiDerLachs @simonbrandhorst)

Speed and memory usage are improved drastically. The main contributing factor is that now `permutation_of_terms` methods can fully take advantage of the type of the monomial resp. module orderings. This then avoids a ton of dynamic dispatches during runtime.

In addition, two smaller changes contribute:
- using `Iterators.reverse` instead of `reverse` avoids allocating a new vector when we just want to iterate backwards over an existing one
- changing `struct FreeModElem` to fully specify the type of its member `coords` gets rid of some additional dynamic dispatch.

Consider this setup:

    R, (x, y, z) = QQ[:x, :y, :z]

    m = ideal(R, gens(R))
    pow_m = m^200;

    F = FreeMod(R, 5)

    M, _ = pow_m*F;

    w = sum(gens(M));
    v = ambient_representative(w);
    g = sum(gens(pow_m));

Then we measure the performance before and after this PR.

Before:

    julia> @btime collect(monomials(v));
      738.822 ms (10193202 allocations: 393.99 MiB)

    julia> @btime collect(monomials(g));
      64.825 ms (1031634 allocations: 37.52 MiB)

    julia> @btime permutation_of_terms(v, default_ordering(parent(v)));
      616.583 ms (7253642 allocations: 263.95 MiB)

    julia> @btime permutation_of_terms(g, default_ordering(parent(g)));
      62.254 ms (950427 allocations: 34.42 MiB)

After:

    julia> @btime collect(monomials(v));
      188.433 ms (1012571 allocations: 43.89 MiB)

    julia> @btime collect(monomials(g));
      20.697 ms (81211 allocations: 3.41 MiB)

    julia> @btime permutation_of_terms(v, default_ordering(parent(v)));
      153.328 ms (73 allocations: 5.21 MiB)

    julia> @btime permutation_of_terms(g, default_ordering(parent(g)));
      18.084 ms (4 allocations: 317.34 KiB)


fingolfin:
I believe that further changes of this kind would allow us to *drastically* reduce the memory usage of the algebraic geometry code (and probably some of our other code, too). E.g. I already pointed out elsewhere that conversion between OSCAR and Singular polynomials incurs up to 2 orders of magnitudes more allocations than I would expect.

I wish Julia would give some more guide rails here to avoid these pitfalls. It took me some hours to finally understand what's going on (admittedly the fact I only worked at this late at night likely didn't help). 

I guess tools like JET might have helped, but we currently have soooo many things it reports (and quite some "false positives", too) that it seems hopeless to find the worst offenders this way (well, at least for me -- perhaps someone who has mastered it better can do better? We'd just need to find such a person).


BTW I also experimented with https://github.com/JuliaLang/AllocCheck.jl here which helped a bit (it found the `reverse` problem) but no tool pointed me at the main culprit: dynamic dispatch to the comparison functions -- and I *still* don't really understand why that caused so many allocations -- that's bad. I.e. I can't fully explain why this PR is so effective in reducing allocations. If someone else can fill that in (e.g. @benlorenz @lgoettgens @thofma ...), please.

thofma:
I think the allocations with dynamic dispatch are due to boxing of the return value (even for immutables). Last time I checked out JET, it gave waaaay too many false positives, often pointing to some Base methods. I played around with the filtering, but I could not make it work to my satisfaction. This was maybe two years ago. Maybe the situation has improved?

src/Modules/ModuleTypes.jl
```diff
@@ -132,7 +132,7 @@ true
 ```
 """
 mutable struct FreeModElem{T} <: AbstractFreeModElem{T}
-  coords::SRow{T} # also usable via coeffs()
+  coords::SRow{T, Vector{T}} # also usable via coeffs()
```
lgoettgens: I think this won't work in general, as there are special implementations for the second type parameter, e.g. for ZZ. (please correct me if I am wrong @thofma)

lgoettgens: See https://github.com/thofma/Hecke.jl/blob/7d8744d1856450d4b1537a36d34e9b4776b92f9b/src/Sparse/ZZRow.jl#L192 and below

thofma: Yes. One probably needs a `sparse_row_type(T)` method (which does not exist so far).

fingolfin: Hrm, that's annoying. Then we need to make this type part of the signature for FreeModElem, which is annoying, as it is kind of an implementation detail. Maybe it suffices to ensure all access to `coords` is through an accessor function and have that add a type assertion?

But yeah it would still need a `sparse_row_type` method...

thofma: I would go via the accessor method plus type assertion.

lgoettgens (changes requested):
To not have this merged by accident.
See https://github.com/oscar-system/Oscar.jl/pull/3679#discussion_r1588880649

fingolfin:
Just converted it to a draft. I have another factor 2 improvements in allocations for `collect(monomials(v))` as well (from 2.4 million down to 1.2 million). Will push update later

Perhaps I'll just remove the SRow related bit and put that into another PR, its contribution is relatively minor.

src/Modules/ModuleTypes.jl
fingolfin: Will do. I an also working on adding `sparse_row_type`

src/Modules/FreeModElem-orderings.jl
```diff
@@ -196,7 +201,11 @@ function Base.iterate(a::GeneralPermutedIterator{:monomials, <:FreeModElem{<:MPo
   state += 1
   state <= length(a.perm) || return nothing
   (i, j) = a.perm[state]
-  return monomial(a.elem[i], j)*parent(a.elem)[i], state
+  mij = monomial(a.elem[i], j)
+  F = parent(a.elem)
+  R = base_ring(F)
+  s = sparse_row(R, [i], [mij])
```
fingolfin: The changes in this `iterate` method reduce allocations from 24 to 12. By adding the undocumented keyword arg `sort=false` we can reduce it further to 9.
```suggestion
  s = sparse_row(R, [i], [mij]; sort=false)
```

I'll update this PR further and will also submit an optimization to Hecke to make this unnecessary (basically: if the vectors have length 1 we don't need to sort them...

fingolfin:
This should be ready for merging now

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -22,6 +22,13 @@ function FreeModElem(c::Vector{T}, parent::FreeMod{T}) where T
   return FreeModElem{T}(sparse_coords,parent)
 end
 
+function FreeModElem(i::Int, x::T, parent::FreeMod{T}) where T
+  @assert 1 <= i <= rank(parent)
+  sparse_coords = sparse_row(base_ring(parent), [i], [x]; sort=false)
```
lgoettgens:
```suggestion
  sparse_coords = sparse_row(base_ring(parent), [i], [x])
```
no longer needed due to https://github.com/thofma/Hecke.jl/pull/1502

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -147,10 +154,11 @@ end
 Return the standard basis of `F`.
 """
 function basis(F::AbstractFreeMod)
-  bas = elem_type(F)[]
+  bas = Vector{elem_type(F)}(undef, dim(F))
+  e = one(base_ring(F))
   for i=1:dim(F)
-    s = Hecke.sparse_row(base_ring(F), [(i, base_ring(F)(1))])
-    push!(bas, F(s))
+    s = sparse_row(base_ring(F), [i], [e]; sort=false)
```
lgoettgens:
```suggestion
    s = sparse_row(base_ring(F), [i], [e])
```
no longer needed due to https://github.com/thofma/Hecke.jl/pull/1502

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -172,7 +180,8 @@ Return the `i`th basis vector of `F`, that is, return the `i`th standard unit ve
 """
 function basis(F::AbstractFreeMod, i::Int)
   @assert 0 < i <= ngens(F)
-  s = Hecke.sparse_row(base_ring(F), [(i, base_ring(F)(1))])
+  e = one(base_ring(F))
+  s = sparse_row(base_ring(F), [i], [e]; sort=false)
```
lgoettgens:
```suggestion
  s = sparse_row(base_ring(F), [i], [e])
```
no longer needed due to https://github.com/thofma/Hecke.jl/pull/1502

src/Modules/UngradedModules/SubquoModuleElem.jl
```diff
@@ -933,7 +933,9 @@ end
 Return the generators of `M`.
 """
 function gens(M::SubquoModule{T}) where T
-  return SubquoModuleElem{T}[gen(M,i) for i=1:ngens(M)]
+  R = base_ring(M)
+  e = R(1)
+  return [SubquoModuleElem{T}(sparse_row(R, [i], [e]; sort=false), M) for i in 1:ngens(M)]
```
lgoettgens:
```suggestion
  return [SubquoModuleElem{T}(sparse_row(R, [i], [e]), M) for i in 1:ngens(M)]
```
no longer needed due to https://github.com/thofma/Hecke.jl/pull/1502

lgoettgens:
```suggestion
  return [gen(M, i) for i in 1:ngens(M)]
```
even shorter

src/Modules/UngradedModules/SubquoModuleElem.jl
```diff
@@ -943,9 +945,7 @@ Return the `i`th generator of `M`.
 """
 function gen(M::SubquoModule{T}, i::Int) where T
   R = base_ring(M)
-  v::SRow{T} = sparse_row(R)
-  v.pos = [i]
-  v.values = [R(1)]
+  v = sparse_row(R, [i], [R(1)]; sort=false)
```
lgoettgens:
```suggestion
  v = sparse_row(R, [i], [R(1)])
```
no longer needed due to https://github.com/thofma/Hecke.jl/pull/1502

lgoettgens (approved):
LGTM. Some minor cleanup since https://github.com/thofma/Hecke.jl/pull/1502 is already merged

thofma (commented):
If it works, it works. Not too familiar with the code, but looks good.

simonbrandhorst (approved):
Thanks a lot for these improvements!

lgoettgens:
Since this is already merged, I'll remove the triage label. If someone still wants to talk about this, please re-add.

PR #3682, opened 2024-05-03
Compositions from JuLie

joschmitt:
Move the compositions from JuLie to OSCAR.
I actually made the iterator call the existing `weak_compositions` iterator; the algorithm in JuLie came down to the same and this way it is less code.

I also moved the `ascending_partitions` out of experimental, named them `ascending_compositions` as discussed at some point and made the function return an iterator. I only kept one of the two implemented algorithms; the runtimes were basically the same.

Related to #3147.


thofma:
@mjrodgers can you have a (superficial) look?

src/Combinatorics/EnumerativeCombinatorics/compositions.jl
```diff
@@ -0,0 +1,376 @@
...
+
+By a composition of `n` we mean a sequence of positive integers whose sum is `n`.
+
+# Examples
+```jldoctest
+julia> C = compositions(4)
+Iterator over the compositions of 4
+
+julia> collect(C)
+8-element Vector{Oscar.Composition{Int64}}:
```
lgoettgens: The type `Partition` is exported and thus prints as `Vector{Partition{Int64}}` in this context. Maybe export `Combination` and `WeakCombination` as well?

mjrodgers:
> @mjrodgers can you have a (superficial) look?

I have looked it over, everything here looks good to me.

src/Combinatorics/EnumerativeCombinatorics/compositions.jl
joschmitt: Yes, why not.

lgoettgens (approved):

joschmitt:
In the 5th attempt the last CI job finally avoided 3676. Please merge.

PR #3683, opened 2024-05-03
Types for rational maps

HechtiDerLachs:
CC: @wdecker , @jankoboehm

### DIFF
--- a/src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Morphisms/Constructors.jl
+++ b/src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Morphisms/Constructors.jl
@@ -201,3 +201,31 @@ function base_change(phi::Any, IP::AbsProjectiveScheme)
   return result, psi
 end
 
+########################################################################
+# Rational maps
+########################################################################
+@doc raw"""
+    rational_map(P::AbsProjectiveScheme, Q::AbsProjectiveScheme, f::Map; check::Bool=true )
+
+Given a homomorphism ``f : T → S`` of the `homogeneous_coordinate_ring`s of `Q` and `P`, respectively, 
+construct the associated rational map of projective varieties.
+"""
+function rational_map(P::AbsProjectiveScheme, Q::AbsProjectiveScheme, f::Map; check::Bool=true )
+  return RationalMap(P, Q, f, check=check)
+end
+
+@doc raw"""
+    rational_map(X::AbsProjectiveScheme, Y::AbsProjectiveScheme, a::Vector{<:RingElem})
+
+Suppose ``X ⊂ ℙʳ`` and ``Y ⊂ ℙˢ`` are projective varieties.
+Construct the rational map associated to the morphism of graded rings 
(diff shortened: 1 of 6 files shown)


src/Rings/MPolyQuo.jl
```diff
@@ -1886,3 +1886,44 @@ end
...
+########################################################################
+
+function tensor_product(A::MPolyRing, B::MPolyRing)
+  kk = coefficient_ring(A)
+  @assert kk === coefficient_ring(B) "coefficient rings do not coincide"
+  all_symb = vcat(symbols(A), symbols(B))
+  r = ngens(A)
+  s = ngens(B)
+  res, _ = polynomial_ring(kk, all_symb, cached=false)
+  return res, hom(A, res, gens(res)[1:r]), hom(B, res, gens(res)[r+1:r+s])
```
fingolfin: I think this would be more natural (and allocates less):
```suggestion
  res, a, b = polynomial_ring(kk, symbols(A), symbols(B); cached=false)
  return res, hom(A, res, a), hom(B, res, b)
```

HechtiDerLachs: Thanks! Didn't know about this constructor.

simonbrandhorst: Out of curiosity.  When parents are created in a function we want `cached=false` even if they are part of the return data?  What is the rationale here?

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Morphisms/Constructors.jl
```diff
@@ -201,3 +201,31 @@ function base_change(phi::Any, IP::AbsProjectiveScheme)
...
 end
 
+########################################################################
+# Rational maps
+########################################################################
+@doc raw"""
+    rational_map(P::AbsProjectiveScheme, Q::AbsProjectiveScheme, f::Map; check::Bool=true )
+
+Given a homomorphism ``f : T → S`` of the `homogeneous_coordinate_ring`s of `Q` and `P`, respectively, 
+construct the associated rational map of projective varieties.
```
simonbrandhorst, line 211: This is defined for schemes but the documentation is for varieties.  
We have a type for projective varieties. I would suggest to use that?

fingolfin:
In triage, @jankoboehm was happy with this PR and @wdecker voiced no objections.

I suggest to address Simon's comment one way or another, and then get this merged RSN.

HereAround (approved):
@wdecker and @fieker and I like this PR. Thank you @HechtiDerLachs

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Morphisms/Constructors.jl
HechtiDerLachs: I thought about using the `AbsProjectiveVariety`, but in practice this will kill the user. We want to be able to compute with schemes which nevertheless have the properties of varieties. These are checked, eventually based on the actual type of the input.

PR #3686, opened 2024-05-03
Some fix for hypercomplexes

HechtiDerLachs:
CC: @wdecker .


HechtiDerLachs:
Failing tests are disabled in #3672 . We can either wait for that one to be merged, or merge this one already. 

But: The test added involves a rather large input. If there are objections to having this in the test-file, then I can try to take care of this. Maybe the surface can be loaded from some database, anyway? @wdecker?

wdecker:
Unfortunately, this surface is not from a data base and is actually just an intermediate product of some adjunction process.

wdecker (approved):

PR #3690, opened 2024-05-06
Use new is_terse and terse functions

fingolfin:


experimental/Schemes/WeilDivisor.jl
```diff
@@ -608,7 +608,7 @@ generated by rational functions ``f₁,…,fᵣ ∈ K(X)``.
 end
 
 function Base.show(io::IO, L::LinearSystem)
-  if get(io, :supercompact, true)
+  if isterse(io)
```
joschmitt:
```suggestion
  if is_terse(io)
```

src/Rings/orderings.jl
```diff
@@ -1719,7 +1723,7 @@ function induce(vars::Vector{Int}, o::ProdOrdering)
 end
 
 @doc raw"""
-    induce(vars::AbstractVector{<:MPolyRingElem}, ord::ModuleOrdering)
+    induce(vars::AbstractVector{<:MPolyRingElem}, ord::MonomialOrdering)
```
joschmitt: There are also some bits of #3679 in here?

lgoettgens:
The printing part of the dev docs need to be adapted as well; just applying your script led to some inconsistencies.

experimental/GModule/GaloisCohomology.jl
```diff
@@ -1390,7 +1390,7 @@ end
 function Base.show(io::IO, m::MIME"text/plain", a::RelativeBrauerGroupElem)
   io = pretty(io)
   print(io, "Element of relative Brauer group of ", Lowercase(), parent(a).k)
-  io = IOContext(io, :supercompact => true, :compact => true)
+  io = terse(io)
```
lgoettgens, line 1393: This no longer setting `:compact` makes the lhs of the below list print verbose, as https://github.com/thofma/Hecke.jl/blob/master/src/NumFieldOrd/NfOrd/Ideal/Ideal.jl#L148 does not use `is_terse`. There are some more occurrences in this file still using `:compact` and thus producing weird results.
This is the reason for failing doctests.

fingolfin: I'll add the `compact` back here, and then will have a look at replacing uses of `compact` in Hecke (so we can drop `compact` here in a future PR)

src/Rings/orderings.jl
fingolfin: Ouch! Good catch, not sure how that happened -- hopefully managed to remove all traces now.

fingolfin:
CI is green now

joschmitt (approved):
Looks good to me

fingolfin:
I've also changed the references to `supercompact` in the documentation to `terse`.

The only references to `supercompact` left are in tests where `sprint `or `repr` are called with `context = :supercompact => true`. I will provide add a replacement function to AA we can use to adjust this kind of test.

fingolfin:
See https://github.com/Nemocas/AbstractAlgebra.jl/pull/1701

lgoettgens (approved):

PR #3692, opened 2024-05-06
Foah reparametrization

emikelsons:
Reparametrized the "F-Theory on all toric hypersurfaces" literature models. Now that the constructor has been updated, we can have the same parametrization as in the paper.


emikelsons:
@HereAround @apturner

PR #3693, opened 2024-05-06
Reparametrized literature models

emikelsons:
Reparametrized the "F-Theory on all toric hypersurfaces" literature models. Now that the constructor has been updated, we can have the same parametrization as in the paper.
@HereAround @apturner


HereAround (approved):
Thank you @emikelsons . That is very much appreciated!

PR #3696, opened 2024-05-06
Use coordinates accessor function

fingolfin:
... instead of directly accessing members. In the future this will allow us to insert type assertions into the accessors for improved type stability

Extracted from #3679 so that it can be tested and merged now.

### DIFF
--- a/src/Modules/UngradedModules/Methods.jl
+++ b/src/Modules/UngradedModules/Methods.jl
@@ -354,7 +354,8 @@ function hom_matrices_helper(f1::MatElem{T}, g1::MatElem{T}) where T
       throw(DomainError("v does not represent a homomorphism"))
     end
     R = base_ring(M)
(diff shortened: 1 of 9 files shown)


joschmitt (approved):

src/Modules/UngradedModules/Methods.jl
```diff
@@ -354,7 +354,8 @@ function hom_matrices_helper(f1::MatElem{T}, g1::MatElem{T}) where T
       throw(DomainError("v does not represent a homomorphism"))
     end
     R = base_ring(M)
-    A = copy_and_reshape(dense_row(repres(v).coords[R, 1:s0*t0], s0*t0), s0, t0)
+    c = coordinates(repres(v))
+    A = copy_and_reshape(dense_row(c[R, 1:s0*t0], s0*t0), s0, t0)
```
fingolfin, line 358: See also PR #3697 for this odd syntax `c[R, 1:s0*t0]`

lgoettgens (approved):

PR #3697, opened 2024-05-07
Remove a weird and broken getindex method

fingolfin:
Also reduce code duplication in another.

This code has been in there since March 2021.

### DIFF
--- a/src/Modules/UngradedModules/Methods.jl
+++ b/src/Modules/UngradedModules/Methods.jl
@@ -233,16 +233,7 @@ ring_map(f::SubQuoHom) = f.ring_map
 #  re-evaluate and use or not
 
 function getindex(r::Hecke.SRow, u::AbstractUnitRange)
-  R = base_ring(r)
-  s = sparse_row(R)
-  shift = 1-first(u)
-  for (p,v) = r
-    if p in u
-      push!(s.pos, p+shift)
-      push!(s.values, v)
-    end
-  end
-  return s
+  return getindex(r, base_ring(r), u)
 end
 
 function getindex(r::Hecke.SRow, R::AbstractAlgebra.Ring, u::AbstractUnitRange)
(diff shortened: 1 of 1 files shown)


src/Modules/UngradedModules/Methods.jl
```diff
@@ -257,21 +248,6 @@ function getindex(r::Hecke.SRow, R::AbstractAlgebra.Ring, u::AbstractUnitRange)
   return s
 end
 
-function getindex(a::Hecke.SRow, b::AbstractVector{Int})
-  if length(a.pos) == 0
-    return a
-  end
-  m = minimum(b)
-  b = sparse_row(parent(a.values[1]))
```
fingolfin, line 265: Note how the input vector `b` is overwritten here, we only return the minimum value from that vector. I think this is a mistake and one of the two should be renamed if one wanted to fix this.

But I kinda hope nothing calls this method, I'd rather be rid of it here than trying to "fix" it.

src/Modules/UngradedModules/Methods.jl
```diff
@@ -233,16 +233,7 @@ ring_map(f::SubQuoHom) = f.ring_map
...
-    if p in u
-      push!(s.pos, p+shift)
-      push!(s.values, v)
-    end
-  end
-  return s
+  return getindex(r, base_ring(r), u)
 end
 
 function getindex(r::Hecke.SRow, R::AbstractAlgebra.Ring, u::AbstractUnitRange)
```
fingolfin, line 239: Note that this enables calls of the form `some_srow[R,1:r]` which I find really disturbing (this is how I learned about this function in the first place: I am touching a place calling it in PR #3696, inside `src/Modules/UngradedModules/Methods.jl`

joschmitt: Sounds to me like someone (TM) should remove this `getindex` method? Why should one specify the ring for the result in a `getindex`?

joschmitt (approved):
These `getindex` seem really dodgy and CI is happy. In my opinion, also the `getindex` with the ring should go completely, but that doesn't need to happen here.

src/Modules/UngradedModules/Methods.jl
jankoboehm: I guess this is supposed to extract the index values specified by the indices in b from the sparse row, returning a new sparse row with the extracted values. So the new sparse row should have a different name than b in the function.

fingolfin: Yeah that is my guess, too. But I argue that even when untangling `b` and the other `b` by renaming one, the semantics of the resulting code are wrong. E.g. it is legal to do `row[ [1,1,1] ]`  and this should return a list/vector/SRow/whatever of length 3 with three equal entries. The code below doesn't do that at all.

I don't want to waste time trying to fix this. If someone really needs this kind of functionality, it is easier to implement it from scratch (and ideally in Hecke, not here)

PR #3698, opened 2024-05-07
Disable a test in elliptic surfaces

joschmitt:
Disables the elliptic surface test that sometimes gets stuck, see #3676.

Something similar is done in #3672, but there seems to be some further discussion in that PR, so I would like to get this in now.

CC @benlorenz @HechtiDerLachs

### DIFF
--- a/test/AlgebraicGeometry/Schemes/elliptic_surface.jl
+++ b/test/AlgebraicGeometry/Schemes/elliptic_surface.jl
@@ -57,6 +57,8 @@
   trivial_lattice(X3)
   =#
 
+  #=
+  This testset is disabled (for now), see https://github.com/oscar-system/Oscar.jl/issues/3676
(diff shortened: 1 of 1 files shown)


benlorenz (approved):

joschmitt:
Please merge.

afkafkafk13:
> Disables the elliptic surface test that sometimes gets stuck, see #3676.
> 
> Something similar is done in #3672, but there seems to be some further discussion in that PR, so I would like to get this in now.
> 
> CC @benlorenz @HechtiDerLachs

#3672 is ready for merging, as soon as the book-test is fixed. @simonbrandhorst 's remark there is non-blocking.

lgoettgens:
Splitting this off makes it easier to revert and bisect in the future, so I will go on with this.

PR #3700, opened 2024-05-07
Fixes `mres` and `nres` in `free_resolution`

RafaelDavidMohr:
This PR will fix the `mres` and `nres` options in `free_resolution` using `prune_with_map`. Previously, using these options
might not return a minimal free resolution.

For using `prune_with_map` conveniently in `free_resolution`, I would like add a `minimal` keyword to `presentation`, if this is set to `true` we would return a minimal presentation, computed from the output of `prune_with_map`.

@wdecker Do we want a user-facing option to compute a minimal presentation? If not, then I will do this just internally.

### DIFF
--- a/src/Modules/UngradedModules/Presentation.jl
+++ b/src/Modules/UngradedModules/Presentation.jl
@@ -559,6 +559,50 @@ function prune_with_map(M::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}} # 
   return M_new, phi
 end
 
+function _presentation_minimal(M::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}}
+  R = base_ring(M)
+
+  # Prepare to set some names
+  br_name = AbstractAlgebra.get_name(R)
+  if br_name === nothing
+    br_name = "br"
+  end
+  
+  M_new, phi = prune_with_map(M)
+  F0 = ambient_free_module(M_new)
+
+  # M_new is a quotient of a free module
+  proj_map = hom(F0, M_new, gens(M_new), check=false)
+  F0_to_M = compose(proj_map, phi)
+  F0_to_M.generators_map_to_generators = true
+  AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(M.sub))")
+
+  K, inc = sub(F0, relations(M_new))
+  F1 = if is_graded(M)
+    graded_free_module(R, degrees_of_generators(K))
+  else
+    free_module(R, ngens(K))
+  end
+  F1_to_F0 = compose(hom(F1, K, gens(K), check=false), inc)
+  AbstractAlgebra.set_name!(F1, "$br_name^$(ngens(F1))")
+
+  # When there is no kernel, clean things up
+  if is_zero(F1)
+    F1 = FreeMod(R, 0)
+    F1_to_F0 = hom(F1, F0, elem_type(F0)[]; check=false)
+  end
+  
+  # prepare the end of the presentation
+  Z = FreeMod(R, 0)
+  AbstractAlgebra.set_name!(Z, "0")
+  M_to_Z = hom(M, Z, elem_type(Z)[zero(Z) for i in 1:ngens(M)]; check=false)
+
+  # compile the presentation complex
+  CC = Hecke.ComplexOfMorphisms(ModuleFP, ModuleFPHom[F1_to_F0, F0_to_M, M_to_Z], check=false, seed = -2)
+  set_attribute!(CC, :show => Hecke.pres_show)
+  return CC
+end
+
 function prune_with_map(F::FreeMod)
   return F, hom(F, F, gens(F))
 end
--- a/src/Modules/UngradedModules/Presentation.jl
+++ b/src/Modules/UngradedModules/Presentation.jl
@@ -563,7 +563,8 @@ function prune_with_map(M::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}} # 
   return M_new, phi
 end
 
-function _presentation_minimal(SQ::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}}
+function _presentation_minimal(SQ::ModuleFP{T};
+                               minimal_kernel::Bool=true) where {T<:MPolyRingElem{<:FieldElem}}
   R = base_ring(SQ)
 
   # Prepare to set some names
@@ -581,7 +582,12 @@ function _presentation_minimal(SQ::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldE
   F0_to_SQ.generators_map_to_generators = true
   AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(F0))")
 
+  # if we want a minimal kernel too, we go to prune_with_map
   K, inc = sub(F0, relations(SQ_new))
+  if minimal_kernel
+    K, inc2 = prune_with_map(K)
+    inc = compose(inc2, inc)
+  end
   F1 = if is_graded(SQ)
     graded_free_module(R, degrees_of_generators(K))
   else
--- a/src/Modules/UngradedModules/Presentation.jl
+++ b/src/Modules/UngradedModules/Presentation.jl
@@ -579,7 +579,7 @@ function _presentation_minimal(SQ::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldE
   proj_map = hom(F0, SQ_new, gens(SQ_new), check=false)
   F0_to_SQ = compose(proj_map, phi)
   F0_to_SQ.generators_map_to_generators = true
-  AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(SQ.sub))")
+  AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(F0))")
 
   K, inc = sub(F0, relations(SQ_new))
   F1 = if is_graded(SQ)
--- a/src/Modules/UngradedModules/Presentation.jl
+++ b/src/Modules/UngradedModules/Presentation.jl
@@ -1,11 +1,15 @@
 ####################
 
 @doc raw"""
-    presentation(M::SubquoModule)
+    presentation(M::SubquoModule; minimal=false)
 
-Return a free presentation of `M`. 
+Return a free presentation of `M`. If `minimal` is set to `true`, the returned
+presentation is minimal.
 """
-function presentation(SQ::SubquoModule)
+function presentation(SQ::SubquoModule;
+                      minimal=false)
+  if minimal
+    return _presentation_minimal(SQ)
   if is_graded(SQ)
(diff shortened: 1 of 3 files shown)


wdecker:
@RafaelDavidMohr As far as I can see, the new functionality is only used with `mres`. If everybody agrees, we do not need the proposed keyword. Or?

wdecker:
@jankoboehm userfacing function named `minimal_presentation`?

jankoboehm:
Yes we want the keyword. There are use-cases for both options.

src/Modules/UngradedModules/Presentation.jl
```diff
@@ -1,11 +1,15 @@
...
-Return a free presentation of `M`. 
+Return a free presentation of `M`. If `minimal` is set to `true`, the returned
+presentation is minimal.
 """
-function presentation(SQ::SubquoModule)
+function presentation(SQ::SubquoModule;
+                      minimal=false)
+  if minimal
+    return _presentation_minimal(SQ)
   if is_graded(SQ)
```
fingolfin:
```suggestion
  elseif is_graded(SQ)
```

src/Modules/UngradedModules/Presentation.jl
```diff
@@ -559,6 +563,50 @@ function prune_with_map(M::ModuleFP{T}) where {T<:MPolyRingElem{<:FieldElem}} #
...
+  end
+  
+  SQ_new, phi = prune_with_map(SQ)
+  F0 = ambient_free_module(SQ_new)
+
+  # M_new is a quotient of a free module
+  proj_map = hom(F0, SQ_new, gens(SQ_new), check=false)
+  F0_to_SQ = compose(proj_map, phi)
+  F0_to_SQ.generators_map_to_generators = true
+  AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(SQ.sub))")
```
fingolfin: Why do you ned to set this name here and below? Since F0 and F1 are free module anyway, couldn't this be done generally as part of the `show` methods for free modules?

As it is, this hardcodes the name of the basering, so if it changes later on, it will look ugly. Likewise if there is no name set, you hardcode `br` forever.

lgoettgens: This is the current standard for presentations/resolutions/etc. and the same at other places. Thus I would keep it like this here right now and would instead think of some way to replace all of these instances by some AA printing magic in the future.

RafaelDavidMohr:
@wdecker @jankoboehm do you have an example where `presentation` produces something non-minimal that I can use as a test for the minimal free resolution?

wdecker:
This example should do:

```julia
X = rational_d9_pi6();
Pn = ambient_coordinate_ring(X)
A = homogeneous_coordinate_ring(X)
n = ngens(Pn)-1
c = codim(X)
FA = free_resolution(A, algorithm = :fres)
C = Oscar.SimpleComplexWrapper(FA.C[0:first(range(FA.C))])
C_simp = simplify(C)
C_shift = shift(C_simp, c)
OmegaPn = graded_free_module(Pn, [n+1])
D = hom(C_shift, OmegaPn)
D_simp = simplify(D)
Z, inc = kernel(D_simp, 0)
B, inc_B = Oscar.boundary(D_simp, 0)
M = SubquoModule(D_simp[0], ambient_representatives_generators(Z), ambient_representatives_generators(B))
```

wdecker:
one more:

S, (x0, x1, x2, x3, x4) = graded_polynomial_ring(GF(3), ["x0", "x1", "x2", "x3", "x4"]);
m = ideal(S, [x1^2+(-x1+x2+x3-x4)*x0, x1*x2+(x1-x3+x4)*x0, x1*x3+(-x1+x4+x0)*x0, x1*x4+(-x1+x3+x4-x0)*x0, x2^2+(x1-x2-x4-x0)*x0, x2*x3+(x1-x2+x3+x4-x0)*x0, x2*x4+(x1+x2-x3-x4-x0)*x0, x3^2+(x3+x4-x0)*x0,x3*x4+(-x3-x4+x0)*x0, x4^2+(x1+x3-x4-x0)*x0]);
Qm, _ = quo(S, m);
FQm = free_resolution(Qm)
M, _ = image(map(FQm, 1))

RafaelDavidMohr:
> one more: S, (x0, x1, x2, x3, x4) = graded_polynomial_ring(GF(3), ["x0", "x1", "x2", "x3", "x4"]); m = ideal(S, [x1^2+(-x1+x2+x3-x4)*x0, x1*x2+(x1-x3+x4)*x0, x1*x3+(-x1+x4+x0)*x0, x1*x4+(-x1+x3+x4-x0)*x0, x2^2+(x1-x2-x4-x0)*x0, x2*x3+(x1-x2+x3+x4-x0)*x0, x2*x4+(x1+x2-x3-x4-x0)*x0, x3^2+(x3+x4-x0)*x0,x3*x4+(-x3-x4+x0)*x0, x4^2+(x1+x3-x4-x0)*x0]); Qm, _ = quo(S, m); FQm = free_resolution(Qm) M, _ = image(map(FQm, 1))

Do we know the correct ranks for the minimal free resolution in this example so I can test?

wdecker:
In `_presentation_minimal` please catch border case of a free module before entering
`AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(SQ.sub))")`

wdecker:
There is a problem with the print function:

julia> R, (x,y,z) = polynomial_ring(QQ, ["x", "y", "z"]);

julia> A = R[x 1 y; 2*x^2 z^2 3*y^2];

julia> M = cokernel(A);

julia> PM = presentation(M)
0 <---- M <---- R^3 <---- R^2

julia> MPM = _presentation_minimal(M)
0 <---- M <---- R^3 <---- R^1

julia> matrix(map(MPM,1))
[-2*x^2 + x*z^2   -3*y^2 + y*z^2]

julia> MPM[0]
Free module of rank 2 over R

RafaelDavidMohr:
> In `_presentation_minimal` please catch border case of a free module before entering `AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(SQ.sub))")`

This case is caught in `presentation` already, and the user will only access `_presentation_minimal` via the `minimal` kw of presentation. Should I nonetheless treat this case seperately?

RafaelDavidMohr:
> M = cokernel(A);

Fixed with https://github.com/oscar-system/Oscar.jl/pull/3700/commits/f2a684b09aa4e0b3d6f1bec15d624479e596465d

wdecker:
I am not sure what our policy here is. It may happen that a future developer may use this internal function and run into the border case. @fingolfin what do you think?

src/Modules/UngradedModules/Presentation.jl
```diff
@@ -1,12 +1,16 @@
...
+presentation is minimal.
 """
-function presentation(SQ::SubquoModule)
-  if is_graded(SQ)
+function presentation(SQ::SubquoModule;
+                      minimal=false)
+  if minimal
+    return _presentation_minimal(SQ)
+  elseif is_graded(SQ)
     return _presentation_graded(SQ)
```
lgoettgens, line 14: How do `minimal=true` and a graded module interact? Due to now calling `_presentation_minimal` here, the graded check is skipped. So `_presentation_minimal` would need to special-case graded modules somehow

RafaelDavidMohr: Internally, `_presentation_minimal` uses `prune_with_map` which has to check for gradedness anyway. So `_presentation_minimal` does not have to do any extra checks.

lgoettgens: Nice, didn't know that

fingolfin:
@wdecker I am afraid I don't understand what you are asking about. What is "this internal function", and which border case?

Happy to talk about it during triage today if anyone would like to.

src/Modules/UngradedModules/Presentation.jl
fingolfin: Yeah if this is pervasive we should not try to deal with it in this PR. But perhaps we can open an issue as a reminder? 

Or perhaps we just add it to https://github.com/oscar-system/Oscar.jl/issues/2891

lgoettgens: See https://github.com/oscar-system/Oscar.jl/issues/3730

lgoettgens:
> @wdecker I am afraid I don't understand what you are asking about. What is "this internal function", and which border case?
> 
> Happy to talk about it during triage today if anyone would like to.

I think I understood the issue and can try to explain it later in triage if wdecker is not there

RafaelDavidMohr:
I added the discussed minimization of the kernel in `_presentation_minimal`. I added it with an optional keyword, so that it can easily be integrated into `free_resolution`.

Currently this uses `prune_with_map`. I tried to use `Singular.mres` but this function seems to ignore the `length` argument. 

@wdecker Should I make a Singular issue regarding this?

ederc:
Can you give a concrete example where `Singular.mres` ignores the `length` argument? If I compute a free resolution with current Oscar master using `algorithm = :mres` and `length` set, it seems to compute up to the correct `length`.

wdecker (approved):
This is ready to go as soon as the tests are green. Please check tests.

RafaelDavidMohr:
> Can you give a concrete example where `Singular.mres` ignores the `length` argument? If I compute a free resolution with current Oscar master using `algorithm = :mres` and `length` set, it seems to compute up to the correct `length`.

Maybe I'm misunderstanding the concept of length of a free resolution, but for example
```
julia> R, (x, y, z) = polynomial_ring(GF(65521), ["x", "y", "z"]);

julia> I = ideal(R, [x^2, x*y, y]);

julia> M = ideal_as_module(I);

julia> sf = Oscar.singular_module(ambient_free_module(M));

julia> sm = Singular.Module(base_ring(sf), [sf(repres(g)) for g in gens(M)]...);

julia> Singular.mres(sm, 1)
Singular resolution: R^1 <- R^2 <- R^1

julia> Singular.mres(sm, 2)
Singular resolution: R^1 <- R^2 <- R^1

julia> Singular.mres(sm, 3)
Singular resolution: R^1 <- R^2 <- R^1
```

wdecker:
@RafaelDavidMohr In your example, the resolution is short, so entering an increased length will not change anything:

julia> R, (x, y) = polynomial_ring(GF(65521), ["x", "y"]);

julia> I = ideal(R, [x^2, x*y, y]);

julia> M = ideal_as_module(I);

julia> FM = free_resolution(M, algorithm = :mres, length = 1)
Free resolution of M
R^3 <---- R^2
0         1

julia> FM = free_resolution(M, algorithm = :mres, length = 0)
Free resolution of M
R^3 <---- R^2 <---- 0
0         1         2


So why now does Singular print this:

julia> Singular.mres(sm, 1)
Singular resolution: R^1 <- R^2 <- R^1

I guess, this is due to the fact that you use `mres`. So you have to go one step further in order to minimalize. @hannes14 ?

PR #3702, opened 2024-05-08
Deprecate another weird getindex method

lgoettgens:
x-ref https://github.com/oscar-system/Oscar.jl/pull/3697#pullrequestreview-2042645170

This time, I deprecated it instead of removing since it worked according to its semantics, and straight up removing it would be a breaking change. Let's wait for the `depwarn=error` CI job to see if I caught all callsites of this inside Oscar.


joschmitt (approved):
Thanks!

fingolfin (approved):

PR #3705, opened 2024-05-08
Minimal associated primes

HechtiDerLachs:
An attempt to make the specialized functionality in Singular for zero dimensional ideals available. 

This seemed to be useful for @simonbrandhorst in some examples, but now I can't even get the tests to terminate. Let's see what the CI says. 

@wdecker : The [documentation](https://www.singular.uni-kl.de/Manual/latest/sing_1261.htm#SEC1342) reads as if only `QQ` was allowed as a coefficient ring. Do you remember whether this is the case? Because it seems to have run also over number fields.


HechtiDerLachs:
@ederc : Could you have a look at the tests? I don't understand what's going wrong. But maybe `minimal_primes` now called `singular_generators` before and that lead to some false caching?

simonbrandhorst:
Thanks for wrapping this @HechtiDerLachs !

ederc:
Well, the tests fail because you try to apply `assPrimes` from Singular on a zero dimensional ideal over a finite field, but `assPrimes` assumes that the ideal lives over `QQ`. AFAIK there is a modular `std` call behind the scenes in `assPrimes`.

HechtiDerLachs:
The problem I meant occured in an earlier test run. Some call to `dim` was complaining that `singular_groebner_generators` did not return a groebner basis. But it doesn't seem to be reproduced anymore; sorry for calling you. 

But what you say about `minAss` is good to know. I will restrict the cases where it's called.

HechtiDerLachs:
@ederc : Now the error is back, see [this test](https://github.com/oscar-system/Oscar.jl/actions/runs/9006575929/job/24744417543?pr=3705).

ederc:
There was a missing caching of the `isGB` flag if a GB was computed during `small_generating_set`. I pushed a fix to your branch.

simonbrandhorst:
is there anything holding this PR up @HechtiDerLachs ?

HechtiDerLachs:
The tests were still failing. I just had a look and it seems there is another bug in `minimal_primes` (due to me, unfortunately). I will have a look.

HechtiDerLachs:
@ederc : I'm sorry, but it looks like I accidentally overwrote your fixes to this branch when doing a rebase. Do you still have them somewhere? If yes, could you push them here again? Thx!

ederc:
I do not have this code anymore, we need to look again where the `isGB` flag needs to be set.

benlorenz:
> @ederc : I'm sorry, but it looks like I accidentally overwrote your fixes to this branch when doing a rebase. Do you still have them somewhere? If yes, could you push them here again? Thx!

The Github UI shows `HechtiDerLachs force-pushed the minimal_associated_primes branch from 9aa5177 to 2cd696e`, and clicking on that 9aa5177 commit should get you to the last commit before the force push, and the [history](https://github.com/oscar-system/Oscar.jl/commits/9aa5177a57c67a042ee5c43719e9fd90ab02a1da) is also available. With that commit hash known you can also try doing `git log 9aa5177` locally since git will keep lost commits for a few days.
You could also look through `git reflog` locally.

ederc:
@HechtiDerLachs Caching `isGB` is back in.

HechtiDerLachs:
Thanks a lot @ederc and @benlorenz ! I was hoping that something like this was possible, but didn't know how. 

Unfortunately it seems that a lot of tests time out. Or something else goes wrong which I do not fully understand, yet. 

Edit: I checked two of the failing tests locally around the point where they were cancelled and they run just fine on my machine.

HechtiDerLachs:
I really can't make sense of the failing tests. Everything that I tried on my local machine about things where the CI gets stuck really goes through for me. And in some cases I can't even find the error messages.

fingolfin:
Triage talked about it yesterday and we suspect that it's the same problem as in #3905 i.e. Singular tries to start subprocesses and gets stuck. So hopefully @hannes14 can take a look when he's back from vacation.

fieker:
Its still under Hans aegis: 
 - it is fixed (in Singular)
 - it needs the release cascade
 - (maybe)

fingolfin:
@hannes14 [updated Singular_jll](https://github.com/JuliaPackaging/Yggdrasil/pull/9333) and just told me that Singular.jl should automatically use that. So I've now restarted the tests, let's see if it helps.

benlorenz:
I looked into the errors / warnings a bit, so far I have found the following:
```
│    Can't call method "instance_for_owner" on an undefined value at /Users/aaruni/Desktop/oscar-runners/runner-4/julia-depot/scratchspaces/d720cf60-89b5-51f5-aff5-213f193123e7/polymake_15999920327334887436_1.10_depstree_v2/share/polymake/perllib/Polymake/Core/BigObject.pm line 775 during global destruction.
```
This is triggered from an atexit handler from the perl code which tries to clean up polymake objects. Not totally sure why this error doesn't appear during normal process exit, maybe because some other finalizers need to run first but these are not triggered correctly for these processes.

```
polymake:  WARNING: Automatic update of the interface definition file /home/runner/.julia/scratchspaces/d720cf60-89b5-51f5-aff5-213f193123e7/polymake_11299506659249835626_1.10_userdir/wrappers.0/apps/fan/cpperl/wrap-check_fan.cpperl refused:
2024-09-04T09:36:57.9086549Z Timestamp file /home/runner/.julia/scratchspaces/d720cf60-89b5-51f5-aff5-213f193123e7/polymake_11299506659249835626_1.10_userdir/wrappers.0/build/Opt/.apps.built is missing unxpectedly.
```
This is also during process exit, maybe because multiple processes are trying to modify the same files at the same time (and this is not really a place where we can add a pidlock). All these processes are trying to save the same new auto-generated code that was probably triggered in the original process.

```
      From worker 4:	libc++abi: terminating due to uncaught exception of type std::__1::bad_function_call: std::exception
      From worker 4:	
      From worker 4:	[68228] signal (6): Abort trap: 6
      From worker 4:	in expression starting at /Users/aaruni/Desktop/oscar-runners/runner-3/_work/Oscar.jl/Oscar.jl/test/AlgebraicGeometry/Schemes/resolutions.jl:1
      From worker 4:	__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
      From worker 4:	Allocations: 509000079 (Pool: 508678466; Big: 321613); GC: 90
```
This is in a [macos job](https://github.com/oscar-system/Oscar.jl/actions/runs/10698681595/job/29658626313#step:9:2171) and I don't really know what exactly triggers this, but this could also be due to code running at exit without some julia exit handlers running before this, maybe related to this:
https://github.com/oscar-system/Polymake.jl/blob/master/src/Polymake.jl#L317-L322

**Note:**
All these subprocesses dying during exit should not really cause the tests to fail (only the doctests due to the printing).

Many of the CI jobs also timed out, something got a lot slower here or got stuck somewhere?

benlorenz:
I ran the Oscar tests on this branch locally and after a while it also got stuck with many processes (22 processes including the main one), all of them seem to be sleeping / waiting.
(The PC has 12 Cores / 24 Threads, AMD Ryzen 9 5900X)

The main process (28895) is deep inside some singular code, called from `minimal_primes`:
```
Test Summary:             | Pass  Total  Time
mpolyquo-localizations.jl |   37     37  2.1s

======================================================================================
Information request received. A stacktrace will print followed by a 1.0 second profile
======================================================================================

cmd: /net/site-local.linux64/julia/julia-1.10.5/bin/julia 28895 running 1 of 1

signal (10): User defined signal 1
__poll at /lib64/libc.so.6 (unknown line)
_Z12slStatusSsiLP6slistsiPi at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z10jjWAIT1ST1P6sleftvS0_ at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z15iiExprArith1TabP6sleftvS0_iPK8sValCmd1iPK13sConvertTypes at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z7yyparsev at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z10iiAllStartP8procinfoPKc13feBufferTypesi at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z8iiPStartP5idrecP6sleftv at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z11iiMake_procP5idrecP11sip_packageP6sleftv at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z6jjPROCP6sleftvS0_S0_ at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_ZL21iiExprArith2TabInternP6sleftvS0_iS0_iPK8sValCmd2iiPK13sConvertTypes.part.85 at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
...
_Z10iiAllStartP8procinfoPKc13feBufferTypesi at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z8iiPStartP5idrecP6sleftv at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z11iiMake_procP5idrecP11sip_packageP6sleftv at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z15ii_CallLibProcMPKcPPvPiP8ip_sringRi at /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so (unknown line)
_Z31call_singular_library_procedureNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEP8ip_sringN5jlcxx8ArrayRefIP11_jl_value_tLi1EEE at /home/datastore/lorenz/software/julia/depot/artifacts/9e729ade26a239c52713155b6c560dd4615b31a5/lib/libsingular_julia.so (unknown line)
_ZNSt17_Function_handlerIFP11_jl_value_tNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEP8ip_sringN5jlcxx8ArrayRefIS1_Li1EEEEPSD_E9_M_invokeERKSt9_Any_dataOS7_OS9_OSC_ at /home/datastore/lorenz/software/julia/depot/artifacts/9e729ade26a239c52713155b6c560dd4615b31a5/lib/libsingular_julia.so (unknown line)
_ZN5jlcxx6detail11CallFunctorIP11_jl_value_tJNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEP8ip_sringNS_8ArrayRefIS3_Li1EEEEE5applyEPKvNS_13WrappedCppPtrESH_P10jl_array_t at /home/datastore/lorenz/software/julia/depot/artifacts/9e729ade26a239c52713155b6c560dd4615b31a5/lib/libsingular_julia.so (unknown line)
call_singular_library_procedure at /home/datastore/lorenz/software/julia/depot/packages/CxxWrap/5IZvn/src/CxxWrap.jl:624
unknown function (ip: 0x148fccf9d9e0)
_jl_invoke at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:2895 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:3077
low_level_caller at /home/datastore/lorenz/software/julia/depot/packages/Singular/A4riV/src/caller.jl:399
assPrimes at /home/datastore/lorenz/software/julia/depot/packages/Singular/A4riV/src/Meta.jl:44
unknown function (ip: 0x148fccf9d2b5)
_jl_invoke at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:2895 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:3077
#minimal_primes#346 at /home/datastore/lorenz/software/julia/Oscar.jl/src/Rings/mpoly-ideals.jl:1072
minimal_primes at /home/datastore/lorenz/software/julia/Oscar.jl/src/Rings/mpoly-ideals.jl:1068 [inlined]
__compute_is_prime__ at /home/datastore/lorenz/software/julia/Oscar.jl/src/Rings/mpoly-ideals.jl:1663
#373 at /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/sxDV6/src/Attributes.jl:357 [inlined]
get! at ./dict.jl:479
get_attribute! at /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/sxDV6/src/Attributes.jl:230 [inlined]
is_prime at /home/datastore/lorenz/software/julia/Oscar.jl/src/Rings/mpoly-ideals.jl:1662
unknown function (ip: 0x148fca651095)
_jl_invoke at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:2895 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:3077
__compute_is_prime__ at /home/datastore/lorenz/software/julia/Oscar.jl/src/Rings/mpolyquo-localizations.jl:2023
#1001 at /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/sxDV6/src/Attributes.jl:357 [inlined]
get! at ./dict.jl:479
get_attribute! at /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/sxDV6/src/Attributes.jl:230 [inlined]
is_prime at /home/datastore/lorenz/software/julia/Oscar.jl/src/Rings/mpolyquo-localizations.jl:2022
unknown function (ip: 0x148fa7ea52f5)
_jl_invoke at /cache/build/builder-amdci4-4/julialang/julia-release-1-dot-10/src/gf.c:2895 [inlined]
```

One Process (1803) is in a wait call:
```
0x0000148febf0fc1f in wait4 () from /lib64/libc.so.6
#0  0x0000148febf0fc1f in wait4 () from /lib64/libc.so.6
#1  0x0000148f53324606 in ssiClose(sip_link*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#2  0x0000148f533232e4 in slClose(sip_link*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#3  0x0000148f532c9445 in iiExprArith1Tab(sleftv*, sleftv*, int, sValCmd1 const*, int, sConvertTypes const*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#4  0x0000148f532a0977 in yyparse() () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#5  0x0000148f532e414c in iiAllStart(procinfo*, char const*, feBufferTypes, int) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
...
```

One Process (1843) is in a futex_wait (called from some openmp exit handlers?):
```
Using host libthread_db library "/lib64/libthread_db.so.1".
futex_wait (val=120, addr=0x14b2b2044) at /workspace/srcdir/gcc-13.2.0/libgomp/config/linux/x86/futex.h:97
97      /workspace/srcdir/gcc-13.2.0/libgomp/config/linux/x86/futex.h: No such file or directory.
#0  futex_wait (val=120, addr=0x14b2b2044) at /workspace/srcdir/gcc-13.2.0/libgomp/config/linux/x86/futex.h:97
#1  do_wait (val=120, addr=<optimized out>) at /workspace/srcdir/gcc-13.2.0/libgomp/config/linux/wait.h:67
#2  gomp_team_barrier_wait_end (bar=0x14b2b2040, state=120) at /workspace/srcdir/gcc-13.2.0/libgomp/config/linux/bar.c:112
#3  0x0000148feaa8c00e in gomp_team_barrier_wait_final (bar=bar@entry=0x14b2b2040) at /workspace/srcdir/gcc-13.2.0/libgomp/config/linux/bar.c:136
#4  0x0000148feaa8a96d in gomp_team_end () at /workspace/srcdir/gcc-13.2.0/libgomp/team.c:956
#5  0x0000148febe7bbd9 in __run_exit_handlers () from /lib64/libc.so.6
#6  0x0000148febe7bd6a in exit () from /lib64/libc.so.6
#7  0x0000148f53313bc2 in m2_end () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#8  0x0000148f53328e10 in ssiRead1(sip_link*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#9  0x0000148f5332c404 in ssiOpen(sip_link*, short, sleftv*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#10 0x0000148f53323174 in slOpen(sip_link*, short, sleftv*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#11 0x0000148f532c9445 in iiExprArith1Tab(sleftv*, sleftv*, int, sValCmd1 const*, int, sConvertTypes const*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
```


And finally there are further 19 processes which are waiting in a read like this:
```
0x0000148fec03a754 in read () from /lib64/libpthread.so.0
#0  0x0000148fec03a754 in read () from /lib64/libpthread.so.0
#1  0x0000148f53ed2f89 in s_getc(s_buff_s*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libpolys.so
#2  0x0000148f53ed31e4 in s_readint(s_buff_s*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libpolys.so
#3  0x0000148f53328dae in ssiRead1(sip_link*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#4  0x0000148f53323421 in slRead(sip_link*, sleftv*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#5  0x0000148f532af29e in jjREAD(sleftv*, sleftv*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
#6  0x0000148f532c9445 in iiExprArith1Tab(sleftv*, sleftv*, int, sValCmd1 const*, int, sConvertTypes const*) () from /home/datastore/lorenz/software/julia/depot/artifacts/178cf93ab258b57deb473bfa44f78c5994f8628c/lib/libSingular.so
...
```

I can keep those processes alive for a while if there is anything I should check, they don't need much memory and all are idle anyway.

fingolfin:
The `atexit` related crashes are fixed in @hannes14 most recent update from last week. But there are still errors / crashes, just again different ones

benlorenz:
There are still exit handlers running
```
│    Can't call method "flags" on an undefined value at /Users/aaruni/Desktop/oscar-runners/runner-2/julia-depot/scratchspaces/d720cf60-89b5-51f5-aff5-213f193123e7/polymake_6360604600323823479_1.10_depstree_v2/share/polymake/perllib/Polymake/Core/BigObject.pm line 1247 during global destruction.
```
which I think is problematic. But at least I don't see any stuck jobs that ran into a timeout.

Many of the failing jobs don't have any logs for the `Run tests` step, some show an exit code of 143 (sigterm) which might indicate the workers running out of memory? (due to multiple processes?)

fingolfin:
@hannes14 so I guess that ball is in your court again ;-).

Out of curiosity, how does Singular determine how many processes to fork? Does it take the number of cores and/or RAM into account?

hannes14:
Singular uses sysconf(_SC_NPROCESSORS_ONLN) resp. sysconf(_SC_NPROCESSORS_CONF), 
RAM is not considered

src/Rings/mpoly-ideals.jl
```diff
@@ -1133,18 +1139,34 @@ function minimal_primes(
...
+    for q in J
+      is_one(q) && continue
+      q in unique_comp && continue
+      push!(unique_comp, q)
+    end
+    J = unique_comp
+    result = typeof(I)[]
+    # `unique!` does not work for lists of ideals. I don't know why, but for the moment we need the 
+    # following workaround.
+    for p in filter!(!is_one, vcat([minimal_primes(j; algorithm, factor_generators=false) for j in J]...))
```
```suggestion
    for p in filter!(!is_one, reduce(vcat, [minimal_primes(j; algorithm, factor_generators=false) for j in J]))
```

src/Rings/mpoly-ideals.jl
```diff
@@ -1191,17 +1197,27 @@ function minimal_primes(
...
       J = K
     end
+
+    unique_comp = typeof(I)[]
+    for q in J
+      is_one(q) && continue
+      q in unique_comp && continue
+      push!(unique_comp, q)
+    end
+    J = unique_comp
```
fingolfin: @HechtiDerLachs I resolved the merge conflicts -- I hope I didn't screw that up too badly...

fingolfin:
Tests pass now, but `test/AlgebraicGeometry/Schemes/Resolution_structure.jl` in Julia 1.10 went from ~64 seconds to ~370 seconds

src/Rings/mpoly-ideals.jl
```diff
@@ -1191,17 +1197,27 @@ function minimal_primes(
...
           push!(K, j + ideal(R, b))
         end
       end
       J = K
     end
+
+    unique_comp = typeof(I)[]
+    for q in J
+      is_one(q) && continue
+      q in unique_comp && continue
```
fingolfin: This compares ideals, won't that trigger additional GB computations and thus be slow in general?? Maybe you just want to remove "obviously" identical ideals (with generators lists being the same, up to reordering?)

src/Rings/mpoly-ideals.jl
```diff
@@ -1126,6 +1126,12 @@ julia> L = minimal_primes(I)
 function minimal_primes(I::MPolyIdeal; algorithm::Symbol = :GTZ, cache::Bool=true)
   has_attribute(I, :minimal_primes) && return get_attribute(I, :minimal_primes)::Vector{typeof(I)}
   R = base_ring(I)
+  if coefficient_ring(R) isa QQField && is_zero(dim(I))
```
fingolfin: `dim(I)` triggers a GB computation?

fingolfin: Triage suggest this would perhaps make sense with `is_known_to_be_zero_dimensional(I)` check -- i.e. if we know this information, we can call the 0-dim method.

But otherwise, leave it singular -- if it detects the 0-dim case it still handles it well.

simonbrandhorst:
I can confirm that the test timings go up. The reason was not the call to `dim` but that 
`Singular.LibAssprimeszerodim.assPrimes` was slower in all cases I could produce. 

Here is an example.
```
julia> R, (x, y) = polynomial_ring(QQ, [:x, :y])
(Multivariate polynomial ring in 2 variables over QQ, QQMPolyRingElem[x, y])

julia> I = ideal(R,[x^5*y - 2*x^4*y^2 - 2*x^4*y + x^3*y^3 + 2*x^3*y^2 - x^2*y^3 + 2*x^2*y^2 + 2*x^2*y + 2*x*y^4 + x*y^3 - 2*x*y^2 - x*y - y^5 - 2*y^4 - y^3, x^6 - 2*x^5 - 3*x^4*y^2 - 2*x^4*y + 2*x^3*y^3 + 3*x^3*y^2 + 2*x^3*y + 2*x^3 + 5*x^2*y^2 + 2*x^2*y - x^2 + 3*x*y^4 - 5*x*y^2 - 2*x*y - 2*y^5 - 4*y^4 - 2*y^3, x^7 + y^3 - 6]
       )
Ideal generated by
  x^5*y - 2*x^4*y^2 - 2*x^4*y + x^3*y^3 + 2*x^3*y^2 - x^2*y^3 + 2*x^2*y^2 + 2*x^
  2*y + 2*x*y^4 + x*y^3 - 2*x*y^2 - x*y - y^5 - 2*y^4 - y^3
  x^6 - 2*x^5 - 3*x^4*y^2 - 2*x^4*y + 2*x^3*y^3 + 3*x^3*y^2 + 2*x^3*y + 2*x^3 + 
  5*x^2*y^2 + 2*x^2*y - x^2 + 3*x*y^4 - 5*x*y^2 - 2*x*y - 2*y^5 - 4*y^4 - 2*y^3
  x^7 + y^3 - 6

julia> @time minimal_primes(I)
  0.031095 seconds (54.79 k allocations: 1.706 MiB)
2-element Vector{MPolyIdeal{QQMPolyRingElem}}:
 Ideal with 2 generators
 Ideal (y^7 + 7*y^6 + 21*y^5 + 35*y^4 + 36*y^3 + 21*y^2 + 7*y - 5, x - y - 1)

julia> @time Oscar.minimal_primes_zero_dim(I)
 20.402501 seconds (108.43 k allocations: 1.781 MiB)
2-element Vector{MPolyIdeal{QQMPolyRingElem}}:
 Ideal with 5 generators
 Ideal with 5 generators

julia> dim(I)
0

```
Therefore it should not be the default algorithm. Maybe there exists input where it is faster. I do not know.
do you ? @hannes14  
Since this PR also contains some fixes, I suggest to move forward with it.

src/Rings/mpoly-ideals.jl
afkafkafk13: @fingolfin : is_one(q) triggers a GB computation, but (at least hopefully) the ideal membership test in 1240 will use those GB from the prior runs of the loop for the comparison.

I rather wonder whether the factorization in Oscar is so much superior that it does not make sense to do a `facstd` of J in Singular and then go on with the components from there. 

Edit: If it is about working over field extensions, I understand the point.

src/Rings/mpoly-ideals.jl
```diff
@@ -1172,6 +1172,24 @@ function minimal_primes(I::MPolyIdeal; algorithm::Symbol = :GTZ, cache::Bool=tru
   return V
 end
 
+@doc raw"""
+    function minimal_primes_zero_dim(I::MPolyIdeal{QQPolyRingElem})
+    
+Return a vector containing the minimal associated prime ideals of `I`.
+
+Calls a specialized variant of `minimal_primes` for zero dimensional ideals 
+with coefficient ring over the rationals.
```
afkafkafk13:
```suggestion
Calls a specialized modular variant of `minimal_primes` for zero dimensional ideals 
with coefficient ring over the rationals.
```

afkafkafk13 (commented):
The two bug fixes make sense. 

The benefit of the new method is rather limited and seems to rely on Singular's parallization via modular.lib. Hence we should at least inform the user that this is using a modular approach. -- minor change, but approval has to wait for it.

src/Rings/mpoly-ideals.jl
thofma: What should one take away from this additional information? That the result may not be correct anymore? Consider me a user/developer, who stumbles upon this method and reads the docstring.

afkafkafk13: Any better hint for the knowledgable user that there is a modular approach and Singular-level parallelization going on in the background is welcome -- but it needs to be of a kind that does not disturb the average user.

(I might not even have included this function into the list of exported commands -- i.e. put a _ in front of it --, if it had been my choice.... It is very often significantly inferior to the usual method from primdec.lib concerning time consumption due to the overhead.)

wdecker: I can already not understand why I should write specialized here. This alone gives no information. Specialized modular is even more confusing. Two comments:
1) This Singular function comes with final tests (depending on the value). I will check asap what precisely these checks are doing and whether they guarantee correctness (correctness in some cases).
2) As we will more and more rely on modular GB based methods, we should develop a general philosophy of telling the truth to the user. In contrast to Magma, we should not cheat here.

With respect to this PR, let us decide on how to handle this as soon as I have checked what the tests do.

afkafkafk13: Indeed the word specialized should be deleted, whereas I would keep modular.

I strongly second the need for a general philosophy as stated in 2. .

W.r.t. this PR: What we definitely need are the two bug fixes.

src/Rings/mpoly-ideals.jl
```diff
@@ -1172,6 +1172,24 @@ function minimal_primes(I::MPolyIdeal; algorithm::Symbol = :GTZ, cache::Bool=tru
   return V
 end
 
+@doc raw"""
+    function minimal_primes_zero_dim(I::MPolyIdeal{QQPolyRingElem})
+    
+Return a vector containing the minimal associated prime ideals of `I`.
+
+Calls a specialized variant of `minimal_primes` for zero dimensional ideals
```
simonbrandhorst:
```suggestion
Calls a modular variant of `minimal_primes` for zero dimensional ideals over a polynomial ring
```

src/Rings/mpoly-ideals.jl
simonbrandhorst: Thanks for the feedback. I removed specialized and added modular. 

So far I did not export the function because I could not determine a clear use-case for it.

fieker:
My 2 pennies:
 - adding modular is pointless as no-one outside knows what this implies. Maybe call it Las Vegas/ Monte Carlo/ .. whatever to indicate that is is not guarateed correct.
 - any function in Singular that is using Singular parallelization, at this point in time, cannot be used in Oscar reliably. Thus I'd question the point of linking them in. It needs structural changes in Singular before they can be used at all. This is under discussion. Even the feasibility is not clear yet. Unfortunately.

wdecker:
Again, we need to be more precise here. Some of this is guaranteed to be correct provided the neccessary checks are on. For the case considered here, I will find out soon.Von meinem iPhone gesendetAm 05.11.2024 um 08:08 schrieb Claus Fieker ***@***.***>:﻿
My 2 pennies:

adding modular is pointless as no-one outside knows what this implies. Maybe call it Las Vegas/ Monte Carlo/ .. whatever to indicate that is is not guarateed correct.
any function in Singular that is using Singular parallelization, at this point in time, cannot be used in Oscar reliably. Thus I'd question the point of linking them in. It needs structural changes in Singular before they can be used at all. This is under discussion. Even the feasibility is not clear yet. Unfortunately.

jankoboehm:
My two pennies: We have to separate three different issues here: 

modular or not: 
modular  is just an algorithmic option to the command and there should always be a non-modular version if possible (since modular in general is not for all inputs faster).

verification: 
 If modular, one should be able to turn verfication on/off via an algorithmic option to the command or via a global "physics/science" mode, or both. As standard, verification has to be on.

resources:
Even on one core modular can be for some algorithms much faster that non-modular. We should enhance Singular (i.e. tasks.lib, resources.lib, parallel.lib) with a special case so that it does not fork if we use a single core. This does not harm even if the forking works smoothly eventually.

afkafkafk13:
> My two pennies: We have to separate three different issues here:
> 
> modular or not: modular is just an algorithmic option to the command and there should always be a non-modular version if possible (since modular in general is not for all inputs faster).

This was, what I had in mind: Be frank about this being a modular method (as opposed to the non-modular approach in primdec.lib).

> verification: If modular, one should be able to turn verfication on/off via an algorithmic option to the command or via a global "physics/science" mode, or both. As standard, verification has to be on.

Here the tests are switched on by default, so this is indeed verified (and certifies correctness according to the underlying article of Idrees, Pfister, Steidel).  But the Oscar-method in this PR currently does not provide the means to turn it off yet. Is it desired to be able to switch it off?

> resources: Even on one core modular can be for some algorithms much faster that non-modular. We should enhance Singular (i.e. tasks.lib, resources.lib, parallel.lib) with a special case so that it does not fork if we use a single core. This does not harm even if the forking works smoothly eventually.

This is a valid point in general, but not of relevance here.  The current PR is an instance where no parallelism is used at all in Singular and computation only passes through the `modular` command.
Edit: I take the last phrase back.

afkafkafk13:
Please revert to **draft** for the moment. This touches on things that need to be checked/cleaned up in Singular before having this in Oscar.

simonbrandhorst:
This should not delay the two bugfixes. I will remove the singular part from this PR and return to non-draft.

afkafkafk13:
Splitting up this PR into the urgent part to be merged asap and the one touching on Singular is fine with me.

HereAround:
Key question: Modular or not modular? We just discussed in triage, and it seems @wdecker and @ederc are looking into this.

Related question: Should the user have an option to choose which algorithm they want to run? This needs to be documented as pointed out by @afkafkafk13.

wdecker:
Key question: Yes, we will soon include modular versions for primary decomposition in general (work by Steenpass/Eder in Singular).
Related question: This has been available for a long time and is well-documented.

HereAround:
@wdecker @hannes14 and @ederc are discussing with Andreas Steenpaß on whether his old libraries for modular GB computation can be used (and if so, how). The hope is to make progress, possibly even complete this task, before Christmas.

Key point: @wdecker expects that this new functionality leads to a significant speed-up and produces verified results.

In circling back to the actual code in this PR: Part of it was merged, and by now the changes look "innocent". This should serve as a good test for the new functionality that @wdecker , @hannes14 and @ederc are working on. The idea is then to include this functionality with the PRs on this new functionality, or merge this PR once said new functionality is available in OSCAR.

PR #3708, opened 2024-05-08
Revert "Disable a test in elliptic surfaces (#3698)"

joschmitt:
With #3672 merged, the test is now commented out twice. To avoid future confusion, I revert my commit.

### DIFF
--- a/test/AlgebraicGeometry/Schemes/elliptic_surface.jl
+++ b/test/AlgebraicGeometry/Schemes/elliptic_surface.jl
@@ -63,8 +63,6 @@
   trivial_lattice(X3)
   =#
 
-  #=
-  This testset is disabled (for now), see https://github.com/oscar-system/Oscar.jl/issues/3676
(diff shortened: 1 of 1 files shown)


lgoettgens:
@HechtiDerLachs do you want the whole testfile staying disabled or just the one offending testset?

HechtiDerLachs:
I would say we keep it as close as possible to the final state after @joschmitt 's changes.

joschmitt:
> I would say we keep it as close as possible to the final state after @joschmitt 's changes.

I'm sorry, I don't understand.
I commented out the single test set 'elliptic parameter' (I believe that this is where CI got stuck) and you commented out the whole file. The changes are orthogonal as far as git is concerned, so there was no merge conflict. It doesn't really matter, but I would like to revert one of the two changes to avoid confusion. I don't care which one.

HechtiDerLachs (approved):
I think this is rather fine then. Sorry about the confusion.

PR #3709, opened 2024-05-08
Attach warning to docs of functions in experimental

### LINKED ISSUE #2011: Make the user aware of that they are using code from `experimental`

lkastner:
preview: https://docs.oscar-system.org/previews/PR3709/


experimental/Experimental.jl
```diff
@@ -57,3 +57,29 @@ for pkg in oldexppkgs
   # Load the package
   include(joinpath(expdir, pkg, "$pkg.jl"))
 end
+
+# We modify the documentation of the experimental part to attach a warning to
+# every exported function that this function is part of experimental.
+# Furthermore we give a link for users to rea up on what this entails.
```
fingolfin:
```suggestion
# Furthermore we give a link for users to read up on what this entails.
```

fingolfin (commented):
Nice!

Do we need a specific Documenter version for this to work? Then perhaps specify that in docs/Project.toml?

experimental/Experimental.jl
```diff
@@ -57,3 +57,29 @@ for pkg in oldexppkgs
...
+#
+# Note that there are functions in the docs of experimental that are not
+# exported. These then also do not get the warning attached.
+using Markdown
+warnexp = Markdown.parse(raw"""
+!!! warning "Experimental"
+    This function is part of the experimental code in Oscar. Please read
+    [here](https://docs.oscar-system.org/dev/Experimental/intro/) for more
```
lgoettgens:
```suggestion
    [here](https://docs.oscar-system.org/v1/Experimental/intro/) for more
```
Or maybe make this dependent on `Oscar.is_dev`?

lgoettgens (commented):
Nice addition!! One small comment above

Project.toml
```diff
@@ -12,6 +12,7 @@ Hecke = "3e1990a7-5d81-5526-99ce-9ba3ff248f21"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
 LazyArtifacts = "4af54fe1-eca0-43a8-85a7-787d91b784e3"
+Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
```
lgoettgens, line 15: this needs a compat of something like `"1.6"` (as it is a stdlib)

lkastner:
> Do we need a specific Documenter version for this to work? Then perhaps specify that in docs/Project.toml?

It is independent of Documenter. This is only using the Julia documentation features. And not in a good way...

fingolfin (approved):
Looks good to me

thofma:
Awesome, thanks @lkastner!

thofma (approved):

PR #3699, opened 2024-05-07
FTheoryTools: Updates to hypersurface model constructor

HereAround:
* Remove method `hypersurface_model(base::NormalToricVariety; completeness_check::Bool = true)`.
(This requires to also remove `hypersurface_model_over_projective_space(d::Int)`, `hypersurface_model_over_hirzebruch_surface(r::Int)` and `hypersurface_model_over_del_pezzo_surface(b::Int)` as well as significant changes to the CI tests of hypersurface models.)
* Overhaul constructor `hypersurface_model(auxiliary_base_vars::Vector{String}, auxiliary_base_grading::Matrix{Int64}, d::Int, fiber_ambient_space::NormalToricVariety, fiber_twist_divisor_classes::Vector{Vector{Int64}}` by providing the hypersurface equation as argument. Sanity checks added and fixed a typo in the hypersurface equation of the F11 hypersurface model.

cc @apturner @emikelsons

PR #3712, opened 2024-05-08
FTheoryTools: More improvements yet

HereAround:
~~For my convenience, this is based on https://github.com/oscar-system/Oscar.jl/pull/3699, which hopefully can be merged very soon. Then this PR here can hopefully be rebased rather easily.~~

This PR makes the following changes:
* Rename `divisors_user_must_specify` to `defining_classes`.
* Rename `divisors_parametrized_by_user_specified_divisors` to `model_sections`.
* Rename `parametrization_of_parametrized_divisors_by_Kbar_and_user_specified_divisors` to `classes_of_model_sections_in_basis_of_Kbar_and_defining_classes`.
* Disable an unused testset.
* Create sections only for those elements listed as `model_sections` in our database.
* Revoke the optional argument `generic` for the literature model constructor. (Initially introduced to keep the book chapter running. I believe to have found a better solution to achieve this.)
* Introduce `classes_of_model_sections(m::AbstractFTheoryModel)`.
* Introduce `defining_classes(m::AbstractFTheoryModel)`.
* Introduce new optional argument `defining_classes` and replace `model_sections` in most places. Leave it in a few places though, to test the book chapter code. Also, include automatic conversion of `toric_line_bundles` and `toric_divisors` into `toric_divisor_classes`.
* Fix a bug in `explicit_model_sections`.

cc @apturner @emikelsons


HereAround:
@apturner What are your thoughts on this PR?

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -122,6 +122,82 @@ function defining_section_parametrization(m::AbstractFTheoryModel)
...
+  "w"   => Divisor class on a normal toric variety
+  "a2"  => Divisor class on a normal toric variety
+  "a1"  => Divisor class on a normal toric variety
+  "a43" => Divisor class on a normal toric variety
+  "a4"  => Divisor class on a normal toric variety
+  "a32" => Divisor class on a normal toric variety
+```
+"""
+@attr Dict{String, ToricDivisorClass} function classes_of_model_sections(m::AbstractFTheoryModel)
+  @req hasfield(typeof(m), :explicit_model_sections) "explicit_model_sections not supported for this F-theory model"
```
apturner, line 156: Two questions here:
1. Should the two strings in these `@req`s start with capital letters to match the other error messages in OSCAR?
2. Do we actually need to check if `explicit_model_sections` exists in order to return classes of model sections? Certainly, taking the classes of the explicit sections is one way to determine the classes, but we could not have made the explicit sections in the first place if we didn't already know the classes. The classes are already stored as a grading matrix for all the literature models, and for model families we may not have explicit sections but we still know their classes. It seems the classes should have their own attribute, and that should be checked here.

experimental/FTheoryTools/docs/src/literature.md
```diff
@@ -13,7 +13,7 @@ su5_tate_model_over_arbitrary_3d_base()
 ```
 More generally, we support literature constructions.
 ```@docs
-literature_model(; doi::String="", arxiv_id::String="", version::String="", equation::String="", model_parameters::Dict{String,<:Any} = Dict{String,Any}(), base_space::FTheorySpace = affine_space(NormalToricVariety, 0), model_sections::Dict{String, <:Any} = Dict{String,Any}(), completeness_check::Bool = true)
+literature_model(; doi::String="", arxiv_id::String="", version::String="", equation::String="", model_parameters::Dict{String,<:Any} = Dict{String,Any}(), base_space::FTheorySpace = affine_space(NormalToricVariety, 0), model_sections::Dict{String, <:Any} = Dict{String,Any}(), defining_classes::Dict{String, <:Any} = Dict{String,Any}(), completeness_check::Bool = true)
```
apturner, line 16: I know we've already discussed this somewhat, but I want to further talk about the `model_sections` argument here. This has to be kept to agree with the book, but it now means something different than the "model_sections" appearing in the user-facing function `classes_of_model_sections`, correct? The `model_sections` argument here, I believe, corresponds to what is now called "defining classes" (without Kbar). Maybe this should be a named argument we keep for backwards compatibility but that goes undocumented?

experimental/FTheoryTools/src/LiteratureModels/constructors.jl
```diff
@@ -172,23 +161,26 @@ function literature_model(model_dict::Dict{String, Any}; model_parameters::Dict{
...
-      @req value isa ToricDivisor "Construction of literature models over concrete bases currently requires toric divisors as model sections"
-    end
 
-    # Are all model sections specified?
-    @req all(k->haskey(model_sections, k), model_dict["model_data"]["divisors_user_must_specify"]) "Not all model sections are specified"
+    # FIXME: Append model_sections to defining_classes. This might need fixing/extending in the future...
+    defining_classes_provided = merge(model_sections, defining_classes)
+    for (key, value) in defining_classes_provided
+      if (value isa ToricDivisor || value isa ToricDivisorClass || value isa ToricLineBundle) == false
+        error("Construction of literature models over concrete bases currently requires defining classes (and model sections) to be provided as toric divisor (classes) or line bundles")
```
apturner, line 169: So which "model sections" is this referring to now? I think it's referring to things like `w` in the SU(5)xU(1) model, which we used to refer to as "model sections" but are now called "defining classes", which you also mention here. Is this phrasing used for backward compatibility? I don't want to confuse people with our new versus old definitions of "model sections". Here for instance, you say the model sections must be given as divisors, divisor classes, or line bundles; but our new use of "model sections" refers to the sections themselves. We should discuss.

apturner (approved):
Thank you @HereAround! I've added some comments for things we should discuss more, but this looks great and can be merged.

PR #3715, opened 2024-05-10
Create misc files to accumulate pirated methods...

lgoettgens:
...to ease moving them downstream eventually.

This approach was suggested in the last triage meeting by (I unfortunately cannot remember).

I furthermore started moving some things that are very obviously capable of being moved downstream. There are many other things already with comments, but I would like to postpone this to another PR.


src/Misc/MoveToAbstractAlgebra.jl
```diff
@@ -0,0 +1,99 @@
...
+Return whether the form corresponding to the matrix `B` is alternating,
+i.e. `B = -transpose(B)` and `B` has zeros on the diagonal.
+Return `false` if `B` is not a square matrix.
+"""
+function is_alternating(B::MatElem)
+   n = nrows(B)
+   n==ncols(B) || return false
+
+   for i in 1:n
+      B[i,i]==0 || return false
```
fingolfin:
```suggestion
      is_zero(B[i,i]) || return false
```

fingolfin: Could also re-format the new files while at it (since theses are new files, it won't affect any open PRs negatively. Note that I am *not* suggesting to put them onto the list of files where we apply code formatting checks)

src/Misc/MoveToAbstractAlgebra.jl
```diff
@@ -0,0 +1,99 @@
...
+
+@doc raw"""
+    upper_triangular_matrix(L)
+
+Return the upper triangular matrix whose entries on and above the diagonal are the elements of `L`.
+
+An exception is thrown whenever the length of `L` is not equal to $n(n+1)/2$,
+for some integer $n$.
+"""
+function upper_triangular_matrix(L)
```
fingolfin: I know this PR just moves things around, but the code as it is definitely is not ready for broad consumption...


That said, it seems AA actually already defines it (and also `lower_triangular_matrix`) but with a much better docstring and a slightly nicer implementation.

So perhaps we can just delete it here?

src/Misc/MoveToNemo.jl
```diff
@@ -0,0 +1,11 @@
...
+# A place to accumulate code that should eventually be moved to Nemo.jl
+###############################################################################
+
+function Hecke.numerator(f::QQPolyRingElem, parent::ZZPolyRing = Hecke.Globals.Zx)
+  g = parent()
+  ccall((:fmpq_poly_get_numerator, Nemo.libflint), Cvoid, (Ref{ZZPolyRingElem}, Ref{QQPolyRingElem}), g, f)
+  return g
+end
+
+Hecke.minpoly(a::QQBarFieldElem) = minpoly(Hecke.Globals.Qx, a)
```
fingolfin, line 11: This can't be moved to Nemo as-is, it references Hecke.Globals.Qx. Similar above with `Hecke.Globals.Zx`.

... unless we also decide to move those to Nemo? Should we? @thofma @fieker

lgoettgens: Yes, that's true. I think this should be done, but we can discuss this elsewhere (in my upcoming issue/PR regarding this in Nemo e.g.) and not need to block this PR by this

docs/src/Groups/matgroup.md
```diff
@@ -79,8 +79,8 @@ pol_elementary_divisors(A::MatElem{T}) where T
 generalized_jordan_block(f::T, n::Int) where T<:PolyRingElem
 generalized_jordan_form(A::MatElem{T}; with_pol=false) where T
 matrix(A::Vector{AbstractAlgebra.Generic.FreeModuleElem{T}}) where T <: RingElem
-upper_triangular_matrix(L)
-lower_triangular_matrix(L)
+upper_triangular_matrix(L::AbstractVector{T}) where {T <: NCRingElement}
+lower_triangular_matrix(L::AbstractVector{T}) where {T <: NCRingElement}
```
fingolfin, line 83: Honestly I would just drop them from the documentation here. I am not sure why we listed them here (in the matrix group section) in the first place. 
```suggestion
```

Actually almost everything in this "Utilities for matrices" section is questionable here. Some should perhaps also be moved. Some should not be documented at all (I don't even understand the docstring for some of them, e.g. `pol_elementary_divisors`)

(This comment is not meant to be a blocker for this PR)

fingolfin (approved):
Fine by me

PR #3716, opened 2024-05-10
star_subdivision: refactor facet_signs to operate on oscar types

benlorenz:
To improve performance and avoid rare crashes:

I don't fully understand why the old code sometimes accesses already deleted `Polymake.Rational` numbers. They are created in the `F*v` multiplication but I don't see why these can be GCed and still be used for some assignment later on.

The crash usually looks like this
```
[28924] signal (11.128): Segmentation fault
in expression starting at REPL[18]:1
__gmpn_copyi_x86_64 at /home/lorenz/software/polymake/julia/julia/julia-1.10.3/bin/../lib/julia/libgmp.so.10 (unknown line)
Allocations: 540062742 (Pool: 540017641; Big: 45101); GC: 81
Segmentation fault
```
and for example appeared in the book tests (`specialized/bies-turner-string-theory-applications`), but also this one looks similar : https://github.com/oscar-system/Oscar.jl/actions/runs/9032323010/job/24820290639#step:9:336

I even got a similar crash while trying to benchmark the old and the new code. When it doesn't crash the old code runs like this:
```julia
julia> @benchmark _facet_signs(mp,new_ray)
BenchmarkTools.Trial: 8571 samples with 1 evaluation.
 Range (min … max):  278.723 μs … 731.682 ms  ┊ GC (min … max): 0.00% … 20.37%
 Time  (median):     324.794 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   591.597 μs ±  12.505 ms  ┊ GC (mean ± σ):  9.74% ±  0.48%

                           ▃█▃   ▅▂                              
  ▂▂▂▂▂▁▁▁▁▁▁▂▂▂▂▂▂▂▂▃▃▂▃▃▃███▅▄███▅▄▃▃▃▃▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  279 μs           Histogram: frequency by time          372 μs <

 Memory estimate: 89.72 KiB, allocs estimate: 9112.
```

And the new code:
```julia
julia> @benchmark _facet_signs(mo,new_ray_o)
BenchmarkTools.Trial: 10000 samples with 3 evaluations.
 Range (min … max):   8.470 μs … 65.220 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      9.949 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   10.206 μs ±  2.345 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▂▆██▇▄▂                                               
  ▂▂▂▂▃▅▆███████▇▅▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  8.47 μs         Histogram: frequency by time        16.7 μs <

 Memory estimate: 10.79 KiB, allocs estimate: 81.
```

This could fix some of the crashes in #3633.

cc: @lkastner

### DIFF
--- a/src/PolyhedralGeometry/PolyhedralFan/standard_constructions.jl
+++ b/src/PolyhedralGeometry/PolyhedralFan/standard_constructions.jl
@@ -112,7 +112,7 @@ function star_subdivision(Sigma::_FanLikeType, new_ray::AbstractVector{<:Integer
   end
   mc_old = maximal_cones(IncidenceMatrix, Sigma)
 
-  facet_normals = pm_object(Sigma).FACET_NORMALS
+  facet_normals = matrix(QQ, pm_object(Sigma).FACET_NORMALS)
   refinable_cones = _get_maximal_cones_containing_vector(Sigma, new_ray)
   @req length(refinable_cones) > 0 "$new_ray not contained in support of fan."
   new_cones = _get_refinable_facets(Sigma, new_ray, refinable_cones, facet_normals, mc_old)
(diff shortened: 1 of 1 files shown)


lkastner (approved):
Thank you. ^^

lkastner (approved):

PR #3717, opened 2024-05-11
Fix vector_space(K, polynomials)

fingolfin:
The `preimage` function this provided was buggy; it failed when the number of monomials involved exceeded the number of polynomials. The new tests catch this.

Also add a rudimentary doc string, and refactor the code to use exponent vectors instead of "monomials".

@Lax202 that should fix the case you had trouble with

### DIFF
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1348,9 +1348,9 @@ end
 
 
 @doc raw"""
-  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <:M PolyRingElem}
+  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
 
-Return a K-vector space `V` and an epimorphism from `V` onto the K-vector
+Return a `K`-vector space `V` and an epimorphism from `V` onto the `K`-vector
 space spanned by the polynomials in `polys`.
 
 Note that all polynomials must have the same parent `R`, and `K` must be equal
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1346,61 +1346,103 @@ function homogeneous_component(R::MPolyDecRing, g::IntegerUnion)
   return homogeneous_component(R, grading_group(R)([g]))
 end
 
-function vector_space(K::AbstractAlgebra.Field, e::Vector{T}; target = nothing) where {T <:MPolyRingElem}
+
+@doc raw"""
+  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <:M PolyRingElem}
+
+Return a K-vector space `V` and an epimorphism from `V` onto the K-vector
+space spanned by the polynomials in `polys`.
+
+Note that all polynomials must have the same parent `R`, and `K` must be equal
+to `base_ring(R)`. The optional keyword argument `target` can be used to
+specify `R` when `polys` is empty.
+
+```jldoctest
+julia> R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"]);
+
+julia> polys = [x + y, x - y, 2x + 3y];
+
+julia> V, VtoPoly = vector_space(QQ, polys)
+(Vector space of dimension 3 over QQ, Map: V -> R)
+
+julia> VtoPoly.(basis(V))
+3-element Vector{QQMPolyRingElem}:
+ x
+ y
+ 0
+```
+"""
+function vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
   local R
-  if length(e) == 0
+  if length(polys) == 0
     R = target
     @assert R !== nothing
   else
-    R = parent(e[1])
+    R = parent(polys[1])
+    @assert target === nothing || target === R
   end
   @assert base_ring(R) == K
-  mon = Dict{elem_type(R), Int}()
-  mon_idx = Vector{elem_type(R)}()
+  expvec = Dict{Vector{Int}, Int}()
+  expvec_idx = Vector{Vector{Int}}()
   M = sparse_matrix(K)
-  last_pos = 1
-  for i = e
+  
+  # take polynomials and turn them into sparse row vectors
+  for f in polys
     pos = Vector{Int}()
     val = Vector{elem_type(K)}()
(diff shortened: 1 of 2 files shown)


src/Rings/mpoly-graded.jl
```diff
@@ -1346,61 +1346,103 @@ function homogeneous_component(R::MPolyDecRing, g::IntegerUnion)
   return homogeneous_component(R, grading_group(R)([g]))
 end
 
-function vector_space(K::AbstractAlgebra.Field, e::Vector{T}; target = nothing) where {T <:MPolyRingElem}
+
+@doc raw"""
+  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <:M PolyRingElem}
```
joschmitt:
```suggestion
  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
```

src/Rings/mpoly-graded.jl
```diff
@@ -1346,61 +1346,103 @@ function homogeneous_component(R::MPolyDecRing, g::IntegerUnion)
...
+
+Return a K-vector space `V` and an epimorphism from `V` onto the K-vector
```
joschmitt:
```suggestion
Return a `K`-vector space `V` and an epimorphism from `V` onto the `K`-vector
```

src/Rings/mpoly-graded.jl
```diff
@@ -1346,61 +1346,103 @@ function homogeneous_component(R::MPolyDecRing, g::IntegerUnion)
...
+  # row reduce
   rref!(M)
 
+  # turn the reduced sparse rows back into polynomials
   b = Vector{elem_type(R)}()
-  for i=1:nrows(M)
-    s = zero(e[1])
-    for (k,v) = M[i]
-      s += v*mon_idx[k]
+  for i in 1:nrows(M)
```
joschmitt, line 1409: Is there a reason to not just go till `rank(M)` (which `rref!` returns)? Or is it necessary that the dimension of the vector space is the number of polynomials?

src/Rings/mpoly-graded.jl
```diff
@@ -1346,61 +1346,103 @@ function homogeneous_component(R::MPolyDecRing, g::IntegerUnion)
...
-function vector_space(K::AbstractAlgebra.Field, e::Vector{T}; target = nothing) where {T <:MPolyRingElem}
+
+@doc raw"""
+  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
+
+Return a `K`-vector space `V` and an epimorphism from `V` onto the `K`-vector
+space spanned by the polynomials in `polys`.
+
+Note that all polynomials must have the same parent `R`, and `K` must be equal
+to `base_ring(R)`. The optional keyword argument `target` can be used to
```
fingolfin, line 1357: Given this requirement on `K`, I wonder why we even have it as function argument? Perhaps we should at least make it optional?

Thoughts @thofma @fieker ?

src/Rings/mpoly-graded.jl
fingolfin: I tried to not change the semantics of this function, and if I understand you right, that change would (and then also the docstring would have to be updated).

What @Lax202 really needs the code for is to get the space of all linear combinations of the given polynomials that equal zero, i.e. the kernel of the morphismus return here (only we can't get it directly, we have to jump through some hoops). I believe the work required for that is identical in either case, so from that perspective I wouldn't mind changing to `rank(M)` here. But other code might be affected, and so while I generally support the idea, I wouldn't want to try it in this PR, as it might be opening a can of worms...

joschmitt: I have no stakes in this. I'm just wondering: For example, in the doctest the vector space is of dimension 3, but the last basis element maps to 0, so why not make it dimension 2?
That it always has been like this, is of course a good argument :)

(I can also try it out once this pull request is merged and see how much breaks.)

src/Rings/mpoly-graded.jl
joschmitt: This goes for some other `vector_space` methods as well:
```
vector_space(K::Field, e::Vector{T}; target) where T<:MPolyRingElem @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-graded.jl:1349
vector_space(K::Field, Q::MPolyQuoRing) @ Oscar ~/.julia/dev/Oscar/src/Rings/MPolyQuo.jl:1272
vector_space(kk::Field, W::Oscar.MPolyQuoLocRing{<:Field, <:FieldElem, <:MPolyRing, <:MPolyRingElem, <:Oscar.MPolyComplementOfKPointIdeal}; ordering) @ Oscar ~/.julia/dev/Oscar/src/Rings/mpolyquo-localizations.jl:2203
vector_space(kk::Field, W::Oscar.MPolyQuoLocRing; ordering) @ Oscar ~/.julia/dev/Oscar/src/Rings/mpolyquo-localizations.jl:2184
```

src/Rings/mpoly-graded.jl
fingolfin: Yeah I agree with you, and am happy to have this changed -- just please not in this PR, so it can be merged sooner rather than later :-)

joschmitt (approved):
The remaining question could also be resolved separately.

src/Rings/mpoly-graded.jl
thofma: It construct a vector space and a map to the polynomial ring. If there are no polynomials, the polynomial ring (target of the map) must be supplied.

thofma (commented):
The code makes sense to me. Not sure if this accounts for much, since I never touched this method :). But if @joschmitt is happy, I am happy.

src/Rings/mpoly-graded.jl
fingolfin: indeed, and that's why we have the `target` KW argument. But even in that scenario we don't need `K` -- after all we have this in the code: `@assert base_ring(R) == K` (where `R` is either `target` or else `parent(polys[1])`)

fingolfin:
@thofma ah I saw your name attached to the code in `tig blame` but that was a red herring from some global refactoring commits -- sorry! The true author of the code seems to be @fieker

src/Rings/mpoly-graded.jl
thofma: Ah oops, I misunderstood, you are right.

joschmitt:
I added a backport label because this constitutes a bugfix. If this gets backported, then #3727 should also be backported.

PR #3723, opened 2024-05-13
Upgrade Schemes to new experimental structure

lgoettgens:
Since we currently only have one small PR editing `experimental/Schemes/`, I took the opportunity to upgrade its source part to the new experimental structure. Moving tests and docs is a topic for the future, but will be easier as it is less disruptive.
To not create any conflicts with https://github.com/oscar-system/Oscar.jl/pull/3639, I did not move the two edited files for now. This can be done in a future PR.
Furthermore, this re-upgrades FTheoryTools again to the new structure. (cc @HechtiDerLachs)


HechtiDerLachs (approved):
Fine with me, thx for the effort.

HereAround (approved):
Thank you @lgoettgens . I like this change!

afkafkafk13 (approved):
Thank you for cleaning this up. 

(By the way: The PR currently blocking ResolutionStructure.jl and BlowupMorphism.jl is nearly ready for merging.)

PR #3726, opened 2024-05-13
[FTheoryTools] Allow tuning sections to zero

HereAround:
@emikelsons and I uncovered that one cannot currently tune sections to zero. This PR should fix this.

Important: I am *not* removing such a trivial section from the model, nor do I purge it from the parametrization of the defining sections (f,g for Weierstrass and ai for Tate) or the parametrization of Weierstrass/Tate/hypersurface equation. As of right now, I believe this to be a feature rather than a bug: You can recover the original model by setting that trivial section to a non-trivial value. If you disagree and want trivial sections purged, then this can of course be adjusted.

cc @apturner @emikelsons

### DIFF
--- a/experimental/FTheoryTools/src/HypersurfaceModels/methods.jl
+++ b/experimental/FTheoryTools/src/HypersurfaceModels/methods.jl
@@ -30,6 +30,11 @@ end
     function tune(h::HypersurfaceModel, input_sections::Dict{String, <:Any}; completeness_check::Bool = true)
 
 Tune a hypersurface model by fixing a special choice for the model sections.
+Note that it is in particular possible to set a section to zero. We anticipate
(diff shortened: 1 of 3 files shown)


HereAround:
~~On 2nd thought, I think it will be better if we remove all trivial sections. Will adjust this PR accordingly later.~~

On 3rd thought - no. Best to keep them and make sure we can reinstate trivial sections. Although, we might at some point want a function `strip_trivial_sections` or alike... But let us cross that bridge once we get to it.

HereAround:
@apturner and @emikelsons , I believe this PR is ready for review.

apturner (approved):
Thanks @HereAround, this looks good. I agree with keeping the sections around even when they have been set to zero.

PR #3727, opened 2024-05-14
Tweak `vector_space(K, polynomials)` a bit more

joschmitt:
As discussed in https://github.com/oscar-system/Oscar.jl/pull/3717#discussion_r1597447298. Let's see whether something breaks. I have the impression this function is only called at one place in the code base.

### DIFF
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1350,7 +1350,7 @@ end
 @doc raw"""
     vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
 
-Return a `K`-vector space `V` and an epimorphism from `V` onto the `K`-vector
+Return a `K`-vector space `V` and an isomorphism from `V` to the `K`-vector
 space spanned by the polynomials in `polys`.
 
 Note that all polynomials must have the same parent `R`, and `K` must be equal
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1269,7 +1269,7 @@ end
     homogeneous_component(R::MPolyDecRing, g::FinGenAbGroupElem)
 
 Given a polynomial ring `R` over a field which is graded by a free
-group of type `FinGenAbGroup`, and given an element `g` of that group,
+group, and given an element `g` of that group,
 return the homogeneous component of `R` of degree `g` as a standard
 vector space. Additionally, return the map which sends an element
 of that vector space to the corresponding monomial in `R`.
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Fine by me :-)

fingolfin:
@fieker is this OK with you (you wrote the function originally I think)

src/Rings/mpoly-graded.jl
```diff
@@ -1348,9 +1348,9 @@ end
 
 
 @doc raw"""
-  vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
+    vector_space(K::Field, polys::Vector{T}; target = nothing) where {T <: MPolyRingElem}
 
-Return a `K`-vector space `V` and an epimorphism from `V` onto the `K`-vector
+Return a `K`-vector space `V` and an isomorphism from `V` to the `K`-vector
```
afkafkafk13, line 1353: Caution: You should not change the (user-visible) behaviour of vector_space(Field, Polys) without looking at the other functions vector_space(Field, something), like vector_space(Field, MPolyQuoLocRing). 
This needs to be consistent in behaviour!!! You are only tweaking one function here.

The main application is to pass from a zero-dimensional quotient ring R/I or slice of a graded ring R_d, which of course happens to carry the structure of a finite dimensional vector space, to an object of type vector space by using a monomial basis of the vector space. However, if this family of functions is provided to the user, it must be consistent in all instances.

afkafkafk13 (changes requested):

src/Rings/mpoly-graded.jl
afkafkafk13: P.S.: I am not against the change per se.

joschmitt: I'm sorry, but I don't understand what should change here.
This is editing the doc string of the function `vector_space(::Field, ::Vector{<:MPolyRingElem})`. The other `vector_space` functions that seem related are:
```
vector_space(K::Field, Q::MPolyQuoRing)
vector_space(kk::Field, W::Oscar.MPolyQuoLocRing{<:Field, <:FieldElem, <:MPolyRing, <:MPolyRingElem, <:Oscar.MPolyComplementOfKPointIdeal}; ordering) 
vector_space(kk::Field, W::Oscar.MPolyQuoLocRing; ordering)
```
Neither of them is documented. Looking at the implementation, I assume that all of them return an isomorphic vector space and not just one that maps surjectively onto the quotient ring. If this assumption is correct, then this pull request will actually increase consistency.

afkafkafk13: Sorry, I abused "changes requested" to buy the time for checking myself that everything is consistent. I should have sent an e-mail instead.
I am now convinced that this does not break anything and does not cause anything inconsistent.

afkafkafk13 (approved):

joschmitt:
It seems I levelled up and gained the power to merge to master :)

joschmitt:
Backport requires #3717.

PR #3728, opened 2024-05-14
Fix Oscar.build()

aaruni96:
Fixes Oscar.build() again, and adds generic CPU target as default.

[skip ci] as it doesn't change anything else which the tests might check for.

An important/questionable change I am making is to set peristent_tasks to false in Aqua.jl. When true, Oscar.build() fails for some reason I don't really understand. I am not sure what the consequences to not testing for that are.

@lgoettgens I think you added the Aqua tests. Do you know?

### DIFF
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -27,11 +27,12 @@ Oscar.system("precompile.jl")
 
 sysimage=joinpath(tmp, "Oscar.$(Libdl.dlext)")
 if !("JULIA_CPU_TARGET" in keys(ENV)) || (ENV["JULIA_CPU_TARGET"] == "")
-  PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_execution_file=CO)
+  println("Building system image for generic target. Use JULIA_CPU_TARGET to change.")
+  target = "generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)"
 else
   target = ENV["JULIA_CPU_TARGET"]
-  PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_execution_file=CO; cpu_target=target)
 end
+PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_execution_file=CO; cpu_target=target)
 
 println("(re)start julia as")
 println("\tjulia -J $(sysimage)")
--- a/system/Build.jl
+++ b/system/Build.jl
@@ -28,7 +28,7 @@ Oscar.system("precompile.jl")
 sysimage=joinpath(tmp, "Oscar.$(Libdl.dlext)")
 if !("JULIA_CPU_TARGET" in keys(ENV)) || (ENV["JULIA_CPU_TARGET"] == "")
   println("Building system image for generic target. Use JULIA_CPU_TARGET to change.")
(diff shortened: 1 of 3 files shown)


thofma:
How long does that take to run typically? Maybe we can add a CI job to not break it again accidentally.

lgoettgens:
Can you put the error log when enabling the Aqua test again here? Then I can have a look tomorrow

lgoettgens:
> Can you put the error log when enabling the Aqua test again here? Then I can have a look tomorrow

Of course you would need to remove the restriction to the short tests for this again, as the Aqua tests are part of the long tests.

system/Build.jl
```diff
@@ -27,11 +27,12 @@ Oscar.system("precompile.jl")
 
 sysimage=joinpath(tmp, "Oscar.$(Libdl.dlext)")
 if !("JULIA_CPU_TARGET" in keys(ENV)) || (ENV["JULIA_CPU_TARGET"] == "")
-  PackageCompiler.create_sysimage([:Oscar], sysimage_path=sysimage, precompile_execution_file=CO)
+  println("Building system image for generic target. Use JULIA_CPU_TARGET to change.")
+  target = "generic;sandybridge,-xsaveopt,clone_all;haswell,-rdrnd,base(1)"
```
fingolfin: That will only work on x86, though, not on e.g. my MacBook. How about this instead:
```suggestion
  target = PackageCompiler.default_app_cpu_target()
```

test/Aqua.jl
```diff
@@ -6,6 +6,7 @@ using Aqua
     ambiguities=false,      # TODO: fix ambiguities
     unbound_args=false,     # TODO: fix unbound type parameters
     piracies=false,         # TODO: check the reported methods to be moved upstream
+    persistent_tasks=false  # TODO: FIXME | setting this to true breaks Oscar.build()
```
lgoettgens:
```suggestion
```
should not be an issue anymore due to `subset="short"`

lgoettgens:
As I understand the running of tests in the pkgimage process is to exercise "common usecases" and compile them into the pkgimage. Nothing of the Aqua tests exercises any code from Oscar, so I think these tests should never be called from the pkgimage creation process.

test/Aqua.jl
aaruni96: Do you still want the error message from when it fails, or do we ignore that since we don't run into it if we only run short tests ?

lgoettgens: I have seen it myself now. This could be due to https://github.com/JuliaTesting/Aqua.jl/issues/292. But as commented here https://github.com/oscar-system/Oscar.jl/pull/3728#issuecomment-2112279986, I don't think you should be running the Aqua tests at all

aaruni96:
> How long does that take to run typically?

With just the short tests, it takes ~40 minutes on my workstation (~30 minutes for the tests, ~10 minutes for the system image building).

As far as I know this cannot be sped up via parallelisation, because all the code needs to execute on the process which will also eventually create the system image.

thofma:
> > How long does that take to run typically?
> 
> With just the short tests, it takes ~40 minutes on my workstation (~30 minutes for the tests, ~10 minutes for the system image building).
> 
> As far as I know this cannot be sped up via parallelisation, because all the code needs to execute on the process which will also eventually create the system image.

Thanks for the numbers. I guess this makes a CI job an option for this, but this can be added later.

fingolfin (approved):

PR #3729, opened 2024-05-14
Remove stuff from docs that should not be there in a 1.0 version

fingolfin:
Not sure if using HTML comments like that works as intended, so don't merge this until someone reviewed the docs preview for this PR.


joschmitt:
In a locally build documentation the comments don't work.

fingolfin:
Doesn't work as I intended, see https://docs.oscar-system.org/previews/PR3729/General/other/ -- so I've now removed the sections.

joschmitt (approved):

PR #3737, opened 2024-05-15
Unify style for 'recursive' kwarg

fingolfin:


lgoettgens (approved):

fieker (commented):
What probolem is actually solved with this? Inparticular that is appears that in kwargs we do not want spaces around '=' while in other places we require them?
What is the improvement? This is going to cause merge conflicts...

ThomasBreuer (commented):
Personally, I prefer whitespace around the `=`. (And I would prefer to have whitespace in many other places.)
If we introduce the general rule to not use whitespace around `=` in assignments to keyword arguments then we should add this rule to 'Developer Documentation'/'Developer Style Guide'/'Code formatting'/Whitespace.

thofma:
I want whitespaces around ` = `.

lgoettgens:
Wanting to have whitespace at kwargs is contradicting the Julia Formatter.toml. If there is the decision to have them, please add `whitespace_in_kwargs=true` there.

I don't really care what the result of this discussion/decision is, as long as we implement it consistently, add it to the formatter config and to the dev docs.

fingolfin:
I am not very invested in this PR at all, nor in which style to use (with our without whitespace around kwarg). I used `recursive=true` simply because that's the default in our formatting config; I'd be equally happy to have that default changed and use the other style.

As to

> This is going to cause merge conflicts

No at the time I submitted this, I carefully made sure that there wouldn't be merged conflicts, and I am confident I can achieve that again (I'd just have to wait till all PRs by @ThomasBreuer are merged and coordinate with him that nothing new is in the works at that precise moment).

> What problem is actually solved with this? 

You can ask the same about any code formatting. Why do we even indent code? Clearly not indenting code is bad, and slightly less clearly there is a point where enforcing certain code formatting rules causes more harm than it helps (by discouraging contributions). But there is no single line to draw as to where code formatting should stop. In this particular case I am not trying to force anyone else to uphold specific formatting rules, 

I just would like internal consistency for `src/Groups`, `src/GAP` and their `test` counterparts (the two trivial changes in `.github` were side effects and I thought "no harm with those", but I am happy to drop them -- though there certainly are no merge conflicts in that file)

And why do I want internal consistency? For the same reason as always, to reduce my mental burden. Some people can edit code that is `formatted( like  ,  this,or =2, so= 1)` and just shrug and for others it causes actual mental anxiety and slows them down (possibly a mild form of what some people would call "neurodivergent"). Like many, many programmers I fall into the latter category.

Note that I am not proposing a general rule, not auto-formatting anything, nor requiring future contributions to follow a specific pattern, or anything like that.


So, I'd like to ask back: what is the *actual, concrete harm done by this PR* ? Is any of you planning to do a major pull requests on the GAP or groups code in OSCAR? Let me know and I'll delay this PR, or leave out the files you want to edit. I've not heard any other objections other than "I prefer another style" -- in that case I'd say Thomas and me are code owners and we decide how the code we maintain looks, and since I am happy to follow with Thomas preference, I see no problem.

PR #3739, opened 2024-05-15
Regular Solids

zkgeiselmann:
We replace the approximate Archimedean and Catalan Solids bei exact ones.


test/PolyhedralGeometry/polyhedron.jl
```diff
@@ -621,12 +621,29 @@ end
...
     @test !is_platonic_solid(j)
     @test !is_archimedean_solid(j)
     @test is_johnson_solid(j)
     @test !is_vertex_transitive(j)
   end
+
+  let sd = archimedean_solid("snub_dodecahedron"), ph = catalan_solid("pentagonal_hexecontahedron")
+    @test ph isa Polyhedron{<:EmbeddedNumFieldElem}
+    @test f_vector(sd) == reverse(f_vector(ph))
+    for i in range(1,n_facets(ph))
```
lgoettgens:
```suggestion
    for i in 1:n_facets(ph)
```

docs/oscar_references.bib
```diff
@@ -2088,3 +2096,27 @@ @Book{Zie95
   year          = {1995},
   doi           = {10.1007/978-1-4613-8431-1}
 }
+
+@Misc{con23,
```
lgoettgens: This (and the following two) bib keys are weird (but forced like this by bibtool). Maybe this is a suitable place to add a custom key, like we did for Stacks as well?

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -807,14 +809,46 @@ julia> n_facets(T)
...
+@doc raw"""
+      snub_cube()
+
+Construct the Snub Cube, an Archimedean solid.
+- [con24*1](@cite)
+
+See also [`archimedean_solid`](@ref).
+"""
+function snub_cube()
+  return load(joinpath(oscardir, "data", "ArchimedeanSolids", "snub_cube.mrdi"))
```
lgoettgens:
```suggestion
  return archimedian_solid("snub_cube")
```
I would find it much more intuitive the other way around. I.e. `archimedian_solid` checks for `archimedian_strings_from_oscar` and if it is contained directly loads the mrdi file. The named constructors then can directly call this with a string.
This would reduce some duplicate code like the folder name of the solids and the file path construction

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -866,7 +900,39 @@ julia> n_facets(T)
 12
 ```
 """
-catalan_solid(s::String) = polyhedron(Polymake.polytope.catalan_solid(s))
+function catalan_solid(s::String)
+  if s in _catalan_strings_from_oscar
+    # code used for generation of loaded files can be found at:
+    # https://github.com/dmg-lab/RegularSolidsSrc
+    return getfield(Main, Symbol(s))()
```
benlorenz: That getfield based on a string is not really good, maybe we could create a dict with name and function, e.g.:
```julia
const _archimedean_strings_from_oscar = Dict("snub_cube" => snub_cube,
                                             "snub_dodecahedron" => snub_dodecahedron)
```
And then use `haskey` and `_archimedean_strings_from_oscar[s]()` ?

(this might require a slight reordering of the functions / this const)

benlorenz (commented):
There are a few `The vertices are realized as floating point numbers.` in the docstrings that can be removed now.

benlorenz:
Please add the new functions to the docs md file (or don't export them, they are not really needed as the polytopes can be accessed via the wrapper as well).

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
lgoettgens: I forgot to send my comments. There is a different approach to this in https://github.com/oscar-system/Oscar.jl/pull/3739#discussion_r1603000534, what do you think about that? This would circumvent all of this string to function matching as it does the inverse

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
benlorenz: I think that is probably cleaner than my suggestion.

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -866,7 +900,39 @@ julia> n_facets(T)
...
+  end
+  pmp = Polymake.polytope.catalan_solid(s)
+  return polyhedron(pmp)
+end
+
+@doc raw"""
+      pentagonal_icositetrahedron()
+
+Construct the Pentagonal Icositetrahedron, a Catalan solid.
```
fingolfin:
```suggestion
Construct the pentagonal icositetrahedron, a Catalan solid.
```
Similar changes need to be made in multiple other places: Catalan, Johnson, Archimedes are names and should stay capitalized, but generic nouns and adjectives like `snub` or `Hexecontahedron` etc. should be lowercase.

docs/oscar_references.bib
fingolfin: How useful is it really to have these as "artificial" bib entries instead of a simple direct link to the wikipedia page, though?

test/PolyhedralGeometry/polyhedron.jl
```diff
@@ -621,12 +621,36 @@ end
...
+    @test ae isa Polyhedron{<:EmbeddedNumFieldElem}
+    @test !is_platonic_solid(ae)
+    @test is_archimedean_solid(ae)
+    @test !is_johnson_solid(ae)
+    @test is_vertex_transitive(ae)
+    @test Polymake.polytope.isomorphic(Oscar.pm_object(ae), Polymake.polytope.archimedean_solid(s))
+  end
+
+  for s in ("pentagonal_icositetrahedron", "pentagonal_hexecontahedron")
+    ce = eval(Symbol(s))()
```
lgoettgens: Eval should not be used unless deeply necessary (here it is not)

thofma: I.e., you can replace
```
for s in ("pentagonal_icositetrahedron", "pentagonal_hexecontahedron")
```
by
```
for s in (pentagonal_icositetrahedron, pentagonal_hexecontahedron)
```

lgoettgens:
After [`79a2cf0` (#3739)](https://github.com/oscar-system/Oscar.jl/pull/3739/commits/79a2cf05d362f2659d4894b0123d38169b90265e), most of the new bib entries are no longer referenced and can thus go.
And I think you forgot to delete the corresponding exports.

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -833,17 +868,17 @@ However, vertex-facet-incidences are correct in all cases.
     - "disdyakis_dodecahedron" : Disdyakis dodecahedron.
           Dual polytope to the truncated cuboctahedron, made of 48 scalene
           triangular facets.
-    - "pentagonal_icositetrahedron" : Pentagonal Icositetrahedron.
+    - "pentagonal_icositetrahedron" : Pentagonal icositetrahedron.
           Dual polytope to the snub cube, made of 24 irregular pentagonal facets.
           The vertices are realized as floating point numbers.
```
benlorenz:
```suggestion
```
This line, and further similar lines, can be removed now.

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -833,17 +868,17 @@ However, vertex-facet-incidences are correct in all cases.
...
-    - "pentagonal_hexecontahedron" : Pentagonal Hexecontahedron.
+    - "pentagonal_hexecontahedron" : Pentagonal hexecontahedron.
           Dual polytope to the snub dodecahedron, made of 60 irregular pentagonal
           facets. The vertices are realized as floating point numbers.
```
benlorenz:
```suggestion
          facets.
```

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -764,30 +766,30 @@ See also [`is_archimedean_solid`](@ref).
...
     - "rhombicuboctahedron" : Rhombicuboctahedron.
           Regular polytope with eight triangular and 18 square facets.
-    - "truncated_cuboctahedron" : Truncated Cuboctahedron.
+    - "truncated_cuboctahedron" : Truncated cuboctahedron.
           Regular polytope with 12 square, eight hexagonal and six octagonal
           facets.
-    - "snub_cube" : Snub Cube.
+    - "snub_cube" : Snub cube.
           Regular polytope with 32 triangular and six square facets.
           The vertices are realized as floating point numbers.
```
benlorenz:
```suggestion
```

src/PolyhedralGeometry/Polyhedron/standard_constructions.jl
```diff
@@ -764,30 +766,30 @@ See also [`is_archimedean_solid`](@ref).
...
           Regular polytope with 20 triangular, 30 square and 12 pentagonal
           facets.
-    - "truncated_icosidodecahedron" : Truncated Icosidodecahedron.
+    - "truncated_icosidodecahedron" : Truncated icosidodecahedron.
           Regular polytope with 30 square, 20 hexagonal and 12 decagonal
           facets.
-    - "snub_dodecahedron" : Snub Dodecahedron.
+    - "snub_dodecahedron" : Snub dodecahedron.
           Regular polytope with 80 triangular and 12 pentagonal facets.
           The vertices are realized as floating point numbers.
```
benlorenz:
```suggestion
```

docs/src/PolyhedralGeometry/Polyhedra/constructions.md
```diff
@@ -133,6 +133,18 @@ regular_120_cell
 regular_600_cell
 ```
 
+Like some of the Johnson solids, the following four Archimedean and Catalan
+solids are constructed using [serialized data](@ref serialization).
```
benlorenz, line 137: The serialization docs page needs a corresponding `## [Some Header](@id serialization)` for this to work.

lkastner (approved):
Thank you.

PR #3740, opened 2024-05-15
Fix citations to CLS11 in toric varieties

paemurru:
The name of the authors of `[CLS11](@cite)` was misspelled in a comment. Only in a comment, but still should be corrected. Instead of the names, I instead put `[CLS11](@cite)`.

### DIFF
--- a/experimental/Schemes/src/NormalToricVarieties/attributes.jl
+++ b/experimental/Schemes/src/NormalToricVarieties/attributes.jl
@@ -1,9 +1,9 @@
 # Compute the ideal sheaves and Weil divisors of the rays of `polyhedral_fan(X)` 
 # according to the "orbit-cone-correspondence", Theorem 3.2.6 in 
-# Cox-Little-Schenk.
+# CLS11.
 #
 # We would like to use Equation (3.2.7) from page 121 of the electronic
-# version of the CLS book. It turned out that there is a misprint in this
+# version of CLS11. It turned out that there is a misprint in this
 # formula in the online version, always leading to the unit ideal. In the
 # print version, which might not be accessible to all of us, the formula
 # has been corrected.
@@ -25,7 +25,7 @@
 #       ideal.
 #
 # What we do above is also supported by the description of divisors 
-# in CLS, Formula (4.3.2): For a ray ρ ∈ Σ(1) the divisor -D with 
+# in CLS11, Formula (4.3.2): For a ray ρ ∈ Σ(1) the divisor -D with 
 # linear system consisting of the rational functions f which vanish 
 # on D is then locally given by 
 #
--- a/experimental/Schemes/src/ToricIdealSheaves/constructors.jl
+++ b/experimental/Schemes/src/ToricIdealSheaves/constructors.jl
@@ -159,7 +159,7 @@ function ideal_sheaf(X::NormalToricVariety, tau::Cone)
   # Maybe it's a mistake in the book and we really need the dual?
   #tau_perp = polarize(tau)
   ideal_dict = IdDict{AbsAffineScheme, Ideal}()
-  # We are using Equation (3.2.7) in CLS to determine the local 
+  # We are using Equation (3.2.7) in CLS11 to determine the local 
   # form of the ideal.
   for U in affine_charts(X)
     cu = cone(U)
@@ -199,7 +199,7 @@ function Base.in(tau::Cone, Sigma::PolyhedralFan)
 end
 
 function _dehomogenize_to_chart(X::NormalToricVariety, I::MPolyIdeal, k::Int)
-  # We use CLS, Proposition 5.2.10 and Exercise 5.2.5.
+  # We use CLS11, Proposition 5.2.10 and Exercise 5.2.5.
   IM = maximal_cones(IncidenceMatrix, X)
   U = affine_charts(X)[k]
 
@@ -210,7 +210,7 @@ function _dehomogenize_to_chart(X::NormalToricVariety, I::MPolyIdeal, k::Int)
 #  phi_s_star = hom(cox_ring(X), help_ring, imgs_phi_star; check=false)
 # 
 #  # Now we need to create alpha*.
-#  # We do this with the description on p. 224 in CLS.
+#  # We do this with the description on p. 224 in CLS11.
 #  hb_U = hilbert_basis(weight_cone(U))
 #  r = matrix(ZZ, rays(U))
 #  imgs_alpha_star = [prod(x_rho[l]^Int(dot(hb_U[j], r[l, :])) for l in 1:nrows(r); init=one(help_ring)) for j in 1:length(hb_U)]


joschmitt:
There is a lot of (automatic) white space changes in here which makes you the author of these lines at first glance. Can you get your editor to not strip the trailing white spaces for this? I know its annoying...

experimental/Schemes/src/NormalToricVarieties/attributes.jl
```diff
@@ -1,9 +1,8 @@
-# Compute the ideal sheaves and Weil divisors of the rays of `polyhedral_fan(X)` 
-# according to the "orbit-cone-correspondence", Theorem 3.2.6 in 
-# Cox-Little-Schenk.
+# Compute the ideal sheaves and Weil divisors of the rays of `polyhedral_fan(X)`
+# according to the "orbit-cone-correspondence", Theorem 3.2.6 in [CLS11](@cite).
 #
 # We would like to use Equation (3.2.7) from page 121 of the electronic
-# version of the CLS book. It turned out that there is a misprint in this
+# version of [CLS11](@cite). It turned out that there is a misprint in this
```
lgoettgens:
```suggestion
# version of CLS11. It turned out that there is a misprint in this
```

experimental/Schemes/src/NormalToricVarieties/attributes.jl
```diff
@@ -1,9 +1,8 @@
-# Compute the ideal sheaves and Weil divisors of the rays of `polyhedral_fan(X)` 
-# according to the "orbit-cone-correspondence", Theorem 3.2.6 in 
-# Cox-Little-Schenk.
+# Compute the ideal sheaves and Weil divisors of the rays of `polyhedral_fan(X)`
+# according to the "orbit-cone-correspondence", Theorem 3.2.6 in [CLS11](@cite).
```
lgoettgens:
```suggestion
# according to the "orbit-cone-correspondence", Theorem 3.2.6 in CLS11.
```
the `@cite` is some documenter.jl feature to add links to the bibliography, so it is nonsense to use in comments

HereAround: I agree with @lgoettgens . Also - unless I recall incorrectly (@HechtiDerLachs may correct me on this), we did not want to make those comments part of the documentation. Rather, they are for internal use to guide developers.

paemurru:
> There is a lot of (automatic) white space changes in here which makes you the author of these lines at first glance. Can you get your editor to not strip the trailing white spaces for this? I know its annoying...

I see, good point.

This should really be added to `.editorconfig`!

experimental/Schemes/src/NormalToricVarieties/attributes.jl
paemurru: Using `[CLS11](@cite)` in comments has the benefit that it is completely unambiguous.
But if writing CLS11 is more common in comments, sure we can do it that way.

joschmitt:
> This should really be added to `.editorconfig`!

It's in there but commented out. Presumably because of problems like this.

paemurru:
I meant that perhaps it should be set to `false` and not commented out, if this is what the convention in OSCAR is

joschmitt:
Well, then make a pull request.

I assume it's less a convention and more accepting reality though.

fingolfin (approved):

fingolfin (dismissed):
Would merge if the whitespace changes were not in there

experimental/Schemes/src/ToricIdealSheaves/constructors.jl
```diff
@@ -159,7 +159,7 @@ function ideal_sheaf(X::NormalToricVariety, tau::Cone)
   # Maybe it's a mistake in the book and we really need the dual?
   #tau_perp = polarize(tau)
   ideal_dict = IdDict{AbsAffineScheme, Ideal}()
-  # We are using Equation (3.2.7) in CLS to determine the local 
+  # We are using Equation (3.2.7) in [CLS11](@cite) to determine the local
```
lkastner: It does not make sense to input `@cite` here since this is just a comment in function and nothing that is rendered in the documentation. It is for developers and developers of the toric geometry part know what CLS refers to.

experimental/Schemes/src/NormalToricVarieties/attributes.jl
```diff
@@ -17,39 +16,39 @@
...
+#       for those vⱼ which are not in τ⟂. Hence, we can use the
+#       monomials corresponding to those vⱼ as generators of the
 #       ideal.
 #
-# What we do above is also supported by the description of divisors 
-# in CLS, Formula (4.3.2): For a ray ρ ∈ Σ(1) the divisor -D with 
-# linear system consisting of the rational functions f which vanish 
-# on D is then locally given by 
+# What we do above is also supported by the description of divisors
+# in CLS, Formula (4.3.2): For a ray ρ ∈ Σ(1) the divisor -D with
```
lkastner: Why change the other `CLS` to `CLS11`, but not this?

experimental/Schemes/src/ToricIdealSheaves/constructors.jl
```diff
@@ -208,21 +208,21 @@ function _dehomogenize_to_chart(X::NormalToricVariety, I::MPolyIdeal, k::Int)
 #  help_ring, x_rho = polynomial_ring(QQ, ["x_$j" for j in indices])
 #  imgs_phi_star = [j in indices ? x_rho[findfirst(k->k==j, indices)] : one(help_ring) for j in 1:n_rays(X)]
 #  phi_s_star = hom(cox_ring(X), help_ring, imgs_phi_star; check=false)
-# 
+#
 #  # Now we need to create alpha*.
 #  # We do this with the description on p. 224 in CLS.
```
lkastner, line 213: What about this?

lkastner (changes requested):
Please get rid of the trailing white space changes, it is really hard to spot the differences otherwise.

Also these are comments in the code, only visible to developers, so `@cite` does not have any meaning here. And if already deciding to change this, why not be consistent?

experimental/Schemes/src/NormalToricVarieties/attributes.jl
lkastner: There is also another one further down.

lkastner:
> This should really be added to `.editorconfig`!

I disagree with this. We cannot formalize everything. The author of a PR needs to use `git diff` to see all changes, assess whether these changes fit with the commit message and the PR description, and then possibly use `git add -p` to only commit those things that fit. If you want to change the line ending, open a separate PR and add the commits to the `.git-blame-ignore-revs`. There will always be situations where unrelated changes happen.

lkastner:
I really don't want to talk about a typo at triage tomorrow. This has been blown out of proportion completely.

lkastner (approved):
Nvm, I fixed this myself.

PR #3745, opened 2024-05-16
Enhance booktests

### LINKED ISSUE #3673: Running booktests creates auxiliary files
After running `Pkg.test(Oscar)` locally, one gets two files `test/p.mrdi` and `test/q.mrdi` that are not ignored by git.
I would expect them to be cleaned up by the test process afterwards or at least being added to the `.gitignore`.

cc @benlorenz @lkastner

lkastner:
- Fixes #3673 by running everything in tempdir
- Freeze versions of autoinstalled packages for booktest chapters


joschmitt (approved):
I can confirm that running the book tests locally doesn't produce any (permanent) files anymore with this. Thanks!

PR #3746, opened 2024-05-16
adding missing reference

wdecker:


lgoettgens (approved):
Since bibtool is happy, I see no reason to delay this.

PR #3752, opened 2024-05-17
CI: disable cache for selfhosted runner

benlorenz:
As suggested in slack, this might help with race-conditions when the cache action modifies the shared depot.

Edit:
Errors like this: https://github.com/oscar-system/Oscar.jl/actions/runs/9114205319/job/25057562605#step:9:255
```
     Testing Running tests...
ERROR: LoadError: SystemError: opening file "/Users/aaruni/.julia/compiled/v1.10/Oscar/aJ3Pg_GQbme.ji": No such file or directory
```
In this case even outside of the extra initialization stuff from Polymake or GAP.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -76,6 +76,7 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
       - uses: julia-actions/cache@v1
+        if: runner.environment != 'self-hosted'
         with:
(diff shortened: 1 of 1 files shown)


aaruni96 (approved):
The macOS tests pass, and the changes look good to me.

fingolfin (approved):
Thank you!

PR #3757, opened 2024-05-17
trim_trailing_whitespace

paemurru:
@joschmitt noticed in https://github.com/oscar-system/Oscar.jl/pull/3740#issuecomment-2117295514 that there is a lot of trailing whitespace in OSCAR code and asked to set `remove_trailing_whitespace` to `false` in the code editor. It would be nice to agree on a convention about this and write it in the `.editorconfig` file.

While there is nothing said about this in the official Julia styleguide, the most popular style guides for the most popular programming languages advise not to have trailing whitespace. I would personally prefer if `remove_trailing_whitespace` to `true`, so this is what I put in this pull request (but we should change it to the consensus opinion once people agree).

This does not influence code diffs since we can always set git and github to ignore whitespace changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 end_of_line = lf
 indent_style = space
 insert_final_newline = true
-#trim_trailing_whitespace = true  # as long as we don't enforce this, this option causes more trouble than it helps
+trim_trailing_whitespace = true
 
 [*.jl]
 indent_size = 2


joschmitt:
I would like to emphasize that I am in favour of having `trim_trailing_whitespace` set to `true`. However, my general opinion on the formatting discussions in the past weeks is that there are far too many people working on OSCAR to enforce anything, if we want to be open to new contributions. Especially, since the contributions are from mathematicians and not professional software developers.

But I already fought (and lost) a battle regarding white space at the beginning of lines in upstream packages this week, so I leave this one to other people. :grimacing:

simonbrandhorst:
The more conventions one has to follow, the harder it is to get started. I do not see the advantage.

.editorconfig
```diff
@@ -9,7 +9,7 @@ charset = utf-8
 end_of_line = lf
 indent_style = space
 insert_final_newline = true
-#trim_trailing_whitespace = true  # as long as we don't enforce this, this option causes more trouble than it helps
+trim_trailing_whitespace = true
```
fingolfin, line 12: This is a no-go. I am the one who added the `.editorconfig` originally in 4d2a248bd2b810cfe92e9e98b9360a88d84cada2 and disabled `trim_trailing_whitespace` in eb442df2dcc970b80395dc4a8434f3c99dcdf802 -- because reality is that this kind of thing is too disruptive as long as we don't combine it with a big "remove all trailing whitespace in all files" commit -- which in itself has the potential of being very disruptive, causing open PRs to have annoying merge conflicts.

Without such a cleanup PR, people with an editor honoring this setting will keep creating PRs with lots of unrelated whitespace changes in them. You dismiss this by saying

> This does not influence code diffs since we can always set git and github to ignore whitespace changes.

but I don't think it is that simple at all. First off, not everyone even knows about that feature, and it is not on by default. Secondly, affecting code review is only one (minor) effect of these whitespace changes. Far more serious is that once again it increases the risk of creating pointless merge conflicts for others.

Therefore I think the only realistic change at this point is this one:
```suggestion
trim_trailing_whitespace = false
```


After the "Begehung", if we find a time when there are few open PRs, I'd open to revisit this and considering doing a mass "remove trailing spaces PR", if the people doing that agree to afterwards go through all open PRs and resolve their merge conflicts (well, at least those which weren't there before already)

fingolfin (changes requested):
I don't think this is acceptable in the current form, see my inline comment for details.

Issue #3749, opened 2024-05-16
Faster tropicalization of linear spaces


oskarhenriksson:
It seems like we could significantly speed up the tropicalization of a linear space given by the kernel of a matrix, by using `Polymake.tropical.matroid_fan` directly on the associated projection matrix, instead of computing the actual matroid and then applying `Polymake.tropical.linear_space`.

Specifically, I'm currently using the following home-made function (built together with @YueRen) to get the Bergman fan as a polyhedral fan (which one, in turn, could easily convert to a tropical linear space).

```
function bergman_fan_from_linear_equation_matrix(linearEquationsMatrix::QQMatrix)

    basisOfComplementTransposed = nullspace(linearEquationsMatrix)[2]
    basisOfComplement = transpose(basisOfComplementTransposed)
    projectionMatrix = basisOfComplementTransposed*inv(basisOfComplement*basisOfComplementTransposed)*basisOfComplement
    
    # Compute Bergman fan as polymake object
    B = Polymake.tropical.matroid_fan{min}(projectionMatrix)

    # Convert to OSCAR polyhedral complex
    B = PolyhedralComplex{QQFieldElem}(B)

    # Get the rays (last entry is the origin, and therefore removed)
    bergmanRays = vertices_and_rays(B)
    @assert last(bergmanRays) isa PointVector "Unexpected behavior"
    @assert iszero(last(bergmanRays)) "Unexpected behavior"
    bergmanRays = matrix(QQ,Vector{QQFieldElem}.(bergmanRays))[1:end-1,:]

    # Get the lineality space (we deprojectivize by adding back the all-one vector)
    bergmanLineality = vcat(matrix(QQ,ones(Int,1,ncols(bergmanRays))),matrix(QQ,lineality_space(B)))

    # Create indicdence matrix (last column contains the vertex index)
    bergmanIncidences = vertex_and_ray_indices(maximal_polyhedra(B))[:,1:end-1]

    return polyhedral_fan(bergmanIncidences, bergmanRays, bergmanLineality, non_redundant=true)
end
```

For comparison, we consider the following matrix:

```
[...]

PR #3759, opened 2024-05-17
TropicalGeometry: new linear_space algorithm for triv val on QQ

YueRen:
First version for issue https://github.com/oscar-system/Oscar.jl/issues/3749

We probably need a similar function for polynomials

### DIFF
--- a/docs/src/TropicalGeometry/linear_space.md
+++ b/docs/src/TropicalGeometry/linear_space.md
@@ -10,10 +10,12 @@ Objects of type `TropicalLinearSpace` need to be embedded, abstract tropical lin
 
 ## Constructors
 In addition to converting from `TropicalVariety`, objects of type `TropicalLinearSpace` can be constructed from:
-- Pluecker vectors over a tropical semiring,
-- Pluecker vectors over a field and a tropical semiring map,
-- matrices over a tropical semiring,
-- matrices over a field and a tropical semiring map.
+1. Pluecker vectors over a tropical semiring: uses a low-level implementation in `polymake`
+2. Pluecker vectors over a field and a tropical semiring map: computes coordinatewise valuation and uses constructor (1.)
+3. matrices over a tropical semiring: computes tropical minors and uses constructor (1.)
+4. matrices over a field and a tropical semiring map.
+    - if matrix over `QQ` and tropical semiring map is trivial, uses an implementation of Rincon's algorithm [Rin13](@cite) in `polymake`
+    - for general input, computes minors and uses constructor (2.)
 ```@docs
 tropical_linear_space
 ```
--- a/src/TropicalGeometry/linear_space.jl
+++ b/src/TropicalGeometry/linear_space.jl
@@ -11,6 +11,7 @@
     multiplicities::Vector{ZZRingElem}
 
     # tropical linear spaces need to be embedded
+    # Note: multiplicities needs to be passed explicitly to avoid complications with a third optional parameter (tropical semiring map)
     function TropicalLinearSpace{minOrMax,true}(Sigma::PolyhedralComplex, multiplicities::Vector{ZZRingElem}) where {minOrMax<:Union{typeof(min),typeof(max)}}
         return new{minOrMax,true}(Sigma,multiplicities)
     end
@@ -65,7 +66,7 @@ function add_missing_lineality_from_polymake(Sigma::PolyhedralComplex)
     VRmat = matrix(QQ,Vector{QQFieldElem}.(collect(VR)))
     L = vcat(matrix(QQ,lineality_space(Sigma)),
              matrix(QQ,ones(Int,1,ambient_dim(Sigma))))
-    return polyhedral_complex(IM,VRmat,rayIndices,L)
+    return polyhedral_complex(IM,VRmat,rayIndices,L; non_redundant=true)
 end
 
 
@@ -208,10 +209,25 @@ function tropical_linear_space(k::Int, n::Int, plueckerVector::Vector, nu::Union
 end
 
 
+# helper function that takes a matrix (tropical or algebraic)
+# and returns the non-zero minors and their indices
+function compute_pluecker_indices_and_vector(A::MatElem)
+    n = ncols(A)
+    k = nrows(A)
+    @req n>=k "matrix for Pluecker vector cannot have more rows than columns"
+    plueckerIndices = AbstractAlgebra.combinations(1:n,k)
+    plueckerVector = AbstractAlgebra.minors(A,k)
+    nonZeroIndices = findall(!iszero,plueckerVector)
+    plueckerIndices = plueckerIndices[nonZeroIndices]
+    plueckerVector = plueckerVector[nonZeroIndices]
+    return plueckerIndices, plueckerVector
+end
+
+
 @doc raw"""
     tropical_linear_space(A::MatElem{<:TropicalSemiringElem}; weighted_polyhedral_complex_only::Bool=false)
 
-Return a tropical linear space whose Pluecker vector are the tropical minors of `A`.  If `weighted_polyhedral_complex==true`, will not cache any extra information.
+Return a tropical linear space whose Pluecker vector are the tropical minors of `A`.  Assumes that `ncols(A)>=nrows(A)`.  If `weighted_polyhedral_complex==true`, will not cache any extra information.
 
 # Examples
 ```jldoctest
@@ -224,14 +240,10 @@ Min tropical linear space
 
 ```
 """
-function tropical_linear_space(A::Union{Matrix{T},MatElem{T}}; weighted_polyhedral_complex_only::Bool=false) where {T<:TropicalSemiringElem}
-    # convert to Oscar matrix (necessary for AbstractAlgebra.minors)
-    A isa Matrix && (A = matrix(parent(first(A)),A))
+function tropical_linear_space(A::MatElem{T}; weighted_polyhedral_complex_only::Bool=false) where {T<:TropicalSemiringElem}
+    @req ncols(A)>=nrows(A) "matrix for Pluecker vector cannot have more rows than columns"
 
-    n = max(nrows(A), ncols(A))
-    k = min(nrows(A), ncols(A))
-    plueckerIndices = AbstractAlgebra.combinations(1:n,k)
-    plueckerVector = AbstractAlgebra.minors(A, k)
+    plueckerIndices,plueckerVector = compute_pluecker_indices_and_vector(A)
     TropL = tropical_linear_space(plueckerIndices, plueckerVector,
                                   weighted_polyhedral_complex_only=weighted_polyhedral_complex_only)
     if !weighted_polyhedral_complex_only
@@ -239,7 +251,10 @@ function tropical_linear_space(A::Union{Matrix{T},MatElem{T}}; weighted_polyhedr
     end
     return TropL
 end
-
+function tropical_linear_space(A::Matrix{T}; weighted_polyhedral_complex_only::Bool=false) where {T<:TropicalSemiringElem}
+    # convert to Oscar matrix (necessary for AbstractAlgebra.minors)
+    return tropical_linear_space(matrix(parent(first(A)),A), weighted_polyhedral_complex_only=weighted_polyhedral_complex_only)
+end
 
 @doc raw"""
     tropical_linear_space(A::MatElem,nu::TropicalSemiringMap; weighted_polyhedral_complex_only::Bool=false)
@@ -261,19 +276,13 @@ Min tropical linear space
 ```
 """
 function tropical_linear_space(A::MatElem, nu::Union{Nothing,TropicalSemiringMap}=nothing; weighted_polyhedral_complex_only::Bool=false)
-    # if nu unspecified, initialize as the trivial valuation + min convention
-    isnothing(nu) && (nu=tropical_semiring_map(base_ring(A)))
-
     # compute reduced row echelon form of A
     # and remove all zero rows so that matrix is of full rank
     _,A = rref(A)
     nonzeroRowIndices = findall(!iszero,[A[i,:] for i in 1:nrows(A)])
     A = A[nonzeroRowIndices,:]
 
-    n = max(nrows(A), ncols(A))
-    k = min(nrows(A), ncols(A))
-    plueckerIndices = AbstractAlgebra.combinations(1:n,k)
-    plueckerVector = AbstractAlgebra.minors(A, k)
+    plueckerIndices,plueckerVector = compute_pluecker_indices_and_vector(A)
     TropL = tropical_linear_space(plueckerIndices, plueckerVector, nu,
                                   weighted_polyhedral_complex_only=weighted_polyhedral_complex_only)
     if !weighted_polyhedral_complex_only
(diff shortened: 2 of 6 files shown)


YueRen:
@lkastner @benlorenz I get errors in the tests because of `subdivision_of_points`.  Is there anything I have to fix in my code or can I just ignore that?

benlorenz:
> @lkastner @benlorenz I get errors in the tests because of `subdivision_of_points`. Is there anything I have to fix in my code or can I just ignore that?

The subdivision test should probably compare as sets instead of vectors, since the maximal cells are just reordered. Please try changing this test to:
```julia
@test issetequal(collect(maximal_cells(square_by_incidence)), collect(maximal_cells(square_by_weights)))
```
Or even
```julia
@test issetequal(maximal_cells(square_by_incidence), maximal_cells(square_by_weights))
```

Not sure why this happens though, how the new code might trigger these reorderings. It might also depend on the order in which all tests are run. There are some runs for 'short' where the test succeeded.
I will try looking into which test might touch some global state.

_Edit:_ the rebase and new merge with master might also make this disappear, then we can take care of this in a separate PR.

But there is another error in the booktests due to changed output where the output in the jlcon (probably `linearSpace1.jlcon`) needs to be adjusted:
```diff
  julia> R,(w,x,y,z) = QQ[\"w\",\"x\",\"y\",\"z\"];
  
  julia> nu = tropical_semiring_map(QQ);
  
  julia> I = ideal([w+x+y+z,w+2*x+5*y+11*z]);
  
  julia> TropL = tropical_linear_space(I,nu)
  Min tropical linear space
  
  julia> vertices_and_rays(TropL)
  5-element SubObjectIterator{Union{PointVector{QQFieldElem}, RayVector{QQFieldElem}}}:
+  [0, 1, 0, 0]
   [0, 0, 0, 0]
   [0, -1, -1, -1]
-  [0, 1, 0, 0]
   [0, 0, 1, 0]
   [0, 0, 0, 1]
  
  julia> maximal_polyhedra(IncidenceMatrix,TropL)
  4×5 IncidenceMatrix
  [1, 2]
- [1, 3]
- [1, 4]
- [1, 5]
+ [2, 3]
+ [2, 4]
+ [2, 5]
```

YueRen:
The pull request should be ready for review.  The code coverage goes down a bit, but currently it is impossible to call that portion of the code without removing attributes which I don't think is possible.

Is there a way for @oskarhenriksson to review the pull request?  The pull request fixes an issue of his.

thofma:
I think anyone with an account can submit a review in the "Files changed" window.

benlorenz:
The book test fails since it still needs to be adjusted for the re-ordered output.

YueRen:
@oskarhenriksson I have updated the tests and made sure the that tropical semiring map is an optional parameter (default is min-convention).  If you have the time (and once most tests have passed successfully), please review the changes!

oskarhenriksson:
@YueRen Thanks a lot! This seems to do what I want, and it seems like it's faster than the original `tropical_linear_space`.

oskarhenriksson:
A small concern though:

I just did some quick experiments, and it seems like a fairly major bottleneck in the new function is the computation of `n_maximal_polyhedra(Sigma)` in [line 293](https://github.com/oscar-system/Oscar.jl/blob/b0350459c34145173447ac8baf63458402295dcc/src/TropicalGeometry/linear_space.jl#L293C39-L293C65). 

For instance, if we take `A` to be a random 5-by-20 matrix with integer entries, the computation of `PolyhedralComplex{QQFieldElem}(Polymake.tropical.matroid_fan{convention(nu)}(A))` takes 4 seconds, whereas the computation of `n_maximal_polyhedra(Sigma)` takes 29 seconds. (For comparison, the whole computation with the old `tropical_linear_space` took 100 seconds.)

It feels like we already should have this information in the Polymake output from `matroid_fan`, so perhaps there could be a faster way to compute it?

oskarhenriksson:
Perhaps one can consider the following instead (to avoid the costly computation of `n_maximal_polyhedra(Sigma)`):

```julia
 # Compute Bergman fan in polymake and convert to polyhedral complex
 B = Polymake.tropical.matroid_fan{convention(nu)}(A)
 Sigma = PolyhedralComplex{QQFieldElem}(B)
 # Add the missing all ones vector to the lineality
 Sigma = Oscar.add_missing_lineality_from_polymake(Sigma)
 multiplicities = ones(ZZRingElem, B.N_MAXIMAL_POLYTOPES)
```

src/TropicalGeometry/linear_space.jl
```diff
@@ -283,14 +284,27 @@ function tropical_linear_space(A::MatElem, nu::Union{Nothing,TropicalSemiringMap
     return TropL
 end
 
+# Special implementation for trivial valuation over the rationals using Polymake.tropical.matroid_fan for better performance
+function tropical_linear_space(A::QQMatrix, nu::TropicalSemiringMap{QQField, Nothing, minOrMax}=tropical_semiring_map(QQ); weighted_polyhedral_complex_only::Bool=false) where {minOrMax<:Union{typeof(min),typeof(max)}}
+    # Compute Bergman fan in polymake and convert to polyhedral complex
+    Sigma = PolyhedralComplex{QQFieldElem}(Polymake.tropical.matroid_fan{convention(nu)}(A))
+    # Add the missing all ones vector to the lineality
+    Sigma = add_missing_lineality_from_polymake(Sigma)
+    multiplicities = ones(ZZRingElem, n_maximal_polyhedra(Sigma))
```
oskarhenriksson: Perhaps one can consider the following instead (to avoid the costly computation of `n_maximal_polyhedra(Sigma)`). With this, the whole computation for a random 5-by-20 matrix `A` goes down from 30 to 4 seconds. (I'm not sure why this info isn't automatically carried over when we broadcast to `PolyhedralComplex{QQFieldElem}`?)

```suggestion
    # Compute Bergman fan in polymake and convert to polyhedral complex
    B = Polymake.tropical.matroid_fan{convention(nu)}(A)
    Sigma = PolyhedralComplex{QQFieldElem}(B)
    # Add the missing all ones vector to the lineality
    Sigma = add_missing_lineality_from_polymake(Sigma)
    multiplicities = ones(ZZRingElem, B.N_MAXIMAL_POLYTOPES)
```

docs/src/TropicalGeometry/linear_space.md
```diff
@@ -10,10 +10,12 @@ Objects of type `TropicalLinearSpace` need to be embedded, abstract tropical lin
...
 In addition to converting from `TropicalVariety`, objects of type `TropicalLinearSpace` can be constructed from:
-- Pluecker vectors over a tropical semiring,
-- Pluecker vectors over a field and a tropical semiring map,
-- matrices over a tropical semiring,
-- matrices over a field and a tropical semiring map.
+1. Pluecker vectors over a tropical semiring: uses a low-level implementation in `polymake`
+2. Pluecker vectors over a field and a tropical semiring map: computes coordinatewise valuation and uses constructor (1.)
+3. matrices over a tropical semiring: computes tropical minors and uses constructor (1.)
+4. matrices over a field and a tropical semiring map.
+    - if matrix over `QQ` and tropical semiring map is trivial, uses a fast specialized implementation in `polymake`
```
oskarhenriksson: Perhaps it could make sense to [cite](https://github.com/oscar-system/Oscar.jl/blob/master/docs/oscar_references.bib) the [paper by Felipe Rincón](https://www.sciencedirect.com/science/article/pii/S0747717112001174?via%3Dihub) on the underlying algorithm here.

src/TropicalGeometry/linear_space.jl
benlorenz: The extra time for `n_maximal_polyhedra` is probably redundancy elimination after adding the extra lineality. And by not doing it here it will just delay it to the first time the user triggers some other computation with the object.

If it is guaranteed that this lineality does not change the combinatorics (and is not implicit in the other lineality rows) then adding the internal `non_redundant` flag for the constructor in `add_missing_lineality_from_polymake` might be a better approach as polymake will then directly store the properties as irredundant description.

docs/src/TropicalGeometry/linear_space.md
YueRen: Good point, it's added in the version I just pushed now.

src/TropicalGeometry/linear_space.jl
YueRen: Done, thanks for the tip!  The `non_redundant` flat will come very useful for me.

joschmitt:
> The code coverage goes down a bit, but currently it is impossible to call that portion of the code without removing attributes which I don't think is possible.

Stupid question: as I understand it, you say that it is impossible to create an example that calls those parts of the code. So why have this code? Do we expect it to be "callable" in the near future?

src/TropicalGeometry/linear_space.jl
```diff
@@ -335,13 +349,72 @@ end
...
+
 @doc raw"""
     pluecker_indices(TropL::TropicalLinearSpace)
 
 Return the Pluecker indices used to construct `TropL`.  Raises an error, if it is not cached.
 """
 function pluecker_indices(TropL::TropicalLinearSpace)
-    @req has_attribute(TropL,:pluecker_indices) "no pluecker indices cached"
+    ensure_pluecker_data!(TropL)
     return get_attribute(TropL,:pluecker_indices)
```
fingolfin: But `ensure_pluecker_data!` does not always set `pluecker_indices`, so shouldn't the `@req` stay, or a similar new test be added?


Looking at the code, how about something like this instead (ideally with a more specific return type for `algebraic_matrix`, but since you effectively use `Any` right now, I put that there):
```julia
@attr Vector{Vector{In64}} function pluecker_indices(TropL::TropicalLinearSpace)
    if has_attribute(TropL,:tropical_matrix)
        A = get_attribute(TropL,:tropical_matrix)
    else
        A = algebraic_matrix(TropL)
    end
    n = ncols(A)
    k = rank(A)
    return AbstractAlgebra.combinations(1:n,k)
end

@attr Any function algebraic_matrix(TropL::TropicalLinearSpace)
    I = get_attribute(TropL,:algebraic_ideal)
    @req I !== nothing "insufficient data to compute algebraic matrix"
    A = matrix([[coeff(g,xi) for xi in gens(base_ring(I))] for g in gens(I)])
    return A
end
```

I think you can rewrite the other function in a similar vein -- and then even dealing with the case I asked about elsewhere ("what if `tropical_matrix ` and `tropical_semiring_map` are known but none of `algebraic_pluecker_vector`, `algebraic_matrix`, `algebraic_ideal`?") becomes relatively simple.

src/TropicalGeometry/linear_space.jl
```diff
@@ -335,13 +349,72 @@ end
...
+        end
+    else
+        # if not, we have to compute the tropical pluecker vector from the tropical data
+        if has_attribute(TropL,:tropical_matrix)
+            A = get_attribute(TropL,:tropical_matrix)
+            n = ncols(A)
+            k = rank(A)
+            plueckerIndices = AbstractAlgebra.combinations(1:n,k)
+            tropicalPlueckerVector = AbstractAlgebra.minors(A, k)
+            println("tropical matrix -> tropical pluecker")
```
fingolfin:
```suggestion
```

src/TropicalGeometry/linear_space.jl
```diff
@@ -335,13 +349,72 @@ end
...
+        # if yes, compute the algebraic matrix and rerun
+        if has_attribute(TropL,:algebraic_ideal)
+            I = get_attribute(TropL,:algebraic_ideal)
+            A = matrix([[coeff(g,xi) for xi in gens(base_ring(I))] for g in gens(I)])
+            set_attribute!(TropL,:algebraic_matrix,A)
+            return ensure_pluecker_data!(TropL)
+        end
+    else
+        # if not, we have to compute the tropical pluecker vector from the tropical data
+        if has_attribute(TropL,:tropical_matrix)
```
fingolfin: I find this function a bit difficult to untangle and I am not sure it is is always used right (see my other comments). In any case, isn't it true that if `tropical_semiring_map` is present, but none of `algebraic_pluecker_vector`, `algebraic_matrix`, `algebraic_ideal` are, then this method will fail -- even if `tropical_matrix` *is* available and thus would allow for the requested computation?

YueRen:
> > The code coverage goes down a bit, but currently it is impossible to call that portion of the code without removing attributes which I don't think is possible.
> 
> Stupid question: as I understand it, you say that it is impossible to create an example that calls those parts of the code. So why have this code? Do we expect it to be "callable" in the near future?

It's so that I do not have to remember to implement that portion when it becomes callable (e.g., when removal of attributes is implemented, or a new constructor of tropical linear space makes that combination of attributes possible).

src/TropicalGeometry/linear_space.jl
YueRen: Ah, you got me there.  That is a very good point, I will take a look at it later!

joschmitt:
> > > The code coverage goes down a bit, but currently it is impossible to call that portion of the code without removing attributes which I don't think is possible.
> > 
> > 
> > Stupid question: as I understand it, you say that it is impossible to create an example that calls those parts of the code. So why have this code? Do we expect it to be "callable" in the near future?
> 
> It's so that I do not have to remember to implement that portion when it becomes callable (e.g., when removal of attributes is implemented, or a new constructor of tropical linear space makes that combination of attributes possible).

Well, it is technically possible to remove attributes:
```
julia> R,(w,x,y,z) = QQ["w","x","y","z"];

julia> I = ideal([w+x+y+z]);

julia> TropL = tropical_linear_space(I);

julia> has_attribute(TropL, :algebraic_pluecker_vector)
true

julia> delete!(TropL.__attrs, :algebraic_pluecker_vector);

julia> has_attribute(TropL, :algebraic_pluecker_vector)
false
```
I wouldn't endorse this kind of stuff, but I think it's better than having untested code.

YueRen:
> > > > The code coverage goes down a bit, but currently it is impossible to call that portion of the code without removing attributes which I don't think is possible.
> > > 
> > > 
> > > Stupid question: as I understand it, you say that it is impossible to create an example that calls those parts of the code. So why have this code? Do we expect it to be "callable" in the near future?
> > 
> > 
> > It's so that I do not have to remember to implement that portion when it becomes callable (e.g., when removal of attributes is implemented, or a new constructor of tropical linear space makes that combination of attributes possible).
> 
> Well, it is technically possible to remove attributes:
> 
> ```
> julia> R,(w,x,y,z) = QQ["w","x","y","z"];
> 
> julia> I = ideal([w+x+y+z]);
> 
> julia> TropL = tropical_linear_space(I);
> 
> julia> has_attribute(TropL, :algebraic_pluecker_vector)
> true
> 
> julia> delete!(TropL.__attrs, :algebraic_pluecker_vector);
> 
> julia> has_attribute(TropL, :algebraic_pluecker_vector)
> false
> ```
> 
> I wouldn't endorse this kind of stuff, but I think it's better than having untested code.

Good point, I'll add it to the test later then.

src/TropicalGeometry/linear_space.jl
YueRen: Good catch, that is very true and definitely shouldn't be the case!

YueRen: Removing the `else` should do the trick, because if the data has been computed, we would have exited the function already.

src/TropicalGeometry/linear_space.jl
YueRen: @fingolfin I like your suggestion, I'll make the necessary changes now!

YueRen: One question: what is the `@attr` for in `@attr Any function algebraic_matrix(TropL::TropicalLinearSpace)`?

joschmitt: `@attr` automatically enables attribute storage for the function. So the result is stored and looked up in the attribute `:algebraic_matrix`. See also the doc string for `@attr`.

YueRen:
I've squashed all commits and rebased to latest master.  Unless there are any objections, I think the pull request is ready to go once all tests are through.

src/TropicalGeometry/linear_space.jl
```diff
@@ -208,10 +209,25 @@ function tropical_linear_space(k::Int, n::Int, plueckerVector::Vector, nu::Union
...
+    plueckerVector = plueckerVector[nonZeroIndices]
+    return plueckerIndices, plueckerVector
+end
+
+
 @doc raw"""
     tropical_linear_space(A::MatElem{<:TropicalSemiringElem}; weighted_polyhedral_complex_only::Bool=false)
 
-Return a tropical linear space whose Pluecker vector are the tropical minors of `A`.  If `weighted_polyhedral_complex==true`, will not cache any extra information.
+Return a tropical linear space whose Pluecker vector are the tropical minors of `A`.  Assumes that `ncols(A)>=nrows(A)`.  If `weighted_polyhedral_complex==true`, will not cache any extra information.
```
lkastner, line 230: You could `@req ncols(A)>=nrows(A)`, it is probably not very expensive.

src/TropicalGeometry/linear_space.jl
```diff
@@ -260,20 +273,17 @@ Min tropical linear space
 
 ```
 """
-function tropical_linear_space(A::MatElem, nu::Union{Nothing,TropicalSemiringMap}=nothing; weighted_polyhedral_complex_only::Bool=false)
-    # if nu unspecified, initialize as the trivial valuation + min convention
-    isnothing(nu) && (nu=tropical_semiring_map(base_ring(A)))
+function tropical_linear_space(A::Union{MatElem,Matrix}, nu::Union{Nothing,TropicalSemiringMap}=nothing; weighted_polyhedral_complex_only::Bool=false)
+    # convert to Oscar matrix (necessary for AbstractAlgebra.minors)
+    A isa Matrix && (A = matrix(parent(first(A)),A))
```
lkastner: Optionally:
```suggestion
tropical_linear_space(A::AbstractMatrix, nu::Union{Nothing,TropicalSemiringMap}=nothing; weighted_polyhedral_complex_only::Bool=false) =
    tropical_linear_space(matrix(parent(first(A)),A), nu; weighted_polyhedral_complex_only)
function tropical_linear_space(A::MatElem, nu::Union{Nothing,TropicalSemiringMap}=nothing; weighted_polyhedral_complex_only::Bool=false)
```
Just make sure to attach the docs to the right function, so probably the top function needs to be moved below.

lkastner (approved):

src/TropicalGeometry/linear_space.jl
YueRen: I like your version better, I'll change it!

src/TropicalGeometry/linear_space.jl
YueRen: Will be done!

YueRen:
Everything should be good to go.  I also double-checked that the docu for the linear spaces looks correct, after incorporating the changes suggested by @lkastner.

PR #3762, opened 2024-05-20
`QuadFormAndIsom`: Serialization infinite order isometries and a minor fix

StevellM:
Up to now, since there is no serialization of `PosInf`, we could not store the order for isometries of infinite order. Since, in the finite or trivial case, this order `n` is at least -1, for the storing I have opted for `n = -2`  in the case of infinite order. 

Eventually, even if we implement the serialization of infinity, I guess it could stay as is: there is no possible conflict, and it is something that humans are not suppose to read.

@antonydellavecchia does it seem good ? Any objection ? 

The extra minor fix is trivial: `OqAinOD`, by construction, sends the identity to the identity (similarly for the other one).

### DIFF
--- a/src/Serialization/Upgrades/1.1.0.jl
+++ b/src/Serialization/Upgrades/1.1.0.jl
@@ -16,6 +16,15 @@ push!(upgrade_scripts_set, UpgradeScript(
       if !haskey(dict[:data], :basis)
         upgraded_dict[:data][:basis] = dict[:data][:lattice][:data][:basis]
       end
+
+      # In the past, we saved only finite order isometries, but the order
+      # was not typed. Now we need to reference the type otherwise we cannot
+      # read old files.
+      if !(dict[:data][:ambient_space][:data][:order] isa Dict)
+        n = Base.parse(Int, dict[:data][:ambient_space][:data][:order])
+        upgraded_dict[:data][:ambient_space][:data][:order] =
+        Dict{Symbol, Any}(:_type => "Base.Int", :data => "$n")
+      end
     end
 
     return upgraded_dict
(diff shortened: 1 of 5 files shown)


antonydellavecchia:
I've added serialization for PosInf, It just seems a bit cleaner, the tests still pass locally with my changes

antonydellavecchia (approved):

StevellM:
Ah I did not know, great! Well I have not added a test with an infinite order isometry.. I could add one to check it properly works, but a bit later today or tomorrow maybe.

antonydellavecchia:
> Ah I did not know, great! Well I have not added a test with an infinite order isometry.. I could add one to check it properly works, but a bit later today or tomorrow maybe.

I just added to this branch

StevellM:
There seem to be a problem, with this fix I cannot reload old serialized file

```
unsupported type '2' for decoding
```
Here is the file in question:
[02-1.txt](https://github.com/oscar-system/Oscar.jl/files/15401363/02-1.txt)

StevellM:
Ok I guess I see the problem: there were saved at a time when "order" was not saved as a "typed object". Now we load it as a typed object so everything breaks... Is it possible to detect it prior to the changes and make sure it reads even such old files (like an upgrade script) ?

src/Serialization/Upgrades/1.1.0.jl
```diff
@@ -16,6 +16,15 @@ push!(upgrade_scripts_set, UpgradeScript(
...
         upgraded_dict[:data][:basis] = dict[:data][:lattice][:data][:basis]
       end
+
+      # In the past, we saved only finite order isometries, but the order
+      # was not typed. Now we need to reference the type otherwise we cannot
+      # read old files.
+      if !(dict[:data][:ambient_space][:data][:order] isa Dict)
+        n = Base.parse(Int, dict[:data][:ambient_space][:data][:order])
+        upgraded_dict[:data][:ambient_space][:data][:order] =
+        Dict{Symbol, Any}(:_type => "Base.Int", :data => "$n")
```
antonydellavecchia, line 26: looks good to me, these lines aren't covered but thats because you dont test an upgrade.
If they are working for you though i'd say this is fine to merge

StevellM:
Is there anything more needed before merging ? Everything seems fine on both sides.

thofma:
Are you happy with this @antonydellavecchia?

antonydellavecchia:
This looks good, and can be merged

PR #3763, opened 2024-05-20
bug: fix galois_group regression

thofma:
- `pmaximal_overorder` requires a prime
- since we don't want to factor, we compute a tame overorder

@fieker: I think the code is still correct, but maybe you want to do trial division + `pmaximal_overorder` call?

CC: @micjoswig (should fix the Ehrhart polynomial problems).


fieker (approved):
We need to work on the interface to the "TameOverorderBL"...

PR #3766, opened 2024-05-21
No Reynolds Operator - Invariant theory of linearly reductive groups

Lax202:
@fingolfin 
We can, with this code:
1. Make linearly reductive groups with just an ideal as input
2. Make an action (representation) of this linearly reductive group with a matrix 
3. Compute fundamental invariants of the invariant ring with only the above information, and no Reynolds operator.

I have also added a simple test from a well known SL(2) result.


wdecker:
@Lax202 Thanks for doing this! You have two functions `no_reynolds_basis` and `inv_gens_no_reynolds`. How about renaming the first one to `basis_no_reynolds`?

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -412,17 +440,27 @@ function Base.show(io::IO, R::RedGroupInvarRing)
...
     n = ncols(rep_mat)
-    m = G.group[2]
-    mixed_ring_xy, x, y, zz = polynomial_ring(G.field, "x"=>1:n, "y"=>1:n, "zz"=>1:m^2)
-    ztozz = hom(R,mixed_ring_xy, gens(mixed_ring_xy)[(2*n)+1:(2*n)+(m^2)])
+    if isdefined(G, :group)
+        m = (G.group[2])^2
+    else
+        m = ngens(base_ring(group_ideal(G)))
+    end
+    mixed_ring_xy, x, y, zz = polynomial_ring(G.field, "x"=>1:n, "y"=>1:n, "zz"=>1:m)
```
fingolfin: All internal polynomial rings like this should be created with `cached=false`.

Also we can use symbols instead of strings for the variable names to avoid some allocations.
```suggestion
    mixed_ring_xy, x, y, zz = polynomial_ring(G.field, :x=>1:n, :y=>1:n, :zz=>1:m; cached=false)
```

Ideally let's make those changes everywhere :-)

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -412,17 +440,27 @@ function Base.show(io::IO, R::RedGroupInvarRing)
...
+    ztozz = hom(R,mixed_ring_xy, gens(mixed_ring_xy)[(2*n)+1:(2*n)+(m)])
```
fingolfin: Does this also work?
```suggestion
    ztozz = hom(R, mixed_ring_xy, zz)
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -195,12 +212,18 @@ end
 
 function Base.show(io::IO, R::RepresentationLinearlyReductiveGroup)
     io = pretty(io)
-    @assert group(group(R))[1] == :SL
-    println(io, "Representation of ", group(group(R))[1], group(group(R))[2])
-    if R.sym_deg[1]
-        print(io, Indent(), "on symmetric forms of degree ", R.sym_deg[2])
-        print(io, Dedent())
+    if isdefined(group(R), :group) && group(group(R))[1] == :SL
```
fingolfin: `group(group(R))` looks really weird. I had to look up what the "out" group does. 

I suggest to replace that `group` method by something like this:
```
group_type(R::RepresentationLinearlyReductiveGroup) = isdefined(R, :group) ? R.group[1] : nothing
group_dim(R::RepresentationLinearlyReductiveGroup) = R.group[2]
```
then you can change the code here to something like
```suggestion
    if group_type(group(R))[1] == :SL
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -195,12 +212,18 @@ end
...
+        if R.sym_deg[1]
+            print(io, Indent(), "on symmetric forms of degree ", R.sym_deg[2])
+            print(io, Dedent())
+        else
+            println(io, Indent(), "with representation matrix")
+            show(io, R.rep_mat)
+            print(io, Dedent())
+        end
     else
+        println(io, "Representation of Linearly Reductive Group")
```
fingolfin:
```suggestion
        println(io, "Representation of linearly reductive group")
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -412,17 +440,27 @@ function Base.show(io::IO, R::RedGroupInvarRing)
...
+    if isdefined(G, :group)
+        m = (G.group[2])^2
+    else
+        m = ngens(base_ring(group_ideal(G)))
+    end
+    mixed_ring_xy, x, y, zz = polynomial_ring(G.field, "x"=>1:n, "y"=>1:n, "zz"=>1:m)
+    ztozz = hom(R,mixed_ring_xy, gens(mixed_ring_xy)[(2*n)+1:(2*n)+(m)])
     genss = [ztozz(f) for f in gens(G.group_ideal)]
     #rep_mat in the new ring
     new_rep_mat = matrix(mixed_ring_xy,n,n,[ztozz(rep_mat[i,j]) for i in 1:n, j in 1:n])
```
fingolfin:
```suggestion
    new_rep_mat = matrix(mixed_ring_xy,[ztozz(rep_mat[i,j]) for i in 1:n, j in 1:n])
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -675,3 +721,82 @@ function affine_algebra(R::RedGroupInvarRing)
...
+    g = gens(I)
+    V = Vector{Int}()
+    n = ncols(mat_)
+    #make a vector of all the degrees of generators of the null cone ideal I
+    for gen in g
+        d = total_degree(gen)
+        if !(d in V)
+            push!(V, Int(d))
+        end
+    end
```
fingolfin: Should `V` be sorted? 

Note that `d in V` can be expensive if `g` is larger (but I guess this is mostly irrelevant here...). Anyway, this *may* be more efficient:
```suggestion
    n = ncols(mat_)
    #make a vector of all the degrees of generators of the null cone ideal I
    V = map(total_degree,g)
    sort!(V)
    unique!(V)
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -675,3 +721,82 @@ function affine_algebra(R::RedGroupInvarRing)
...
+    #change base ring of mat_
+    R = base_ring(mat_)
+    mixed_ring_xy = base_ring(I)
+    mapp = hom(R, mixed_ring_xy, gens(mixed_ring_xy))
+    new_rep_mat = matrix(mixed_ring_xy,n,n,[mapp(mat_[i,j]) for i in 1:n, j in 1:n])
```
fingolfin:
```suggestion
    new_rep_mat = matrix(mixed_ring_xy, [mapp(mat_[i,j]) for i in 1:n, j in 1:n])
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -437,22 +479,30 @@ function proj_of_image_ideal(G::LinearlyReductiveGroup, rep_mat::MatElem)
...
     n = ncols(rep_mat)
-    m = G.group[2]
+    if isdefined(G, :group)
+        m = (group_dim(G))^2
+    else
+        m = ngens(base_ring(group_ideal(G)))
+    end
     #use parallelised groebner bases here. This is the bottleneck!
-    return eliminate(W[1], gens(mixed_ring_xy)[(2*n)+1:(2*n)+(m^2)]), W[2]
+    return eliminate(W[1], gens(mixed_ring_xy)[(2*n)+1:(2*n)+(m)]), W[2]
```
fingolfin: Does this also work? In that case you can get rid of the computation of `m` above. Also the same then could be done in `generators` a few lines below.
```suggestion
    return eliminate(W[1], gens(mixed_ring_xy)[(2*n)+1:end]), W[2]
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -412,20 +444,30 @@ function Base.show(io::IO, R::RedGroupInvarRing)
...
+        m = (group_dim(G))^2
+    else
+        m = ngens(base_ring(group_ideal(G)))
+    end
+    mixed_ring_xy, x, y, zz = polynomial_ring(G.field, :x=>1:n, :y=>1:n, :zz=>1:m; cached=false)
+    ztozz = hom(R, mixed_ring_xy, zz)
     genss = [ztozz(f) for f in gens(G.group_ideal)]
     #rep_mat in the new ring
-    new_rep_mat = matrix(mixed_ring_xy,n,n,[ztozz(rep_mat[i,j]) for i in 1:n, j in 1:n])
+    new_rep_mat = matrix(mixed_ring_xy,[ztozz(rep_mat[i,j]) for i in 1:n, j in 1:n])
```
fingolfin: Using `map_entries` is more efficient. Should be used here and in other similar places
```suggestion
    new_rep_mat = map_entries(ztozz, rep_mat)
```

experimental/InvariantTheory/test/runtests.jl
```diff
@@ -63,4 +63,40 @@
...
+    @test length(FF) == 2
+    M = representation_matrix(r)
+    ringg = parent(M[1,1])
+    z = gens(ringg)
+    f = z[1]*z[4] - z[2]*z[3] - 1
+    G = linearly_reductive_group(ideal([f]))
+    R = representation_reductive_group(G,M)
+    RG = invariant_ring(R)
+    F = fundamental_invariants(RG)
+    @test length(F) == 27
```
Lax202, line 101: - When the user constructs the group as SL(n), they are automatically using the Reynold's operator from Cayley's omega process to compute fundamental invariants. 
- When the user creates the group using only the defining ideal, there is no known Reynolds operator. Therefore the algorithm which does not require a Reynold's operator is used to compute fundamental invariants (or rather, just generators of the invariant ring?). 

This example shows the difference between the two algorithms. SL(2) over degree 4 polynomials has 2 fundamental invariants when using Cayley's omega process. Without a Reynolds operator, we get 27 generators.

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -33,7 +35,7 @@ mutable struct LinearlyReductiveGroup
     function LinearlyReductiveGroup(sym::Symbol, m::Int, fld::Field) #have not decided the representation yet
         #check char(fld)
         @assert sym == :SL && characteristic(fld) == 0
-        R, _ = polynomial_ring(fld, :z => (1:m,1:m))
+        R, _ = polynomial_ring(fld, :z => (1:m,1:m);cached=false)
```
fingolfin:
```suggestion
        R, _ = polynomial_ring(fld, :z => (1:m,1:m); cached=false)
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -412,20 +445,30 @@ function Base.show(io::IO, R::RedGroupInvarRing)
...
-    print(io, Indent(),  "under group action of ", R.group.group[1], R.group.group[2])
-    print(io, Dedent())
+    if isdefined(group(R), :group)
+        print(io, Indent(),  "under group action of ", group_type(group(R)), group_dim(group(R)))
+        print(io, Dedent())
+    else
+        println(io, Indent(),  "under group action of linearly reductive group defined by")
+        print(io, group_ideal(group(R)))
+        print(io, Dedent())
+    end
```
fingolfin:
```suggestion
    print(io, Indent(), "under group action of ")
    if isdefined(group(R), :group)
        print(io, group_type(group(R)), group_dim(group(R)))
    else
        println(io, "linearly reductive group defined by")
        print(io, group_ideal(group(R)))
    end
    print(io, Dedent())
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -437,22 +480,24 @@ function proj_of_image_ideal(G::LinearlyReductiveGroup, rep_mat::MatElem)
...
+        m = (group_dim(G))^2
+    else
+        m = ngens(base_ring(group_ideal(G)))
+    end
     gbasis = gens(X) 
     length(gbasis) == 0 && return gbasis
     mixed_ring_xy = parent(gbasis[1])
     #evaluate at gens(mixed_ring_xy)[n+1:2*n] = 0
-    V = vcat(gens(mixed_ring_xy)[1:n], [0 for i in 1:n], gens(mixed_ring_xy)[2*n+1:2*n+m^2])
+    V = vcat(gens(mixed_ring_xy)[1:n], [0 for i in 1:n], gens(mixed_ring_xy)[2*n+1:2*n+m])
```
fingolfin: Can't you also do this here, and then get rid of the definition of `m` above?
```suggestion
    V = vcat(gens(mixed_ring_xy)[1:n], [0 for i in 1:n], gens(mixed_ring_xy)[2*n+1:end])
```
also `gens` allocates, so even better:
```suggestion
    xyz = gens(mixed_ring_xy)
    V = vcat(xyz[1:n], [0 for i in 1:n], xyz[2*n+1:end])
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -475,7 +520,7 @@ function inv_generators(I::MPolyIdeal, G::LinearlyReductiveGroup, ringg::MPolyRi
     #we need the representation matrix and determinant of group matrix to be in basering(I)
     mixed_ring_xy = parent(genss[1])
     R = base_ring(M)
-    m = G.group[2]
+    m = group_dim(G)
```
fingolfin: Can't you again eliminate `m` here and also below (there by using `end`)?

In any case I would use `xyz = gens(mixed_ring_xy)` to avoid the repeated calls to `gens`

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -484,7 +529,7 @@ function inv_generators(I::MPolyIdeal, G::LinearlyReductiveGroup, ringg::MPolyRi
     new_det = mapp_(det_)
 
     #now we apply reynolds operator to genss
-    if G.group[1] == :SL #TODO other types of reductive groups
+    if group_type(G) == :SL #TODO other types of reductive groups
         new_gens_wrong_ring = [reynolds_function(genss[i], new_rep_mat, new_det, m) for i in 1:length(genss)]
```
fingolfin:
```suggestion
        new_gens_wrong_ring = [reynolds_function(g, new_rep_mat, new_det, m) for g in genss]
```

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -484,7 +529,7 @@ function inv_generators(I::MPolyIdeal, G::LinearlyReductiveGroup, ringg::MPolyRi
...
     else 
         return nothing
```
fingolfin, line 532: Below the fold:
- replace `V = vcat(V,...` by `append!(V, ...`
- `zeros(ringg, n+m^2)` can probably be changed to `zeros(ringg, length(xyz) - n)`
- replace `elemm != 0` by `!is_zero(elemm)`

fingolfin (approved):

experimental/InvariantTheory/src/InvariantTheory.jl
```diff
@@ -675,3 +720,78 @@ function affine_algebra(R::RedGroupInvarRing)
   end
   return domain(R.presentation), R.presentation
 end
+
+
+###################
+#In the case there is no reynolds operator
+###################
+
+#Algorithm 4.5.1 in Computational Invariant Theory by Derksen and Kemper
```
fingolfin: please add "second edition" or so (and/or reference `[DK15]` which is the key we use in our .bib file

experimental/InvariantTheory/src/InvariantTheory.jl
Lax202: m is required in this function as it is used in the Reynolds operator.

PR #3770, opened 2024-05-22
undo a renaming in a list of type names

ThomasBreuer:
The documentation of serializable objects lists types, thus `polynomial_ring` is not what one wants here.

Concerning the entry `Polynomial`: This variable does not exist, `PolyRingElem` would be the type in question.
However, I am not sure whether `SomethingElem` types should be mentioned here at all:

Can elements of a serializable object perhaps be assumed to be serializable?
If yes then `PolyRingElem` can be omitted, if no then I should add also the element types of the group types that are listed in the file.

### DIFF
--- a/docs/src/General/serialization.md
+++ b/docs/src/General/serialization.md
@@ -34,8 +34,8 @@ SimplicialComplex
 ### Commutative Algebra
 ```julia
 Ideal
-Polynomial
-polynomial_ring
+PolyRing
+PolyRingElem
 ```
 
 ### Groups


fingolfin:
I agree that it makes sense to omit `PolyRingElem` based on what is in this list so far.

Of course there are also many other things "missing", e.g. various types of fields. But perhaps that's intentional? Unclear to me.

antonydellavecchia:
Related to this 
https://github.com/oscar-system/Oscar.jl/pull/3768

antonydellavecchia:
Changes here have been merged in another branch

Issue #1920, opened 2023-02-10
Reducing usage of or even removing `groebner_assure`, `oscar_assure`, `singular_assure`


fingolfin:
In a recent PR, I saw code like this:
```julia
  groebner_assure(r.I, ordering)
  oscar_assure(r.I.gb[ordering])
  singular_assure(r.I.gb[ordering])
  SG = r.I.gb[ordering].gens.S
```
Why do we do it this way, instead of providing accessors for things like `I.gb[ordering].gens.S` that automatically do this for us? So let's say we
have a method 
```julia
function singular_groeber_gens(I::MPolyIdeal, ordering::MonomialOrdering)
  groebner_assure(r.I, ordering)
  singular_assure(r.I.gb[ordering])
  return I.gb[ordering].gens.S
end
```
and then original code above would just become
```julia
  SG = singular_groeber_gens(r.I, ordering)
```

Is there a fundamental problem with this alternative approach that I am missing?

Advantages of this approach include:
- we reduce the number of places that deal with implementation details (ideally most code would not need to know how and where Groebner bases are stored, and how they are computed
  -  this has many advantages: it makes future refactoring easier, makes it easier to write generic code, helps ensure code invariants ("if that value is set then this value also is set"), etc.
- the intent of the code is clearer: instead of randomly sprinkling `*_assure` calls, you just say what you want (e.g. in the example above, it is not readily clear whether and why `oscar_assure` is actually needed)
- the code arguably gets easier to read

thofma:
Isn't this exactly what we did? The code you quoted is itself part of the `*_assure` function and it is called in the accessors here:
https://github.com/oscar-system/Oscar.jl/blob/169ccdf41f7ac1a55e828530253aa215b1a11534/src/Rings/MPolyQuo.jl#L55-L58

And everywhere these accessors are called, which is why this fundamental change in the behaviour only took a few lines in #1917.

Maybe one could save a few lines in total by refactoring it yet more, but why?

ederc:
Yes, in the quotient business we are doing this. But @fingolfin is right, we should do this also in other settings, the direct call of `...gens.S` is widespread.

fingolfin:
Yes, the example was from your PR and I realize that PR implemented a step in the right direction -- but I disagree that there are just a few lines to save; I did not target that PR specifically (and hence explicitly did not reference it), it just was a convenient source for an example. I could have picked another one of the hundreds which are remaining. Indeed:
```
[...]

PR #3775, opened 2024-05-22
Removes some more singular_assure()  calls

ederc:
See #1920.


afkafkafk13 (approved):
Thank you for the cleaning up.

fingolfin (approved):
Thank you!!!

PR #3776, opened 2024-05-22
add `schur_cover`

ThomasBreuer:
(suggested by @fieker)

Note that the GAP function `EpimorphismSchurCover` does not necessarily return a map from a f. p. group.


fieker (approved):
thanks

fingolfin (approved):

Issue #280, opened 2021-02-12
Purge the examples folder


thofma:
I just spent too much time tracking down a bug which appeared when changing the version from `0.5.1-DEV` to `0.5.1`. The example folder files <strike>requires</strike> include code which is necessary for the non-example tests to pass. But the examples folder is only incuded, if `is_dev` is true.

fieker:
On Fri, Feb 12, 2021 at 02:26:22AM -0800, thofma wrote:
> I just spent too much time tracking down a bug which appeared when changing the version from `0.5.1-DEV` to `0.5.1`. The example folder code requires code which is necessary for the tests to pass. But the examples folder is only incuded, if `is_dev` is true.
> 
>
Move the stuff that is automatically tested in dev into experimental?
Where it is always included?

thofma:
https://github.com/oscar-system/Oscar.jl/blob/master/src/Oscar.jl#L218-L226

thofma:
I don't understand why this was not moved to experimental.

fieker:
Essentially this is a purely "experimental" implementation from a Fach Praktikum. Wolfram is waiting for the link to the Singular library
Smilar: All of Delphine's stuff should be there, but that needs coordination with her to avoid merge conflicts and frustartion

thofma:
@delphinepol Once your recent PR is merged, could you make a PR to move your things from `examples` to `experimental`? That would be great, thanks.

delphinepol:
OK !

lgoettgens:
I think that most of the code will probably be broken right now as all of this stuff is not covered by CI

lgoettgens:
Triage:
@fieker likes the examples folder. Someone (TM) would need to go through all of the examples code, fix it, remove stuff that is now in `src` or `experimental`, ...

Maybe @thofma or @fingolfin can be there next triage and bring up their opinion?

PR #3777, opened 2024-05-22
Remove `examples/BinomialIdeal.jl`

lgoettgens:
Progress towards https://github.com/oscar-system/Oscar.jl/issues/280.

All of this (except some hacky helper functions) has been migrated to `src/Rings/binomial_ideal.jl` already.
Furthermore, I maybe found a copy-paste error. Someone with knowledge about the code should check the line with the added comment.


fingolfin (approved):
Looks good to me

src/Rings/binomial_ideals.jl
```diff
@@ -293,7 +293,7 @@ function cellular_decomposition_macaulay(I::MPolyIdeal)
   end
   @assert !isone(I)
   @assert is_binomial(I)
-  return _cellular_decomposition(I)
+  return _cellular_decomposition(I) # Should this be _cellular_decomposition_macaulay(I)?
```
fingolfin: Good catch
```suggestion
  return _cellular_decomposition_macaulay(I)
```

PR #3779, opened 2024-05-22
No longer load code dependent on `is_dev`

lgoettgens:
This could lead to very hard to track down bugs in released versions (as it is not reproducible in the local clone).

I would even consider backporting this to 1.0. What do you guys think?

### DIFF
--- a/src/Oscar.jl
+++ b/src/Oscar.jl
@@ -283,13 +283,6 @@ include("Serialization/main.jl")
 
 include("../experimental/Experimental.jl")
 
-if is_dev
-#  include("../examples/ModStdNF.jl")
-#  include("../examples/ModStdQ.jl")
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
Looks good to me. I'd be fine with backporting, as I think nothing is accessing this (as it is hidden in a module `PrimDec` without any exports nor imports (other than `import Singular`), and there seems to be no methods installed for functions from outside `PrimDec` either. So e.g. we have this:
```
julia> Oscar.PrimDec.primary_decomposition
primary_decomposition (generic function with 1 method)

julia> Oscar.primary_decomposition
primary_decomposition (generic function with 7 methods)
```

But I think @benlorenz should decide

benlorenz (approved):
I think this is fine for merging and backporting.

PR #3780, opened 2024-05-22
Implement `simplify` for `ComplexOfMorphisms`

HechtiDerLachs:
Now the code for `canonical_bundle` looks sane. 

What happened? There was no method `simplify(::ComplexOfMorphism)`. And there could not be any reasonable such method which would not switch to the new lazy complexes as return values, since the `ComplexOfMorphism`s would need a lot of extra care to become sufficiently lazy. Hence I added that method and everything works just fine. No dubious calls to special methods in the code for `canonical_bundle` anymore. 

Ping @wdecker @jankoboehm

### DIFF
--- a/experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
+++ b/experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
@@ -332,6 +332,14 @@ function can_compute(fac::BaseChangeFromOriginalFactory, phi::AbsHyperComplexMor
 end
 
 
+function simplify(c::FreeResolution)
+  return simplify(SimpleComplexWrapper(c.C))
+end
+
+function simplify(c::ComplexOfMorphisms)
+  return simplify(SimpleComplexWrapper(c))
+end
+
 function simplify(c::AbsHyperComplex{ChainType, MorphismType}) where {ChainType, MorphismType}
   @assert dim(c) == 1 "complex must be one-dimensional"
   chain_fac = SimplifiedChainFactory(c)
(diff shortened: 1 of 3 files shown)


fingolfin:
I am confused by the title, "switch to lazy complexes". The other complexes are also lazy.

lgoettgens:
I am confused as well. You don't touch any code inside `canonical_bundle`, so how could become sane now?

HechtiDerLachs:
I do touch code in `canonical_bundle`. Before you had to use a special method of `free_resolution` which takes the desired type for the output (`Oscar.SimpleFreeResolution`) as first argument. The type is not exported and it would be weird to explain to any outsider why you would have to make this sophisticated special call to get the unleashed functionality of the hypercomplexes. Now you can just call the classical `free_resolution` and the passage to hypercomplexes is done by the corresponding new method for `simplify` one line below. This makes the use more streamlined and intuitive and it "hides" the passage to hypercomplexes. 

Earlier on I was deliberately hiding the access to the hypercomplexes a bit, since at the time I didn't want to interfere with the classical `ComplexOfMorphism`s too much. But now that people want to use it and we eventually fuse them with the existing functionality, this is a step forward. 

As I said: Having `simplify` return a `ComplexOfMorphisms` would most likely not be desirable, so the implementation given here seems legitimate.

HechtiDerLachs:
I was unaware of the existence of `canonical_bundle` as an Oscar function, hence the confusion. Before, I only changed the body of such a function with the same name which was declared locally in a test file. After consulting with @wdecker today, I now adjusted the actual OSCAR code.

jankoboehm:
It does exactly what we discussed, so I think we can go ahead.

wdecker (approved):

PR #3783, opened 2024-05-23
Book Metadata

aaruni96:
Add some more book metadata from https://link.springer.com/book/9783031621260 to README and CITATION.


HereAround (approved):
Thank you @aaruni96 . Looks good to me!

PR #3786, opened 2024-05-23
Make simplify_light return maps also for FreeModule case

### LINKED ISSUE #1745: Ext/ Tor outside sensible paramters crash
**Describe the bug**
Calling ext/ tor with a large index causes error - instead of returing s.th. trivial

**To Reproduce**
```
R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
A = R[x; y]
B = R[x^2; x*y; y^2; z^4]
M = subquotient(A, B)
F = free_module(R, 1)
Q, _ = quo(F, [x*F[1]])
```
Then `tor(Q, M, 20)` or `ext(Q, M, 20)` will not work

aaruni96:
Fixes #1745

### DIFF
--- a/src/Modules/UngradedModules/HomologicalAlgebra.jl
+++ b/src/Modules/UngradedModules/HomologicalAlgebra.jl
@@ -233,7 +233,7 @@ function tor(M::ModuleFP, N::ModuleFP, i::Int)
   return simplify_light(homology(lifted_resolution,i))[1]
 end
 
-simplify_light(F::FreeMod) = F
+simplify_light(F::FreeMod) = (F, identity_map(F), identity_map(F))
 
 #TODO, mF
 #  (hom lift) => hom and tensor functor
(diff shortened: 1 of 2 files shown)


afkafkafk13 (approved):
Thank you. Now the return values are consistent.

fieker:
On Thu, May 23, 2024 at 07:35:18AM -0700, Anne Frühbis-Krüger wrote:
> @afkafkafk13 approved this pull request.
> 
> Thank you. Now the return values are consistent.
> 
> 
Seperate question: why two maps? Wouldn't a single isomorphism (with the
preimage installed) be the natural interface?

afkafkafk13:
> On Thu, May 23, 2024 at 07:35:18AM -0700, Anne Frühbis-Krüger wrote: @afkafkafk13 approved this pull request. Thank you. Now the return values are consistent.
> Seperate question: why two maps? Wouldn't a single isomorphism (with the preimage installed) be the natural interface?
> […](#)
> -- Reply to this email directly or view it on GitHub: [#3786 (review)](https://github.com/oscar-system/Oscar.jl/pull/3786#pullrequestreview-2074171486) You are receiving this because you are subscribed to this thread. Message ID: ***@***.***>

If you are interested in moving objects/subobjects back and forth, this is indeed the case. But if you need to give one map or the other as data to further computations, you need the map. Unfortunately, there is no reason to believe that the same direction would suffice for all applications which the user might have in mind.

PR #3787, opened 2024-05-23
enable broadcasting for `GAPGroupElement`

ThomasBreuer:
(suggested by @fieker)

Should this eventually be moved to `AbstractAlgebra.GroupElem`?
And wouldn't broadcasting be interesting also for `FinGenAbGroupElem` or `AbstractAlgebra.AdditiveGroupElem`?

### DIFF
--- a/src/Groups/types.jl
+++ b/src/Groups/types.jl
@@ -28,6 +28,9 @@ i.e., if `g` is a `GAPGroupElem`, then `GapObj(g)` is the `GapObj` underlying `g
 """
 abstract type GAPGroupElem{T<:GAPGroup} <: AbstractAlgebra.GroupElem end
 
+## Scalar for broadcasting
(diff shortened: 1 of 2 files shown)


lgoettgens:
I like this, but prefer to directly add it to AbstractAlgebra instead. https://github.com/Nemocas/AbstractAlgebra.jl/pull/1714 will take care of that.

ThomasBreuer:
Thanks @lgoettgens.
(Once the change in AbstractAlgebra.jl becomes available in Oscar.jl, I will still add the proposed test to Oscar.jl.)

lgoettgens:
> Thanks @lgoettgens. (Once the change in AbstractAlgebra.jl becomes available in Oscar.jl, I will still add the proposed test to Oscar.jl.)

They are released now as part of AA 0.41.7

PR #3788, opened 2024-05-23
Use the 'official' partitions in intersection theory

joschmitt:
As pointed out in https://github.com/oscar-system/Oscar.jl/pull/3389#discussion_r1600763120.


afkafkafk13 (approved):
Thank you!
Yes, this does the job and uses the official function.

lgoettgens (approved):

PR #3789, opened 2024-05-24
Removes singular_assure() calls from mpoly-local implementation

ederc:
See #1920.

### DIFF
--- a/src/Rings/mpoly-local.jl
+++ b/src/Rings/mpoly-local.jl
@@ -341,24 +341,27 @@ function ideal(Rx::MPolyRingLoc, g::Singular.sideal)
   return MPolyIdealLoc(Rx, g)
 end
 
-# Computes the Singular.jl data of an MPolyIdealLoc if it is not defined yet.
-function singular_assure(I::MPolyIdealLoc)
-  singular_assure(I.gens)
-end
-
-function singular_assure(I::IdealGensLoc)
+function singular_generators(I::IdealGensLoc)
   if !isdefined(I, :S)
-    R = base_ring(I.Ox)
-    m = I.Ox.max_ideal
-    Q = I.Ox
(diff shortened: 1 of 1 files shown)


joschmitt (dismissed):
Looks good to me.
Stupid question though: is anyone using this code? From what I can see, in the algebraic geometry code people seem to use the stuff from `mpoly-localizations.jl`.

ederc:
> Looks good to me. Stupid question though: is anyone using this code? From what I can see, in the algebraic geometry code people seem to use the stuff from `mpoly-localizations.jl`.

I have no idea. We discussed the `singular_assure()` issue on Wednesday, no one from the Geometry crew wanted the code to be removed. In the end I do not care.

joschmitt:
I'm just wondering how much more time and energy we (or more precisely you) should put into this, if this code could just go.

@wdecker @HechtiDerLachs @afkafkafk13 @simonbrandhorst @jankoboehm does anybody know about the status of this file? Why do we have `MPolyLocRing` and `MPolyRingLoc`?

HechtiDerLachs:
I will make a PR to remove `src/Rings/mpoly-local.jl` if that's OK with everybody. 

If you disagree: Please speak up!

afkafkafk13:
> I will make a PR to remove `src/Rings/mpoly-local.jl` if that's OK with everybody.
> 
> If you disagree: Please speak up!

I am not against it, but I would like to ask you to pay special attention not to break the following functionalities:
- duValSing.jl
- SpaceGerms.jl (in particular rational_point_coordinates)
- Tjurina.jl

from experimental/src/Schemes. I hope that everything is consistent, but these are the parts where the local ordering are used. If something breaks, then it will most probably be visible there.

afkafkafk13:
This will most probably be obsoleted by #3790

PR #3790, opened 2024-05-24
Remove mpoly-local.jl.

HechtiDerLachs:
As discussed last Wednesday.

ping: @joschmitt , @ederc , @afkafkafk13 , @fingolfin 

closes #3789


test/Rings/mpoly-local.jl
```diff
@@ -28,17 +28,19 @@ end
...
+  @test fraction(c2) == (x*(x + y))//((1 + x)*(3 + x - y))
 
   (x, y, z) = map(Q, (x, y, z))
-  @test 1//x + 1//y == (x + y)//(x*y)
-  @test_throws Exception x//(x - 2)
+  @test inv(x) + inv(y) == (x + y)*inv(x*y)
+  @test_throws Exception x*inv(x - 2)
 end
 
+#=
```
HechtiDerLachs: We should remove this eventually. Let me know if this is OK.

src/Rings/mpoly.jl
```diff
@@ -200,7 +200,7 @@ mutable struct IdealGens{S}
       r.gens = BiPolyArray(Ox, S)
       r.isGB = S.isGB
       r.isReduced = isReduced
-      if T <: Union{MPolyRing, MPolyRingLoc}
+      if T <: Union{MPolyRing}
```
afkafkafk13: I don't see why you would still need the union here.

test/Rings/mpoly-local.jl
afkafkafk13: Indeed, talking about a groebner basis w.r.t. local ordering is not consistent with the notation (both in textbooks and in other parts of Oscar). I agree that this test should be removed.

afkafkafk13 (commented):
No objections. Let us see whether problems pop up in the tests. (I keep my fingers crossed)

joschmitt:
The function `minimal_generators` is still listed in `src/exports.jl` and should be removed now. From what I can see, it is replaced by `minimal_generating_set` in functionality, so one could add a deprecation?
Maybe also the types should be deprecated (seeing that we are now 1.0 and stable and so on)?

HechtiDerLachs:
> Maybe also the types should be deprecated

It looks like the types were never exported. Do we need to deprecate them nevertheless?

joschmitt:
> > Maybe also the types should be deprecated
> 
> It looks like the types were never exported. Do we need to deprecate them nevertheless?

Right, probably not necessary then.

HechtiDerLachs:
So far tests were only failing because of some Aqua issues. Fingers crossed...

joschmitt:
> So far tests were only failing because of some Aqua issues. Fingers crossed...

That's the undefined export of `minimal_generators`.

joschmitt:
Ugh, `MPolyRingElemLoc` is exported...

src/Rings/Rings.jl
```diff
@@ -20,7 +20,7 @@ include("MPolyMap/MPolyQuo.jl")
 include("MPolyMap/AffineAlgebras.jl")
 include("MPolyMap/base_change.jl")
 
-include("mpoly-local.jl")
+#include("mpoly-local.jl")
```
lgoettgens:
```suggestion
```
if the file no longer exists, the commented out include just creates confusion

src/deprecations.jl
```diff
@@ -123,6 +123,10 @@ Base.@deprecate_binding StdSpec StdAffineScheme
 # deprecated for 1.1
 @deprecate morphism_of_projective_schemes morphism
 
+# deprecated from src/Rings/mpoly-local.jl
```
lgoettgens:
```suggestion
```
there is no similar comment elsewhere in this file

src/deprecations.jl
```diff
@@ -123,6 +123,10 @@ Base.@deprecate_binding StdSpec StdAffineScheme
...
+@deprecate minimal_generators minimal_generating_set
+Base.@deprecate_binding MPolyRingElemLoc MPolyLocRingElem
```
lgoettgens: Please put them to the bottom of the file to keep it easier to track historical changes

test/Rings/mpoly-local.jl
```diff
@@ -5,8 +5,8 @@
   I = ideal(Q, [x - 2, (y - 1)^2*z])
```
lgoettgens: Since the corresponding src file is removed, this file should go as well. The contained tests could get moved to `test/Rings/mpoly-localization.jl` instead

lgoettgens (commented):
some style comments. please wait for the CI to succeed first before investing any time in resolving them

joschmitt (commented):
This looks good to me from a technical point of view, but I would prefer if someone who knows about the history of this file could approve.

ederc (commented):
I am not the right one to review this PR.

HechtiDerLachs:
Brief account of the history of this file: Afaik this was due to Raul Epure, probably as an early proof of concept. I am not aware of anyone using these things really. They served for me as a blueprint for the implementation of my more general localizations. I build them up in parallel back then, since I did not feel sufficiently mature yet to immediately replace already grown structures in OSCAR. Eventually it was never really decided how to handle this parallelism and we should probably have removed it already for Oscar 1.0. However, it never seemed really urgent and so it got lost. 

Everything that works for the "old" localizations can now also be handled with the new ones (for `MPolyComplementOfKPointIdeal` as a multiplicative set). So no functionality is lost by removing this.

afkafkafk13 (approved):
From the technical side I am happy with this. It seems that it salvages everything crucial to the localization framework (mpoly-localizations.jl).

Concerning the history of the removed file, I can say that it goes back to experiments predating the localization framework introduced by @HechtiDerLachs and that by now nothing relies (or seems to rely) on these experiments anymore. Unfortunately, the original author of the file ( @raulepure ) is no longer available on the team to approve this PR.

I do not feel qualified to give the only and final approval, but I confidently give an approval. We should ask about objections from the others at triage on Wednesday -- just to be on the safe side.

simonbrandhorst (approved):

fingolfin (approved):
Fine by me but let's discuss it in triage

fingolfin:
Question during triage: in the cases were both the new and old implementation are applicable, did you make any performance comparisons? If the new code is not too much worse (or even better) then that's fine.


Other than that, there were no objections to merging. I'll wait to hear from @HechtiDerLachs about the performance bit.

ederc:
When this PR is merged one can just close #3789.

HechtiDerLachs:
For the timing I can only run the old tests vs. the new tests.

Old:
```julia
julia> @time begin
  R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
  m = ideal(R, [y - 1, x - 2, z - 3])
  Q = localization(R, m)
  I = ideal(Q, [x - 2, (y - 1)^2*z])
  J = ideal(Q, [x - 2, y - 1])
  a = Q(x//(1 + x))
  b = Q((x + y)//(3 + x - y))

  K1 = I+J
  K2 = I*J
  K3 = J^3
  c1 = a+b
  c2 = a*b

  @test iszero(a-a)
  @test a*deepcopy(a) == a^2
  @test det(matrix(Q, [1 x; y z])) == Q(z-x*y)

  @test K1.gens.O == Q.([x - 2, y^2*z - 2*y*z + z, y - 1])
  @test K2.gens.O == Q.([x^2 - 4*x + 4, x*y - x - 2*y + 2, x*y^2*z - 2*x*y*z + x*z - 2*y^2*z + 4*y*z - 2*z, y^3*z - 3*y^2*z + 3*y*z - z])
  @test K3.gens.O == Q.([x^3 - 6*x^2 + 12*x - 8, x^2*y - x^2 - 4*x*y + 4*x + 4*y - 4, x*y^2 - 2*x*y + x - 2*y^2 + 4*y - 2, y^3 - 3*y^2 + 3*y - 1])
  @test c1.frac == (x*(3 + x - y) + (1 + x)*(x + y))//((1 + x)*(3 + x - y))
  @test c2.frac == (x*(x + y))//((1 + x)*(3 + x - y))

  (x, y, z) = map(Q, (x, y, z))
  @test 1//x + 1//y == (x + y)//(x*y)
  @test_throws Exception x//(x - 2)
end;
  0.002512 seconds (6.49 k allocations: 283.660 KiB)
```
New:
```julia
julia> @time begin
         R, (x, y, z) = polynomial_ring(QQ, ["x", "y", "z"])
         m = ideal(R, [y - 1, x - 2, z - 3])
         Q = localization(R, m)
         I = ideal(Q, [x - 2, (y - 1)^2*z])
         J = ideal(Q, [x - 2, y - 1])
         a = Q(x//(1 + x))
         b = Q((x + y)//(3 + x - y))

         K1 = I+J
         K2 = I*J
         K3 = J^3
         c1 = a+b
         c2 = a*b

         @test iszero(a-a)
         @test a*deepcopy(a) == a^2
         @test det(matrix(Q, [1 x; y z])) == Q(z-x*y)

         @test gens(K1) == Q.([x - 2, y^2*z - 2*y*z + z, x - 2, y - 1])
         @test gens(K2) == Q.([x^2 - 4*x + 4, x*y - x - 2*y + 2, x*y^2*z - 2*x*y*z + x*z - 2*y^2*z + 4*y*z - 2*z, y^3*z - 3*y^2*z + 3*y*z - z])
         @test gens(K3) == Q.([x^3 - 6*x^2 + 12*x - 8, x^2*y - x^2 - 4*x*y + 4*x + 4*y - 4, x^2*y - x^2 - 4*x*y + 4*x + 4*y - 4, x*y^2 - 2*x*y + x - 2*y^2 + 4*y - 2, x^2*y - x^2 - 4*x*y + 4*x + 4*y - 4, x*y^2 - 2*x*y + x - 2*y^2 + 4*y - 2, x*y^2 - 2*x*y + x - 2*y^2 + 4*y - 2, y^3 - 3*y^2 + 3*y - 1])
         @test fraction(c1) == (x*(3 + x - y) + (1 + x)*(x + y))//((1 + x)*(3 + x - y))
         @test fraction(c2) == (x*(x + y))//((1 + x)*(3 + x - y))

         (x, y, z) = map(Q, (x, y, z))
         @test inv(x) + inv(y) == (x + y)*inv(x*y)
         @test_throws Exception x*inv(x - 2)
       end;
  0.001596 seconds (3.31 k allocations: 134.988 KiB)
```
Is merging fine with everyone, then? 

Edit: Interestingly the second block took 55 seconds (!) in the first run due to precompilation. But later runs yield better results, as it seems.

joschmitt:
I can confirm the timings, so I merge this now.

Issue #3785, opened 2024-05-23
`min_weights` returns zero weight vector for non-trivial subdivisions


YueRen:
@oskarhenriksson and I encountered the following bug while trying to figure out the Pluecker vector of the snowflake tropical linear space:

```
using Oscar
B = [ 1 0 1 0 0 0;
      1 0 0 1 0 0;
      1 0 0 0 1 0;
      1 0 0 0 0 1;
      0 1 1 0 0 0;
      0 1 0 1 0 0;
      0 1 0 0 1 0;
      0 1 0 0 0 1;
      0 0 1 0 1 0;
      0 0 1 0 0 1;
      0 0 0 1 1 0;
      0 0 0 1 0 1 ]
C = IncidenceMatrix([[1, 2, 5, 6],[3, 4, 7, 8],[9, 10, 11, 12]])
S = subdivision_of_points(B,C)
maximal_cells(S) # checking that S has non-trivial maximal cells
min_weights(S) # returned weight vector is 0 and induces trivial subdivision
```

Here the output:
```
julia> maximal_cells(S)
3-element SubObjectIterator{Vector{Int64}}:
 [1, 2, 5, 6]
 [3, 4, 7, 8]
 [9, 10, 11, 12]

julia> min_weights(S)
12-element Vector{Int64}:
 0
 0
 0
 0
 0
 0
 0
 0
 0
 0
 0
 0

```

YueRen:
Equally surprising, the following two regular subdivisions (with non-trivial weight vectors) are trivial.  Surely one of them shouldn't be?

```
julia> w = [0,0,1,1,0,0,1,1,2,2,2,2];

julia> maximal_cells(subdivision_of_points(B,-w))
1-element SubObjectIterator{Vector{Int64}}:
 [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]

julia> maximal_cells(subdivision_of_points(B,+w))
1-element SubObjectIterator{Vector{Int64}}:
 [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]

```

oskarhenriksson:
In case it helps, the secondary cone of `S` has no rays, and the lineality space is as follows:

```
julia> SC = secondary_cone(S)
Polyhedral cone in ambient dimension 12

julia> rays_modulo_lineality(SC)
[...]

PR #3791, opened 2024-05-24
Some requests in polyhedral geometry

### LINKED ISSUE #3748: Casting `PolyhedralFan` to `PolyhedralComplex`
It happens now and then that I want to convert a polyhedral fan to a polyhedral complex in Oscar. There are already versions of `polyhedral_complex` that do this for polyhedra and tropical varieties, so perhaps it could make sense to have one also for fans? 

I'm currently using the following function locally (thanks for @lkastner for helping me with this over on Slack!). Perhaps something along these lines should be in Oscar? 

```
function polyhedral_complex_from_fan(F::PolyhedralFan)
    K = F.parent_field
    rml, L = rays_modulo_lineality(F)
    rays_and_vertex = vcat(matrix(K, rml), zero_matrix(ZZ, 1, ambient_dim(F)))
    cells = maximal_cones(IncidenceMatrix, F)
    new_cells = IncidenceMatrix(nrows(cells), ncols(cells)+1);
    new_cells[1:nrows(cells), 1:ncols(cells)] = cells;
    for i in 1:nrows(cells) 
        new_cells[i, ncols(cells)+1] = 1
    end
    return polyhedral_complex(new_cells, rays_and_vertex, collect(1:length(rml)))
end
```

lkastner:
- Implement `polyhedron(Cone)`
- Implement `polyhedral_complex(PolyhedralFan)`
- Add warning for subdivision of points #3785

### DIFF
--- a/src/PolyhedralGeometry/PolyhedralComplex/constructors.jl
+++ b/src/PolyhedralGeometry/PolyhedralComplex/constructors.jl
@@ -210,7 +210,7 @@ function polyhedral_complex(F::PolyhedralFan{T}) where {T<:scalar_types}
       Polymake.give(pmo_in, "MAXIMAL_CONES_LINEAR_SPAN_NORMALS"),
     )
   end
-  return PolyhedralComplex{T}(pmo_out, F.parent_field)
+  return PolyhedralComplex{T}(pmo_out, coefficient_field(F))
 end
 
 ###############################################################################
--- a/src/PolyhedralGeometry/PolyhedralComplex/constructors.jl
+++ b/src/PolyhedralGeometry/PolyhedralComplex/constructors.jl
@@ -178,26 +178,37 @@ Turn a polyhedral fan into a polyhedral complex.
 function polyhedral_complex(F::PolyhedralFan{T}) where {T<:scalar_types}
   pmo_in = pm_object(F)
   pmo_out = Polymake.fan.PolyhedralComplex{_scalar_type_to_polymake(T)}()
-  for prop in ["RAYS", "INPUT_RAYS"]
+  for prop in ["RAYS", "INPUT_RAYS", "FACET_NORMALS"]
     if Polymake.exists(pmo_in, prop)
       Polymake.take(pmo_out, prop, embed_at_height_one(Polymake.give(pmo_in, prop), true))
     end
   end
-  for prop in ["INPUT_LINEALITY", "LINEALITY_SPACE"]
+  for prop in ["INPUT_LINEALITY", "LINEALITY_SPACE", "LINEAR_SPAN_NORMALS"]
     if Polymake.exists(pmo_in, prop)
       Polymake.take(pmo_out, prop, embed_at_height_one(Polymake.give(pmo_in, prop), false))
     end
   end
-  for prop in ["MAXIMAL_CONES", "INPUT_CONES"]
+  for prop in ["MAXIMAL_CONES", "INPUT_CONES", "MAXIMAL_CONES_FACETS"]
     if Polymake.exists(pmo_in, prop)
       inprop = Polymake.give(pmo_in, prop)
-      outprop = Polymake.IncidenceMatrix(nrows(inprop), ncols(inprop)+1)
+      outprop = if inprop isa Polymake.IncidenceMatrix
+        Polymake.IncidenceMatrix(nrows(inprop), ncols(inprop) + 1)
+      else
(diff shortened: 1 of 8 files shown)


lkastner:
cc @YueRen @oskarhenriksson @benlorenz

YueRen (approved):
Looks good, thanks a lot!

src/PolyhedralGeometry/PolyhedralComplex/constructors.jl
```diff
@@ -170,6 +170,38 @@ end
...
+      inprop = Polymake.give(pmo_in, prop)
+      outprop = Polymake.IncidenceMatrix(nrows(inprop), ncols(inprop)+1)
+      outprop[1:end, 2:end] = inprop
+      for i in 1:nrows(inprop)
+        outprop[i,1] = 1
+      end
+      Polymake.take(pmo_out, prop, outprop)
+    end
+  end
+  return PolyhedralComplex{T}(pmo_out, F.parent_field)
```
alexej-jordan:
```suggestion
  return PolyhedralComplex{T}(pmo_out, coefficient_field(F))
```
We should avoid directly accessing fields of a struct and use the specific functions instead.

src/PolyhedralGeometry/Polyhedron/constructors.jl
```diff
@@ -290,6 +294,29 @@ convex_hull(
...
+    if Polymake.exists(pmo_in, prop)
+      Polymake.take(pmo_out, prop, embed_at_height_one(Polymake.give(pmo_in, prop), true))
+    end
+  end
+  for prop in ["INPUT_LINEALITY", "LINEALITY_SPACE", "EQUATIONS", "LINEAR_SPAN"]
+    if Polymake.exists(pmo_in, prop)
+      Polymake.take(pmo_out, prop, embed_at_height_one(Polymake.give(pmo_in, prop), false))
+    end
+  end
+  return Polyhedron{T}(pmo_out, C.parent_field)
```
alexej-jordan:
```suggestion
  return Polyhedron{T}(pmo_out, coefficient_field(C))
```
As above

src/PolyhedralGeometry/PolyhedralComplex/constructors.jl
```diff
@@ -170,6 +170,49 @@ end
...
 
+@doc raw"""
+    polyhedral_complex(F::PolyhedralFan)
+
+Turn a polyhedral fan into a polyhedral complex.
+"""
+function polyhedral_complex(F::PolyhedralFan{T}) where {T<:scalar_types}
+  pmo_in = pm_object(F)
+  pmo_out = Polymake.fan.PolyhedralComplex{_scalar_type_to_polymake(T)}()
+  for prop in ["RAYS", "INPUT_RAYS", "FACET_NORMALS"]
```
fingolfin: Looking at this I wondered "hmm, how many allocations will this make? Can it reuse the strings? What about the vector?"

Turns out if you change this from a vector to a tuple, you can avoid all allocations in that line. I.e.:
```suggestion
  for prop in ("RAYS", "INPUT_RAYS", "FACET_NORMALS")
```

To demonstrate the difference:
```julia
julia> function f(s)
         for prop in ["MAXIMAL_CONES", "INPUT_CONES", "MAXIMAL_CONES_FACETS"]
           s == prop && return true
         end
         return false
       end
f (generic function with 1 method)

julia> function g(s)
         for prop in ("MAXIMAL_CONES", "INPUT_CONES", "MAXIMAL_CONES_FACETS")
           s == prop && return true
         end
         return false
       end
g (generic function with 1 method)

julia> @btime f("MAXIMAL_CONES")
  25.435 ns (1 allocation: 80 bytes)
true

julia> @btime g("MAXIMAL_CONES")
  1.500 ns (0 allocations: 0 bytes)
true
```
Mind you, either is plenty fast and I assume it won't matter for `polyhedral_complex` which hopefully nobody will call in a tight loop ;-) so this is *not* a change request -- I just wondered about the behaviour of this code myself and thought I'd share my findings for future reference :-)

lkastner: Nice, thank you.

YueRen: I learned something new in julia today.  So is it generally recommended to use tuples over vectors when (easily) possible?

PR #3792, opened 2024-05-24
`FreeResolution` as hypercomplex

HechtiDerLachs:
This is a prototype for how `FreeResolution` could be turned into a hypercomplex with minimal effort. 

I made the PR so that this approach can be discussed and tried out. Nothing of this needs to be done/decided urgently! I just wanted to have a concrete suggestion on the table so that everyone can get an impression about the implications. 

Some tests have been disabled for the time being. This is mostly due to the fact that for hypercomplexes the functor `hom(-, M)` takes
```
   0        1        2
   A  <---  B  <---  C
```
to 
```
    -2            -1            0
  Hom(C, M) <-- Hom(B, M) <-- Hom(A, M)
```
rather than 
```
    0            1              2
  Hom(A, M) --> Hom(B, M) --> Hom(C, M)
```
I would very much prefer not to change this convention. 

If this behaviour is disliked for `FreeResolution`s, there are still some things we can adjust for the particular type. Feel free to try it out and let me know your preferences whenever it is convenient.

### DIFF
--- a/experimental/DoubleAndHyperComplexes/src/DoubleAndHyperComplexes.jl
+++ b/experimental/DoubleAndHyperComplexes/src/DoubleAndHyperComplexes.jl
@@ -1,4 +1,7 @@
 include("Objects/Types.jl")
+### Modified FreeResolution
+include("FreeResolutionTypes.jl")
+
 include("Objects/Attributes.jl")
 include("Objects/tensor_products.jl")
 include("Objects/tor.jl")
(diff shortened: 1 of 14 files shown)


HechtiDerLachs:
Don't be confused by the size of the PR! This is mostly due to the need to bring everything about `FreeResolution` to `experimental/DoubleAndHyperComplexes` due to the dependency on the types only introduced there. In fact, I changed very little about the actual implementations for `FreeResolution`. The only difference is that for things like `hom`, `tensor_product`, etc. it is now calling the generic methods for hypercomplexes.

PR #3794, opened 2024-05-24
Implement `change_base_ring` for hypercomplexes

HechtiDerLachs:
@micjoswig : This is a prototype for what you asked for during the meeting. Have a look at the tests to see how to get started. 

Currently there is little functionality supported for `ModuleFP`s over the integers. If you show me the type of complexes you're interested in, then I can work on making them available here.


micjoswig:
Thanks. I will look it this later, after July 10.

HechtiDerLachs:
@micjoswig : Any new opinions on this? If yes, we might put this PR on the list for discussion for the upcoming Wednesday meeting.

micjoswig:
The relevant complexes don't exist in Oscar yet. My basic example would be the boundary complex of a finite simplicial complex.

It is not very difficult. But we need to decide which part we take from polymake (or none at all).

@lkastner @benlorenz

HechtiDerLachs:
Thanks for the reply. 

To my understanding we can proceed with this PR in the meantime. How to bring over applications from Polymake is a different topic, I think. And we will need this functionality for other things, too.

fieker:
Janko is (organizing or doing a) review

fingolfin:
@HechtiDerLachs made some updates but tests are failing now. He'll work on it and together with @jankoboehm get this merged.

fingolfin:
tests are passing now, and here in triage @jankoboehm said he is fine with this being merged, so I'll just do that.

PR #3795, opened 2024-05-24
Adjusting docu following PR #3700

wdecker:


afkafkafk13:
Slightly unrelated question: Do we currently also handle the case of free resolutions over localizations at point ideals? If so, do we want to mention that in this case the minimal free resolution is unique, too?

wdecker:
> Slightly unrelated question: Do we currently also handle the case of free resolutions over localizations at point ideals? If so, do we want to mention that in this case the minimal free resolution is unique, too?

We eventually must do this.  One problem is that Singular currently does not compute minimal such resolutions: It only removes constant units.

afkafkafk13:
> > Slightly unrelated question: Do we currently also handle the case of free resolutions over localizations at point ideals? If so, do we want to mention that in this case the minimal free resolution is unique, too?
> 
> We eventually must do this. One problem is that Singular currently does not compute minimal such resolutions: It only removes constant units.

Can you specify an example for this behaviour? I would like to better understand the problem. 'mres' in Singular always seemed to do the right thing w.r.t. a local ordering like 'ds'. In particular, I witnessed instances where non-constant units were handled properly.

afkafkafk13 (approved):
Good to go.

Maybe open a (feature request) issue with lower priority as a reminder that we need to add minimal resolution w.r.t local orderings into the framework and documentation.

wdecker:
> > > Slightly unrelated question: Do we currently also handle the case of free resolutions over localizations at point ideals? If so, do we want to mention that in this case the minimal free resolution is unique, too?
> > 
> > 
> > We eventually must do this. One problem is that Singular currently does not compute minimal such resolutions: It only removes constant units.
> 
> Can you specify an example for this behaviour? I would like to better understand the problem. 'mres' in Singular always seemed to do the right thing w.r.t. a local ordering like 'ds'. In particular, I witnessed instances where non-constant units were handled properly.

I do not have an explicit example. Recall, however, that Singular just mimicks local rings with the help of local orderings, having no arithmetic for local rings, but offering some computations which are done over polynomial rings. So how do we represent the inverse of a unit such as 1+x in this context?

jankoboehm:
I guess you don't and use Mora's division?

afkafkafk13:
> I guess you don't and use Mora's division?

exactly -- an this is fine, as long as you want to work with ideals, not elements!

wdecker:
> I do not have an explicit example. Recall, however, that Singular just mimicks local rings with the help of local orderings, having no arithmetic for local rings, but offering some computations which are done over polynomial rings. So how do we represent the inverse of a unit such as 1+x in this context?

I actually was wrong here. So we should indeed proceed with discussing free resolutions in the local case. @HechtiDerLachs What is already available?

HechtiDerLachs:
As far as I know, local orderings are still not functional for modules, so we can't use them in Oscar. 
But we can do the following:
```julia
R, (x, y, z) = QQ[:x, :y, :z]
U = complement_of_point_ideal(R, [0, 0, 0])
L, loc = localization(R, U)
M = R[x y z; y z x-1]
I = ideal(minors(M, 2))
LI = loc(I)
L1 = FreeMod(L, 1)
LI, inc = LI*L1
M = cokernel(inc)
# Free resolution basically localizes a free resolution over R at this point
res, aug = free_resolution(Oscar.SimpleFreeResolution, M)
# We get the usual complex from the Hilbert-Burch theorem
matrix(map(res, 1))
matrix(map(res, 2))
# Now we minimize
c = simplify(res)
# We get a Koszul complex, but with denominators
matrix(map(c, 1))
matrix(map(c, 2))
```
Again, this is using the `simplify` for hypercomplexes. It works generically without problems. Maybe it needs some tuning in larger applications, but we would look into that once people have an application.

afkafkafk13:
To me, this is not the way to go for the minimal free resolutions in the local case. We should rely on the Singular implementation in that case.
Please compare, what happens for I= \langle (x-1)*z,(z-1)*y,(y-1)*x  \rangle in the localization of the polynomial ring at the origin. (This is of course the maximal ideal of the ring, written in a non-standard way.)

Singular code:
` 
ring r=0,(x,y,z),ds;
ideal I=(x-1)*z,y*(z-1),x*(y-1);
list li=mres(I,0);
li;
`
Output:
`[1]:
   _[1]=x
   _[2]=y
   _[3]=z
[2]:
   _[1]=x*gen(3)-z*gen(1)
   _[2]=x*gen(2)-y*gen(1)
   _[3]=y*gen(3)-z*gen(2)
[3]:
   _[1]=x*gen(3)-y*gen(1)+z*gen(2)
`

In your approach, we obtain something correct, but still carrying along unnecessary units:
`julia> matrix(map(c, 1))
[y*z - y]
[x*z - z]
[x*y - x]

julia> matrix(map(c, 2))
[ x*y - x                       0             -y*z + y]
[-x*y + x                 x*y - x   -x*z + y*z - y + z]
[ y*z - z   y^2*z - y^2 - y*z + y         -y*z^2 + y*z]

julia> matrix(map(c, 3))
[y*z - y + z   y*z - y   -x]

`
If we are in real life situations in localizations at the origin and you are only interested in ideals, modules, but not in elements, it is often vital to treat units in the Mora-way and obtain something which is easier to handle in future computations.

jankoboehm:
What we are lacking for module orderings is the conversion from an Oscar ordering to a Singular ordering. As far as I remember that was only implemented for ring orderings.

PR #3800, opened 2024-05-27
To fix issue with Betti Tables

### LINKED ISSUE #3798: Problems with function `betti_table`
@Lax202 

1) The function seems to print an extra empty line now?

2) In this loop in the body defining `betti_table`, gradedness is checked for the free part of `F`, but not for the modules at cohomological degree -1, -2:

for i in 0:n
    module_degrees = F[i].d
    module_degrees === nothing && error("One of the modules in the graded free resolution is not graded.")
    for degree in module_degrees
      idx = (i, degree)
      generator_count[idx] = get(generator_count, idx, 0) + 1
    end
  end

Please remove test in the loop and, instead, start the function body with something like

  `@assert is_standard_graded(base_ring(F)) "resolution must be defined over a standard graded ring"`
  `@assert is_graded(F) "resolution must be graded"`

Lax202:
This is an attempt to fix [Issue #3798](https://github.com/oscar-system/Oscar.jl/issues/3798) @wdecker 

- I have removed the extra line at the end of the show function of the Betti table. 
- I am unsure about the negative degrees in the Betti table. I've made some kind of change but don't have an example yet to test it on. Help with an example with negative cohomological degrees would be great.

### DIFF
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -1455,14 +1455,12 @@ total: 1  3  2
 """
 function betti_table(F::FreeResolution; project::Union{FinGenAbGroupElem, Nothing} = nothing, reverse_direction::Bool=false)
   @assert is_graded(F) "resolution must be graded"
-  @assert is_standard_graded(base_ring(F)) "resolution must be defined over a standard graded ring"
   generator_count = Dict{Tuple{Int, Any}, Int}()
   C = F.C
   rng = Hecke.map_range(C)
   n = first(rng)
   for i in 0:n
     module_degrees = F[i].d
-    module_degrees === nothing && error("One of the modules in the graded free resolution is not graded.")
     for degree in module_degrees
       idx = (i, degree)
       generator_count[idx] = get(generator_count, idx, 0) + 1
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -2573,7 +2573,6 @@ julia> betti_table(FA)
 ------------------
 total: 1  5  6  2  
 
-
 julia> minimal_betti_table(FA)
        0  1  2  3  
 ------------------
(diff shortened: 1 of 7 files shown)


wdecker:
@Lax202 Please, your changes with respect to catching non-gradedness are not helpful. You ONLY need to erase as well as copy and paste as I described in the issue:

Please remove test in the loop and, instead, start the function body with something like

@assert is_standard_graded(base_ring(F)) "resolution must be defined over a standard graded ring"
@assert is_graded(F) "resolution must be graded"

That is all.

Lax202:
@wdecker sorry for misunderstanding. I have added the assert statements.

wdecker:
Please delete this line:

module_degrees === nothing && error("One of the modules in the graded free resolution is not graded.")

wdecker (approved):

benlorenz:
> joschmitt added the backport 1.0.x label yesterday

I would rather not backport this as this is not really a bugfix and it _does_ change the output of the book examples (even if it is just whitespace). If I understand it correctly this is about doing extra checks to catch invalid input?

joschmitt:
> > joschmitt added the backport 1.0.x label yesterday
> 
> I would rather not backport this as this is not really a bugfix and it _does_ change the output of the book examples (even if it is just whitespace). If I understand it correctly this is about doing extra checks to catch invalid input?

Ah, okay, sorry. I just read "fix issue".

wdecker:
> _does_ change the output of the book examples

Indeed? In the latex version of the book it looks okay.

benlorenz:
The CI job for the book tests is complaining about the removed lines, e.g.:
```diff
  julia> betti(free_resolution(defining_ideal(D)))
  0  1
  -----------
  2    : 3  2
  -----------
  total: 3  2
- 
  
  julia> Oscar.rat_normal_curve_anticanonical_map(D)
```

But it turns out in the book there is no extra new line. This was probably not flagged because the old Documenter based checking needed some extra filtering which removed the extra newline. So backporting this PR would change the output of 1.0.x but on the other hand adjust it to match the output in the book.

benlorenz:
So on second thought I think it does make sense to backport this since it will bring 1.0 and the book closer together even if it does change the output of 1.0 slightly.

Please adjust the remaining jlcons, I think `exres2.jlcon` and `param.jlcon`, and once it is merged I will check if it applies cleanly to the release branch.

PR #3802, opened 2024-05-27
improve `map_word`

### LINKED ISSUE #3796: `map_word` inconsistency for free group with zero generators
```
julia> FF = free_group(1)
Free group of rank 1

julia> map_word(one(FF), [-1])
1

julia> FF = free_group(0)
Free group of rank 0

julia> map_word(one(FF), Int[]) |> typeof
Nothing
```
I think it should error or also return `1`.

@ThomasBreuer can you have a look?

ThomasBreuer:
do not use `init == nothing`

resolves #3796

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1863,6 +1863,7 @@ function map_word(g::FPGroupElem, genimgs::Vector; genimgs_inv::Vector = Vector(
   G = parent(g)
   Ggens = gens(G)
   if length(Ggens) == 0
+    @req !(init == nothing) "use '; init =...' if there are no generators"
     return init
   end
   gX = GapObj(g)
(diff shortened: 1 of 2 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -1863,6 +1863,7 @@ function map_word(g::FPGroupElem, genimgs::Vector; genimgs_inv::Vector = Vector(
   G = parent(g)
   Ggens = gens(G)
   if length(Ggens) == 0
+    @req !(init == nothing) "use '; init =...' if there are no generators"
```
```suggestion
    @req init !== nothing "use '; init =...' if there are no generators"
```
comparisons with `nothing` should be done using `===` or `!==` (instead of `==` or `!=`)

thofma (approved):
Thanks!

PR #3803, opened 2024-05-27
bug fix for automorphism group of simplicial complex

antonydellavecchia:
Corrected the translation from polymake permutation generatorsto Oscar perm group generators


lkastner (approved):
Now that I understand the difference between `cperm` and `perm` this is ok. (c=cyclic for future reference)

PR #3804, opened 2024-05-28
editorconfig: explicitly disable trim_trailing_whitespace

fingolfin:
... for the reasons described [here](https://github.com/oscar-system/Oscar.jl/pull/3757/files#r1612405431) and discussed previously on multiple occasions. But in a nutshell: don't add or remove whitespace in dozens of unrelated locations just because you fix a typo in a file, as otherwise you cause extra work for other people with no good justification, and will ultimately discourage some people from contributing.

We could perhaps one day turn this one, assuming we first remove all trailing whitespace everywhere (but even then I am not sure it's a good idea as not everyone will have their editor set up to honor `.editorconfig`, and so these will be re-introduced over time. Anyway, that's a future discussion as far as I am concerned, for perhaps after the Begehung).

Pinging all other project leads (@fieker @micjoswig @simonbrandhorst @thofma @wdecker), as I was told in private that there is "disagreement among OSCAR leaders" about this, so y'all should have a chance to object (or agree).

Closes #3757

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@ charset = utf-8
 end_of_line = lf
 indent_style = space
 insert_final_newline = true
-#trim_trailing_whitespace = true  # as long as we don't enforce this, this option causes more trouble than it helps
+trim_trailing_whitespace = false
 
 [*.jl]
 indent_size = 2


fieker:
Do not change unrelated code unless you have to. It's a pain for others having PR open or working on it. If we all agree on a global beautification, it will be in a quite period of stability, not before the "Begehung", and after a long discussion.

Just to add: out core buisness is math, not software engineering...

fieker (approved):

micjoswig:
In addition to my general approval: (1) I am not very religious about white space, but I like clean code.  (2) Currently, we have recommendations, which are not strictly enforced. Fine with me. (3) In the future we could change this to some stricter set of rules, but then they would need to be enforced automatically (with or without git).

micjoswig (approved):

lkastner:
I really want to stress that if your editor automatically fixes code that you are not actively working with, just because you are working on this file, then this is a clientside problem.

thofma (approved):
I concur.

simonbrandhorst (approved):

fingolfin:
@wdecker also agrees.

PR #3808, opened 2024-05-28
doc: use set_current_module in .md files, too

fingolfin:


lgoettgens:
Looks good!

Can you do the same in all `experimental/*/docs/*.md` files as well?

fingolfin:
@lgoettgens ahh I forgot about those. Will do!

lgoettgens:
And I think that this block to put in all md files is mentioned somewhere in the docs, ideally this should be adapted as well

fingolfin:
Huh, interesting: in `docs/src/InvariantTheory/finite_groups.md` they just use
```
DocTestSetup = Oscar.doctestsetup()
```
and it seems to work. That's a bit better, I'll try if that works everywhere.

But I've now also asked the Documenter people if there isn't a way to set this in a single place, globally...

fingolfin:
> And I think that this block to put in all md files is mentioned somewhere in the docs, ideally this should be adapted as well

I agree it would be reasonable to have such an explanation, but I didn't find any in the existing docs. Did I miss it?

lgoettgens:
> Huh, interesting: in `docs/src/InvariantTheory/finite_groups.md` they just use
> 
> ```
> DocTestSetup = Oscar.doctestsetup()
> ```
> 
> and it seems to work. That's a bit better, I'll try if that works everywhere.

Nice find! Yeah, this should work. 

> But I've now also asked the Documenter people if there isn't a way to set this in a single place, globally...

That would be great

lgoettgens:
> > And I think that this block to put in all md files is mentioned somewhere in the docs, ideally this should be adapted as well
> 
> I agree it would be reasonable to have such an explanation, but I didn't find any in the existing docs. Did I miss it?

I didn't find it as well.

lgoettgens (approved):

PR #3814, opened 2024-05-29
add a paragraph about CTRL-C in Julia and GAP

ThomasBreuer:


docs/src/General/other.md
```diff
@@ -62,6 +62,18 @@ This section describes differences between GAP and Oscar.
...
+
+  When an error occurs or when the user hits ctrl-C in a GAP session,
+  usually a break loop is entered,
+  from which one can either try to continue the computations,
+  by entering `return`, or return to the GAP prompt, by entering `quit`;
+  in the latter case, some objects may be corrupted afterwards.
+
+  In a Julia session, one gets automatically back to the Julia prompt
+  when an error occurs or when the user hits ctrl-C,
+  and again some objects may be corrupted afterwards.
```
fingolfin, line 75: As I stated on Monday: this corruption issue in principle affects everything, including (AFAIK) code provided by the other cornerstones, and even native Julia code. So it feels odd to me to single out that it can happen when calling into GAP.

ThomasBreuer: The section is about differences between Julia and GAP, here the fact that GAP sessions have a break loop and Julia has not.

fingolfin (approved):

PR #3815, opened 2024-05-31
Remove unnecessary `set_name!` calls in resolutions and presentations

### LINKED ISSUE #3730: Fix name handling of elements of complexes/presentations/resolutions
x-ref https://github.com/oscar-system/Oscar.jl/pull/3700#discussion_r1596686047 cc @fingolfin 

A lot of code creating complexes/presentations/resolutions extracts the name of the base ring once and then uses it to set names of the parts using something like
```julia
AbstractAlgebra.set_name!(F0, "$br_name^$(ngens(mod))")
```

If the name of the base ring changes, the name of the parts still references the old name of the base ring.

Such a use case should be covered by the AbstractAlgebra fancy/magic name printing. I will have a look if this can be achieved using the current AA functionality and otherwise try to come up with some new interface in AA.

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/3730.

This does not change any actual printing (as long as the ring keeps it name) due to the `extra_name` method in
https://github.com/lgoettgens/Oscar.jl/blob/477a289cdbe6becb3ea648aa7f6d85fbc02086c1/src/Modules/UngradedModules/FreeMod.jl#L95-L104

One case where there is a change in behavior is when e.g. the ring is no longer present in the original variable:
```julia
julia> R, (w, x, y, z) = graded_polynomial_ring(QQ, ["w", "x", "y", "z"]);

julia> Z = R(0); O = R(1); B = [Z Z Z O; w*y w*z-x*y x*z-y^2 Z];

julia> A = transpose(matrix(B));

julia> M = graded_cokernel(A);

julia> PM1 = presentation(M)
0 <---- M <---- R^2 <---- R^4

julia> S = R; R = "foo";

julia> PM1
0 <---- M <---- R^2 <---- R^4			# master
0 <---- M <---- S^2 <---- S^4			# this PR
```

Furthermore, together with https://github.com/thofma/Hecke.jl/pull/1532, add parentheses around non-named rings in presentations/resolutions:
```julia
julia> R = "foo"; S = "foo";
julia> PM1 = presentation(M)
0 <---- M <---- Graded multivariate polynomial ring in 4 variables over QQ^2 <---- Graded multivariate polynomial ring in 4 variables over QQ^4		# master
0 <---- M <---- (Graded multivariate polynomial ring in 4 variables over QQ)^2 <---- (Graded multivariate polynomial ring in 4 variables over QQ)^4		# this PR and https://github.com/thofma/Hecke.jl/pull/1532

### CODE REVIEW

**Comment by HechtiDerLachs** [comment]
Unfortunately, I'm afraid, I'm not the correct person to review or approve this PR. 

I almost never cared about the printing of anything and I don't know how the internals even work. Whether or not things still behave correctly with these changes should probably be evaluated by @wdecker and @jankoboehm .

File: src/Modules/UngradedModules/Presentation.jl [hunk 1]
```diff
@@ -20,6 +20,7 @@ function presentation(SQ::SubquoModule;
   end
 
   # Old code left for debugging
```
fingolfin, line 22: I really don't like this kind of thing, this commented code will just bitrot.

If it is a useful for debugging, how about turning it into an actual method? (Obviously not addressed at you, @lgoettgens)

In any case, personally I'd prefer if it was removed here and instead of commented out.

fingolfin (commented):
Thank you, I am absolutely in favor of this.

However we should indeed coordinate this with @jankoboehm @wdecker @afkafkafk13 @simonbrandhorst and possibly others

afkafkafk13:
> Thank you, I am absolutely in favor of this.
> 
> However we should indeed coordinate this with @jankoboehm @wdecker @afkafkafk13 @simonbrandhorst and possibly others

I have no objections either. But this question should be addressed to @jankoboehm @wdecker who have the broad view on the current state of presentations and free resolutions

fingolfin:
Let's talk about (and hopefully merge) this during triage this week

fingolfin:
Triage (including Wolfram and Janko) decided we can merge this. As to deleting the commented out code: figure who commented it out (@HechtiDerLachs ?) and double check that it is OK to remove.

lgoettgens:
Can we merge this and handle the commented out code separately? I can find multiple occurrences of `# Old code left` in the Module code, all from @HechtiDerLachs according to `git blame` that could get tackled together

fingolfin (approved):

Issue #129, opened 2020-07-12
Cache `symmetric_group(n)`?


fingolfin:
Perhaps we ought to cache the output of `symmetric_group(n)` somehow, seeing as we call it so often. Alternatively, perhaps we can avoid creating it in more cases, resp. make its creation lazy (so we only create it if the user really requests the parent explicitly)

fingolfin:
An appealing way to take care of this would be to use a `WeakValueDicts` -- we could use https://github.com/travigd/WeakValueDicts.jl for this, or roll our own. With an additional "purge_cache" method perhaps (we may want to think about a global way to allow a user to "clear all caches" or "clear some caches")

fieker:
On Fri, Feb 12, 2021 at 07:29:11AM -0800, Max Horn wrote:
> An appealing way to take care of this would be to use a `WeakValueDicts` -- we could use https://github.com/travigd/WeakValueDicts.jl for this, or roll our own. With an additional "purge_cache" method perhaps (we may want to think about a global way to allow a user to "clear all caches" or "clear some caches")
> 
In Hecke/src/Hecke.jl

function print_cache()
  print_cache(find_cache(Nemo))
  print_cache(find_cache(Nemo.Generic))
  print_cache(find_cache(Hecke))
end

function clear_cache()
  clear_cache(find_cache(Nemo))
  clear_cache(find_cache(Nemo.Generic))
  clear_cache(find_cache(Hecke))
end


which basically empties all global Arrays and Dicts

fingolfin:
That's a neat hack, but at least for the purpose discussed here, a `WeakValueDict` is strictly superior, as it will "just work" for many cases, without the user having to know about `clear_cache`; also, clearing *all* caches is a rather brutal operation.

In fact, `clear_cache` can even have a negative impact: if an object is still being referenced, then removing it from the caches won't GC it; and if then we need just that object from the cache, we'll end up allocating a second copy; so in that way, `clear_cache` can even increase memory consumption.

fingolfin:
We discussed this during triage: the plan is this: we cache the `symmetric_group(n)`, but only for the "element constructors" like `perm`, `cperm`, `@perm`

Also we can look into implementing a "finitary $\mathrm{Sym}(\mathbb{N})$ to more closely match GAP, and use *that* as parent for "parentless permutations". And then discuss semantics...

fingolfin:
Resolved in PR #4582

PR #3817, opened 2024-05-31
change `_common_parent_group`

### LINKED ISSUE #894: parents and degrees of PermGroup
```
julia> G = symmetric_group(2)
Sym( [ 1 .. 2 ] )

julia> degree(G)
2

julia> G = symmetric_group(1)
Group(())

julia> degree(G)
0
```
This should be 1.

The next:
```
julia> G = symmetric_group(3)
Sym( [ 1 .. 3 ] )

julia> U = sub(G, [G([1,2,3])])[1]
Group(())
julia> U^U([1,2,3])
Group(())

julia> ans([1,2,3])
ERROR: ArgumentError: the element does not embed in the group
Stacktrace:
 [1] (::PermGroup)(L::Vector{Int64})
   @ Oscar ~/.julia/dev/Oscar/src/Groups/perm.jl:162
 [2] top-level scope
   @ REPL[85]:1
```
so the conjugation looses the degree...

ThomasBreuer:
- define and implement the meaning for the group types,
- add tests for the supported situations

resolves #894

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -253,9 +253,12 @@ end
 #   The operation is allowed whenever the two groups have the same `.G` field,
 #   and the parent is the full automorphism group of that group.
 #
-# - `DirectProductGroup`, `DirectProductGroup`:
-#   The operation is allowed whenever the two groups have the same `.L`
-#   field, and the parent is the direct product of these groups.
+# - `DirectProductGroup`, `DirectProductGroup` and
+#   `SemidirectProductGroup`, `SemidirectProductGroup` and
+#   `WreathProductGroup`, `WreathProductGroup`:
+#   The operation is allowed whenever the two groups have the same `.Xfull`
+#   field, and the parent is the direct/semidirect/wreath product of these
+#   groups.
 #
 # For other types of groups, we throw an exception if the parent groups are
 # not equal and their `GapObj`s are not equal.
@@ -325,8 +328,20 @@ end
 
 function _common_parent_group(x::DirectProductGroup, y::DirectProductGroup)
   x === y && return x
-  @req x.L == y.L "the groups belong to different full groups"
-  return direct_product(x.L)
+  @req x.Xfull === y.Xfull "the groups belong to different full groups"
+  return DirectProductGroup(x.Xfull, x.L, x.Xfull, true)
+end
+
+function _common_parent_group(x::SemidirectProductGroup, y::SemidirectProductGroup)
+  x === y && return x
+  @req x.Xfull === y.Xfull "the groups belong to different full groups"
+  return SemidirectProductGroup{typeof(x.N), typeof(x.H)}(x.Xfull, x.N, x.H, x.f, x.Xfull, true)
+end
+
+function _common_parent_group(x::WreathProductGroup, y::WreathProductGroup)
+  x === y && return x
+  @req x.Xfull === y.Xfull "the groups belong to different full groups"
+  return WreathProductGroup(x.Xfull, x.G, x.H, x.a, x.Xfull, true)
 end
 
 # generic method
@@ -460,6 +475,10 @@ div_right(x::GAPGroupElem, y::GAPGroupElem) = group_element(parent(x), (GapObj(x
 div_left(x::GAPGroupElem, y::GAPGroupElem) = group_element(parent(x), (GapObj(y) \ GapObj(x))::GapObj)
 
 Base.conj(x::GAPGroupElem, y::GAPGroupElem) = group_element(_common_parent_group(parent(x), parent(y)), (GapObj(x) ^ GapObj(y))::GapObj)
+
+# AbstractAlgebra defines `x^y` for group elements of the *same* type only.
+Base.:^(x::GAPGroupElem, y::GAPGroupElem) = Base.conj(x, y)
+
 
 """
     comm(x::GAPGroupElem, y::GAPGroupElem)
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -224,53 +224,64 @@ end
 #
 # - `PermGroup`, `PermGroup`:
 #   The operation is allowed whenever the parents have the same degree,
-#   and the parent is the symmetric group of that degree.
+#   then the parent of the result is the symmetric group of that degree.
 #
 # - `PcGroup`, `PcGroup` and
 #   `FPGroup`, `FPGroup`:
-#   The operation is allowed whenever the parents belong to the same `GapObj`
-#   (in the sense of `==´), and the first of the two groups is taken.
+#   The operation is allowed whenever the parents have the same `GapObj`
+#   (in the sense of `===´),
+#   then the first of the two groups is taken as the parent of the result.
 #
 # - `SubPcGroup`, `SubPcGroup` and
 #   `SubFPGroup`, `SubFPGroup`:
-#   The operation is allowed whenever the parents have the same `full_group`
-#   (in the sense of `==´), and the first of the two groups is taken.
+#   The operation is allowed whenever the `full_group` fields of the parents
+#   have the same `GapObj`
+#   (in the sense of `===´),
+#   then the first of the two groups is taken as the parent of the result.
 #
 # - `SubPcGroup`, `PcGroup` and
 #   `PcGroup`, `SubPcGroup` and
 #   `SubFPGroup`, `FPGroup` and
 #   `FPGroup`, `SubFPGroup`:
 #   The operation is allowed whenever the `full_group` of the `SubPcGroup`
-#   (`SubFPGroup`) is equal to the `PcGroup` (`FPGroup`)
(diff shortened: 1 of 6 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
 
+
+# We allow arithmetic operations between two group elements with
+# *nonidentical parents*.
+# In this case, the parent of the resulting element is set according to
+# the following rules, depending on the types of the parents.
+#
+# - `PermGroup`, `PermGroup`:
+#   The operation is allowed whenever the parents have the same degree,
+#   and the parent is the symmetric group of that degree.
```
fingolfin: Maybe
```suggestion
#   then the new parent is the symmetric group of that degree.
```
or
```suggestion
#   then the parent of the result is the symmetric group of that degree.
```

src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
+#
+# - `PcGroup`, `PcGroup` and
+#   `FPGroup`, `FPGroup`:
+#   The operation is allowed whenever the parents belong to the same `GapObj`
```
fingolfin: I don't understand this. When does a `PcGroup` group "belong" to a `GapObj`?

src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
+# - `DirectProductGroup`, `DirectProductGroup`:
+#   The operation is allowed whenever the two groups have the same `.L`
+#   field, and the parent is the direct product of these groups.
+#
+# For other types of groups, we throw an exception if the parent groups are
+# not equal and their `GapObj`s are not equal.
+#
+# Note that we cannot guarantee that the groups objects in question are
+# *identical*, only that their `GapObj`s are *equal*:
+# On the GAP side, the same group can be created several times,
```
fingolfin: Maybe add that this applies to perm groups, but not to pc and fp groups?

src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
+# the parent of the result).
+#
+function _common_parent_group(x::PermGroup, y::PermGroup)
+  x === y && return x
+  @req degree(x) == degree(y) "the groups have different degrees"
+  return symmetric_group(degree(x))
+end
+
+function _common_parent_group(x::PcGroup, y::PcGroup)
+  (x === y || GapObj(x) == GapObj(y)) && return x
```
fingolfin: Does `==` for `PcGroup` do anything besides compare identity?

If so, wouldn't this be sufficient?
```suggestion
  (GapObj(x) === GapObj(y)) && return x
```

src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
+  throw(ArgumentError("the groups are not compatible"))
+end
+
+function _common_parent_group(x::FPGroup, y::FPGroup)
+  (x === y || GapObj(x) == GapObj(y)) && return x
```
fingolfin: Same questions as for `PCGroup`
```suggestion
  (GapObj(x) === GapObj(y)) && return x
```

src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
+  throw(ArgumentError("the groups are not compatible"))
+end
+
+function _common_parent_group(x::SubPcGroup, y::SubPcGroup)
+  x === y && return x
+  @req x.full_group == y.full_group "the groups belong to different full groups"
+  return x.full_group
```
fingolfin: So this is not type stable then, it sometimes gives a `SubPcGroup` and sometimes a `PcGroup`. Hrm, that's annoying, but I see no immediate way to resolve it, unless we were to make the "full group" deliberately a `SubPcGroup` that is an index 1 subgroup of the `PcGroup`...

src/Groups/GAPGroups.jl
```diff
@@ -216,27 +216,126 @@ function rand_pseudo(G::GAPGroup; radius::Int = 10)
...
+  return y
+end
+
+function _common_parent_group(x::PcGroup, y::SubPcGroup)
+  @req y.full_group == x "the groups belong to different full groups"
+  return x
+end
+
+function _common_parent_group(x::SubFPGroup, y::SubFPGroup)
+  x === y && return x
```
fingolfin, line 319: Same as for `PcGroup` :-/

src/Groups/perm.jl
```diff
@@ -286,7 +286,9 @@ function cperm()
 end
 
 function cperm(L1::AbstractVector{T}, L::AbstractVector{T}...) where T <: IntegerUnion
-  return prod([PermGroupElem(symmetric_group(maximum(y)), GAPWrap.CycleFromList(GAP.Obj([Int(k) for k in y]))) for y in [L1, L...]])
+  cycles = [L1, L...]
+  n = maximum(map(maximum, cycles))
+  return prod([PermGroupElem(symmetric_group(n), GAPWrap.CycleFromList(GAP.Obj([Int(k) for k in y]))) for y in cycles])
```
fingolfin, line 291: Argh, this reminds me I really need to submit a PR for my rewrite of `cperm` and friends.


Also, this method is still inefficient because it creates a fresh `symmetric_group` each time. Either we cache those (see issue #129) or perhaps we should just disallow this without an explicit parent group (well, not immediately to stay compatible with 1.0, but at least we could deprecate this). Thoughts, @fieker ?

fingolfin (commented):
Thank you! Some quick thoughts

src/Groups/GAPGroups.jl
ThomasBreuer: The question is what we want to support.
Currently the following works.
```
julia> g = small_group(24, 12);

julia> gg = GAP.Globals.Group(GAP.Globals.GeneratorsOfGroup(GapObj(g)));

julia> newg = PcGroup(gg);

julia> newg == g
true
```
If we want to achieve that two `PcGroup`s are equal only if their `GapObj`s are identical then we have to change `PcGroup`. Likewise for `FPGroup`s. I do not expect difficulties from such a change.

src/Groups/GAPGroups.jl
ThomasBreuer: (see the comment for `PcGroup`)

src/Groups/GAPGroups.jl
ThomasBreuer: Taking that index 1 subgroups in `PcGroup`s occur already in many places because of these type conventions (the first member in `derived_series(G)` etc.), we can return always a `SubPcGroup`.
Currently the `SubPcGroup` stores the full `PcGroup`, thus fetching this `.full_group` is cheap, compared to wrapping a new `SubPcGroup`. Perhaps the `SubPcGroup` should store an index 1 subgroup instead?

src/Groups/GAPGroups.jl
ThomasBreuer: See the code example about `PcGroup` below.

src/Groups/GAPGroups.jl
ThomasBreuer: O.k., let us say "have the same `GapObj`".

src/Groups/GAPGroups.jl
ThomasBreuer: Hmm, if we want that a `SubPcGroup` stores a `SubPcGroup` as its `full_group` then we have the logical problem that the `SubPcGroup` constructor takes a `GapObj` and calls `SubPcGroup` for its full group, an infinite recursion ...

src/Groups/GAPGroups.jl
```diff
@@ -224,50 +224,64 @@ end
...
 #   `SubFPGroup`, `FPGroup` and
 #   `FPGroup`, `SubFPGroup`:
 #   The operation is allowed whenever the `full_group` of the `SubPcGroup`
-#   (`SubFPGroup`) is equal to the `PcGroup` (`FPGroup`)
-#   (in the sense of `==´), and the `PcGroup` (`FPGroup`) is taken.
+#   (`SubFPGroup`) and the `PcGroup` (`FPGroup`) have the same `GapObj`
+#   (in the sense of `===´),
+#   then the `full_group` of the `SubPcGroup` (`SubFPGroup`) is taken
+#   as the parent of the result.
+#   Note that we take a group of type `SubPcGroup` (`SubFpGroup`).
```
fingolfin, line 251: Perhaps we should explain why (i.e. type stability, e.g. when multiplying elements with identical parent)

fingolfin (approved):

PR #3818, opened 2024-06-03
Book2

fieker:


experimental/GModule/Brueckner.jl
```diff
@@ -304,67 +333,70 @@ function lift(C::GModule, mp::Map)
...
+
+
+    #TODO: not all "chn" yield distinct groups - the factoring by the
+    #      co-boundaries is missing
+    #      not all "epi" are epi, ie. surjective. The part of the thm
+    #      is missing...
+    # (Thm 15, part b & c) (and the weird lemma)
+
+  allG = []
+
```
fingolfin: Avoid code instability?
```suggestion
```

fieker: done - but playing some more

thofma:
Effects only experimental, so let's get this in.

PR #3819, opened 2024-06-03
updates related to polymake 4.12

benlorenz:
- rounding for OscarNumber
- fix for graph edge iterators
- add some type mapping for nodemaps for polymake posets
- delayed loading for graphmaps when loading polymake objects from dicts

todo:
- ~~add wrappers for some poset functionality~~ #3928


lgoettgens:
I think it would make sense to split this PR into two: one regarding the small fixes for polymake 4.12 (which seems to be almost finished) and a separate one for the post functionality.
This would make it easier for other people interested in posets (like e.g. @felix-roehrich) to take part in a discussion there, without holding back the other changes unnecessarily long.
(and that's how we do it with the other cornerstones usually as well, i.e. don't mix new features and adaptions to an update too much to keep everything transparent and comprehensible.)

benlorenz:
> I think it would make sense to split this PR into two: one regarding the small fixes for polymake 4.12 (which seems to be almost finished) and a separate one for the post functionality. This would make it easier for other people interested in posets (like e.g. @felix-roehrich) to take part in a discussion there, without holding back the other changes unnecessarily long. (and that's how we do it with the other cornerstones usually as well, i.e. don't mix new features and adaptions to an update too much to keep everything transparent and comprehensible.)

That makes sense, I will split this up.
_Edit:_ done in #3928

lgoettgens:
The error in CI seems like something real: https://github.com/oscar-system/Oscar.jl/actions/runs/9927077065/job/27421571688?pr=3819#step:9:3937
However, it only occurs on macOS.

benlorenz:
> The error in CI seems like something real: [oscar-system/Oscar.jl/actions/runs/9927077065/job/27421571688?pr=3819#step:9:3937](https://github.com/oscar-system/Oscar.jl/actions/runs/9927077065/job/27421571688?pr=3819#step:9:3937) However, it only occurs on macOS.

Yes, I am aware of that, there are some extra conversion helpers missing for Pair/Tuple types, I need to add them in Polymake.jl. (The reason is that on Linux CxxLong is just an alias for Int64, but on macOS CxxLong is a separate type, and then some default constructions don't match as expected)

antonydellavecchia (approved):

PR #3820, opened 2024-06-03
Fix some minor inconsistencies in the manual

fingolfin:


src/Groups/action.jl
```diff
@@ -1,13 +1,13 @@
-"""
-A *group action* of a group G on a set Ω (from the right) is defined by
-a map μ: Ω × G → Ω that satisfies the compatibility conditions
-μ(μ(x, g), h) = μ(x, g*h) and μ(x, one(G)) == x for all x ∈ Ω.
+raw"""
```
fingolfin: @ThomasBreuer this free standing string duplicates parts of `action.md` but is not attached to any object, so as far as I know, it is not reachable from the REPL. So what is the point of having it around? Am I missing something?

ThomasBreuer: This definition was the starting point of the file. As far as I remember, we did not have `action.md` at that time.
We can remove the string.

src/Rings/mpoly-affine-algebras.jl
```diff
@@ -1481,7 +1481,7 @@ $A\hookrightarrow A_1\times \dots\times A_r=\overline{A}$
 of $A$ as does `normalize(A)`, but return additionally the `delta invariant` of $A$,
 that is, the dimension
 
-$\dim_K(\overline{A}/A)$.
+$\dim_K(\overline{A}/A).$
```
ThomasBreuer, line 1484: Why? I think the dot marks the end of the sentence, it does not belong to the expression in math mode.

fingolfin: Because this is actually rendered as display math (so think of this as `\[A\].` versus `\[A.\]`). Hence in the current version the `.` ends up on a new line after the formula.

joschmitt: Yes, but the `$` are interpreted like `\[`/`\]` in normal latex, so the dot gets printed in a newline.

joschmitt: Funny, we answered in the same moment. :smile:

ThomasBreuer (approved):
Looks good, thanks.

PR #3821, opened 2024-06-03
experimental: Gröbner Walk

ooinaruhugh:
This PR supercedes #3131 and provides some variants of the Gröbner Walk.
Currently, this includes
- Standard walk from Collart, Kalkbrener and Mall (1997)
- Generic walk from Fukuda (2007)
- Perturbed walk from Amrhein, Gloor and Küchlin (1997)

Further things we want to do:
- #3131 included some more variants, we're still in the process of revising those. 
- Also, the code probably needs some tidying up.
- We're currently investigating whether we can improve the performance of this implementation. 
- Tests


experimental/GroebnerWalk/Project.toml
```diff
@@ -0,0 +1,8 @@
+name = "GroebnerWalk"
+uuid = "6dc8777d-2b0c-4de5-9e3c-426c1820f6f9"
+authors = ["Kamillo Ferry <kafe@kafe.dev>", "Francesco Nowell <francesconowell@gmail.com>"]
+version = "1.0.1"
```
fingolfin: I am not sure we can support "sub-projects" like that.

More importantly, I think code with a license deviating from the license of OSCAR overall is a no-go (at least inside of OSCAR, of course a seperate package based on OSCAR can do anything it wants). Especially a viral one like GPL3 -- including this here would risk "tainting" the whole project with that license... Where does this license choice come from? Is this because your code is based on GPL code written and licensed as such by others and hence must be GPL; or is this your own choice?

ooinaruhugh: Regarding the license, we chose GPLv3 because we thought OSCAR is licensed under GPLv3 as well.  This might as well be changed to GPLv3+, that's fine.

Also, regarding "sub-projects", when reworking the code from #3131, I tried to incorporate as many suggestions as possible from there.

That's why we chose to make this an experimental package since 
- then we could develop the code separately from OSCAR and not clutter the main OSCAR module,
- there is a sizeable set of helper functions which might or might not stay, and
- this looked like the appropriate place while we bring the code up to standard.

Ultimately, the goal is to be part of the main OSCAR package and provide functionality similar to `groebner_basis` and its variants. I'm happy to hear about suggestions on how to make this happen.

thofma: The things in `experimental/` are not julia packages themselves, but just submodules living in a separate folder. These submodules are loaded like normal Oscar code in `src/`, the only difference is that they are marked in the documentation as "experimental" (and things in `src/` should not depend on things `experimental/`). They don't have a separate license or a separate Project.toml or UUID.

P.S.: This is described in https://docs.oscar-system.org/dev/Experimental/intro/ (but things might not be as clear as we had hoped, so please let us know if there is something that can be improved).

fingolfin: Yes my apologies for the confusion about the license, it was ne who was  confused about that for a moment ("brain fart") :oops:. Of course OSCAR is GPL 3 *or later* (but this is not clear in the README so we should clarify it). Still a separate license file is quite unusual, as is a separate Project.toml.

ooinaruhugh: > The things in `experimental/` are not julia packages themselves, but just submodules living in a separate folder. These submodules are loaded like normal Oscar code in `src/`, the only difference is that they are marked in the documentation as "experimental" (and things in `src/` should not depend on things `experimental/`). They don't have a separate license or a separate Project.toml or UUID.
> 
> P.S.: This is described in https://docs.oscar-system.org/dev/Experimental/intro/ (but things might not be as clear as we had hoped, so please let us know if there is something that can be improved).

Ah, okay. I understood it that in theory you would just take a package that is to be moved to OSCAR and just plop it down in `experimental/` (and further down the road get the submodule absorbed in OSCAR). It wasn't clear that a separate `Project.toml` is specifically discouraged.

ooinaruhugh:
I talked to @afkafkafk13 recently and we decided that the progress up to now is ready for review. New features will come with a later PR, of course corrections to the current code will be added.

experimental/GroebnerWalk/src/GroebnerWalk.jl
```diff
@@ -0,0 +1,22 @@
...
+include("common.jl")
+
+include("standard_walk.jl")
+include("generic_walk.jl")
+include("perturbed_walk.jl")
+
+import Oscar: weight_ordering, ZZRingElem, MonomialOrdering, ZZMatrix, IdealGens
+import Oscar.Orderings: MatrixOrdering, _support_indices
+
+export groebner_walk
```
lgoettgens:
```suggestion
import Oscar: weight_ordering, ZZRingElem, MonomialOrdering, ZZMatrix, IdealGens
import Oscar.Orderings: MatrixOrdering, _support_indices

export groebner_walk

include("markedGB.jl")
include("common.jl")

include("standard_walk.jl")
include("generic_walk.jl")
include("perturbed_walk.jl")
```
You usually want to put imports/exports before the content of the module. Otherwise, the imports are not available yet when you try to use them

experimental/GroebnerWalk/src/GroebnerWalk.jl
```diff
@@ -0,0 +1,22 @@
...
+
+function __init__()
+    add_verbosity_scope(:groebner_walk)
+
+    set_verbosity_level(:groebner_walk, 0)
```
lgoettgens:
```suggestion
```
this is not needed, see e.g. https://github.com/oscar-system/Oscar.jl/blob/344710c8292b58fd42fca71c67278e1a6d20fdbf/src/Oscar.jl#L143-L179 for the place where the other scopes are introduced

experimental/GroebnerWalk/test/runtests.jl
```diff
@@ -0,0 +1 @@
+include("standard_walk.jl")
```
lgoettgens: this file is redundant, as Oscar automatically runs all files from the test directory when running tests

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,170 @@
...
+  end
+
+  Gb = groebner_basis(I; ordering=start, complete_reduction=true)
+  Gb = walk(Gb)
+
+  return Oscar.IdealGens(Gb, target; isGB=true)
+end
+
+#TODO docstring
+function weight_ordering(w::Vector{ZZRingElem}, o::MonomialOrdering)
```
lgoettgens: This function is an enhancement to Oscar itself and does not need to get put inside `experimental`.
Please put it below https://github.com/oscar-system/Oscar.jl/blob/d1075791e3baadea53b705cf2b70040e8548428e/src/Rings/orderings.jl#L1283 or, even better, just change the signature to `weight_ordering(w::Vector{<:IntegerUnion}, o::MonomialOrdering)` to make one method work for both types of inputs

experimental/GroebnerWalk/README.md
```diff
@@ -0,0 +1,42 @@
...
+using Oscar
+
+R, (x,y) = QQ[:x, :y]                  # define ring ...
+I = ideal([y^4+ x^3-x^2+x,x^4])        # ... and ideal
+```
+
+Then, we can pass the ideal to `groebner_walk` to calculate the Gröbner basis.
+Here, we want a Gröbner basis with respect to the lexicographic ordering on `R`.
+```julia
+using GroebnerWalk
```
lgoettgens: This example seems to be a bit outdated

experimental/GroebnerWalk/benchmark/agk.jl
```diff
@@ -0,0 +1,16 @@
...
+    o2 = weight_ordering([0,0,0,1,1], degrevlex(R))
+
+    F = [
+        u + u^2 - 2*v - 2*u^2*v + 2*u*v^2 - x,
+        -6*u + 2*v + v^2 - 5*v^3 + 2*u*v^2 - 4*u^2*v^2 - y,
+        -2 + 2*u^2 + 6*v - 3*u^2*v^2 - z
+    ]
+
+    return ideal(F), o2, o1
+end
```
lgoettgens:
```suggestion
end

```
all files in Oscar should end with a newline

experimental/GroebnerWalk/test/runtests.jl
ooinaruhugh: runtests.jl is deleted

experimental/GroebnerWalk/test/standard_walk.jl
```diff
@@ -0,0 +1,23 @@
+R1, (x,y,z) = polynomial_ring(QQ, ["x", "y", "z"])
```
lgoettgens: Could you move all of these definitiona into your `@testset`? Background to this is that `@testset` introduced a new scope, and without this change you pollute the global namespace, and thus maybe change results of tests run after your file

experimental/GroebnerWalk/benchmark/cyclic.jl
```diff
@@ -0,0 +1,78 @@
...
+    
+        z0*z1*z2*z3*z4*z5*z6 + z1*z2*z3*z4*z5*z6*z7 + z2*z3*z4*z5*z6*z7*z0
+        + z3*z4*z5*z6*z7*z0*z1 + z4*z5*z6*z7*z0*z1*z2 + z5*z6*z7*z0*z1*z2*z3
+        + z6*z7*z0*z1*z2*z3*z4 + z7*z0*z1*z2*z3*z4*z5,
+
+        z0*z1*z2*z3*z4*z5*z6*z7 - 1
+    ]
+
+    return ideal(F), lex(R), default_ordering(R)
+end
```
lgoettgens:
```suggestion
end

```
Same issue for many more of your files

lgoettgens: Please check all other files as well

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,161 @@
...
+2 -> x + y^12 - y^8 + y^4
+with respect to the ordering
+matrix_ordering([x, y], [1 0; 0 1])
+```
+"""
+function groebner_walk(
+  I::MPolyIdeal, 
+  target::MonomialOrdering = lex(base_ring(I)),
+  start::MonomialOrdering = default_ordering(base_ring(I));
+  perturbation_degree = length(gens(base_ring(I))), # meaning, n=#gens(R)
```
lgoettgens:
```suggestion
  perturbation_degree = ngens(base_ring(I)), # meaning, n=#gens(R)
```

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,161 @@
...
+end
+
+@doc raw"""
+    same_cone(G::Oscar.IdealGens, T::MonomialOrdering)
+
+Checks whether the leading terms of G with respect to the matrix ordering given by T agree 
+with the leading terms of G with respect to the current ordering.
+This means they are in the same cone of the Groebner fan. (cf. Lemma 2.2, Collart, Kalkbrener and Mall 1997)
+"""
+same_cone(G::Oscar.IdealGens, T::MonomialOrdering) = all(leading_term.(G; ordering=T) .== leading_term.(G; ordering=ordering(G)))
```
lgoettgens: since this returns a bool, I personally would find it more intuitive if it was named something like `is_same_cone` or `defines_same_cone`.

experimental/GroebnerWalk/README.md
```diff
@@ -0,0 +1,39 @@
...
+```julia
+R, (x,y) = QQ[:x, :y]                  # define ring ...
+I = ideal([y^4+ x^3-x^2+x,x^4])        # ... and ideal
+```
+
+Then, we can pass the ideal to `groebner_walk` to calculate the Gröbner basis.
+Here, we want a Gröbner basis with respect to the lexicographic ordering on `R`.
+```julia
+using Oscar
+using Oscar.GroebnerWalk
```
lgoettgens:
```suggestion
```
you shouldn't need this second using line as `groebner_walk` is exported (at least after my other comments below)

experimental/GroebnerWalk/benchmark/benchmarks.jl
```diff
@@ -0,0 +1,60 @@
+using Oscar
+using GroebnerWalk
```
lgoettgens: this won't work anymore
(and there are many more occurrences of `using GroebnerWalk` in your PR that need to be adapted)

experimental/GroebnerWalk/results.csv
```diff
@@ -0,0 +1,30 @@
...
+katsura6-QQ,1.003294291,,35.098429459,
+simple,1.422464625,0.274453459,,2.475e-5
+agk4-QQ,11.076574625,55.562715875,,543.319831
+agk4-Fp,3.492181375,51.427781875,,0.094963834
+tran3.3-QQ,0.640738125,24.688099959,,2323.13229425
+tran3.3-Fp,0.254231917,14.402583584,,0.003486333
+newellp1-QQ,51.5992475,1529.494271334,,811.051479458
+newellp1-Fp,9.365612875,1140.743606416,,0.165012125
+simple,1.149968916,0.25867925,,2.5334e-5
+katsura6-QQ,0.813826709,35.404766583,,
```
lgoettgens, line 30: I don't really see the benefit of having this file in Oscar. It seems more like this is your personal benchmark result that's used in some paper.

experimental/GroebnerWalk/src/GroebnerWalk.jl
```diff
@@ -0,0 +1,19 @@
...
+include("generic_walk.jl")
+include("perturbed_walk.jl")
+
+export groebner_walk
+
+function __init__()
+    add_verbosity_scope(:groebner_walk)
+end
+
+end
```
lgoettgens, line 35:
```suggestion
end

export groebner_walk
```
If you want to have functions available for Oscar users, you need to export them twice (once inside the module, and once at the bottom of the file).
Once you do this, all `using Oscar.GroebnerWalk` should be unnecessary.

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,161 @@
...
+    groebner_walk(
+      I::MPolyIdeal, 
+      target::MonomialOrdering = lex(base_ring(I)),
+      start::MonomialOrdering = default_ordering(base_ring(I));
+      perturbation_degree = length(gens(base_ring(I))),
+      algorithm::Symbol = :standard
+    )
+
+Compute a reduced Groebner basis w.r.t. to a monomial ordering by converting it using the Groebner Walk.
+The Groebner Walk is proposed by Collart, Kalkbrener & Mall (1997).
```
lgoettgens: Please put your references into `doc/oscar_references.bib` and cite them here using `[BIBKEY](@cite)`.
For more details, see https://docs.oscar-system.org/dev/DeveloperDocumentation/documentation/#Updating-the-bibliography

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,161 @@
...
+One can choose a strategy of:
+- Standard Walk (:standard) computes the Walk like as presented in Cox, Little & O'Shea (2005).
+- Generic Walk (:generic) computes the Walk as presented in Fukuda, Jensen, Lauritzen & Thomas (2005).
+- Perturbed Walk (:perturbed, with p = degree of the perturbation) computes the Walk as presented in Amrhein, Gloor & Küchlin (1997).
```
lgoettgens: this looks a bit redundant to the algorithm keyword argument explanation below. Maybe these two lists can be merged?

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,161 @@
...
+julia> R,(x,y) = polynomial_ring(QQ, ["x","y"]);
+
+julia> I = ideal([y^4+ x^3-x^2+x,x^4]);
+
+julia> groebner_walk(I, lex(R))
+Gröbner basis with elements
+1 -> x + y^12 - y^8 + y^4 r
+2 -> y^16
+with respect to the ordering
+lex([x, y])
```
lgoettgens: These doctest outputs are a bit outdated. You might need to re-run them and update the printing

experimental/GroebnerWalk/src/common.jl
```diff
@@ -0,0 +1,161 @@
...
+
+julia> groebner_walk(I, lex(R); algorithm=:perturbed)
+Gröbner basis with elements
+1 -> x + y^12 - y^8 + y^4
+2 -> y^16
+with respect to the ordering
+lex([x, y])
+
+julia> set_verbosity_level(:groebner_walk, 1);
```
lgoettgens, line 45:
```suggestion
julia> set_verbosity_level(:groebner_walk, 1);

```
empty line missing

experimental/GroebnerWalk/src/generic_walk.jl
```diff
@@ -0,0 +1,245 @@
...
+    difference_lead_tail(MG::MarkedGroebnerBasis)
+
+Computes $a - b$ for $a$ a leading exponent and $b$ in the tail of some $g\in MG$. 
+"""
+function difference_lead_tail(MG::MarkedGroebnerBasis)
+  (G,Lm) = gens(MG), markings(MG)
+  lead_exp = leading_exponent_vector.(Lm)
+  
+  v = map(zip(lead_exp, exponents.(G))) do (l,T)
+    Ref(l) .- T
```
lgoettgens, line 72: what's the purpose of using `Ref` here? I unfortunately don't see the issue you are trying to work around to think about more julian or more efficient ways to do it.

experimental/GroebnerWalk/src/generic_walk.jl
```diff
@@ -0,0 +1,245 @@
...
+
+@doc raw"""
+    facet_initials(MG::MarkedGroebnerBasis, v::Vector{ZZRingElem})
+
+Given a marked Gröbner basis `MG` and a facet normal `v`, computes the Gröbner basis of initial forms 
+of `MG` by truncating to all bounding vectors parallel to `v`.
+"""
+function facet_initials(MG::MarkedGroebnerBasis, v::Vector{ZZRingElem})
+  R = base_ring(MG)
+  inwG = Vector{MPolyRingElem}()
```
lgoettgens:
```suggestion
  inwG = elem_type(R)[]
```
this initializes the vector with the concrete mpolynomial type, instead of using the abstract mpoly type (that doesn't know about it's coefficient type)

experimental/GroebnerWalk/src/generic_walk.jl
```diff
@@ -0,0 +1,245 @@
...
+
+@doc raw"""
+    lift_generic(MG::MarkedGroebnerBasis, H::MarkedGroebnerBasis)
+
+Given a marked Gröbner basis `MG` generating an ideal $I$ and a reduced marked Gröbner basis `H` of initial forms,
+lift H to a marked Gröbner basis of I (with unknown ordering) by subtracting initial forms according to Fukuda, 2007.
+"""
+function lift_generic(MG::MarkedGroebnerBasis, H::MarkedGroebnerBasis)
+  return map(1:length(H.gens)) do i
+      H[i] - normal_form(H[i], MG)
```
lgoettgens:
```suggestion
    H[i] - normal_form(H[i], MG)
```

experimental/GroebnerWalk/src/generic_walk.jl
```diff
@@ -0,0 +1,245 @@
...
+    less_than_zero(M::ZZMatrix, v::Vector{ZZRingElem})
+
+Checks whether  $Mv <_{\mathrm{lex}} 0$.
+"""
+function less_than_zero(M::ZZMatrix, v::Vector{ZZRingElem})
+  if is_zero(v)
+    return false
+  else
+    i = 1
+  end
```
lgoettgens:
```suggestion
  end
  i = 1
```

experimental/GroebnerWalk/src/markedGB.jl
```diff
@@ -0,0 +1,121 @@
...
+import Base: length, getindex, setindex!
+
+#constructs a "markedGB" consisting of two lists 
+#gens, a vector of generators of a Gröbner basis
+#markings, a vector "Markings" of G, corresponding to leading terms w.r.t some monomial ordering
+#TODO: maybe initialize s.t. G is reduced/monic? 
+
+struct MarkedGroebnerBasis
+    gens::Vector{<:MPolyRingElem}
+    markings::Vector{<:MPolyRingElem}
```
lgoettgens: do both of these vectors contain elements over the same mpoly ring? If yes, it would be very beneficial to add a type parameter to `MarkedGroebnerBasis` and then use this to make these fields of type `Vector{T}`. This helps julia very much to infer types and make the resulting code more efficient

experimental/GroebnerWalk/src/markedGB.jl
```diff
@@ -0,0 +1,121 @@
...
+
+    function MarkedGroebnerBasis(gens::Vector{<:MPolyRingElem}, markings::Vector{<:MPolyRingElem})
+        if length(gens) != length(markings)
+            throw(ArgumentError("Inputs are of different length"))
+        else
```
lgoettgens: There is `@req` to make easier input sanitation (see https://docs.oscar-system.org/dev/Hecke/manual/developer/macros/#@req)

ooinaruhugh:
The test for our code fail because the function that we export doesn't seem to be accessible. Looking at the other experimental packages wasn't really enlightening, because there the package is not imported and functions are "just used".

experimental/GroebnerWalk/src/markedGB.jl
```diff
@@ -0,0 +1,121 @@
...
+            if 0 in gens
+                throw(ArgumentError("Gröbner basis contains the zero polynomial"))
+            end
+            new(gens, markings) 
+        end
+    end
+end
+
+base_ring(G::MarkedGroebnerBasis) = parent(first(G.gens))
+length(G::MarkedGroebnerBasis) = length(gens(G))
```
lgoettgens:
```suggestion
length(G::MarkedGroebnerBasis) = ngens(G)
ngens(G::MarkedGroebnerBasis) = length(gens(G))
```
everything implementing `gens` should implement `ngens` as well

experimental/GroebnerWalk/src/markedGB.jl
```diff
@@ -0,0 +1,121 @@
...
+        else
+            if 0 in gens
+                throw(ArgumentError("Gröbner basis contains the zero polynomial"))
+            end
+            new(gens, markings) 
+        end
+    end
+end
+
+base_ring(G::MarkedGroebnerBasis) = parent(first(G.gens))
```
lgoettgens, line 22:
```suggestion
base_ring(G::MarkedGroebnerBasis) = parent(first(G.gens))
```
how do you make sure that `G.gens` has at least one element?

lgoettgens (commented):
some more misc comments.
I don't have any knowledge about the involved mathematics, thus these are only comments about efficient of your code and some consistency with the rest of Oscar

lgoettgens:
> The test for our code fail because the function that we export doesn't seem to be accessible. Looking at the other experimental packages wasn't really enlightening, because there the package is not imported and functions are "just used".

This is indeed the standard in Oscar. submodules are not expected to be used with `using Oscar.submodule`, but you should add double exports instead (see https://github.com/oscar-system/Oscar.jl/pull/3821#discussion_r1688231270)

experimental/GroebnerWalk/benchmark/agk.jl
```diff
@@ -0,0 +1,18 @@
+<<<<<<< HEAD
```
lgoettgens:
```suggestion
```
some obsolete merge artifact

experimental/GroebnerWalk/results.csv
afkafkafk13: While this is in experimental, I would keep that file. It can certainly help to detect regression, because it incorporates information on standardized examples.

The question, on how to keep or drop this data later on, when this package moves to src, is not relevant for the moment.

experimental/GroebnerWalk/benchmark/benchmarks.jl
ederc, line 1: Can one remove this empty file?

experimental/GroebnerWalk/benchmark/agk.jl
ederc, line 1: Concerning this benchmark and the others below: There is `src/Rings/special_ideals.jl` in which someone already has added the `katsura-n` examples. Maybe we can add your benchmarks there and remove these different files, I assume your benchmarks are also of interest to other developers/users.

experimental/GroebnerWalk/docs/doc.main
ederc, line 1: We should have a short, but still correct doc information about how to call your GBwalk functionality.

experimental/GroebnerWalk/benchmark/benchmarks.jl
ooinaruhugh: When adding an empty line at the end of some files, I might have accidentally blanked some of those. I think missed this one when fixing that.

experimental/GroebnerWalk/src/generic_walk.jl
ederc: This should also be changed in `autoreduce()`.

experimental/GroebnerWalk/src/generic_walk.jl
ooinaruhugh: `l` is a Vector and `T` is a Vector of Vector. Since I want to subtract every element from `T` from `l`, I need to fix `l` for broadcasting with `Ref`.

ooinaruhugh:
The recent forced pushes come from me messing up the exchange between separate repository and this PR. I'm working on getting the suggested edits from here in there.

I do see the appeal of small PRs now...

experimental/GroebnerWalk/src/markedGB.jl
ooinaruhugh: Not at all, but we were implementing this similarly to `IdealGens`.

Also, the reason we had to do this is pretty much the need for a version where markings can be specified, instead of an ordering.

experimental/GroebnerWalk/benchmark/agk.jl
ooinaruhugh: I started adding the ideals to a `special-ideals.jl` in the experimental module. I will add the others with references, as soon as this is done I will move this to `src/Rings/special_ideals.jl` (in a future PR)

fingolfin:
We just discussed this during triage. @ederc is fine with the "big picture" (we can improve details later). @afkafkafk13 is as well happy with it, but would like to have another look before approving.

Since this is big and has been around, it indeed seems sensible to me to merged this rather sooner and fine tune it later, but I'll leave it to the experts to make the final call.

ooinaruhugh:
Yeah, I agree with that. As far as I can tell, I addressed most of the previous remarks and the implemented algorithms do work correctly. I would also prefer to do any additions in a future PR.

afkafkafk13 (approved):

PR #3822, opened 2024-06-03
Added more attributes including gauge algebras

emikelsons:
- Added torsion sections to the attributes
- Added gauge algebras for almost all models including parametric models
- Fixed some arxiv ids(removed v2)
- Changed model information so that the gauge algebra constructor can parse it.


emikelsons:
@HereAround @apturner

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -1018,3 +1018,31 @@ function zero_section(m::AbstractFTheoryModel)
...
+
+```jldoctest
+julia> t = literature_model(arxiv_id = "1408.4808", equation = "3.190", type = "hypersurface")
+Assuming that the first row of the given grading is the grading under Kbar
+
+Hypersurface model over a not fully specified base
+
+julia> gauge_algebra(t)
+5-element Vector{Any}:
+ Special linear Lie algebra of degree 2 over QQBar
```
fingolfin: So a "gauge algebra" is a list of five Lie algebras? :-)

experimental/FTheoryTools/src/AbstractFTheoryModels/methods.jl
```diff
@@ -624,6 +630,40 @@ function set_zero_section(m::AbstractFTheoryModel, desired_value::Vector{String}
   set_attribute!(m, :zero_section => [f(eval_poly(l, R)) for l in desired_value])
 end
 
+function set_gauge_algebra(m::AbstractFTheoryModel, algebras::Vector{String})
+gauge_algebras = [];
```
fingolfin: Please indent the content of your functions, and avoiding trailing semicolons.

Also, initializing this to `[]` results in a `Vector{Any}` -- but perhaps a stricter type can be used?

experimental/FTheoryTools/src/AbstractFTheoryModels/methods.jl
```diff
@@ -624,6 +630,40 @@ function set_zero_section(m::AbstractFTheoryModel, desired_value::Vector{String}
...
+C = algebraic_closure(QQ);
+for i in 1:length(algebras)
+    g = algebras[i];
```
fingolfin: Do you use `i` in here? If not just do
```suggestion
for g in algebras
```

fingolfin (commented):
Just a few quick comments :-)

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
emikelsons: As far as I am aware, there is no way to form a direct sum yet and so this is really placeholder/workaround so that we can specify the action of the gauge group after information about the discrete factors is added.

HereAround: I mentioned this to @lgoettgens on slack, so we can discuss more.

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -1018,3 +1018,31 @@ function zero_section(m::AbstractFTheoryModel)
...
+
+Return the gauge algebra of the given model.
+If no gauge algebra is known, an error is raised.
+This information is typically available for all models, however.
+
+```jldoctest
+julia> t = literature_model(arxiv_id = "1408.4808", equation = "3.190", type = "hypersurface")
+Assuming that the first row of the given grading is the grading under Kbar
+
+Hypersurface model over a not fully specified base
```
lgoettgens, line 1034: there is some weird looking whitespace in this output (the empty line). is this intended?

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
lgoettgens: once there is a construction of direct sums of Lie algebras, this can get adapted :)

experimental/FTheoryTools/src/AbstractFTheoryModels/methods.jl
```diff
@@ -624,6 +630,39 @@ function set_zero_section(m::AbstractFTheoryModel, desired_value::Vector{String}
...
+C = algebraic_closure(QQ);
+for g in algebras
+    if g == "0"
+      continue
+    end
+    if g == "u(1)"
+      algebra = lie_algebra(C,1,[C(1im)*identity_matrix(C,1)],["i"]);
+      push!(gauge_algebras, algebra);
+    elseif g[1:2] == "su"
+      algebra = special_linear_lie_algebra(C,eval(Meta.parse(g[4:end-1])));
```
lgoettgens:
```suggestion
      algebra = special_linear_lie_algebra(C,parse(Int, g[4:end-1]));
```
If you just expect integers there and not mathematical expressions, this is wayyy safer as you don't allow arbitrary code execution.

emikelsons: This will sometimes contain mathematical expressions as well

emikelsons: Along the lines of "2*5+1"

HereAround: To elaborate more:

The `FTheoryTools` database contains "instructions" for infinite family of geometries. One such family has "instructions" that allow to create one geometry for each non-negative integer k. The k-th geometry in this family comes associated with the Lie algebra `su(2k+1)`.

The above line should 1) replace the character/string `k` by a user given integer value (say 5), 2) evaluate the resulting mathematical expression (2*5+1) to an integer (11) and 3) pass that integer to the Lie algebra constructor.

Suggestions for this @lgoettgens ?

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
HereAround: The empty line is intended.

My eyes are missing the full stop after `Kbar`. If I recall correctly, this is a convention across `OSCAR`.

(@apturner I wonder if we want to switch off the warning/information about the Kbar grading sometimes. E.g. if the model is created from the database. Thoughts?)

HereAround:
The failure in the doctests looks non-critical:
* Previous printout included: "SU(2k+1)",
* New printout includes (for k = 5): "SU(2*5+1)".

Similar to the above comment (evaluating this expression for a user specified integer k to an integer and then passing that integer to the Lie algebra constructor), it would be great to evaluate this expression further before printing it. So that maybe it could become

* Yet new printout includes (for k = 5): "SU(11)".

@lgoettgens do you maybe have a suggestion?

(If I recall correctly, @apturner and I accepted this during the initial setup of `FTheoryTools`. But it seems about time to improve this.)

experimental/FTheoryTools/src/AbstractFTheoryModels/methods.jl
emikelsons: This is now of the form parse(Int... as the eval(Meta.parse()) is now done elsewhere

thofma:
I would like to second @lgoettgens concerns about blindly calling `eval` on user supplied input. If it is a polynomial in `k`, why not parse it as a polynomial in `k` and evaluate that?

emikelsons:
> I would like to second @lgoettgens concerns about blindly calling `eval` on user supplied input. If it is a polynomial in `k`, why not parse it as a polynomial in `k` and evaluate that?

What are the other options to evaluate this in Oscar? It will be an arithmetic expression as 'k' will have already been substituted for some integer

thofma:
Can you clarify what exactly must be evaluated? Somewhere it was mentioned that one wants to evaluate `"2k + 1"` at some value of `k`?

emikelsons:
Well currently the user inputted value for 'k' is first substituted into the expression and then the expression to be evaluated/parsed later becomes something like '2*5 + 1', if the original k was say 5.

thofma:
Then you can still use `eval_poly`:
```
julia> Oscar.eval_poly("2*5 + 1", polynomial_ring(ZZ, Symbol[])[1])
11
```

experimental/FTheoryTools/src/LiteratureModels/constructors.jl
```diff
@@ -313,6 +313,8 @@ function literature_model(model_dict::Dict{String, Any}; model_parameters::Dict{
     
     for (key, val) in model_parameters
       map!(x -> nested_string_map(s -> replace(s, key => string(val)), x), values(model_dict["model_data"]))
+      map!(x -> nested_string_map(s -> replace(s, key => string(val)), x), values(model_dict["model_descriptors"]))
+      map!(x -> nested_string_map(s -> replace(s, r"\(([^)]+)\)" => dim -> string("(", Oscar.eval_poly(string.(match(r"\(([^)]+)\)", dim).captures[1]), polynomial_ring(ZZ, Symbol[])[1]),")")), x), values(model_dict["model_descriptors"]))
```
thofma:
```suggestion
      map!(x -> nested_string_map(s -> replace(s, r"\(([^)]+)\)" => dim -> string("(", Oscar.eval_poly(string.(match(r"\(([^)]+)\)", dim).captures[1]), ZZ),")")), x), values(model_dict["model_descriptors"]))
```
It noticed that one can actually do `Oscar.eval_poly("2*2+1", ZZ)`.

experimental/FTheoryTools/src/LiteratureModels/Models/model1408_4808-11-WSF.json
```diff
@@ -41,8 +41,8 @@
     "model_descriptors": {
         "type": "weierstrass",
         "description": "F-theory weierstrass model dual to hypersurface model with fiber ambient space F_11",
-        "gauge_algebra": ["su3","su2","u1"],
-        "global_gauge_quotients": [["Z3", [1, 3]], ["Z2", [2, 3]]]
+        "gauge_algebra": ["su(3)","su(2)","u(1)"],
+        "global_gauge_quotients": [["diagonal_matrix(root_of_unity(C,3),3)"], ["-identity_matrix(C,2)"], ["diagonal_matrix(root_of_unity(C,6,-1),1)"]]
```
HereAround, line 45: I suppose this another place holder? For I should have expected some groups or group homomorphisms here. Could you state what is currently missing in OSCAR? Matrix Lie groups perse, or rather "just" some necessary homomorphisms?

emikelsons: In general Matrix Lie groups over the complex numbers are missing. This is an attempted workaround due to that fact. As I understand here for each component of the gauge algebra I've listed the elements of the center that we are quotienting by to replicate the corresponding action of the gauge quotient on the center of the lie algebra. Together this should be enough to reconstruct the gauge group, but I've not yet tried.

HereAround: I see. Thank you. We should document this, so that we still understand in a year or two what the meaning was.  And why (a.k.a. what feature is missing). Likewise, I suggest a comment why the gauge algebra is currently a vector of algebras rather than their direct sum.

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
apturner: I think it is fine to turn off the warning about Kbar for the literature models, since the users don't provide the grading anyway

experimental/FTheoryTools/src/LiteratureModels/Models/model1408_4808-11-WSF.json
apturner: This was the notation that Mikelis and I decided on when we last spoke. The global gauge quotient is given explicitly as an element of the center of the simply connected Lie *group* corresponding to the Lie algebra, so this does unambiguously specify the gauge group.

As for the direct sum, it sounds like that may be on the way. @lgoettgens mentioned that he might have an old branch where he previously did some work toward implementing this.

HereAround: That is all fine (for now). Even if we do not want to state that the direct sum is currently WIP (why? can be changed once it is there), it would not hurt to explain for outsiders/users how to read the global gauge quotients, correct?

apturner: Yes, documenting the notation for the quotients seems wise. As for the WIP direct sum, that seems like something for the comments? I'm fine mentioning it in the documentation though if you prefer.

HereAround: Yeah, that is probably the way to go. (I don't want @lgoettgens to feel pressured. Of course we are looking forward to this feature. But then, it is not super urgent.)

emikelsons: Where should I place a comment about the notation for the quotients?

HereAround: I believe we want a method `global_gauge_quotients(m::AbstractFTheoryModel)`, similar to `gauge_algebra(m::AbstractFTheoryModel)`. Just as for `gauge_algebra(m::AbstractFTheoryModel)`, you could provide details of the implementation in the doc string/the example of this function.

Btw., those doc strings need to be added to the documentation. That is, you need to modify the corresponding files in the `FTheoryTools/docs` subfolder. In the case at hand, take a look at `Oscar.jl/experimental/FTheoryTools/docs/src/generalities.md`.

HereAround:
Thank you @emikelsons . Looks good to me.

HereAround (approved):

PR #3823, opened 2024-06-04
Small changes to experimental docs

joschmitt:
* The introduction to `Experimental` mentioned the `PlaneCurves` directory that was removed in #3179.
* I replaced the word "package" by "project". I always found it weird that we talk about experimental packages because they are not packages in the Julia sense (or behave like packages in GAP for example) and apparently it leads to confusion as in #3821.


afkafkafk13 (approved):
This clarifies this point for external contributors. Thank you.

lkastner (approved):

PR #3827, opened 2024-06-04
Fix `is_smooth` for projective space

paemurru:
Previously, taking `base_ring` from a quotient ring $S = R/I$ gave the ring $R$, but if $S$ was not a quotient ring, then it gave the base field of $S$, introducing an error in case the argument of `is_smooth` is the projective space.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
@@ -53,7 +53,8 @@ function _jacobian_criterion(X::CoveredScheme{<:Field})
 
   dec_info = decomposition_info(default_covering(X))
   for (V, fs) in dec_info
-    R = base_ring(OO(V))
+    R_quotient = OO(V)
+    typeof(R_quotient)<:MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
     I = saturated_ideal(defining_ideal(V))
     mat = jacobian_matrix(R, gens(I))
     sing_locus = ideal(R, fs) + ideal(R, minors(mat, codim(V)))
--- a/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
@@ -54,7 +54,7 @@ function _jacobian_criterion(X::CoveredScheme{<:Field})
   dec_info = decomposition_info(default_covering(X))
   for (V, fs) in dec_info
     R_quotient = OO(V)
-    typeof(R_quotient)<:MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
+    R_quotient isa MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
     I = saturated_ideal(defining_ideal(V))
     mat = jacobian_matrix(R, gens(I))
     sing_locus = ideal(R, fs) + ideal(R, minors(mat, codim(V)))
--- a/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
+++ b/src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
@@ -54,7 +54,7 @@ function _jacobian_criterion(X::CoveredScheme{<:Field})
   dec_info = decomposition_info(default_covering(X))
   for (V, fs) in dec_info
     R_quotient = OO(V)
-    R_quotient isa MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
+    R = R_quotient isa MPolyRing ? R_quotient : base_ring(R_quotient)
     I = saturated_ideal(defining_ideal(V))
     mat = jacobian_matrix(R, gens(I))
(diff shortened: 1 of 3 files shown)


src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -53,7 +53,8 @@ function _jacobian_criterion(X::CoveredScheme{<:Field})
 
   dec_info = decomposition_info(default_covering(X))
   for (V, fs) in dec_info
-    R = base_ring(OO(V))
+    R_quotient = OO(V)
+    typeof(R_quotient)<:MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
```
lgoettgens:
```suggestion
    R_quotient isa MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
```
small nitpick: this is the preferred way in julia land if I remember correctly

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
```diff
@@ -53,7 +53,8 @@ function _jacobian_criterion(X::CoveredScheme{<:Field})
 
   dec_info = decomposition_info(default_covering(X))
   for (V, fs) in dec_info
-    R = base_ring(OO(V))
+    R_quotient = OO(V)
+    R_quotient isa MPolyQuoRing ? R = base_ring(R_quotient) : R = R_quotient
```
afkafkafk13: Are you 100% sure that you never see an MPolyLocalizedRing or an MPolyQuoLocalizedRing here? This is the general case of covered schemes here, not the projective scheme case. We could see for instance a complement of a hypersurface as a chart (which gives the right answer for base_ring, by the way). 

I would do this fix the other way around: 

R_quotient isa MPolyRing ? R = R_quotient : R = base_ring(R_quotient)

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Properties.jl
```diff
@@ -17,7 +17,7 @@ end
 end
 
 @doc raw"""
-    is_smooth(P::AbsProjectiveScheme; algorithm=:default) -> Bool
+    is_smooth(P::AbsProjectiveScheme{<:Any, <:MPolyQuoRing}; algorithm=:default) -> Bool
```
afkafkafk13: We could rather catch the easy corner case than exclude it in the docs. 
It is better, if we have the compatibility functions to do this in one place (what you do below anyway), than several users doing it each by themselves after reading this in the docs.

afkafkafk13 (commented):
Thank you. This bug-fix is really important.

I have a few comments on details.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Properties.jl
simonbrandhorst: A solution could be the following: Define

`is_smooth(P::AbsProjectiveScheme; algorithm=:default)`
and document just this method. Then have the following two methods
`is_smooth(P::AbsProjectiveScheme{<:any, <:MPolyQuoRing}; algorithm=:default)`
`is_smooth(P::AbsProjectiveScheme{<:any, <:MPolyRing}; algorithm=:default)`

perhaps the `any` needs to be replaced by `Field` ? ... I have little experience with algebraic geometry over general rings ... best consult a textbook to be sure.

afkafkafk13: We already have is_smooth for MPolyRing below in the new code. Your suggestion is the way to go.

Over general rings, smooth means Jacobi-criterion satisfied and regular means dim = edim. Please keep in mind that regular still allows fibre-singularities e.g. over ZZ.

src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl
paemurru: I am not entirely sure how the algorithms would work for localized rings, in particular if the localization results in a nonfinitely generated ring. So I left it as it is for now.

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Properties.jl
```diff
@@ -101,24 +122,22 @@ function is_smooth(P::AbsProjectiveScheme; algorithm=:default)
...
     ))
   end
-  R = base_ring(homogeneous_coordinate_ring(P))
+  R = ambient_coordinate_ring(P)
   I = defining_ideal(P)
   mat = jacobian_matrix(R, gens(I))
   sing_locus = ideal(R, minors(mat, codim(P)))
-  return dim(sing_locus + I) <= 0
+  irrelevant_ideal = ideal(R, gens(R))
+  return is_one(saturation(sing_locus, irrelevant_ideal))
```
afkafkafk13: I think that you want 
                       sing_locus = ideal(R, minors(mat,codim(P)) + I
Otherwise, you could also get points outside of P, because you never add generators of I in lines 139 and 140, whereas this happened in the deleted line 121.

afkafkafk13 (approved):

PR #3834, opened 2024-06-09
refactor vertex-edge graph and dual graph of a polyhedron

micjoswig:
This is the obvious minimal implementation for vertex edge graphs and dual graphs of polyhedra; taken from polymake.

The question is how to deal with unbounded polyhedra, in particular those with a nontrivial lineality space.  The functions will work in these cases, too.  However, the behavior might be unexpected.  polymake considers the projectivized quotient modulo linealities, which is always a polytope.  Oscar, on the other hand, follows a more naive semantics.

Opinions?

### DIFF
--- a/src/Combinatorics/Graphs/functions.jl
+++ b/src/Combinatorics/Graphs/functions.jl
@@ -112,6 +112,7 @@ end
 
 @doc raw"""
     rem_edge!(g::Graph{T}, s::Int64, t::Int64) where {T <: Union{Directed, Undirected}}
+    rem_edge!(g::Graph{T}, e::Edge) where {T <: Union{Directed, Undirected}}
 
 Remove edge `(s,t)` from the graph `g`.
 Return `true` if there was an edge from `s` to `t` and it got removed, `false`
@@ -175,6 +176,8 @@ end
 
 Remove the vertex `v` from the graph `g`. Return `true` if node `v` existed and
 was actually removed, `false` otherwise.
+Please note that this will shift the indices of the vertices with index larger than `v`,
+but it will preserve the vertex ordering.
 
 # Examples
 ```jldoctest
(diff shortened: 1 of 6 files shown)


micjoswig:
And, of course, we need to care of cones, polyhedral complexes etc, too.

src/Combinatorics/Graphs/functions.jl
```diff
@@ -902,20 +936,21 @@ end
...
 ################################################################################
 @doc raw"""
-    edgegraph(p::Polyhedron)
+    vertex_edge_graph(p::Polyhedron)
 
 Return the edge graph of a `Polyhedron`, vertices of the graph correspond to
 vertices of the polyhedron, there is an edge between two vertices if the
 polyhedron has an edge between the corresponding vertices. The resulting graph
 is `Undirected`.
+The keyword argument can be used to consider the polyhedron modulo its lineality space.
```
lkastner:
```suggestion
If the polyhedron has lineality, then it has no vertices or bounded edges, so the `vertex_edge_graph` will be the empty graph.
In this case, the keyword argument can be used to consider the polyhedron modulo its lineality space.
```

lkastner (approved):

benlorenz:
I have changed this code to use (and rename) the existing `dualgraph` and `edgegraph` functions. I have kept aliases for the old names.
And I have adjusted them to deal with unbounded polyhedra:
- The vertex-edge-graph only considers proper vertices (unless using the `modulo_lineality` flag), i.e. it represents the bounded part of the polyhedron.
- The dual-graph corresponds to all (possibly unbounded) facets except for the facet at infinity. And facets that intersect only at infinity are not connected by an edge in the graph.

micjoswig:
> I have changed this code to use (and rename) the existing `dualgraph` and `edgegraph` functions. I have kept aliases for the old names. And I have adjusted them to deal with unbounded polyhedra:

Sounds perfect.  Thanks-

src/aliases.jl
```diff
@@ -48,3 +50,7 @@ function number_of_weak_compositions end
 @alias n_generators ngens
 @alias n_rows nrows
 @alias n_variables nvars
+
+# for backwards compatibility
+@alias dualgraph dual_graph
+@alias edgegraph vertex_edge_graph
```
lgoettgens, line 56: Shouldn't these rather be `@deprecate`? Anyway, if they should be phased out they should be in `deprecations.jl` and not `aliases.jl`. `aliases.jl` only contains the aliases that we plan to keep forever

lkastner: We discussed this internally and this is the result.

PR #3836, opened 2024-06-10
fix Ehrhart

fieker:


src/NumberTheory/GaloisGrp/GaloisGrp.jl
```diff
@@ -2664,7 +2664,11 @@ function blow_up(G::PermGroup, C::GaloisCtx, lf::Vector, con::PermGroupElem=one(
 
   icon = inv(con)
 
-  gs = map(Vector, gens(G))
+  if ngens(G) == 0
+    gs = Vector{Int}[]
+  else
+    gs = map(Vector, gens(G))
+  end
```
lgoettgens:
```suggestion
  gs = Vector.(gens(G))
```
this infers as `Vector{Vector{Int64}}` as well and is a bit easier to read

ThomasBreuer (approved):
Yes, this fixes one problem with empty lists of generators. There are probably more such places in the code.

src/NumberTheory/GaloisGrp/GaloisGrp.jl
fingolfin: That's interesting, but I still would like to understand why `map(Vector, gens(G))` doesn't work anymore.

fingolfin: Best would be to explicitly specify the return type, then Julia does not have to guess. That is, either
```suggestion
  gs = map(Vector{Int}, gens(G))
```
or
```suggestion
  gs = Vector{Int}.(gens(G))
```

src/NumberTheory/GaloisGrp/GaloisGrp.jl
```diff
@@ -2664,7 +2664,7 @@ function blow_up(G::PermGroup, C::GaloisCtx, lf::Vector, con::PermGroupElem=one(
 
   icon = inv(con)
 
-  gs = map(Vector, gens(G))
+  gs = map(Vector{Int}, gens(G))
```
lgoettgens, line 2667:
```suggestion
  gs = Vector{Int}.(gens(G))
```
I am pretty sure the current form won't work, as mapping over an empty list and broadcasting over an empty list infer differently.
In case the tests fail, you could try to adapt the suggestion

fingolfin: Mapping like that works and produces type stable code if the "function" being applied is a concrete type.

Issue #2455, opened 2023-06-07
In OSCAR, polynomial rings should all be constructed with `cached = false`


HereAround:
This was brought to my attention by @fieker. Apparently, this happens in particular for schemes. Hence, I cc @HechtiDerLachs , so we can discuss this at some point.

fingolfin:
There is some relevant text on the caches in the AA documentation, as part of the ring interface. We should check if this text still reflects what we want, and then copy it (after editing) into the OSCAR dev docs.

Also we need to think about allowing optional specification of parents in more functions, e.g. `hilbert_series`; there is an open PR #2084 by @HechtiDerLachs on this, which we need to get finished and merged.

fingolfin:
The overall situation is still rather bad, it wouldn't hurt to evangelize this again, and task some people to work on it.

Running
```
git grep -n '\bpolynomial_ring(' src | fgrep -v 'julia> ' | fgrep -v 'src/InvariantTheory' | egrep -v 'cached\s*=\s*false' 
```
to find examples in Oscar.jl where code creates *cached* polynomial rings, one finds (excluding some obvious false hits):

- [ ] [src/AlgebraicGeometry/Curves/ParametrizationPlaneCurves.jl:25](https://github.com/oscar-system/Oscar.jl/blob/bfd7624b3f2ed4c057770440756fe9066ad297e5/src/AlgebraicGeometry/Curves/ParametrizationPlaneCurves.jl#L25):
      `    newring, _ = polynomial_ring(K, symbols(R))`
- [ ] [src/AlgebraicGeometry/Schemes/AffineSchemeOpenSubscheme/Rings/Methods.jl:246](https://github.com/oscar-system/Oscar.jl/blob/bfd7624b3f2ed4c057770440756fe9066ad297e5/src/AlgebraicGeometry/Schemes/AffineSchemeOpenSubscheme/Rings/Methods.jl#L246):
      `  S, t = polynomial_ring(W, ["t$i" for i in 1:r])`
- [ ] [src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl:298](https://github.com/oscar-system/Oscar.jl/blob/bfd7624b3f2ed4c057770440756fe9066ad297e5/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl#L298):
      `  KL, z = polynomial_ring(k, new_symb)`
[...]

PR #3838, opened 2024-06-10
Do not cache polynomial rings

ederc:
This PR disables caching of `polynomial_ring` in `Rings/*`. See discussion in #2455.


ederc:
@HechtiDerLachs Any idea what to do with these tests from `test/Rings/ReesAlgebra.jl`?
```
  # Both algebras should be the same despite being created differently.
  @test all(x->iszero(RM(x)), gens(modulus(Rg)))
  @test all(x->iszero(Rg(x)), gens(modulus(RM)))
```

HechtiDerLachs:
I'm afraid, I do not yet fully understand the question, sorry. Should I run the test on this branch and look into the error messages?

ederc:
These tests will fail since the corresponding requirement that the parents are equal no longer holds. As discussed in #2455 this is the behaviour wanted. I just do not know how to adjust the corresponding tests, this is something I would like you to decide.

HechtiDerLachs:
Alright, thanks for the clarification. At the moment I can not so easily switch branches since I'm currently working on something else, but I will look into it the next possible moment.

fingolfin:
Orthogonal to this issue, but: @HechtiDerLachs you can have multiple Oscar.jl clones in parallel (or even use a single clone with multiple worktrees). So that you can e.g. have Oscar master and your active development branch and *another* branch all ready and running in different Julia sessions... I can happily help you set that up, too (not saying this to put pressure on you to look into this issue, just to improve your workflow in the future)

HechtiDerLachs:
@ederc : Could you replace the lines in the test by the following:
```julia
@test all(x->iszero(evaluate(x, gens(RM))), gens(modulus(Rg)))
@test all(x->iszero(evaluate(x, gens(Rg))), gens(modulus(RM)))
```
I think that should do the job. At least it does for me locally. 
Thank you and sorry about the delay.

ederc:
@HechtiDerLachs Thanks, I applied your changes.

joschmitt (approved):
Thank you!

PR #3839, opened 2024-06-11
Fibration hopping presentation

HechtiDerLachs:
These are some changes I had to make in order for the code from our book chapter with @simonbrandhorst to run again. Unfortunately, as it seems, our code has not been continuously tested, so we did not catch any regressions beforehand. Probably there is more to follow.


experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -842,6 +842,7 @@ function maximal_associated_points(I::AbsIdealSheaf;
     covering=default_covering(scheme(I)),
     algorithm::Symbol=:GTZ
   )
+  return _maximal_associated_points(I; covering, algorithm, use_decomposition_info=true)
```
HechtiDerLachs: The "old" routine was not necessarily returning false results, but it was just hung up on the particular input from our book chapter and never finished. I found that with `_maximal_associated_points` we get through. Since there is no general speed regression from this (as I already measured on another account), I suggest to just switch to the new implementation.

experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -1200,6 +1201,10 @@ function Base.show(io::IO, I::AbsIdealSheaf)
 end
 
 function Base.show(io::IO, ::MIME"text/plain", I::PrimeIdealSheafFromChart)
+  if has_attribute(I, :name)
```
simonbrandhorst, line 1157:
```suggestion
  io = pretty(io)
  if has_attribute(I, :name)
```

experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -1200,6 +1201,10 @@ function Base.show(io::IO, I::AbsIdealSheaf)
...
+    print(io, get_attribute(I, :name))
+    return
+  end
```
joschmitt, line 1160: I think we have the macro `@show_special` https://nemocas.github.io/AbstractAlgebra.jl/dev/misc/#AbstractAlgebra.PrettyPrinting.@show_special for this.

experimental/Schemes/src/IdealSheaves.jl
simonbrandhorst: If disabled like this, the old code will just rot. Either it is to be removed completely (but then we should first investigate why it hangs and be sure that the new implementation is reliably faster and not just on this notebook),  or we keep it alive, then both implementations have to be tested.

experimental/Schemes/src/IdealSheaves.jl
lgoettgens: I am afraid that this does not work here as the schemes code has its own name handling that is somewhat orthogonal to the AA one

experimental/Schemes/src/IdealSheaves.jl
HechtiDerLachs: @afkafkafk13 : The code in question to be removed is your "old" code for decomposing ideal sheaves. It stems from a time where a lot of the lazyfication had not yet been implemented. Now it can in principle be rewritten in a much more conceptual form with no significant deficits in speed. But I think, you are still the expert on this, since you have probably implemented this already several times in previous versions even in Singular. 

Hence my question: Do you think that in the new implementation in `_maximal_associated_points` I have overlooked some tuning that you have thought of but I didn't?

afkafkafk13: The new code looks really fine. Much better than the one from last year. The whole environment has made a lot of progress. 

Please add a test for roughly the following (before removing the old code):
Create a 2-dim. scheme X with two disjoint parts, choose an ideal sheaf on it which has at least 3 components describing subschemes of dimensions 1 and 0 respectively (so not pure-dimensional). Then blow up in a point on one component of X and then in a point on the other. Now decompose the  strict transform of the ideal sheaf.

This request is not about an optimization, but just to have a simple and inexpensive worst-case scenario in the tests.

HechtiDerLachs: Thanks for having a look. But does such a test not already exist for the old code?

afkafkafk13: No it doesn't, but I wanted to add something of that kind the next time I touch code in IdealSheaves.jl. Between the time of its writing and now, we got blow-up, normalization, lazy gluings and some more stuff. The code here needs to be tested with these in some inexpensive way.... If I come up with a good example, I will let you know.

HechtiDerLachs: I wrote a test now. But I'm waiting with pushing it to see the current test run finish first.

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -2013,7 +2015,7 @@ function _pushforward_lattice_along_isomorphism(step::MorphismFromRationalFuncti
   for D in lat_X
     @assert length(components(D)) == 1 "divisors in the algebraic lattice must be prime"
     I = first(components(D))
-    @assert is_prime(I)
+    @assert I isa PrimeIdealSheafFromChart || has_attribute(I, :is_prime) "ideal sheaf must be known to be prime"
```
afkafkafk13: This might be slightly dangerous. We cannot be sure that noone ever sets a value of false to the attribute.... The value should still be checked for completeness.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -775,6 +775,12 @@ function +(a::T, b::T) where {T<:MPolyQuoLocRingElem}
   if lifted_denominator(a) == lifted_denominator(b) 
     return (parent(a))(lifted_numerator(a) + lifted_numerator(b), lifted_denominator(a), check=false)
   end
+  gcd_ab = gcd([lifted_denominator(b), lifted_denominator(a)])
+  p = divexact(lifted_denominator(a), gcd_ab)
+  q = divexact(lifted_denominator(b), gcd_ab)
+  new_den = p*lifted_denominator(b)
+  return (parent(a))(lifted_numerator(a)*q + lifted_numerator(b)*p, new_den, check=false)
+
   return (parent(a))(lifted_numerator(a)*lifted_denominator(b) + lifted_numerator(b)*lifted_denominator(a), lifted_denominator(a)*lifted_denominator(b), check=false)
```
afkafkafk13: Please do not forget to remove the old return statements of this commit before merging.

experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -1972,7 +1946,10 @@ function is_subset(P::PrimeIdealSheafFromChart, I::AbsIdealSheaf)
   X = scheme(P)
   @assert X === scheme(I)
   U = original_chart(P)
-  return is_subset(P(U), I(U))
+  is_subset(P(U), I(U)) || return false
+  # I might have support outside the support of P. Hence, we can not 
+  # avoid the full check, as it seems.
+  return all(is_subset(P(U), I(U)) for U in affine_charts(X))
```
simonbrandhorst:
```suggestion
  # I might have support outside the support of P. Hence, we can not 
  # avoid the full check, as it seems.
  return all(is_subset(P(U), I(U)) for U in affine_charts(X) if !is_one(I(U)))
```

HechtiDerLachs:
Now the notebook runs through, including pushforward and pullbacks of divisors.

experimental/Schemes/src/elliptic_surface.jl
simonbrandhorst:
```suggestion
    @assert I isa PrimeIdealSheafFromChart || (has_attribute(I, :is_prime) &&  is_prime(P)) "ideal sheaf must be known to be prime"
```

src/Rings/mpolyquo-localizations.jl
```diff
@@ -770,6 +775,12 @@ function +(a::T, b::T) where {T<:MPolyQuoLocRingElem}
...
```
simonbrandhorst:
```suggestion
```

simonbrandhorst (approved):

experimental/Schemes/src/IdealSheaves.jl
simonbrandhorst: This is an optimisation. The rationale is that `is_subset` probably computes a groebner basis of `I(U)` anyways.
Thus `is_one(I(U))` should come at little extra cost. If it is one, there is nothing to test.

HechtiDerLachs:
Current timings for the code of the notebook in a hot session: 
`1502.247163 seconds (4.97 G allocations: 295.001 GiB, 43.90% gc time, 3.97% compilation time: 95% of which was recompilation)`

simonbrandhorst:
Starting from scratch, computing the `algebraic_lattice` takes over `GF(7)`
`439.719678 seconds (1.73 G allocations: 68.728 GiB, 53.91% gc time, 3.15% compilation time: 83% of which was recompilation)`
Over `QQ` 
`535.421043 seconds (1.62 G allocations: 63.388 GiB, 54.89% gc time, 11.54% compilation time: 75% of which was recompilation)`
It is remarkable that 50% of the time is spent in garbage collection. Surely another 40% are spent in allocating the 60GiB

afkafkafk13 (approved):

HechtiDerLachs:
> It is remarkable that 50% of the time is spent in garbage collection.

It is, but I am not surprised. I suppose this comes from massive evaluation of polynomials due to the computation of gluings and probably also due to copying to and from the Singular side. I'm not sure to which extent this can be avoided, really.

HechtiDerLachs:
New run from scratch: `1242.058935 seconds (5.28 G allocations: 316.266 GiB, 20.46% gc time, 20.33% compilation time: <1% of which was recompilation)`
Interestingly it takes *longer* on a hot session: `1541.989411 seconds (4.93 G allocations: 288.714 GiB, 48.18% gc time, 0.11% compilation time)`.

PR #3840, opened 2024-06-11
add `galois_orbit_sum` for group characters

ThomasBreuer:


thofma:
Thanks! Not sure about the name. `galois_orbit_sum` is a bit longer, but maybe a bit more self-explanatory?

ThomasBreuer:
O.k., I will change the name.

thofma (approved):

thofma:
Thanks!

ThomasBreuer:
Funny.
After changing the name of the new function, the test `1.10, short, macOS, RPTU` failed.

joschmitt:
> Funny. After changing the name of the new function, the test `1.10, short, macOS, RPTU` failed.

I assume that this is an issue with the RPTU Mac (and not with the pull request) like I reported here https://oscar-system.slack.com/archives/C5B2H11BQ/p1715673722611139. I don't think that this issue is tracked somewhere?

ThomasBreuer:
Yes, this problem is not caused by the pull request.
The error message says that some file from a GAP package cannot be read.

PR #3845, opened 2024-06-12
interface to GAP's group recognition (first attempt)

ThomasBreuer:
This code provided essentially a new type, the `show` methods derived from GAP's `ViewObj` method for its `IsRecogNode` objects, and a few wrappers for functions dealing with the objects.
Let us discuss this basic setup, and how we want to proceed.

### DIFF
--- a/src/Groups/recog.jl
+++ b/src/Groups/recog.jl
@@ -1,118 +1,225 @@
-@attributes mutable struct GroupRecognitionNode{T <: GAPGroup}
+@attributes mutable struct GroupRecognitionTree{T <: GAPGroup}
   input_group::T
-  gap_node::GapObj
+  gap_tree::GapObj
 
-  function GroupRecognitionNode{T}(G::GAPGroup, gap_node::GapObj) where T
-     res = new{T}(G, gap_node)
+  function GroupRecognitionTree{T}(G::GAPGroup, gap_tree::GapObj) where T
+     @req G isa PermGroup || G isa MatrixGroup "only matrix and permutation groups are supported"
+     res = new{T}(G, gap_tree)
      return res
   end
 end
 
-GapObj(node::GroupRecognitionNode) = node.gap_node
+GapObj(tree::GroupRecognitionTree) = tree.gap_tree
 
-# did the recognition not fail?
-is_ready(node::GroupRecognitionNode) = GAP.Globals.IsReady(node.gap_node)
 
-# does the node describe a leaf?
-is_leaf(node::GroupRecognitionNode) = GAP.Globals.IsLeaf(node.gap_node)
+"""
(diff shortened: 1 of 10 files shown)


src/Groups/recog.jl
```diff
@@ -0,0 +1,118 @@
...
+  function GroupRecognitionNode{T}(G::GAPGroup, gap_node::GapObj) where T
+     res = new{T}(G, gap_node)
+     return res
+  end
+end
+
+GapObj(node::GroupRecognitionNode) = node.gap_node
+
+# did the recognition not fail?
+is_ready(node::GroupRecognitionNode) = GAP.Globals.IsReady(node.gap_node)
```
fingolfin:
```suggestion
is_ready(node::GroupRecognitionNode) = GAP.Globals.IsReady(node.gap_node)::Bool
```

src/Groups/recog.jl
```diff
@@ -0,0 +1,118 @@
...
+
+# does the node describe a leaf?
+is_leaf(node::GroupRecognitionNode) = GAP.Globals.IsLeaf(node.gap_node)
```
fingolfin:
```suggestion
is_leaf(node::GroupRecognitionNode) = GAP.Globals.IsLeaf(node.gap_node)::Bool
```

src/Groups/recog.jl
```diff
@@ -0,0 +1,118 @@
...
+# the group stored in `node`, provided that recognition did not fail;
+# the `GapObj`s of its elements store memory information
+function group(node::GroupRecognitionNode)
+  @req is_ready(node) "recognition failed, no group stored"
+  return _oscar_group(GAP.Globals.Grp(node.gap_node))
+end
+
+function Base.show(io::IO, node::GroupRecognitionNode)
+  if is_terse(io)
+    print(io, "Recognition node of a group")
```
fingolfin: Maybe "Recognition tree" would be more apt and less confusing for users? (I've contemplate changing this name in `recog` itself but since there it'd require changing a lot of code, I've put it off for now)

fingolfin (approved):
That seems like a great start.

There is one function missing:  in GAP a group recognition object allows doing `g in recogNode` which solves the membership problem for the given group element `g`.

(Of course what we'd also want is functions for computing a word/SLP representing `g` in terms of some generators of `G`, but unfortunately this functionality is not complete in `recog` -- @danielrademacher implemented the relevant bits for the leaf cases he solved in his PhD, but we now also need to work on the general infrastructure to get it. We'll work on that, but it won't be there for the Begehung)

fingolfin (approved):
Looks good to me. Shall we just merge it now so that work on a notebook can start? We can still refine it afterwards.

PR #3837, opened 2024-06-10
Upgrade `GModule` to new experimental structure

lgoettgens:
This is the last experimental project with the old structure.
Once this is merged, we can follow up with some cleaning and refactoring of the experimental inclusion code.

To reduce the friction: Now that https://github.com/oscar-system/Oscar.jl/pull/3818 is merged, does anybody currently have an unmerged branch around this that could get conflicts? In particular @mjrodgers @ThomasBreuer @fieker @fingolfin

PR #3849, opened 2024-06-12
Remove the "old experimental structure"

lgoettgens:
With https://github.com/oscar-system/Oscar.jl/pull/3837 merged, there are no more experimental projects adhering to the "old standard". So let's remove the workaround for them and adapt the documentation.


fingolfin (approved):
Finally ;-)

PR #3852, opened 2024-06-12
Export fpField, FpField explicitly, cleanup

fingolfin:


experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -23,10 +23,11 @@ function pop_largest_factor!(f::Fac{ZZRingElem})
 end
 
 # TODO : Should be fixed in Nemo
-function (k::Nemo.FpField)(a::Vector)
+function (k::FpField)(a::Vector)
```
fingolfin, line 26: @mjrodgers I think you added this -- in how far is this a "fix"? Is it adding a missing API that is needed for generic programming or so? Just wondering what the use case is.

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
```diff
@@ -23,10 +23,11 @@ function pop_largest_factor!(f::Fac{ZZRingElem})
...
   @assert isone(length(a))
   return k(a[1])
 end
+
 function (k::FqPolyRepField)(a::Vector)
```
fingolfin, line 31: I guess this is similar to the above

fingolfin: @mjrodgers  Also about `coords` methods a few lines down, where it says:
```
# TODO : Should be fixed in Hecke for prime fields
function coords(x::FinFieldElem)
  return absolute_coordinates(x)
end
function coords(x::PrimeFieldElem)
  return [x]
end
```
but there is no `coords` in Hecke, I think, so I am not sure that is about?

src/imports.jl
```diff
@@ -205,9 +207,5 @@ end
...
 import cohomCalg_jll
 
-import Nemo: is_cyclo_type
-import Nemo: is_maxreal_type
-import Nemo: ZZModRing  # FIXME: remove if/once Nemo exports this
-import Nemo: zzModRing  # FIXME: remove if/once Nemo exports this
-import Nemo: FpField  # FIXME: remove if/once Nemo exports this
-import Nemo: fpField  # FIXME: remove if/once Nemo exports this
+import Nemo: FpField, fpField  # FIXME: remove if/once Nemo exports this
+export FpField, fpField  # FIXME: remove if/once Nemo exports this
```
lgoettgens:
```suggestion
import Nemo: FpField, fpField  # FIXME: remove once https://github.com/Nemocas/Nemo.jl/pull/1789 is available
export FpField, fpField  # FIXME: remove once https://github.com/Nemocas/Nemo.jl/pull/1789 is available
```
adding a PR link makes it easier in the future to check if the FIXME can be taken care of

lgoettgens (approved):

lgoettgens:
I would even prefer to not have these comments at all but instead release a patch version of Nemo right after https://github.com/Nemocas/Nemo.jl/pull/1789 finishes CI and updating the Nemo compat here. (You already made the extra step and fix it in Nemo, so why add another temporary workaround?)

lgoettgens (approved):
Thanks!

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
mjrodgers: This is quite old now, and I'm guessing it can be removed with the migration to the unified finite field type. I'm playing around with some things and seeing how to bring this experimental package up to date with some of the intervening OSCAR changes.

experimental/StandardFiniteFields/src/StandardFiniteFields.jl
mjrodgers: There was an issue at the time that `absolute_coordinates` was not defined for fields of prime order (which meant they needed to be handled as a special case), so this was a workaround. I think now that the finite fields are unified, this can be removed; I'm looking into this.

fieker:
On Mon, Jun 17, 2024 at 04:40:43AM -0700, Morgan Rodgers wrote:
> @mjrodgers commented on this pull request.
> 
> 
> 
> >    @assert isone(length(a))
>    return k(a[1])
>  end
> +
>  function (k::FqPolyRepField)(a::Vector)
> 
> There was an issue at the time that `absolute_coordinates` was not defined for fields of prime order (which meant they needed to be handled as a special case), so this was a workaround. I think now that the finite fields are unified, this can be removed; I'm looking into this.
> 
Be careful: replacing Native.GF(3) by GF(3) easily incurs a factor 10
slow down...

mjrodgers:
> Be careful: replacing Native.GF(3) by GF(3) easily incurs a factor 10 slow down...

I just realized that this is the reason; I will at least check if these fixes are incorporated yet to Nemo/Hecke (or if they should be). I think the first is implemented in Nemo but maybe not exported. I am not sure if there is any demand for `absolute_coordinates` to be defined for native finite fields with prime order, but I could see an argument (if only because I personally find it annoying when something fails just because the obvious return value would be trivial).

fingolfin:
It's always helpful if a method is available for all implementations of a given concept (here: finite fields), so that one can write code targeting them in a uniform way -- at least when it conceptually and structurally makes sense, which seems like the case here?

PR #3856, opened 2024-06-13
polynomial_ring cached=false for combinatorics

ederc:
Some tests have now problems with different parent rings, see discussion at #2455.

Maybe @bschroter can help fixing the tests?

### DIFF
--- a/src/Combinatorics/Matroids/properties.jl
+++ b/src/Combinatorics/Matroids/properties.jl
@@ -884,8 +884,9 @@ function tutte_connectivity(M::Matroid)
 end
 
 @doc raw"""
+    tutte_polynomial(M::Matroid)
     tutte_polynomial(M::Matroid; parent::ZZMPolyRing)
-    tutte_polynomial(R::ZZMPolyRing, M::Matroid)
+    tutte_polynomial(parent::ZZMPolyRing, M::Matroid)
 
 Return the Tutte polynomial of `M`. This is polynomial in the variables x and y with integral coefficients.
 See Section 15.3 in [Oxl11](@cite).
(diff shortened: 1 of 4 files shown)


src/Combinatorics/Matroids/properties.jl
```diff
@@ -918,7 +918,7 @@ q^3 - 7*q^2 + 14*q - 8
 ```
 """
 function characteristic_polynomial(M::Matroid)
```
fingolfin: These functions touched here should all at some point be changed allow passing in the polynomial ring, as discussed [here in our dev docs](https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/#Optional-arguments-for-parents-of-return-values)

(To be clear, that's not a request to change this PR)

src/Combinatorics/Matroids/properties.jl
```diff
@@ -936,7 +936,7 @@ q^2 - 6*q + 8
 ```
 """
 function reduced_characteristic_polynomial(M::Matroid)
-    R, q = polynomial_ring(ZZ, 'q')
+    R, q = polynomial_ring(ZZ, 'q'; cached=false)
     p = characteristic_polynomial(M)
```
fingolfin:
```suggestion
    p = characteristic_polynomial(M)
    R = parent(p)
```

src/Combinatorics/Matroids/properties.jl
```diff
@@ -936,7 +936,7 @@ q^2 - 6*q + 8
...
     c = Vector{Int}(undef,degree(p))
```
fingolfin: Huh... that seems risky, couldn't one have examples where an `Int` would over- or underflow?

src/Combinatorics/Matroids/properties.jl
```diff
@@ -884,7 +884,8 @@ function tutte_connectivity(M::Matroid)
 end
 
 @doc raw"""
-    tutte_polynomial(M::Matroid)
+    tutte_polynomial(M::Matroid; parent::ZZMPolyRing)
```
fingolfin, line 888: Can we indicate somehow that `parent` is optional? E.g.
```suggestion
    tutte_polynomial(M::Matroid[; parent::ZZMPolyRing])
```
but I am not sure this is universally understood. Or perhaps
```suggestion
    tutte_polynomial(M::Matroid; parent::ZZMPolyRing = ...)
```
and/or just say in the text that a parent can be *optionally* passed via kwarg or as first argument or so

lgoettgens (approved):

PR #3858, opened 2024-06-13
polynomial_ring cached=false for tropical geometry

ederc:
See #2455.

Tests need to be adjusted, maybe @YueRen can provide some help.


YueRen:
@ederc Thanks for the changes, much appreciated!  I will look over them tomorrow.

YueRen:
(This is more a memo to myself, though comments are welcome) Okay, here is why the change is problematic and what I plan to do to fix it:

Consider a polynomial ring `K[x1,..,xn]` over a valued field `K`.  In addition to the original polynomial ring `K[x1,..,xn]`, computations in tropical geometry require 
- a polynomial ring `R[x1,...,xn]` over the valuation ring `R` (for tropical Groebner bases)
- a polynomial ring `k[x1,...,xn]` over the residue field field `k` (this is where initial ideals live)

Currently, I just (re-)create `R[x1,...,xn]` and `k[x1,...,xn]` from `K[x1,...,xn]` and blindly trust that it always ends up being the same polynomial ring.  This is probably not very smart.

What should happen is that `K[x1,..,xn]` should store `R[x1,..,xn]` and `k[x1,..,xn]` as attributes.  This will require lots of changes, so it might take a bit of time (ETA August).

fingolfin:
Why should `K[x1,..,xn]` store `R[x1,..,xn]` and `k[x1,..,xn]` as attributes? That would be a bit unusual. Perhaps it is unavoidable for some reason, but normally I would expect that there is some kind of parent object (perhaps a "tropical variety" or something like that) which contains these rings.

I had a quick look at the changes in this PR and the tropical code in general but it wasn't apparent to me at all how and when these

YueRen:
> Why should `K[x1,..,xn]` store `R[x1,..,xn]` and `k[x1,..,xn]` as attributes? That would be a bit unusual. Perhaps it is unavoidable for some reason, but normally I would expect that there is some kind of parent object (perhaps a "tropical variety" or something like that) which contains these rings.
> 
> I had a quick look at the changes in this PR and the tropical code in general but it wasn't apparent to me at all how and when these

I should probably save it in `TropicalSemiringMap` then.  `TropicalSemiringMap`'s main job is to encode a valuation on `K` and a choice between `min`-convention and `max`-convention, and it is always passed along in tropical (algebraic) computations.  It could also store a dictionary, which maps `K[x]` to `R[x]` and `k[x]` (in case multiple polynomial rings have the same `K`).

YueRen:
I just realized that this branch is on @ederc's fork of Oscar.  What is the simplest way for me to rebase it to the newest master and add changes on top of it?

benlorenz:
> I just realized that this branch is on @ederc's fork of Oscar. What is the simplest way for me to rebase it to the newest master and add changes on top of it?

Github shows `Maintainers are allowed to edit this pull request.` so you should be able to push to his branch.
With pure git something like this should work:
```
git fetch origin
git fetch origin pull/3858/head:tropcache
git checkout tropcache
git rebase origin/master
git commit ....
git push --force git@github.com:ederc/Oscar.jl tropcache:caching-4
```
Otherwise you could try the `gh` cli.

YueRen:
@benlorenz Thx for the info!  I didn't know that automatically allowing maintainer to push onto your remote branches was possible.

YueRen (approved):
Apologies, I ended up changing more than necessary, so I will try to summarize them (in decreasing number of lines affected):

1. removed various `debug*.jl` files that must have been checked in by accident
2. added `get_polynomial_ring_for_*` functions in `groebner_basis.jl` and `initial.jl` that construct the desired polynomial rings and cache it in an attribute of the original polynomial ring for later reuse.  Before, the code simply constructed the desired polynomial rings and relied on OSCAR to ensure the same ring were being reused.  And caching it in the original polynomial ring made the most sense, since the only other non-optional input is the weight vector.  All polynomial rings constructed in the code should now have `cached=false`
3. removed various `perturbation` optional named inputs, as that feature was simply not used
4. restructured the (tropical) `groebner_basis` command to cover all ideals.

src/TropicalGeometry/initial.jl
```diff
@@ -1,9 +1,41 @@
...
+#
+################################################################################
+function get_polynomial_ring_for_initial(R::MPolyRing, nu::TropicalSemiringMap)
+    @req coefficient_ring(R)==valued_field(nu) "coefficient ring is not valued field"
+
+    if !has_attribute(R, :tropical_geometry_polynomial_rings_for_initial)
+        set_attribute!(R, :tropical_geometry_polynomial_rings_for_initial, Dict{TropicalSemiringMap,MPolyRing}())
+    end
+
+    polynomialRingsForInitial = get_attribute(R, :tropical_geometry_polynomial_rings_for_initial)
```
fingolfin:
```suggestion
    polynomialRingsForInitial = get_attribute!(R, :tropical_geometry_polynomial_rings_for_initial) do
        return Dict{TropicalSemiringMap,MPolyRing}()
    end::Dict{TropicalSemiringMap,MPolyRing}
```

fingolfin:
The RPTU runner don't seem to be picking up jobs. There was a temporary outage of the cooling system on campus, and they had to shut down all math department compute servers. Supposedly they are all back now except for wawa, but I wonder if perhaps something went wrong with bring the CI runners back?

@aaruni96 could you have a look?

fingolfin (approved):

joschmitt:
According to the book tests, the output in the book example `groebner.jlcon` has changed:
```
  julia> R,(x1,x2,x3,x4) = QQ[\"x1\",\"x2\",\"x3\",\"x4\"];
  
  julia> I = ideal([x1+2*x2-3*x3, 3*x2-4*x3+5*x4]);
  
  julia> nu_2 = tropical_semiring_map(QQ,2);
  
  julia> GBI = groebner_basis(I,nu_2,[0,0,0,0])
  2-element Vector{QQMPolyRingElem}:
-  x1 + 2*x2 - 3*x3
+  3*x1 - x3 - 10*x4
   3*x2 - 4*x3 + 5*x4
  
  julia> inI = initial(I,nu_2,[0,0,0,0])
  Ideal generated by
    x1 + x3
    x2 + x4
  
  julia> GBI = groebner_basis(I,nu_2,[1,0,0,1])
  2-element Vector{QQMPolyRingElem}:
   3*x2 - 4*x3 + 5*x4
-  -x1 - 2*x2 + 3*x3
+  -3*x1 + x3 + 10*x4
  
  julia> inI = initial(I,nu_2,[1,0,0,1])
  Ideal generated by
    x2
    x3
  
  julia> coefficient_ring(inI)
  Prime field of characteristic 2
```
(Apart from the usual garbage collector issue with the book tests.)

YueRen:
@joschmitt Good catch, I've changed the book tests accordingly.  Both new and old results are correct, since they are contained in the original ideal and give the same polynomials over `GF(2)`.

PR #3864, opened 2024-06-17
Sets poly ring caching to false for invariant theory

ederc:
See #2455.


joschmitt (approved):
Looks good, thanks! (I pushed a commit that fixes the formatting.)

PR #3865, opened 2024-06-17
Sets poly ring caching to false for algebraic geometry

ederc:
See #2455.

Some tests needs adjustments, maybe @HechtiDerLachs can have a look.

### DIFF
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
@@ -18,7 +18,7 @@ function fiber_product(
   )
   Y = domain(f)
   X = codomain(f)
-  X == codomain(g) || error("maps need to have the same codomain")
+  X === codomain(g) || error("maps need to have the same codomain")
   Z = domain(g)
   YxZ, pY, pZ = product(Y, Z)
   RX = ambient_coordinate_ring(X)
--- a/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
+++ b/src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Methods.jl
@@ -478,9 +478,36 @@ The optional arguments `domain_map` and `codomain_map` can be used
 to specify the morphisms `b₁` and `b₂`, respectively.
 """
 function base_change(phi::Any, f::AbsAffineSchemeMor;
-    domain_map::AbsAffineSchemeMor=base_change(phi, domain(f))[2],
-    codomain_map::AbsAffineSchemeMor=base_change(phi, codomain(f))[2]
+    domain_map::Union{AbsAffineSchemeMor, Nothing} = nothing,
+    codomain_map::Union{AbsAffineSchemeMor, Nothing} = nothing
   )
+  # We need to cater for the case where `phi` is a natural inclusion. 
+  # In that case, we have the same `ambient_coordinate_ring` for domain and codomain
+  # and we need to make sure, its base change is performed only once.
+  if domain_map === nothing 
+    ambient_coordinate_map = nothing # initialize the variable
+    if codomain_map !== nothing && ambient_coordinate_ring(domain(f)) === ambient_coordinate_ring(codomain(f))
+      # The `ambient_coordinate_ring` is already determined by the map on the codomain
+      R = ambient_coordinate_ring(domain(f))
+      R_red = ambient_coordinate_ring(domain(codomain_map))
+      ambient_ring_map_domain = hom(R, R_red, phi, gens(R_red))
+    else
+      _, ambient_ring_map_domain = _change_base_ring(phi, ambient_coordinate_ring(domain(f)))
+    end
+    _, domain_map = base_change(phi, domain(f); ambient_ring_map=ambient_ring_map_domain)
+  end
+  if codomain_map === nothing 
+    ambient_ring_map_codomain = nothing # initialize the variable
+    if ambient_coordinate_ring(domain(f)) === ambient_coordinate_ring(codomain(f))
+      # The `ambient_ring` is already determined by the map on the domain
+      R = ambient_coordinate_ring(domain(f))
+      R_red = ambient_coordinate_ring(domain(domain_map))
+      ambient_ring_map_codomain = hom(R, R_red, phi, gens(R_red))
+    else
+      ambient_ring_map_codomain = _change_base_ring(phi, ambient_coordinate_ring(codomain(f)))
+    end
+    _, codomain_map = base_change(phi, codomain(f); ambient_ring_map=ambient_ring_map_codomain)
+  end
   X = domain(f)
   Y = codomain(f)
   XX = domain(domain_map)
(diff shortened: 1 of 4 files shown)


src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Constructors.jl
```diff
@@ -176,7 +176,7 @@ with coordinates [y1, y2, y3, y4, y5]
 ```
 """
 function affine_space(kk::BRT, n::Int; variable_name="x") where {BRT<:Ring}
-  R, _ = polynomial_ring(kk, [variable_name * "$i" for i in 1:n])
+  R, _ = polynomial_ring(kk, [variable_name * "$i" for i in 1:n]; cached=false)
```
simonbrandhorst: in the constructors for `affine_space` the ring should be cached!  If desired, we can add a keyword
`cached` and pass that on to `polynomial_ring`

src/AlgebraicGeometry/Schemes/ProjectiveSchemes/Objects/Constructors.jl
```diff
@@ -89,7 +89,7 @@ Multivariate polynomial ring in 3 variables over QQ graded by
 """
 function projective_space(A::Ring, var_symb::Vector{<:VarName})
   n = length(var_symb)
-  S, _ = graded_polynomial_ring(A, Symbol.(var_symb))
+  S, _ = graded_polynomial_ring(A, Symbol.(var_symb); cached=false)
```
simonbrandhorst:
```suggestion
  S, _ = graded_polynomial_ring(A, Symbol.(var_symb))
```
Same for the constructors for `projective_space`

src/AlgebraicGeometry/Schemes/AffineSchemes/Objects/Constructors.jl
lkastner: Why?

simonbrandhorst: For the same reason that `polynomial_ring` is cached by default.  (whether or not that reason is good)

lkastner: Cool, so you want the following behaviour?
```
julia> A1 = affine_space(NormalToricVariety, 3)
Normal toric variety

julia> A2 = affine_space(NormalToricVariety, 3)
Normal toric variety

julia> A1 == A2
false

julia> A1 = affine_space(QQ, 3)
Affine space of dimension 3
  over rational field
with coordinates [x1, x2, x3]

julia> A2 = affine_space(QQ, 3)
Affine space of dimension 3
  over rational field
with coordinates [x1, x2, x3]

julia> A1 == A2
true
```

simonbrandhorst: Good catch. In this case I would say that the `NormalToricVariety` has the unexpected behavior and should cache.

simonbrandhorst: (or possibly implement equality different)

lkastner: Currently rings are special as they basically only have one implementation. This is not true for varieties which exist as `AffineVariety`, `NormalToricVariety`, `Scheme`, etc. Your suggestion then needs to be passed on to every other implementation.

simonbrandhorst: Yes.

fingolfin: I don't understand why the ring here should be cached. I could see that perhaps it might be useful to have the output of `affine_space` itself be cached, but that would mean adding a `cached::Bool` kwarg to `affine_space` and then adding a cache.

lkastner: > Yes.

So should caching then be extended to all standard constructions, like del pezzo, Hirzebruch, etc? Otherwise it would be inconsistent.

simonbrandhorst: The reason is that these affine spaces appear as keys in certain dictionaries. Sometimes we want different spaces.

lkastner: Ok, now you make a dictionary:
```
julia> A1 = affine_space(QQ, 3);

julia> A2 = affine_space(QQ, 3);

julia> d = Dict(A1=>1)
Dict{AffineVariety{QQField, QQMPolyRing}, Int64} with 1 entry:
  Affine 3-space over QQ with coordinates [x1, x2, x3] => 1

julia> d[A2]
ERROR: KeyError: key Affine 3-space over QQ with coordinates [x1, x2, x3] not found
Stacktrace:
 [1] getindex(h::Dict{AffineVariety{QQField, QQMPolyRing}, Int64}, key::AffineVariety{QQField, QQMPolyRing})
   @ Base ./dict.jl:498
 [2] top-level scope
   @ REPL[30]:1

julia> A1 == A2
true
```
I don't understand what you are doing.

Also you could make a `Dict{Scheme, FOO}` and then your proposed equality would cause this `Dict` to behave differently for different schemes.

simonbrandhorst: We  do IDdicts

simonbrandhorst: From my point of view this would be a breaking change of behavior.

lkastner:
```
julia> idd = IdDict(A1=>1)
IdDict{AffineVariety{QQField, QQMPolyRing}, Int64} with 1 entry:
  Affine 3-space over QQ with coordinates [x1, x2, x3] => 1

julia> idd[A1]
1

julia> idd[A2]
ERROR: KeyError: key Affine 3-space over QQ with coordinates [x1, x2, x3] not found
Stacktrace:
 [1] getindex(d::IdDict{AffineVariety{QQField, QQMPolyRing}, Int64}, key::Any)
   @ Base ./iddict.jl:108
 [2] top-level scope
   @ REPL[35]:1
```

simonbrandhorst: > Currently rings are special as they basically only have one implementation. This is not true for varieties which exist as `AffineVariety`, `NormalToricVariety`, `Scheme`, etc. Your suggestion then needs to be passed on to every other implementation.

How does having several implementations relate to caching?
(Besides, we have many finite fields als well as far as I remember, we also have several implementations of the field of rational numbers etc.)
Do we have a rationale what is in the global cache and what not?

In any case I would be fine with caching `affine_space` as @fingolfin suggests. (but I don't care enough to implement it)

lkastner: > How does having several implementations relate to caching?

I illustrated this with the example `affine_space(NormalToricVariety, *)`, as caching would then have to be implemented for every new implementation of `affine_space` to ensure consistent behaviour. Opposed to only once if there is only one implementation.

> but I don't care enough to implement it

Ok, so who is going to write conformance tests for the caching mechanisms you want?

simonbrandhorst: Well, do as you see fit. I can live entirely without global caching.

fingolfin:
We just discussed this in triage. Overall we said: a new affine (or projective or whatever) space should have a new (non-cached) polynomial ring.

Of course if someone wants to cache these spaces at some point, they can do that; but then this would not even involve the `cached` argument for the polynomial ring because we'd be taking the affine space from its own cache.

As such, adding `cached=false` to the polynomial ring constructors seems correct. Nobody objected in the triage meeting to this claim.


Independent of this there are still some failing tests, someone (e.g. @HechtiDerLachs ?) needs to look into those.

fingolfin:
This PR is now quite minimal but shows a genuine issue and requires help by experts on the code in `src/AlgebraicGeometry/Schemes/AffineSchemes/`... So I guess @HereAround @simonbrandhorst @afkafkafk13 ... ?

simonbrandhorst:
@HechtiDerLachs I tried to figure out what is going on but failed. 
Would you please take a look? It is your code.

HechtiDerLachs:
The problem seems to be bigger. I indeed overlooked the implicit use of some caching for base changes. I would prefer to use this Friday to make use of all the other people being around to get some different stuff done. I will get back to this issue next week.

fingolfin:
That makes sense. Thank you, @HechtiDerLachs

PR #3868, opened 2024-06-18
Add option to endow tensor product with product ordering

joschmitt:
@wdecker this is what we talked about: I added a keyword argument `use_product_ordering` to `tensor_product`.

Also fixes some typos in the implementation and adds very basic tests.

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -1885,36 +1885,50 @@ symbols(A::MPolyQuoRing) = symbols(base_ring(A))
 # Tensor products of rings
 ########################################################################
 
-function tensor_product(A::MPolyRing, B::MPolyRing)
+function tensor_product(A::MPolyRing, B::MPolyRing; use_product_ordering::Bool = false)
   kk = coefficient_ring(A)
   @assert kk === coefficient_ring(B) "coefficient rings do not coincide"
   res, a, b = polynomial_ring(kk, symbols(A), symbols(B); cached=false)
+  if use_product_ordering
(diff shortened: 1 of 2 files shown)


src/Rings/MPolyQuo.jl
```diff
@@ -1885,36 +1885,50 @@ symbols(A::MPolyQuoRing) = symbols(base_ring(A))
...
   return res, hom(A, res, pr.(inc_A.(gens(RA))); check=false), hom(B, res, pr.(inc_B.(gens(RB))); check=false)
 end
 
-function tensor_product(A::MPolyQuoRing, B::MPolyRing)
+function tensor_product(A::MPolyQuoRing, B::MPolyRing; use_product_ordering::Bool = false)
   RA = base_ring(A)
   P, inc_A, inc_B = tensor_product(RA, B)
   I = ideal(P, inc_A.(gens(modulus(A))))
-  res, pr = quo(R, I)
-  return res, hom(A, res, pr.(inc_A.(gens(RA))); check=false), compose(inc_B, pr)
```
joschmitt, line 1917: @HechtiDerLachs I replaced the `compose` by a "manual" construction of the map because `pr` is a `MapFromFunc` so the result of `compose` is a generic `CompositeMap` with which one can't do much.

ederc (approved):

src/Rings/MPolyQuo.jl
HechtiDerLachs: Looks fine. If it works, it works.

HechtiDerLachs (approved):

wdecker:
@joschmitt thx!

PR #3871, opened 2024-06-18
[FTheoryTools] Refactor F-Theory QSMs

HereAround:
* Introduce functions to access the data on the F-theory QSMs conveniently.
* Add a lot of documentation to explain what this data is/how to use it.
* Clean-up. (Move auxiliary type of types file and remove unused function.)
* Include doc-tests for all newly added functions.

@apturner Can you please double-check on the names for the new functions and let me know if they make sense. We export those here, and thus should make a careful choice.

@emikelsons The models in this PR are special instances of the F11-model that you learned about a lot recently.


experimental/FTheoryTools/docs/src/literature.md
```diff
@@ -120,3 +120,80 @@ Provided that a resolution for a model is known, we can (attempt to) resolve the
...
+hodge_h22(m::AbstractFTheoryModel)
+```
+
+More recently, a research program estimated the exact massless spectra of the F-theory QSMs
+(cf. [Bie24](@cite)). These studies require yet more information about the F-theory QSM geometries,
+which are supported by our database.
+
+First, recall that (currently), the base of an F-theory QSM is a 3-dimensional toric variety B3.
+Let s in H^0(B3, Kbar_B3), then V(s) is a K3-surface. Moreover, let xi be the coordinates
+of the Cox ring of B3. Then V(xi) is a divisor in B3. Consequently, Ci = V(xi) cap V(s)
```
apturner, line 157: This should probably not just be plaintext, but we can fix that later due to time sensitivity

experimental/FTheoryTools/docs/src/literature.md
```diff
@@ -120,3 +120,80 @@ Provided that a resolution for a model is known, we can (attempt to) resolve the
...
+is a divisor in the K3-surface V(s). For the root bundle counting program, these curves Ci are
+of ample importance (cf. [Bie24](@cite)). We support the following information on these curves:
+```@docs
+genera_of_ci_curves(m::AbstractFTheoryModel)
+degrees_of_kbar_restrictions_to_ci_curves(m::AbstractFTheoryModel)
+topological_intersection_numbers_among_ci_curves(m::AbstractFTheoryModel)
+indices_of_trivial_ci_curves(m::AbstractFTheoryModel)
+topological_intersection_numbers_among_non_trivial_ci_curves(m::AbstractFTheoryModel)
```
apturner: I would probably replace `_non_trivial_` here with `_nontrivial_`. Both "non-trivial" and "nontrivial" are in common use, and this function name already has so many underscores in it

experimental/FTheoryTools/src/types.jl
```diff
@@ -143,3 +143,80 @@ end
   end
 
 end
+
+
+##############################################################################
+# 3 Struct for Quadrillion F-theory Standard Models when read in from database
+##############################################################################
+
+struct QSMModel
```
apturner, line 152: This should be a subtype of AbstractFTheoryModel, probably. And I suppose you intentionally made it immutable?

apturner (approved):
Thank you @HereAround

experimental/FTheoryTools/docs/src/literature.md
HereAround: Agreed.

experimental/FTheoryTools/src/types.jl
HereAround: No. This is just a helper type to facilitate reading from the QSM database. This reading process returns an element of this type. I then refactor it into a proper QSM model with set attributes.

Presumably this could be improved in the future.

experimental/FTheoryTools/docs/src/literature.md
HereAround: Ok. Let me adjust this...

experimental/FTheoryTools/src/types.jl
apturner: Understood, then it makes sense as is

experimental/FTheoryTools/docs/src/literature.md
HereAround: Just updated the PR accordingly.

HereAround: Third attempt now. The doc tests just keep failing repeatedly due to printing changes in dicts. :(

Issue #3870, opened 2024-06-18
`groebner_basis` of modules and PR #3868


wdecker:
The new functionality provided by PR #3868 does not (yet) fit with the `groebner_basis` function. Can someone with knowledge  please adjust the `groebner_basis` function and add an example to the tests?

julia> D, (w, x, y, z) = polynomial_ring(QQ, ["w", "x", "y", "z"]);

julia> C, (s,t) = polynomial_ring(QQ, ["s", "t"]);

julia> T, _ = tensor_product(C, D, use_product_ordering = true);

julia> F = free_module(T, 2);

julia> G = gens(T);

julia> S, _ = sub(F, [G[1]*gens(F)[1]]);

julia> groebner_basis(S);
ERROR: type ProdOrdering has no field vars

ederc:
It looks like when checking for the embedded ring ordering there is no handling of a possible product ordering. This is something @RafaelDavidMohr and I should investigate

joschmitt:
Fixed by #3873.

PR #3873, opened 2024-06-19
Checking for product orderings

ederc:
This fixes issue #3870. It checks if an embedded ring ordering in a module ordering is a product ordering and handles this case correspondingly.

@RafaelDavidMohr: Please have a look if this is the correct way to fix this issue.


RafaelDavidMohr:
Yes, this looks good to me!

wdecker (approved):

PR #3874, opened 2024-06-19
rearrange how Oscar loads GAP packages at runtime

ThomasBreuer:
- GAP package OscarInterface: add dependency Polycyclic (formally needed)

- loading order of GAP packages:
  - in the beginning, install recog via a URL
  - load OscarInterface in the end

(This has an effect only if the environment variable `GAP_BARE_DEPS` is set at the time when GAP.jl's `__init__` gets called. Currently we cannot achieve this from Oscar.jl.)


ThomasBreuer:
The failure of the `Invalidations` test is subtle.

(This is not the usual difference in the count of invalidations, there is really an error message,
but this appears in a piece of GAP code that should not get touched after the latest commit.)

The point seems to be that the latest commit fixes an error that occurs when Oscar gets loaded. The `Invalidations` test compares old and new situation, and for that it loads Oscar in the old situation, which runs into the error.
(If this is the case then the problem should disappear when the `Invalidations` test runs after danielrademacher/recog/pull/2 has been merged.)

src/utils/install.jl
```diff
@@ -0,0 +1,58 @@
...
+  dirs = GAP.Globals.List(info, i -> i.InstallationPath)
+  repo = GAP.Globals.Filename(GAP.Globals.List(dirs, GAP.Globals.Directory), GapObj(".git"));
+  if repo !== GAP.Globals.fail
+    # the package is already installed, update it
+    return GAP.Globals.UpdatePackage(name, interactive)
+  end
+
+  ! GAP.Globals.PKGMAN_IsValidTargetDir(dir) && return false
+  if branch == ""
+    exec = GAP.Globals.PKGMAN_Exec(GapObj("."), GapObj("git"), GapObj("clone"), url, dir)
```
fingolfin, line 29: This code is indeed a very 1-1 translation of the GAP code, otherwise bits like this here really should have been written as
```suggestion
    exec = run(`git clone $url $dir`)
```
(after changing `url` and `dir` to be Julia strings, of course).

But given that this function is meant as a temporary hack, since you already wrote and tested it this way, let's keep it that way then

fingolfin (approved):

src/utils/install.jl
ThomasBreuer: My idea was that we need something a bit different from what PackageManager provides, and that there will be a few necessary improvements which then might be carried back to PackageManager. Let us see.

fingolfin: of course

benlorenz:
As discussed with @fingolfin I removed this from the 1.1 branch again since this seems a bit risky as it might affect other installed Oscar version via the shared gaproot.

PR #3876, opened 2024-06-19
Modules over NCRing

Lax202:
Requires [#1322](https://github.com/thofma/Hecke.jl/pull/1322) from Hecke, which aims to support sparse matrices over NCRing.


src/Modules/UngradedModules/FreeMod.jl
```diff
@@ -75,6 +75,7 @@ free_module(R::MPolyRing, p::Int, name::VarName = :e; cached::Bool = false) = Fr
 free_module(R::MPolyQuoRing, p::Int, name::VarName = :e; cached::Bool = false) = FreeMod(R, p, name, cached = cached)
 free_module(R::MPolyLocRing, p::Int, name::VarName = :e; cached::Bool = false) = FreeMod(R, p, name, cached = cached)
 free_module(R::MPolyQuoLocRing, p::Int, name::VarName = :e; cached::Bool = false) = FreeMod(R, p, name, cached = cached)
+free_module(R::NCRing, p::Int, name::VarName = :e; cached::Bool = false) = FreeMod(R, p, name, cached = cached)
```
fingolfin, line 78: If we do this then we can effectively delete the four methods above as they are all special cases no of this one. And also adjust the docstring.

But then ideally we also test this over a bunch of other rings `R`, e.g. `ZZ`, `QQ`, `cylotomoic_field(10)`, a free associative algebra, etc.

But I wonder how complete and correct it will be in all those cases, e.g. when `R = ZZ` ?

thofma:
Are these left or right modules? What is the design idea for handling left and right modules?

Is this supposed to do something useful for all rings or for some specific class of rings?

thofma:
Also, there is already
```
free_module(R::NCRing, rank::Int64; cached) @ AbstractAlgebra ~/AbstractAlgebra/src/FreeModule.jl:18
```
so this here won't work.

src/imports.jl
```diff
@@ -166,6 +166,7 @@ let exclude_hecke = [
     :change_uniformizer,
     :coefficients,
     :exponent_vectors,
+    :free_module,
```
thofma, line 169: No, this is not how this is supposed to be used.

wdecker:
> Are these left or right modules? What is the design idea for handling left and right modules?
> 

The idea is to focus on left modules and treat right modules as left modules over the opposite algebra.

> Is this supposed to do something useful for all rings or for some specific class of rings?

What I have in mind here are PBW algebras and their quotients by two-sided ideals.

thofma:
> > Is this supposed to do something useful for all rings or for some specific class of rings?
> 
> What I have in mind here are PBW algebras and their quotients by two-sided ideals.

Then I would suggest adding methods for those types instead of removing `free_module` methods that other people have implemented.

wdecker:
That is, actually, what I had in mind.

Von meinem iPhone gesendetAm 19.06.2024 um 19:25 schrieb Tommy Hofmann ***@***.***>:﻿


Is this supposed to do something useful for all rings or for some specific class of rings?

What I have in mind here are PBW algebras and their quotients by two-sided ideals.

Then I would suggest adding methods for those types instead of removing free_module methods that other people have implemented.

src/Modules/UngradedModules/FreeMod.jl
fingolfin: As was pointed out elsewhere, for now this is really intended for PBW algebras, so I'd restrict the type here.

Also, looking at the code further, one very quickly runs into code that insists on using `ModuleGens`, which means this free module implementation in its current form is quite closely tied to rings backed by Singular, which is another reason to restrict it here for now.

fieker:
On Wed, Jun 19, 2024 at 07:32:19AM -0700, Tommy Hofmann wrote:
I fear we need to bite the bullet and sort the modules a bit better.
There are
 - dense free_modules (AA)
 - sparse free modules (O)
Currently, you get the sparse version if this is a mpoly (or related)
ring, and dense otherwise.
Looking at this, we need to separate the dense/ sparse differently.
Suggestions welcome, in particular those, that do not neccessitate
rewriting the Geometry or the other parts, too much, after the Begehung.

Furthermore, I'm not sure we have a concept for the NC-world that has
any consense, maybe we also need to discuss this a bit more.

Based on vague hearsay, I am not convinced that wrapping blindly
Singular features is going to be a winning move here.

Maybe this draft PR should be considered as a call-of-interest or even
call-of-intend, but now the discussion can start.
@wdecker, @thofma, @Lax202, @fingolfin, @jankoboehm

wdecker:
> Based on vague hearsay, I am not convinced that wrapping blindly Singular features is going to be a winning move here.

As you can see from what is already there for PBW algebras and their quotients, we are NOT blindly wrapping Singular code. We build up OSCAR data structures  and refer to Singular for the Gröbner basis computations. The latter is wise as Singular's implementation is fast: It is very close to what we do in the commutative world. Now the task is setting up modules. This can be done in complete analogy to what we already have over polynomial rings and their quotients and can be started just now. So what points do you wish to discuss prior to that?

src/Modules/UngradedModules/FreeMod.jl
jankoboehm: Sure `ModuleGens` has features for Singular double book-keeping, but you don't have to to use it (e.g. over localizations we don't). In the end, one has to say how to do the curcial toolbox (kernel/lift etc) by implementing the new signatures. Then everything works (surely it can also work over `ZZ` using Smith, or over fields, indeed we use it over fields for cohomology). 

However, in the setting we have in mind here the implementation is very similar to the commutative polynomial case and will be driven by the Buchberger machinery of Singular. So one can even do the implementation one abstraction layer lower, implementing the Buchberger toolbox (normal form, std, liftstd, syz etc). Certainly some issues will show up by going to the non-commutative setting, but nothing serious, everything is very similar to the commutative setting.

jankoboehm: I agree that we should add to the signatures, instead of generalizing to `NCRing`.

jankoboehm:
I don't really see a use case of a dense vector format in geometry. The machinery works well for mpoly related but should also useful over fields, Euclidean rings etc, where you can also encounter a sparse setting (although some constructs are overkill here). For example in some cohomology computations one has to handle very large, very sparse matrices over fields. On the other hand, I don't think that the dense machine as it exists has to made run over mpoly settings. So the distrinction as it exists is actually pretty well adjusted to the use cases.

fieker:
On Thu, Jun 20, 2024 at 10:20:33AM -0700, Janko Boehm wrote:
> I don't really see a use case of a dense vector format in geometry. The machinery works well for mpoly related but should also useful over fields, Euclidean rings etc, where you can also encounter a sparse setting (although some constructs are overkill here). For example in some cohomology computations one has to handle very large, very sparse matrices over fields. On the other hand, I don't think that the dense machine as it exists has to made run over mpoly settings. So the distrinction as it exists is actually pretty well adjusted to the use cases.
> 
The "global assumption" is, as far as I can see
 - dense: you have a working rref: so field or euc in reality
 - sparse: you have a working Groebner (+Syz), so mpoly in reality
of course, one can add the mpoly case to dense and the euc case to
sparse...

The cohomology: this is a generic problem, our (low-level) lin alg (det,
rref, solve) should automatically switch between sparse and dense
depending on input. For hnf over Z this is happening already...

fingolfin:
Replaced by PR #3900

PR #3879, opened 2024-06-20
Improve type stability of tropical code

fingolfin:
Also remove some outdated claims from docstrings.

@YueRen this is far from ideal as e.g. specifying a return type of `Vector` is still imprecise and in particular this is not a concrete type. I tried being more specific but I found no clear patterns, and in fact in general no way to deduce from just the type of e.g. a `TropicalLinearSpace` what kind of polynomials are stored in it. Perhaps that's a hint that there should be some more type parameters. Or perhaps it is OK and just the best we can do, also fine by me.

### DIFF
--- a/src/TropicalGeometry/hypersurface.jl
+++ b/src/TropicalGeometry/hypersurface.jl
@@ -189,7 +189,7 @@ Return the polynomial over a valued field used to construct `TropH`.  Raises an 
 """
 function algebraic_polynomial(TropH::TropicalHypersurface)
     @req has_attribute(TropH,:algebraic_polynomial) "no algebraic polynomial cached"
-    return get_attribute(TropH,:algebraic_polynomial)
+    return get_attribute(TropH, :algebraic_polynomial)::MPolyRingElem
 end
 
 
@@ -239,7 +239,7 @@ julia> maximal_cells(sop)
 """
 function dual_subdivision(TropH::TropicalHypersurface{minOrMax,true}) where minOrMax
     @req has_attribute(TropH,:dual_subdivision) "no dual subdivision cached"
-    return get_attribute(TropH,:dual_subdivision)
+    return get_attribute(TropH, :dual_subdivision)::SubdivisionOfPoints
 end
 
 
@@ -248,9 +248,9 @@ end
 
 Return the tropical polynomial used to construct `TropH`.  Raises an error if it is not cached.
 """
-function tropical_polynomial(TropH::TropicalHypersurface)
+function tropical_polynomial(TropH::TropicalHypersurface{minOrMax}) where minOrMax
     @req has_attribute(TropH,:tropical_polynomial) "no tropical polynomial cached"
-    return get_attribute(TropH,:tropical_polynomial)
+    return get_attribute(TropH,:tropical_polynomial)::Generic.MPoly{TropicalSemiringElem{minOrMax}}
 end
 
 
--- a/src/TropicalGeometry/hypersurface.jl
+++ b/src/TropicalGeometry/hypersurface.jl
@@ -250,7 +250,7 @@ Return the tropical polynomial used to construct `TropH`.  Raises an error if it
 """
 function tropical_polynomial(TropH::TropicalHypersurface{minOrMax}) where minOrMax
     @req has_attribute(TropH,:tropical_polynomial) "no tropical polynomial cached"
-    return get_attribute(TropH,:tropical_polynomial)::Generic.MPoly{TropicalSemiringElem{minOrMax}}
+    return get_attribute(TropH,:tropical_polynomial)::mpoly_type(TropicalSemiringElem{minOrMax})
 end
 
 
--- a/src/TropicalGeometry/linear_space.jl
+++ b/src/TropicalGeometry/linear_space.jl
@@ -374,13 +374,9 @@ end
 @doc raw"""
     pluecker_indices(TropL::TropicalLinearSpace)
 
-Return the Pluecker indices used to construct `TropL`.  Raises an error, if it is not cached.
+Return the Pluecker indices used to construct `TropL`.
 """
-function pluecker_indices(TropL::TropicalLinearSpace)
-    if has_attribute(TropL,:pluecker_indices)
-        return get_attribute(TropL,:pluecker_indices)
-    end
-
+@attr Vector{Vector{Int}} function pluecker_indices(TropL::TropicalLinearSpace)
     # there are two ways to compute the pluecker indices:
     # - from the algebraic matrix
     # - from the tropical matrix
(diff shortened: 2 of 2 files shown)


src/TropicalGeometry/hypersurface.jl
```diff
@@ -248,9 +248,9 @@ end
 
 Return the tropical polynomial used to construct `TropH`.  Raises an error if it is not cached.
 """
-function tropical_polynomial(TropH::TropicalHypersurface)
+function tropical_polynomial(TropH::TropicalHypersurface{minOrMax}) where minOrMax
     @req has_attribute(TropH,:tropical_polynomial) "no tropical polynomial cached"
-    return get_attribute(TropH,:tropical_polynomial)
+    return get_attribute(TropH,:tropical_polynomial)::Generic.MPoly{TropicalSemiringElem{minOrMax}}
```
lgoettgens:
```suggestion
    return get_attribute(TropH,:tropical_polynomial)::mpoly_type(TropicalSemiringElem{minOrMax})
```
we have a nice getter for this

src/TropicalGeometry/linear_space.jl
```diff
@@ -459,35 +441,30 @@ end
 
 Return the tropical matrix used to construct `TropL`.  Raises an error, if it is not cached.
 """
-function tropical_matrix(TropL::TropicalLinearSpace)
+function tropical_matrix(TropL::TropicalLinearSpace{minOrMax}) where minOrMax
     @req has_attribute(TropL,:tropical_matrix) "no tropical matrix cached"
-    return get_attribute(TropL,:tropical_matrix)
+    return get_attribute(TropL, :tropical_matrix)::Generic.MatSpaceElem{TropicalSemiringElem{minOrMax}}
```
lgoettgens:
```suggestion
    return get_attribute(TropL, :tropical_matrix)::dense_matrix_type(TropicalSemiringElem{minOrMax})
```

YueRen (dismissed):

YueRen:
One question: What are the differences between 
* `MPolyRingElem` and `mpoly_type`?
* `MatElem`, `MatSpaceElem`, and `dense_matrix_type`?

They seem functionally similar to me.

joschmitt:
> One question: What are the differences between
> 
>     * `MPolyRingElem` and `mpoly_type`?
> 
>     * `MatElem`, `MatSpaceElem`, and `dense_matrix_type`?
> 
> 
> They seem functionally similar to me.

`MPolyRingElem` is an abstract type that has all the concrete types for multivariate polynomials as subtypes
```
help?> MPolyRingElem
search: MPolyRingElem fpMPolyRingElem ZZMPolyRingElem QQMPolyRingElem FqMPolyRingElem FpMPolyRingElem MPolyQuoRingElem MPolyDecRingElem zzModPolyRingElem ZZModPolyRingElem LaurentMPolyRingElem zzModMPolyRingElem

  No documentation found.

  Summary
  ≡≡≡≡≡≡≡

  abstract type MPolyRingElem{T}

  Subtypes
  ≡≡≡≡≡≡≡≡

  AbstractAlgebra.Generic.MPoly{T<:RingElement}
  FpMPolyRingElem
  FqMPolyRingElem
  LazyPoly{T<:RingElement, PR<:MPolyRing{T<:RingElement}}
  MPolyDecRingElem{T, S}
  QQMPolyRingElem
  Singular.spoly{T<:RingElem}
  ZZMPolyRingElem
  fpMPolyRingElem
  fqPolyRepMPolyRingElem
  zzModMPolyRingElem
```
`mpoly_type(T)` gives the concrete type for a ring or field of type `T`.

I would use `MPolyRingElem` in the signature of a function, so for example `MPolyRingElem{<:FieldElem}` for any multivariate polynomial over a field. To initialize a `Vector` or to specify the type in this `attr` business, I would use the `mpoly_type(T)` function.

It's the same for `MatElem` vs `dense_matrix_type`. `MatSpaceElem` doesn't exist anymore I think?

fingolfin:
`AbstractAlgebra.Generic.MatSpaceElem` is a generic matrix implementation in AA which is used as fallback whenever we don't have a better implementation.

fingolfin:
@YueRen if you are OK with this, please approve the PR :-)

src/TropicalGeometry/hypersurface.jl
```diff
@@ -248,9 +248,9 @@ end
 
 Return the tropical polynomial used to construct `TropH`.  Raises an error if it is not cached.
 """
-function tropical_polynomial(TropH::TropicalHypersurface)
+function tropical_polynomial(TropH::TropicalHypersurface{minOrMax}) where minOrMax
```
YueRen, line 251:
```suggestion
function tropical_polynomial(TropH::TropicalHypersurface{minOrMax,true}) where minOrMax
```

`TropicalHypersurface` has two parameters.

src/TropicalGeometry/hypersurface.jl
```diff
@@ -248,9 +248,9 @@ end
...
     @req has_attribute(TropH,:tropical_polynomial) "no tropical polynomial cached"
-    return get_attribute(TropH,:tropical_polynomial)
+    return get_attribute(TropH,:tropical_polynomial)::mpoly_type(TropicalSemiringElem{minOrMax})
```
YueRen, line 253: Is there a reason why we use `mpoly_type` instead of `MPolyRingElem` as in the change in line 192?

src/TropicalGeometry/linear_space.jl
```diff
@@ -459,35 +441,30 @@ end
 
 Return the tropical matrix used to construct `TropL`.  Raises an error, if it is not cached.
 """
-function tropical_matrix(TropL::TropicalLinearSpace)
+function tropical_matrix(TropL::TropicalLinearSpace{minOrMax}) where minOrMax
```
YueRen, line 444:
```suggestion
function tropical_matrix(TropL::TropicalLinearSpace{minOrMax,true}) where minOrMax
```

`TropicalLinearSpace` has two parameters (of which the second is currently always `true`).

src/TropicalGeometry/hypersurface.jl
YueRen: The second parameter is always `true` (inherited from `TropicalVarietySupertype`).

YueRen (changes requested):

src/TropicalGeometry/hypersurface.jl
fingolfin: But one can omit trailing type parameters.

src/TropicalGeometry/hypersurface.jl
lgoettgens: Is in line 192 the coefficient type of this polynomial determined at compile time? if yes, `mpoly_type` should be used there as well

fingolfin: Yes: the code in line 192 is too unstable, so the best we can do there is to say "it will be some kind of multivariate polynomial".

There just is no way to specify a more concrete type *by only using the argument types* -- remember, we want the compiler to be able to deduce the types at *compile time*, not at runtime.

At first I thought the result always was a `QQMPolyRingElem` but in one of your book examples it is a polynomial over a rational function field.

As I suggested in the description of this PR already: to improve this, you'd have to add some type parameter(s) to `TropicalHypersurface`, `TropicalLinearSpace`, and possibly others.

src/TropicalGeometry/hypersurface.jl
fingolfin: (and if they are not used in a given context, it is usually a good idea to omit them, too)

YueRen: Ah, I understand now. Thanks for the clarification!

YueRen (approved):

Issue #3881, opened 2024-06-21
Elliptic fibration book example errors


joschmitt:
The book example in the file `specialized/brandhorst-zach-fibration-hopping/vinberg_3.jlcon` errors both on OSCAR 1.1 and current master:
```
julia> K = QQ;

julia> Kt, t = polynomial_ring(K, :t);

julia> Ktf = fraction_field(Kt);

julia> E = elliptic_curve(Ktf, [0, -t^3, 0, t^3, 0]);

julia> P = E([t^3, t^3]);

julia> Y2 = elliptic_surface(E, 2, [P]);

julia> S = weierstrass_model(Y2)[1];

julia> basisNSY2, _, NSY2 = algebraic_lattice(Y2);

julia> fibers_in_Y2 = [QQ.(vec(collect(i))) for i in [
        [4   2   0   0   0   0   0   0   0   -4   -4   -8   -7   -6   -5   -4   -3   -2   -1   0],
        [1   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0],
        [5   2   -2   -3   -4   -3   -2   -1   -2   -5   -4   -8   -7   -6   -5   -4   -3   -2   -1   0],
        [4   2   -2   -4   -6   -9//2   -3   -3//2   -7//2   -3   -5//2   -5   -9//2   -4   -7//2   -3   -5//2   -2   -3//2   0],
        [2   1   -1   -2   -3   -2   -1   0   -2   -1   -1   -2   -2   -2   -2   -2   -2   -1   0   1],
        [2   1   0   0   0   0   0   0   0   -2   -2   -4   -4   -4   -4   -3   -2   -1   0   1]
       ]];

julia> @assert all(inner_product(ambient_space(NSY2), i,i) == 0 for i in fibers_in_Y2)

julia> [representative(elliptic_parameter(Y2, f)) for f in fibers_in_Y2[4:6]]
ERROR: AssertionError: found
Stacktrace:
 [1] horizontal_decomposition(X::EllipticSurface{QQField, AbstractAlgebra.Generic.FracFieldElem{…}}, F::Vector{QQFieldElem})
   @ Oscar ~/.julia/dev/Oscar/experimental/Schemes/src/elliptic_surface.jl:1349
 [2] elliptic_parameter(X::EllipticSurface{QQField, AbstractAlgebra.Generic.FracFieldElem{…}}, F::Vector{QQFieldElem})
[...]

PR #3883, opened 2024-06-23
Booktests: try running vinberg3 in CI

benlorenz:
Lets see how long it takes there and whether it succeeds, ref #3881, cc @simonbrandhorst.

### DIFF
--- a/test/book/test.jl
+++ b/test/book/test.jl
@@ -32,7 +32,7 @@ isdefined(Main, :FakeTerminals) || include(joinpath(pkgdir(REPL),"test","FakeTer
 
              # somewhat slow (~300s)
              "cornerstones/polyhedral-geometry/ch-benchmark.jlcon",
-             "specialized/brandhorst-zach-fibration-hopping/vinberg_3.jlcon",
+             #"specialized/brandhorst-zach-fibration-hopping/vinberg_3.jlcon",
             ]
 
   dispsize = (40, 130)


benlorenz:
It did indeed fail:
```julia
  julia> [representative(elliptic_parameter(Y2, f)) for f in fibers_in_Y2[4:6]]
- 3-element Vector{AbstractAlgebra.Generic.FracFieldElem{QQMPolyRingElem}}:
-  (y//z)//((x//z)*t)
-  ((y//z) + t^3)//((x//z)*t - t^4)
-  ((y//z) + t^3)//((x//z) - t^3)
+ ERROR: AssertionError: found
+ Stacktrace:
+  [1] horizontal_decomposition(X::EllipticSurface{QQField, AbstractAlgebra.Generic.FracFieldElem{…}}, F::Vector{QQFieldElem})
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/experimental/Schemes/src/elliptic_surface.jl:1349
+  [2] elliptic_parameter(X::EllipticSurface{QQField, AbstractAlgebra.Generic.FracFieldElem{QQPolyRingElem}}, F::Vector{QQFieldElem})
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/experimental/Schemes/src/elliptic_surface.jl:1419
+  [3] (::var\"#23#24\")(f::Vector{QQFieldElem})
+    @ Main.__255 ./none:0
+  [4] iterate
+    @ ./generator.jl:47 [inlined]
+  [5] collect(itr::Base.Generator{Vector{Vector{QQFieldElem}}, var\"#23#24\"})
+    @ Base ./array.jl:834
+  [6] top-level scope
+    @ REPL[48]:1
+ Some type information was truncated. Use `show(err)` to see complete types.
```

And it added about 13 minutes to the booktest job. I guess we could afford this since it still finished before a bunch of other jobs in the same run but it is still quite a lot.

simonbrandhorst:
Thanks for the effort! I agree that 13 minutes is (too?) long for a regular test on every PR. Is there a way to run it once for every release?

simonbrandhorst:
In the sense that we have,
tests, long tests, very long tests (but only very few ones)

simonbrandhorst:
Seems like we can speed up things by a factor of  50 ... so we will not need very long tests. Best wait before taking any further steps.

simonbrandhorst:
Since there will be no fix in the near future, I close it. Thanks @benlorenz  for bringing this up.

PR #3884, opened 2024-06-23
speedup elliptic surfaces

simonbrandhorst:
50x Speedup. Therefore please backport. 
The speedup is relying on the lazy data structures of @HechtiDerLachs 
which are now further exploited.

### DIFF
--- a/experimental/Schemes/src/IdealSheaves.jl
+++ b/experimental/Schemes/src/IdealSheaves.jl
@@ -913,8 +913,9 @@ function maximal_associated_points(
     # we can at least remember some partial information 
     for U in patches(covering)
       if isone(I(U))
+        I_one = ideal(OO(U), one(OO(U)))
         for P in comps
-          object_cache(P)[U] = I(U)
+          object_cache(P)[U] = I_one
         end
       end
     end
--- a/experimental/Schemes/src/IdealSheaves.jl
+++ b/experimental/Schemes/src/IdealSheaves.jl
@@ -869,7 +869,8 @@ function maximal_associated_points(
     I::AbsIdealSheaf; 
     covering=default_covering(scheme(I)), 
     use_decomposition_info::Bool=true,
-    algorithm::Symbol=:GTZ
+    algorithm::Symbol=:GTZ,
+    mode::Symbol=:save_gluings
   )
   X = scheme(I)
   comps = AbsIdealSheaf[]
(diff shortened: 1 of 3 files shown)


experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -899,8 +900,27 @@ function maximal_associated_points(
     end
   end
 
+  if mode==:save_gluings && length(comps)==1
+    R = radical(I)
```
afkafkafk13, line 904: I assume that radical first checks whether the ideal sheaf is prime and then does not compute anything, right? So this is about storing the right attributes.

experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -899,8 +900,27 @@ function maximal_associated_points(
...
+    set_attribute!(R, :is_prime=>true)
```
afkafkafk13, line 905: Shouldn't this be set automatically for PrimeIdealSheafFromChart? Shouldn't this be inherited through radical in this case?

HechtiDerLachs (approved):
Looks good. Thanks for the effort!

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -948,15 +948,46 @@ end
 
 Return the fiber components of the fiber over the point $P \in C$.
 """
-function fiber_components(S::EllipticSurface, P)
+function fiber_components(S::EllipticSurface, P; algorithm=:exceptional_divisors)
```
afkafkafk13, line 951: Don't you want to state what type P should have? You expect it to be of a certain form later on.

If this P is an automatically generated object from further up, then you might want to give it an attribute which allows you a sanity check inside the method before continuing.

afkafkafk13 (commented):
Looks good overall. Please answer the (minor) remarks before I approve this. (They do not concern the functionality of the code, but the step from 'solves my problem' to 'is safe to use also by some others'.)

experimental/Schemes/src/IdealSheaves.jl
simonbrandhorst: Not quite. Since `length(comps)==1`, the support of `I` is irreducible. Therefore the radical of `I` is prime.

experimental/Schemes/src/IdealSheaves.jl
simonbrandhorst: `radical(I)` will be a `RadicalIdealSheaf` and therefore the type does not know that it is prime.

experimental/Schemes/src/elliptic_surface.jl
simonbrandhorst: `P` should be a rational point on a covered scheme. But there is no type/interface for these yet. There is a similar mess in other places.

experimental/Schemes/src/IdealSheaves.jl
afkafkafk13: The maths was clear from the beginning. 
From the comment below I deduce that this is to move it to the type RadicalIdealSheaf, which is fine with me and confirms my guess (but in a slightly different way). I would still set is_prime first and then do the radical -- and make sure that is_prime is checked and transferred to the new object by radical, if applicable.

experimental/Schemes/src/IdealSheaves.jl
afkafkafk13: see above.

experimental/Schemes/src/elliptic_surface.jl
afkafkafk13: ok. This needs to be cleaned up at some point, but definitely not now. I agree.

experimental/Schemes/src/IdealSheaves.jl
HechtiDerLachs: I was wondering whether the decomposition leading to `comp` did not produce further local representatives of the prime ideal sheaf already . Maybe it's worth caching those results in the `RadicalIdealSheaf` already?

afkafkafk13: If we have the space for it....

PR #3886, opened 2024-06-25
Matroid serialisation and matroid encodings

Sequenzer:
This adds the ability to save matroids using the `mrdi` file format. Currently this only allows you to save matroids whose groundset is of type `Vector{Int}`, arbitrary groundsets are not possible with this implementation. Maybe something for @antonydellavecchia to look into.

This also adds a functions, `matroid_hex`, which is a way of encoding a matroid into a string using revlex_bases_encoding.

`matroid_hex` is an encoding used in the paper [Quantum automorphisms of matroids](https://arxiv.org/abs/2312.13464).

The encoding style also gets a function to retrieve the matroid from the encoded string `matroid_from_matroid_hex`. Suggestions for the names of the functions are welcome :)

This is part of the step-by-step merge of the [QuantumAutomorphismGroups.jl](https://github.com/dmg-lab/QuantumAutomorphismGroups.jl) package. 

@danteluber Please Look at this!


src/Combinatorics/Matroids/properties.jl
```diff
@@ -1030,6 +1030,133 @@ function min_revlex_basis_encoding(M::Matroid)
...
+
+@doc raw"""
+    matroid_from_matroid_hex(str::AbstractString)
+
+Returns a matroid from a string of hex characters.
+
+# Examples
+To retrieve the fano matroid from its hex encoding write:
+
+```
```
benlorenz:
```suggestion
```jldoctest
```

test/Combinatorics/Matroids/Matroids.jl
```diff
@@ -368,4 +368,13 @@
...
    end
+    
+  @testset "matroid6 and matroid_hex" begin
+    M = fano_matroid() 
+    N = uniform_matroid(2, 4)
+    M1 = matroid_from_matroid6(matroid6(M))
+    N1 = matroid_from_matroid6(matroid6(N))
+    @test is_isomorphic(M, M1)
+    @test is_isomorphic(N, N1)
+  end
```
lgoettgens, line 380: This testblock claims to test both encodings, but only contains tests for `matroid6`. Could you add corresponding tests for `matroid_hex` as well?

src/Combinatorics/Matroids/properties.jl
```diff
@@ -1030,6 +1030,58 @@ function min_revlex_basis_encoding(M::Matroid)
...
+function matroid_hex(M::Matroid)::String
+  rvlx = min_revlex_basis_encoding(M)
+  r,n = rank(M), length(M) 
+
+  v = _revlex_basis_to_vector(rvlx)
+  for _ in 1:(4-length(v)%4)
+    pushfirst!(v,0)
+  end
+  v = reshape(v,4,:)
+  v = [string(parse(Int,join(v[:,j]),base=2),base=16) for j in 1:size(v)[2]]
```
antonydellavecchia: you should in general be a little more generous with your spaces, see https://docs.oscar-system.org/stable/DeveloperDocumentation/styleguide/#Whitespace
```suggestion
  v = [string(parse(Int, join(v[:, j]), base=2), base=16) for j in 1:size(v)[2]]
```

src/Serialization/Combinatorics.jl
```diff
@@ -19,6 +19,30 @@ function load_object(s::DeserializerState, g::Type{Graph{T}}) where T <: Union{D
...
+## Matroid
+###############################################################################
+@register_serialization_type Matroid "Matroid"
+
+function save_object(s::SerializerState, m::Matroid) 
+  @req m.groundset isa Vector{Int} "Groundset must be a Vector{Int}"
+
+  save_data_dict(s) do
+    save_object(s, pm_object(m), :matroid)
+    save_object(s, m.groundset, :groundset)
```
antonydellavecchia:
```suggestion
    save_object(s, matroid_groundset(m), :groundset)
```

src/Serialization/Combinatorics.jl
```diff
@@ -19,6 +19,30 @@ function load_object(s::DeserializerState, g::Type{Graph{T}}) where T <: Union{D
...
+    save_object(s, m.gs2num, :gs2num)
```
antonydellavecchia:
```suggestion
    save_object(s, create_gs2num(m), :gs2num)
```

src/Combinatorics/Matroids/properties.jl
```diff
@@ -1030,6 +1030,58 @@ function min_revlex_basis_encoding(M::Matroid)
...
+
+```jldoctest
+julia> matroid_from_matroid_hex("r3n7_3f7eefd6f")
+Matroid of rank 3 on 7 elements
+
+```
+"""
+function matroid_from_matroid_hex(str::AbstractString)::Matroid
+  sep = split(str,"_")
+  (r,n) = parse.(Int,split(sep[1][2:end],"n"))
```
antonydellavecchia, line 1078: Have you thought of this? not sure which is better, I do think it might be a little easier to recognize whats going on here with regex.
```suggestion
  (r, n) = string.(match(r"r([0-9]+)n([0-9]+)", sep[1]).captures)
```

Sequenzer: I have thought of this but I believe it's slower.

antonydellavecchia: also, there seems to be no error handling, this function should throw an error when the string isn't valid

src/Serialization/Combinatorics.jl
```diff
@@ -19,6 +19,30 @@ function load_object(s::DeserializerState, g::Type{Graph{T}}) where T <: Union{D
   return g(smallobj)
 end
 
+###############################################################################
+## Matroid
+###############################################################################
+@register_serialization_type Matroid "Matroid"
+
+function save_object(s::SerializerState, m::Matroid) 
+  @req m.groundset isa Vector{Int} "Groundset must be a Vector{Int}"
```
antonydellavecchia:
```suggestion
  @req groundset(m) isa Vector{Int} "Groundset must be a Vector{Int}"
```

src/Combinatorics/Matroids/properties.jl
Sequenzer: I have added some error handling to this function. But I can only check if the syntax is correct, it is impossible to check if its a matroid without constructing it.

antonydellavecchia (approved):

thofma:
I am not sure if the "draft" status means that more comments are welcome, but here is a minor comment. The function signatures are sometimes annotated with an explicit `convert` on the return type, e.g.,
```
function matroid_hex(M::Matroid)::String
```
We usually don't do this and I think in none of the cases here it is necessary.

Issue #3850, opened 2024-06-12
Combinatorics: compositions, partitions, Young tableaux, Schur polynomials


JohnAAbbott:
**Describe the bug**
Several minor problems (see _To reproduce_ below)

**To Reproduce**
Steps to reproduce the behavior, please provide a code snippet that triggers
the bug.
```
using Oscar
# Many problems when trying to specify Int8 entries in partitions/compositions

# number_of_compositions(n)  seems OK  for n Int, Int8, ZZRingElem etc

# compositions(n)  initially OK  for n Int, Int8, ZZRingElem etc  **BUT**

collect(compositions(Int8(4)))     # Error: cannot convert
collect(compositions(ZZ(4)))       # works!

compositions(Int8(4), 2)           # Error: cannot convert
compositions(Int8(4), Int8(2))     # Error: cannot convert
collect(compositions(4, Int8(2)))  # OK :-)

# weak_compositions seem to be OK
collect(weak_compositions(Int8(4), 2))  # OK :-)


# Trouble with filter and partitions iterator

iter = partitions(25);
filter(c -> length(c) == 1, iter)  # MethodError


coins = [1, 2, 5, 10, 20, 50];   # Euro coins
iter = partitions(99, coins);    # iterator over ways of giving 99 cents change
filter(c -> length(c) < 8, iter) # --> MethodError



tab = young_tableau([ [1,2,3,4], [5,6], [7] ])
H = hook_lengths(tab)  # MethodError   (!unfortunate!)
H = hook_lengths(shape(tab));
weight(H)  # BoundsError

tab1 = young_tableau([[-1]]);
is_standard(tab1)      # BoundsError
is_semistandard(tab1)  # works OK

# iterators: just generate a big list internally, not "proper" iterators
std_tab = standard_tableaux(13); # just generates a big list (I guess)
std_tab  # gives Base.Generator{...}
for t in std_tab
  println(t)   # prints out just "Young tableau" (lots of times)
end
# L = collect(std_tab); println(L[1]); # does print out the tableau

P = partition([4,2,1]);
number_of_standard_tableaux(P)
std_tab = standard_tableaux(P);
std_tab[1]   # MethodError

L = collect(std_tab);
L[1]  # works as desired

### --------------------------------------------

# Schur polynomials

# Manual may be wrong: \lambda is a partition with n parts
# (and not a partition of n).  I would also discourage use
# of the schur_polynomial without the ring (because then each
[...]

PR #3887, opened 2024-06-25
Iteratorify some tableaux

joschmitt:
Turn `semistandard_tableaux(shape)` and `standard_tableaux` into an actual iterator internally.
Also adds tests/fixes for different `Integer` types. `ZZRingElem` doesn't work in general because the entries are often used as indices.

The only function left, which still first fills an array internally and then iterates the entries, is `semistandard_tableaux(shape, weight)`.
This function is unfortunately implemented recursively, so it would need to be reimplemented completely. There is no reference for the used algorithm given; I assume it was implemented by a Bachelor's student in JuLie originally. @ulthiel Is there a reference for this algorithm?

### DIFF
--- a/src/Combinatorics/EnumerativeCombinatorics/tableaux.jl
+++ b/src/Combinatorics/EnumerativeCombinatorics/tableaux.jl
@@ -500,29 +500,29 @@ function semistandard_tableaux(s::Vector{T}, weight::Vector{T}) where T <: Integ
   n_max = sum(s)
   @req n_max == sum(weight) "sum(s) == sum(weight) required"
 
-  tabs = Vector{YoungTableau}()
+  tabs = Vector{YoungTableau{T}}()
   if isempty(s)
-    push!(tabs, young_tableau(Vector{Int}[], check = false))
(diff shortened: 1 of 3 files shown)


JohnAAbbott:
The doc for YTs mentions that the tableaux entries are "usually" (positive) integers in the range `1..n` where `n` is the number of cells in the YT.  This **suggests** two possible modifications to the UI:

1. make the constructor for a YT accept only integers in the range `1..n`; offer a boolean kwarg which allows positive entries from a wider range -- I have asked a few people, and no-one seemed ever to have seen non-positive entries in a YT
2. insist that entries are of type `Int`: in ptic, disallow `ZZRingElem`; allowing `Int8` or other "small machine integer types" seems to bring no benefit, now that iterator are genuine iterators (previously `Int8` brought a minor advantage because less memory was needed for storing the full list)

What do you think about these suggestions?

fingolfin (approved):
Looks good to me

joschmitt:
I copied the open questions to #3850.

PR #3890, opened 2024-06-26
undo most of #3874

ThomasBreuer:
Return to the situation where the *released* version of GAP's `recog` package gets installed if necessary, *not* a current version from github.

We keep the change that the `OscarInterface` package gets loaded in the end, and that it requires the `Polycyclic` package.

I think that once there will be a new released `recog` version, the `install` call for `recog` in Oscar will run into an error unless we are then able to start GAP without loaded packages. The point is that if a version is already loaded that lies in a path different from the path of the newly installed version, the GAP code in `recog` that gets called for processing the documentation will throw an error. This behaviour of `recog` is not nice, the people who invented it did probably not think of the situation that the package gets installed automatically by `PackageManager`.


fingolfin:
Thank you!


> This behaviour of recog is not nice, the people who invented it did probably not think of the situation that the package gets installed automatically by PackageManager

They absolutely did not, because when I added `regen_doc.g` in 2017, PackageManager did not even exist :-). And to this day nobody apparently tried to install recog via PackageManager either (or at least they did not bother to write an issue for `recog`), otherwise "we" would have looked into fixing it long ago. (Though looking at the changes you made, they seem to me more like a workaround for a bug in PackageManager than a fix? Note that makedoc.g in virtually all GAP packages really expects to be run from the directory it lives in; that it apparently often works otherwise as well is just luck (whether good or bad luck is a matter of opinion, after all it hid the bug in PackageManager for a long time)

fingolfin (approved):

PR #3892, opened 2024-06-26
change the handling of the `init` argument in `map_word`

### LINKED ISSUE #3885: `map_word` ignores `init` in some cases
```julia
G = abelian_group(FPGroup, [2])
A = abelian_group(2)
```

then

```julia
map_word(one(G), gens(A), init = zero(A))
```

will fail as it calls for one(A[1]) which does not exist. It ignores the init....

ThomasBreuer:
resolves #3885

- use `init` whenever it is different from `nothing` (this differs from the documented behaviour for the Julia functions `sum` and `prod`)

- extend the `map_word` method for `PcGroupElem` to `SubPcGroupElem`

- add a special `map_word` method where the images are `FinGenAbGroupElem`s, where evaluating means to add images instead of multiplying them

- extend the tests

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -1944,7 +1944,9 @@ and $R_j =$ `imgs[`$i_j$`]`$^{e_j}$.
 
 If `g` is an element of (a subgroup of) a finitely presented group
 then the result is defined as `map_word` applied to a representing element
-of the underlying free group.
+of the underlying free group of `full_group(parent(g))`.
+In particular, `genimgs` are interpreted as the images of the generators
+of this free group, not of `gens(parent(g))`.
 
 If the first argument is a vector `v` of integers $k_i$ or pairs `k_i => e_i`,
 respectively,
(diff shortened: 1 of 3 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -1930,7 +1930,7 @@ end
 
 
 @doc raw"""
-    map_word(g::FPGroupElem, genimgs::Vector; genimgs_inv::Vector = Vector(undef, length(genimgs)), init = nothing)
+    map_word(g::Union{FPGroupElem, SubFPGroupElem}, genimgs::Vector; genimgs_inv::Vector = Vector(undef, length(genimgs)), init = nothing)
```
fieker, line 1933: question: for sub-pc and sub-fp I have to specify images for the generators of the full group or for the generators of the sub group?

ThomasBreuer: For the full group, since the given word refers to the generators of the full group (same for `PcGroup` vs. `SubPcGroup`).
I will change the description in order to emphasize this.

fingolfin (approved):

PR #3893, opened 2024-06-27
Enhance `show` for `IdealGens`

### LINKED ISSUE #3322: MINOR/LOW-PRIORITY: unhelpful compact/supercompact printing of Groebner basis
**Describe the bug**
It is unhelpful to print ``with elements...``:  see the recorded output immediately below
```
Dict{MonomialOrdering, Oscar.IdealGens{QQMPolyRingElem}} with 1 entry:
  degrevlex([x, y]) => Gröbner basis with elements…
```
**To Reproduce**
Steps to reproduce the behaviour, please provide a code snippet that triggers
the bug.
```
using Oscar
P, (x,y) = polynomial_ring(QQ, ["x", "y"]);
I = ideal([x,y]);
groebner_basis(I)
I.gb
```

**Expected behavior**
I suggest either simply  ``Gröbner basis`` or maybe ``Gröbner basis with 2 elements``.
It would also be helpful to indicate a corresponding monomial ordering, but that might not be compact, so should probably only be given when printing out with maximum details (if at all).

**System (please complete the following information):**
Please paste the output of `Oscar.versioninfo(full=true)` below. 
```
OSCAR version 0.15.0-DEV - #master, a91222cd18 -- 2024-02-05 21:24:09 +0000
...
```

**Additional context**
**LOW PRIORITY**
I wonder if similar behaviour occurs in other analogous contexts?

lgoettgens:
Resolves https://github.com/oscar-system/Oscar.jl/issues/3322.

Let's first do some bikeshedding here, and I will only fix the doctests once we are happy with the printing of the example from https://github.com/oscar-system/Oscar.jl/issues/3322. All changes to `Gröbner basis` will then be applied analogously to the cases of `Standard basis` and `Ideal generating system` as well.

current master:
```julia
julia> P, (x,y) = polynomial_ring(QQ, ["x", "y"]);

julia> I = ideal([x,y]);

julia> g = groebner_basis(I)
Gröbner basis with elements
1 -> y
2 -> x
with respect to the ordering
degrevlex([x, y])

julia> [g]
1-element Vector{Oscar.IdealGens{QQMPolyRingElem}}:
 Gröbner basis with elements
1 -> y
2 -> x
with respect to the ordering
degrevlex([x, y])

julia> show(Oscar.terse(stdout), g)
Gröbner basis with elements
1 -> y
2 -> x
with respect to the ordering
degrevlex([x, y])

julia> I.gb
Dict{MonomialOrdering, Oscar.IdealGens{QQMPolyRingElem}} with 1 entry:
  degrevlex([x, y]) => Gröbner basis with elements…
```

Proposal 1:
- keep current printing as detailed mode, only add some indentation (and some sanity things like protecting `Gröbner` from decapitalization)
- oneline mode shows `length` and the ordering
- terse mode only prints `Gröbner basis`
```julia
julia> g = groebner_basis(I)
Gröbner basis with elements
  1 -> y
  2 -> x
with respect to the ordering
  degrevlex([x, y])

julia> [g]
1-element Vector{Oscar.IdealGens{QQMPolyRingElem}}:
 Gröbner basis with 2 elements w.r.t. degrevlex([x, y])

julia> show(Oscar.terse(stdout), g)
Gröbner basis

julia> I.gb
Dict{MonomialOrdering, Oscar.IdealGens{QQMPolyRingElem}} with 1 entry:
  degrevlex([x, y]) => Gröbner basis with 2 elements w.r.t. degrevlex([x, y])
```

Proposal 2:
- as Proposal 1, but remove the enumeration in detailed mode, as this is different from how we do it for other similar things, like printing the ideal `I` in detailed mode
```julia
julia> g = groebner_basis(I)
Gröbner basis with elements
  y
  x
with respect to the ordering
  degrevlex([x, y])

# other printing as in Proposal 1
```

Please write what you think in general about this change, and all small details you are (un)happy with, in particular:
- Proposal 1 vs. 2, i.e. show indices or not?
- print ordering in oneline mode?

Let me ping @fingolfin @ederc @jankoboehm @RafaelDavidMohr @JohnAAbbott as you might have an opinion on this, and please ping other people I might forgot.


fingolfin:
I am fine with either color for the bikeshed.

simonbrandhorst:
I would go for Proposal 2. So far I never needed to know which was the n-th element in a Gröbner basis.
But if someone things this is useful, we can go for 1. I do not have a strong opinion here.

lgoettgens:
I now adapted everything according to 2, and fixed the doctests and the book printing. Could you maybe briefly talk at triage tomorrow about this? I unfortunately cannot attend but I would be happy to hear opinions/approvals from people.

JohnAAbbott:
I personally think that option 2 (not showing the indexes) is fine.  I believe that Wolfram had, at one point, expressed a different opinion.  Are there any examples of other list-like structures which print out as `index -> value`?  I do not regard `Dict` as being list-like.

lgoettgens:
> Are there any examples of other list-like structures which print out as `index -> value`? I do not regard `Dict` as being list-like.

I know of no type (except current IdealGens) neither in Oscar nor julia base doing this.

fingolfin:
So we just talked about this during triage. @jankoboehm prefers Proposal 1, as it helps when the polynomials are longer than a line to find the "start" of each polynomial and also to know the index of polynomial in the list.

@wdecker is not here but @fieker predicts that removing this numbering will not be well-received (and recommends a nuclear shelter for the person doing it, although that's just hearsay! Wolfram is not here after all)

Nobody else had a strong opinion either way. Since this change also technically is "smaller" than proposal 2 (i.e. it is closer to the current state), I would suggest we do proposal 1 at least for now.

wdecker:
As people already predicted, I prefer Proposal 1.

lgoettgens:
This now changed back to proposal 1, i.e. with the indices. I hope that I caught all places in the doctests where this needs to be changed again.

joschmitt (approved):
Yesterday, everyone was happy during the meeting, so let's merge this.

PR #3896, opened 2024-06-28
Add direct sums of Lie algebras

lgoettgens:
~~This currently contains https://github.com/oscar-system/Oscar.jl/pull/3894 and needs to get rebased once that is merged. So please skip the first commit when reviewing this.~~

This is only a first sketch, there will possibly be changes (and documentation) at a later point.
I am aware that this does not (yet) is completely as https://github.com/oscar-system/Oscar.jl/issues/3059 wants it to be, put this is in the big part due to some generic functionality not being there to get used and me thus needing to include some of the generic functions myself (but restricted to the types here). All of this can be significantly simplified once there has been some more progress towards https://github.com/oscar-system/Oscar.jl/issues/3059.


HereAround (approved):
Thank you @lgoettgens looks great to me.

PR #3898, opened 2024-06-29
Banff presentation rebased

HechtiDerLachs:
Rebase of #3891 with some fixes.

### DIFF
--- a/experimental/Schemes/src/IdealSheaves.jl
+++ b/experimental/Schemes/src/IdealSheaves.jl
@@ -1625,10 +1625,11 @@ function produce_object(F::PrimeIdealSheafFromChart, U2::AbsAffineScheme)
       pb_f_res = restricted_map(pb_f)
       @assert domain(pb_f_res) === ambient_coordinate_ring(V2)
       Q = preimage(pb_f_res, F(domain(f)))
-      result = ideal(OO(V2), gens(Q))
-      @assert !isone(Q)
-      set_attribute!(result, :is_prime=>true)
+      rest = OOX(V2, U2)
+      result = ideal(OO(U2), rest.(gens(Q)))
+      @hassert :IdealSheaves 1 !isone(Q)
       @hassert :IdealSheaves 1 is_prime(result)
+      set_attribute!(result, :is_prime=>true)
       return result
       I2 = F(codomain(g))
       I = pullback(g)(I2)
@@ -1693,7 +1694,9 @@ function produce_object(I::PullbackIdealSheaf, U::AbsAffineScheme)
   if any(x->x===U, patches(dom))
     f_loc = f_cov[U]
     V = codomain(f_loc)
-    return pullback(f_loc)(original_ideal_sheaf(I)(V))
+    KK = original_ideal_sheaf(I)(V)
+    @assert base_ring(KK) === OO(V)
+    return pullback(f_loc)(KK)
   end
 
   # We are in a chart below a patch in the domain covering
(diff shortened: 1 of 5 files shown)


HechtiDerLachs:
It looks like some changes introduced here are quite expensive. `vinberg1.jlcon` did terminate, but probably took a long time. I will investigate and do some comparisons.

benlorenz:
> It looks like some changes introduced here are quite expensive. `vinberg1.jlcon` did terminate, but probably took a long time. I will investigate and do some comparisons.

The timings don't look that bad, `vinberg_1` took about 1 minute:
```
Sat, 29 Jun 2024 19:39:02 GMT    vinberg_1.jlcon
....
Sat, 29 Jun 2024 19:40:26 GMT    vinberg_2.jlcon: skip
```
There are quite a few GC pauses in between but that is not unusual.

The booktests ran from `19:15` until it got stuck in `zariski.jlcon` at `19:42`, which is the same time and result as we have on master. It will hopefully be fixed via singular.

HechtiDerLachs:
Thanks for having a look! 

I was afraid that the booktests only got cancelled, because the `vinberg_1.jlcon` had already taken up too much of the time. But indeed I have no trouble running this test locally. 

I also tried the testfile for `elliptic_surfaces.jl` and got the following timings.
Before:
```
606.089320 seconds (3.03 G allocations: 116.741 GiB, 8.48% gc time)

620.284879 seconds (3.03 G allocations: 115.920 GiB, 8.75% gc time)
```
With this branch:
```
601.933683 seconds (2.93 G allocations: 114.751 GiB, 8.59% gc time)

662.196737 seconds (2.84 G allocations: 108.817 GiB, 8.91% gc time)
```
So memory consumption decreased with no significant impact on computing time. I think this is all good then.

experimental/Schemes/src/IdealSheaves.jl
```diff
@@ -1621,6 +1621,16 @@ function produce_object(F::PrimeIdealSheafFromChart, U2::AbsAffineScheme)
...
+      pb_f = pullback(f)::AbsLocalizedRingHom
+      pb_f_res = restricted_map(pb_f)
+      @assert domain(pb_f_res) === ambient_coordinate_ring(V2)
+      Q = preimage(pb_f_res, F(domain(f)))
+      rest = OOX(V2, U2)
+      result = ideal(OO(U2), rest.(gens(Q)))
+      @hassert :IdealSheaves 1 !isone(Q)
+      @hassert :IdealSheaves 1 is_prime(result)
+      set_attribute!(result, :is_prime=>true)
+      return result
```
simonbrandhorst: Please either remove the old code or add an algorithm option if you think that it still performs better in some cases.

experimental/Schemes/src/WeilDivisor.jl
```diff
@@ -479,13 +538,53 @@ function _self_intersection(I::AbsIdealSheaf)
   return get_attribute(I, :_self_intersection)::Int
 end
 
+function _is_known_to_be_one(I::AbsIdealSheaf, U::AbsAffineScheme;
```
simonbrandhorst, line 541: May I suggest to name this function 
`has_is_one(I::AbsIdealSheaf, U::AbsAffineScheme)` 
and in its body to call
`has_is_one(I::Ideal)` ? 
We have the `has_is_...` syntax already in some other places. 
It means: has cached the value already ( or it is trivial to compute).

simonbrandhorst (approved):

experimental/Schemes/src/WeilDivisor.jl
HechtiDerLachs: I saw `has_is_one` being introduced. However, it is something different: Here we ask whether an ideal sheaf is known to be isomorphic to the unit ideal sheaf. `has_is_one` is asking whether it's easy to decide this question. The implementations are different and it would be more work to adapt this here. Since it's an internal function, I suggest to leave this for the moment.

PR #3899, opened 2024-07-01
Add `smaller_degree_permutation_representation`, lazy double cosets

simonbrandhorst:
Adds 
`double_cosets_representatives_and_sizes`
and 
`smaller_degree_permutation_representation`.
If you point me to the right place, I will happily add these functions to the docs.

### DIFF
--- a/src/Groups/cosets.jl
+++ b/src/Groups/cosets.jl
@@ -24,7 +24,7 @@ function _group_coset(G::GAPGroup, H::GAPGroup, repr::GAPGroupElem, side::Symbol
 end
 
 function ==(x::GroupCoset, y::GroupCoset)
-   return x.X == y.X && x.side == y.side
+   return GapObj(x) == GapObj(y) && x.side == y.side
 end
 
 function Base.show(io::IO, ::MIME"text/plain", x::GroupCoset)
@@ -32,7 +32,7 @@ function Base.show(io::IO, ::MIME"text/plain", x::GroupCoset)
   io = pretty(io)
   println(io, "$side coset of ", Lowercase(), x.H)
   print(io, Indent())
-  println(io, "with representative ", x.repr)
+  println(io, "with representative ", representative(x))
   print(io, "in ", Lowercase(), x.G)
   print(io, Dedent())
 end
@@ -44,7 +44,7 @@ function Base.show(io::IO, x::GroupCoset)
   else
     print(io, "$side coset of ")
     io = pretty(io)
-    print(terse(io), Lowercase(), x.H, " with representative ", x.repr)
+    print(terse(io), Lowercase(), x.H, " with representative ", representative(x))
   end
 end
 
@@ -73,8 +73,8 @@ Right coset of Sym(3)
 ```
 """
 function right_coset(H::GAPGroup, g::GAPGroupElem)
-   @req GAPWrap.IsSubset(parent(g).X, H.X) "H is not a subgroup of parent(g)"
-   return _group_coset(parent(g), H, g, :right, GAP.Globals.RightCoset(H.X,g.X))
+   @req GAPWrap.IsSubset(GapObj(parent(g)), GapObj(H)) "H is not a subgroup of parent(g)"
+   return _group_coset(parent(g), H, g, :right, GAPWrap.RightCoset(GapObj(H), GapObj(g)))
 end
 
 """
@@ -101,8 +101,8 @@ Left coset of Sym(3)
 ```
 """
 function left_coset(H::GAPGroup, g::GAPGroupElem)
-   @req GAPWrap.IsSubset(parent(g).X, H.X) "H is not a subgroup of parent(g)"
-   return _group_coset(parent(g), H, g, :left, GAP.Globals.RightCoset(GAP.Globals.ConjugateSubgroup(H.X,GAP.Globals.Inverse(g.X)),g.X))
+   @req GAPWrap.IsSubset(GapObj(parent(g)), GapObj(H)) "H is not a subgroup of parent(g)"
+   return _group_coset(parent(g), H, g, :left, GAPWrap.RightCoset(GAPWrap.ConjugateSubgroup(GapObj(H), GAPWrap.Inverse(GapObj(g))), GapObj(g)))
 end
 
 
@@ -126,24 +126,24 @@ Base.:*(g::GAPGroupElem, H::GAPGroup) = left_coset(H,g)
 function Base.:*(c::GroupCoset, y::GAPGroupElem)
    @assert y in c.G "element not in the group"
    if c.side == :right
-      return right_coset(c.H, c.repr*y)
+      return right_coset(c.H, representative(c)*y)
    else
-      return left_coset(c.H^y, c.repr*y)
+      return left_coset(c.H^y, representative(c)*y)
    end
 end
 
 function Base.:*(y::GAPGroupElem, c::GroupCoset)
    @assert y in c.G "element not in the group"
    if c.side == :left
-      return left_coset(c.H, y*c.repr)
+      return left_coset(c.H, y*representative(c))
    else
-      return right_coset(c.H^(y^-1), y*c.repr)
+      return right_coset(c.H^(y^-1), y*representative(c))
    end
 end
 
 function Base.:*(c::GroupCoset, d::GroupCoset)
    @req (c.side == :right && d.side == :left) "Wrong input"
-   return double_coset(c.H, c.repr*d.repr, d.H)
+   return double_coset(c.H, representative(c)*representative(d), d.H)
 end
 
 """
@@ -238,7 +238,7 @@ julia> is_bicoset(fH)
 true
 ```
 """
-is_bicoset(C::GroupCoset) = GAPWrap.IsBiCoset(C.X)
+is_bicoset(C::GroupCoset) = GAPWrap.IsBiCoset(GapObj(C))
 
 """
     right_cosets(G::GAPGroup, H::GAPGroup; check::Bool=true)
@@ -317,6 +317,8 @@ struct SubgroupTransversal{T<: GAPGroup, S<: GAPGroup, E<: GAPGroupElem} <: Abst
    X::GapObj               # underlying *right* transversal in GAP
 end
 
+GAP.julia_to_gap(T::SubgroupTransversal) = T.X
+
 function Base.show(io::IO, ::MIME"text/plain", x::SubgroupTransversal)
   side = x.side === :left ? "Left" : "Right"
   println(io, "$side transversal of length $(length(x)) of")
@@ -343,7 +345,7 @@ Base.hash(x::SubgroupTransversal, h::UInt) = h # FIXME
 Base.length(T::SubgroupTransversal) = index(Int, T.G, T.H)
 
 function Base.getindex(T::SubgroupTransversal, i::Int)
-  res = group_element(T.G, T.X[i])
+  res = group_element(T.G, GapObj(T)[i])
   if T.side === :left
     res = inv(res)
   end
@@ -397,7 +399,7 @@ function right_transversal(G::T1, H::T2; check::Bool=true) where T1 <: GAPGroup 
      _check_compatible(G, H)
    end
    return SubgroupTransversal{T1, T2, eltype(T1)}(G, H, :right,
-              GAP.Globals.RightTransversal(G.X, H.X))
+              GAPWrap.RightTransversal(GapObj(G), GapObj(H)))
 end
 
 """
@@ -437,11 +439,11 @@ function left_transversal(G::T1, H::T2; check::Bool=true) where T1 <: GAPGroup w
      _check_compatible(G, H)
    end
    return SubgroupTransversal{T1, T2, eltype(T1)}(G, H, :left,
-              GAP.Globals.RightTransversal(G.X, H.X))
+              GAPWrap.RightTransversal(GapObj(G), GapObj(H)))
 end
 
 Base.IteratorSize(::Type{<:GroupCoset}) = Base.SizeUnknown()
-Base.iterate(G::GroupCoset) = iterate(G, GAPWrap.Iterator(G.X))
+Base.iterate(G::GroupCoset) = iterate(G, GAPWrap.Iterator(GapObj(G)))
 
 function Base.iterate(G::GroupCoset, state)
   GAPWrap.IsDoneIterator(state) && return nothing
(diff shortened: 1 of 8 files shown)


src/Groups/cosets.jl
```diff
@@ -575,6 +575,24 @@ function double_cosets(G::T, H::GAPGroup, K::GAPGroup; check::Bool=true) where T
...
 
+raw"""
+    double_cosets_representatives_and_sizes(G::GAPGroup, H::GAPGroup, K::GAPGroup)
+    
+Return representatives of the double coset `H\G/K` and their size. 
+
+This function is faster and consumes less memory than [`double_cosets`](@ref)
+since no double cosets are created and stored.    
+"""
+function double_cosets_representatives_and_sizes(G::T, H::GAPGroup, K::GAPGroup) where T <: GAPGroup
```
fingolfin: First off, I support adding this wrapper if it is something that benefits your work right now!

However, I'd really prefer to not document this so that we can eventually remove it again. That's because the "expensive" part with `DoubleCosets` and `DoubleCosetsNC` really is that creating GAP objects is much more expensive than creating Julia objects, and so is avoided. I mean, this is the implementation:
```gap
InstallGlobalFunction( DoubleCosets, function(G,U,V)
  if not IsSubset(G,U) and IsSubset(G,V) then
    Error("not contained");
  fi;
  return DoubleCosetsNC(G,U,V);
end );

InstallMethod(DoubleCosetsNC,"generic",true,
  [IsGroup,IsGroup,IsGroup],0,
function(G,U,V)
  return List(DoubleCosetRepsAndSizes(G,U,V),i->DoubleCoset(U,i[1],V,i[2]));
end);
```

But this is quite different with Julia objects. So what we should do ultimately (and I explicitly don't mean Simon in this PR, but rather Thomas and myself in a future PR) is to change `double_cosets` to call `DoubleCosetRepsAndSizes` instead of `DoubleCosets`/`DoubleCosetsNC`. GAP does not implement much functionality for double cosets anyway (basically: equality test, membership test, querying size and representative, `rand`, `rand_pseudo`, iteration/enumeration, printing, conversion to a list), and we can just re-implement all of this in a few dozen (or at most few hundred) lines of rather simple code. But initially, we can just store a representative and size; and an optional, "lazy" pointer to a GAP double coset object, which we only created when needed. And then we can re-implement functionality gradually.

We could also change `double_cosets` to return an iterator to make it even more "lazy", but that's separate step.

src/Groups/perm.jl
```diff
@@ -149,6 +149,18 @@ function perm(L::AbstractVector{<:IntegerUnion})
   return PermGroupElem(symmetric_group(length(L)), GAPWrap.PermList(GapObj(L;recursive=true)))
 end
 
+"""
+  smaller_degree_permutation_representation(G::PermGroup) -> PermGroup, map
```
fingolfin:
```suggestion
    smaller_degree_permutation_representation(G::PermGroup) -> PermGroup, map
```

src/Groups/cosets.jl
```diff
@@ -575,6 +575,24 @@ function double_cosets(G::T, H::GAPGroup, K::GAPGroup; check::Bool=true) where T
...
+  dcs = GAP.Globals.DoubleCosetRepsAndSizes(G.X,H.X,K.X)
+  res = Vector{Tuple{elem_type(T), ZZRingElem}}(undef, length(dcs))
+  for i in 1:length(res)
+    g = group_element(G, dcs[i][1])
+    n = GAP.gap_to_julia(ZZRingElem, dcs[i][2])
```
fingolfin:
```suggestion
    n = ZZ(dcs[i][2])
```

src/Groups/perm.jl
```diff
@@ -149,6 +149,18 @@ function perm(L::AbstractVector{<:IntegerUnion})
...
+  
+Return an isomorphic permutation group of smaller degree and the isomorphism.
+"""
+function smaller_degree_permutation_representation(G::PermGroup)
+     mp = GAP.Globals.SmallerDegreePermutationRepresentation(GapObj(G))
+     img = PermGroup(GAP.Globals.Image(mp))
+     return img, GAPGroupHomomorphism(G, img, mp)
+end
+
+simplify(G::PermGroup) = smaller_degree_permutation_representation(G)
```
fingolfin: Why both names? Let's pick one?

simonbrandhorst: `simplify` does not have a specified behavior. It returns whatever a developer considers "simpler" than the input. Would this for instance include reducing the size of a generating set using `small_generating_set`?

But then `simplify` is nice for non-experts who are not aware which kinds of simplifications could be useful.

src/Groups/cosets.jl
simonbrandhorst: Yes, I am in a situation where `double_cosets` ooms my machine at 64 GB. 
It seems that `DoubleCosetsRepsAndSizes` sometimes gets through. Since it looked useful, I wrapped it for the sake of others.

simonbrandhorst: Should I remove it from the exports?

src/Groups/perm.jl
```diff
@@ -149,6 +149,18 @@ function perm(L::AbstractVector{<:IntegerUnion})
   return PermGroupElem(symmetric_group(length(L)), GAPWrap.PermList(GapObj(L;recursive=true)))
 end
 
+"""
+    smaller_degree_permutation_representation(G::PermGroup) -> PermGroup, map
+  
+Return an isomorphic permutation group of smaller degree and the isomorphism.
```
ThomasBreuer: We cannot guarantee a strictly smaller degree. It may happen that the returned isomorphism is the identity map.

src/Groups/perm.jl
ThomasBreuer: As far as I understand, `simplify` is defined for various objects, and user who know about its existence may be tempted to call `simplify` with a group. I think this makes sense only if it is documented in which respect the return value may differ from the input.

For example, it makes a difference whether `simplify` is allowed to return an isomorphic group (that acts in a different way on sets on which the input group acts) or whether `simplify` returns "the same" group (in the sense that it has the same elements) but with a "nicer" set of generators.

It is important that `simplify` does not change its input in place, i.e., there should never be a `simplify!` for groups.

simonbrandhorst: I would say that the only requirement on simplify is to return an isomorphic object plus the isomorphism.
This could be a global definition of simplify for any kind of parent object. 

But if this is controversial, I can just delete it from this PR.

fieker:
On Tue, Jul 02, 2024 at 02:21:19PM -0700, Simon Brandhorst wrote:
> 
> 
> 
> > @@ -149,6 +149,18 @@ function perm(L::AbstractVector{<:IntegerUnion})
>    return PermGroupElem(symmetric_group(length(L)), GAPWrap.PermList(GapObj(L;recursive=true)))
>  end
>  
> +"""
> +  smaller_degree_permutation_representation(G::PermGroup) -> PermGroup, map
> +  
> +Return an isomorphic permutation group of smaller degree and the isomorphism.
> +"""
> +function smaller_degree_permutation_representation(G::PermGroup)
> +     mp = GAP.Globals.SmallerDegreePermutationRepresentation(GapObj(G))
> +     img = PermGroup(GAP.Globals.Image(mp))
> +     return img, GAPGroupHomomorphism(G, img, mp)
> +end
> +
> +simplify(G::PermGroup) = smaller_degree_permutation_representation(G)
> 
> I would say that the only requirement on simplify is to return an isomorphic object plus the isomorphism.
> This could be a global definition of simplify for any kind of parent object. 
> 
> But if this is controversial, I can just delete it from this PR. 

I'd say it mildly controversal: reducing the number of generators while
keeping the degree is also one of my hobbies...

ThomasBreuer:
Concerning `simplify`:
If the idea is to get some "nicer" isomorphic group and the isomorphism to it then it would be natural to switch to a pc group from a permutation group, provided that this permutation group is solvable. However, then the type of the result of `simplify` would not be known in advance.

I think that if `simplify` shall be supported for groups then we have to document in each case what `simplify` means.
If there are several aspects that can be addressed (degree, number of generators) then one could think about an optional argument that specifies the wishes.

src/exports.jl
```diff
@@ -483,6 +483,7 @@ export domain_covering
 export dominates
 export double_coset
 export double_cosets
+export double_cosets_representatives_and_sizes
```
fingolfin: As suggested by Simon, let's just not export this
```suggestion
```

src/Groups/perm.jl
fingolfin: Indeed this docstring should clarify that this is a randomized algorithm. So repeatedly calling this can give different results. And as Thomas suggested, it might even fail to produce any reduction.

There is also a `cheap` option in GAP, which we could expose here as well.

I would suggest looking at the GAP docstring for this function and then liberally borrowing from that, possibly even the example for a `jldoctest`.

If you need this *now* we can of course also do these improvements later, and for now just add a `TODO` comment suggesting these improvements, i.e.: `TODO: implement suggestions from comments on PR #3899` or so

simonbrandhorst:
@fingolfin I followed your suggestion and delayed the creation of the gap object until asked for.

There are two points I am not sure about:
I had to turn `GroupDoubleCoset` into a `mutable struct` to allow for later storing of the corresponding gap object.
Could this have performance implications?

I decided to store the size of the double coset. However, I was unable to predict the output type of `GAPWrap.Size(SomeDoubleCoset)`. Sometimes it is `Int64` and sometimes a gap integer.
We could always convert to `ZZRingElem`. But again this seems too expensive to do for 10^6 cosets.

@ThomasBreuer would you please take over from here?

fingolfin:
Something went wrong there, this PR contains many commits that are already on master -- perhaps you can rebase it on current master once more (i.e. on commit 5f6b05e632e8cf6331371528e410d62cde0a7494 at the time I am writing this)

src/Groups/cosets.jl
```diff
@@ -457,22 +457,42 @@ end
...
 """
-struct GroupDoubleCoset{T <: GAPGroup, S <: GAPGroupElem}
+mutable struct GroupDoubleCoset{T <: GAPGroup, S <: GAPGroupElem}
 # T=type of the group, S=type of the element
    G::T
    H::GAPGroup
    K::GAPGroup
    repr::S
    X::GapObj
+   size
```
fingolfin: Either do
```suggestion
   size::GapInt
```
or
```suggestion
   size::ZZRingElem
```
I tend to prefer the latter.

src/Groups/cosets.jl
```diff
@@ -457,22 +457,42 @@ end
 Group double coset.
 Two double cosets are equal if, and only if, they contain the same elements.
 """
-struct GroupDoubleCoset{T <: GAPGroup, S <: GAPGroupElem}
+mutable struct GroupDoubleCoset{T <: GAPGroup, S <: GAPGroupElem}
```
fingolfin: An alternative to making the whole struct `mutable` would be to change the field `X` to `X::Ref{GapObj}`. Then one can initialize that field to `Ref{GapObj}()` and later assign a GAp object via `my_double_coset.X[] = some_gap_obj`

src/Groups/cosets.jl
```diff
@@ -640,7 +671,7 @@ function intersect(V::AbstractVector{Union{<: GAPGroup, GroupCoset, GroupDoubleC
    else
       G = V[1].G
    end
-   l = GAP.Obj([v.X for v in V])
+   l = GAP.Obj([GapObj(v) for v in V])
```
fingolfin: Would this also work (untested)?
```suggestion
   l = GAP.Obj(V; recursive = true)
```

src/Groups/cosets.jl
```diff
@@ -558,31 +578,42 @@ julia> double_cosets(G,H,K)
...
     order(C::Union{GroupCoset,GroupDoubleCoset})
 
 Return the cardinality of the (double) coset `C`.
 """
-order(C::Union{GroupCoset,GroupDoubleCoset}) = GAPWrap.Size(C.X)
-Base.length(C::Union{GroupCoset,GroupDoubleCoset}) = GAPWrap.Size(C.X)
+order(C::Union{GroupCoset,GroupDoubleCoset})
+
+function order(C::GroupCoset)
+  return GAPWrap.Size(GapObj(C))
```
fingolfin: This is actually arguable wrong (not your fault, of course, it is like this in the existing code), as you noticed elsewhere. The result will either be an Int, or a GAP integer which is useless to most code. 

The proper solution here is to do the same as we did for `order` for e.g. GAP groups: `order` always returns a `ZZRingElem`. And for speed fanatics, who "know what they are doing" we can offer a method that allows specifying the return type. As in:
```julia
function order(::Type{T}, x::Union{GAPGroupElem, GAPGroup}) where T <: IntegerUnion
   ord = GAPWrap.Order(GapObj(x))
   if ord === GAP.Globals.infinity
      throw(InfiniteOrderError(x))
   end
   return T(ord)
end
```

src/Groups/cosets.jl
```diff
@@ -457,22 +457,42 @@ end
...
+  obj.X = GAP.Globals.DoubleCoset(H, x, K)
+  return obj.X
+end
 
 Base.hash(x::GroupDoubleCoset, h::UInt) = h # FIXME
 Base.eltype(::Type{GroupDoubleCoset{T,S}}) where {T,S} = S
 
 function ==(x::GroupDoubleCoset, y::GroupDoubleCoset)
-   return x.X == y.X
+   # TODO:can we avoid creation of a gap object here?
```
fingolfin: Sure we can, though it's not clear to me whether it would be worth the effort for now.

But one could add some cheap pretest(s): if the order of the two double cosets is known and differs, we can immediately return `false`

ThomasBreuer:
Concerning the question
> @ThomasBreuer would you please take over from here?

Yes, I can do this. Currently this looks to me like first addressing the comments added by @fingolfin.

simonbrandhorst:
@fingolfin  I reset the index to master. Should be fine now.

@ThomasBreuer Thank you! Let me know if you need anything from me.

simonbrandhorst:
Thank you. Looks good to me. (Can't approve my own PR though).

src/Groups/cosets.jl
```diff
@@ -558,31 +572,42 @@ julia> double_cosets(G,H,K)
...
-Base.length(C::Union{GroupCoset,GroupDoubleCoset}) = GAPWrap.Size(C.X)
+order(C::Union{GroupCoset,GroupDoubleCoset}) = order(ZZRingElem, C)
+
+function order(::Type{T}, C::GroupCoset) where T <: IntegerUnion
+  return T(GAPWrap.Size(GapObj(C)))
+end
+
+function order(::Type{T}, C::GroupDoubleCoset) where T <: IntegerUnion
+  if !isassigned(C.size)
+    C.size[] = T(GAPWrap.Size(GapObj(C)))
```
fingolfin:
```suggestion
    C.size[] = ZZRingElem(GAPWrap.Size(GapObj(C)))
```

src/Groups/cosets.jl
```diff
@@ -558,31 +572,42 @@ julia> double_cosets(G,H,K)
...
+  end
+  return C.size[]
```
fingolfin:
```suggestion
  return T(C.size[])
```

src/Groups/perm.jl
```diff
@@ -149,6 +149,28 @@ function perm(L::AbstractVector{<:IntegerUnion})
...
+julia> s, _ = sylow_subgroup(g, 3);
+
+julia> rho = smaller_degree_permutation_representation(s)
+(Permutation group of degree 3 and order 3, Hom: s -> permutation group)
+```
+"""
+function smaller_degree_permutation_representation(G::PermGroup)
+     mp = GAP.Globals.SmallerDegreePermutationRepresentation(GapObj(G))
+     img = PermGroup(GAP.Globals.Image(mp))
+     return img, GAPGroupHomomorphism(G, img, mp)
```
fingolfin:
```suggestion
  mp = GAP.Globals.SmallerDegreePermutationRepresentation(GapObj(G))
  img = PermGroup(GAP.Globals.Image(mp))
  return img, GAPGroupHomomorphism(G, img, mp)
```

fingolfin (approved):

PR #3831, opened 2024-06-06
Construct Lie algebras from root system

lgoettgens:
Adds a constructor `lie_algebra(::Field, ::RootSystem)`, and redirects the constructors given a Dynkin type to this instead of transforming a `GAP.Globals.SimpleLieAlgebra`.
(@HereAround @emikelsons this allows the construction of F4)

To the reviewers: Please don't try to understand the algorithm nor the concrete implementation, I've already spent way to much time doing that myself. And there are pretty comprehensive tests that check that the resulting Lie algebras satisfy the Jacobi identity, and that GAP can detect the Dynkin type again.
Note to GAP: The `iso_oscar_gap` method had to be adapted, because GAP by default decided to take a different Chevalley basis and in particular a different order of simple roots, which made functions like `HighestWeightModule` and `DominantCharacter` (that take a weight in terms of fundamental weights) behave unexpectedly. As a workaround, we now construct and pass a GAP root system.
I am aware of some optimizations in `_struct_consts` and `_N_matrix`, in particular w.r.t. the number of calls to `is_root_with_index` etc. That's something I plan to optimize in the future, but I would be happy to have this already merged prior to that. If you see any other obvious performance improvements, please let me know.

PR #3913, opened 2024-07-04
Make LieAlgebraModule tests better gc-able

lgoettgens:
Adresses the memory issues in 1.11-nightly and nightly tests reported in https://github.com/oscar-system/Oscar.jl/pull/3831#issuecomment-2208905534.

The excessive memory seems to come from the GAP iso adaption in https://github.com/oscar-system/Oscar.jl/pull/3831. Before this PR, the whole computation was done in GAP, the plan for the future is to not need GAP at all, but in the current state a lot of data has to be transferred (once per testcase).
If seems to help to run each testcase in its single scope to make everything garbage collectable right after the testcase ended (instead of currently keeping everything in memory until the `@testset` is over).
Furthermore I disabled the longest running testcase. The correctness of everything should still be covered by the other cases.

Let's hope that all of this helps with the CI failures.

### DIFF
--- a/experimental/LieAlgebras/test/LieAlgebraModule-test.jl
+++ b/experimental/LieAlgebras/test/LieAlgebraModule-test.jl
@@ -711,23 +711,45 @@
 
   @testset "dim_of_simple_module" begin
     # All concrete test results have been computed using the LiE CAS (http://wwwmathlabo.univ-poitiers.fr/~maavl/LiE/) v2.2.2
-    @test (@inferred dim_of_simple_module(
-      Int, lie_algebra(QQ, :A, 6), [1, 3, 5, 0, 1, 0]
-    )) == ZZ(393513120)
-    @test (@inferred dim_of_simple_module(
-      ZZRingElem, lie_algebra(QQ, :B, 7), [7, 2, 5, 1, 0, 2, 6]
-    )) == ZZ(307689492858882008424585750)
-    @test (@inferred dim_of_simple_module(lie_algebra(QQ, :C, 3), [3, 3, 3])) == 262144
-    @test (@inferred dim_of_simple_module(
-      Int128, lie_algebra(QQ, :D, 5), [1, 2, 3, 4, 5]
-    )) == Int128(591080490000)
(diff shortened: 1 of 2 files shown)


benlorenz:
Thanks for the quick improvements, memory usage is still quite large, but both jobs did succeed on the first try. I did a restart of the whole job to be sure.

For 1.11:
```
GC: pause 3754.52ms. collected 6522.264534MB. incr 
Heap stats: bytes_mapped 9474.31 MB, bytes_resident 9302.19 MB,
heap_size 9856.99 MB, heap_target 13168.92 MB, Fragmentation 0.295
Test Summary:                | Pass  Broken  Total     Time
LieAlgebras.LieAlgebraModule | 7431      49   7480  1m16.7s
-> Testing experimental/LieAlgebras/test/LieAlgebraModule-test.jl took: runtime 51.598 seconds + compilation 26.46 seconds + recompilation 0.0 seconds, 29.259 GiB
```
and nightly:
```
GC: pause 3664.34ms. collected 6586.752552MB. incr 
Heap stats: bytes_mapped 9602.34 MB, bytes_resident 9410.86 MB,
heap_size 9853.67 MB, heap_target 13062.80 MB, Fragmentation 0.296
Test Summary:                | Pass  Broken  Total     Time
LieAlgebras.LieAlgebraModule | 7431      49   7480  1m17.2s
-> Testing experimental/LieAlgebras/test/LieAlgebraModule-test.jl took: runtime 52.564 seconds + compilation 25.279 seconds + recompilation 0.721 seconds, 29.219 GiB
```
Previously this file had about 56 GiB of allocations when it did succeed.

Maybe we should also move this test-file to the `long` subset, which includes the more demanding tests and usually finishes first anyway. (Just add it to `test_large` in `test/runtests.jl`).

lgoettgens:
I'll wait for this test run to finish and then put the line into `test_long`. If all then three CI runs succeed without problems, this is good enough for me for the time being.

And I'll move the rewrite of the second part (to no longer need the GAP conversion) a bit up on my priority list.

benlorenz (approved):
thanks!

PR #3916, opened 2024-07-04
add tables of marks

ThomasBreuer:
(I need this for the forthcoming talk at ICMS.)

- code
- tests
- documentation
- and improve the caching of character tables from the library: normalize the key such that the Oscar table gets created just once from the underlying GAP table

### DIFF
--- a/docs/oscar_references.bib
+++ b/docs/oscar_references.bib
@@ -386,6 +386,16 @@
   doi           = {10.1112/S0025579300011773}
 }
 
+@Book{Bur55,
+  author        = {Burnside, W.},
+  title         = {Theory of groups of finite order},
+  mrnumber      = {69818},
+  note          = {2d ed},
+  publisher     = {Dover Publications, Inc., New York},
+  pages         = {xxiv+512},
+  year          = {1955}
+}
+
 @Book{C-MLS20,
   author        = {Cisneros Molina, Jose Luis and Le, Dung Trang and Seade, Jose},
   title         = {Handbook of Geometry and Topology of Singularities {I}},
@@ -1655,6 +1665,15 @@
   publisher     = {Cambridge University Press, Cambridge},
   pages         = {xi+200},
   year          = {1984}
+}
+
+@Misc{MNP24,
+  author        = {Merkwitz, Thomas and Naughton, Liam and Pfeiffer, Götz},
+  title         = {TomLib, The GAP Library of Tables of Marks, Version 1.2.11},
+  note          = {GAP package},
+  year          = {2024},
+  month         = {1},
+  url           = {https://gap-packages.github.io/tomlib/}
 }
 
 @Article{MR20,
--- a/src/Groups/tom.jl
+++ b/src/Groups/tom.jl
@@ -23,14 +23,45 @@ Objects of type `GAPGroupTableOfMarks` support [`get_attribute`](@ref).
   isomorphism::Map  # isomorphism from `group` to a group in GAP
 
   function GAPGroupTableOfMarks(G::Union{GAPGroup, FinGenAbGroup}, tom::GapObj, iso::Map)
+    @req GAPWrap.IsTableOfMarks(tom) "tom must be a GAP table of marks"
     return new(tom, G, iso)
   end
 
   function GAPGroupTableOfMarks(tom::GapObj)
     # group and isomorphism are left undefined
+    @req GAPWrap.IsTableOfMarks(tom) "tom must be a GAP table of marks"
     return new(tom)
   end
 end
+
+
+# If `tblid` is an admissible name for a library character table
+# then translate it to the GAP name of the corresponding
+# library table of marks if there is one.
+function _table_of_marks_name_in_GAP(tblid::String)
+  info = GAPWrap.LibInfoCharacterTable(GapObj(tblid))
+  info === GAP.Globals.fail && return false, ""
+
+  tblid = string(info.firstName::GapObj)
+  info = GAP.Globals.LIBLIST.TOM_TBL_INFO::GapObj
+  pos = findfirst(isequal(GapObj(lowercase(tblid))), Vector{GapObj}(info[2]))
+  pos === nothing && return false, ""
+
+  return true, string(info[1][pos]::GapObj)
+end
+
+# If `tomid` is the GAP name for a library table of marks
+# then translate it to the GAP name of the corresponding
+# library character table.
+# (There is always such a character table.)
+function _character_table_name_in_GAP(tomid::String)
+  info = GAP.Globals.LIBLIST.TOM_TBL_INFO
+  pos = findfirst(isequal(GapObj(lowercase(tomid))), Vector{GapObj}(info[1]))
+  pos === nothing && return false, ""
+  info = GAPWrap.LibInfoCharacterTable(info[2][pos])
+  return true, string(info.firstName)
+end
+
 
 # access to field values via functions
 GapObj(tom::GAPGroupTableOfMarks) = tom.GAPTable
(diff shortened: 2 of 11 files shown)


docs/oscar_references.bib
```diff
@@ -386,6 +386,16 @@ @Article{Bis96
   doi           = {10.1112/S0025579300011773}
 }
 
+@Book{Bur55,
+  author        = {Burnside, W.},
+  title         = {Theory of groups of finite order},
+  mrnumber      = {69818},
+  note          = {2d ed},
```
fingolfin, line 393: Just a guess, but maybe it should be this?
```suggestion
  note          = {2nd ed},
```

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+
+Objects of type `GAPGroupTableOfMarks` support [`get_attribute`](@ref).
+"""
+@attributes mutable struct GAPGroupTableOfMarks <: GroupTableOfMarks
+  GAPTable::GapObj  # the GAP table of marks object
+  group::Union{GAPGroup, FinGenAbGroup}    # the underlying group, if any
+  isomorphism::Map  # isomorphism from `group` to a group in GAP
+
+  function GAPGroupTableOfMarks(G::Union{GAPGroup, FinGenAbGroup}, tom::GapObj, iso::Map)
+    return new(tom, G, iso)
```
fingolfin, line 27: Maybe add here and in the next constructor a check verifying that `tom` actually is a table of marks?

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+4: 1 1 1 1
+```
+"""
+function table_of_marks(G::Union{GAPGroup, FinGenAbGroup})
+  return get_attribute!(G, :table_of_marks) do
+    iso = isomorphism_to_GAP_group(G)
+    gaptom = GAPWrap.TableOfMarks(codomain(iso))
+    return GAPGroupTableOfMarks(G, gaptom, iso)
+  end
+end
```
fingolfin: To make this type stable and also a bit more concise, how about using the `@attr` property:
```suggestion
@attr GAPGroupTableOfMarks function table_of_marks(G::Union{GAPGroup, FinGenAbGroup})
  iso = isomorphism_to_GAP_group(G)
  gaptom = GAPWrap.TableOfMarks(codomain(iso))
  return GAPGroupTableOfMarks(G, gaptom, iso)
end
```

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+
+# A table of marks constructed from a group is stored in this group.
+# Tables of marks from the library are cached in the dictionary
+# `tables_of_marks_by_id`,
+# in order to achieve that fetching the same table twice yields the same
+# object.
```
fingolfin, line 113: Then perhaps use a `AbstractAlgebra.WeakValueCache` (resp. its alias `AbstractAlgebra.CacheDictType`)

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+const tables_of_marks_by_id = Dict{String, Union{GAPGroupTableOfMarks, Nothing}}()
+
+"""
+    table_of_marks(id::String)
+
+Return the table of marks
+for which `id` is an admissible name in GAP's library of tables of marks.
```
fingolfin, line 120: Is there a cheap way to test if a string is an admissible name?

Is there a way to get a list of all admissible names?

docs/oscar_references.bib
ThomasBreuer: If depends.
If we want to be consistent with [MathSciNet](https://mathscinet.ams.org/mathscinet/article?mr=69818) then we should keep it as it is.

fingolfin (approved):
oops I just noticed I never sent the comments I added during the weekend :-(.

Anyway, these are all minor things and could also be addressed in a follow-up PR, if at all

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+  return get!(tables_of_marks_by_id, id) do
+    tom = GAPWrap.TableOfMarks(GapObj(id))
+    tom === GAP.Globals.fail && return nothing
+    if GAP.Globals.HasUnderlyingGroup(tom)
+      G = _oscar_group(GAPWrap.UnderlyingGroup(tom))
+      iso = isomorphism_to_GAP_group(G)
+      return GAPGroupTableOfMarks(G, tom, iso)
+    else
+      return GAPGroupTableOfMarks(tom)
+    end
```
fingolfin, line 144:
```suggestion
    end::GAPGroupTableOfMarks
```
to make it type stable / allow Julia to determine the return type correctly

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+"""
+function character_table(tom::GAPGroupTableOfMarks)
+  tbl = get_attribute!(tom, :character_table) do
+    if GAPWrap.IsLibTomRep(GapObj(tom))
+      info = GAP.Globals.LIBLIST.TOM_TBL_INFO
+      pos = findfirst(isequal(GapObj(lowercase(identifier(tom)))), Vector{GapObj}(info[1]))
+      pos !== GAP.Globals.fail && return character_table(string(info[2][pos]))
+    end
+    return character_table(group(tom))
+  end
```
fingolfin:
```suggestion
  end::GAPGroupCharacterTable
```

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+
+  set_attribute!(tbl, :table_of_marks, tom)
```
fingolfin, line 632: Can't you move that into the do-end block? Then this method could also make use of `@attr`

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+
+julia> table_of_marks(tbl) == table_of_marks(g)
+true
+```
+"""
+function table_of_marks(tbl::GAPGroupCharacterTable)
+  tom = get_attribute!(tbl, :table_of_marks) do
+    isdefined(tbl, :group) && return table_of_marks(group(tbl))
+    return table_of_marks(identifier(tbl))
+  end
```
fingolfin:
```suggestion
  end::GAPGroupTableOfMarks
```

src/Groups/tom.jl
```diff
@@ -0,0 +1,613 @@
...
+
+  set_attribute!(tom, :character_table, tbl)
```
fingolfin, line 656: As above, if this was moved into the do-end block, this method could use `@attr`

src/Groups/tom.jl
ThomasBreuer: O.k.

src/Groups/tom.jl
ThomasBreuer: O.k. (as a continuation of #3927)

src/Groups/tom.jl
ThomasBreuer: o.k.

src/Groups/tom.jl
ThomasBreuer: o.k.

src/Groups/tom.jl
ThomasBreuer: yes, this works

src/Groups/tom.jl
ThomasBreuer: Then it may happen that the Julia objects get garbage collected if no variable keeps them alive, and the underlying GAP objects are still cached on the GAP side, thus creating the Julia object anew will yield different Julia objects pointing to identical GAP objects. I do not see a problem with that, but it looks strange.

PR #3919, opened 2024-07-05
Adv/fix fq field elem load

antonydellavecchia:
Fix regression with FqFieldElem


thofma:
Maybe add a `FFF = GF(7)` as a test case to cover the degree one case?

PR #3922, opened 2024-07-06
adjusting code in intersection theory

wdecker:
@thofma There are several places where we have to adjust the code to current OSCAR. Here is a first PR in this direction. Please double check.


thofma:
looks good

thofma (approved):

PR #3923, opened 2024-07-08
TropicalGeometry + PolyhedralGeometry: fixing references in documentation

YueRen:
Beginning fixing some documentation as discussed in https://github.com/oscar-system/Oscar.jl/discussions/3915

Does anybody know how to correctly add a reference to the function `rays` for polyhedral complexes?  Currently, ``[`rays(::PolyhedralComplex)`](@ref)`` links to `rays` for `K3Chamber`.

### DIFF
--- a/src/PolyhedralGeometry/PolyhedralComplex/properties.jl
+++ b/src/PolyhedralGeometry/PolyhedralComplex/properties.jl
@@ -34,7 +34,7 @@ Return an iterator over the vertices of `PC` in the format defined by `as`. The
 vertices are defined to be the zero-dimensional faces, so if `P` has lineality,
 there are no vertices, only minimal faces.
 
-See also [`minimal_faces`](@ref) and [`rays`](@ref).
+See also [`minimal_faces`](@ref minimal_faces(PC::PolyhedralComplex{T}) where {T<:scalar_types}) and [`rays`](@ref rays(PC::PolyhedralComplex{T}) where {T<:scalar_types}).
 
 Optional arguments for `as` include
 * `PointVector`.
@@ -135,7 +135,7 @@ _vertices(
     vertices_and_rays(PC::PolyhedralComplex)
 
 Return the vertices and rays of `PC` as a combined set, up to lineality. This
-function is mainly a helper function for [`maximal_polyhedra`](@ref).
+function is mainly a helper function for [`maximal_polyhedra`](@ref maximal_polyhedra(PC::PolyhedralComplex{T}) where {T<:scalar_types}).
 
 # Examples
 ```jldoctest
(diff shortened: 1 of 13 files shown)


src/TropicalGeometry/variety_supertype.jl
```diff
@@ -43,7 +43,7 @@ end
 @doc raw"""
     ambient_dim(TropV::TropicalVariety)
 
-Return the ambient dimension of `TropV`.  Requires `TropV` to be embedded.
+See [`ambient_dim(::PolyhedralComplex)`](@ref).
```
fingolfin: I don't understand. In how far does the documentation for `ambient_dim(::PolyhedralComplex)` explain the behavior of `ambient_dim(TropV::TropicalVariety)` ? It feels to me as if there is some unspoken connection missing.

YueRen: Right.  The unspoken connection is that tropical varieties are weighted polyhedral complexes.  As such, they should be inheriting all functions that are available to polyhedral complexes.

It is briefly mentioned at the beginning of the Tropical Varieties / Hypersurface / Linear Spaces / Curves sections, but I'm open to adding that bit of information to the docstrings of all or selected constructors:
https://docs.oscar-system.org/stable/TropicalGeometry/variety/

YueRen:
Modulo bugs that are currently affecting the entire documentation, see https://github.com/oscar-system/Oscar.jl/issues/4009, I think the pull request is ready to be reviewed.  

The goal of the pull request is to make the documentation more explicit whilst keeping it slim by adding references to same-named functions in polyhedral geometry.

YueRen:
@lkastner Could you review the changes for me, please?  I've had to add some bits to the documentation in polyhedral geometry, let me know if you are okay with that.

src/PolyhedralGeometry/PolyhedralComplex/properties.jl
```diff
@@ -253,7 +253,7 @@ translation `p+L`, where `p` is only unique modulo `L`. The return type is a
 dict, the key `:base_points` gives an iterator over such `p`, and the key
 `:lineality_basis` lets one access a basis for the lineality space `L` of `PC`.
 
-See also [`vertices`](@ref) and [`lineality_space`](@ref).
+See also [`vertices(as::Type{PointVector{T}}, PC::PolyhedralComplex{T}) where {T<:scalar_types}`](@ref) and [`lineality_space`](@ref).
```
thofma, line 256:
```suggestion
See also [`vertices(PointVector{FieldElem}}, Polyhedron{FieldElem}})`](@ref) and [`lineality_space`](@ref).
```
I think this should also work. Not sure if it looks nicer though. It is both ugly.

thofma:
```suggestion
See also [`vertices`](@ref vertices(as::Type{PointVector{T}}, PC::PolyhedralComplex{T}) where {T<:scalar_types}) and [`lineality_space`](@ref).
```
This should do the trick

YueRen: One moment, I'll give it a try on my machine (I remember it not working yesterday evening).

YueRen: Works like a charm, not sure what I did yesterday. :-/

thofma: It took me a few tries of adding/removing `` ` `` before I found a working version.

lkastner (approved):

PR #3926, opened 2024-07-10
[ToricVarieties] Fix for blowup corresponding to already existing ray

HereAround:
@HechtiDerLachs @lkastner Does this seem to make sense?


lkastner:
I am a bit confused since you are talking about the identity morphism. But as we have it now, the `star_subdivision` may still return a different fan, even if the ray used exists already:
```
julia> R = matrix(QQ, [[1,0,0],[1,1,0],[1,0,1],[1,1,1]]);

julia> IM = IncidenceMatrix([[1,2,3,4]]);

julia> f = polyhedral_fan(IM, R)
Polyhedral fan in ambient dimension 3

julia> sf = star_subdivision(f, [1,0,0])
Polyhedral fan in ambient dimension 3

julia> n_maximal_cones(sf)
2

julia> n_rays(sf) == n_rays(f)
true
```
This happens if the fan is not simplicial and is an important step when resolving singularities, so I don't think the `star_subdivision` is wrong here. Is this properly dealt with by choosing the identity map?

HechtiDerLachs (approved):
From my side this looks reasonable: Catching the trivial case where blowdown is an isomorphism with a special constructor. We should make sure that this does not unnecessarily break type stability, though. 

But @lkastner 's comments might be valid. I know too little about the internals of `star_subdivision` to say more about that.

experimental/Schemes/src/ToricBlowups/types.jl
```diff
@@ -11,6 +11,28 @@
...
+
+    # Compute position of new ray and verify that it is already among the rays of the given variety v
+    rs = matrix(ZZ, rays(v))
+    position_new_ray = nothing
+    for i in 1:nrows(rs)
+      if new_ray == rs[i, :]
+        position_new_ray = i
+        break
+      end
+    end
```
HechtiDerLachs:
```suggestion
    position_new_ray = findfirst(i->new_ray==rs[i, :], 1:nrows(rs))
```
Wouldn't this work as well? Also one might consider putting this behind some `@check` flag?

HereAround:
> I am a bit confused since you are talking about the identity morphism. But as we have it now, the `star_subdivision` may still return a different fan, even if the ray used exists already:
> 
> ```
> julia> R = matrix(QQ, [[1,0,0],[1,1,0],[1,0,1],[1,1,1]]);
> 
> julia> IM = IncidenceMatrix([[1,2,3,4]]);
> 
> julia> f = polyhedral_fan(IM, R)
> Polyhedral fan in ambient dimension 3
> 
> julia> sf = star_subdivision(f, [1,0,0])
> Polyhedral fan in ambient dimension 3
> 
> julia> n_maximal_cones(sf)
> 2
> 
> julia> n_rays(sf) == n_rays(f)
> true
> ```
> 
> This happens if the fan is not simplicial and is an important step when resolving singularities, so I don't think the `star_subdivision` is wrong here. Is this properly dealt with by choosing the identity map?

Thank you @lkastner. Great catch. I did not think of this.

Also, we currently raise an error if the ray is already contained in the given fan. I have modified this code segment here already, but apparently not yet so that your example runs.

I suggest to check for equality of the new (after executing the blowup) and old variety and issue the identity blowdown whenever they are identical.

And of course, I will include your example in the tests.

I will now modify this PR accordingly.

experimental/Schemes/src/ToricBlowups/types.jl
HereAround: Will try this. Thank you for the suggestion.

HereAround: Just included this in the code.

HechtiDerLachs (commented):
Looks good from my side. If @lkastner is happy, then it's good to go.

fingolfin:
What is the status of this? Note that  #3934 is now merged; does this PR require adjustments for that? Also I think this is waiting for @lkastner to approve (but maybe first something needs to be changed already?!)

HereAround:
Yes, this is waiting for @lkastner to approve. (I forgot to discuss this PR last week in Durham.)

As far as I can tell, no adjustments needed as of right now.

HereAround:
@lkastner @HechtiDerLachs and I just talked about this PR on slack. We realized, that the proposed code in this PR gives an incorrect result in one particular geometry. @lkastner is investigating this a bit more. For the time being, I will revert this PR to a draft, so I cannot be merged (accidentally).

paemurru:
Caught another bug related to this.

```
julia> P3 = projective_space(NormalToricVariety, 3)
Normal toric variety

julia> (x1,x2,x3,x4) = gens(cox_ring(P3))
4-element Vector{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}:
 x1
 x2
 x3
 x4

julia> I = ideal([x2])
Ideal generated by
  x2
  x3

julia> bP3 = domain(blow_up(P3, I))
ERROR: BoundsError: attempt to access 4×3 Polymake.LibPolymake.MatrixAllocated{Polymake.LibPolymake.Rational} at index [5, 1:3]
```

lkastner:
@paemurru Do you know what the exceptional divisor is if I blow up in an existing ray of the fan? There are two cases:
1. The blow up does nothing, then I suspect the exceptional divisor should be 0.
2. The blow up leads to a refinement of the fan, e.g. the fan could consist of only the cone over a square, then blowing up in a ray would add the diagonal. I am not sure what the exceptional divisor is here.

paemurru:
I wrote a longer answer here under point (2)
https://github.com/oscar-system/Oscar.jl/pull/4096#issue-2528450566

In short: we should rename it to `exceptional_prime_divisor` and it should always be a prime divisor (just like when blowup up a prime Cartier divisor the exceptional-divisor-of-the-blowup is the prime Cartier divisor, even though the morphism is an isomorphism). (Perhaps we should have a type `ToricPrimeDivisor` so that this is always guaranteed.)

If anyone needs, we could later add `exceptional_blowup_divisor` for the effective Cartier divisor and `exceptional_locus` for the locus where the morphism is not an isomorphism.

lkastner:
> I wrote a longer answer here under point (2) [#4096 (comment)](https://github.com/oscar-system/Oscar.jl/pull/4096#issue-2528450566)
> 

I am aware of this PR but it does not answer my question. I know that what currently is implemented may not be exceptional. So my question is: What is the correct exceptional divisor in this situation? Is there one at all?

HereAround:
Sitting with @HechtiDerLachs at our OSCAR workshop. Here is our take on the matter @lkastner:


There are two notions of "exceptional divisor" which are mathematically different a priori: 
 * a) the locus where the blowup morphism is not an isomorphism
 * b) the locally principal ideal sheaf `J` which comes out of blowing up some ideal sheaf `I` downstairs with the standard `Proj` construction (see Hartshorne). This is a delicate version of a pullback.

@paemurru : Could you briefly say which names these two should have in your opinion?

As for the cases discussed, here is what we believe a) and b) should be:

> @paemurru Do you know what the exceptional divisor is if I blow up in an existing ray of the fan? There are two cases:
> 
>     1. The blow up does nothing, then I suspect the exceptional divisor should be 0.

For a) @HechtiDerLachs  does not remember whether we have agreed on anything in the toric world, yet. @HereAround suggests, this should be a toric divisor (and is currently implemented as such). The question is: Is this mathematically correct if we return the zero divisor here whenever the blowup does not change the geometry? If not: What would we need to change?  The zero divisor is not prime. Is that a problem?

For b) we have to look things up in Hartshorne again. This can probably be postponed?

> 
>     2. The blow up leads to a refinement of the fan, e.g. the fan could consist of only the cone over a square, then blowing up in a ray would add the diagonal. I am not sure what the exceptional divisor is here.

The following is for the case that the refinement does *not* produce a new ray:

For b) @HechtiDerLachs and @HereAround both suggest that in this case the exceptional divisor for b) should be the Cartier divisor associated to the ideal sheaf for the old ray in the new fan. That should be the toric divisor for that ray. 

For a) We have to determine what is the correct notion for a) in this case. @HereAround and @HechtiDerLachs both do not see it right now. 

Finally, in the case of a star refinement which produces a new ray, @HechtiDerLachs and @HereAround believe that the notions a) and b) coincide and the exceptional divsior is the toric divisor associated to the new ray.

paemurru:
> @paemurru : Could you briefly say which names these two should have in your opinion?

(a) `exceptional_locus`, since it is not necessarily a divisor
(b) `exceptional_blowup_divisor`. In the literature, "exceptional divisor" sometimes refers to the exceptional locus, which can be confusing

paemurru:
A general comment: there is no existing definition of "exceptional divisor" of adding a ray to the fan of a toric variety. There is no definition for "exceptional divisor" of weighted star subdivision. So, there is no existing "correct" definition. I do not think it would be useful to have this. On the other hand, the exceptional prime divisor of the added/chosen ray is rather useful.

paemurru:
If you wanted to have a definition of exceptional divisor for a weighted star subdivision (adding a ray or choosing an existing ray), I would first canonically associate a blowup along a Rees algebra to it, and then define the exceptional divisor to be the effective Cartier divisor of that blowup (Definition 3.2.2 of [Rydh, Quek - Weighted blow-ups - Draft 2](https://people.kth.se/~dary/weighted-blowups20220329.pdf)).

One choice for a canonical Rees algebra is the one given by the order of vanishing along the prime divisor corresponding to the chosen ray. In this case, the exceptional divisor would be a positive integer multiple of the prime Weil divisor associated to the chosen ray.

paemurru:
> Is this mathematically correct if we return the zero divisor here whenever the blowup does not change the geometry?

If you want to return the "exceptional locus", this is fine, but if you want to return an "exceptional divisor", it does not make sense to return the zero divisor in this case.

> The following is for the case that the refinement does not produce a new ray:
For b) @HechtiDerLachs and @HereAround both suggest that in this case the exceptional divisor for b) should be the Cartier divisor associated to the ideal sheaf for the old ray in the new fan. That should be the toric divisor for that ray.

The toric divisor corresponding to the ray is not necessarily Cartier.

> Finally, in the case of a star refinement which produces a new ray, @HechtiDerLachs and @HereAround believe that the notions a) and b) coincide and the exceptional divsior is the toric divisor associated to the new ray.

No, these do not coincide in general. For example, take the affine 2-space with rays (1, 0) and (0, 1). If you add the ray with coordinates (a, b) greater than 1 and gcd(a, b) = 1, then the prime divisor corresponding to that new ray is not Cartier.

lkastner:
I made a few things here and there more robust:
- Compare primitive ray generators instead of possibly rational multiples of rays
- `_strict_transform` uses `index_of_new_ray` to access the new variable instead of `eval_poly`
- `==` of `ToricDivisor` now errors if the underlying space is not the same
- `blow_up` now picks non-existent variable name in case of this variable name being used already

HereAround:
> I made a few things here and there more robust:
> 
>     * Compare primitive ray generators instead of possibly rational multiples of rays
> 
>     * `_strict_transform` uses `index_of_new_ray` to access the new variable instead of `eval_poly`
> 
>     * `==` of `ToricDivisor` now errors if the underlying space is not the same
> 
>     * `blow_up` now picks non-existent variable name in case of this variable name being used already

Sounds and looks great to me. I would approve, if I had not opened this PR initially. ;)

In any case - as long as I am concerned - this is good to go.

HereAround:
Thank you for working on this @lkastner . Truly appreciated!

benlorenz (approved):
lgtm

PR #3927, opened 2024-07-10
Add/move some type assertions

fingolfin:
... to allow Julia to infer types. Also make use of `@attr` in a few more places.

### DIFF
--- a/experimental/Schemes/src/FunctionFields.jl
+++ b/experimental/Schemes/src/FunctionFields.jl
@@ -78,7 +78,7 @@ represented by
 function function_field(X::AbsCoveredScheme; check::Bool=true)
   return get_attribute!(X, :function_field) do
     VarietyFunctionField(X, check=check)
-  end::VarietyFunctionField  # TODO: make the type fully concrete
+  end::VarietyFunctionField  # can't make this more concrete just using the type of X
 end
 
 ########################################################################
--- a/experimental/Schemes/src/FunctionFields.jl
+++ b/experimental/Schemes/src/FunctionFields.jl
@@ -78,7 +78,7 @@ represented by
 function function_field(X::AbsCoveredScheme; check::Bool=true)
   return get_attribute!(X, :function_field) do
     VarietyFunctionField(X, check=check)
-  end
+  end::VarietyFunctionField  # TODO: make the type fully concrete
 end
 
 ########################################################################
--- a/src/AlgebraicGeometry/ToricVarieties/cohomCalg/special_attributes.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/cohomCalg/special_attributes.jl
@@ -184,7 +184,7 @@ function all_cohomologies(l::ToricLineBundle)
         
         # read out the result
         stdout = read(out, String)
-        result = [ZZRingElem(parse(Int, c)) for c in split(chop(chop(split(stdout, "{" )[4])), ",")]
+        result = [parse(ZZRingElem, c) for c in split(chop(chop(split(stdout, "{" )[4])), ",")]
         
         # consistency check
         if length(result) != dim(v)+1
@@ -193,7 +193,7 @@ function all_cohomologies(l::ToricLineBundle)
         
         # return result
         return result
-    end
+    end::Vector{ZZRingElem}
 end
 
 
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -1507,7 +1507,7 @@ julia> character_parameters(character_table("S5"))
 julia> character_parameters(character_table("M11"))
 ```
 """
-@attr Vector function character_parameters(tbl::GAPGroupCharacterTable)
+@attr Union{Nothing, Vector} function character_parameters(tbl::GAPGroupCharacterTable)
     GAPt = GapObj(tbl)
     GAPWrap.HasCharacterParameters(GAPt) || return nothing
     paras = Vector{GAP.Obj}(GAPWrap.CharacterParameters(GAPt))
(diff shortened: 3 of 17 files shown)


experimental/Schemes/src/FunctionFields.jl
```diff
@@ -78,7 +78,7 @@ represented by
 function function_field(X::AbsCoveredScheme; check::Bool=true)
   return get_attribute!(X, :function_field) do
     VarietyFunctionField(X, check=check)
-  end
+  end::VarietyFunctionField  # TODO: make the type fully concrete
```
HechtiDerLachs: Thanks for the attempt! 

However, I'm afraid that making the type fully concrete is a bit too much to ask here. We do not have the getters to deduce all the type parameters from the type of the scheme. For instance, there is one chosen chart `U` of `X` which is used to represent elements of this field. Even the type of the `ambient_polynomial_ring` of `U` can not easily be deduced from `X`. But it is used in the type parameters for the function field.

src/Rings/MPolyMap/AffineAlgebras.jl
```diff
@@ -271,7 +271,7 @@ function _groebner_data(F::AffAlgHom)
 
     fs = map(lift, _images(F))
     return S2toT(modulus(S)) + ideal(T, [ gen(T, n + i) - S2toT(fs[i]) for i in 1:m ])
-  end
+  end # TODO: should add a type assertion here
```
HechtiDerLachs:
```suggestion
  end::MPolyIdeal{elem_type(K)}
```
would be my guess.

HechtiDerLachs (approved):
I had a look at the schemes and toric varieties and some parts of the algebra. Thanks for the effort put into this. It looks rather good overall.

ThomasBreuer (approved):
The point seems to be:
Whenever possible, use `@attr` instead of `get_attribute!(...) do`.
If this is not possible then describe the type of the return value via a type assertion.

In situations where I have used attributes, I have always copied the code lines in question from somewhere else. Thus it is helpful if the Oscar code obeys these rules. Perhaps the documentation of attributes (in AbstractAlgebra) should also state the rules.

src/Rings/MPolyMap/AffineAlgebras.jl
joschmitt: I added `::MPolyIdeal{mpoly_type(K)}`.

src/Groups/group_characters.jl
```diff
@@ -1507,13 +1507,11 @@ julia> character_parameters(character_table("S5"))
...
 ```
 """
-function character_parameters(tbl::GAPGroupCharacterTable)
-    return get_attribute!(tbl, :character_parameters) do
-      GAPt = GapObj(tbl)
-      GAPWrap.HasCharacterParameters(GAPt) || return nothing
-      paras = Vector{GAP.Obj}(GAPWrap.CharacterParameters(GAPt))
-      return _translate_parameter_list(paras)
-    end
+@attr Vector function character_parameters(tbl::GAPGroupCharacterTable)
```
joschmitt:
```suggestion
@attr Union{Nothing, Vector} function character_parameters(tbl::GAPGroupCharacterTable)
```

src/Groups/group_characters.jl
```diff
@@ -1537,13 +1535,11 @@ julia> class_parameters(character_table("S5"))
...
 ```
 """
-function class_parameters(tbl::GAPGroupCharacterTable)
-    return get_attribute!(tbl, :class_parameters) do
-      GAPt = GapObj(tbl)
-      GAPWrap.HasClassParameters(GAPt) || return nothing
-      paras = Vector{GAP.Obj}(GAPWrap.ClassParameters(GAPt))
-      return _translate_parameter_list(paras)
-    end
+@attr Vector function class_parameters(tbl::GAPGroupCharacterTable)
```
joschmitt:
```suggestion
@attr Union{Nothing, Vector} function class_parameters(tbl::GAPGroupCharacterTable)
```

lkastner (approved):
I can only speak for the toric part though.

experimental/Schemes/src/FunctionFields.jl
fingolfin:
```suggestion
  end::VarietyFunctionField  # can't make this more concrete just using the type of X
```

fingolfin:
@ThomasBreuer the rule really is: the return types of `get_attribute` or `get_attribute!` are always `Any` so to ensure type stability you should always combine these with type annotations, and *in the right spot* (i.e. applied to their return value, not to the return value of any `do`-block they might call).

Using `@attr` is not necessary for that, but using it arguably makes it a bit easier to get this right, and makes it clear what the pattern is in a method / ensures things are "as usual" (e.g. "method name and attribute name match")

src/AlgebraicGeometry/ToricVarieties/cohomCalg/special_attributes.jl
```diff
@@ -184,7 +184,7 @@ function all_cohomologies(l::ToricLineBundle)
         
         # read out the result
         stdout = read(out, String)
-        result = [ZZRingElem(parse(Int, c)) for c in split(chop(chop(split(stdout, "{" )[4])), ",")]
+        result = [parse(ZZRingElem, c) for c in split(chop(chop(split(stdout, "{" )[4])), ",")]
```
fingolfin, line 187: @HereAround FYI

(I also think `function all_cohomologies(l::ToricLineBundle)` could be changed to use `@attr`, i.e. use `@attr Vector{ZZRingElem} function all_cohomologies(l::ToricLineBundle)` plus the corresponding tweaks to its body. But that would involve changing its indentation (or leave it mis-indented) and since it doesn't give any other immediate benefit, I didn't do it.

PR #3929, opened 2024-07-12
Intersection theory: code correction, new example

wdecker:


experimental/IntersectionTheory/src/Main.jl
```diff
@@ -187,19 +187,20 @@ end
 # AbstractVarietyMap
 #
 @doc raw"""
-    hom(X::AbstractVariety, Y::AbstractVariety, fˣ::Vector)
-    hom(X::AbstractVariety, Y::AbstractVariety, fˣ::Vector, fₓ)
+     hom(X::AbstractVariety, Y::AbstractVariety, fˣ::Vector, fₓ = nothing; inclusion::Bool = false, symbol::String = "x")
```
lgoettgens:
```suggestion
    hom(X::AbstractVariety, Y::AbstractVariety, fˣ::Vector, fₓ = nothing; inclusion::Bool = false, symbol::String = "x")
```
wrong indentation

experimental/IntersectionTheory/src/Main.jl
```diff
@@ -187,19 +187,20 @@ end
...
 In case of an inclusion $i:X\hookrightarrow Y$ where the class of `X` is not
-present in the Chow ring of `Y`, use the argument `inclusion=true`.
-A copy of `Y` will be created, with extra classes added so that one can
-pushforward classes on `X`.
+present in the Chow ring of `Y`, use the argument `inclusion = true`. Then,
+a copy of `Y` will be created, with extra classes added so that one can
+pushforward all classes on `X`.
 """
-function hom(X::AbstractVariety, Y::AbstractVariety, fˣ::Vector, fₓ=nothing; inclusion::Bool=false, symbol::String="x")
+function hom(X::AbstractVariety, Y::AbstractVariety, fˣ::Vector, fₓ = nothing; inclusion::Bool = false, symbol::String = "x")
```
lgoettgens, line 203: According to https://docs.oscar-system.org/dev/DeveloperDocumentation/styleguide/#Unicode, we shouldn't use unicode in identifier names like the $f^x$ here

experimental/IntersectionTheory/src/Main.jl
```diff
@@ -1768,14 +1793,14 @@ function abs_flag(dims::Vector{Int}; base::Ring=QQ, symbol::String="c")
...
-    abstract_flag_variety(F::AbstractBundle, dims::Vector{Int}; symbol::String="c")
+    abstract_flag_variety(F::AbstractBundle, dims::Int...; symbol::String = "c")
+    abstract_flag_variety(F::AbstractBundle, dims::Vector{Int}; symbol::String = "c")
 
 Given integers, say, $d_1, \dots, d_{k}$ or a vector of such integers, and given an
 abstract bundle $F$, return the abstract flag variety (flag bundle) of nested sequences 
 of subspaces of dimensions $d_1; \dots, d_{k}$ in the fibers of $F$.
 """
-function abstract_flag_variety(F::AbstractBundle, d::Int; symbol::String="c") abstract_flag_variety(F, [d], symbol=symbol) end
+function abstract_flag_variety(F::AbstractBundle, dims::Int...; symbol::String = "c") abstract_flag_variety(F, collect(dims), symbol=symbol) end
```
lgoettgens: for readability it would be better to have 
```suggestion
function abstract_flag_variety(F::AbstractBundle, dims::Int...; symbol::String = "c")
  abstract_flag_variety(F, collect(dims), symbol=symbol)
end

```
or
```suggestion
abstract_flag_variety(F::AbstractBundle, dims::Int...; symbol::String = "c") = abstract_flag_variety(F, collect(dims), symbol=symbol)

```

experimental/IntersectionTheory/src/Main.jl
thofma: I suggest we ignore this for now, and adjust it once Wolfram is done with his changes.

wdecker: I agree with @thofma , there is a lot of unicode in this code which makes the code much more readable.

jankoboehm (approved):

PR #3936, opened 2024-07-14
Minor tweaks

fingolfin:
- **Fix abelian_closure docstring**
- **Sort src/exports.jl**
- **Merge Oscar.disc_log and Oscar.DiscLog.disc_log**
- **Add AlgClosure{T} docstring**

### DIFF
--- a/src/Rings/AlgClosureFp.jl
+++ b/src/Rings/AlgClosureFp.jl
@@ -20,6 +20,13 @@ import ..Oscar: algebraic_closure, base_field, base_ring, base_ring_type, charac
   elem_type, embedding, has_preimage_with_preimage, IntegerUnion, is_unit, map_entries,
   minpoly, parent_type, promote_rule, roots
 
+"""
+    AlgClosure{T} <: AbstractAlgebra.Field
+
+Type for the algebraic closure of a finite field.
+
+See [`algebraic_closure`](@ref).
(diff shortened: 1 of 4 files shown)


src/Rings/AlgClosureFp.jl
```diff
@@ -20,6 +20,13 @@ import ..Oscar: algebraic_closure, base_field, base_ring, base_ring_type, charac
...
   minpoly, parent_type, promote_rule, roots
 
+"""
+    AlgClosure{T} <: AbstractAlgebra.Field
+
+Type for the algebraic closure of a finite field.
+
+See [`algebraic_closure`](@ref).
+"""
 struct AlgClosure{T} <: AbstractAlgebra.Field
```
fingolfin, line 30: This type probably should perhaps be renamed to `AlgClosureField`. Or perhaps `FqBarField` to match `QQBarField` ?

Any change should of course also be reflected in the name of `AlgClosureElem`.

lgoettgens: I agree. `AlgClosure` doesn't convey the meaning that it is about char p, and not the same as `QQBarField`.

joschmitt (approved):
I'm in favour of aligning the names `AlgClosure` and `QQBarField` as well.

PR #3937, opened 2024-07-15
Make Lie algebra construction faster

lgoettgens:
Follow-up to https://github.com/oscar-system/Oscar.jl/pull/3831 that implements some performance improvements and fixes some type instabilities and other issues reported by JET.

The following diagram shows the enormous improvement. In particular, we are now almost a third faster than GAP for all of these test cases.

comparison_normalized_linear

created using:
```julia
cases=[(:A, 1), (:A, 2), (:A, 3), (:A, 4), (:A, 5), (:A, 6), (:A, 7), (:A, 8), (:B, 2), (:B, 3), (:B, 4), (:B, 5), (:B, 6), (:B, 7), (:B, 8), (:D, 4), (:D, 5), (:D, 6), (:D, 7), (:D, 8), (:E, 6), (:E, 7), (:E, 8), (:F, 4), (:G, 2)]
gaptimes = [begin @info case; @belapsed GAP.Globals.SimpleLieAlgebra($(GAP.Obj(String(case[1]))), $(case[2]), $(GAP.Globals.Rationals)) end for case in cases]
oscartimes = [begin @info case; @belapsed lie_algebra($QQ, $(case[1]), $(case[2])) end for case in cases]

using StatsPlots, Plots

groupedbar([[1 for _ in cases] (oscartimes_master./gaptimes) (oscartimes_5b6483./gaptimes)]; xticks=(eachindex(cases), join.(cases)),label=["GAP (normalized)" "OSCAR (master)" "OSCAR (PR)"],bar_width=0.65,ylims=(-Inf,4))
```


experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -21,7 +23,13 @@
 
     R = new(mat)
     R.positive_roots = map(r -> RootSpaceElem(R, r), pos_roots)
+    R.positive_roots_map = Dict(
+      (coefficients(root), ind) for (ind, root) in enumerate(R.positive_roots::Vector{RootSpaceElem})
+    )
```
lgoettgens, line 29: pre-computing this and using this in `is_root` instead of a linear search is the main speed improvement of this PR

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -438,13 +450,13 @@ function Base.:(*)(q::RationalUnion, r::RootSpaceElem)
 end
 
 function Base.:(+)(r::RootSpaceElem, r2::RootSpaceElem)
-  @req root_system(r) === root_system(r2) "$r and $r2 must belong to the same root space"
+  @req root_system(r) === root_system(r2) "parent root system mismatch"
```
lgoettgens, line 455: these strings get interpolated even in the case that no error is thrown. Withouth string interpolation, there is basically no overhead

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -104,6 +114,10 @@ end
   return cartan_bilinear_form(cartan_matrix(R); check=false)
 end
 
+@attr QQMatrix function bilinear_form_QQ(R::RootSystem)
```
joschmitt: I think something like
```suggestion
@attr QQMatrix function bilinear_form(::QQField, R::RootSystem)
```
would be more OSCAR-y?

lgoettgens: Yeah, but this does not allow `@attr` to work. Furthermore, this is just supposed to be internal

experimental/LieAlgebras/test/RootSystem-test.jl
```diff
@@ -54,6 +73,30 @@
       @test all(i -> positive_coroot(R, i) == positive_coroots(R)[i], 1:npositive_roots)
       @test all(i -> negative_coroot(R, i) == negative_coroots(R)[i], 1:npositive_roots)
       @test simple_coroots(R) == positive_coroots(R)[1:rk]
+      @test all(is_coroot_with_index(r) == (true, i) for (i, r) in enumerate(coroots(R)))
```
lgoettgens, line 76:
```suggestion
      @test all(is_coroot, coroots(R))
      @test all(is_coroot_with_index(r) == (true, i) for (i, r) in enumerate(coroots(R)))
```

experimental/LieAlgebras/src/RootSystem.jl
lgoettgens: I renamed it now to `_bilinear_form_QQ` to make it more visible that its internal

felix-roehrich (approved):
Looks good, thanks.

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -274,7 +288,7 @@ This is a more efficient version for `positive_coroots(R)[i]`.
 Also see: `positive_coroots`.
 """
 function positive_coroot(R::RootSystem, i::Int)
-  return R.positive_coroots[i]::DualRootSpaceElem
+  return (R.positive_coroots::Vector{DualRootSpaceElem})[i]
```
fingolfin, line 291: How about this to avoid the repeated type annotations? (But I am fine with this either way)
```suggestion
  return positive_coroots(R)[i]
```

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -638,36 +671,55 @@ function height(r::DualRootSpaceElem)
...
 end
 
 function is_positive_coroot_with_index(r::DualRootSpaceElem)
-  i = findfirst(==(r), positive_coroots(root_system(r)))
+  i = get(root_system(r).positive_coroots_map, coefficients(r), nothing)
   if isnothing(i)
     return false, 0
   else
     return true, i
   end
```
fingolfin, line 700: Obv. this is existing code, and it is fine, but just as food for thought:
```suggestion
  i = get(root_system(r).positive_coroots_map, coefficients(r), 0)
  return !is_zero(i), i
```

fingolfin (approved):

experimental/LieAlgebras/src/RootSystem.jl
lgoettgens: for the positive (co)roots this indeed works. For the negative ones, however, this is not optimal as it would first negate all positive roots to then only take one of them. And I rather have positive and negative consistent here.

experimental/LieAlgebras/src/RootSystem.jl
lgoettgens: will be included in the next round of refactoring :)

PR #3943, opened 2024-07-16
tropical determinant via Hungarian method

micjoswig:
The previous implementation (via Leibniz' rule) was not useful.  It is now replaced via polymake's implementation of the Hungarian method.

To see the difference, here are two computations with v.1.1.1:
```

julia> @time det(matrix(tropical_semiring(), rand(0:9, 100, 100)))
ERROR: OverflowError: 100 is too large to look up in the table; consider using `factorial(big(100))` instead
Stacktrace:
 [1] factorial_lookup
   @ ./combinatorics.jl:19 [inlined]
 [2] factorial
   @ ./combinatorics.jl:27 [inlined]
 [3] AbstractAlgebra.Generic.AllPerms(n::Int64)
   @ AbstractAlgebra.Generic ~/.julia/packages/AbstractAlgebra/EYoda/src/generic/GenericTypes.jl:57
 [4] iterate
   @ ~/.julia/packages/AbstractAlgebra/EYoda/src/generic/PermGroups.jl:634 [inlined]
 [5] det(A::AbstractAlgebra.Generic.MatSpaceElem{TropicalSemiringElem{typeof(min)}})
   @ Oscar ~/.julia/packages/Oscar/38HwK/src/TropicalGeometry/matrix.jl:35
 [6] macro expansion
   @ ./timing.jl:279 [inlined]
 [7] top-level scope
   @ ./REPL[10]:1

julia> @time Polymake.tropical.tdet(matrix(tropical_semiring(), rand(0:9, 100, 100)))
  0.016758 seconds (328.42 k allocations: 3.419 MiB, 22.80% compilation time)
pm::TropicalNumber<pm::Min, pm::Rational>
0
```


src/TropicalGeometry/matrix.jl
```diff
@@ -14,7 +14,9 @@
 @doc raw"""
     det(A::Generic.MatSpaceElem{<: TropicalSemiringElem})
 
-Return the tropical determinant of `A`.
+Return the tropical determinant of `A`.  That is, this function evaluates the tropicalization of the ordinary determinant considered as a multivariate polynomial.
```
YueRen:
```suggestion
Return the tropical determinant of `A`.  That is, this function evaluates the tropicalization of the ordinary determinant considered as a multivariate polynomial at `A`.
```

YueRen (changes requested):
A very welcome change!

micjoswig:
@danteluber please have a look

src/TropicalGeometry/matrix.jl
```diff
@@ -30,12 +32,9 @@ julia> det(A)
 ```
 """
 function det(A::Generic.MatSpaceElem{R}) where {R<:Union{TropicalSemiringElem,MPolyRingElem{<:TropicalSemiringElem},PolyRingElem{<:TropicalSemiringElem}}}
```
YueRen: This function is also being called for matrices over `MPolyRingElem{<:TropicalSemiringElem}` and `PolyRingElem{<:TropicalSemiringElem}`, which causes problems in the tests.

We probably want two `det` functions:
-  `function det(A::Generic.MatSpaceElem{TropicalSemiringElem})` (this function)
-  `function det(A::Generic.MatSpaceElem{R}) where {R<:Union{MPolyRingElem{<:TropicalSemiringElem},PolyRingElem{<:TropicalSemiringElem}}}` (the old `det` function)

YueRen (dismissed):

src/TropicalGeometry/matrix.jl
micjoswig: Yes, I saw that.  The following test fails
```
R, (x,y) = T[:x,:y]
A = Matrix(identity_matrix(R,2)) # (julia) matrix over tropical polynomials
@test det(A) == one(R)
```
I am not quite sure if your old was actually correct.  The Leibniz formula cannot be evaluated over an arbitrary semiring, because it requires the multiplicative unit to have an additive inverse.

The reason why ordinary determinants of matrices with ordinary polynomials as coefficients make sense is that polynomial rings don't have zero divisors, whence they embed into their field of fractions.  Ordinary determinants only make sense over ordinary fields.

Your formula computes the permanent, which kind of makes sense over semirings.  So, if we should revive your version, then under the name `permanent`.

Please enlighten me about your goals.  As far as I can tell for now, the test cases giving an error should be removed.

micjoswig: Note: For coefficients in the tropical semiring there is no difference between the (tropical) determinant and the (tropical) permanent.  The reason is that +1 and -1 in any valued field both tropicalize to 0.

YueRen: I added determinants of matrices over the tropical polynomial ring without a second thought, and I don't really have any goals with them.

If determinants over the tropical polynomial ring don't make any sense mathematically, then I agree that we should remove those failing tests and possibly even change the tropical `det` function so that it is only called for matrices over the tropical semiring.

That way, if the user tries to compute `det` of a matrix of tropical polynomials, OSCAR will use the generic implementation and the user will immediately get an error that negation is not defined over the tropical semiring.

src/TropicalGeometry/matrix.jl
```diff
@@ -30,12 +32,9 @@ julia> det(A)
...
-    detA = zero(base_ring(A))
-    nrows(A)!=ncols(A) && return detA # return tropical zero if matrix not square
-    for sigma in AbstractAlgebra.SymmetricGroup(nrows(A)) # otherwise follow Leibniz Formula
-        detA += prod([ A[i,sigma[i]] for i in 1:nrows(A) ])
-    end
-    return detA
+    T = base_ring(A)
+    nrows(A)!=ncols(A) && return zero(T) # return tropical zero if matrix not square
```
lkastner: I am curious about this choice: Why not an error? For non-tropical matrices I get an error:
```
julia> A = matrix(ZZ, [0 1 2; 3 4 5])
[0   1   2]
[3   4   5]

julia> det(A)
ERROR: Non-square matrix
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] det(x::ZZMatrix)
   @ Nemo ~/.julia/packages/Nemo/WUukL/src/flint/fmpz_mat.jl:749
 [3] top-level scope
   @ REPL[3]:1
```

src/TropicalGeometry/matrix.jl
micjoswig: OK, I removed those tests and the function definition for matrices with polynomials.

src/TropicalGeometry/matrix.jl
micjoswig: Suggestion accepted.

src/TropicalGeometry/matrix.jl
```diff
@@ -29,15 +31,8 @@ julia> det(A)
...
-    return detA
-end
-
-function det(A::Matrix{R}) where {R<:Union{TropicalSemiringElem,MPolyRingElem{<:TropicalSemiringElem},PolyRingElem{<:TropicalSemiringElem}}}
-    return det(matrix(parent(first(A)),A))
+function det(A::Generic.MatSpaceElem{<:TropicalSemiringElem})
+    nrows(A)!=ncols(A) && error("Non-square matrix")
+    T = base_ring(A)
+    return T(Polymake.tropical.tdet(A))
 end
```
YueRen, line 43:
```suggestion
end

function det(A::Matrix{TropicalSemiringElem}) 
    return det(matrix(parent(first(A)),A))
end
```

We need a function like this for julia matrices over tropical semirings.  This is causing the error in the tests, I believe.

YueRen (changes requested):

YueRen (approved):
Everything looks great.  Thanks for this very useful feature!

src/TropicalGeometry/matrix.jl
```diff
@@ -29,15 +31,13 @@ julia> det(A)
...
 ```
 """
-function det(A::Generic.MatSpaceElem{R}) where {R<:Union{TropicalSemiringElem,MPolyRingElem{<:TropicalSemiringElem},PolyRingElem{<:TropicalSemiringElem}}}
-    detA = zero(base_ring(A))
-    nrows(A)!=ncols(A) && return detA # return tropical zero if matrix not square
-    for sigma in AbstractAlgebra.SymmetricGroup(nrows(A)) # otherwise follow Leibniz Formula
-        detA += prod([ A[i,sigma[i]] for i in 1:nrows(A) ])
-    end
-    return detA
+function det(A::Generic.MatSpaceElem{<:TropicalSemiringElem})
```
lgoettgens:
```suggestion
function det(A::MatrixElem{<:TropicalSemiringElem})
```
This should work for the abstract AA matrix type as well

src/TropicalGeometry/matrix.jl
```diff
@@ -14,7 +14,9 @@
 @doc raw"""
     det(A::Generic.MatSpaceElem{<: TropicalSemiringElem})
```
lgoettgens:
```suggestion
    det(A::MatrixElem{<: TropicalSemiringElem})
```
Here as well

PR #3944, opened 2024-07-16
Linear combinations of set partitions

sebvz777:
We added the implementation of linear combinations of set-partitions.
See for example Chapter 5 in [Gro20] in the Oscar references.


experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,241 @@
...
+julia> S, d = polynomial_ring(QQ, "d")
+(Univariate polynomial ring in x over QQ, d)
+julia> a = linear_partition([(set_partition([1, 2], [1, 1]), S(4)), 
+(set_partition([1, 1], [1, 1]), 4*d)])
+julia> add(a, a)
+LinearSetPartition(Dict(set_partition([1, 2], [1, 1]) => S(8), 
+SetPartition([1, 1], [1, 1]) => 8*d))
+```
+"""
+function add(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}) where
```
lgoettgens: Why do you have both `add` and `+`? I think only having `+` would be enough here. In particular, `add` should not get exported IMO.
(same applies to `substract` and `scale`)

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,241 @@
...
+where the first element in the tuple is a `RingElement` and the second
+a `SetPartition`. Furthermore simplify the term before initializing
+the `LinearSetPartition` object with the corresponding dict.
+
+# Examples
+```jldoctest
+julia> S, d = polynomial_ring(QQ, "d")
+(Univariate polynomial ring in x over QQ, d)
+julia> linear_partition([(set_partition([1, 1], [1, 1]), S(4)), (set_partition([1, 1], [1, 1]), 4*d)])
+LinearSetPartition(Dict(set_partition([1, 1], [1, 1]) => 4 + 4*d))
```
lgoettgens:
```suggestion
julia> S, d = polynomial_ring(QQ, "d")
(Univariate polynomial ring in x over QQ, d)

julia> linear_partition([(set_partition([1, 1], [1, 1]), S(4)), (set_partition([1, 1], [1, 1]), 4*d)])
LinearSetPartition(Dict(set_partition([1, 1], [1, 1]) => 4 + 4*d))
```
spacing is off here and in some other doctests

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,241 @@
...
+    return LinearSetPartition{S, T}(coeffs)
+end
+
+"""
+    coefficients(p::LinearSetPartition{S, T}) where { S <: AbstractPartition, T <: RingElement }
+
+Get the constructor field `coefficients`, the partition term, in form of a `Dict` from
+`SetPartition` to coefficient.
+""" 
+function coefficients(p::LinearSetPartition{S, T}) where { S <: AbstractPartition, T <: RingElement }
```
lgoettgens:
```suggestion
function coefficients(p::LinearSetPartition)
```
at places like this, where each type variable (here `S` and `T`) only occur at a single position (i.e. NOT in multiple types in the signature and NOT in the function body) and you don't restrict their type (the restrictions here are identical to the struct definition), you don't need to give the type parameters at all

lgoettgens (commented):
Some comments from just shortly looking at the changes.
Do you plan to have linear combinations for the other types of partitions as well?
From my view of not knowing anything about the concrete mathematics involved: I would expect that functions like `+(::SetPartition, ::SetPartition)` and `*(::RingElement, ::SetPartition)` exist and return a `LinearSetPartition`. Is something like this sensible (in a mathematical sense) and planned to get added?

fingolfin:
This seems to build on the PR #2977 by @pinguly so perhaps @pinguly would like to comment on this PR / review it? As far as I can tell, both of you are students of Moritz Weber, so it would also be useful to know what role this all plays within the TRR?

fingolfin:
@mjrodgers  please have a look at this, too

pinguly:
This pull request extends the previous one by adding linear combinations of set-partitions. These appear for example in study of quantum groups (Moritz Weber) and more generally tensor categories (Ulrich Thiel). As a concrete example, one could use these objects to implement the Temperly-Lieb algebra/category in the future.

@lgoettgens  It is not clear to me what type the result of `+(::SetPartition, ::SetPartition)` should have or in which ring the coefficients should live. If one chooses `ZZ`, is there a simple way to allow the addition with partitions over other rings. Or should `+(::LinearSetPartition, ::LinearSetPartition)` not depend on the type of the underlying ring and just try to add the coefficients?

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
+"""
+    LinearSetPartition
+
+LinearSetPartition represents a linear combination of set-partitions. See Chapter 5 in [Gro20](@cite).
```
lgoettgens: This is not entirely true, as it can also represent a linear combination of spatial partitions etc.
I find this name a bit misleading, as it implies that this only wraps SetParition and not AbstractPartition. What do you think about e.g. LinearPartition{...}?

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+"""
+function linear_partition(coeffs::Dict{S, T}) where {S <: AbstractPartition, T <: RingElement}
+    return LinearSetPartition{S, T}(coeffs)
+end
+
+"""
+    coefficients(p::LinearSetPartition)
+
+Get the constructor field `coefficients`, the partition term, in form of a `Dict` from
+`SetPartition` to coefficient.
```
lgoettgens: Here is another reference to just SetPartition

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+    stackdict[p] = q
+    return q
+end
+
+"""
+    linear_partition(term::Vector{Tuple{S, T}}) where { S <: AbstractPartition, T <: RingElement }
+
+Return a `LinearSetPartition` generated from the Vector `term` of 2-tuples,
+where the first element in the tuple is a `RingElement` and the second
+a `SetPartition`. Furthermore simplify the term before initializing
```
lgoettgens: This is in the wrong order and furthermore refs SetPartition.

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+the `LinearSetPartition` object with the corresponding dict.
+
+# Examples
+```jldoctest
+julia> S, d = polynomial_ring(QQ, "d")
+(Univariate polynomial ring in d over QQ, d)
+
+julia> linear_partition([(set_partition([1, 1], [1, 1]), S(4)), (set_partition([1, 1], [1, 1]), 4*d)])
```
lgoettgens: Maybe add another method for this where the coefficient is not type restricted at all, but instead require the coeff ring as a first parameter, and then try to coerce all coefficients from the list using `R(coeff)`. This would make it easier for users to construct these

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 4, SetPartition([1, 1], [1, 1]) => 4*d))
+
+julia> a + a
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 8, SetPartition([1, 1], [1, 1]) => 8*d))
+```
+"""
+function +(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}) where 
+        { S <: AbstractPartition, T <: RingElement } 
+
+    result = deepcopy(p)
```
lgoettgens: It would be more efficient (and imo semantically sensible) to just deep copy `coefficients(p)`, and then modifying this

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+julia> linear_partition([(set_partition([1, 1], [1, 1]), S(4)), (set_partition([1, 1], [1, 1]), 4*d)])
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 1], [1, 1]) => 4*d + 4))
+```
+"""
+function linear_partition(term::Vector{Tuple{S, T}}) where { S <: AbstractPartition, T <: RingElement }
+    return linear_partition(Dict{S, T}(x[1] => x[2] for x in simplify_operation(term)))
+end
+
+"""
+    +(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}) where
```
lgoettgens: At some time we agreed to not have docstrings on basic mathematical functions like +, if they don't give any relevant insight. Background was to reduce the amount of output when typing `?+` in the repl.
The doctest can be moved to the usual tests

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+# Examples
+```jldoctest
+julia> S, d = polynomial_ring(QQ, "d")
+(Univariate polynomial ring in d over QQ, d)
+
+julia> scale(S(1) / S(2), linear_partition(Dict(set_partition([1, 2], [1, 1]) => S(8), set_partition([1, 1], [1, 1]) => 8*d)))
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 4, SetPartition([1, 1], [1, 1]) => 4*d))
+```
+"""
+function scale(a::RingElement, p::LinearSetPartition{S, T}) where
```
lgoettgens: This function should be renamed * and then doesn't need a docstring (see +)

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+julia> a = linear_partition([(set_partition([1, 2], [1, 1]), S(4)), (set_partition([1, 1], [1, 1]), 4*d)])
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 4, SetPartition([1, 1], [1, 1]) => 4*d))
+
+julia> compose(a, a, d)
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 16*d + 16, SetPartition([1, 1], [1, 1]) => 16*d^2 + 16*d))
+```
+"""
+function compose(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}, d::T) where 
+        { S <: AbstractPartition, T <: RingElement }
+    result = Dict{S, T}()
```
lgoettgens: Don't you need to check that both linear partitions are indeed defined over the same ring? Eg in the case of number fields you cannot deduce that they have the same ring just from the type parameter T.
(same issue for + and tensorprodukt)

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+    return linear_partition(result)
+end 
+
+function ⊗(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}) where 
+        { S <: AbstractPartition, T <: RingElement }
+    return tensor_product(p, q)
+end
+
+"""
+    -(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}) where
```
lgoettgens: Docstring not needed, see +

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
...
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 4, SetPartition([1, 1], [1, 1]) => 4*d))
+
+julia> a - a
+LinearSetPartition{SetPartition, QQPolyRingElem}(Dict{SetPartition, QQPolyRingElem}())
+```
+"""
+function -(p::LinearSetPartition{S, T}, q::LinearSetPartition{S, T}) where 
+        { S <: AbstractPartition, T <: RingElement }
+    return p + scale(-1, q)
+end
```
lgoettgens: What about adding a unary - method? This could then of course be used here as `p + (-q)`

experimental/SetPartitions/src/SetPartitions.jl
```diff
@@ -19,11 +21,16 @@ import Oscar:
     parent,
     degree,
     tensor_product,
-    @req
+    @req,
+    RingElement,
+    scale,
```
lgoettgens: This Import together with the export below may export some oscar functionality that was be thought of to be kept internal. But as written above, I think your method scale shouldn't be there at all, and then you can get rid of both import and export as well

experimental/SetPartitions/src/Util.jl
```diff
@@ -123,3 +123,46 @@ function _add_partition_top_bottom(vector::Vector{Dict{Int, Set{T}}}, p::T) wher
...
+        end
+        
+        if !(i1 in keys(partitions))
+            partitions[i1] = i2
+        else
+            deleteat!(partition_sum, findall(x -> x == (i1, i2), partition_sum))
+            deleteat!(partition_sum, findall(x -> x == (i1, get(partitions, i1, -1)), partition_sum))
+            push!(partition_sum, (i1, get(partitions, i1, -1) + i2))
+            partitions[i1] = get(partitions, i1, -1) + i2
+        end
```
lgoettgens: This function seems overly complicated to me. An honestly, I don't understands what it wants to achieve, in particular the associavity in the docstring I cannot relate with what's written here. Could you maybe explain here what this is supposed to do, and then we can see how to improve this?

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,235 @@
+"""
+    LinearSetPartition
+
+LinearSetPartition represents a linear combination of set-partitions. See Chapter 5 in [Gro20](@cite).
+"""
+struct LinearSetPartition{S <: AbstractPartition, T <: RingElement}
+    coefficients :: Dict{S, T}
```
lgoettgens: I think it would be sensible to store the base ring here as well, as it cannot be reconstructed in all cases (e.g. empty linear comb), and provide it using `base_ring`. This also enables checking the base ring in operations (see comments below)

lgoettgens (commented):
Some more comments

experimental/SetPartitions/src/Util.jl
sebvz777: Hi, first of all thanks for your feedback! :)
The reason why you are not understanding it could be because I confused associativity with distributivity and made the function way more complicated than it should be... so thanks for pointing it out. I improved the code and corrected the docstring. I hope that this can clear things up!

experimental/SetPartitions/src/LinearSetPartition.jl
```diff
@@ -0,0 +1,186 @@
...
+
+    function LinearSetPartition{S, T}(coeffs::Dict{S, T}) where {S <: AbstractPartition, T <: RingElement}
+        key_types = Set(typeof(k) for k in keys(coeffs))
+        @req length(key_types) <= 1 "keys in the coefficients dictionary must be of the same subtype of AbstractPartition"
```
lgoettgens:
```suggestion
        @req isconcretetype(S) "Linear combinations are only defined for concrete subtypes of AbstractPartition"
```
I think this is what you actually want here, as `S` may otherwise be `AbstractPartition` even though all keys in the dictionary are of the same concrete type

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,201 @@
...
+See for example Chapter 5 in [Gro20](@cite) for further information on linear 
+combinations of partitions.
+"""
+struct LinearPartition{S <: AbstractPartition, T <: RingElem}
+    base_ring :: Ring  # parent_type(T)
+    coefficients :: Dict{S, T}
+
+    function LinearPartition{S, T}(ring::Ring, coeffs::Dict{S, T}) where {S <: AbstractPartition, T <: RingElem}
+        @req isconcretetype(S) "Linear combinations are only defined for concrete subtypes of AbstractPartition"
+        @req (typeof(ring) == parent_type(T)) "Ring type is not the parent of the coefficient type"
```
lgoettgens:
```suggestion
        @req ring isa parent_type(T) "Ring type is not the parent of the coefficient type"
```
This is a bit more robust

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,201 @@
...
+        for (_, c) in coeffs
```
lgoettgens:
```suggestion
        for c in values(coeff)
```
I think this exists and is more efficient

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,201 @@
...
+
+"""
+    base_ring(p::LinearPartition{S, T}) where {S <: AbstractPartition, T <: RingElem}
+
+Return the underlying coefficient ring of `p`.
+""" 
+function base_ring(p::LinearPartition{S, T}) where {S <: AbstractPartition, T <: RingElement}
+    return p.base_ring::parent_type(T) 
+end
+
```
lgoettgens, line 74: Please implement `base_ring_type(::Type{LinearPartition{S, T}) where... = parent_type(T)` as well. This makes it possible to do type assertions in other parts of the code that then get evaluated at compile time

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,201 @@
...
+function ==(p::LinearPartition, q::LinearPartition)
+    return base_ring(p) == base_ring(q) && coefficients(p) == coefficients(q)
+end
+
+function deepcopy_internal(p::LinearPartition, stackdict::IdDict)
+    if haskey(stackdict, p)
+        return stackdict[p]
+    end
+    q = linear_partition(
+            deepcopy_internal(base_ring(p), stackdict),
```
lgoettgens:
```suggestion
            base_ring(p),
```
In most cases you don't want to copy the base ring as well when copying a partition. If you do this, then e.g. `x + deepcopy(x)` won't work (e.g. You cannot use deep copy when defining arithmetics)

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,201 @@
...
+    end
+    return linear_partition(base_ring(p), result)
+end
+
+"""
+    compose(p::LinearPartition{S, T}, q::LinearPartition{S, T}, d::T) where 
+        { S <: AbstractPartition, T <: RingElement }
+
+Multiply each coefficient from `p` with each coefficient from `q`
+as well as perform a composition between each SetPartition part
```
lgoettgens: Another ref to SetParition

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,201 @@
...
+    end
+    q = linear_partition(
+            deepcopy_internal(base_ring(p), stackdict),
+            deepcopy_internal(coefficients(p), stackdict))
+    stackdict[p] = q
+    return q
+end
+
+function +(p::LinearPartition{S, T}, q::LinearPartition{S, T}) where 
+        { S <: AbstractPartition, T <: RingElement }
```
lgoettgens: You restricted the type to `T <: RingElem`. This change should be reflected in all `where`s in function headers and docstrings as well

experimental/SetPartitions/src/Util.jl
```diff
@@ -123,3 +123,47 @@ function _add_partition_top_bottom(vector::Vector{Dict{Int, Set{T}}}, p::T) wher
...
+    partitions = Dict{S, T}()
+
+    for (i1, i2) in partition_sum
+        if iszero(i2)
+            continue
+        end
+        partitions[i1] = get(partitions, i1, 0) + i2
+    end
+    
+    return [(s, t) for (s, t) in partitions]
```
lgoettgens:
```suggestion
    return [(s, t) for (s, t) in partitions if !iszero(t)]
```
You need to filter out zeros in the end again. Because some partition may be contained twice in the list, where the coefficients are additive inverse to each other

lgoettgens (commented):
Some more comments

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,207 @@
...
+"""
+function linear_partition(ring::Ring, coeffs::Dict{S, <: Any}) where {S <: AbstractPartition}
+    ring_coeffs = Dict(p => ring(c) for (p, c) in coeffs)
+    return LinearPartition{S, elem_type(ring)}(ring, ring_coeffs)
+end
+
+"""
+    linear_partition(ring::Ring, terms::Vector{Tuple{S, <: Any}}) where {S <: AbstractPartition}
+
+Return a `LinearPartition` generated from the Vector `term` of 2-tuples,
```
lgoettgens:
```suggestion
Return a `LinearPartition` generated from the vector `term` of 2-tuples,
```

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,207 @@
...
+where the first element in the tuple is a `RingElement` and the second
+a `SetPartition`. Furthermore simplify the term before initializing
```
lgoettgens: that is not true. from the signature, the first element in the tuple is a AbstractPartition.

Furthermore, you should mention `ring` in the docstring, and that the second elements of the tuples get tried to get coerced into it

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,207 @@
...
+    q = linear_partition(
+            base_ring(p),
+            deepcopy_internal(coefficients(p), stackdict))
+    stackdict[p] = q
+    return q
+end
+
+function +(p::LinearPartition{S, T}, q::LinearPartition{S, T}) where 
+        { S <: AbstractPartition, T <: RingElem } 
+    @req (base_ring(p) == base_ring(q)) "Linear partitions are defined over different base rings"
```
lgoettgens:
```suggestion
    @req base_ring(p) == base_ring(q) "Linear partitions are defined over different base rings"
```

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,207 @@
...
+
+"""
+    compose(p::LinearPartition{S, T}, q::LinearPartition{S, T}, d::T) where 
+        { S <: AbstractPartition, T <: RingElem }
+
+Return the composition between `p` and `q`.
+
+The composition is obtained by multiplying each coefficient
+of `p` with each coefficient of `q` and composing the corresponding
+partitions.
```
lgoettgens: This docstring should mention `d` and its role in the composition.

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,207 @@
...
+julia> a = linear_partition(S, [(set_partition([1, 2], [1, 1]), 4), (set_partition([1, 1], [1, 1]), 4*d)])
+LinearPartition{SetPartition, QQPolyRingElem}(Univariate polynomial ring in d over QQ, Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 4, SetPartition([1, 1], [1, 1]) => 4*d))
+
+julia> compose(a, a, d)
+LinearPartition{SetPartition, QQPolyRingElem}(Univariate polynomial ring in d over QQ, Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 16*d + 16, SetPartition([1, 1], [1, 1]) => 16*d^2 + 16*d))
+```
+"""
+function compose(p::LinearPartition{S, T}, q::LinearPartition{S, T}, d::T) where 
+        { S <: AbstractPartition, T <: RingElem }
+    @req (base_ring(p) == base_ring(q)) "Linear partitions are defined over different base rings"
```
lgoettgens:
```suggestion
    @req base_ring(p) == base_ring(q) "Linear partitions are defined over different base rings"
```

experimental/SetPartitions/src/LinearPartition.jl
```diff
@@ -0,0 +1,207 @@
...
+julia> a = linear_partition(S, [(set_partition([1, 2], [1, 1]), 4), (set_partition([1, 1], [1, 1]), 4*d)])
+LinearPartition{SetPartition, QQPolyRingElem}(Univariate polynomial ring in d over QQ, Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 2], [1, 1]) => 4, SetPartition([1, 1], [1, 1]) => 4*d))
+
+julia> tensor_product(a, a)
+LinearPartition{SetPartition, QQPolyRingElem}(Univariate polynomial ring in d over QQ, Dict{SetPartition, QQPolyRingElem}(SetPartition([1, 1, 2, 2], [1, 1, 2, 2]) => 16*d^2, SetPartition([1, 2, 3, 3], [1, 1, 3, 3]) => 16*d, SetPartition([1, 2, 3, 4], [1, 1, 3, 3]) => 16, SetPartition([1, 1, 2, 3], [1, 1, 2, 2]) => 16*d))
+```
+"""
+function tensor_product(p::LinearPartition{S, T}, q::LinearPartition{S, T}) where 
+        { S <: AbstractPartition, T <: RingElem }
+    @req (base_ring(p) == base_ring(q)) "Linear partitions are defined over different base rings"
```
lgoettgens:
```suggestion
    @req base_ring(p) == base_ring(q) "Linear partitions are defined over different base rings"
```

experimental/SetPartitions/src/SetPartitions.jl
```diff
@@ -51,6 +63,10 @@ export set_partition
 export spatial_partition
 export upper_colors
 export upper_points
+export simplify_operation_zero
+export simplify_operation
+export linear_partition
```
lgoettgens: the rest of the export list seems to be sorted alphabetically...

experimental/SetPartitions/src/SetPartitions.jl
```diff
@@ -94,3 +112,6 @@ export set_partition
 export spatial_partition
 export upper_colors
 export upper_points
+export simplify_operation_zero
+export simplify_operation
```
lgoettgens:
```suggestion
```
see above

experimental/SetPartitions/src/SetPartitions.jl
```diff
@@ -51,6 +63,10 @@ export set_partition
...
```
lgoettgens:
```suggestion
```
in my opinion, these should not get exported, as they are just used to convert user input to something consistent in the constructor. So a user should never need to call these by themselves

experimental/SetPartitions/src/SetPartitions.jl
```diff
@@ -94,3 +112,6 @@ export set_partition
...
+export linear_partition
```
lgoettgens: alphabetical sorting, see above

experimental/SetPartitions/test/LinearComb-test.jl
```diff
@@ -0,0 +1,27 @@
+@testset "LinearCombinations of SetPartitions" begin
```
lgoettgens: This testfile is named `LinearComb-test.jl`, although it tests the features from `LinearPartition.jl`. That seems to be a little mismatch.

lgoettgens (commented):
Some more minor nitpicks. Once these are adapted, this is fine from my POV. (but @fingolfin asked @mjrodgers to review this as well in https://github.com/oscar-system/Oscar.jl/pull/3944#issuecomment-2247409245)

Two additional things came to my mind while reading through this again. This is nothing that needs to be changed asap, just for @sebvz777 and @pinguly to think about for future work.
1. The name `LinearPartition` seems kind of weird. Firstly, it is not an `AbstractPartition` by itself, although the name may suggest that. Secondly, in the literature I found (and you referenced), these things are either called "linear combination of partitions" or "linear category of partitions", but never "linear partition".
2. For fixed `S` and `T` and `d` (third arg of `compose`), the objects of type `LinearPartition{S,T}` form a ring. So it would make sense to think about adding a parent object (aka the ring itself) and implementing the ring interface from AA.

pinguly:
@lgoettgens Regarding the two additional remarks:
1) I agree that `LinearPartition` does not quiet fit. It should be something like `LinearCategoryOfPartitionsElem` but this is rather long.
2)  For fixed `S`, `T` and `d` the linear partitions do not form a ring but a category since the composition is only defined if the points agree (we should assert and document this more explicitly). So one could implement a parent category object in the future when categories are included in Oscar.

mjrodgers (approved):
I agree with this most recent batch of suggestions made by @lgoettgens and think this looks good to merge into experimental as an initial commit for this package.

experimental/SetPartitions/src/Util.jl
```diff
@@ -123,3 +123,47 @@ function _add_partition_top_bottom(vector::Vector{Dict{Int, Set{T}}}, p::T) wher
...
+    simplify_operation(partition_sum::Vector{Tuple{S, T}}) where { S <: AbstractPartition, T <: RingElement }
+
+Simplify the vector representation of a `LinearPartition` in terms of distributivity.
+
+# Examples
+```jldoctest
+julia> S, d = polynomial_ring(QQ, "d")
+(Univariate polynomial ring in d over QQ, d)
+
+julia> simplify_operation([(set_partition([1, 1], [1, 1]), S(10)), (set_partition([1, 1], [1, 1]), 4*d)])
```
lgoettgens:
```suggestion
julia> Oscar.SetPartitions.simplify_operation([(set_partition([1, 1], [1, 1]), S(10)), (set_partition([1, 1], [1, 1]), 4*d)])
```

lgoettgens (approved):
All comments from above are resolved now. Once the doctests are happy, this is good to go from my POV. The above snipped should be enough to fix the doctests (I think)

PR #3945, opened 2024-07-16
Make global variables const

lgoettgens:
to allow for better type inference and more efficient compiler-generated code


src/Serialization/main.jl
```diff
@@ -315,8 +315,7 @@ function register_serialization_type(ex::Any, str::String, uses_id::Bool, uses_p
       serialize_with_params(T::Type{<:$ex}) = $uses_params
 
       # only extend serialize on non std julia types
-      non_oscar_types = Union{Number, String, Bool, Symbol, Vector, Tuple, Matrix, NamedTuple, Dict, Set}
```
lgoettgens, line 318: this line used to re-define this global variable over and over again, with every call to `@register_serialization_type`

joschmitt (approved):
Thank you!

PR #3946, opened 2024-07-16
[ToricVarieties] Overhaul equality of normal toric varieties

### LINKED ISSUE #3934: Equality (in the sense of telling identically the same - NOT isomorphic) of toric varieties malfunctions
The following example is from @lkastner . It produces varieties that are identical, but the check returns false.

```julia
julia> p2 = projective_space(NormalToricVariety, 2)
Normal toric variety

julia> fp2 = polyhedral_fan(p2)
Polyhedral fan in ambient dimension 2

julia> sfp2 = star_subdivision(fp2, [1,0])
Polyhedral fan in ambient dimension 2

julia> rays(sfp2) == rays(fp2)
true

julia> maximal_cones(sfp2) == maximal_cones(fp2)
false

julia> n_maximal_cones(sfp2) == n_maximal_cones(fp2)
true

julia> blp2 = normal_toric_variety(sfp2)
Normal toric variety

julia> p2 == blp2
false
```

I understand that a working equality check is computationally intense.

In the Friday meeting (just briefly discussed the matter there), I heard that irrespective of this performance issue, we should implement the correct equality check. So we should probably try to implement this.

HereAround:
This PR aims to close https://github.com/oscar-system/Oscar.jl/issues/3934 as follows:

* Overwrite the `==` method: Return true if `===` is true and otherwise raise an error.
* Add documentation which explains why this computation is very demanding.
* Add a (stupid) (as in not very performant) equality method  `slow_equal`. I have not exported this method (exports are hard to revert, we do not want anybody to use this method unless they really, really want to etc.). Still, the method is documented.
* Throughout OSCAR, `==` of two toric varieties was used in a number of places. It seemed reasonable to change some of them to `===`.

### DIFF
--- a/docs/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties.md
+++ b/docs/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties.md
@@ -16,6 +16,46 @@ the affine and non-affine case:
     Transformations for non-standard lattices will have to be done by the user.
 
 
+## Equality of Normal Toric Varieties
+
+!!! warning
+    Equality of normal toric varieties is computationally very demanding.
+    We have therefore made special design decisions for the `==` method.
+
+In OSCAR, the `==` operator is reserved to check if two normal toric varieties are identical,
+meaning their underlying polyhedral fans are the same. However, this check is computationally
+challenging due to several reasons:
+
+- The ray generators might be scaled.
+- The ray generators might be stored in different orders.
+- The maximal (polyhedral) cones of the polyhedral fan might be stored in different orders.
+- If we fall back on polyhedral fan equality, lineality of the cones must also be considered.
+
+To avoid this computational bottleneck, we have specially designed the `==` method.
+It checks if the memory locations of the two objects in question are identical. If 
+so, our `==` method returns `true`. Otherwise, it raise an error.
+
+Note that triple equality `===` (i.e. the check of equal the memory locations) is always
+supported for normal toric varieties. We recommend using it.
+
+However, if you truly need to check for two normal toric varieties to be mathematically
+identical, then you will need to add a custom method. This method could look as follows:
+
+```julia
+function slow_equal(tv1::NormalToricVariety, tv2::NormalToricVariety)
+  tv1 === tv2 && return true
+  ambient_dim(tv1) == ambient_dim(tv2) || return false
+  f_vector(tv1) == f_vector(tv2) || return false
+  return Set(maximal_cones(tv1)) == Set(maximal_cones(tv2))
+end
+```
+
+Please note that this method `slow_equal` is not performant, that we currently (summer 2024)
+have no intentions in adding this function to OSCAR nor to make improvements to its performance.
+Rather, expect this method to be slow, potentially painfully so.
+
+
+
 ## Constructors
 
 ### Affine Toric Varieties
--- a/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
@@ -179,6 +179,17 @@ end
 
 
 ######################
+# Equality
+######################
+
+function Base.:(==)(tv1::NormalToricVariety, tv2::NormalToricVariety)
+  tv1 === tv2 && return true
+  error("Equality of normal toric varieties is computationally very demanding. More details in the documentation.")
+end
+
+
+
+######################
 # Display
 ######################
 
--- a/experimental/Schemes/src/IdealSheaves.jl
+++ b/experimental/Schemes/src/IdealSheaves.jl
@@ -466,7 +466,11 @@ end
 function ==(I::AbsIdealSheaf, J::AbsIdealSheaf)
   I === J && return true
   X = space(I)
-  X == space(J) || return false
+  if X isa NormalToricVariety
+    X === space(J) || return false
+  else
+    X == space(J) || return false
+  end
   for U in basic_patches(default_covering(X))
     is_subset(I(U), J(U)) && is_subset(J(U), I(U)) || return false
   end
--- a/src/AlgebraicGeometry/ToricVarieties/Proj/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/Proj/constructors.jl
@@ -42,7 +42,7 @@ function _projectivization_and_total_space(is_proj::Bool, E::Vector{T}) where T 
 
   is_proj && length(E) == 1 && return v
 
-  @req all(i -> toric_variety(E[i]) == v, eachindex(E)) "The divisors are defined on different toric varieties."
+  @req all(i -> toric_variety(E[i]) === v, eachindex(E)) "The divisors are defined on different toric varieties."
 
   if is_proj
     PF_fiber = normal_fan(simplex(length(E) - 1))
--- a/src/AlgebraicGeometry/ToricVarieties/ToricMorphisms/constructors.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/ToricMorphisms/constructors.jl
@@ -190,7 +190,7 @@ end
 ####################################################
 
 function Base.:(==)(tm1::ToricMorphism, tm2::ToricMorphism)
-    return domain(tm1) == domain(tm2) && codomain(tm1) == codomain(tm2) && grid_morphism(tm1) == grid_morphism(tm2)
+    return domain(tm1) === domain(tm2) && codomain(tm1) === codomain(tm2) && grid_morphism(tm1) == grid_morphism(tm2)
 end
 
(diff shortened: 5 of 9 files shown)


src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -178,6 +178,62 @@ end
...
+- The ray generators might be stored in different orders.
+- The maximal (polyhedral) cones of the polyhedral fan might be stored in different orders.
+- If we fall back on polyhedral fan equality, lineality of the cones must also be considered.
+
+To avoid this computational bottleneck, we have designed OSCAR to *not* implement the `==`
+method for toric varieties. Instead, using `==` will always raise an error, as shown below.
+
+However, triple equality `===` is supported and will return `true` if the two instances of
+normal toric varieties refer to the same information on the hard drive (as sequence of bits and
+bytes) and `true` otherwise. We recommend using `===` instead of `==`.
```
lgoettgens:
```suggestion
bytes) and `false` otherwise. We recommend using `===` instead of `==`.
```

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -178,6 +178,62 @@ end
...
+- The ray generators might be scaled.
+- The ray generators might be stored in different orders.
+- The maximal (polyhedral) cones of the polyhedral fan might be stored in different orders.
+- If we fall back on polyhedral fan equality, lineality of the cones must also be considered.
+
+To avoid this computational bottleneck, we have designed OSCAR to *not* implement the `==`
+method for toric varieties. Instead, using `==` will always raise an error, as shown below.
+
+However, triple equality `===` is supported and will return `true` if the two instances of
+normal toric varieties refer to the same information on the hard drive (as sequence of bits and
```
lgoettgens: This is technically not correct. === checks if the memory locations are identical. This can return false, even if both objects have an identical memory representation, just being stored at different places in memory.

HereAround: Thank you.

HereAround: I have just fixed this locally and will push the corresponding change in a moment.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
HereAround: Fixed.

docs/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties.md
```diff
@@ -15,6 +15,11 @@ the affine and non-affine case:
     The lattice is always assumed to be the standard lattice $\mathbb{Z}^n$.
     Transformations for non-standard lattices will have to be done by the user.
 
+!!! warning
+```@docs
+    Base.:(==)(tv1::NormalToricVariety, tv2::NormalToricVariety)
+```
```
lgoettgens, line 51: Have you verified that this looks fine in a builded documentation? 
I am not sure that markdown understands here what you want to do

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -178,6 +178,61 @@ end
...
+true
+
+julia> function capture_error_message()
+         try
+           p2 == f2
+         catch e
+           println(e.msg)
+         end
+       end
+capture_error_message (generic function with 1 method)
```
lgoettgens: This function is not needed. You can just let the doctest print the error, and truncate them with `[...]` once line numbers etc occurr

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -178,6 +178,61 @@ end
...
+
+julia> capture_error_message()
+Equality of normal toric varieties is not implemented, as computationally very demanding. More details in the documentation.
+```
+"""
+function Base.:(==)(tv1::NormalToricVariety, tv2::NormalToricVariety)
+  error("Equality of normal toric varieties is not implemented, as computationally very demanding. More details in the documentation.")
```
fingolfin: Minor suggestion: add something like `tv1 === tv2 && return true`.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -178,6 +178,61 @@ end
...
+         try
+           p2 == f2
+         catch e
+           println(e.msg)
+         end
+       end
+capture_error_message (generic function with 1 method)
+
+julia> capture_error_message()
+Equality of normal toric varieties is not implemented, as computationally very demanding. More details in the documentation.
```
lgoettgens:
```suggestion
julia> p2 == f2
ERROR: Equality of normal toric varieties is not implemented, as computationally very demanding. More details in the documentation.
[...]
```
Something like this

HereAround: Ok. Will try this out. Thank you!

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
HereAround: Done.

docs/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties.md
HereAround: Not yet... Actually, I wanted to circumvent this by writing things up a little different, so that I will not have to worry about this issue.

docs/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties.md
```diff
@@ -15,6 +15,19 @@ the affine and non-affine case:
...
+!!! warning
+    Equality of normal toric varieties is computationally very demanding.
+    We have therefore made special design decisions for the `==` method.
+    Details below.
+
+
+## Equality of Normal Toric Varieties
+
+```@docs
+Base.:(==)(tv1::NormalToricVariety, tv2::NormalToricVariety)
```
lkastner: At some point there was some discussion whether we should add documentation to such functions that have many implementations, e.g. for `iszero` and such. @fingolfin do you remember this? I thought the outcome was not to document for these cases. The remarks could also just go in the docs folder directly.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
```diff
@@ -178,6 +178,92 @@ end
...
+ERROR: Equality of normal toric varieties is computationally very demanding. More details in the documentation.
+[...]
+
+julia> Oscar.slow_equal(p2, f2)
+false
+```
+"""
+function slow_equal(tv1::NormalToricVariety, tv2::NormalToricVariety)
+  return Set(maximal_cones(tv1)) == Set(maximal_cones(tv2))
+end
```
lkastner: Personally I would just recommend to leave out this oneliner, since everyone needing the check can do that by themselves and verify that it is a bad idea. However, if you insist, you should prepend some easy checks:
```suggestion
function slow_equal(tv1::NormalToricVariety, tv2::NormalToricVariety)
  tv1 === tv2 && return true
  ambient_dim(tv1) == ambient_dim(tv2) || return false
  f_vector(tv1) == f_vector(tv2) || return false
  return Set(maximal_cones(tv1)) == Set(maximal_cones(tv2))
end
```

thofma: One other option would be to hijack the method `issetequal` (`is_set_equal`). It is a bit more descriptive compared to `slow_equal`. Just some backseat-programming, feel free to ignore it.

docs/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties.md
HereAround: I think you are right: I have just removed the docstring and added the details directly to the documentation.

src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/constructors.jl
HereAround: Along your suggestion, I have removed the implementation of this function from OSCAR. Still, since we have the blueprint now and this might be useful to people, I have added this as code block to the documentation.

lkastner (approved):

PR #3950, opened 2024-07-17
quantum_autmorphism_groups of matroid

Sequenzer:
Add the functions `quantum_permutation_group` and `quantum_automorphism_group` to Oscar.

Both the quantum permutation group and the quantum automorphism group (of matroids) are defined in [Quantum Automorphisms of Matroids](https://arxiv.org/abs/2312.13464).

Since both of these mathematical objects are defined as quotients of a non-commutative polynomial ring, I decided to represent them in OSCAR as just the vector of generators.

All functions are basically done. Suggestions for better tests are welcome.


src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+
+# Examples
+
+```jldoctest
+M = uniform_matroid(3,4);
+qAut = quantum_automorphism_group(M,:bases);
+length(gens(qAut))
+# output
+
+2040
```
lgoettgens:
```suggestion
julia> M = uniform_matroid(3,4);

julia> qAut = quantum_automorphism_group(M,:bases);

julia> length(gens(qAut))
2040
```

fingolfin (commented):
Hi there, thanks for your contribution! I left a few suggestions for improving the code

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+  @req structure in [:bases, :circuits, :flats] "structure must be one of :bases, :circuits, :flats"
+
+  n = length(M)
+  grdSet = matroid_groundset(M)
+
+  b    = [[] for _ in 1:n]
+  nb   = [[] for _ in 1:n]
+  rels = [[] for _ in 1:n]
+
+  sets  = eval(structure)(M)
```
fingolfin, line 97: Use of `eval` is a code smell. Here it seems you want to get the Oscar function with the corresponding name, so you could do this (untested)
```suggestion
  sets  = getproperty(Oscar, structure)(M)
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+
+  new_relations = []
+
+  for relation in relation_indices
+    temp = one(A)
+    for gen in relation
+      temp = temp * u[gen[1], gen[2]]
+    end
+    push!(new_relations,temp)
+  end
```
lgoettgens, line 171:
```suggestion
  new_relations = elem_type(A)[prod(gen -> u[gen[1], gen[2]], relation; init=one(A))) for relation in relation_indices]
```
not pushing all the time but doing a list comprehension is more efficient, as constantly pushing means that julia needs to realloc space multiple times

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+  return ideal([gens(SN)..., new_relations...])
+
+end
+
+@doc raw"""
+    quantum_automorphism_group(G::Graph)
+
+Get the Ideal that defines the quantum automorphism group of a graph.
```
lgoettgens, line 179:
```suggestion
Get the ideal that defines the quantum automorphism group of a graph.
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+
+# Examples
+```jldoctest
+G = graph_from_edges([[1,2],[2,4]])
+qAut = quantum_automorphism_group(G)
+gens(qAut)
+
+# output
+
+168
```
lgoettgens, line 189:
```suggestion
julia> G = graph_from_edges([[1,2],[2,4]]);

julia> qAut = quantum_automorphism_group(G);

julia> gens(qAut)
168
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+  u = permutedims(reshape(gens(A),(n,n)),[2,1])
+
+  nonedges = Tuple{Int,Int}[] # This should be n^2 - |E| many nonedges
+  for i in 1:n 
+    for j in i:n
+      if !has_edge(G,i,j)
+        push!(nonedges,(i,j))
+      end
+    end
+  end
```
lgoettgens, line 208:
```suggestion
  nonedges = Tuple{Int,Int}[(i, j) for i in 1:n for j in i:n if !has_edge(G,i,j)] # This should be n^2 - |E| many nonedges
```

lgoettgens (commented):
some misc julia/style things I noticed while reading through this. I have no knowledge about the mathematics involved

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: Oh, thank you i didn't know about that function :)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+end
+
+@doc raw"""
+    _quantum_automorphism_group_indices(M::Matroid, structure::Symbol=:bases)
+
+Get the indices of the relations that define the quantum automorphism group of a matroid for a given structure.
+
+# Examples
+```jldoctest
+M = uniform_matroid(3,4)
```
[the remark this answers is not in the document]
Sequenzer: I believe I did it correctly now.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,228 @@
...
+        rel = []
+        for i in 1:size
+          push!(rel,(nonset[i],set[i]))
+        end
+        push!(rels[size],rel)
+      end
+    end
+  end
+
+  rels = unique.(rels)
```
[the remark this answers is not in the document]
fingolfin: Ah but this was using a broadcast before, so that would change the semantics.

Dunno if this would work?
```suggestion
  unique!.(rels)
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+  nb   = [[] for _ in 1:n]
+  rels = [[] for _ in 1:n]
+
+  sets  = getproperty(Oscar, structure)(M)
+  sizes = unique(map(x -> length(x), sets))
+  tempGrdSet = reduce(vcat,[grdSet for i in 1:n])
+
+  for size in sizes 
+    size == 0 && continue
+    powerSet = unique(sort.(Oscar.subsets(tempGrdSet,size)))
```
fingolfin, line 82:
```suggestion
    powerSet = unique(sort.(subsets(tempGrdSet, size)))
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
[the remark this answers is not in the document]
lgoettgens: good catch! your suggestion should work

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+
+    setsOfSize = filter(x->length(x)==size,sets)
+    nonSets = setdiff(powerSet,setsOfSize) 
+
+    for set in setsOfSize
+      append!(b[size],collect(Oscar.permutations(set)))
+    end
+    for nonset in nonSets
+      append!(nb[size],collect(Oscar.permutations(nonset)))
```
fingolfin, line 91:
```suggestion
      append!(nb[size], collect(permutations(nonset)))
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
fingolfin: The drawback is IMHO readability. The pushing isn't too bad, Julia amortizes this (and one give use `sizehint!` to minimize extra allocations).

But definitely should use `prod` :-).

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+julia> G = graph_from_edges([[1,2],[2,4]]);
+
+julia> qAut = quantum_automorphism_group(G);
+
+julia> length(gens(qAut));
+168
+```
+"""
+function quantum_automorphism_group(G::Graph)
+  @req typeof(G) == Graph{Undirected} "Graph must be undirected"
```
fingolfin, line 171:
```suggestion
  @req G isa Graph{Undirected} "Graph must be undirected"
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+  for i in 1:n 
+    for j in i:n
+      if !has_edge(G,i,j)
+        push!(nonedges,(i,j))
+      end
+    end
+  end
+
+
+  edgs = map(edg -> (src(edg), dst(edg)), collect(edges(G))) # This should be |E| many edges
```
fingolfin, line 189: No need to collect an iterator if you want to, well, iterate, over it :-)
```suggestion
  edgs = map(edg -> (src(edg), dst(edg)), edges(G)) # This should be |E| many edges
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+
+  new_relations = []
+  function _addrelations(edge,nonedg)
+    r1 = u[edge[1],nonedg[1]] * u[edge[2],nonedg[2]]
+    r2 = u[edge[2],nonedg[1]] * u[edge[1],nonedg[2]]
+    r3 = u[edge[1],nonedg[2]] * u[edge[2],nonedg[1]] 
+    r4 = u[edge[2],nonedg[2]] * u[edge[1],nonedg[1]]
+    append!(new_relations,[r1,r2,r3,r4])
+    return;
```
fingolfin, line 198:
```suggestion
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+  end
+  for edg in edgs
+    for nonedg in nonedges
+      _addrelations(edg,nonedg)
+    end
+  end
```
fingolfin, line 204: `new_relations` is a `Vector{Any}` here. You can avoid it e.g. like this:
```suggestion
  function _addrelations(edge,nonedg)
    r1 = u[edge[1],nonedg[1]] * u[edge[2],nonedg[2]]
    r2 = u[edge[2],nonedg[1]] * u[edge[1],nonedg[2]]
    r3 = u[edge[1],nonedg[2]] * u[edge[2],nonedg[1]] 
    r4 = u[edge[2],nonedg[2]] * u[edge[1],nonedg[1]]
    return [r1,r2,r3,r4]
  end
  new_relations = [_addrelations(edg, nonedg) for edg in edgs for nonedg in nonedges]
```

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -139,7 +139,7 @@ _to_lpring(a::FreeAssAlgebra, deg_bound::Int) = Singular.FreeAlgebra(base_ring(a
 @doc raw"""
     groebner_basis(I::FreeAssAlgIdeal, deg_bound::Int=-1; protocol::Bool=false)
 
-Return the Groebner basis of `I` with respect to the degree bound `deg_bound`. If `protocol` is `true`, the protocol of the computation is also returned. The default value of `deg_bound` is `-1`, which means that no degree bound is imposed, which leads to a computation that uses a much slower algorithm, that may not terminate, but returns a full groebner basis if it does.
+Return the Groebner basis of `I` with respect to the degree bound `deg_bound`. If `protocol` is `true`, the protocol of the computation is also returned. The default value of `deg_bound` is `-1`, which means that no degree bound is imposed, resulting in a computation that is usually slower, but will return a full Groebner basis if there exists a finite one.
```
fingolfin, line 142: Better to avoid too long lines (cut-off somewhere between 80-100 columns).

E.g.
```suggestion
Return the Groebner basis of `I` with respect to the degree bound `deg_bound`.
If `protocol` is `true`, the protocol of the computation is also returned.
The default value of `deg_bound` is `-1`, which means that no degree bound is
imposed, resulting in a computation that is usually slower, but will return a
full Groebner basis if there exists a finite one.
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+
+  n = length(M)
+  grdSet = matroid_groundset(M)
+
+  b    = [[] for _ in 1:n]
+  nb   = [[] for _ in 1:n]
+  rels = [[] for _ in 1:n]
+
+  sets  = getproperty(Oscar, structure)(M)
+  sizes = unique(map(x -> length(x), sets))
```
fingolfin, line 77:
```suggestion
  sizes = unique!(map(length, sets))
```
or
```suggestion
  sizes = unique!(length.(sets))
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+        rel = []
+        for i in 1:size
+          push!(rel,(set[i],nonset[i]))
+        end
+        push!(rels[size],rel)
+        rel = []
+        for i in 1:size
+          push!(rel,(nonset[i],set[i]))
+        end
+        push!(rels[size],rel)
```
fingolfin, line 104:
```suggestion
        push!(rels[size], [(set[i],nonset[i]) for i in 1:size])
        push!(rels[size], [(nonset[i],set[i]) for i in 1:size])
```

fingolfin: This avoids type instabilities (`rel = []` creates a `Vector{Any}`) and unnecessary extra allocations

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,206 @@
...
+"""
+function _quantum_automorphism_group_indices(M::Matroid, structure::Symbol=:bases)
+  @req structure in [:bases, :circuits, :flats] "structure must be one of :bases, :circuits, :flats"
+
+  n = length(M)
+  grdSet = matroid_groundset(M)
+
+  b    = [[] for _ in 1:n]
+  nb   = [[] for _ in 1:n]
+  rels = [[] for _ in 1:n]
```
fingolfin, line 74: These are all of type `Vector{Vector{Any}}`. Ideally they should have a concrete type, if possible, via suitable type annotations.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: oh, yeah that was dumb xD

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: Ahh thank you. I solved this differently tho since this suggestion doesn't work.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: I changed it to what lars suggested.

Sequenzer:
I believe, I did something addressing all of your comments, Thank you guys :)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,176 @@
...
+      append!(nb[size],collect(permutations(nonset)))
+    end
+    for set in b[size]
+      for nonset in nb[size]
+        push!(rels[size], [(set[i],nonset[i]) for i in 1:size])
+        push!(rels[size], [(nonset[i],set[i]) for i in 1:size])
+      end
+    end
+  end
+  println(typeof(rels))
```
lgoettgens, line 100:
```suggestion
```
missed print statement

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,175 @@
...
+
+  for size in sizes 
+    size == 0 && continue
+    powerSet = unique(sort.(subsets(tempGrdSet,size)))
+
+    setsOfSize = filter(x->length(x)==size,sets)
+    nonSets = setdiff(powerSet,setsOfSize) 
+
+    for set in setsOfSize
+      append!(b[size],collect(permutations(set)))
```
lgoettgens: the function `permutations` used here is from the `experimental` folder and should thus not be used in code in `src`. @fingolfin do you by chance  know the syntax how to do this properly?

Sequenzer: I switched to an alternative way. Lets hope the test succeed now.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+
+julia> length(idx)
+1920
+```
+"""
+function _quantum_automorphism_group_indices(M::Matroid, structure::Symbol=:bases)
+  @req structure in [:bases, :circuits, :flats] "structure must be one of :bases, :circuits, :flats"
+
+  n = length(M)
+  grdSet = matroid_groundset(M)
```
dcorey2814, line 72: make work with arbitrary groundset (I think such a function exists.)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+
+julia> length(gens(qAut))
+2040
+```
+"""
+function quantum_automorphism_group(
+  M::Matroid,
+  structure::Symbol=:bases)
+
+  @req matroid_groundset(M) isa Vector{Int} "Matroid must have integer groundset"
```
bschroter: It looks like, this isn't really necessary to restrict the ground set to Integers. Of course this is a bit of additional work and requires to adjust the function _quantum_automorphism_group_indices, but would align with the functionality that exits for matroids.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+
+    for set in setsOfSize
+      append!(b[size],[[set[g(i)] for i in 1:size] for g in S])
+    end
+    for nonset in nonSets
+      append!(nb[size],[[nonset[g(i)] for i in 1:size] for g in S])
+    end
+    for set in b[size]
+      for nonset in nb[size]
+        push!(rels[size], [(set[i],nonset[i]) for i in 1:size])
```
dcorey2814:
```suggestion
        push!(rels[size], [(set[i], nonset[i]) for i in 1:size])
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+
+  sets  = getproperty(Oscar, structure)(M)
+  sizes = unique!(map(length, sets))
+  tempGrdSet = reduce(vcat,[grdSet for i in 1:n])
+
+  for size in sizes 
+    size == 0 && continue
+    powerSet = unique(sort.(subsets(tempGrdSet,size)))
+
+    setsOfSize = filter(x->length(x)==size,sets)
```
dcorey2814:
```suggestion
    setsOfSize = filter(x->length(x)==size, sets)
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+  nb   = [Vector{Int}[] for _ in 1:n]
+  rels = [Vector{Tuple{Int,Int}}[] for _ in 1:n]
+
+  sets  = getproperty(Oscar, structure)(M)
+  sizes = unique!(map(length, sets))
+  tempGrdSet = reduce(vcat,[grdSet for i in 1:n])
+
+  for size in sizes 
+    size == 0 && continue
+    powerSet = unique(sort.(subsets(tempGrdSet,size)))
```
dcorey2814:
```suggestion
    powerSet = unique(sort.(subsets(tempGrdSet, size)))
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+  n = length(M)
+  grdSet = matroid_groundset(M)
+
+  b    = [Vector{Int}[] for _ in 1:n]
+  nb   = [Vector{Int}[] for _ in 1:n]
+  rels = [Vector{Tuple{Int,Int}}[] for _ in 1:n]
+
+  sets  = getproperty(Oscar, structure)(M)
+  sizes = unique!(map(length, sets))
+  tempGrdSet = reduce(vcat,[grdSet for i in 1:n])
```
dcorey2814:
```suggestion
  tempGrdSet = reduce(vcat, [grdSet for i in 1:n])
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+    size == 0 && continue
+    powerSet = unique(sort.(subsets(tempGrdSet,size)))
+
+    setsOfSize = filter(x->length(x)==size,sets)
+    nonSets = setdiff(powerSet,setsOfSize) 
+
+    S = symmetric_group(size)
+
+    for set in setsOfSize
+      append!(b[size],[[set[g(i)] for i in 1:size] for g in S])
```
dcorey2814:
```suggestion
      append!(b[size], [[set[g(i)] for i in 1:size] for g in S])
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+    end
+    for nonset in nonSets
+      append!(nb[size],[[nonset[g(i)] for i in 1:size] for g in S])
```
dcorey2814:
```suggestion
      append!(nb[size], [[nonset[g(i)] for i in 1:size] for g in S])
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+    end
+    for set in b[size]
+      for nonset in nb[size]
+        push!(rels[size], [(set[i],nonset[i]) for i in 1:size])
+        push!(rels[size], [(nonset[i],set[i]) for i in 1:size])
+      end
+    end
+  end
+  rels = unique!.(rels)
+  return Vector{Vector{Tuple{Int,Int}}}(reduce(vcat,rels))
```
dcorey2814:
```suggestion
  return Vector{Vector{Tuple{Int,Int}}}(reduce(vcat, rels))
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+
+@doc raw"""
+    quantum_automorphism_group(M::Matroid, structure::Symbol=:bases)
+
+Return the Ideal that defines the quantum automorphism group of a matroid for a given structure.
+
+# Examples
+
+```jldoctest
+julia> M = uniform_matroid(3,4);
```
dcorey2814: Add or replace with a more interesting example. Matroid of K4?

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
dcorey2814: Modify when doing arbitrary ground set.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+
+julia> idx = Oscar._quantum_automorphism_group_indices(M,:bases);
+
+julia> length(idx)
+1920
+```
+"""
+function _quantum_automorphism_group_indices(M::Matroid, structure::Symbol=:bases)
```
dcorey2814, line 66: make output as an iterator.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+```jldoctest
+julia> G = graph_from_edges([[1,2],[2,4]]);
+
+julia> qAut = quantum_automorphism_group(G);
+
+julia> length(gens(qAut))
+168
+```
+"""
+function quantum_automorphism_group(G::Graph)
```
dcorey2814: If the quantum automorphism group of a directed graph is that of the undirected graph, then have the input be either directed or undirected, then just convert to undirected.

dcorey2814: Add option where the generators are expressed as the magic unitary commuting with the adjacency matrix.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,177 @@
...
+  edgs = map(edg -> (src(edg), dst(edg)), edges(G)) # This should be |E| many edges
+
+  function _addrelations(edge,nonedg)
+    r1 = u[edge[1],nonedg[1]] * u[edge[2],nonedg[2]]
+    r2 = u[edge[2],nonedg[1]] * u[edge[1],nonedg[2]]
+    r3 = u[edge[1],nonedg[2]] * u[edge[2],nonedg[1]] 
+    r4 = u[edge[2],nonedg[2]] * u[edge[1],nonedg[1]]
+    return [r1,r2,r3,r4]
+  end
+  new_relations = reduce(vcat,[_addrelations(edg, nonedg) for edg in edgs for nonedg in nonedges])
```
dcorey2814: include nonedge, edge pairs. (You only have edge, nonedge pairs.)

test/Combinatorics/Matroids/Matroids.jl
```diff
@@ -354,7 +354,46 @@
         M = cycle_matroid(g)
         @test degree(automorphism_group(M)) == 5
     end
+    @testset "quantum_automorphism_group" begin
```
dcorey2814, line 357: add test for circuits.

dcorey2814 (changes requested):

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: I played around with it a bit but i don't know an easy way to do this.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: I will not do that feature for now. Since it should give the same result.

Sequenzer: I changed the required type for this function to `Graph{Undirected}` simply because I don't really know how it works in the directed case.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: Maybe @lgoettgens @fingolfin have an opinion on this. Since the output of the functions is already a vector, I don't know if it's smart to do an iterator on top. Alternatively, I could define a custom iterator by rewriting the function, but that feels like it would be a lot of work :)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: I fixed it thank you.

fingolfin:
Thanks to @Sequenzer for addressing the barrage of comments and suggestions by @lgoettgens and myself, well done :-)

Just as note: since GitHub's PR review feature is really stupid and often prefers showing resolved & outdated source code comments over newer, unresolved comments (which are then hidden behind a "show more comments..." button/link), I've deleted a bunch of, well, resolved & outdated comments by @lgoettgens and myself, to "unclog" the system a bit (this is of course crude. GH should fix their system, but I've been waiting for a long time now for that, so I don't think we can wait for it)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
fingolfin: If you generate a vector anyway, then just returning an iterator over it is pointless and even detrimental.

It is different if you can write a truly iterative version that only produces its result piecewise; then that *can* be useful. It depends on your usecases, though: if you always need all the results at once anyway, there is no gain. If OTOH you have algorithms that my have a result after just looking at the first 10 out of 2000 results, then obviously it can be helpful.

Looking at your code it seems to be pretty natural to turn it into an iterator, but I can't judge beyond that whether it would be a good idea.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,180 @@
...
+    S = symmetric_group(size)
+
+    for set in setsOfSize
+      append!(b[size], [[set[g(i)] for i in 1:size] for g in S])
+    end
+    for nonset in nonSets
+      append!(nb[size], [[nonset[g(i)] for i in 1:size] for g in S])
+    end
+    for set in b[size]
+      for nonset in nb[size]
```
fingolfin: As far as I can tell (I might be dead wrong!), you only need `b[size]` during that iteration, but not before or after. So you are wasting memory by keeping its value for previous iterations around? So then this would also work:
```suggestion
  rels = [Vector{Tuple{Int,Int}}[] for _ in 1:n]

  sets  = getproperty(Oscar, structure)(M)
  sizes = unique!(map(length, sets))
  tempGrdSet = reduce(vcat,[grdSet for i in 1:n])

  for size in sizes 
    size == 0 && continue
    powerSet = unique(sort.(subsets(tempGrdSet, size)))

    setsOfSize = filter(x->length(x)==size, sets)
    nonSets = setdiff(powerSet, setsOfSize) 

    S = symmetric_group(size)

    b = Vector{Int}[]
    for set in setsOfSize
      append!(b, [[set[g(i)] for i in 1:size] for g in S])
    end

    nb = Vector{Int}[]
    for nonset in nonSets
      append!(nb, [[nonset[g(i)] for i in 1:size] for g in S])
    end

    for set in b
      for nonset in nb
```
Actually you could then even eliminate `b` outright:
```suggestion
  rels = [Vector{Tuple{Int,Int}}[] for _ in 1:n]

  sets  = getproperty(Oscar, structure)(M)
  sizes = unique!(map(length, sets))
  tempGrdSet = reduce(vcat,[grdSet for i in 1:n])

  for size in sizes 
    size == 0 && continue
    powerSet = unique(sort.(subsets(tempGrdSet, size)))

    setsOfSize = filter(x->length(x)==size, sets)
    nonSets = setdiff(powerSet, setsOfSize) 

    S = symmetric_group(size)

    nb = Vector{Int}[]
    for nonset in nonSets
      append!(nb, [[nonset[g(i)] for i in 1:size] for g in S])
    end
    for set0 in setsOfSize, g in S
      set = [set0[g(i)] for i in 1:size]
      for nonset in nb
```
But you may not wish to do that to preserve symmetry (and since other things in this code seem even more expensive, it probably doesn't help much anyway)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,180 @@
...
+
+  new_relations = elem_type(A)[prod(gen -> u[gen[1], gen[2]], relation; init=one(A)) for relation in relation_indices]
+  return ideal([gens(SN)..., new_relations...])
+
+end
+
+@doc raw"""
+    quantum_automorphism_group(G::Graph)
+
+Return the Ideal that defines the quantum automorphism group of a graph.
```
fingolfin:
```suggestion
Return the ideal that defines the quantum automorphism group of the graph `G`.
```

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,180 @@
...
+"""
+function quantum_automorphism_group(G::Graph{Undirected})
+
+  n = nv(G)
+
+  SN = quantum_symmetric_group(n)
+  A = base_ring(SN)
+  u = permutedims(reshape(gens(A),(n, n)),[2, 1])
+
+  nonedges = Tuple{Int,Int}[(i, j) for i in 1:n for j in i:n if !has_edge(G,i,j)] # This should be n^2 - |E| many nonedges
```
fingolfin: What is `E`? If you can make |E| concrete then it might be better to turn this comment into an assertion?

Here's a guess
```suggestion
  nonedges = Tuple{Int,Int}[(i, j) for i in 1:n for j in i:n if !has_edge(G,i,j)]
  @assert length(nonedges) == n^2 - n_edges(G)
```

fingolfin: (To be clear: the code is overall in pretty great shape now, feel free to ignore my additional suggestions)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,180 @@
...
+
+  function _addrelations(edge,nonedg)
+    r1 = u[edge[1], nonedg[1]] * u[edge[2], nonedg[2]]
+    r2 = u[edge[2], nonedg[1]] * u[edge[1], nonedg[2]]
+    r3 = u[edge[1], nonedg[2]] * u[edge[2], nonedg[1]] 
+    r4 = u[edge[2], nonedg[2]] * u[edge[1], nonedg[1]]
+    return [r1,r2,r3,r4]
+  end
+  new_relations = reduce(vcat,[vcat(_addrelations(edg, nonedg), _addrelations(nonedg, edg)) for edg in edgs for nonedg in nonedges])
+  return ideal([gens(SN)..., unique(new_relations)...])
```
fingolfin: To be clear, the following is relatively minor. But: if the goal is to make this code optional, then it shouldn't use `vcat` and `reduce`, but rather a straight-forward loop pushing things into a vector for which the max capacity was set beforehand. Like this change perhaps (untested), which avoids allocation a length 4 vector for each call of `_addrelations`, followed by another allocation for the inner `vcat`, etc. etc. -- this function should only allocate for the result vector perhaps two times.
```suggestion
  new_relations = copy(gens(SN))
  sizehint!(new_relations, ngens(SN) + 8*length(edgs)*length(nonedges))
  function _addrelations(edge, nonedg)
    r1 = u[edge[1], nonedg[1]] * u[edge[2], nonedg[2]]
    r2 = u[edge[2], nonedg[1]] * u[edge[1], nonedg[2]]
    r3 = u[edge[1], nonedg[2]] * u[edge[2], nonedg[1]] 
    r4 = u[edge[2], nonedg[2]] * u[edge[1], nonedg[1]]
    push!(new_relations, r1, r2, r3, r4)
  end
  for edg in edgs, nonedg in nonedges
    _addrelations(edg, nonedg)
    _addrelations(nonedg, edg)
  end
  unique!(new_relations)
  return ideal(new_relations)
```

dcorey2814 (approved):

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: I applied this suggestion.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,229 @@
...
+
+  n = length(M)
+  grdSet = matroid_groundset(M)
+
+  rels = Vector{Tuple{Int,Int}}[]
+
+  sets  = getproperty(Oscar, structure)(M)
+  sizes = unique!(map(length, sets))
+
+  function _qAut_size()
```
fingolfin: Why is this a nested function?

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,229 @@
...
+    setSizes = map(length, sets)
+
+    SetsCount = [0 for _ in 1:maximum(setSizes)]
```
fingolfin: Rename `SetsCount` to `setsCount` or even `sets_count`?

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,229 @@
...
+    foreach(size->SetsCount[size] += 1, setSizes)
+    for i in 1:length(SetsCount)
+      if SetsCount[i] == 0 || SetsCount[i] == n^i
+        SetsCount[i] = 0
+        continue
+      end
+      nonSetsCount = n^i - (SetsCount[i] * factorial(i))
```
fingolfin: I have a hard time understanding this code :-( probably because I am missing some matroid theory?

So `sets` contains all bases/circuits/flats, `sizes` all sizes of those that occur; `setSizes` then is the same as `sizes` except duplicates aren't removed; you then count for each size how often it turns up.

Side remark: By chance (for unrelated reasons) yesterday I asked on the Julia slack about a convenient function doing such a thing and @lgoettgens pointed out `multiset` to me -- you might consider using `setsCount = multiset(length.(sets))` which returns an `MSet` which counts for each key how often it occurs in it.


Aaanyway: for a set of size $n$ there are $\binom{n}{i}$ subsets of size $i$. So why do you check against `SetsCount[i] == n^i` and not `SetsCount[i] == binomial(n,i)` ?

Unless those sets are actually also multisets and may contain repetitions, then `n^i` makes of course sense.

But then why are the non-sets computed as `n^i - (SetsCount[i] * factorial(i))` and not `n^i - SetsCount[i]` ?


Perhaps all of this is obvious when one is deep in this, but perhaps also some comments could be added, and perhaps some variables have bad names (e.g. if these "sets" are not "sets")

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,229 @@
...
+  end
+  for edge in edgs, nonedg in nonedges
+    _addrelations(edge,nonedg)
+    _addrelations(nonedg,edge)
+  end
+  unique!(new_relations)
+  return ideal(new_relations)
+end
+
+struct MultiPartitionIterator{T}
```
fingolfin, line 197: Oh that seems quite nice :-)

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
```diff
@@ -0,0 +1,229 @@
...
+  end
+  @assert length(rels) == vector_size "The number of relations is not correct $(length(rels)) != $(vector_size)"
+
+  return Vector{Vector{Tuple{Int,Int}}}(rels)
+end
+
+@doc raw"""
+    quantum_automorphism_group(M::Matroid, structure::Symbol=:bases)
+
+Return the Ideal that defines the quantum automorphism group of a matroid for a given structure.
```
fingolfin: I  have a feeling of deja vu ;-)
```suggestion
Return the ideal that defines the quantum automorphism group of a matroid for a given structure.
```

Sequenzer: Yes, I wasn't able to commit to your previous suggestion :) sooorry.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: So the answer is yes, these sets could technically be multisets containing repetitions, but for the implemented axiom structures they don't contain any duplicates. This means that the check against `SetsCount[i] == n^i` is useless at best.

We count with `n^i -(SetsCount[i] * factorial(i))`, because `sets = getproperty(Oscar, structure)(M)` will only return one permutation for each set of the chosen structure. In the cited paper we argue that instead of choosing sets from the powerset of the groundset, we need to generalise to new bases/circuits/hyperplanes which are a subset of all tuples consisting of elements of the groundset. In the case of bases, for example, this means that a basis is a tuple such that it has no duplicate elements and the support defines a basis in the classical sense.

As an example: If `{1,2,3}` is a classical basis, then `(1,2,3), (2,1,3), (2,3,1), (1,3,2), (3,1,2), (3,2,1)` are bases in our world.

As for the name `sets`, do you have a better suggestion? I chose `sets` because it is the only term that fits all the different structures that come from basis sets, independent sets, etc. but yeah it might be confusing haha
  
I will tinker around a bit more.

src/Combinatorics/Matroids/quantum_automorphism_groups.jl
Sequenzer: I believe it is more fitting there, but idk what the policy is. 

But yeah it doesn't have to be a function. I will remove that.

Sequenzer:
So I for my part am done with it. I asked Dan to look at the mathematical aspects of it and he believes the code is correct.

So this should be ready to merge :)

docs/src/Combinatorics/matroids.md
```diff
@@ -123,4 +123,8 @@ chow_ring(M::Matroid; ring::Union{MPolyRing,Nothing}=nothing, extended::Bool=fal
 augmented_chow_ring(M::Matroid)
 ```
 
-
+### Quantum Automorphisms
+```@docs
+quantum_symmetric_group(n::Int)
+function quantum_automorphism_group(M::Matroid, structure::Symbol)
```
benlorenz:
```suggestion
quantum_automorphism_group(M::Matroid, structure::Symbol)
```

PR #3951, opened 2024-07-17
`QuadFormAndIsom`: fix permutation of signatures

StevellM:
We want to identify signatures of hermitian forms over a given cyclotomic field which are identical up to changing a "fixed root of unity". The set of corresponding permutations of places is not equal to the full symmetric group (of appropriate degree). In the previous version of the code, I was identifying all signatures which were similar under any permutation of the real places, which was mathematically wrong. This fixes this issue.

Thanks to @thofma for the help and the snippet of code in `_permutations_of_places`.


StevellM:
After discussion, this is still not completely correct. Better fix to come soon.

PR #3958, opened 2024-07-20
chore: bump dependencies

thofma:


test/book/cornerstones/groups/auxiliary_code/main.jl
```diff
@@ -1,5 +1,5 @@
 import Pkg
-Pkg.add(name="GenericCharacterTables", version=v"0.2"; io=devnull)
+Pkg.add(name="GenericCharacterTables", version="0.3"; io=devnull)
```
lgoettgens: This will choose the latest 0.3.x release of GCT.
Other possible options are:
1. Fix a specific version
```suggestion
Pkg.add(name="GenericCharacterTables", version=v"0.3.1"; io=devnull)
```
2. Use the latest version (even breaking)
```suggestion
Pkg.add(name="GenericCharacterTables"; io=devnull)
```

lgoettgens: (all of these of course wait for @fingolfin to release a 0.3.1 version)

benlorenz:
It looks like this needs another update for the GenericCharacterTables: 
https://github.com/oscar-system/Oscar.jl/actions/runs/10042679900/job/27753670388?pr=3958#step:9:534

The code also uses `Generic.UniversalPolyRing` with two type parameters:
https://github.com/oscar-system/GenericCharacterTables.jl/blob/cd48ab9c38d8476b1bfcedd47e68e1b3d2e688d7/src/CharTable.jl#L24
https://github.com/oscar-system/GenericCharacterTables.jl/blob/cd48ab9c38d8476b1bfcedd47e68e1b3d2e688d7/src/CharTable.jl#L33

And there seem to be some deprecated calls to `is_power` (https://github.com/Nemocas/Nemo.jl/pull/1830) in the docs:
https://github.com/oscar-system/Oscar.jl/blob/e548b0db6ab82c81dd88518a570e1a3fb7bf3adc/docs/src/Rings/rational.md?plain=1#L271-L291
(and maybe more?)

lgoettgens:
See https://github.com/oscar-system/GenericCharacterTables.jl/pull/87 for resolving the remaining problems with GCT.jl

lgoettgens:
GCT.jl 0.3 includes a breaking change for the OSCAR book. See https://github.com/oscar-system/GenericCharacterTables.jl/issues/88 for details. This issue will unfortunately block this PR here.

test/book/cornerstones/groups/auxiliary_code/main.jl
lgoettgens: If/once https://github.com/oscar-system/GenericCharacterTables.jl/pull/90 is merged and released, this instead should be changed to
```suggestion
Pkg.add(name="GenericCharacterTables", version="0.2"; io=devnull)
```

(@benlorenz it is unfortunately not possible to to more involved compat strings here. The only possibility is to provide a string missing the patch version (or missing patch and minor), which makes julia choose any of them. As long as you don't have any version installed and have a rather new general registry (as both is the case in CI), this should then choose the newest applicable patch version.
The only other possibility would be to fix a concrete version (like 0.2.2) here, and need to bump it with every release of GCT.)

lgoettgens (approved):
Finally! LGTM

lgoettgens:
@benlorenz could you please have a look at the change in the booktest setup? I think this should be more or less what we talked about yesterday, but I wanna make sure that it is. For more details see https://github.com/oscar-system/Oscar.jl/pull/3958#discussion_r1688043189

lgoettgens:
> @benlorenz could you please have a look at the change in the booktest setup? I think this should be more or less what we talked about yesterday, but I wanna make sure that it is. For more details see [#3958 (comment)](https://github.com/oscar-system/Oscar.jl/pull/3958#discussion_r1688043189)

We just had a look at this together and he approved the change orally.

benlorenz (approved):

PR #3959, opened 2024-07-20
Intersection theory: Schubert calculus

wdecker:


wdecker:
This can be merged as soon as tests are green.

joschmitt (approved):

PR #3963, opened 2024-07-23
booktest: explicitly destroy mongodb db and client

benlorenz:
Maybe this helps for #3961

Note: this PR targets: #3939


benlorenz:
With the explicit destroy the tests went through (including vinberg2): https://github.com/oscar-system/Oscar.jl/actions/runs/10062115932/job/27813830951?pr=3963

Now trying with a less hacky way by just reassigning the variables.

The reason why this did no appear earlier is that the new tests cause the db connection to timeout which causes an issue with atexit hooks (for the global `mongoc_cleanup`) running before the finalizer of the connection object.

PR #3964, opened 2024-07-23
A few toric fixes and additions

HereAround:
* Support functionality for trivial line bundle on toric variety.
* Fix a typo in the documentation.
* Remove some (presumably?) outdated comments in the documentation.

### DIFF
--- a/docs/src/AlgebraicGeometry/ToricVarieties/ToricLineBundles.md
+++ b/docs/src/AlgebraicGeometry/ToricVarieties/ToricLineBundles.md
@@ -31,6 +31,7 @@ the inverse by invoking `inv`.
 anticanonical_bundle(v::NormalToricVarietyType)
 canonical_bundle(v::NormalToricVarietyType)
 structure_sheaf(v::NormalToricVarietyType)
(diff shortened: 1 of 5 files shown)


lkastner (approved):

HereAround:
Those failures in the intersection theory/abstract varieties are surprising. @thofma any ideas?

thofma:
The PR here adds `trivial_line_bundle`, which breaks `IntersectionTheory` (since `FTheoryTools` is loaded first). It now needs to be imported in `IntersectionTheory`. Adding a line at https://github.com/oscar-system/Oscar.jl/blob/master/experimental/IntersectionTheory/src/IntersectionTheory.jl#L9 should fix it.

HereAround:
To clarify, there are no changes to FTheoryTools in this PR, but rather in the Oscar source on toric geometry. Still, seems that this needs importing in the file you mentioned. I will try to fix that.

thofma:
I see. The cure is the same.

HereAround:
> I see. The cure is the same.

Thank you. Just fixed it (or so I believe).

PR #3968, opened 2024-07-24
Stop pretending OscarInterface has a manual

### LINKED ISSUE #3967: Error `#E  component 'PDFFile' must be bound to a string denoting a relative path to a readable file` in macOS CI
Seen on a recent master, e.g. in https://github.com/oscar-system/Oscar.jl/actions/runs/10066028282/job/27826455858#step:9:239.
It seems to only occur on the macOS runners.

The only reference to `PDFFile` I found in this repo is https://github.com/oscar-system/Oscar.jl/blob/8163ba8856c7c5d7643c4f7c78a082eabbe09575/gap/OscarInterface/PackageInfo.g#L71 so I suspect the GAP interface to be somehow the culprit of this.
Maybe @fingolfin or @ThomasBreuer have some insight into this?

More breadcrumbs:
- The first CI run on master where this happened is https://github.com/oscar-system/Oscar.jl/actions/runs/9740722514/job/26878424643, which ran on https://github.com/oscar-system/Oscar.jl/commit/5d4e35298bd6c84e66bcb1224b462f7cc516c0bb (https://github.com/oscar-system/Oscar.jl/pull/3890), which was a GAP-related change.
- All changes in dependencies between this run and the previous (which didn't have the error) were Singular.jl v0.23.2->v0.23.3 and libsingular_julia_jll v0.45.1+0->v0.45.2+0.

fingolfin:
We might add one in the future, but for now there is none.

This may resolve (edit: it does not) #3967 but I am not sure  as I don't fully understand what causes that one.

### DIFF
--- a/gap/OscarInterface/PackageInfo.g
+++ b/gap/OscarInterface/PackageInfo.g
@@ -64,14 +64,7 @@ Status := "other",
 
 AbstractHTML   :=  "",
 
-PackageDoc := rec(
-  BookName  := "OscarInterface",
-  ArchiveURLSubset := ["doc"],
-  HTMLStart := "doc/chap0_mj.html",
-  PDFFile   := "doc/manual.pdf",
-  SixFile   := "doc/manual.six",
-  LongTitle := "GAP interface to OSCAR",
-),
+PackageDoc := [],
 
 Dependencies := rec(
   GAP := ">= 4.12",


lgoettgens:
The message still occurs in https://github.com/oscar-system/Oscar.jl/actions/runs/10074052750/job/27849342956?pr=3968#step:9:235, so https://github.com/oscar-system/Oscar.jl/issues/3967 doesn't get resolved

ThomasBreuer (approved):
The proposed changes are a correction, the files claimed in the old version do not exist, and an empty list as `PackageDoc` value is allowed.

PR #3975, opened 2024-07-26
FTheoryTools: Add comment to explain similar terminology; doc fix

apturner:
Minimal PR to explain confusing terminology once and for all in a comment near the top of the literature model constructor file.

@HereAround


HereAround (approved):
Thank you @apturner. Looks good to me!

PR #3978, opened 2024-07-26
[FTheoryTools] Add model 1511.03209

HereAround:
@antonydellavecchia This model requires a couple of HUGE polynomials. It is a lot faster to save them as .mrdi (thanks to your hard work!) and load them, rather than to compute them anew. Drawback: The .mrdi files are large. The largest file that we need is about 130MB in size and I could not check it into this PR by github file size restriction to at most 100MB.

~~I will try to add the file here in a moment.~~ It seems github does not allow me to upload a .mrdi file. So I will share it via slack with you.

Any suggestions on how to proceed?

(It goes without saying that this PR will fail without having that very file in `experimental/FTheoryTools/src/LiteratureModels/Models/1511.03209`.)


HereAround:
Here is the file (despite github saying "We don't support that file type).

HereAround:
cc @apturner @emikelsons

fingolfin:
Adding a file of this size to the repository is not acceptable. We really need a separate means for data storage.

At the same time I wonder how efficient the encoding is. How many terms do you polynomials have? Are the homogeneous? What's the coefficient ring?

HereAround:
> Adding a file of this size to the repository is not acceptable. We really need a separate means for data storage.
> 

Fully agreed. I have alerted @antonydellavecchia on slack, and we already had a brief conversation on the matter. I am sure we can find a good solution. For instance, for another set of F-theory models, the required data is downloaded when needed. But here, the file(s) are big/massive...

> At the same time I wonder how efficient the encoding is. How many terms do you polynomials have? Are the homogeneous? What's the coefficient ring?

A few numbers regarding the largest polynomial (a6) i.e. the one whose file I could not even temporarily attach to this PR, due to github's file size restrictions:
* The polynomial a6 lives in a polynomial ring with 101 indeterminate over Q. This polynomial ring is graded by Z^98. (Let me avoid the exact gradings, unless truly necessary.)
* The polynomial a6 is homogeneous with regard to this grading. (Let me not spell out that degree, unless truly needed.)
* The polynomial a6 consists of roughly 230.000 monomials.
* The coefficients of said polynomial a6 are taken to be random integers, if I recall correctly in the range from -250.000 to 250.000 (a6 is to model a generic section of a particular line bundle, for which the coefficients of the monomials should be sufficiently different from each other).

Maybe that helps to gauge the situation better?

HereAround:
After thinking this over a bit more, I have "minimized" this PR. It now just includes the build instructions, so that the creation of the model will take a long, long time. But as such, I believe this PR could be merged (provided the tests do not fail).

I have the .mrdi files (and so do @apturner and @emikelsons ), so that we can continue our work by loading from those files.

Whenever @antonydellavecchia is ready, we can enhance those .mrdi files by using the attribute saving framework as well as the serialization of the Lie algebras. With this, I believe we can then create a useful artifact.

Does that seem to make sense?

(Note that this PR includes the changes in https://github.com/oscar-system/Oscar.jl/pull/4002, as mutual changes in the metadata index files happened/to avoid a merge conflict.)

emikelsons (approved):
Looks good and most of the tests passed

HereAround:
@antonydellavecchia I will open a new PR shortly to discuss/implement the future serialization improvements for the `FTheoryTools`, including an artifact for the model added in this PR. For now, I am satisfied with being in principle able to create this model from scratch within `OSCAR`, even though this might take about 30 minutes on my laptop.

PR #3980, opened 2024-07-29
Serialization of Lie algebras

antonydellavecchia:
Implements serialization for Lie Algebras

### DIFF
--- a/experimental/LieAlgebras/src/serialization.jl
+++ b/experimental/LieAlgebras/src/serialization.jl
@@ -1,10 +1,16 @@
-Oscar.encode_type(::Type{<: AbstractLieAlgebra}) = "AbstractLieAlgebra"
-@register_serialization_type AbstractLieAlgebra uses_id
+@register_serialization_type AbstractLieAlgebra
 
-function save_object(s::SerializerState, L::AbstractLieAlgebra{T}) where T <: FieldElem
+function save_object(s::SerializerState, L::AbstractLieAlgebra)
   save_data_dict(s) do
-    save_typed_object(s, coefficient_ring(L), :base_field)
-    save_object(s, _struct_consts(L), :struct_const)
+    save_typed_object(s, coefficient_ring(L), :base_ring)
+    save_object(s, _struct_consts(L), :struct_consts)
     save_object(s, symbols(L), :symbols)
   end
+end
+
+function load_object(s::DeserializerState, ::Type{AbstractLieAlgebra})
+  R = load_typed_object(s, :base_ring)
+  struct_consts = load_object(s, :struct_consts)
+  s = load_object(s, :symbols)
+  return lie_algebra(R, struct_consts, s; check=false)
 end
--- a/experimental/LieAlgebras/src/serialization.jl
+++ b/experimental/LieAlgebras/src/serialization.jl
@@ -4,13 +4,18 @@
 #
 ###############################################################################
 
-@register_serialization_type AbstractLieAlgebra uses_id
+const lie_algebra_serialization_attributes = [
+  :is_abelian, :is_nilpotent, :is_perfect, :is_simple, :is_solvable
+]
+
+@register_serialization_type AbstractLieAlgebra uses_id lie_algebra_serialization_attributes
 
 function save_object(s::SerializerState, L::AbstractLieAlgebra)
   save_data_dict(s) do
     save_typed_object(s, coefficient_ring(L), :base_ring)
     save_object(s, _struct_consts(L), :struct_consts)
     save_object(s, symbols(L), :symbols)
+    save_root_system_data(s, L)
     save_attrs(s, L)
   end
 end
@@ -20,11 +25,15 @@ function load_object(s::DeserializerState, ::Type{<:AbstractLieAlgebra})
   struct_consts = load_object(s, Matrix, (sparse_row_type(R), R), :struct_consts)
   symbs = load_object(s, Vector, Symbol, :symbols)
   L = lie_algebra(R, struct_consts, symbs; check=false)
-  load_attrs!(s, L)
+  load_root_system_data(s, L)
+  load_attrs(s, L)
   return L
 end
 
-@register_serialization_type LinearLieAlgebra uses_id
+@register_serialization_type LinearLieAlgebra uses_id [
+  lie_algebra_serialization_attributes;
+  [:type, :form]
+]
 
 function save_object(s::SerializerState, L::LinearLieAlgebra)
   save_data_dict(s) do
@@ -32,6 +41,7 @@ function save_object(s::SerializerState, L::LinearLieAlgebra)
     save_object(s, L.n, :n)
     save_object(s, matrix_repr_basis(L), :basis)
     save_object(s, symbols(L), :symbols)
+    save_root_system_data(s, L)
     save_attrs(s, L)
   end
 end
@@ -42,11 +52,12 @@ function load_object(s::DeserializerState, ::Type{<:LinearLieAlgebra})
   basis = load_object(s, Vector, (dense_matrix_type(R), matrix_space(R, n, n)), :basis)
   symbs = load_object(s, Vector, Symbol, :symbols)
   L = lie_algebra(R, n, basis, symbs; check=false)
-  load_attrs!(s, L)
+  load_root_system_data(s, L)
+  load_attrs(s, L)
   return L
 end
 
-@register_serialization_type DirectSumLieAlgebra uses_id
+@register_serialization_type DirectSumLieAlgebra uses_id lie_algebra_serialization_attributes
 
 function save_object(s::SerializerState, L::DirectSumLieAlgebra)
   save_data_dict(s) do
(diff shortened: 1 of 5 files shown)


lgoettgens:
@antonydellavecchia I started duplicating everything that you added for the other two types of Lie algebras. All of them now have different problems I am afraid I need your help with.

1. AbstractLieAlgebra
needs serialization of SRow
example:
```julia
L = lie_algebra(QQ, :A, 2)
```

2. LinearLieAlgebra
saving and loading works, but it yields identical (`===`) objects. And it seems that it does not call the specific load function at all, as loading fails after calling `Oscar.reset_global_serializer_state()` with some problem with deserializing matrices.
example:
```julia
L = general_linear_lie_algebra(QQ, 2)
```

3. DirectSumLieAlgebra
examples:
```julia
# empty direct sum
L = direct_sum(QQ, LieAlgebra{QQFieldElem}[])

# non-trivial example
L1 = general_linear_lie_algebra(QQ, 2)
L2 = lie_algebra(QQ, :A, 2)
L = direct_sum(L1, L1, L2) # it is important that the first two summands are still identical after saving and loading
```

4. Element types for all of these three
but they are essentially the same, as they just contain a parent pointer and a coeff vector (or rather a 1xn matrix)
examples (for all `L` as above):
```julia
x = basis(L, 1) # for one elem
xs = basis(L) # for a vector of elems
```

5. Testing
How do I actually test if saving and loading are correct, if they give me the identical (`===`) object? I rather just compare some properties between original and deserialized.

experimental/LieAlgebras/src/serialization.jl
```diff
@@ -0,0 +1,125 @@
...
+      if has_attribute(L, bool_prop)
+        save_object(s, get_attribute(L, bool_prop), bool_prop)
+      end
+    end
+    for symbol_prop in (:type,)
+      if has_attribute(L, symbol_prop)
+        save_object(s, get_attribute(L, symbol_prop), symbol_prop)
+      end
+    end
+    # TODO: handle root_system
```
lgoettgens: @HereAround @emikelsons please wait for this TODO to be resolved (ETA end of August) before putting Lie algebras inside your FTheoryTools serialization. Or, alternatively, have a script that re-creates the serialized models so that we can just re-run it once the serialization of Lie algebras changes

src/Serialization/containers.jl
```diff
@@ -533,3 +533,31 @@ function load_object(s::DeserializerState, ::Type{<: Set}, params::Ring)
...
+  values = entry_type[]
+  load_array_node(s) do _
+    push!(pos, load_object(s, Int, 1))
+    if serialize_with_params(entry_type)
+      push!(values, load_object(s, entry_type, params, 2))
+    else
+      push!(values, load_object(s, entry_type, 2))
+    end
+  end
+  return sparse_row(params, pos, values)
```
lgoettgens:
```suggestion
  return sparse_row(params, pos, values; sort=false)
```
Since we assume that the file already has the correct format, we can assume here that the positions in `pos` are already sorted, and so Hecke does not need to do that again

src/Serialization/main.jl
```diff
@@ -74,8 +74,11 @@ end
 const reverse_type_map = Dict{String, Type}()
 
 function encode_type(::Type{T}) where T
-  error("Unsupported type '$T' for encoding. to add support see
- https://docs.oscar-system.org/stable/DeveloperDocumentation/serialization/ \n")
+  error(
+    """Unsupported type '$T' for encoding. To add support see
+    https://docs.oscar-system.org/stable/DeveloperDocumentation/serialization/
+    """
+  )
```
lgoettgens: this does not change anything in the output, but makes the file work better with syntax highlighting (don't ask me why)

src/Serialization/main.jl
```diff
@@ -365,6 +368,51 @@ macro register_serialization_type(ex::Any, args...)
...
+# Utility macro
+"""
+    @import_all_serialization_functions
+
+This macro imports all serialization related functions that one may need for implementing
+serialization for custom types from Oscar into the current module.
+One can instead import the functions individually if needed but this macro is provided
+for convenience.
+"""
+macro import_all_serialization_functions()
```
lgoettgens: This is the macro that we talked about this afternoon. If you have a better name suggestion, please use that instead.
And I think it would be good if you look through the list below to see if these are indeed the functions that users should/may use to implement their own serialization (inspired by the already existing serializations in Oscar)

src/Serialization/serializers.jl
```diff
@@ -53,6 +53,11 @@ end
 
 function end_dict_node(s::SerializerState)
   write(s.io, "}")
+
+  if s.new_level_entry
+    # makes sure that entries after empty dicts add comma
+    s.new_level_entry = false
+  end
```
lgoettgens: similar to the case for arrays below, serializing an empty dict somewhere in the tree produced syntactically invalid JSON. this seems to fix it

lgoettgens (commented):
@antonydellavecchia Thanks a lot for your help with this! I added some comments to things that came up after our discussion when I played around some more.
The problem with empty vectors is worse than expected, see https://github.com/oscar-system/Oscar.jl/issues/3983 for details what I found while digging around.
Can you take it from here? Once you are happy with all the changes in this PR, please write your approval in a comment (since Github won't allow you to do it the usual way) and I am then happy to merge.

src/Serialization/containers.jl
lgoettgens: And of course the same in https://github.com/oscar-system/Oscar.jl/blob/251b831144612cb08fcde2be58d40f59d6a15fac/src/Serialization/Rings.jl#L438 and https://github.com/oscar-system/Oscar.jl/blob/251b831144612cb08fcde2be58d40f59d6a15fac/src/Serialization/Rings.jl#L446.

experimental/LieAlgebras/src/serialization.jl
lgoettgens: Edit: I already managed to resolve this today. But you should have a reliable way to re-generate your FTheoryTools models anyway, since there certainly will be changes to the Lie algebra serialization. (Not necessarily breaking, but adding more information about them and thus e.g. speeding up some operations on them)

lgoettgens:
I added serialization for some more Lie algebra related types. This is starting to make fun :sweat_smile: 
@antonydellavecchia could you have a look at the new implementations as well? They seem to work, but maybe I did something too hacky somewhere.

src/Serialization/serializers.jl
antonydellavecchia: thanks!

experimental/LieAlgebras/src/serialization.jl
```diff
@@ -0,0 +1,248 @@
...
+  L = lie_algebra(R, struct_consts, symbs; check=false)
+  load_attrs!(s, L)
+  return L
+end
+
+@register_serialization_type LinearLieAlgebra uses_id
+
+function save_object(s::SerializerState, L::LinearLieAlgebra)
+  save_data_dict(s) do
+    save_typed_object(s, coefficient_ring(L), :base_ring)
```
antonydellavecchia, line 40: we could add a use_params for each LieAlgebra type, then the save_type_params function would take care of the coefficient ring, this would probably also make DirectSumAlgebras files a bit neater.

lgoettgens:
I'll try it out after my vacation (until 19.08.)

experimental/LieAlgebras/src/serialization.jl
lgoettgens: I tried it locally, but I prefer the current form

lgoettgens:
https://github.com/oscar-system/Oscar.jl/pull/4020 looks really interesting, but will introduce conflicts. Thus I suggest we wait with this until https://github.com/oscar-system/Oscar.jl/pull/4020 is merged. To nevertheless get the bugfixes etc. from this PR in master in a timely manner, I extracted them to https://github.com/oscar-system/Oscar.jl/pull/4021

lgoettgens:
This is now rebased on top of https://github.com/oscar-system/Oscar.jl/pull/4020. Once that and https://github.com/oscar-system/Oscar.jl/pull/4030 are merged, I can simplify this PR a bit

experimental/LieAlgebras/test/setup_tests.jl
```diff
@@ -106,7 +108,7 @@ if !isdefined(Main, :lie_algebra_conformance_test)
         @test (x * (y * z)) + (y * (z * x)) + (z * (x * y)) == zero(L)
       end
     end
-
+
```
lgoettgens:
```suggestion

```

lgoettgens (approved):
This is now finished from my POV. @antonydellavecchia could you please have a look if you see any issues and if not, proceed with the merge?

experimental/LieAlgebras/src/serialization.jl
HereAround: Sorry, this slipped my attention.

All good. I have opened https://github.com/oscar-system/Oscar.jl/pull/4008 to coordinate this and many other serialization changes for `FTheoryTools`. In other words, waiting for you and @antonydellavecchia to tell us when/how to proceed.

PR #3981, opened 2024-07-29
Remove custom non-conformant `Base.:/` method for `MPolyLocRingElem`

### LINKED ISSUE #1885: Behavior of `//` for elements in localization is inconsistent with the system
The general rule for ring elements and which makes it possible to write generic algorithms is that `//` behaves as follows:
- If `x, y` are `FieldElem`, then `x//y == divexact(x, y) == x * inv(y)`.
- If `x, y` are `RingElem`, then `x//y` constructs an element of the field of fractions. It should not be used for division.

Obviously we should not deviate from this. The various localization elements do not adhere to this rule:

```julia
julia> R, (x, y, z) = PolynomialRing(QQ, ["x", "y", "z"])
(Multivariate Polynomial Ring in x, y, z over Rational Field, fmpq_mpoly[x, y, z])

julia> P = ideal(R, [x])
ideal(x)

julia> U = complement_of_ideal(P)
complement of ideal(x)

julia> Rloc, iota = localization(U);

julia> f = iota(x)//iota(y)
x//y

julia> typeof(f)
MPolyLocalizedRingElem{FlintRationalField, fmpq, FmpqMPolyRing, fmpq_mpoly, MPolyComplementOfPrimeIdeal{FlintRationalField, fmpq, FmpqMPolyRing, fmpq_mpoly}}

julia> parent(iota(x)//iota(y)) == Rloc
true
```

Would be good if we could fix this.

CC: @wdecker @HechtiDerLachs 

Edit: Nowadays we have a total field of fractions, so we could make `x//y` construct an element in the total field of fractions in general, and in the fraction field if `is_domain_type()` is true.

HechtiDerLachs:
This is to address the inconsistency pointed out by @simonbrandhorst in #1885.

### DIFF
--- a/src/Rings/mpoly-localizations.jl
+++ b/src/Rings/mpoly-localizations.jl
@@ -1161,10 +1161,6 @@ function Base.:(/)(a::ZZRingElem, b::T) where {T<:MPolyLocRingElem}
   return (parent(b))(a//fraction(b))
 end
 
-function Base.:(/)(a::T, b::T) where {T<:MPolyLocRingElem}
-  return divexact(a, b)
-end
-
 function ==(a::T, b::T) where {T<:MPolyLocRingElem}
   parent(a) == parent(b) || error("the arguments do not have the same parent ring")
   return fraction(a) == fraction(b)
--- a/src/Rings/mpoly-localizations.jl
+++ b/src/Rings/mpoly-localizations.jl
@@ -1162,12 +1162,7 @@ function Base.:(/)(a::ZZRingElem, b::T) where {T<:MPolyLocRingElem}
 end
 
 function Base.:(/)(a::T, b::T) where {T<:MPolyLocRingElem}
-  parent(a) == parent(b) || error("the arguments do not have the same parent ring")
-  g = gcd(numerator(a), numerator(b))
-  c = divexact(numerator(a), g)
-  d = divexact(numerator(b), g)
-  numerator(fraction(b)) in inverted_set(parent(b)) || error("the second argument is not a unit in this local ring")
-  return (parent(a))(fraction(a) // fraction(b), check=false)
+  return divexact(a, b)
 end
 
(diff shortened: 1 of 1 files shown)


simonbrandhorst:
We have `Base.:/(x::RingElem, y::RingElem) = divexact(x, y; check=true)`  from abstract algebra
so the method can be deleted altogether?

HechtiDerLachs:
Tests seem to be fine. So yes: deleting the method completely sounds reasonable. I will wait for the whole test run to finish and then push the changes.

simonbrandhorst (approved):

src/Rings/mpoly-localizations.jl
```diff
@@ -1162,12 +1162,7 @@ function Base.:(/)(a::ZZRingElem, b::T) where {T<:MPolyLocRingElem}
 end
 
 function Base.:(/)(a::T, b::T) where {T<:MPolyLocRingElem}
```
fingolfin:
```suggestion
```

src/Rings/mpoly-localizations.jl
```diff
@@ -1162,12 +1162,7 @@ function Base.:(/)(a::ZZRingElem, b::T) where {T<:MPolyLocRingElem}
...
-  parent(a) == parent(b) || error("the arguments do not have the same parent ring")
-  g = gcd(numerator(a), numerator(b))
-  c = divexact(numerator(a), g)
-  d = divexact(numerator(b), g)
-  numerator(fraction(b)) in inverted_set(parent(b)) || error("the second argument is not a unit in this local ring")
-  return (parent(a))(fraction(a) // fraction(b), check=false)
+  return divexact(a, b)
 end
```
fingolfin:
```suggestion
```

HechtiDerLachs:
Thanks!

PR #1658, opened 2022-10-27
Matroid strata and realization spaces

dcorey2814:
This computes the coordinate ring for a matroid stratum of the Grassmannian using matrix coordinates.

PR #3982, opened 2024-07-29
Add docu for matroid realization spaces as affine schemes

HechtiDerLachs:
As it turned out during the ICMS, these features were apparently not known to (some of?) the developers of matroid realization spaces. So I thought it might be worthwhile to add some documentation. 

Still looking forward to testing and improving our smoothness tests for more complicated matroids!

ping: @bschroter , @dcorey2814 , @danteluber .

### DIFF
--- a/experimental/MatroidRealizationSpaces/docs/src/introduction.md
+++ b/experimental/MatroidRealizationSpaces/docs/src/introduction.md
@@ -36,7 +36,7 @@ Every `MatroidRealizationSpace` is an instance of an affine scheme.
 For those cases where implementations exist, the
 entire functionality provided for `AbsAffineScheme`s applies 
 to matroid realization spaces. For example:
-```julia
+```jldoctest
 julia> RM = realization_space(pappus_matroid(), ground_ring=QQ)
 The realization space is
   [1   0   1   0   x2   x2                 x2^2    1    0]
--- a/experimental/MatroidRealizationSpaces/docs/src/introduction.md
+++ b/experimental/MatroidRealizationSpaces/docs/src/introduction.md
@@ -31,4 +31,55 @@ If ``B`` is the polynomial ring `ambient_ring(RS)`, ``I`` the ideal
 `inequations(RS)`, then the coordinate ring of the realization space
 ``\mathcal{R}(M)`` is isomorphic to ``U^{-1}B/I``.  
 
+### Matroid realization spaces as affine schemes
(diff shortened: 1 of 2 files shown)


experimental/MatroidRealizationSpaces/docs/src/introduction.md
```diff
@@ -31,4 +31,55 @@ If ``B`` is the polynomial ring `ambient_ring(RS)`, ``I`` the ideal
 `inequations(RS)`, then the coordinate ring of the realization space
 ``\mathcal{R}(M)`` is isomorphic to ``U^{-1}B/I``.  
 
+### Matroid realization spaces as affine schemes
+Every `MatroidRealizationSpace` is an instance of an affine scheme.  
+For those cases where implementations exist, the
+entire functionality provided for `AbsAffineScheme`s applies 
+to matroid realization spaces. For example:
+```julia
```
lgoettgens:
```suggestion
```jldoctest
```
If the runtime is not absurd high, it would be great to make this a doctest instead, to make sure that this does not bitrot

LukasKuehne:
Hi,
I have a question on this PR: does that mean that a matroid realization space is now automatically returned as an affine scheme? Does this has any performance implications? The realization space itself is quite costly sometimes and we therefore wanted to make the computation modular, i.e. for instance not automatically compute the saturation as this is too slow in many instances but only do it if a certain flag is true.
Best,
Lukas

HechtiDerLachs:
> does that mean that a matroid realization space is now automatically returned as an affine scheme?

Yes. But that's the case since #1658. This PR does not change anything about that. It only documents these features. 

> Does this has any performance implications?

Not that I know of. Feel free to try it out!

> The realization space itself is quite costly sometimes and we therefore wanted to make the computation modular,

I do not quite understand. If I call `realization_space(my_matroid)`, then I get an object which is a `MatroidRealizationSpace`. It is as costly as calling your constructor and I did not interfere with any of your code there. If you plan to make any further changes or improvements on your realization spaces, this should also not be a problem.

In any case: Being an instance of an affine scheme should not affect any performance whatsoever. It just means that 
  * the type `MatroidRealizationSpace` is not hanging below `Any`, but below `AbsAffineScheme` 
  * that a function `underlying_scheme(::MatroidRealizationSpace)` has been implemented
  * that the result of the latter function is cached in an additional field of the `MatroidRealizationSpace` which is left empty by default. 
 
Nothing more, nothing less.

PR #3900, opened 2024-07-01
For Modules over PBWAlgRing and PBWAlgQuo

Lax202:
This is a draft pull request for the work to extend FreeMod, SubModuleofFreeModule, SubquoModule to PBWAlgRing and PBWAlgQuo. The aim is to extend the Groaner machinery as well.

PR #3988, opened 2024-08-02
Type changes for modules over non commutative rings

HechtiDerLachs:
Building on the work by @Lax202 in #3900, I deliberately extended the signatures to also allow for rather general non-commutative rings. I tested this code locally in another branch with a generic type for exterior algebras and the changes you see here were those necessary to make everything run. 

This is a suggestion only. If people feel uncomfortable with the generality here, please feel free to close the PR.

Note that reasonable testing is not yet possible, since the latest Hecke version is needed for that.

### DIFF
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -543,8 +543,7 @@ end
 
 const AnyGradedRingElem = Union{<:MPolyDecRingElem, <:MPolyQuoRingElem{<:MPolyDecRingElem},
                                 <:MPolyLocRingElem{<:Ring, <:RingElem, <:MPolyDecRing},
-                                <:MPolyQuoLocRingElem{<:Ring, <:RingElem, <:MPolyDecRing}, 
-                                <:NCRingElement
+                                <:MPolyQuoLocRingElem{<:Ring, <:RingElem, <:MPolyDecRing}
                                }
 
 @doc raw"""
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -5,14 +5,14 @@
 ###############################################################################
 
 @doc raw"""
-    graded_free_module(R::NCRing, p::Int, W::Vector{FinGenAbGroupElem}=[grading_group(R)[0] for i in 1:p], name::String="e")
+    graded_free_module(R::AdmissibleModuleFPRing, p::Int, W::Vector{FinGenAbGroupElem}=[grading_group(R)[0] for i in 1:p], name::String="e")
 
 Given a graded ring `R` with grading group `G`, say,
 and given a vector `W` with `p` elements of `G`, create the free module $R^p$ 
 equipped with its basis of standard unit vectors, and assign weights to these 
 vectors according to the entries of `W`. Return the resulting graded free module.
 
-    graded_free_module(R::NCRing, W::Vector{FinGenAbGroupElem}, name::String="e")
+    graded_free_module(R::AdmissibleModuleFPRing, W::Vector{FinGenAbGroupElem}, name::String="e")
 
 As above, with `p = length(W)`.
 
@@ -36,7 +36,7 @@ julia> graded_free_module(R, [G[1], 2*G[1]])
 Graded free module R^1([-1]) + R^1([-2]) of rank 2 over R
 ```
 """
-function graded_free_module(R::NCRing, p::Int, W::Vector{FinGenAbGroupElem}=[grading_group(R)[0] for i in 1:p], name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, p::Int, W::Vector{FinGenAbGroupElem}=[grading_group(R)[0] for i in 1:p], name::String="e")
   @assert length(W) == p
   @assert is_graded(R)
   all(x -> parent(x) == grading_group(R), W) || error("entries of W must be elements of the grading group of the base ring")
@@ -45,7 +45,7 @@ function graded_free_module(R::NCRing, p::Int, W::Vector{FinGenAbGroupElem}=[gra
   return M
 end
 
-function graded_free_module(R::NCRing, p::Int, W::Vector{Any}, name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, p::Int, W::Vector{Any}, name::String="e")
   @assert length(W) == p
   @assert is_graded(R)
   p == 0 || error("W should be either an empty array or a Vector{FinGenAbGroupElem}")
@@ -53,12 +53,12 @@ function graded_free_module(R::NCRing, p::Int, W::Vector{Any}, name::String="e")
   return graded_free_module(R, p, W, name)
 end
 
-function graded_free_module(R::NCRing, W::Vector{FinGenAbGroupElem}, name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, W::Vector{FinGenAbGroupElem}, name::String="e")
   p = length(W)
   return graded_free_module(R, p, W, name)
 end
 
-function graded_free_module(R::NCRing, W::Vector{Any}, name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, W::Vector{Any}, name::String="e")
   p = length(W)
   @assert is_graded(R)
   p == 0 || error("W should be either an empty array or a Vector{FinGenAbGroupElem}")
@@ -67,7 +67,7 @@ function graded_free_module(R::NCRing, W::Vector{Any}, name::String="e")
 end
 
 @doc raw"""
-    graded_free_module(R::NCRing, W::Vector{<:Vector{<:IntegerUnion}}, name::String="e")
+    graded_free_module(R::AdmissibleModuleFPRing, W::Vector{<:Vector{<:IntegerUnion}}, name::String="e")
 
 Given a graded ring `R` with grading group $G = \mathbb Z^m$, 
 and given a vector `W` of integer vectors of the same size `p`, say, create the free 
@@ -75,11 +75,11 @@ module $R^p$ equipped with its basis of standard unit vectors, and assign weight
 vectors according to the entries of `W`, converted to elements of `G`. Return the 
 resulting graded free module.
 
-    graded_free_module(R::NCRing, W::Union{ZZMatrix, Matrix{<:IntegerUnion}}, name::String="e")
+    graded_free_module(R::AdmissibleModuleFPRing, W::Union{ZZMatrix, Matrix{<:IntegerUnion}}, name::String="e")
 
 As above, converting the columns of `W`.
 
-    graded_free_module(R::NCRing, W::Vector{<:IntegerUnion}, name::String="e")
+    graded_free_module(R::AdmissibleModuleFPRing, W::Vector{<:IntegerUnion}, name::String="e")
 
 Given a graded ring `R` with grading group $G = \mathbb Z$, 
 and given a vector `W` of integers, set `p = length(W)`, create the free module $R^p$ 
@@ -113,7 +113,7 @@ julia> FF == FFF
 true
 ```
 """
-function graded_free_module(R::NCRing, W::Vector{<:Vector{<:IntegerUnion}}, name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, W::Vector{<:Vector{<:IntegerUnion}}, name::String="e")
   @assert is_zm_graded(R)
   n = length(W[1])
   @assert all(x->length(x) == n, W)
@@ -122,14 +122,14 @@ function graded_free_module(R::NCRing, W::Vector{<:Vector{<:IntegerUnion}}, name
   return graded_free_module(R, length(W), d, name)
 end
 
-function graded_free_module(R::NCRing, W::Union{ZZMatrix, Matrix{<:IntegerUnion}}, name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, W::Union{ZZMatrix, Matrix{<:IntegerUnion}}, name::String="e")
   @assert is_zm_graded(R)
   A = grading_group(R)
   d = [A(W[:, i]) for i = 1:size(W, 2)]
   return graded_free_module(R, size(W, 2), d, name)
 end
 
-function graded_free_module(R::NCRing, W::Vector{<:IntegerUnion}, name::String="e")
+function graded_free_module(R::AdmissibleModuleFPRing, W::Vector{<:IntegerUnion}, name::String="e")
   @assert is_graded(R)
   A = grading_group(R)
   d = [W[i] * A[1] for i in 1:length(W)]
@@ -652,7 +652,7 @@ function graded_map(A::MatElem)
   return graded_map(Fcdm, A)
 end
 
-function graded_map(F::FreeMod{T}, A::MatrixElem{T}; check::Bool=true) where {T <: NCRingElement}
+function graded_map(F::FreeMod{T}, A::MatrixElem{T}; check::Bool=true) where {T <: AdmissibleModuleFPRingElem}
   R = base_ring(F)
   G = grading_group(R)
   source_degrees = Vector{eltype(G)}()
@@ -669,7 +669,7 @@ function graded_map(F::FreeMod{T}, A::MatrixElem{T}; check::Bool=true) where {T 
   return phi
 end
 
-function graded_map(F::FreeMod{T}, V::Vector{<:AbstractFreeModElem{T}}; check::Bool=true) where {T <: NCRingElement}
+function graded_map(F::FreeMod{T}, V::Vector{<:AbstractFreeModElem{T}}; check::Bool=true) where {T <: AdmissibleModuleFPRingElem}
   R = base_ring(F)
   G = grading_group(R)
   nrows = length(V)
@@ -697,7 +697,7 @@ function graded_map(F::FreeMod{T}, V::Vector{<:AbstractFreeModElem{T}}; check::B
 end
 
 
-function graded_map(F::SubquoModule{T}, V::Vector{<:ModuleFPElem{T}}; check::Bool=true) where {T <: NCRingElement}
+function graded_map(F::SubquoModule{T}, V::Vector{<:ModuleFPElem{T}}; check::Bool=true) where {T <: AdmissibleModuleFPRingElem}
   R = base_ring(F)
   G = grading_group(R)
   nrows = length(V)
--- a/src/Modules/ModulesGraded.jl
+++ b/src/Modules/ModulesGraded.jl
@@ -5,14 +5,14 @@
 ###############################################################################
 
 @doc raw"""
-    graded_free_module(R::Ring, p::Int, W::Vector{FinGenAbGroupElem}=[grading_group(R)[0] for i in 1:p], name::String="e")
+    graded_free_module(R::NCRing, p::Int, W::Vector{FinGenAbGroupElem}=[grading_group(R)[0] for i in 1:p], name::String="e")
 
 Given a graded ring `R` with grading group `G`, say,
 and given a vector `W` with `p` elements of `G`, create the free module $R^p$ 
 equipped with its basis of standard unit vectors, and assign weights to these 
 vectors according to the entries of `W`. Return the resulting graded free module.
 
-    graded_free_module(R::Ring, W::Vector{FinGenAbGroupElem}, name::String="e")
+    graded_free_module(R::NCRing, W::Vector{FinGenAbGroupElem}, name::String="e")
 
 As above, with `p = length(W)`.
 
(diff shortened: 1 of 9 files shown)


src/Modules/ModulesGraded.jl
```diff
@@ -543,7 +543,8 @@ end
 
 const AnyGradedRingElem = Union{<:MPolyDecRingElem, <:MPolyQuoRingElem{<:MPolyDecRingElem},
                                 <:MPolyLocRingElem{<:Ring, <:RingElem, <:MPolyDecRing},
-                                <:MPolyQuoLocRingElem{<:Ring, <:RingElem, <:MPolyDecRing}
+                                <:MPolyQuoLocRingElem{<:Ring, <:RingElem, <:MPolyDecRing}, 
+                                <:NCRingElement
```
HechtiDerLachs: Now that I look at it again: Maybe we take out this line here. It was a custom fix to make my stuff on exterior algebras run.

src/Modules/UngradedModules/DirectSum.jl
```diff
@@ -92,7 +81,14 @@ function direct_product(M::ModuleFP{T}...; task::Symbol = :prod) where T
...
   projection_dictionary = IdDict{Int,ModuleFPHom}()
   injection_dictionary = IdDict{Int,ModuleFPHom}()
-  set_attribute!(s, :projection_morphisms => projection_dictionary, :injection_morphisms => injection_dictionary)
+  ranges = sizehint!(Vector{UnitRange{Int}}(), length(M))
+  i=0
+  for x in M
+    j = i + ngens(x)
+    push!(ranges, i+1:j)
+    i = j
+  end
```
HechtiDerLachs: This might be debateable but I witnessed that using direct sums/products can really slow down performance in some cases. My solution was to also get access to the index ranges of the distinct summands in order to write fast code. I think that storing this information in a separate attribute does not hurt, does it?

src/Modules/UngradedModules/DirectSum.jl
```diff
@@ -19,34 +19,23 @@ function direct_product(F::FreeMod{T}...; task::Symbol = :prod) where {T}
...
-    end
-    if i<length(F)
-      e *= "0"
-    end
-    e*=")"
-
-    for t = F[i].S
-      push!(G.S, Symbol(s*string(t)*e))
-    end
-  end
```
HechtiDerLachs: This has shown to eat a lot of resources for large direct sums. I tried to address this by using `join` and friends below to eventually allow for more compiler optimization.

src/Modules/ModulesGraded.jl
```diff
@@ -960,7 +960,9 @@ end
 
 
 function degrees_of_generators(M::SubModuleOfFreeModule{T}; check::Bool=true) where T
-  return map(gen -> degree(gen; check), gens(M))
+  return get_attribute(M, :degrees_of_generators) do
+    map(gen -> degree(gen; check), gens(M))
+  end::Vector{FinGenAbGroupElem}
```
HechtiDerLachs: Computing the degrees of the generators over and over again is a waste of resources. I made it an attribute for the moment. We could also turn it into a field of the struct.

src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -488,8 +488,8 @@ function _simple_kernel(h::FreeModuleHom{<:FreeMod, <:FreeMod})
   g = images_of_generators(h)
   b = ModuleGens(g, G, default_ordering(G))
   M = syzygy_module(b)
-  v = elem_type(F)[sum(c*F[i] for (i, c) in coordinates(repres(v)); init=zero(F)) for v in gens(M)]
-  I, inc = sub(F, v)
+  v = elem_type(F)[F(coordinates(repres(w))) for w in gens(M) if !is_zero(w)]
+  #v = elem_type(F)[sum(c*F[i] for (i, c) in coordinates(repres(v)); init=zero(F)) for v in gens(M)]
   return sub(F, v)
```
HechtiDerLachs:
```suggestion
  return sub(F, v)
```

HechtiDerLachs (commented):
I added some more things which I found to be necessary to make modules over non-commutative rings work. 

Among other things I did some profiling with my code and also added some fixes to save resources in general.

joschmitt:
Would it be possible to put the fixes and performance optimizations in a separate pull request? I assume we definitely want to have 76c3599 independent of whether the decision is to have modules over non-commutative rings in this generality or not.

HechtiDerLachs:
Yes, I can do that. But I wanted to first get a full picture of how this should be split up or not. If everything was approved eventually, then I could save the extra time on the manual splitting and testing, right? If it's urgent, though, I can separate certain things. Let me know if this is the case. 

Edit: Also, I think, it would make sense to include the tests written by @Lax202 . They just don't run at this point, because a later Hecke version is required (one which accepts sparse rows over non-commutative rings).

Edit: In the end I think, you're right. We will probably want the changes in different patches either way.

joschmitt:
> Yes, I can do that. But I wanted to first get a full picture of how this should be split up or not. If everything was approved eventually, then I could save the extra time on the manual splitting and testing, right? If it's urgent, though, I can separate certain things. Let me know if this is the case.
> 
> Edit: Also, I think, it would make sense to include the tests written by @Lax202 . They just don't run at this point, because a later Hecke version is required (one which accepts sparse rows over non-commutative rings).

I just think it would make sense to keep fixes separate from new features if possible. None of this is urgent (for me), but I assume that the whole pull request has to wait for some people to return from holidays, a new Hecke release and in general people to agree on something to be merged. The fix for the radical or the optimization with making `degrees_of_generators` an attribute for example should be non-controversial and could be merged now.

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -277,6 +277,10 @@ end
 *(a::Integer, b::AbstractFreeModElem) = parent(b)(base_ring(parent(b))(a)*coordinates(b))
 *(a::QQFieldElem, b::AbstractFreeModElem) = parent(b)(base_ring(parent(b))(a)*coordinates(b))
 
+function *(a::T, b::AbstractFreeModElem{T}) where T 
+  return parent(b)(base_ring(parent(b))(a)*coordinates(b))
+end
```
HechtiDerLachs:
```suggestion
function *(a::T, b::AbstractFreeModElem{T}) where {T <: NCRingElem}
  parent(a) === base_ring(parent(b)) && return parent(b)(a*coordinates(b))
  return parent(b)(base_ring(parent(b))(a)*coordinates(b))
end
```

HechtiDerLachs:
I updated this PR to only comprise the changes necessary to allow for non-commutative rings. Everything else should already be in #3995 .

Still I think we should carry over the tests due to @Lax202 from #3900. But this requires the new Hecke ~which we don't have yet~. 

ping @jankoboehm

joschmitt:
> Still I think we should carry over the tests due to @Lax202 from #3900. But this requires the new Hecke which we don't have yet.

There was a Hecke release yesterday.

HechtiDerLachs:
Looks like the Hecke release is breaking the doctests?

joschmitt:
> Looks like the Hecke release is breaking the doctests?

@thofma ?

thofma:
Was not caught in the downstream tests unfortunately, but should be fixed by #3999.

HechtiDerLachs:
Tests pass. 

We need to decide on the following:

  1. Do we want to allow modules over non-commutative rings in general? In #3900 the signatures were only widened to handle `PBWAlgElem`s, but not `NCRingElem`s. Personally, I do not see a good reason to make such restrictions, but maybe people had different ideas about this?
  2. How do we handle right-multiplication? At the moment every `ModuleFP` over a non-commutative ring is a priori considered to be a left-module. For subquos, I'm afraid, we have to stick to such a convention/restriction, eventually. But for free modules and `SubModuleOfFreeModule` we could in principle also allow for right-multiplication. At the moment some methods for this exist, but they will throw an error. I think this is a good compromise for the time being so that no wrong code is introduced, but yet the whole functionality can be extended if we decide to do so. 
  3. @Lax202 put some good work into making `SRow`s accept non-commutative rings. Sparse matrices, however, do not work over such rings, yet. They are used for modules in various places, even though not at the level of core functionality. But we should support `SMat` over non-commutative rings. 

For a full fledged support of left-, right-, and bi-modules we should then eventually discuss whether it might be worth introducing new types. Alternatively, one could probably also go via attributes of the existing modules, but as you can guess I would not be a fan of such an approach. I think some stuff is already done for ideals and we could peak at the structure there to learn? Either way: I think this PR can be merged without going into that discussion for the moment: All modules are naturally treated as left-modules for the time being and that can be extended at a later date to whatever we decide to do.

thofma:
If I remember correctly, one of Oscars design principle says that we use right actions, so the modules should be right modules by default. It is also the convention that we use for the modules over noncommutative rings that we have already, e.g., modules over finite-dimensional algebras.

Maybe it is a also a good question to ask what the goal of these modules is? Catching all `NCRingElem` makes little sense, since the concepts that are trying to be used here (normal forms? standard bases? subquo?) are applicable only to a very small subset of modules/rings. So why not stick PWB algebras, where this has a chance of working?

wdecker:
> If I remember correctly, one of Oscars design principle says that we use right actions, so the modules should be right modules by default. It is also the convention that we use for the modules over noncommutative rings that we have already, e.g., modules over finite-dimensional algebras.

In the docu we write: `When implementing functionality for PBW-algebras, taking the opposite algebra into account often allows one to focus on left ideals, left modules, and left Gröbner bases.` So under the hood, when it comes to computations, we follow the principle of Singular. Nevertheless, for the applications, we need all three types of directions: left, right, and twosided  (some operations may even change the direction).   Accordingly, we have `mutable struct PBWAlgIdeal{D, T, S}`, where `D` stands for the direction. For me it does not make sense to have one of these directions as the default. I think you cannot compare the situation here with the situations where the above mentioned design principle applies.

thofma:
It is fine with me if this is how the PBW algebras and their modules should be implemented. But what is implemented here is
https://github.com/oscar-system/Oscar.jl/blob/9b269b819d9cba4676fc03e110fa38495c8a12f6/src/Modules/UngradedModules/FreeModElem.jl#L264-L270
which is, I think, both not adhering to the Oscar standard and against the idea to not have a default.

HechtiDerLachs:
> one of Oscars design principle says that we use right actions, so the modules should be right modules by default.

I was afraid this was the case. However, this would be a whole different load of work! All of the code for `ModuleFP`s has originally been developed for commutative rings and the implementation uses left-multiplication. If we were to make the `ModuleFP`s right modules by default, we would have to go through the entire code base and exchange `a*v` by `v*a` for scalars `a` and `ModuleFPElem`s/`SRow`s/... `v`. We can do that, but that would be an expensive task and not covered in this small PR.

> Catching all `NCRingElem` makes little sense

So does catching all `RingElem`s for the existing implementations. Yet, the signatures said `Ring` and `RingElem` and not `MPolyRing` and `MPolyRingElem`. For the generic types I would therefore be in favor of widening the signature to `NCRingElem`, regardless of what implementations we really do offer in the end. 

For example, I have implemented a custom type `ExteriorAlgebra{T}` for exterior algebras over arbitrary rings. This is not a `PBWAlgebra`, because the ground ring is not necessarily a field and I wanted to be able to write generic code which does not bind to the singular backend directly. I am using the `ModuleFP`s over `ExteriorAlgebra{T}`s and it works just fine. But that would not be possible, if the signature was restricted to `PBWAlgElem`s. Plus: If we made that restriction, then we would also disallow external users to use the Oscar modules with their custom non-commutative rings.

HechtiDerLachs:
> which is, I think, both not adhering to the Oscar standard and against the idea to not have a default.

Probably true. I would like to hear an executive decision about this. Personally, I do not really care, but this should be decided somehow. These lines are just a minimal fix to make things half-way coherent for the moment. And given that, as I explained above, the current implementation of the generic module code is unable to support right-modules, this is the only feasible solution for the moment.

wdecker:
> both not adhering to the Oscar standard and against the idea to not have a default.

@thofma Oh, I see now. I perfectly agree with what you write.

thofma:
> > one of Oscars design principle says that we use right actions, so the modules should be right modules by default.
> 
> I was afraid this was the case. However, this would be a whole different load of work! All of the code for `ModuleFP`s has originally been developed for commutative rings and the implementation uses left-multiplication. If we were to make the `ModuleFP`s right modules by default, we would have to go through the entire code base and exchange `a*v` by `v*a` for scalars `a` and `ModuleFPElem`s/`SRow`s/... `v`. We can do that, but that would be an expensive task and not covered in this small PR.

First of all, I am sympathetic to small changes that enable some nice applications. But the fundamental design decision on how modules over noncommutative rings should be handled is probably beyond a small PR:
- The concept of left/right/bi and how it is handled is not thought through in my opinion.
- This is not something experimental and once it is in, it will never be done properly.
- It should take into account what is there already in terms of modules over noncommutative rings and should not go exactly the opposite direction.

So I don't thinks this is a good argument for going against the design principle and the rest of Oscar.

Changing the signature to only allow `PBWAlgebra` would be a hack to get something working, without us committing to some design. Although given the answer by Wolfram, I am also not sure if this is what is required/wished for in case of `PBWAlgebra`s.

HechtiDerLachs:
> But the fundamental design decision on how modules over noncommutative rings should be handled is probably beyond a small PR

Given the above discussion, you are probably right. I understand your concerns. Even though this is rather sad, because I would have liked to have these changes supported in Oscar with a view towards cohomology computations via exterior algebras. But if I have to wait for a coherent framework of modules over non-commutative rings to first be developed, then that is what it is. In that case we can close the PR and just continue with #3900.

jankoboehm:
There is a chance that we can follow the situation for ideals. Then what we have in this PR would exactly be what is needed, there is just some logic missing on top: We reduce the right case to the left, and the two-sided to the right. What would be needed is that the underlying algorithmic oracles can support that. Groebner bases should.

wdecker:
> two-sided to the right.

For PBW algebras we compute a LEFT GB first and enlarge is successively until it is a right GB, too.

HechtiDerLachs:
I want to apologize to anyone involved here. Today I was informed that similar discussions had already taken place elsewhere, also offline. I didn't mean to overthrow or question the considerations and conclusions that were eventually reached at other occasions. I suggest to continue any discussion about these matters in person when we meet again.

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -254,28 +254,20 @@ function Base.deepcopy_internal(a::AbstractFreeModElem, dict::IdDict)
...
+*(a::Any, b::AbstractFreeModElem) = parent(b)(base_ring(parent(b))(a)*coordinates(b))
 
-function *(a::MPolyRingElem, b::AbstractFreeModElem) 
-  if parent(a) !== base_ring(parent(b))
-    return base_ring(parent(b))(a)*b # this will throw if conversion is not possible
-  end
-  return parent(b)(a*coordinates(b))
+function *(a::T, b::AbstractFreeModElem{T}) where {T <: NCRingElem}
+  parent(a) === base_ring(parent(b)) && return parent(b)(a*coordinates(b))
+  return parent(b)(base_ring(parent(b))(a)*coordinates(b))
```
HechtiDerLachs:
```suggestion
function *(a::T, b::AbstractFreeModElem{T}) where {T <: NCRingElem}
  @assert is_left(parent(b)) "left multiplication is not defined for non-left module $(parent(b))"
  parent(a) === base_ring(parent(b)) && return parent(b)(a*coordinates(b))
  return parent(b)(base_ring(parent(b))(a)*coordinates(b))
```

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -254,28 +254,20 @@ function Base.deepcopy_internal(a::AbstractFreeModElem, dict::IdDict)
...
 end
 
-function *(a::RingElem, b::AbstractFreeModElem) 
-  if parent(a) !== base_ring(parent(b))
-    return base_ring(parent(b))(a)*b # this will throw if conversion is not possible
-  end
-  return parent(b)(a*coordinates(b))
+function *(b::AbstractFreeModElem{T}, a::T) where {T <: NCRingElem}
+  error("all modules are left-modules by default")
```
HechtiDerLachs:
```suggestion
function *(b::AbstractFreeModElem{T}, a::T) where {T <: NCRingElem}
  @assert is_right(parent(b)) "right multiplication not defined for non-right module $(parent(b))"
  error("right multiplication is not supported at the moment")
```

src/Modules/UngradedModules/FreeModElem.jl
```diff
@@ -254,28 +254,20 @@ function Base.deepcopy_internal(a::AbstractFreeModElem, dict::IdDict)
...
 end
 
-*(a::Int, b::AbstractFreeModElem) = parent(b)(a*coordinates(b))
-*(a::Integer, b::AbstractFreeModElem) = parent(b)(base_ring(parent(b))(a)*coordinates(b))
-*(a::QQFieldElem, b::AbstractFreeModElem) = parent(b)(base_ring(parent(b))(a)*coordinates(b))
+function *(b::AbstractFreeModElem{T}, a::Any) where {T <: RingElem}
+  error("scalar multiplication from the right is not defined; use left-multiplication instead")
+end
```
HechtiDerLachs:
```suggestion
function *(b::AbstractFreeModElem{T}, a::Any) where {T <: RingElem}
  error("scalar multiplication from the right is not yet supported")
end

# Methods to determine whether a module is a left-, right-, or bi-module. 
# We plan to have flags set for this. But for the moment the generic code only supports left-multiplication, 
# so we can decide this from the type alone. How we do it in the long run is not yet decided, but in either case
# we want to use these functions to decide as they are already there for ideals. 
is_left(M::ModuleFP) = is_left_module(typeof(M))
is_left(::Type{T}) where {RET<:RingElem, T<:ModuleFP{T}} = true
is_left(::Type{T}) where {RET<:NCRingElem, T<:ModuleFP{T}} = true # Left multiplication is generically supported

is_right(M::ModuleFP) = is_right_module(typeof(M))
is_right(::Type{T}) where {RET<:RingElem, T<:ModuleFP{T}} = true
is_right(::Type{T}) where {RET<:NCRingElem, T<:ModuleFP{T}} = false # Right multiplication is not supported by the generic code at the moment, but we plan to do so eventually. 

is_two_sided(M::ModuleFP) = is_right_module(typeof(M))
is_two_sided(::Type{T}) where {RET<:RingElem, T<:ModuleFP{T}} = true
is_two_sided(::Type{T}) where {RET<:NCRingElem, T<:ModuleFP{T}} = false # see above

```

HechtiDerLachs (commented):
A first write-up of today's discussion. 

In addition, we will insert a type union `AdmissibleModuleFPRingElem = Union{<:RingElem, <:PBWAlgElem, <:PBWAlgQuoElem, ...}` and use that in the restriction for the type parameters of the module types and functions.

In the long run we plan to support right-modules similar to the case of ideals: For an `NCRing` `A` there is an "opposite algebra" `A_op` and a bijective morphism `phi : A -> A_op`. This should extend generically to modules `phi : M -> M_op`. For a right-module `M` we can then define multiplication for elements `a` in `A` and `v` in `M` via 
`v*a = inv(phi)(phi(a)*phi(v))`. The opposite algebra and -modules are, of course, cached.

thofma:
I had a short discussion with @wdecker and we concluded that there should be `left_free_module`/`right_free_module`, but no `free_module` for noncommutative rings.

src/Modules/UngradedModules/FreeModElem.jl
lgoettgens: In L280 and L284 I think you forgot to update the rhs when copying it there

HechtiDerLachs:
What happened to this PR? I tried to push some more fixes, but it rejected me. When I pulled, I got all that stuff from other people. Was this a mistake?

HechtiDerLachs:
Tests pass. From my side this is good to go.

thofma:
Did you by any chance miss https://github.com/oscar-system/Oscar.jl/pull/3988#issuecomment-2308907093? I am happy to talk about this/do the changes next week in person.

thofma:
Does this here only implement `FreeMod` but not `free_module`?

HechtiDerLachs:
Yes, I think, I took out the `free_module` implementation from Lakshmi already before for the obvious reasons. But that reminds me that I didn't yet add the other constructors.

thofma:
OK. Not sure what you plan to add exactly, but with @wdecker we agreed to not have `free_module` but only `left_free_module/right_free_module` in the general (noncommutative) case. We can also talk about this on Monday.

HechtiDerLachs:
Yes, let's finalize this on Monday together then.

thofma (approved):

PR #3989, opened 2024-08-02
A function checking if a tropical polytope is ordinarily convex (ie a polytrope)

Sami-Halaseh:


Sami-Halaseh:
@ooinaruhugh @lenweis

Sami-Halaseh:
@danteluber

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,30 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
     return helper(transpose(A),nca,nra)
   end
 end
+
+@doc raw"""
+    is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
```
fingolfin: This docstring does not fit the function it is attached to

Sami-Halaseh: Sorry, this is my first time creating a pull request and I didn't submit the most recent changes. It should be 'correct' now.

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,29 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
     return helper(transpose(A),nca,nra)
   end
 end
+
+@doc raw"""
+    is_polytrope(A::QQMatrix)
```
YueRen:
```suggestion
    is_polytrope(A::QQMatrix, minOrMax::Union{typeof(min),typeof(max)})
```

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,29 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+julia> A = QQ[0 0 1; 0 1 0; 0 3 3]
+[0   0   1]
+[0   1   0]
+[0   3   3]
+
+julia> is_polytrope(A,min)
+true
+```
+"""
+function is_polytrope(A::QQMatrix, MorM)
```
YueRen:
```suggestion
function is_polytrope(A::QQMatrix, MoM::Union{typeof(min),typeof(max)})
```

src/TropicalGeometry/matrix.jl
YueRen: Additionally, the convention in the tropical part of OSCAR is that `min` is default.  If you think that is more confusing than helpful for polytrope, then please ignore that convention.  Otherwise please add it as follows `minOrMax::Union{typeof(min),typeof(max)}=min`.

YueRen (changes requested):
Dear @Sami-Halaseh,

Thanks for the pull-request!  Overall it looks great, I just have a few super-minor changes.  

One general question:  Wouldn't it be more sensible to make the input a polytope?

Best regards, Yue.

src/TropicalGeometry/matrix.jl
```diff
@@ -76,27 +76,43 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
-  Polymake.Shell.CV = pPMCV
+function is_polytrope(A::MatrixElem{<:TropicalSemiringElem})
+  #=Convert the matrix of tropical numbers into matrix of rational numbers in
+  to pass it into polymake's tropical convex hull function =#
+  ncA = ncols(A)
+  nrA = nrows(A)
+  QQA = zero_matrix(QQ,nrA,ncA)
+  for i in 1:nrA
+    for j in 1:ncA
+      @req A[i,j] != zero(tropical_semiring(convention(A))) "If infty is an entry then the tropical polytope is not bounded"
```
YueRen:
```suggestion
      @req !iszero(A[i,j]) "If infty is an entry then the tropical polytope is not bounded"
```

This is a bit shorter.  Also, are you sure that you want to raise an error (as opposed to simply returning `false`)?

src/TropicalGeometry/matrix.jl
```diff
@@ -76,27 +76,43 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+  PMCV = P.POLYTOPE_MAXIMAL_COVECTORS
+  Polymake.Shell.CV = PMCV
+  #Get the length out of the polymake shell
   Polymake.shell_execute(raw"""$tmp = $CV->size;""")
   l = Polymake.Shell.tmp
-  return l == 1
-end
\ No newline at end of file
+  l != 1 && return false
+  return true
```
YueRen: Is there any difference between this and the previous
```
  return l == 1
```
?

If not, I'd prefer the previous version which is a line shorter.

YueRen (changes requested):

src/TropicalGeometry/matrix.jl
Sami-Halaseh: I think that is better, I have made the change

src/TropicalGeometry/matrix.jl
Sami-Halaseh: Yes I think returning `false` is a better idea

Sami-Halaseh:
> Dear @Sami-Halaseh,
> 
> Thanks for the pull-request! Overall it looks great, I just have a few super-minor changes.
> 
> One general question: Wouldn't it be more sensible to make the input a polytope?
> 
> Best regards, Yue.

@YueRen thank you for the feedback. I have made the changes you suggested over a few commits (sorry about that). One thing I have not addressed is this comment you made. This is actually a confusion of mine. Are tropical polytopes a type in OSCAR? I thought the way do tropical convex hull computations was to somehow go through polymake. For example
```
tropical > $H = new Polytope<Min>(POINTS=>[[0,1,0],[0,0,1],[1,0,0]]);
julia> H = Polymake.Shell.H
name: H
type: Polytope<Min, Rational>
...
julia> typeof(H)
Polymake.LibPolymake.BigObjectAllocated
```
Here it seems like I get two different types. The one when I pull $P into julia out of Polymake says something about type Polytope<Min, Rational>. But then the typeof gives something else. Excuse my ignorance here. At MEGA you suggested I speak with Michael Joswig about tropical convexity in OSCAR. I have reached out about a meeting so hopefully that will help soon.

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,70 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
     return helper(transpose(A),nca,nra)
   end
 end
+
+@doc raw"""
+    is_polytrope(A::MatrixElem{<:TropicalSemiringElem})
+
+Check if a tropical polytope is ordinarily convex (ie a polytrope)
```
YueRen:
```suggestion
Check if the tropical convex hull of the rows of `A` is ordinarily convex (ie a polytrope)
```

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,70 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+  #Get the length out of the polymake shell
+  Polymake.shell_execute(raw"""$tmp = $CV->size;""")
+  l = Polymake.Shell.tmp
+  return l == 1
+end
+
+@doc raw"""
+    is_polytrope(A::QQMatrix, minOrMax::Union{typeof(min),typeof(max)}=min)
+
+Check if a tropical polytope is ordinarily convex (ie a polytrope)
```
YueRen:
```suggestion
Check if the `minOrMax`-tropical convex hull of the rows of `A` is ordinarily convex (ie a polytrope)
```

YueRen (changes requested):
Dear @Sami-Halaseh,

The pull request looks great, maybe just two little changes to the docstring to make it clearer and it's good to be merged.  If the changes works for you, you can just ask github to apply my suggestions (no need to manually change them yourself).

Best regards, Yue.

Sami-Halaseh:
> Dear @Sami-Halaseh,
> 
> The pull request looks great, maybe just two little changes to the docstring to make it clearer and it's good to be merged. If the changes works for you, you can just ask github to apply my suggestions (no need to manually change them yourself).
> 
> Best regards, Yue.

Hey Yue, thanks for the feedback. I think those suggestions are good and I have committed them. We are talking about a tropical polytope/polyhedron type but it looks like it will take some time to flesh out what it should look like and how to put it into OSCAR. For now I think its good for this function to work on matrices.
Sami

thofma:
It is kind of opposite the design of Oscar to have something like
```
function is_polytrope(A::QQMatrix)
```
For the same reason we don't have `is_bijective(A::QQMatrix)`, although one could interpret `A` in some (arbitrary) way as a linear map between vector spaces.

YueRen:
> It is kind of opposite the design of Oscar to have something like
> 
> ```
> function is_polytrope(A::QQMatrix)
> ```
> 
> For the same reason we don't have `is_bijective(A::QQMatrix)`, although one could interpret `A` in some (arbitrary) way as a linear map between vector spaces.

Then how about renaming the function to `do_rows_form_polytrope`?

benlorenz:
Maybe we can keep this function internal (non-exported) for now, until we have the tropical polytope type? Then we don't need to commit to a public interface function but it is still possible to be used for development.

YueRen:
@benlorenz That's also a good solution (removing the export part of the changes).  @Sami-Halaseh What do you say?  

1. Rename the function to `do_rows_form_polytrope` (or something similar; optional) 
2. Remove the export of the function so that it remains private 

and wait for the tropical polytope type before having a public `is_polytrope` function?

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,70 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+  tempP = Polymake.tropical.Polytope{convention(A)}(POINTS=QQA)
+  #=Compute the tropical convex hull again, this time with 
+  the minimal generating set of vertices of tropical polytope=#
+  P = Polymake.tropical.Polytope{convention(A)}(POINTS=tempP.VERTICES)
+  PMCV = P.POLYTOPE_MAXIMAL_COVECTORS
+  Polymake.Shell.CV = PMCV
+  #Get the length out of the polymake shell
+  Polymake.shell_execute(raw"""$tmp = $CV->size;""")
+  l = Polymake.Shell.tmp
+  return l == 1
```
benlorenz:
```suggestion
  return length(P.POLYTOPE_MAXIMAL_COVECTORS) == 1
```
If you change the required Polymake version in the Project.toml to 0.11.20 you can avoid the (slightly fragile) polymake shell commands.

I added `Array{IncidenceMatrix}` as a wrapped type:
```julia
julia> c = Polymake.tropical.cyclic{min}(3,4);

julia> length(c.POLYTOPE_MAXIMAL_COVECTORS)
9

julia> c.POLYTOPE_MAXIMAL_COVECTORS[1]
4×4 IncidenceMatrix
[2, 3, 4]
[1]
[1]
[1]
```

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,70 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+```
+"""
+function is_polytrope(A::QQMatrix, minOrMax::Union{typeof(min),typeof(max)}=min)
+  tempP = Polymake.tropical.Polytope{minOrMax}(POINTS=A)
+  P = Polymake.tropical.Polytope{minOrMax}(POINTS=tempP.VERTICES)
+  pPMCV = P.POLYTOPE_MAXIMAL_COVECTORS
+  Polymake.Shell.CV = pPMCV
+  Polymake.shell_execute(raw"""$tmp = $CV->size;""")
+  l = Polymake.Shell.tmp
+  return l == 1
```
benlorenz:
```suggestion
  return length(P.POLYTOPE_MAXIMAL_COVECTORS) == 1
```
as above

Sami-Halaseh:
Thank you for the feedback everyone. @YueRen I made those two changes. @benlorenz Thank you for the fix, the perl shell code was definitely not a good way to get the length. The wrapped type is much better, we just couldn't figure out how to do that properly.
We are working on a tropical polytope type in OSCAR. There are some bigger things to figure out since it would be nice for them to be able to exist in tropical projective space, not just the tropical torus. Hopefully there will be something to show in the coming months.

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,63 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+Check if the tropical convex hull of the rows of `A` is ordinarily convex (ie a polytrope)
+
+# Examples
+```jldoctest
+julia> A = tropical_semiring()[0 0 1; 0 1 0; 0 3 3]
+[(0)   (0)   (1)]
+[(0)   (1)   (0)]
+[(0)   (3)   (3)]
+
+julia> do_rows_form_polytrope(A)
```
benlorenz:
```suggestion
julia> Oscar.do_rows_form_polytrope(A)
```

src/TropicalGeometry/matrix.jl
```diff
@@ -74,3 +74,63 @@ function is_tropically_generic(A::MatrixElem{<:TropicalSemiringElem})
...
+Check if the `minOrMax`-tropical convex hull of the rows of `A` is ordinarily convex (ie a polytrope)
+
+# Examples
+```jldoctest
+julia> A = QQ[0 0 1; 0 1 0; 0 3 3]
+[0   0   1]
+[0   1   0]
+[0   3   3]
+
+julia> do_rows_form_polytrope(A,min)
```
benlorenz:
```suggestion
julia> Oscar.do_rows_form_polytrope(A, min)
```

test/TropicalGeometry/matrix.jl
```diff
@@ -17,3 +17,22 @@ end
   @test is_tropically_generic(A) == false
   @test is_tropically_generic(transpose(A)) == false
 end
+
+@testset "testing for polytropes" begin
```
benlorenz, line 21:
```suggestion
@testset "testing for polytropes" begin
  do_rows_form_polytrope = Oscar.do_rows_form_polytrope
```

benlorenz (commented):
This now needs the `Oscar.` prefix until exported again.

YueRen (approved):
Thanks for incorporating all suggestions!  Looking forward to having polytropes in OSCAR, in the meantime this function is an excellent bandaid.

micjoswig:
This is the wrong algorithm; must be replaced.  Will be dealt with in #4001 .

@ooinaruhugh

PR #3993, opened 2024-08-06
fix Lloyd's problem - kind off

fieker:
He reported problems with fixed_field, however the galois group was wrong...
Problem(s)
 - isinteger uses the reco stuff from poly factor in Hecke. It assumes that the precision is strictly increasing and that only the max precision is used, Galois needs to jump down (at least for eficiency)
 - if the reco for a huge precisino is used on a small element, it will always succeed, being mapped to identity in this case
 - the extra bits of precision were ignored
 - the test for isinteger should possibly actually call is_integer in the end to use the minpoly test Alternatively: extra precision is used here
 - there are 2 different denominators involved here:
   - to make the roots of the poly algebraic integers
   - the difference bewtween any_order and the ring of integers Currently, the rel galois code will fails for non-monic polys - I think

### DIFF
--- a/src/NumberTheory/GaloisGrp/GaloisGrp.jl
+++ b/src/NumberTheory/GaloisGrp/GaloisGrp.jl
@@ -25,6 +25,7 @@ function __init__()
   GAP.Packages.load("ferret"; install=true)
 
   Hecke.add_verbosity_scope(:GaloisGroup)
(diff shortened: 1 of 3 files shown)


thofma (approved):
Did not follow all the details, but if it works, it works.

Issue #3992, opened 2024-08-05
Crashes with empty ideals in Singular


HechtiDerLachs:
To reproduce:
```julia
P, (x, y) = QQ[:x, :y]
I = ideal(P, elem_type(P)[])
radical_membership(x, I) # Boom!
```
~Fixed in #3988 .~ Fixed in #3995.

It might be worthwhile to investigate this nevertheless. The crash looked really bad with memory access issues in the singular kernel. Maybe something else should also be fixed on another level? (ping @hannes14 )

fingolfin:
Resolved now but @HechtiDerLachs wanted to talked with @hannes14 to see if there is more problems with zero ideals ...

@afkafkafk13 points out that having an empty list of generators might be a problem for Singular. So maybe the code which translates our ideals into Singular ideals should also be adjusted to work with ideals with empty generators?

@hannes14 will take a look

thofma:
From triage: The original issue is fixed. If we find a new error due to "empty ideals", a new issue should be created.

PR #3995, opened 2024-08-06
Some module improvements

HechtiDerLachs:
Everything from #3988 which is not about non-commutative rings; in particular a fix to #3992.

### DIFF
--- a/src/Modules/UngradedModules/FreeModuleHom.jl
+++ b/src/Modules/UngradedModules/FreeModuleHom.jl
@@ -488,8 +488,8 @@ function _simple_kernel(h::FreeModuleHom{<:FreeMod, <:FreeMod})
   g = images_of_generators(h)
   b = ModuleGens(g, G, default_ordering(G))
   M = syzygy_module(b)
-  v = elem_type(F)[sum(c*F[i] for (i, c) in coordinates(repres(v)); init=zero(F)) for v in gens(M)]
-  I, inc = sub(F, v)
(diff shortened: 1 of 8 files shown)


src/Modules/UngradedModules/FreeModuleHom.jl
```diff
@@ -511,6 +510,8 @@ function kernel(h::FreeModuleHom{<:FreeMod, <:SubquoModule})
   g = vcat(g, relations(M))
   R = base_ring(G)
   H = FreeMod(R, length(g))
+  # This code is also used by graded modules and we need to care for that.
+  is_graded(h) && is_homogeneous(h) && set_grading!(H, degree.(g))
```
joschmitt, line 514: Is it possible to add a meaningful test for this? (I assume that the current tests don't really cover this case because they worked before.)

src/Rings/mpoly-ideals.jl
```diff
@@ -1608,6 +1608,7 @@ false
 ```
 """
 function radical_membership(f::T, I::MPolyIdeal{T}) where T
+  iszero(I) && return iszero(f)
```
joschmitt, line 1611: Maybe also add a test for this, to be safe.

joschmitt (approved):
Thank you!

This looks reasonable enough and the CI is happy (besides the GC issues with the book tests).

joschmitt:
Now the book tests got stuck and timed out apparently during the vinberg2 jlcon. I restart for now.

PR #3998, opened 2024-08-08
Intersection theory: Correct one method for the function `abstract_fl…

wdecker:
…ag_variety and rename this method to function `abstract_flag_bundle`


wdecker:
I have also added a new example for the function `degeneracy_locus` which makes implicitly use of the corrected method

joschmitt (approved):

PR #4004, opened 2024-08-13
Turn `semistandard_tableaux(shape, weight)` into an iterator

joschmitt:
Turn the last method from JuLie into a proper iterator.

This was a recursive function before, which I more or less managed to make iterative without too many changes. I don't know whether the algorithm implemented is published anywhere.
It seems to be a bit faster now for whatever reason.

### DIFF
--- a/src/Combinatorics/EnumerativeCombinatorics/tableaux.jl
+++ b/src/Combinatorics/EnumerativeCombinatorics/tableaux.jl
@@ -492,109 +492,198 @@ end
 Return an iterator over all semistandard Young tableaux with shape `s` and given
 weight. This requires that `sum(s) = sum(weight)`.
 """
-function semistandard_tableaux(s::Vector{T}, weight::Vector{T}) where T <: Integer
-  # TODO: This function first fills an array and then returns an iterator over
-  # this array. This was a stopgap, to have a stable API for version 1.0.
-  # The output should be a dedicated iterator type which constructs the tableaux
-  # on demand (if possible).
-  n_max = sum(s)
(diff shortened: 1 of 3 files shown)


thofma:
Could you have a quick glance @mjrodgers?

src/Combinatorics/EnumerativeCombinatorics/tableaux.jl
```diff
@@ -492,109 +492,199 @@ end
...
-      return
-    end
+  return nothing
+end
 
-    #here starts the main part of the function
-    tracker_row = zeros(Int, weight[n])
+function _fill_boxes!(n::Int, S::SemiStandardTableauxFixedShapeAndWeight{T}, state::SemiStandardTableauxFixedShapeAndWeightState{T}) where T
+  # Note: n might be different from state.n if the calling function did not update
+  # state yet
```
mjrodgers: Is there a way to work around this, maybe just ensure the iterator updates the state before calling this? I'm not sure this really matters though.

mjrodgers (approved):
I think this looks good, and glad the speed is improved a bit too. I support merging this, it's definitely good to have this as an iterator.

src/Combinatorics/EnumerativeCombinatorics/tableaux.jl
joschmitt: Yes, you're right. I changed so that `state.n` is updated before this function is called. (I still think it's clearer if `n` is an explicit argument of the function, so I left this as it was.)

PR #4005, opened 2024-08-13
Include correct doc string

joschmitt:
Without this, Documenter seems to include every `ideal` doc string.

Same as #4003 but now with a branch in oscar-system so that there is a preview: https://docs.oscar-system.org/previews/PR4005/AlgebraicGeometry/Schemes/RationalPointsProjective/ .


thofma (approved):

thofma:
Looks good

PR #4008, opened 2024-08-13
[FTheoryTools] More on serialization of FTheoryTools

HereAround:
As mentioned in the other PRs, it seemed natural to group all the different serialization discussions we had in one PR. For this, I need your help @antonydellavecchia whenever you are ready (not urgent).

Points to be addressed in this PR include (but are not necessarily limited to the following):
* The coordinates of the base space and ambient space should be remembered upon serialization. I will try to add a first step towards this goal in this PR/draft. As I see it, this requires that we add a list of strings (or symbols?) to the serialization of the models. An update of the QSM model artifacts is then needed.

The below points were initially meant to be fixed in this PR, but we (@antonydellavecchia and I) decided that it will be better to deal with them in separate PRs.

* Along these lines, we could also consider serialzing G4-fluxes (a.k.a. cohomology classes on toric spaces). At least for the QSMs, there is each time a specifically chosen G4-fluxes (cf. https://arxiv.org/pdf/2102.10115, equ. 2.25).
* Maybe wait for https://github.com/oscar-system/Oscar.jl/pull/3980 and include the Lie algebras of the FTheoryModels in the serialization. In particular, this should be also applied to the QSM models.
* Add an artifact for the model discussed in https://github.com/oscar-system/Oscar.jl/pull/3978. I have the corresponding .mrdi files and will be happy to share e.g. in slack or email. (This cannot be added here as github does not seem to support .mrdi files yet. And then, the files are about 0.5GB in size. So maybe best not to add them here.)
* Make use of the attribute saver that @antonydellavecchia is working on (cf. https://github.com/oscar-system/Oscar.jl/pull/4020). For instance, currently the known resolutions are not saved for FTheory models.

cc @antonydellavecchia @emikelsons @apturner

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,10 +1,10 @@
 [QSMDB]
-git-tree-sha1 = "5c82e5ff3aa99f2aa660c9ffb3caa57ee535cc21"
+git-tree-sha1 = "52686066016440cf2e6e286a923aed887658543c"
 lazy = true
 
      [[QSMDB.download]]
-     sha256 = "3f69aae0fca02d74764f709c622388c564e6e144a8b11347dadef5725a65b0a2"
-     url = "https://github.com/oscar-system/Oscar.jl/releases/download/archive-tag-1/qsmdb.tar.gz"     
+     sha256 = "2b0a368c07d368f3973352b7f6e856214e25383d2cc3a78c70415afe247a997e"
+     url = "https://github.com/oscar-system/Oscar.jl/releases/download/archive-tag-1/QSMDB-2.tar.gz"     
 
 [gap_extraperfect]
 git-tree-sha1 = "084fa12573e5089ceb3299f9d341f244b415da52"
--- a/experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
+++ b/experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
@@ -1741,7 +1741,7 @@ end
 
 
 @doc raw"""
-    simplified dual_graph(m::AbstractFTheoryModel)
+    simplified_dual_graph(m::AbstractFTheoryModel)
 
 This method returns the simplified dual graph of the QSM model in question.
 Note that no labels are (currently) attached to the vertices/nodes or edges.
--- a/experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
+++ b/experimental/FTheoryTools/src/HypersurfaceModels/constructors.jl
@@ -72,7 +72,7 @@ function hypersurface_model(base::NormalToricVariety, fiber_ambient_space::Norma
   @req length(ds) == 1 "Inconsistency in determining the degree of the hypersurface equation"
   @req ds[1] == divisor_class(anticanonical_divisor_class(ambient_space)).coeff "Degree of hypersurface equation differs from anticanonical bundle"
   explicit_model_sections = Dict{String, MPolyRingElem}()
-  gens_S = gens(cox_ring(ambient_space))
+  gens_S = gens(cox_ring(base))
   for k in 1:length(gens_S)
     explicit_model_sections[string(gens_S[k])] = gens_S[k]
   end
--- a/experimental/FTheoryTools/src/Serialization/hypersurface_models.jl
+++ b/experimental/FTheoryTools/src/Serialization/hypersurface_models.jl
@@ -66,16 +66,17 @@ end
 ##########################################
 
 function load_object(s::DeserializerState, ::Type{<:HypersurfaceModel}, params::Tuple{NormalToricVariety, NormalToricVariety, NormalToricVariety, <:MPolyRing, <:MPolyRing})
-  base_space, ambient_space, fiber_ambient_space, R1, R2 = params
+  base_space, amb_space, fiber_ambient_space, R1, R2 = params
   defining_equation = load_object(s, MPolyRingElem, R1, :hypersurface_equation)
   defining_equation_parametrization = load_object(s, MPolyRingElem, R2, :hypersurface_equation_parametrization)
   explicit_model_sections = haskey(s, :explicit_model_sections) ? load_typed_object(s, :explicit_model_sections) : Dict{String, MPolyRingElem}()
   defining_classes = haskey(s, :defining_classes) ? load_typed_object(s, :defining_classes) : Dict{String, ToricDivisorClass}()
-  model = HypersurfaceModel(explicit_model_sections, defining_equation_parametrization, defining_equation, base_space, ambient_space, fiber_ambient_space)
+  model = HypersurfaceModel(explicit_model_sections, defining_equation_parametrization, defining_equation, base_space, amb_space, fiber_ambient_space)
   model.defining_classes = defining_classes
   attrs_data = haskey(s, :__attrs) ? load_typed_object(s, :__attrs) : Dict{Symbol, Any}()
   for (key, value) in attrs_data
     set_attribute!(model, Symbol(key), value)
   end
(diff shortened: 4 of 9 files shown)


HereAround:
@antonydellavecchia , as we discussed during the coding sprint in KL, I will move most of the above ToDos to https://github.com/oscar-system/Oscar.jl/issues/2699 and will then try to brush up this PR.

HereAround:
@antonydellavecchia I have brushed up this PR. In particular, I have updated the QSMDB. I am not sure if we can modify the current QSMDB artifact (attached to a release tag of OSCAR) I have moved it to https://martinbies.github.io/Materials/QSMDB/qsmdb.tar.gz, where I can modify/edit it more conveniently. Not sure if this is the best move though and am happy to change this. Please let me know what you think.

In any case - I would expect that with the current changes, all tests pass and thus, this PR is getting ready to be merged.

HereAround:
@antonydellavecchia I am puzzled why the tests fail. The new artifact files should not cause the problem in the failures. So I am wondering if somehow OSCAR fails to fetch those new artifact files?

HereAround:
@antonydellavecchia As discussed in slack, I have added the new files to the release tag now. Hopefully the tests do succeed now.

Artifacts.toml
```diff
@@ -1,10 +1,10 @@
 [QSMDB]
-git-tree-sha1 = "5c82e5ff3aa99f2aa660c9ffb3caa57ee535cc21"
+git-tree-sha1 = "930fd374bd91a4b19d7adbf0cd4b5517c31b5ab2"
 lazy = true
 
      [[QSMDB.download]]
```
antonydellavecchia, line 5: I am not sure if this name now needs to change maybe?

antonydellavecchia:
```suggestion
     [[QSMDB-2.download]]
```
Maybe this? but I am not sure, I would have to experiment as well

HereAround:
Ok. I think I have finally succeeded. At least on my local system it works.

I have renamed the previous QSMDB into QSMDB-old and added the new version as QSMDB. The latter is larger in size by about 1MB than the old file. I believe this is because the hypersurface models do remember more details now than they used to at the time when the initial database was created.

benlorenz:
> Ok. I think I have finally succeeded. At least on my local system it works.
> 
> I have renamed the previous QSMDB into QSMDB-old and added the new version as QSMDB. The latter is larger in size by about 1MB than the old file. I believe this is because the hypersurface models do remember more details now than they used to at the time when the initial database was created.

I changed the name back (of the file in the github release), the old file must stay exactly as it is. It is still used in the existing oscar releases.
I moved the new file to `QSMDB-2.tar.gz`.

Artifacts.toml
benlorenz: No, the name of the artifact should stay unless you want to change all occurrences...

Artifacts.toml
```diff
@@ -1,9 +1,9 @@
 [QSMDB]
-git-tree-sha1 = "5c82e5ff3aa99f2aa660c9ffb3caa57ee535cc21"
+git-tree-sha1 = "ee9e07704b9e42ddd8f1d8751873f9e12a133318"
 lazy = true
 
      [[QSMDB.download]]
-     sha256 = "3f69aae0fca02d74764f709c622388c564e6e144a8b11347dadef5725a65b0a2"
+     sha256 = "f2ce521a152738af28e957502a8fad3e84d9962937490e080a9b4bd598a8c191"
      url = "https://github.com/oscar-system/Oscar.jl/releases/download/archive-tag-1/qsmdb.tar.gz"
```
benlorenz:
```suggestion
     url = "https://github.com/oscar-system/Oscar.jl/releases/download/archive-tag-1/QSMDB-2.tar.gz"     
```

benlorenz:
The name of the artifact and the filename in the url do not need to match, with the files renamed again as mentioned above the following works fine:
```
lorenz@eddie /tmp/artifacttest $ cat Artifacts.toml 
[QSMDB]
git-tree-sha1 = "ee9e07704b9e42ddd8f1d8751873f9e12a133318"
lazy = true

     [[QSMDB.download]]
     sha256 = "f2ce521a152738af28e957502a8fad3e84d9962937490e080a9b4bd598a8c191"
     url = "https://github.com/oscar-system/Oscar.jl/releases/download/archive-tag-1/QSMDB-2.tar.gz"     

lorenz@eddie /tmp/artifacttest $ julia-latest 
[...]
julia> using LazyArtifacts

julia> LazyArtifacts.ensure_artifact_installed("QSMDB", LazyArtifacts.find_artifacts_toml("."))
     Failure artifact: QSMDB
  Downloaded artifact: QSMDB
"/home/lorenz/.julia/artifacts/ee9e07704b9e42ddd8f1d8751873f9e12a133318"

shell> ls /home/lorenz/.julia/artifacts/ee9e07704b9e42ddd8f1d8751873f9e12a133318
1014.mrdi  1212.mrdi  138.mrdi	 1647.mrdi  1982.mrdi  2236.mrdi  2482.mrdi  2744.mrdi	2939.mrdi  3138.mrdi  3315.mrdi  3520.mrdi  377.mrdi   3965.mrdi  4175.mrdi  535.mrdi  772.mrdi
1015.mrdi  121.mrdi   139.mrdi	 1650.mrdi  1983.mrdi  2245.mrdi  2483.mrdi  274.mrdi	2942.mrdi  3140.mrdi  3326.mrdi  3521.mrdi  3782.mrdi  3967.mrdi  417.mrdi   557.mrdi  77.mrdi
...
```

_Edit:_ That one download failure is probably because Pkg will try the package servers first and then use the specified url.

HereAround:
Thank you @benlorenz ! I believe this PR is now ready to be reviewed.

lgoettgens:
There are failures in the doctests

HereAround:
> There are failures in the doctests

Yeah, @benlorenz already commented on them above:

"That one download failure is probably because Pkg will try the package servers first and then use the specified url."

I do not know (yet) how to avoid this. Suggestions welcome.

Hhhmm... just realized that there are other failures too...

benlorenz:
The failure I was referring to was just this line in my small example:
```
     Failure artifact: QSMDB
```
But this is resolved automatically by retrying:
```
  Downloaded artifact: QSMDB
```

benlorenz:
The map creation (for `map_from_torusinvariant_weil_divisor_group_to_class_group`) needs to make sure that the codomain is correct for the case that the class group attribute was loaded from a file and thus exists without the map.

I pushed a fix to this branch which appends another (identity) map to the correct class group in that case:
https://github.com/HereAround/Oscar.jl/blob/6604d6e5c17f656b3b7589236ff5d8594fc37051/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl#L752-L759

HereAround:
> The map creation (for `map_from_torusinvariant_weil_divisor_group_to_class_group`) needs to make sure that the codomain is correct for the case that the class group attribute was loaded from a file and thus exists without the map.
> 
> I pushed a fix to this branch which appends another (identity) map to the correct class group in that case: https://github.com/HereAround/Oscar.jl/blob/6604d6e5c17f656b3b7589236ff5d8594fc37051/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/attributes.jl#L752-L759

Awesome! Very nice fix. Thank you very much indeed @benlorenz ! :)

HereAround:
(Nitpick: I put the fix to class group as the first commit: In this way, I do expect that all commits work individually, and so the changes in this PR can be rebased onto the master branch, rather than squashed.)

HereAround:
@antonydellavecchia could you please take a look at this PR and review it? (@apturner tells me that he does not feel competent enough to gauge the changes here, as they are mostly about serialization in OSCAR, which is your domain of expertise @antonydellavecchia ).

antonydellavecchia (approved):

PR #3812, opened 2024-05-29
Experimental: Algebraic Statistics

bkholler:
This is work in progress for implementing graphical and phylogenetic models from algebraic statistics. This builds upon the draft pull request "Adding Graphical Models in Experimental" #3621.

PR #4010, opened 2024-08-14
first attempts at writing the serializers for the new types PhylogeneticModel and GroupBasedPhylogeneticModel

mariusneu:
I am currently trying to serialize the new data types introduced in the algebraic statistics (sub)project on algebraic phylogenetics by @marinagarrote et al (cf. #3812). Following the documentation and writing the code as naive as possible, I got problems when testing the functions by trying to save an instance of GroupPhylogeneticModel, with the following error message: `Unsupported type 'GroupBasedPhylogeneticModel' for encoding.`


experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,58 @@
+@register_serialization_type PhylogeneticModel uses_params 
+@register_serialization_type GroupBasedPhylogeneticModel uses_params
```
antonydellavecchia: if you set `use_params` the save mechanices will expect there to be a save_type_params function implemented.

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,58 @@
+@register_serialization_type PhylogeneticModel uses_params 
+@register_serialization_type GroupBasedPhylogeneticModel uses_params 
+
+function save_object(s::SerializerState, pm::PhylogeneticModel)
+    save_data_dict(s) do 
+        save_object(s, pm.graph, :tree) # already implemented
```
antonydellavecchia: does graph(pm) work?  in general you should avoid acessing properties and use functions, and if the functions don't exist you should considering writting them.

experimental/AlgebraicStatistics/src/serialization.jl
mariusneu: Removed that for now

experimental/AlgebraicStatistics/src/serialization.jl
mariusneu: Now I use functions throughout the code instead of properties.

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,41 @@
...
+    save_object(s, probability_ring(pm), :probability_ring) # already implemented
+    save_object(s, root_distribution(pm), :root_distribution) # needs to be implemented, or we change the entry type to QQFieldElem
+    save_object(s, transition_matrices(pm), :transition_matrices) # needs to be implemented
+  end
+end
+
+function save_object(s::SerializerState, pm::GroupBasedPhylogeneticModel)
+  save_data_dict(s) do
+    save_object(s, phylogenetic_model(pm), :phylomodel) # details see above
+    save_object(s, fourier_ring(pm), :fourier_ring) # already implemented
```
antonydellavecchia: this should probably be saved as a reference, the best thing to do for now is just
```suggestion
    save_typed_object(s, fourier_ring(pm), :fourier_ring) # already implemented
```

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,41 @@
...
+    save_object(s, fourier_parameters(pm), :fourier_params) # needs to be implemented
+    save_object(s,group_of_model(pm), :group) # already implemented
+  end
+end
+
+
+function load_object(s::DeserializerState, ::Type{<: PhylogeneticModel})
+  graph = load_object(s, Graph{Directed}, :tree)
+  n_states = load_object(s, Int64, :states)
+  prob_ring = load_object(s, QQMPolyRing, :probability_ring)
```
antonydellavecchia: this would then become
```suggestion
  prob_ring = load_typed_object(s, :probability_ring)
```

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,41 @@
+@register_serialization_type PhylogeneticModel 
+@register_serialization_type GroupBasedPhylogeneticModel 
+
+function save_object(s::SerializerState, pm::PhylogeneticModel)
+  save_data_dict(s) do 
+    save_object(s, graph(pm), :tree) # already implemented 
+    save_object(s, number_states(pm), :states) # already implemented 
+    save_object(s, probability_ring(pm), :probability_ring) # already implemented
```
antonydellavecchia:
```suggestion
    save_typed_object(s, probability_ring(pm), :probability_ring) # already implemented
```

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,41 @@
...
+  prob_ring = load_object(s, QQMPolyRing, :probability_ring)
+  root_distr = load_object(s, Vector{Any}, :root_distribution)
+  trans_matrices = load_object(s, Dict{Edge, MatElem{QQMPolyRingElem}}, :trans_matrices)
+    
+  return PhylogeneticModel(graph, n_states, prob_ring, root_distr, trans_matrices)
+end 
+
+function load_object(s::DeserializerState, ::Type{<: GroupBasedPhylogeneticModel})
+  phylomodel = load_object(s, PhylogeneticModel, :phylomodel)
+  fourier_ring = load_object(s, QQMPolyRing, :fourier_ring)
```
antonydellavecchia:
```suggestion
  fourier_ring = load_typed_object(s, :fourier_ring)
```

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,41 @@
+@register_serialization_type PhylogeneticModel 
+@register_serialization_type GroupBasedPhylogeneticModel 
+
+function save_object(s::SerializerState, pm::PhylogeneticModel)
+  save_data_dict(s) do 
+    save_object(s, graph(pm), :tree) # already implemented 
+    save_object(s, number_states(pm), :states) # already implemented 
+    save_object(s, probability_ring(pm), :probability_ring) # already implemented
+    save_object(s, root_distribution(pm), :root_distribution) # needs to be implemented, or we change the entry type to QQFieldElem
```
antonydellavecchia: you can using save_typed_object here if the type isn't always know, or if your using a vector any because not all entries are the same type you'll need to use a tuple

src/Serialization/Combinatorics.jl
```diff
@@ -19,6 +20,19 @@ function load_object(s::DeserializerState, g::Type{Graph{T}}) where T <: Union{D
   return g(smallobj)
 end
 
+function save_object(s::SerializerState, e::Edge)
+  save_data_dict(s) do
```
antonydellavecchia: this can be done using save_data_array instead, it'll make file sizes smaller

experimental/AlgebraicStatistics/src/serialization.jl
```diff
@@ -0,0 +1,42 @@
...
+    save_object(s, phylogenetic_model(pm), :phylomodel) # details see above
+    save_typed_object(s, fourier_ring(pm), :fourier_ring) # already implemented
+    save_typed_object(s, fourier_parameters(pm), :fourier_params) # needs to be implemented
+    save_typed_object(s, group_of_model(pm), :group) # already implemented
+  end
+end
+
+
+function load_object(s::DeserializerState, g::Type{PhylogeneticModel})
+  graph = load_object(s, Graph{Directed}, :tree)
```
antonydellavecchia: here you would need to use 
```suggestion
  graph = load_typed_object(s, :tree)
```

mariusneu:
In commit 8723fc9, tests did not pass so I investigated it and found out that apparently object identity does not the same with polymake-derived objects? Therefore, the graph objects of the original model and the loaded one were not identical, although they were the same tree. I wrote a workaround using `adjacency_matrix(graph)`

fingolfin:
What the status of this PR? (Besides the fact that it has many merge conflicts now)

antonydellavecchia:
Serialization of phylogenetic trees will now be handle after the refactoring of the algebraic stats section.

PR #4012, opened 2024-08-15
Optimize printing for ideals with many generators

joschmitt:
I managed to create an ideal with ~9000 generators, for which the `oneline` printing took several seconds (to eventually just put `Ideal with 9000 generators`).
This should make the printing faster in this case.


lgoettgens (approved):
Lgtm

PR #4013, opened 2024-08-15
support group extensions of matrix groups

ThomasBreuer:
- admit `Oscar.GAPGroupElem` not only `Oscar.BasicGAPGroupElem` in evaluating 2-cochains (we need in particular `Oscar.MatrixGroupElem`)

- add a `isomorphism(::Type{FPGroup}, A::AbstractAlgebra.Generic.FreeModule)` method, which is needed by `extension` if we want to support G-modules created from matrix groups acting on vector spaces

- add tests

### DIFF
--- a/docs/src/Groups/pcgroup.md
+++ b/docs/src/Groups/pcgroup.md
@@ -14,6 +14,7 @@ map_word(g::Union{PcGroupElem, SubPcGroupElem}, genimgs::Vector; genimgs_inv::Ve
 Julia has the following functions that allow to generate polycyclic groups:
 ```@docs
 abelian_group(::Type{T}, v::Vector{Int}) where T <: GAPGroup
+elementary_abelian_group
 cyclic_group
 dihedral_group
 quaternion_group
--- a/experimental/GModule/src/Cohomology.jl
+++ b/experimental/GModule/src/Cohomology.jl
@@ -1705,10 +1705,10 @@ returns (I, q), (hom(Z[G], C), B)
 function dimension_shift(C::GModule)
   G = C.G
   if isa(C.M, FinGenAbGroup)
-    zg, ac, em = Oscar.GModuleFromGap.natural_gmodule(FinGenAbGroup, G, ZZ)
+    zg, ac, em = regular_gmodule(FinGenAbGroup, G, ZZ)
     Z = Hecke.zero_obj(zg.M)
   elseif isa(C.M, AbstractAlgebra.FPModule{<:FieldElem})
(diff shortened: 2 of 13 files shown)


fieker:
Mostly happy, but in Cohomology.m we already have 
```
# return an f.p. group `F` and an isomorphism `M -> F`
function fp_group_with_isomorphism(M::AbstractAlgebra.FPModule{<:FinFieldElem})
  p, mp = pc_group_with_isomorphism(M, refine = false)
  mf = isomorphism(FPGroup, p)
  return codomain(mf), mf*mp
end
```
I think we need to consolidate this (and make a consistent interface?) At least as a promise, the "old" function is more general as in accepts not only prime fields, and not only free_modules.
I have no idea if the way via pc-group is efficient though, so an independent implementatino might be useful

fieker:
What about also adding (maybe later)
 - natural_gmodule for matrix groups
 - natural_representation

ThomasBreuer:
Concerning a consistent interface, the code anyhow needs `isomorphism`, thus the natural solution seems to be to move the general implementations from `pc_group_with_isomorphism` and `fp_group_with_isomorphism` to corresponding `isomorphism` methods.

Concerning `natural_gmodule`, there are currently two methods which return a G-module of dimension equal to the group order (what is usually called regular module).
Thus I did not dare to use the name natural module for the G-module that is given by the action of a matrix group of degree n via right multiplication on a n-dimensional vector space.
(Similarly for the action of a permutation group of degree n on a free module of dimension n.)

fieker:
The existing functions will have to be renamed into `regular_gmodule`. And possibly also tied to `getindex(R::Ring, G::Group)` to support `Z[G]`

fieker (approved):

ThomasBreuer:
O.k., I can adjust the pull request according to the above discussion.

ThomasBreuer:
Concerning the idea

> The existing functions will have to be renamed into `regular_gmodule`. And possibly also tied to `getindex(R::Ring, G::Group)` to support `Z[G]`

There is already a matching `getindex` method in Hecke that constructs a `GroupAlgebra` object.
Thus the question is what we want here.

ThomasBreuer:
@fieker I have updated the pull request as we had discussed.
See the commit message of 757ad59 for the details.

fieker:
on first glance: looks good!

fingolfin:
I'll discuss this next week with @fieker and @ThomasBreuer to see if we can merge (and I'll try to get in a review I before that, too)

fingolfin (approved):
Looks good to me as well, @fieker should be in the office within the next hour and we can then let him merge it

PR #4015, opened 2024-08-15
[FTheoryTools] Fix constructor for 1511.03209 model

HereAround:
A small fix needed.

cc @emikelsons


emikelsons (approved):
Looks good and the tests pass

PR #4016, opened 2024-08-16
Update `generic_section` to allow specified range for random integer coefficients

apturner:
Provide keyword argument `range` for `generic_section` of toric line bundles.


HereAround (approved):
Thank you @apturner . Looks good to me!

PR #4020, opened 2024-08-19
Serialize Types with Attributes

antonydellavecchia:
A first pull request on serialization of attributes. Adds the keyword argument `with_attrs`. 

Includes updates to register_serialization_type that allows developers to add a vector of symbols
referring to attributes they would like to be serialized with the given type.

This is a first step for the serialization involved in https://github.com/oscar-system/Oscar.jl/pull/4008
and https://github.com/oscar-system/Oscar.jl/pull/3980


src/Serialization/main.jl
```diff
@@ -236,6 +246,12 @@ function load_object(s::DeserializerState, T::Type, params::Any, key::Union{Symb
   end
 end
 
+function load_attrs(s::DeserializerState, obj::T) where T
+  for attr in attrs_list(s, T)
+    haskey(s, attr) && set_attribute(obj, load_typed_object(s, attr))
```
lgoettgens:
```suggestion
    haskey(s, attr) && set_attribute!(obj, load_typed_object(s, attr))
```

experimental/AlgebraicStatistics/src/AlgebraicStatistics.jl
```diff
@@ -6,6 +6,7 @@ include("PhylogeneticModels.jl")
 include("PhylogeneticAuxiliary.jl")
 include("PhylogeneticParametrization.jl")
 
+
```
antonydellavecchia:
```suggestion
```

src/Serialization/main.jl
```diff
@@ -186,6 +190,12 @@ function save_type_params(s::SerializerState, obj::Any, key::Symbol)
   save_type_params(s, obj)
 end
 
+function save_attrs(s::SerializerState, obj::T) where T
+  for attr in attrs_list(s, T)
+    has_attribute(obj, attr) && save_typed_object(s, get_attribute(obj, attr), attr)
+  end
```
lgoettgens: Wouldn't it make sense to put attributes into their own dict a level deeper (like in https://github.com/oscar-system/Oscar.jl/pull/3980/files#diff-a782bbd162452110d68eee32ed25d1a4269c059bedb762b91fd168937768c8a3R78)? this makes it easier to distinguish attributes from the other serialized parts of the objects, and might reduce some name conflicts

src/Serialization/ToricGeometry.jl
```diff
@@ -1,13 +1,30 @@
...
+  attrs = attrs_list(s, T)
+
+  if !isnothing(attrs) && any([has_attribute(ntv, attr) for attr in attrs])
+    save_data_dict(s) do
+      save_attrs(s, ntv)
+      save_object(s, ntv.polymakeNTV, :pm_data)
+    end
+  else
+    save_object(s, ntv.polymakeNTV)
+  end
```
lgoettgens: I think it would be cleaner to always put the polymake thing behind the `pm_data` key, independent of the existence of attributes. This would reduce both `save_object` and `load_object` to a single code path.

But maybe its just easiest if (once the attribute serialization works) you split this up into a different PR regarding toric varieties and discuss it with the people involved there (aka not me)

src/Serialization/main.jl
```diff
@@ -310,6 +334,9 @@ function register_serialization_type(ex::Any, str::String, uses_id::Bool, uses_p
       # Types like ZZ, QQ, and ZZ/nZZ do not require ids since there is no syntactic
       # ambiguities in their encodings.
 
+      # add list of possible attributes to save for a given type to a global dict
+      register_attr_list($ex, $attrs)
```
lgoettgens:
```suggestion
      Oscar.register_attr_list($ex, $attrs)
```
to make it consistent with the rest of this function once https://github.com/oscar-system/Oscar.jl/pull/4021 is merged

src/Serialization/main.jl
```diff
@@ -186,6 +190,12 @@ function save_type_params(s::SerializerState, obj::Any, key::Symbol)
   save_type_params(s, obj)
 end
 
+function save_attrs(s::SerializerState, obj::T) where T
```
lgoettgens, line 196: This and `load_attrs` should be added to the `using Oscar:...` list here (https://github.com/oscar-system/Oscar.jl/pull/4021/files#diff-a00d43318a99f7f3073a8eeebdf4de11fe5c61405e45cd918241c100b82bed64R390) once https://github.com/oscar-system/Oscar.jl/pull/4021 is merged

src/Serialization/ToricGeometry.jl
antonydellavecchia: yes, it would make the code a lot simpler, if this is what we want to do though, it will require doing this for a lot of other types as well, maybe not all in the same pull request, but we also require upgrade scripts for each type. The reason I did it this way for now because it saves me writing the upgrade scripts for now.

lgoettgens: Ah, I understand. Of course, this is something that can be adapted step by step for different types

antonydellavecchia: yes, I choose a particularly annoying example to handle to start, but for any type that already uses a save_dict_data, we can just include the save_attrs function somewhere in the body and be done

src/Serialization/main.jl
antonydellavecchia: I see arguements for both sides, and I think it might be better to just have all properties at the same level.
I also dont think this would lead to name conflicts, I think there would be a deeper issue if there was a property and an attribute with the same name but different values

src/Serialization/ToricGeometry.jl
lgoettgens: But even in these cases I think another level in the dict would quite beneficial (see https://github.com/oscar-system/Oscar.jl/pull/4020#discussion_r1721489079)

src/Serialization/main.jl
lgoettgens: Fine with me. At least, if all types use this helper function, it is consistent between the different types.

test/Serialization/ToricGeometry.jl
```diff
@@ -2,9 +2,11 @@
   mktempdir() do path
     @testset "NormalToricVariety" begin
       pp = projective_space(NormalToricVariety, 2)
-      test_save_load_roundtrip(path, pp) do loaded
+      R = cox_ring(pp)
+      test_save_load_roundtrip(path, pp; with_attrs=true) do loaded
         @test rays(pp) == rays(loaded)
         @test ray_indices(maximal_cones(pp)) == ray_indices(maximal_cones(loaded))
+        @test R == get_attribute(loaded, :cox_ring)
```
lgoettgens: since `NormalToricVariety` serializes with `uses_id`, this test does not test anything here, as already `pp === loaded` (cf. our discussion in Durham)

test/Serialization/setup_tests.jl
```diff
@@ -7,11 +7,11 @@ using JSONSchema, Oscar.JSON
 if !isdefined(Main, :test_save_load_roundtrip)
   mrdi_schema = Schema(JSON.parsefile(joinpath(Oscar.oscardir, "data", "schema.json")))
 
-  function test_save_load_roundtrip(func, path, original::T; params=nothing) where {T}
+  function test_save_load_roundtrip(func, path, original::T; params=nothing, with_attrs=false) where {T}
```
lgoettgens:
```suggestion
  function test_save_load_roundtrip(func, path, original::T; params=nothing, with_attrs::Bool=false) where {T}
```

src/Serialization/serializers.jl
```diff
@@ -31,6 +31,11 @@ mutable struct SerializerState
   refs::Vector{UUID}
   io::IO
   key::Union{Symbol, Nothing}
+  type_attr_map::Dict{Type, Vector{Symbol}}
+end
+
+function attrs_list(s::SerializerState, T::Type)
+  return get(s.type_attr_map, T, nothing)
```
lgoettgens:
```suggestion
  return get(s.type_attr_map, T, Dict{Type, Vector{Symbol}}())
```
from a usability perspective (I am currently trying to make this PR work together with https://github.com/oscar-system/Oscar.jl/pull/3980) I find this desireable, because I don't have to take care of any possibility of nothings myself (as e.g. in https://github.com/oscar-system/Oscar.jl/pull/4020/files#diff-45089e89c18427deb0ceef25fae6c753eb01fd978a80d507ca509e3caff7bf37R11)

src/Serialization/serializers.jl
```diff
@@ -31,6 +31,11 @@ mutable struct SerializerState
   refs::Vector{UUID}
   io::IO
   key::Union{Symbol, Nothing}
+  type_attr_map::Dict{Type, Vector{Symbol}}
+end
+
+function attrs_list(s::SerializerState, T::Type)
```
lgoettgens: This function does not work for parameterized types. I believe this is due to `register_serialization_type` only registering the attribute list to the base part of the type, but `attrs_list` is passed the concrete type and thus finds nothing.
For some concrete example, you can have a look at the branch https://github.com/oscar-system/Oscar.jl/tree/lg/serialize-Lie-attrs-test (merge of this branch with https://github.com/oscar-system/Oscar.jl/pull/3980 plus https://github.com/oscar-system/Oscar.jl/pull/4020/files#r1721805863 plus some minor changes to make them load together) and execute the following:
```julia-repl
julia> L = lie_algebra(QQ, :A, 2);

julia> is_abelian(L)
false

julia> save("tmp.mrdi", L; with_attrs=true)
```
`L` is of type `AbstractLieAlgebra{QQFieldElem}` which has serialization of the attribute `is_abelian` enabled. However, this attribute does not show up in `tmp.mrdi`.

test/Serialization/ToricGeometry.jl
antonydellavecchia: yes, I also realized this must be the case since the fix with the set_attribute -> set_attribute!

src/Serialization/serializers.jl
antonydellavecchia: I see the issue, maybe we can do something similar to encode_type, but, i'll need to think about this tomorrow

src/Serialization/serializers.jl
```diff
@@ -31,6 +31,11 @@ mutable struct SerializerState
   refs::Vector{UUID}
   io::IO
   key::Union{Symbol, Nothing}
+  type_attr_map::Function
+end
+
+function attrs_list(s::SerializerState, T::Type)
+  return get(s.type_attr_map, T, Dict{Type, Vector{Symbol}}())
```
lgoettgens:
```suggestion
  return get(s.type_attr_map, T, Symbol[])
```
this of course needs to be the correct default...

src/Serialization/serializers.jl
lgoettgens: I have found a IMO nicer solution in terms of `encode_type` (which I find nice because this is already used in the serialization process as a type identifier). Have a look in 781fa8e6ea7375d21e7894fae550ea84b38eb3bc (this needs some stupid fix for the Lie algebras to work that is included in the branch as well)

lgoettgens: this would be an alternative to your two commits from this morning

antonydellavecchia: yes, i also thought about this, but we can avoid strings, this may be the best alternative

```
julia> T = typeof(sl2)
AbstractLieAlgebra{QQFieldElem}

julia> Base.typename(T).wrapper
AbstractLieAlgebra
```

lgoettgens: that's an internal and undocumented implementation detail of julia Base. I am not convinced that it's a good idea to rely on that

benlorenz: I don't think that is going to disappear anytime soon:
https://github.com/search?q=typename+wrapper+language%3Ajulia&type=code
It is mostly missing a consensus on the name and a PR:
https://github.com/JuliaLang/julia/issues/35543

lgoettgens: But isn't this exactly what we are using `encode_type` currently for? I don't understand why we should use two different methods to do basically the same thing (convert a type to an identifier)

antonydellavecchia: ok, we agree, we'll use encode_type

src/Serialization/serializers.jl
```diff
@@ -212,26 +218,37 @@ end
 
 state(s::OscarSerializer) = s.state
 
-function serializer_open(io::IO, T::Type{<: OscarSerializer})
+function serializer_open(
+  io::IO,
+  T::Type{<: OscarSerializer},
+  type_attr_map::S) where S <: Union{Dict{Type, Vector{Symbol}}, Nothing}
```
lgoettgens:
```suggestion
  type_attr_map::S) where S <: Union{Dict{String, Vector{Symbol}}, Nothing}
```

src/Serialization/serializers.jl
```diff
@@ -212,26 +218,37 @@ end
...
   # some level of handling should be done here at a later date
-  return T(SerializerState(true, UUID[], io, nothing))
+  return T(SerializerState(true, UUID[], io, nothing, type_attr_map))
 end
 
-function deserializer_open(io::IO, T::Type{JSONSerializer})
+function deserializer_open(
+  io::IO,
+  T::Type{JSONSerializer},
+  type_attr_map::S) where S <:Union{Dict{Type, Vector{Symbol}}, Nothing}
```
lgoettgens:
```suggestion
  type_attr_map::S) where S <:Union{Dict{String, Vector{Symbol}}, Nothing}
```

src/Serialization/serializers.jl
```diff
@@ -212,26 +218,37 @@ end
...
-  return T(DeserializerState(obj, nothing, refs))
+  
+  return T(DeserializerState(obj, nothing, refs, type_attr_map))
 end
 
-function deserializer_open(io::IO, T::Type{IPCSerializer})
+function deserializer_open(
+  io::IO,
+  T::Type{IPCSerializer},
+  type_attr_map::S) where S <:Union{Dict{Type, Vector{Symbol}}, Nothing}
```
lgoettgens:
```suggestion
  type_attr_map::S) where S <:Union{Dict{String, Vector{Symbol}}, Nothing}
```

test/Serialization/ToricGeometry.jl
antonydellavecchia: i added a fix to actually test this now

src/Serialization/main.jl
antonydellavecchia: I pushed a commit, that puts in the attributes into anested layer

src/Serialization/main.jl
```diff
@@ -502,20 +548,21 @@ function save(io::IO, obj::T; metadata::Union{MetaData, Nothing}=nothing,
...
   end
   Base.Filesystem.rename(temp_file, filename) # atomic "multi process safe"
   return nothing
 end
 
 """
-    load(io::IO; params::Any = nothing, type::Any = nothing)
-    load(filename::String; params::Any = nothing, type::Any = nothing)
+    load(io::IO; params::Any = nothing, type::Any = nothing, with_attrs::Bool=false)
+    load(filename::String; params::Any = nothing, type::Any = nothing, with_attrs::Bool=false)
```
lgoettgens:
```suggestion
    load(io::IO; params::Any = nothing, type::Any = nothing, with_attrs::Bool=true)
    load(filename::String; params::Any = nothing, type::Any = nothing, with_attrs::Bool=true)
```
change in default needs to be reflected here as well

src/Serialization/main.jl
```diff
@@ -658,7 +709,8 @@ function load(io::IO; params::Any = nothing, type::Any = nothing,
   end
 end
 
-function load(filename::String; params::Any = nothing, type::Any = nothing)
+function load(filename::String; params::Any = nothing,
+              type::Any = nothing, with_attrs::Bool=false)
```
lgoettgens:
```suggestion
              type::Any = nothing, with_attrs::Bool=true)
```
default here should be the same as for the `load(::IO)` dispatch

test/Serialization/setup_tests.jl
```diff
@@ -7,18 +7,18 @@ using JSONSchema, Oscar.JSON
 if !isdefined(Main, :test_save_load_roundtrip)
   mrdi_schema = Schema(JSON.parsefile(joinpath(Oscar.oscardir, "data", "schema.json")))
 
-  function test_save_load_roundtrip(func, path, original::T; params=nothing) where {T}
+  function test_save_load_roundtrip(func, path, original::T; params=nothing, with_attrs::Bool=false) where {T}
```
lgoettgens: maybe rename the kwarg `with_attrs` to something like `save_with_attrs` to underline that this is only used for saving but not for loading?

src/Serialization/serializers.jl
```diff
@@ -31,6 +31,11 @@ mutable struct SerializerState
   refs::Vector{UUID}
   io::IO
   key::Union{Symbol, Nothing}
+  type_attr_map::Dict{String, Vector{Symbol}}
+end
+
+function attrs_list(s::Union{SerializerState, DeserializerState}, T::Type)
+  return get(s.type_attr_map, encode_type(T), Symbol[])
 end
```
lgoettgens: this needs to be moved behind the DeserializerState struct definition

lgoettgens:
I'll try to get this branch now working with the Lie algebras to see if that is indeed what I imagined it to be

test/Serialization/setup_tests.jl
antonydellavecchia: well I just changed it so default behaviour is to load all attributes, so if you save with attributes it will load them all.
If you don't save any then load with behave as usual

lgoettgens: ok, true. didn't think of that. thanks for the explanation

src/Serialization/serializers.jl
```diff
@@ -212,26 +214,43 @@ end
...
+  type_attr_map::S) where S <:Union{Dict{String, Vector{Symbol}}, Nothing}
   # Using a JSON3.Object from JSON3 version 1.13.2 causes
   # @everywhere using Oscar
   # to hang. So we use a Dict here for now.
 
   obj = JSON.parse(io, dicttype=Dict{Symbol, Any})
-  return T(DeserializerState(obj, nothing, nothing))
+  return T(DeserializerState(obj, nothing, nothing, type_attr_map))
 end
+state_types = Union{SerializerState, DeserializerState}
```
lgoettgens:
```suggestion

const state_types = Union{SerializerState, DeserializerState}
```
the `const` is important for type inference

src/Serialization/main.jl
```diff
@@ -239,6 +251,14 @@ function load_object(s::DeserializerState, T::Type, params::Any, key::Union{Symb
   end
 end
 
+function load_attrs(s::DeserializerState, obj::T) where T
+  load_node(s, :attrs) do _
```
lgoettgens:
```suggestion
  haskey(s, :attrs) && load_node(s, :attrs) do _
```
this makes it possible to load old files that have no `:attrs` saved

src/Serialization/main.jl
```diff
@@ -189,6 +193,14 @@ function save_type_params(s::SerializerState, obj::Any, key::Symbol)
   save_type_params(s, obj)
 end
 
+function save_attrs(s::SerializerState, obj::T) where T
+  save_data_dict(s, :attrs) do
+    for attr in attrs_list(s, T)
+      has_attribute(obj, attr) && save_typed_object(s, get_attribute(obj, attr), attr)
+    end
+  end
```
lgoettgens: this, currently, always saves a dict, which may often be empty. In particular, this also gets saved when passing `with_attrs=false`. To circumvent this, you could add something like the following:
```suggestion
  if any(attr -> has_attribute(obj, attr), attrs_list(s, T))
    save_data_dict(s, :attrs) do
      for attr in attrs_list(s, T)
        has_attribute(obj, attr) && save_typed_object(s, get_attribute(obj, attr), attr)
      end
    end
  end
```

src/Serialization/ToricGeometry.jl
```diff
@@ -1,13 +1,30 @@
...
 
-function save_object(s::SerializerState, ntv::NormalToricVarietyType)
-  save_object(s, ntv.polymakeNTV)
+@register_serialization_type NormalToricVariety uses_id [:cox_ring]
+
+
+function save_object(s::SerializerState, ntv::T) where T <: NormalToricVarietyType
+  attrs = attrs_list(s, T)
+
+  if !isnothing(attrs) && any([has_attribute(ntv, attr) for attr in attrs])
```
lgoettgens: `attrs` will never be `nothing`, I think you need `isempty` for what you had here earlier.

you could, in theory, just save NTV from now on always as a dicts with `pm_data` and attrs, and only handle the old format in loading. this would make the saving already simpler, and it would not double the check for the presence of any attribute with https://github.com/oscar-system/Oscar.jl/pull/4020/files#r1723508786

antonydellavecchia: yes, I think the check here instead makes sense, this would then altert he toric variety files and this will introduce a breaking change, i'll need to write an upgrade script + probably upgrade martin's artifacts for f theory.

lgoettgens: cf https://github.com/oscar-system/Oscar.jl/pull/4020#discussion_r1721493505
tldr: I am not involved in NTV, so feel free to ignore my comments about that and instead talk to the people having a say there :)

src/Serialization/main.jl
antonydellavecchia: ill add the check here to not block lie algebra serialization, but also keep the check in toric varieties for now, and see if other similar examples popup and adjust them all at the same time.

lgoettgens: Great, thanks!

lgoettgens (approved):

lgoettgens:
@antonydellavecchia are you happy with having this merged once CI succeeds?

antonydellavecchia:
> @antonydellavecchia are you happy with having this merged once CI succeeds?

I would @benlorenz to have a quick look before

src/Serialization/main.jl
```diff
@@ -435,11 +477,14 @@ include("Upgrades/main.jl")
 # Interacting with IO streams and files
 
 """
-    save(io::IO, obj::Any; metadata::MetaData=nothing)
-    save(filename::String, obj::Any, metadata::MetaData=nothing)
+    save(io::IO, obj::Any; metadata::MetaData=nothing, with_attrs::Bool=false)
+    save(filename::String, obj::Any, metadata::MetaData=nothing, with_attrs::Bool=false)
```
lgoettgens: This now longer matches the changed default

antonydellavecchia: thanks!

benlorenz (approved):

PR #4022, opened 2024-08-19
change the documentation of `is_conjugate`, `is_conjugate_with_data`

### LINKED ISSUE #4017: Spell out what `^z` is in the documentation
I am writing a function which returns some "conjugating element" in some specific context, so I wanted to be a law abiding citizen and match our convention. I had hoped that `is_conjugate` would reveal what the convention is, but
```
  is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)

  Return whether x and y are conjugate elements in G, i.e., there is an element z in G such that
  x^z equals y. To also return the element z, use is_conjugate_with_data.

  ─────────────────────────────────────────────────────────────────────────────────────────────────

  is_conjugate(G::GAPGroup, H::GAPGroup, K::GAPGroup)

  Return whether H and K are conjugate subgroups in G, i.e., whether there exists an element z in G
  such that H^z equals K. To also return the element z use is_conjugate_with_data.
```
both do not tell me what `^z` is. I know that I can find the answer by meditating on left respectively right actions, but maybe we could just add it to the docstring?

I am happy to add it myself, once we decided that this is not a bad idea.

CC: @ThomasBreuer @fingolfin

ThomasBreuer:
Resolves #4017:
Write `z^-1*x*z` instead of `x^z`, etc.

(And fix cross-references between `is_conjugate` and `is_conjugate_with_data`.)

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -864,8 +864,9 @@ end
     is_conjugate(G::GAPGroup, x::GAPGroupElem, y::GAPGroupElem)
 
 Return whether `x` and `y` are conjugate elements in `G`,
-i.e., there is an element `z` in `G` such that `x^z` equals `y`.
-To also return the element `z`, use [`is_conjugate_with_data`](@ref).
+i.e., there is an element `z` in `G` such that `inv(z)*x*z` equals `y`.
+To also return the element `z`, use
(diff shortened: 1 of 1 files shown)


thofma (approved):

thofma:
Merci

PR #4023, opened 2024-08-19
Add Vinberg's algorithm

ElMerkel:
I added an implementation of Vinberg's algorithm for computing fundamental roots of Lorentzian lattices.

@simonbrandhorst 
@StevellM


docs/doc.main
```diff
@@ -83,7 +83,8 @@
         "Hecke/manual/quad_forms/genusherm.md",
         "Hecke/manual/quad_forms/integer_lattices.md",
         "Hecke/manual/quad_forms/Zgenera.md",
-        "Hecke/manual/quad_forms/discriminant_group.md"
+        "Hecke/manual/quad_forms/discriminant_group.md",
+        "LinearAlgebra/vinberg.md"
```
fingolfin:
```suggestion
        "LinearAlgebra/vinberg.md",
```

docs/oscar_references.bib
```diff
@@ -2170,6 +2177,17 @@ @Misc{VE22
   url           = {https://gap-packages.github.io/sglppow/}
 }
 
+@InCollection{Vin75,
+  author 	= {Vinberg, E. B.},
+  title 	= {Some arithmetical discrete groups in {L}obacevskii spaces},
+  booktitle 	= {Discrete subgroups of {L}ie groups and applications to moduli ({I}nternat. {C}olloq., {B}ombay, 1973)},
```
fingolfin:
```suggestion
  title 	= {Some arithmetical discrete groups in Lobacevskii spaces},
  booktitle 	= {Discrete subgroups of Lie groups and applications to moduli (Internat. Colloq., Bombay, 1973)},
```

docs/oscar_references.bib
```diff
@@ -2170,6 +2177,17 @@ @Misc{VE22
...
+  series 	= {Tata Inst. Fundam. Res. Stud. Math.},
+  volume 	= {No. 7},
+  pages 	= {323--348},
+  publisher 	= {Published for the Tata Institute of Fundamental Research, Bombay by Oxford University Press, Bombay},
```
fingolfin: Hmm, did you run `bibtool` [as described here](https://docs.oscar-system.org/stable/DeveloperDocumentation/documentation/#Updating-the-bibliography)?

docs/oscar_references.bib
```diff
@@ -2161,6 +2161,13 @@ @Article{Tay87
   primaryclass  = {math.GR}
 }
 
+@PhDThesis{Tur17,
+  author    	= {Turkalj, I.},
+  title		= {Reflective Lorentzian lattices of signature (5, 1)},
+  year 		= {2017},
+  school  	= {TU Dortmund}
```
fingolfin: There is also a paper, could that be cited instead? See https://doi.org/10.1016/j.jalgebra.2018.06.013

If not, a URL or better a DOI to the thesis would be great

docs/oscar_references.bib
fingolfin: Also, if there is a DOI available, please add it as well.

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
+```@meta
+CurrentModule = Oscar
+DocTestSetup = Oscar.doctestsetup()
+```
+
+# Vinberg's algorithm
+
+ A Lorentzian lattice $L$ is an integral $\Z$ - lattice in hyperbolic space.
```
fingolfin:
```suggestion
 A Lorentzian lattice $L$ is an integral $\Z$-lattice in hyperbolic space.
```

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ $L$ is called reflective if the set of fundamental roots $\~{R}(L)$ is finite.
+
+ See for example [Tur17](@cite) for the theory of Arithmetic Reflection Groups and Reflective Lorentzian Lattices.
+
+## Description 
+ The algorithm constructs a fundamental polyhedron $P$ for a Lorentzian lattice $L$ by computing its fundamental roots $r$.
+
+ Choose $v_0$ with $v_0^2 > 0$ as a point that $P$ should contain.
```
fingolfin: Maybe like this?
```suggestion
 Choose $v_0$ in $L$ with $v_0^2 > 0$ as a point that $P$ should contain.
```

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+
+ Let $Q$ be the corresponding gram matrix of $L$. A fundamental root $r$ satisfies
```
fingolfin:
```suggestion
 Let $Q$ be the corresponding Gram matrix of $L$. A fundamental root $r$ satisfies
```
Also, corresponding to what? Normally I'd expect "corresponding to a certain basis", but I see no basis stated here?

fingolfin: Also, is that supposed to be the definition of "fundamental root"? Or a list of a-posterior properties? That should be made clear (e.g. you could mark words being defined by using italics or bold font)

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ - the vector $r$ is primitive
+ - reflection by $r$ preserves the lattice, i.e. $\frac{2}{r^2}*r*Q$ is an integer matrix.
```
fingolfin: This bullet point ends with a period, the others don't. perhaps end all of them with a comma or semicolon, except for the last one (which should end with a period)?

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ - the pair $(r, v_0)$ is positive oriented, i.e. $(r, v_0) > 0$
+ - the product $(r, \~{r}) \geq \ 0$ for all roots $\~{r}$ already found
+ This implies that $r^2$ divides $2*i$ for $i$ being the level of $Q$, i.e. the last invariant of the smith normal form of $Q$. 
+
+ $P$ can be constructed by solving $(r, v_0) = n$ and $r^2 = k$ by increasing order of the value $\frac{n^2}{k}$ and $r$ satisfying the above conditions.
+
+ Perhaps $v_0$ lies on one or on several roots. Then $P$ is cannot be uniquely determined.
```
fingolfin: Grammar:
```suggestion
 Perhaps $v_0$ lies on one or on several roots. Then $P$ is not uniquely determined.
```
But I would also say this sentence is overall a bit awkward. How about this:
```suggestion
 If $v_0$ lies on multiple roots then $P$ is not uniquely determined.
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,295 @@
+###################################################################
```
fingolfin, line 1: I find it confusing that the source code for this is in `NumberTheory` but the documentation in `LinearAlgebra`. I would rather keep those in sync. If this belongs to LinearAlgebra, then let's add a `src/LinearAlgebra/` dir?

Then again, should this even go directly into `src`, or perhaps first into `experimental`? There are many things still a bit rough and unclear to me, e.g. the use of matrices to represent lattices.

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ In that case we need a direction vector $v_1$ that satisfies
+ - the pair $(v_0, v_1)$ lies orthogonal on each other, i.e. $(v_0, v_1) = 0$
+ - for all possible roots $\~{v}$ with $(v_0, \~{v}) = 0$ it holds $(\~{v}, v_1) \neq 0$ 
+
+ With $v_0$ and $v_1$ fixed $P$ can be uniquely determined for any choice of root lengths and maximal distance $(v_0, r)$.
+ We choose the first roots $r$ by increasing order of the value $\frac{(\~{r}, v_1)}{r^2}$ for all possible roots $\~{v}$ with $(v_0, \~{v}) = 0$.
+ For any other root length we continue as stated above.
+ 
+ For proofs of the statements above and further explanations see [Vin75](@cite)
```
fingolfin:
```suggestion
 For proofs of the statements above and further explanations see [Vin75](@cite).
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,295 @@
...
+  v1 = transpose(direction_vector)
+  @req (v0*Q*v1)[1, 1] == 0 "The direction vector has to be orthogonal on v0"
+  @req all((v*Q*v1)[1, 1] != 0 for v in possible_vec) "The direction vector cannot be orthogonal on one of the possible roots"
+  return v1
+end
+
+@doc raw"""
+    vinberg_algorithm(Q::ZZMatrix, upper_bound::ZZRingElem; v0::ZZMatrix, root_lengths::Vector{ZZRingElem}, direction_vector::ZZMatrix) -> Vector{ZZMatrix}
+
+Return the roots r of a given hyperbolic reflection lattice Q with squared length contained in root_lengths and
```
fingolfin: The keywords such as `r` should be marked as code, i.e.:
```suggestion
Return the roots `r` of a given hyperbolic reflection lattice `Q` with squared length contained in `root_lengths` and 
```
and so on in the lins below.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,295 @@
...
+"""
+function _check_direction_vector(Q::ZZMatrix, v0::ZZMatrix, possible_vec::Vector{ZZMatrix}, direction_vector::ZZMatrix)
+  v1 = transpose(direction_vector)
+  @req (v0*Q*v1)[1, 1] == 0 "The direction vector has to be orthogonal on v0"
+  @req all((v*Q*v1)[1, 1] != 0 for v in possible_vec) "The direction vector cannot be orthogonal on one of the possible roots"
+  return v1
+end
+
+@doc raw"""
+    vinberg_algorithm(Q::ZZMatrix, upper_bound::ZZRingElem; v0::ZZMatrix, root_lengths::Vector{ZZRingElem}, direction_vector::ZZMatrix) -> Vector{ZZMatrix}
```
fingolfin: `Q` is a matrix, not a lattice, though the text below refers to it as "a lattice". This always opens up ambiguities as to how to interpret that matrix. I thought we had types specifically to represent lattices, such as `ZZLat`, why not use that?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,295 @@
...
+        signal += 1
+        continue
+      end
+      v0 = matrix(ZZ, 1, l, v0_)
+      if (v0*Q*transpose(v0))[1, 1] > 0
+        return v0
+      end
+      signal += 1
+    end
+    @req signal < 100000000 "Choose another Q"
```
fingolfin: No need to duplicate the bound here; if we get here, then something went wrong
```suggestion
    error("Choose another Q")
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,295 @@
...
+      if !_is_primitive(v0_)
+        signal += 1
+        continue
+      end
+      v0 = matrix(ZZ, 1, l, v0_)
+      if (v0*Q*transpose(v0))[1, 1] > 0
+        return v0
+      end
+      signal += 1
+    end
```
fingolfin:
```suggestion
    for signal in 1:100000000
      v0_ = rand(-30:30, l)
      if !_is_primitive(v0_)
        continue
      end
      v0 = matrix(ZZ, 1, l, v0_)
      if (v0*Q*transpose(v0))[1, 1] > 0
        return v0
      end
    end
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,295 @@
+###################################################################
+# Vinberg's algorithm
+###################################################################
+
+@doc raw"""
+    _check_v0(Q::ZZMatrix, v0::ZZMatrix) -> ZZMatrix
+
+Checks if the inner product of the given v0 is positive or generates such a v0 if none is given.
```
fingolfin:
```suggestion
Checks if the inner product of the given `v0` is positive or generates such a `v0` if none is given.
```
Similar in all other docstrings.

fieker:
Needs sorting and possibly moved to experimental? @simonbrandhorst any idea? It's your baby....(indirectly)

simonbrandhorst:
@fieker Yes, I commissioned this and will take care of this PR.
@ElMerkel Thanks for the very nice work! More comments will follow later on.

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
+```@meta
+CurrentModule = Oscar
+DocTestSetup = Oscar.doctestsetup()
+```
+
+# Vinberg's algorithm
+
+ A Lorentzian lattice $L$ is an integral $\Z$-lattice in hyperbolic space.
```
simonbrandhorst:
```suggestion
 A Lorentzian lattice $L$ is an integral $\Z$-lattice of signature $(s_+, s_-)$ with $s_+=1$ and $s_->0$. 
```

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ The algorithm constructs a fundamental polyhedron $P$ for a Lorentzian lattice $L$ by computing its fundamental roots $r$.
+
+ Choose $v_0$ in $L$ with $v_0^2 > 0$ as a point that $P$ should contain.
+
+ Let $Q$ be the corresponding gram matrix of $L$. A fundamental root $r$ satisfies
+ - the vector $r$ is primitive
+ - reflection by $r$ preserves the lattice, i.e. $\frac{2}{r^2}*r*Q$ is an integer matrix.
+ - the pair $(r, v_0)$ is positive oriented, i.e. $(r, v_0) > 0$
+ - the product $(r, \~{r}) \geq \ 0$ for all roots $\~{r}$ already found
+ This implies that $r^2$ divides $2*i$ for $i$ being the level of $Q$, i.e. the last invariant of the smith normal form of $Q$.
```
simonbrandhorst:
```suggestion
 This implies that $r^2$ divides $2*i$ for $i$ being the level of $Q$, i.e. the last invariant of the Smith normal form of $Q$. 
```

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ Let $Q$ be the corresponding gram matrix of $L$. A fundamental root $r$ satisfies
+ - the vector $r$ is primitive
+ - reflection by $r$ preserves the lattice, i.e. $\frac{2}{r^2}*r*Q$ is an integer matrix.
+ - the pair $(r, v_0)$ is positive oriented, i.e. $(r, v_0) > 0$
+ - the product $(r, \~{r}) \geq \ 0$ for all roots $\~{r}$ already found
+ This implies that $r^2$ divides $2*i$ for $i$ being the level of $Q$, i.e. the last invariant of the smith normal form of $Q$. 
+
+ $P$ can be constructed by solving $(r, v_0) = n$ and $r^2 = k$ by increasing order of the value $\frac{n^2}{k}$ and $r$ satisfying the above conditions.
+
+ Perhaps $v_0$ lies on one or on several roots. Then $P$ is cannot be uniquely determined.
```
simonbrandhorst:
```suggestion
 If $v_0$ lies a root hyperplane, then $P$ is not uniquely determined.
```

docs/src/LinearAlgebra/vinberg.md
```diff
@@ -0,0 +1,53 @@
...
+ In that case we need a direction vector $v_1$ that satisfies
+ - the pair $(v_0, v_1)$ lies orthogonal on each other, i.e. $(v_0, v_1) = 0$
```
simonbrandhorst:
```suggestion
 - $v_1$ is orthogonal to $v_0$ ```

File: docs/src/LinearAlgebra/vinberg.md [hunk 21]
```diff
@@ -0,0 +1,53 @@
...
+ - for all possible roots $\~{v}$ with $(v_0, \~{v}) = 0$ it holds $(\~{v}, v_1) \neq 0$ 
+
+ With $v_0$ and $v_1$ fixed $P$ can be uniquely determined for any choice of root lengths and maximal distance $(v_0, r)$.
```
simonbrandhorst:
```suggestion
 With $v_0$ and $v_1$ fixed $P$ is uniquely determined for any choice of root lengths and maximal distance $(v_0, r)$.
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    l = nrows(Q)
+    for signal in 1:100000000
+      v0_ = rand(-30:30, l)
+      if !_is_primitive(v0_)
+        continue
+      end
+      v0 = matrix(ZZ, 1, l, v0_)
+      if (v0*Q*transpose(v0))[1, 1] > 0
+        return v0
+      end
```
simonbrandhorst: Assuming that `Q` is of signature `(1, n)` a random search can have a small success rate.
A better way is do diagonalise the form over the rationals and take the generator of the positive square and rescaled to be primitive.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+Checks if the inner product of the given v0 is positive or generates such a v0 if none is given.
+"""
+function _check_v0(Q::ZZMatrix, v0::ZZMatrix)
+  if iszero(v0)
+    l = nrows(Q)
+    for signal in 1:100000000
+      v0_ = rand(-30:30, l)
+      if !_is_primitive(v0_)
+        continue
+      end
```
simonbrandhorst: you could divide by the gcd to make it primitive instead?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+Returns all possible tuples (n, l) with n a natural number and l contained in root_lengths,
+sorted by increase of the value n^2//l, stopping by upper_bound.
+"""
+function _distance_indices(upper_bound::ZZRingElem, root_lengths::Vector{ZZRingElem})
+
+  result = Tuple{Int,ZZRingElem}[]
+
+  for l in root_lengths
+    x0 = sqrt(Float64(abs(l * upper_bound))) # consider n^2//l < upper_bound => n < sqrt(l*upper_bound)
+    x = Int(floor(x0))
```
simonbrandhorst:
```suggestion
    x = isqrt(x0*upper_bound)
```
avoids rounding errors

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+  for r in roots
+    if (r*Qv)[1, 1] < 0
+      return false
+    end
+  end
+
+  if (v*Q*transpose(v0))[1, 1] < 0
+    return false
+  end
+  return true
```
simonbrandhorst: could be shortened to 
```suggestion
  return (v*Q*transpose(v0))[1, 1] > 0
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+end
+
+@doc raw"""
+    _distance_0(Q::ZZMatrix, v0::ZZMatrix, root_lengths::Vector{ZZRingElem}) -> Vector{ZZMatrix}
+
+Return the roots which are orthogonal on v0.
```
simonbrandhorst:
```suggestion
Return the roots which are orthogonal to v0.
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+
+After gathering all vectors v which are orthogonal on v0 with length contained in root_lengths,
```
simonbrandhorst:
```suggestion
After gathering all vectors v which are orthogonal to v0 with length contained in root_lengths,
```

src/NumberTheory/vinberg.jl
simonbrandhorst: The function seems to do more than just returning the roots orthogonal to `v0`. 
So what exactly is the output?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+we execute the algorithm again by replacing v0 with the direction vector v1.
+"""
+function _distance_0(Q::ZZMatrix, v0::ZZMatrix, root_lengths::Vector{ZZRingElem}, direction_vector::ZZMatrix)
+  roots = ZZMatrix[]
+  possible_vec = ZZMatrix[]
+  # _short_vectors_gram is more efficient than short_vectors_affine
+  # Thus we have to make some preparations
+  bm = saturate(kernel(Q*transpose(v0); side=:left))
```
simonbrandhorst: I would have expected the kernel to be saturated already? (Since `QQ` is a `ZZMatrix`).

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    push!(possible_vec_sort, (v, n, k))
+  end
+  sort!(possible_vec_sort; by=x -> (x[2]^2) // abs(x[3]))
+
+  # We execute the algorithm with replacing v0 by v1
+  for (v, n) in possible_vec_sort
+    if n < 0
+      v = -v
+    end
+    if all((v*Q*transpose(w))[1, 1] >= 0 for w in roots)
```
simonbrandhorst:
```suggestion
    vQ = v*Q
    if all((vQ*transpose(w))[1, 1] >= 0 for w in roots)
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+  possible_vec_sort = Tuple{ZZMatrix,RationalUnion,RationalUnion}[]
+  for v in possible_vec
+    vQ = v * Q
+    n = (vQ*v1)[1, 1]
+    k = (vQ*transpose(v))[1, 1]
+    push!(possible_vec_sort, (v, n, k))
+  end
+  sort!(possible_vec_sort; by=x -> (x[2]^2) // abs(x[3]))
+
+  # We execute the algorithm with replacing v0 by v1
```
simonbrandhorst: which algorithm? What do you mean?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    if (v0*Q*v1)[1, 1] != 0
+      signal += 1
+      continue
+    end
+    @hassert :Vinberg 1 (transpose(v1)*Q*v1)[1, 1] < 0
+    if all((v*Q*v1)[1, 1] != 0 for v in possible_vec)
+      # To run the algorithm it suffices to require that v*Q*v1 != 0, but we get a random (right) solution. 
+      # This is the case because it is a random choice in which direction the fundamental cone 
+      # is built everytime we use the algorithm.
+      # To get always the same solution one could require that v*Q*v1 > 0, but this has very huge impact on the efficiency.
```
simonbrandhorst: nah, still would depend on the choices of signs of the `possible_vec` made.
If one wants non-deterministic outputs, one should supply `v1`.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    end
+    signal += 1
+  end
+  @req signal < 10000000 "Choose another v0" # break the algorithm if no v1 was found
+end
+
+@doc raw"""
+    _check_direction_vector(Q::ZZMatrix, v0::ZZMatrix, possible_vec::Vector{ZZMatrix}, direction_vector::ZZMatrix) -> ZZMatrix
+
+We check if the given direction vector holds the following requirements:
```
simonbrandhorst:
```suggestion
We check if the given direction vector satisfies the following requirements:
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    elseif l < 10
+      v1_ = rand(-25:25, l)
+    else
+      v1_ = rand(-20:20, l)  # for higher l it is not necessary/efficient to choose a big random range
+    end
+    v1 = matrix(QQ, l, 1, v1_)
+    if (v0*Q*v1)[1, 1] != 0
+      signal += 1
+      continue
+    end
```
simonbrandhorst: Why do you impose that v1 and v0 are orthogonal? This should not be necessary. 
Besides it is inefficient to search a vector v1 orthogonal to v0 randomly. This should be done by solving the `ZZ`-linear system of equations instead. 
```suggestion
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+  end
+  return roots
+end
+
+@doc raw"""
+    _generate_direction_vector(Q::ZZMatrix, v0::ZZMatrix, possible_vec::Vector{ZZMatrix}) -> ZZMatrix
+
+Since no direction vector v1 was given, we need to find one which satisfies the following conditions:
+- v0.v1 = 0
```
simonbrandhorst: I think this is not necessary, as just the orthogonal projection of v1 to $v_0^\perp$ should matter anyways.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    signal += 1
+  end
+  @req signal < 10000000 "Choose another v0" # break the algorithm if no v1 was found
+end
+
+@doc raw"""
+    _check_direction_vector(Q::ZZMatrix, v0::ZZMatrix, possible_vec::Vector{ZZMatrix}, direction_vector::ZZMatrix) -> ZZMatrix
+
+We check if the given direction vector holds the following requirements:
+- v0.v1 = 0
```
simonbrandhorst: Not necessary. See above.

src/NumberTheory/vinberg.jl
simonbrandhorst: Indeed, I agree with fingolfin please prepare a dispatch via `ZZLat`.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+  @req (v0*Q*v1)[1, 1] == 0 "The direction vector has to be orthogonal on v0"
+  @req all((v*Q*v1)[1, 1] != 0 for v in possible_vec) "The direction vector cannot be orthogonal on one of the possible roots"
+  return v1
+end
+
+@doc raw"""
+    vinberg_algorithm(Q::ZZMatrix, upper_bound::ZZRingElem; v0::ZZMatrix, root_lengths::Vector{ZZRingElem}, direction_vector::ZZMatrix) -> Vector{ZZMatrix}
+
+Return the roots r of a given hyperbolic reflection lattice Q with squared length contained in root_lengths and 
+by increasing order of the value ((r.v0)^2)//(r^2), stopping by upper_bound.
```
simonbrandhorst:
```suggestion
by increasing order of the value ``((r.v_0)^2)/(r^2)``, stopping by upper_bound. 
```
You can also use LaTeX in docstrings.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+If root_lengths is not defined it takes all possible values of r^2.
+If v0 lies on several possible roots and if there is no given direction vector it is a random choice which reflection chamber
+next to v0 will be computed.
+
+# Arguments
+- 'Q': symmetric ZZ matrix
```
simonbrandhorst:
```suggestion
- 'Q': symmetric ZZ matrix -- the gram matrix
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+
+Return the roots r of a given hyperbolic reflection lattice Q with squared length contained in root_lengths and 
+by increasing order of the value ((r.v0)^2)//(r^2), stopping by upper_bound.
+If root_lengths is not defined it takes all possible values of r^2.
+If v0 lies on several possible roots and if there is no given direction vector it is a random choice which reflection chamber
+next to v0 will be computed.
+
+# Arguments
+- 'Q': symmetric ZZ matrix 
+- 'upper_bound': the integer upper bound of the value ((r.v0)^2)//(r^2)
```
simonbrandhorst: do we have a reason to require that this bound is an integer?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+- 'v0': row vector with $v^2 > 0$
```
simonbrandhorst:
```suggestion
- 'v0': row vector with $v_0^2 > 0$
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+- 'root_lengths': the possible integer values of r^2
+- 'direction_vector': row vector v1 with v0.v1 = 0 and v.v1 != 0 for all possible roots v with v.v0 = 0
+
+The output is given in the ambient representation.
```
simonbrandhorst: If `Q` were a lattice you could state that the output were represented with respect to the basis of the ambient space. But I think for efficiency reasons we do not want this.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+  v0 = _check_v0(Q, v0)
+  if isempty(root_lengths)
+    real_root_lengths = _all_root_lengths(Q)
+  else
+    real_root_lengths = _check_root_lengths(Q, root_lengths) # sort out impossible lengths
+  end
+  iteration = _distance_indices(upper_bound, real_root_lengths) # find the right order to iterate through v.v0 and v^2
+  roots = _distance_0(Q, v0, real_root_lengths, direction_vector) # special case v.v0 = 0
+  for (n, k) in iteration # we  search for vectors which solve n = v.v0 and k = v^2
+    possible_Vec = Oscar.short_vectors_affine(Q, v0, QQ(n), k)
```
simonbrandhorst:
```suggestion
    possible_Vec = short_vectors_affine(Q, v0, QQ(n), k)
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,292 @@
...
+    for v in possible_Vec
+      if !_is_primitive(v)
+        continue
+      end
+      if _crystallographic_condition(Q, v)
+        if _check_coorientation(Q, roots, v, v0)
```
simonbrandhorst, line 271: The check that `v.v0>0` should be superflous since `v.v0=n>0`.

simonbrandhorst, line 271: This is the hot loop. So these two function will be called many times. Can you make them non-allocating?

src/NumberTheory/vinberg.jl
simonbrandhorst:
```suggestion
  for (n, k) in iteration # we  search for vectors which solve n = v.v0 and k = v^2
    @vprintln :Vinberg 1 "computing roots of squared length v^2=$(k) and v.v0 = $(n)" 
    possible_Vec = Oscar.short_vectors_affine(Q, v0, QQ(n), k)
```

simonbrandhorst:
The math looks good to me.

src/NumberTheory/vinberg.jl
ElMerkel: No. Actually it doesn't even have to be a rational number. I am going to change that again.

src/NumberTheory/vinberg.jl
ElMerkel: I am not sure about what you mean by that. Now I tried to explain more how the function works. But the output are the roots orthogonal to v_0. Do you mean that in this function the reflection chamber is picked which will be computed afterwards?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,291 @@
...
+    end
+  end
+
+  return (v*Q*transpose(v0))[1, 1] > 0
+end
+
+@doc raw"""
+    _distance_0(Q::ZZMatrix, v0::ZZMatrix, root_lengths::Vector{ZZRingElem}) -> Vector{ZZMatrix}
+
+Return the roots which are orthogonal to `v0`.
```
simonbrandhorst:
```suggestion
Return the fundamental roots which are orthogonal to `v0`.
```
A vector `r` in `L` is called a root if $(r.r) \neq 0$ and the reflection in `r` preserves `L`. 
It seems that you are looking for fundamental roots instead.

simonbrandhorst: Please state what the direction vector is and that it equals `v1` in your notation.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,291 @@
...
+    end
+  end
+  return roots
+end
+
+@doc raw"""
+    _generate_direction_vector(Q::ZZMatrix, v0::ZZMatrix, possible_vec::Vector{ZZMatrix}) -> ZZMatrix
+
+Since no direction vector `v1` was given, we need to find one which satisfies $(\~{v}, v_1) \neq 0$ 
+for all possible roots $\~{v}$ with $(v_0, \~{v}) = 0$
```
simonbrandhorst:
```suggestion
Return a direction vector `v1`  satisfying $(\~{v}, v_1) \neq 0$ 
for all `possible_vec` $\~{v}$ with $(v_0, \~{v}) = 0$
```
Your docstring reads more like a comment in the code. But a docstring cannot assume that kind of context. 
Just state what the input is and what the output is. 
Further `v0` is not needed as input. You could remove that and compute `l` from `Q`.

src/NumberTheory/vinberg.jl
simonbrandhorst: @ElMerkel You marked this as resolved, but I cannot find the method with `ZZLat` as argument. Did you upload your changes?

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,291 @@
...
+      push!(result, (n, l))
+    end
+  end
+
+  sort!(result; by=x -> (x[1]^2) // abs(x[2]))
+  return result
+end
+
+@doc raw"""
+    _is_primitive(v)
```
simonbrandhorst: The return value of `is_something` methods is usually a bool. 
Therefore a more appropriate name for this method would be something like `_rescale_primitive`.

src/NumberTheory/vinberg.jl
ElMerkel: Oh I did not know that you can see when I mark a conversation as resolved. I thought that I can use this for having a better overall view of the feedback. I am sorry, I can imagine that this led to confusion. The code should be uploaded now.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,322 @@
...
+Check if `v` is primitive, which is equivalent to checking whether the greatest common divisor of the entries of `v`
+is equal to 1. If `v` is not primitive, we divide `v` by its gcd.
+"""
+function _rescale_primitive(v::ZZMatrix)
+  gcd = reduce(gcd, v)
+  if isone(gcd)
+    return v
+  else
+    return (1 // gcd * v)
+  end
```
fingolfin: Minor comment: I would recommend not re-using a name like `gcd` for a local variable, perhaps call it `g`.

More importantlu, though, I don't think the `else` part works, or at least it doesn't when I tried it locally:
```
julia> v = matrix(ZZ, 6 * [1 2 ; 3 4])
[ 6   12]
[18   24]

julia> g = reduce(gcd, v)
6

julia> (1 // g * v)
ERROR: Cannot promote to common type
```
Even if it works, I would expect this to return a `QQMatrix`, which is probably not what you want.  However, this works:
```
julia> v / g
[1   2]
[3   4]
```

So overall perhaps use this?
```suggestion
  g = reduce(gcd, v)
  if isone(g)
    return v
  else
    return v / g
  end
```
or even
```suggestion
  g = reduce(gcd, v)
  return isone(g) ? v : v / g
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,322 @@
...
+  if iszero(v0)
+    Eigenvalues = eigenvalues(QQ, Q)
+    stopping_condition = 0
+    negative_eigenvalue = 0
+    for x in Eigenvalues
+      if x < 0
+        negative_eigenvalue = x
+        break
+      end
+      stopping_condition =+
```
fingolfin: Looks as if something is wrong here? Perhaps this was meant?
```suggestion
      stopping_condition += 1
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,322 @@
...
+    end
+    @req stopping_condition < length(Eigenvalues) "Q is not of signature (1, n)"
```
fingolfin: Actually, if I understand this code right, perhaps you could write it like this?
```suggestion
    ev = eigenvalues(QQ, Q)
    pos = findfirst(is_negative, ev)
    @req pos === nothing "Q is not of signature (1, n)"
    x = ev[pos]
```

docs/oscar_references.bib
ElMerkel: I could not find a DOI.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,310 @@
...
+function _check_v0(Q::ZZMatrix, v0::ZZMatrix)
+  if iszero(v0) # generate `v0` by finding the negative eigenvalue `neg_ev` of `Q`. We set `v0` as a primitive generator of the eigenspace of `neg_ev`
+    ev = eigenvalues(QQ, Q)
+    pos_neg_ev = findfirst(is_negative, ev) # find the position `pos_neg_ev` of the negative eigenvalue `neg_ev` 
+    @req pos_neg_ev === nothing "Q is not of signature (1, n)"
+    neg_ev = ev[pos_neg_ev]
+    return _rescale_primitive(ZZ.(eigenspace(Q, neg_ev)))
+  else
+    @req (v0*Q*transpose(v0))[1, 1] > 0 "v0 has non positive inner product"
+    @req isone(reduce(gcd, v)) "v0 is not primitive"
```
simonbrandhorst:
```suggestion
    @req isone(reduce(gcd, v0)) "v0 is not primitive"
```

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,310 @@
...
+Returns all possible tuples $(n, l)$ with `n` a natural number and `l` contained in `root_lengths`,
+sorted by increase of the value $\frac{n^2}{l}$, stopping by `upper_bound`.
+"""
+function _distance_indices(upper_bound, root_lengths::Vector{ZZRingElem})
+
+  result = Tuple{Int,ZZRingElem}[]
+
+  for l in root_lengths
+    x0 = sqrt(big(abs(l * upper_bound))) # consider for the `upper_bound` $u$ $\frac{n^2}{l} < u \implies n < \sqrt{l*u}$
+    x = isqrt(x0 * upper_bound)
```
simonbrandhorst, line 80:
```suggestion
    x = isqrt(abs(l * upper_bound))
```

docs/oscar_references.bib
simonbrandhorst: It is an old paper.

simonbrandhorst:
I did some local changes but could not figure out how to push them here. Hence I close this and open another PR.

PR #4030, opened 2024-08-20
[Lie algebras] misc bugfixes, root system in Lie alg interface, `dim_of_simple_module` implementation in julia

lgoettgens:
contains:

- misc bugfixes (missing `@attr`, typos)
- some progress towards a unified root system interface for all Lie algebra subtypes
- add a pure julia implementation of Weyl's dimension formula that is used if a root system is available (instead of converting to GAP). @benlorenz this is a first step towards https://github.com/oscar-system/Oscar.jl/pull/3913#issuecomment-2209374911

This contains all parts of https://github.com/oscar-system/Oscar.jl/pull/3980 that are not serialization related to allow for a quicker merge of the bugfixes also included there.


experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -162,10 +128,37 @@ end
...
+###############################################################################
+
+has_root_system(L::AbstractLieAlgebra) = isdefined(L, :root_system)
+
+function root_system(L::AbstractLieAlgebra)
+  @req has_root_system(L) "no root system known."
+  return L.root_system
+end
+
+function chevalley_basis(L::AbstractLieAlgebra)
```
felix-roehrich, line 148: It seems this function was only moved, why was the doc string deleted?

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -883,6 +883,10 @@ function is_dominant(w::WeightLatticeElem)
   return all(>=(0), coefficients(w))
 end
 
+function is_integral(w::WeightLatticeElem)
```
felix-roehrich: This does not make sense. WeightLatticeElems are always integral.

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
lgoettgens: The docstring is now attached to `chevalley_basis(::LieAlgebra)`, as it contained nothing specific for the Lie algebra subtype

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -906,6 +910,66 @@ function root_system(w::WeightLatticeElem)
...
 end
 
+###############################################################################
+# more functions
+
+function dot(r::RootSpaceElem, w::WeightLatticeElem)
+  @req root_system(r) === root_system(w) "parent root system mismatch"
+
+  return sum(
+    r[i] * cartan_symmetrizer(cartan_matrix(root_system(r)); check=false)[i] * w[i] for
```
felix-roehrich: We wanted to cache the symmetrizer, right? Either way it shouldn't be in the sum expression.

felix-roehrich (changes requested):

experimental/LieAlgebras/src/RootSystem.jl
lgoettgens: Oops, yeah. Forgot that :sweat_smile:

felix-roehrich (approved):

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -162,10 +128,37 @@ end
 #
 ###############################################################################
 
-function is_abelian(L::AbstractLieAlgebra)
+@attr function is_abelian(L::AbstractLieAlgebra)
   return all(e -> iszero(length(e)), _struct_consts(L))
```
fingolfin: Unrelated, just wondering: would this also work?
```suggestion
  return all(is_empty, _struct_consts(L))
```

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -437,6 +437,45 @@ Return `true` if `L` is solvable, i.e. the derived series of `L` terminates in $
...
+#   Root system getters
+#
+###############################################################################
+
+@doc raw"""
+    has_root_system(L::LieAlgebra) -> Bool
+
+Return whether a root system for `L` is known.
+
+This function should be implemented by subtypes that support root systems.
```
fingolfin: Another off-topic comment, but: perhaps we should agree to whom docstrings of exported (documented) functions are addressed. I'd say it should be users. Of course sometimes there is also a point in adding some remarks for *implementors*. But I think those then should be marked explicitly as such, e.g. via a callout (so the `!!!` syntax). That said, in this case I think one could just rephrase it slightly:
```suggestion
This is only implemented for Lie algebras that support root systems.
```
or so?

fingolfin (approved):

experimental/LieAlgebras/src/LieAlgebra.jl
lgoettgens: Thanks for the remark. I collected the information about which functions need to be implemented by subtypes that support root systems to the general list about subtype-implemented functions to the top. Furthermore, the docstrings are adapted to reflect what users need to know. (Note that I don't have to mention subtypes with no root system support there, as in these cases `has_root_system` should always return `false`, i.e. the `NotImplemented()` should never trigger.). Changes in 7ea927d5e34d33d0677ec409a5c5e01b1c1e9885

experimental/LieAlgebras/src/AbstractLieAlgebra.jl
lgoettgens: Yes, this should work, although I don't like that `is_empty` for `SRow` is using the abstract iterator fallback from julia base, and this name doesn't survey what I wanna test here. (`SRow`s iterate only their non-zero entries).

But the following should work and is indeed nicer than the current version.
```suggestion
  return all(iszero, _struct_consts(L))
```

PR #4031, opened 2024-08-20
Kashiwara crystals and Littelmann path model

felix-roehrich:
I am currently working on realizing the LS path model in OSCAR. The main goal is to replicate and extend the functionality currently provided in the QuaGroup package, while making it more convenient to use.

Since the path model is a Kashiwara crystal, I also want to provide an abstract type `AbstractCrystal`, since many things can be defined using only the abstract structure. Finding proper names for the functions proves rather difficult, since they have no explicit names in literature, so I would like to gather some suggestions.

Currently I have the following:
- $\tilde e_i(b), \tilde f_i(b)$ = `ealpha(b::AbstractCrystalElem, i::Int), falpha(b::AbstractCrystalElem, i::Int)`. The naming here was taken from the QuaGroup package, since I couldn't find anything better.
- $\epsilon_i(b), \varphi_i(b)$ = `eps(b::AbstractCrystalElem, i::Int), phi(b::AbstractCrystalElem, i::Int)`. Since `eps` already existed I hijacked it, not very happy with current names.
- $wt(b)$ = `weight(b::AbstractCrystalElem)`


experimental/LieAlgebras/src/LieAlgebras.jl
```diff
@@ -130,10 +133,14 @@ export coxeter_matrix
 export derived_algebra
 export dim_of_simple_module
 export dominant_character
+export dominant_path
+export ealpha, ealpha!
```
lgoettgens: I don't think these names are particularly readable. What about something like `e_alpha` or maybe `e_operator`?

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
+##### Kashiwara Crystal #####
+
+abstract type AbstractCrystal <: AbstractAlgebra.Set end
+
+abstract type AbstractCrystalElem <: AbstractAlgebra.SetElem end
+
+# AbstractCrystalElem required methods
+
+function Base.eps(b::AbstractCrystalElem)
```
lgoettgens: please don't overload the julia base function that gives you the machine precision

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+  error("not implemented")
+end
+
+function phi(b::AbstractCrystalElem)
+  error("not implemented")
+end
+
+@doc raw"""
+    Apply the root operator $\tilde e_i$ to `b` in place and return the result.
```
lgoettgens:
```suggestion
    ealpha!(b::AbstractCrystalElem, i::Int)

Apply the root operator $\tilde e_i$ to `b` in place and return the result.
```
This and the next few docstrings are indented weirdly

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+    Return the weight of `b`.
+"""
+function weight(b::AbstractCrystalElem)
+  error("not implemented")
+end
+
+# AbstractCrystal optional methods
+
+@doc raw"""
+    is_upper_normal(B::AbstractCrystal) -> bool
```
lgoettgens:
```suggestion
    is_upper_normal(B::AbstractCrystal) -> Bool
```
types in julia are uppercase (more similar places below)

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+Return the result of applying the root operator $\tilde e_i$ to `b` `n`-times.
+"""
+function ealpha(b::AbstractCrystalElem, i::Int, n::Int)
+  return ealpha!(deepcopy(b), i, n)
+end
+
+@doc raw"""
+    ealpha(b::AbstractCrystalElem, i::Vector{Int}, n::Vector{Int}) -> AbstractCrystalElem
+
+Return the result of applying the root operator $\tilde e_i$ to `b`.
```
lgoettgens: `n` is missing from the docstring

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+  s = zero(rdec)
+  b = deepcopy(p)
+  for i in 1:length(rdec)
+    s[i] = Base.eps(b, rdec[i]) # TODO: this will not work for non normal crystals
+    ealpha!(b, rdec[i], s[i])
+  end
+  return s
+end
+
+# LS Path Model
```
lgoettgens: maybe split up the file into 1. abstract crystal stuff and 2. path models?

lgoettgens (commented):
some random things I noticed while looking through this.

lgoettgens:
Maybe @fingolfin or @ulthiel have some opinion on how to name the abstract crystal functions (see initial comment by felix), or can ping some people that might want to take part in this?

experimental/LieAlgebras/docs/src/crystals.md
```diff
@@ -0,0 +1,19 @@
+# Crystals
+
+According to Kashiwara a crystal is a set $B$ together with the following maps
```
fingolfin, line 3: What are "the following maps" ?

fingolfin:
I have no idea at all what a "Kashiwara crystal" is, or an "abstract crystal". After some quick googling it doesn't seem to have any obvious connection to "usual" crystals such as studied in chemistry, and for which crystallographic groups are made, so I don't feel qualified to comment.

experimental/LieAlgebras/docs/src/crystals.md
```diff
@@ -0,0 +1,19 @@
...
+
+This is realised in OSCAR through the `AbstractCrystal` and `AbstractCrystalElem` interfaces and the following methods
```
fingolfin, line 5: This sentence should end with a period or colon.

experimental/LieAlgebras/docs/src/crystals.md
```diff
@@ -0,0 +1,19 @@
...
+
+```@docs
+ealpha(::AbstractCrystalElem, i::Int)
+falpha(::AbstractCrystalElem, i::Int)
+weight(::AbstractCrystalElem)
+```
+
+Additionally the following methods are provided
```
fingolfin, line 13: This sentence should end with a period or colon.

experimental/LieAlgebras/src/LieAlgebras.jl
```diff
@@ -83,8 +83,9 @@ import ..Oscar:
   ⊕,
   ⊗
 
-import Base: getindex, deepcopy_internal, hash, issubset, iszero, parent, zero
+import Base: getindex, deepcopy_internal, hash, issubset, iszero, max, parent, zero
 
+export AbstractCrystal, AbstractCrystalElem
```
fingolfin: off-topic question for @lgoettgens: these `export` statements all have to be duplicate. Would it make sense to put them into an `exports.jl` and then just `include` that twice, in both locations?

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+function weight(b::AbstractCrystalElem)
+  error("not implemented")
+end
+
+# AbstractCrystal optional methods
+
+@doc raw"""
+    is_upper_normal(B::AbstractCrystal) -> bool
+
+Return `true` if `B` is upper normal, and `false` if `B` is not upper normal or it is unknown.
```
fingolfin: Is this a standard term, that everyone working on these things would immediately understand? If not perhaps it could be briefly defined here?

Alternatively resp. in addition, a reference should be added to the bibliography that defines all these terms, and your `.md` file should point out this reference: "We follow the terminology and definitions in CITE".

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
+##### Kashiwara Crystal #####
+
+abstract type AbstractCrystal <: AbstractAlgebra.Set end
```
fingolfin: Please add a docstring to all new types that briefly explains what it is about (this could also contain a citation of the reference work I mention in another comment)

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+Return the result of applying the root operator $\tilde e_i$ to `b`.
+"""
+function ealpha(b::AbstractCrystalElem, i::Int)
+  return ealpha!(deepcopy(b), i)
+end
+
+@doc raw"""
+    ealpha(b::AbstractCrystalElem, i::Int, n::Int) -> AbstractCrystalElem
+
+Return the result of applying the root operator $\tilde e_i$ to `b` `n`-times.
```
fingolfin: I would recommend to avoid putting formulas directly adjacent. How about this (and of course similarly in other places that do this):
```suggestion
Return the result of `n`-times applying the root operator $\tilde e_i$ to `b`.
```
(There is probably a more elegant way to phrase this, perhaps our native speakers @JohnAAbbott or @mjrodgers have a suggestion)

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+  
+  for l in length(i):-1:1
+    falpha!(b, i[l], n[l])
+  end
+  return b
+end
+
+# other
+
+function adapted_string(p::AbstractCrystalElem, rdec::Vector{Int})
```
fingolfin: This is undocumented, and so looks like an internal helper. So why is it exported?

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+  for i in 1:length(rdec)
+    s[i] = Base.eps(b, rdec[i]) # TODO: this will not work for non normal crystals
+    ealpha!(b, rdec[i], s[i])
+  end
+  return s
+end
+
+# LS Path Model 
+
+struct LSPathModel
```
fingolfin: Please add a docstring

experimental/LieAlgebras/src/PathModel.jl
```diff
@@ -0,0 +1,617 @@
...
+end
+
+# LSPathModelElem
+
+struct LSPathSegment
+  t::QQFieldElem
+  w::WeylGroupElem
+end
+
+function Base.:(==)(s1::LSPathSegment, s2::LSPathSegment)
```
fingolfin, line 96: When defining a custom `==` method, make sure to also (two argument) `hash` method if there is even a tiny chance of anyone ever trying to use these as keys in a dictionary or apply `unique` to them etc.

Seems you have one for `LSPathModelElem`, but not for `LSPathSegment`; and the former ends up hashing elements of the latter type, so I guess you really need this.

thofma:
> I have no idea at all what a "Kashiwara crystal" is, or an "abstract crystal". After some quick googling it doesn't seem to have any obvious connection to "usual" crystals such as studied in chemistry, and for which crystallographic groups are made, so I don't feel qualified to comment.

Maybe this is a good reason to not call this `AbstractCrystal`.

Edit: There are also crystals in connection with crystalline cohomology.

experimental/LieAlgebras/src/WeylGroup.jl
```diff
@@ -490,6 +490,9 @@ end
 
 function Base.iterate(iter::ReducedExpressionIterator, word::Vector{UInt8})
   rk = rank(root_system(parent(iter.el)))
+  if isempty(word)
+    return nothing
+  end
```
fingolfin, line 499:
```suggestion
  isempty(word) && return nothing
  rk = rank(root_system(parent(iter.el)))
```

experimental/LieAlgebras/src/PathModel.jl
mjrodgers: I think the original is good, but without the hyphen: 
Return the result of applying the root operator to `b` `n` times.

Or alternately
Applies the root operator to `b` `n` times.
Applies the root operator `n` times to `b`.

Edit: Looks better maybe with the `b` and `n` separated.

experimental/LieAlgebras/src/LieAlgebras.jl
lgoettgens: That's a great idea. I just checked, and the two lists (already on current master), differ by some things (some names get exported just into the Oscar namespace but not exported from Oscar, and thus make it easier for other devs to use them and thus work against https://github.com/oscar-system/Oscar.jl/issues/3408 by accident). I'll prepare a PR for streamlining the two lists and putting them somewhere separate. This PR can then get rebased afterwards.

experimental/LieAlgebras/src/WeylGroup.jl
lgoettgens: This change is already included in https://github.com/oscar-system/Oscar.jl/pull/4030, so it can be dropped from this PR once https://github.com/oscar-system/Oscar.jl/pull/4030 is merged.

experimental/LieAlgebras/src/PathModel.jl
JohnAAbbott: Possibly:  result of `n`-fold application of the root operator to `b`.
For readability we do need to keep `b` and `n` separated (by more than just a space).

felix-roehrich:
> Maybe this is a good reason to not call this `AbstractCrystal`.
> 
> Edit: There are also crystals in connection with crystalline cohomology.

There is a difference between a Kashiwara crystal and an (abstract) crystal in the sense of Kashiwara, where the first one is a very specific crystal and the second came from abstracting the first. That's why naming it `KashiwaraCrystal` is bad, maybe `AbstractKashiwaraCrystal` is good enough to avoid confusion (even if `KashiwaraCrystal` is added at some point).

For the same reason naming `falpha` instead `kashiwara_f` is bad.

experimental/LieAlgebras/src/WeylGroup.jl
lgoettgens:
```suggestion
  rk = rank(root_system(parent(iter.el)))
```
#4030 is merged now

fingolfin:
@felix-roehrich what's the status of this PR? It was last updated 9 months ago and has a bunch of merge conflicts by now.

In principle this code looks reasonable to me, and since it is in experimental, I see no deep reason against merging it, at least if you are interested in "completing" it in the future (whatever that means). Was the naming issue ever settled? The PR seems to be using `KashiwaraCrystal` right now?

PR #4035, opened 2024-08-21
Updates to FreeAssAlgIdeal

### LINKED ISSUE #4043: Streamline `FreeAssAlgebra` and `Hecke.AssociativeAlgebraElem`
At the Friday meeting today it was discussed that while @Sequenzer works on associative algebras in Oscar, we can use the opportunity to streamline the already existing things in AbstractAlgebra and Hecke (both naming and interface).
From a first look, there seems to be:
- `AbstractAlgebra.FreeAssAlgebra`
- `AbstractAlgebra.FreeAssAlgElem`
- `Hecke.AbstractAssociativeAlgebraElem`
- `Hecke.AssociativeAlgebraElem`

I haven't checked what all of them offer, what the differences are, possibilities, etc. This issue exists just to have some place for discussions.

cc @Sequenzer @thofma @micjoswig @fieker

Sequenzer:
- Added `deg_bound` field to `FreeAssAlgIdeal` struct for speed improvement
- Modified constructor to initialize `deg_bound` and `gb`.
- Implemented `Base.show` for `FreeAssAlgIdeal`. 
- Added equality and membership functions for `FreeAssAlgIdeal`.
- Updated `ideal` function. 
- Enhanced `groebner_basis` functions with `interreduce` option.
- Added `is_groebner_basis` and `interreduce!` functions.
- Added some more test for integrated testing with `quantum_automorphism_groups`.
- Renamed `FreeAssAlg` -> `FreeAssociativeAlgebra`


src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -98,24 +110,26 @@ true
...
 
-  lp_I = Singular.Ideal(lpring, lpring.(gb))
-  return iszero(reduce(lpring(a), lp_I))
+function Base.in(a::FreeAssAlgElem, I::FreeAssAlgIdeal)
+  return ideal_membership(a, I)
 end
 
-function Base.in(a::FreeAssAlgElem, I::FreeAssAlgIdeal, deg_bound::Int)
-  return ideal_membership(a, I, deg_bound)
+function Base.in(I1, I2::FreeAssAlgIdeal)
```
joschmitt: Maybe this should rather be `is_subset` instead of `in`?

lgoettgens: I think this should be as similar as possible (w.r.t naming an interface) to ideals in MPolyRing. (not sure right now how they do this particular thing though)

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -17,18 +17,24 @@ Two-sided ideal of a free associative algebra with elements of type `T`.
...
     return r
   end
 end
 
-function AbstractAlgebra.expressify(a::FreeAssAlgIdeal; context = nothing)
-  return Expr(:call, :ideal, [expressify(g, context = context) for g in collect(a.gens)]...)
+Base.show(io::IO, a::FreeAssAlgIdeal) = print(io, "Ideal of ", base_ring(a), " with ", number_of_generators(a), " generators")
+
+function Base.:(==)(I2::FreeAssAlgIdeal, I1::FreeAssAlgIdeal)
+  return is_subset(I1,I2) && is_subset(I2,I1)
```
lgoettgens:
```suggestion
  check_base_rings(I1, I2)
  I1 === I2 && return true
  gens(I1) == gens(I2) && return true
  return is_subset(I1, I2) && is_subset(I2, I1)
```
some optimizations copied from the MPolyIdeal implementation at https://github.com/oscar-system/Oscar.jl/blob/c97f041686553748072fd508e54d0eee53f5d612/src/Rings/mpoly-ideals.jl#L1504-1509

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -98,24 +111,26 @@ true
...
-  lp_I = Singular.Ideal(lpring, lpring.(gb))
-  return iszero(reduce(lpring(a), lp_I))
+function Base.in(a::FreeAssAlgElem, I::FreeAssAlgIdeal)
+  return ideal_membership(a, I)
 end
 
-function Base.in(a::FreeAssAlgElem, I::FreeAssAlgIdeal, deg_bound::Int)
-  return ideal_membership(a, I, deg_bound)
+function is_subset(I1::FreeAssAlgIdeal, I2::FreeAssAlgIdeal)
+  return all(x -> in(x, I2), gens(I1))
```
lgoettgens:
```suggestion
  check_base_rings(I1, I2)
  return all(x -> in(x, I2), gens(I1))
```

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -37,16 +43,20 @@ Return the two-sided ideal of $R$ generated by $g$.
...
 end
 
 function ideal(g::Vector{<:FreeAssAlgElem})
   @assert length(g) > 0 "cannot infer base ring"
   algebra = parent(g[1]) 
   return ideal(algebra, g)
+end
 
+function ideal(g::IdealGens{T}, check_groebner_basis::Bool=false) where T <: FreeAssAlgElem
+  return FreeAssAlgIdeal(g, check_groebner_basis)
```
lgoettgens:
```suggestion
function ideal(g::IdealGens{T}; check_groebner_basis::Bool=false) where T <: FreeAssAlgElem
  return FreeAssAlgIdeal(g; check_groebner_basis)
```
for readability, this should better be a kwarg

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -17,18 +17,24 @@ Two-sided ideal of a free associative algebra with elements of type `T`.
 mutable struct FreeAssAlgIdeal{T} <: Ideal{T}
   gens::IdealGens{T}
   gb::IdealGens{T}
-
-  function FreeAssAlgIdeal(R::FreeAssAlgebra, g::Vector{T}) where T <: FreeAssAlgElem
+  deg_bound::Int
+  function FreeAssAlgIdeal(g::IdealGens{T}, check_groebner_basis::Bool=false) where T <: FreeAssAlgElem
```
lgoettgens:
```suggestion
  function FreeAssAlgIdeal(g::IdealGens{T}; check_groebner_basis::Bool=false) where T <: FreeAssAlgElem
```
see below

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -17,18 +17,24 @@ Two-sided ideal of a free associative algebra with elements of type `T`.
...
     r = new{T}()
-    r.gens = IdealGens(R, g)
+    r.gens = g
+    @req !check_groebner_basis || base_ring(base_ring(g)) isa Field "Groebner basis check requires a field base ring"
+    if check_groebner_basis && base_ring(base_ring(g)) isa Field && is_groebner_basis(g)
```
lgoettgens: I don't really understand what this new arg `check_groebner_basis` is supposed to do?
Could you write some explanation about that or maybe just put that directly into a docstring for `ideal(::IdeaGens; check_groebner_basis::Bool)`?

lgoettgens (commented):
some misc comments of things that confused me

src/Rings/FreeAssAlgIdeal.jl
```diff
@@ -17,18 +17,24 @@ Two-sided ideal of a free associative algebra with elements of type `T`.
...
+    end
     return r
   end
 end
 
-function AbstractAlgebra.expressify(a::FreeAssAlgIdeal; context = nothing)
-  return Expr(:call, :ideal, [expressify(g, context = context) for g in collect(a.gens)]...)
+Base.show(io::IO, a::FreeAssAlgIdeal) = print(io, "Ideal of ", base_ring(a), " with ", number_of_generators(a), " generators")
+
+function Base.:(==)(I2::FreeAssAlgIdeal, I1::FreeAssAlgIdeal)
```
lgoettgens: every type that implements `==` must also implement `hash(::T, ::UInt)` to allow for correctness with julia Base (see e.g. https://github.com/oscar-system/Oscar.jl/issues/2222).
As there is no normal form for these ideals, one needs to use the limited information one has to construct a really bad (but valid) hash function, e.g. as
```julia
function Base.hash(I::FreeAssAlgIdeal, h::UInt)
  return hash(base_ring(I), h)
end
```

src/Serialization/Algebras.jl
```diff
@@ -1,28 +1,28 @@
 ################################################################################
-# FreeAssAlgebra
+# FreeAssociativeAlgebra
 
 # Free associative algebra serialization
-@register_serialization_type FreeAssAlgebra uses_id
+@register_serialization_type FreeAssociativeAlgebra uses_id
```
lgoettgens, line 5: @antonydellavecchia Iirc this renaming here and the two further ones in lines 21 and 56 in this file need some mention in a serialization upgrade script. Do we already have some code snippet to put in there for such a simple type renaming or, if not, could you put together something for it?

Sequenzer: I believe @antonydellavecchia will write an update script...

antonydellavecchia: including something like this in src/Serialization/Upgrades
and naming the file 1.2.0.jl, I copied this script from 0.15.0.jl,
or wtv the next version of oscar will be I guess 1.2.0

```
push!(upgrade_scripts_set, UpgradeScript(
  v"1.2.0",
  function upgrade_1_2_0(s::UpgradeState, dict::Dict)

    upgraded_dict = dict

    renamings = Dict{String,String}([
      ("FreeAssAlgebra", "FreeAssociativeAlgebra"),
      ("FreeAssAlgElem", "FreeAssociativeAlgebraElem"),
      ("FreeAssAlgIdeal", "FreeAssociativeAlgebraIdeal"),
    ])

    function upgrade_type(d::String)
      return get(renamings, d, d)
    end
    function upgrade_type(d::Dict)
      upg_d = d
      upg_d[:name] = get(renamings, d[:name], d[:name])
      if d[:params] isa Dict && haskey(d[:params], :_type)
        upg_d[:params][:_type] = upgrade_type(d[:params][:_type])
      elseif d[:params] isa Vector
        upg_d[:params] = [upgrade_type(v) for v in d[:params]]
      end
      return upg_d
    end

    if haskey(dict, :_type)
      upgraded_dict[:_type] = upgrade_type(dict[:_type])
    end
        
    if haskey(dict, :data) && dict[:data] isa Dict
      upgraded_dict[:data] = upgrade_1_2_0(s, dict[:data])
    end

    if haskey(dict, :_refs)
      upgraded_refs = Dict()
      for (k, v) in dict[:_refs]
        upgraded_refs[k] = upgrade_1_2_0(s, v)
      end
      upgraded_dict[:_refs] = upgraded_refs
    end

    return upgraded_dict
  end
))
```

lgoettgens: Great! Thanks @antonydellavecchia. Some remarks:
- we are only in the 1.2.0 cycle, so all 1.3.0 should be changed to 1.2.0.
- near the bottom are still two old 0_15_0, should be 1_2_0 as well
- the renamings list is missing the ideal

@Sequenzer can you please incorporate this script with the changes lined out here? I will try it out once we have this in this PR.

Edit: I changed the above code comment and included the changes there directly

src/Rings/FreeAssAlgIdeal.jl
Sequenzer: Somehow I didn't see those comments

src/Rings/FreeAssAlgIdeal.jl
Sequenzer: It is supposed to check if the generator is already a Groebner basis and then add it to the .gb immediately. But in retrospect, I do not think it helps that much, especially if it is hidden behind a kwarg. I remove it.

src/Serialization/Algebras.jl
lgoettgens: I verified that this upgrade script works when trying to load some simple objects that were saved with Oscar 1.1.1. (For things that were saved with 1.2.0-DEV versions the upgrade script does not run, but this is not relevant for this PR)

lgoettgens:
@antonydellavecchia It seems that [`8990f92` (#4035)](https://github.com/oscar-system/Oscar.jl/pull/4035/commits/8990f925154bba28bf26ddd2c7a63b168a3cd1f1) broke CI

fingolfin:
I'll take a look at this PR this afternoon, and either merge it or request changes. If I haven't done it by the end of the day, feel free to just merge, I can always get changes in later as well.

lgoettgens (approved):
All comments have been addressed. And Max didn't do a review yesterday, so let's merge this for now.

PR #4036, opened 2024-08-21
LieAlgebras: Streamline exports in `Oscar.LieAlgebras`

lgoettgens:
as suggested by @fingolfin in https://github.com/oscar-system/Oscar.jl/pull/4031#discussion_r1724643328.


experimental/LieAlgebras/src/Combinatorics.jl
```diff
@@ -3,11 +3,11 @@
 # more efficient implementation.
 
 function combinations(n::Integer, k::Integer)
-  return sort(subsets(n, k))
+  return sort(AbstractAlgebra.combinations(n, k))
```
fingolfin, line 6: That's an unrelated change? (Not that I terribly mind, just am surprised)

lgoettgens: When fixing the calls to this method outside of Oscar.LieAlgebras I found this as an alternative. It does not make any speed difference in the limited places it is called anyway. So this change is more a note to me to see if I can get rid of Oscar.LieAlgebras.combinations in the near future.

experimental/LieAlgebras/src/LieAlgebras.jl
```diff
@@ -85,120 +85,18 @@ import ..Oscar:
...
-export LieSubalgebra
-export LinearLieAlgebra, LinearLieAlgebraElem
-export RootSpaceElem
-export RootSystem
-export WeightLatticeElem
-export WeylGroup, WeylGroupElem
-export WeylOrbitIterator
+include("exports.jl")
 
+# only export into Oscar, not from Oscar
```
fingolfin: I find this hard to understand. Perhaps you meant this?
```suggestion
# The following export statements export things *into* module Oscar;
# they are however for internal use and thus are not exported *from* Oscar.
```

Could we perhaps instead replace these by `import` statements, i.e. after module `LieAlgebras` is closed, do:
```
import LieAlgebras: _is_direct_sum
```
and so on? That way nothing would have to be explained. But maybe this doesn't work for some technical reason I am not aware of? (Also: of course this is not a required change for this PR, just a thought)

src/exports.jl
```diff
@@ -1333,6 +1333,7 @@ export ring_elem_type
 export ring_type
 export rising_factorial
 export root
+export roots
```
fingolfin, line 1336: again, I am not really objecting, just surprised as this seems to go beyond the PR description, so I am double checking to verify this is intended?

lgoettgens: roots is reexported from Hecke anyway. Oscar.LieAlgebras was kind of inconsistent about exporting it.
And it felt somehow wrong to export it from Oscar in the LieAlgebras experimental folder, although it is not introduced there (as everything else that is exported there) but imported from Oscar/Hecke/...
This felt cleaner than just removing the export from OSCAR entirely and rely on the reexport

experimental/LieAlgebras/src/LieAlgebras.jl
lgoettgens: There is a semantical difference (for users, not for the compiler) between exports and imports. This was e.g. the reason that I revamped the in/export between AA and AA.Generic in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1537. And there is tooling like https://github.com/ericphanson/ExplicitImports.jl that suggest to change imports, but they won't do anything about exports.
Thus, I would like to keep them as exports (at least for now), but I am happy to take your comment suggestion here :)

lgoettgens:
Bump @fingolfin

lgoettgens:
Bump @fingolfin . It would be great to get this in before https://github.com/oscar-system/Oscar.jl/pull/4053 to avoid merge conflicts.

fingolfin (approved):

fingolfin:
Sorry I didn't see your "bump" here -- I regularly get flooded with GitHub notifications. I recommend pinging me on Slack again, that has a much better signal-to-noise ratio

PR #4038, opened 2024-08-21
Fix #4018

### LINKED ISSUE #4018: `grassmann_pluecker_ideal` saved GB wrong
I'm very happy that OSCAR has a constructor for Pluecker ideals, but the output is not very usable.  For example, asking for its `dim` raises and error and asking for its `degree` returns something wrong:

```
julia> I = grassmann_pluecker_ideal(2,5)
Ideal generated by
  x[1]*x[8] - x[2]*x[6] + x[3]*x[5]
  x[1]*x[9] - x[2]*x[7] + x[4]*x[5]
  x[1]*x[10] - x[3]*x[7] + x[4]*x[6]
  x[2]*x[10] - x[3]*x[9] + x[4]*x[8]
  x[5]*x[10] - x[6]*x[9] + x[7]*x[8]

julia> dim(I)
ERROR: Not a Groebner basis
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] dimension(I::Singular.sideal{Singular.spoly{Singular.n_Q}})
   @ Singular ~/.julia/packages/Singular/A4riV/src/ideal/ideal.jl:105
 [3] __compute_dim__(I::MPolyIdeal{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
   @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-ideals.jl:1811
 [4] #379
   @ ~/.julia/packages/AbstractAlgebra/tABFQ/src/Attributes.jl:357 [inlined]
 [5] get!(default::Oscar.var"#379#380"{MPolyIdeal{MPolyDecRingElem{…}}}, h::Dict{Symbol, Any}, key::Symbol)
   @ Base ./dict.jl:479
 [6] get_attribute!
   @ ~/.julia/packages/AbstractAlgebra/tABFQ/src/Attributes.jl:230 [inlined]
 [7] dim(I::MPolyIdeal{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})
   @ Oscar ~/.julia/dev/Oscar/src/Rings/mpoly-ideals.jl:1806
 [8] top-level scope
   @ REPL[19]:1

julia> degree(I)
1

```

joschmitt:
(Sorry for the duplicate, #4037 was the wrong branch.)

This fixes #4018 for me.

~Apparently, one should not set `keep_ordering = true` in this case. Otherwise the ordering on the Singular side is set to what is nowadays called `internal_ordering`, so lex in this case.~
I don't understand what the point of `keep_ordering` is. Does somebody know?

### DIFF
--- a/src/Rings/mpoly-graded.jl
+++ b/src/Rings/mpoly-graded.jl
@@ -1502,8 +1502,8 @@ mutable struct HilbertData
     @req coefficient_ring(R) isa AbstractAlgebra.Field "The coefficient ring must be a field"
     @req all(is_homogeneous, gens(I)) "The generators of the ideal must be homogeneous"
 
-    G = groebner_assure(I)
-    cf = Singular.hilbert_series_data(G.S, W)
+    S = singular_groebner_generators(I, false, true)
+    cf = Singular.hilbert_series_data(S, W)
     return new(cf, W, I)
   end
   function HilbertData(B::IdealGens)
--- a/src/Rings/mpoly.jl
+++ b/src/Rings/mpoly.jl
@@ -383,7 +383,11 @@ function singular_generators(B::IdealGens, monorder::MonomialOrdering=default_or
   isdefined(B, :ord) && B.ord == monorder && monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord && return B.gens.S
   SR = singular_poly_ring(B.Ox, monorder)
   f = Singular.AlgebraHomomorphism(B.Sx, SR, gens(SR))
-  return Singular.map_ideal(f, B.gens.S)
+  S = Singular.map_ideal(f, B.gens.S)
+  if isdefined(B, :ord) && B.ord == monorder
+    S.isGB = B.isGB
+  end
+  return S
 end
 
 @doc raw"""
@@ -745,7 +749,7 @@ function singular_assure(I::IdealGens)
     I.gens.Sx = singular_poly_ring(I.Ox; keep_ordering = I.keep_ordering)
     I.gens.S = Singular.Ideal(I.Sx, elem_type(I.Sx)[I.Sx(x) for x = I.O])
   end
-  if I.isGB
+  if I.isGB && (!isdefined(I, :ord) || I.ord == monomial_ordering(I.gens.Ox, internal_ordering(I.gens.Sx)))
     I.gens.S.isGB = true
   end
 end
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -2112,7 +2112,7 @@ function grassmann_pluecker_ideal(ring::MPolyRing,
     return coeff_ring(numerator(c))
   end
   if !is_graded(ring)
-	 ring, _ = grade(ring)
+    ring, _ = grade(ring)
   end
   h = hom(base_ring(I), ring, coeffmap, gens(ring))
   converted_generators = elem_type(ring)[h(g) for g in groebner_basis(I; ordering = degrevlex(base_ring(I)))]
--- a/test/Rings/mpoly.jl
+++ b/test/Rings/mpoly.jl
@@ -300,9 +300,18 @@ end
 end
 
 @testset "Grassmann Plücker Relations" begin
-    R, x = graded_polynomial_ring(residue_ring(ZZ, 7)[1], "x" => (1:2, 1:3))
-    test_ideal =ideal([x[1, 2]*x[2, 2] + 6*x[2, 1]*x[1, 3] + x[1, 1]*x[2, 3]])
-    @test grassmann_pluecker_ideal(R, 2, 4) == test_ideal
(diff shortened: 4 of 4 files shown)


fingolfin:
Maybe @ederc or @jankoboehm know more?

joschmitt:
I realize now that my "fix" only works by coincidence, that is, because the ordering we want on the Singular side in this example happens to be the default (degrevlex).
I will try to produce a minimal example of what I mean tomorrow.

wdecker:
Tracing the bug here, I see a couple of things with idealGens, groebner_assure etc which I do not quite understand. I must discuss this with @ederc, @hannes14.
@joschmitt For example, making another change in the function `grassmann_pluecker_ideal`, the "problem disappears":

 return ideal(IdealGens(base,base.(gens(I)), o;
                   keep_ordering=false,
                   isReduced=false,
                   isGB=true))

joschmitt:
I did another fix now. I am more confident of this, but someone who knows about these `groebner_assure`, `singular_assure`, ... etc should have a close look.
What seems weird to me (and what caused the problems here) is that the ordering set in `G.ord` of an `IdealGens` object `G` may not agree with the ordering of the Singular ring in `G.gens.Sx`. So in particular `G.gens.O` may be a Gröbner basis with respect to `G.ord`, but the same polynomials in `G.gens.S` on the Singular side are "not a Gröbner basis" because `G.gens.Sx` has a different ordering.

src/Rings/mpoly-graded.jl
```diff
@@ -1502,8 +1502,8 @@ mutable struct HilbertData
     @req coefficient_ring(R) isa AbstractAlgebra.Field "The coefficient ring must be a field"
     @req all(is_homogeneous, gens(I)) "The generators of the ideal must be homogeneous"
 
-    G = groebner_assure(I)
-    cf = Singular.hilbert_series_data(G.S, W)
+    S = singular_groebner_generators(I, false, true)
+    cf = Singular.hilbert_series_data(S, W)
```
joschmitt, line 1506: I assume that we want `singular_groebner_generators` here and not just `groebner_assure`. The doc string of `groebner_assure` explains that one should not rely on the basis being defined on the Singular side.
I also assume that we need a global ordering here; otherwise one can change the `true` to `false`.

src/Rings/mpoly.jl
```diff
@@ -383,7 +383,11 @@ function singular_generators(B::IdealGens, monorder::MonomialOrdering=default_or
   isdefined(B, :ord) && B.ord == monorder && monomial_ordering(B.Ox, Singular.ordering(base_ring(B.S))) == B.ord && return B.gens.S
   SR = singular_poly_ring(B.Ox, monorder)
   f = Singular.AlgebraHomomorphism(B.Sx, SR, gens(SR))
-  return Singular.map_ideal(f, B.gens.S)
+  S = Singular.map_ideal(f, B.gens.S)
+  if isdefined(B, :ord) && B.ord == monorder
+    S.isGB = B.isGB
+  end
+  return S
```
joschmitt, line 390: This should fix the main problem in #4018: `singular_generators` would not carry the information that the polynomials form a Gröbner basis with respect to `monorder` over to the Singular side.

src/Rings/mpoly.jl
```diff
@@ -745,7 +749,7 @@ function singular_assure(I::IdealGens)
     I.gens.Sx = singular_poly_ring(I.Ox; keep_ordering = I.keep_ordering)
     I.gens.S = Singular.Ideal(I.Sx, elem_type(I.Sx)[I.Sx(x) for x = I.O])
   end
-  if I.isGB
+  if I.isGB && I.ord == monomial_ordering(I.gens.Ox, internal_ordering(I.gens.Sx))
```
joschmitt: This was wrong: the polynomial ring on the Singular side might have a monomial ordering different from `I.ord`.

fingolfin:
We discussed this during triage. The plan is to wait until after the OSCAr workshop in two weeks, during which hopefully the underlying issue can be resolved.

If this for some reason doesn't happen, we can discuss again if we want to merge this for the time being until a "proper" fix can be implemented.

ederc:
I think that the fixes @joschmitt provides in this PR are the correct ones. There is, as @wdecker pointed out, another problem with `groebner_assure`, but we can handle this in a different PR. So for me this PR is good to go.

ederc:
For the above mentioned `groebner_assure` fix see #4077.

ederc (approved):

joschmitt:
At last week's triage meeting, there was talk about an underlying issue and @fingolfin's summary above says that this pull request should wait, at least until the coding sprint next week. I don't want to restart the discussion; if somebody thinks this should be merged "as is", please do so. Otherwise I'm happy to talk about it next week.

fingolfin:
Since the "deeper" issue is fixed and @ederc thinks this PR is good to go, I'll merge it now.

PR #3979, opened 2024-07-28
[FTheoryTools] More on Chern classes and topological invariants

HereAround:
cc @apturner @emikelsons

PR #4044, opened 2024-08-26
[FTheoryTools] Added more checks for G4 fluxes

emikelsons:
- Added verticality check for G4 fluxes
- Changed zero_section method to now return a cohomology class in the toric ambient space. Old version of the method is now under zero_section_coordinates(this is outdated, see below)
- Added D3-tadpole cancellation check for G4 fluxes.


HereAround:
I have just merged https://github.com/oscar-system/Oscar.jl/pull/3979. It seems this leads to to merge conflicts (see below the CI test results). @emikelsons , could you please rebase to the current master.

HereAround:
I would think that the two test failures are unrelated to your changes, but we could see this more clearly once the tests are executed with your PR rebase to the current master.

experimental/FTheoryTools/src/G4Fluxes/properties.jl
```diff
@@ -42,7 +42,7 @@ true
 
   # Compute the cohomology class corresponding to the hypersurface equation
   cy = polynomial(cohomology_class(toric_divisor_class(ambient_space(m), degree(hypersurface_equation(m)))))
-
+
```
HereAround: Is this needed?

experimental/FTheoryTools/src/G4Fluxes/properties.jl
```diff
@@ -52,9 +52,67 @@ true
   # now execute elementary checks of the quantization condition
   for i in 1:length(c_ds)
     for j in i:length(c_ds)
-      numb = integrate(cohomology_class(ambient_space(m), twist_g4 * cy * c_ds[i] * c_ds[j]); check = false)
+      numb = integrate(cohomology_class(ambient_space(m), twist_g4 * c_ds[i] * c_ds[j] * cy); check = false)
```
HereAround, line 55: I suppose this an unrelated change? Is it critical?

experimental/FTheoryTools/src/G4Fluxes/properties.jl
```diff
@@ -52,9 +52,67 @@ true
...
+`true` and otherwise `false`
+
+```jldoctest
+julia> qsm_model = literature_model(arxiv_id = "1903.00009", model_parameters = Dict("k" => 4))
+Hypersurface model over a concrete base
+
+julia> divs = torusinvariant_prime_divisors(ambient_space(qsm_model));
+
+julia> e1 = cohomology_class(divs[35]);e2 = cohomology_class(divs[32]);e4 = cohomology_class(divs[34]);u = cohomology_class(divs[33]);v = cohomology_class(divs[30]);pb_Kbar = cohomology_class(sum([divs[k] for k in 1:29]));
+
```
HereAround: I suggest to split those commands over several lines for readability.

experimental/FTheoryTools/src/G4Fluxes/properties.jl
emikelsons: I don't think anything is actually there. At least for me it shows up like any line break.(Edit: Ok, I now see there are 2 additional spaces)

experimental/FTheoryTools/src/G4Fluxes/properties.jl
```diff
@@ -52,9 +52,67 @@ true
...
+Cohomology class on a normal toric variety given by 1//2*x5*x32 + x5*x33 + 3//2*x5*x34 + x5*x35 + 1//2*x5*x36 + 1//2*x6*x32 + x6*x33 + 3//2*x6*x34 + x6*x35 + 1//2*x6*x36 - x7*x12 + 1//2*x7*x32 + x7*x33 + 3//2*x7*x34 + x7*x35 + 1//2*x7*x36 - 4*x8*x23 - 2*x10*x23 + 2*x15*x26 + 1//2*x15*x32 + x15*x33 + 3//2*x15*x34 + x15*x35 + 1//2*x15*x36 + x16^2 + 1//2*x16*x32 + x16*x33 + 3//2*x16*x34 + x16*x35 + 1//2*x16*x36 + 11*x17*x24 + x17*x32 + 2*x17*x33 + 3*x17*x34 + 2*x17*x35 + x17*x36 + 5*x18^2 - 7//4*x18*x25 + x18*x32 + 2*x18*x33 + 3*x18*x34 + 2*x18*x35 + x18*x36 + 3//2*x19*x32 + 3*x19*x33 + 9//2*x19*x34 + 3*x19*x35 + 3//2*x19*x36 - 4*x20^2 - 44//3*x20*x21 + 7//2*x20*x25 + x20*x32 + 2*x20*x33 + 3*x20*x34 + 2*x20*x35 + x20*x36 - 11//6*x21^2 + 77//6*x21*x24 + x21*x32 + 2*x21*x33 + 3*x21*x34 + 2*x21*x35 + x21*x36 - 31//6*x22^2 - 5//6*x22*x23 - 97//12*x22*x25 - 2*x23^2 + 4*x24^2 + 17//6*x24*x27 + 1//2*x24*x32 + x24*x33 + 3//2*x24*x34 + x24*x35 + 1//2*x24*x36 - 7//4*x25^2 + 1//2*x25*x32 + x25*x33 + 3//2*x25*x34 + x25*x35 + 1//2*x25*x36 + 1//3*x26*x27 - 5//6*x27^2 + 1//2*x27*x32 + x27*x33 + 3//2*x27*x34 + x27*x35 + 1//2*x27*x36 - 1//2*x28^2 + 7//6*x29^2 - 5//2*x35*x36
```
HereAround: Maybe the output for this one could be suppressed? Again, I am thinking about the readability of the docs?

experimental/FTheoryTools/src/LiteratureModels/Models/model1903_00009.json
```diff
@@ -59,6 +59,7 @@
     "fiber_ambient_space_weights": [[1,0,-1,-1,0], [0,0,0,1,0], [0,0,1,-1,0], [1,-1,-1,0,-1], [0,0,0,0,1], [0,1,0,0,-1], [1,-1,0,0,0]],
     "D1": [0, 0],
     "D2": [0, 0],
-    "hypersurface_equation": "s1*e1^2*e2^2*e3*e4^4*u^3+s2*e1*e2^2*e3^2*e4^2*u^2*v+s3*e2^2*e3^3*u*v^2+s5*e1^2*e2*e4^3*u^2*w+s6*e1*e2*e3*e4*u*v*w+s9*e1*v*w^2"
+    "hypersurface_equation": "s1*e1^2*e2^2*e3*e4^4*u^3+s2*e1*e2^2*e3^2*e4^2*u^2*v+s3*e2^2*e3^3*u*v^2+s5*e1^2*e2*e4^3*u^2*w+s6*e1*e2*e3*e4*u*v*w+s9*e1*v*w^2",
+    "zero_section": "v"
```
HereAround: Is the zero section for this model really v? I do not recall it at the top of my head, just want to double-check. (Probably you already did that.)

experimental/FTheoryTools/src/G4Fluxes/properties.jl
emikelsons: Not really critical. This happened as a result of some testing, but I left it as it made sense to me to have the hyperplane class at the end so that the expression is more readily comparable to the doc description.

experimental/FTheoryTools/src/LiteratureModels/Models/model1903_00009.json
emikelsons: I saw it mentioned in the paper, it passed all of the tests and that was good enough for me

experimental/FTheoryTools/src/G4Fluxes/properties.jl
emikelsons: You're right I will make the docs more readable

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -1026,13 +1026,35 @@ Assuming that the first row of the given grading is the grading under Kbar
...
+
+@doc raw"""
+    zero_section(m::AbstractFTheoryModel)
+
+Return the zero section of a model as a cohomology class in the toric ambient space.
+If no zero section is known, an error is raised.
+This information is not typically available for
+Weierstrass and global Tate models, whose zero sections are known.
+
+```jldoctest
```
lgoettgens:
```suggestion
```jldoctest; setup = :(Oscar.LazyArtifacts.ensure_artifact_installed("QSMDB", Oscar.LazyArtifacts.find_artifacts_toml(Oscar.oscardir)))
```

experimental/FTheoryTools/src/G4Fluxes/properties.jl
```diff
@@ -52,9 +52,68 @@ true
...
+
+G4-fluxes are subject to verticality conditions described in [Lin16](@cite).
+It is hard to verify that these condition are met. However,
+we can execute a number of simple consistency checks, by
+verifying that $\int_{Y}{G_4 \wedge [D_1] \wedge [zero section]} = 0$ and $\int_{Y}{G_4 \wedge [D_1] \wedge [D_2]} = 0$
+for all toric base divisors $D_1$ and $D_2$. If all of these
+simple consistency checks are met, this method will return
+`true` and otherwise `false`
+
+```jldoctest
```
lgoettgens:
```suggestion
```jldoctest; setup = :(Oscar.LazyArtifacts.ensure_artifact_installed("QSMDB", Oscar.LazyArtifacts.find_artifacts_toml(Oscar.oscardir)))
```

lgoettgens (commented):
I found two missing `ensure_artifacts_installed`. Please make sure to always include this line in doctests accessing the QSMDB: As the order of doctests in non-deterministic, any doctest might be the first one accessing the QSMDB. And if this setup is then missing, it will make the doctest fail. See e.g. of such an instance.

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
emikelsons: I was wondering what was the purpose of this, if the tests seemingly passed. This changes makes sense.

HereAround:
> I found two missing `ensure_artifacts_installed`. Please make sure to always include this line in doctests accessing the QSMDB: As the order of doctests in non-deterministic, any doctest might be the first one accessing the QSMDB. And if this setup is then missing, it will make the doctest fail. See e.g. of such an instance.

Thank you for catching this @lgoettgens !

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -1245,7 +1245,7 @@ function euler_characteristic(m::AbstractFTheoryModel; check::Bool = true)
 
   # Check if the answer is known
   if has_attribute(m, :euler_characteristic)
-    return get_attribute(m, :euler_characteristic)::CohomologyClass
+    return get_attribute(m, :euler_characteristic)
```
lgoettgens: Why does this get removed here?

HereAround: The type is wrong. It should be changed to `::Int` (not dropped).

HereAround: Or another type for integers. Euler characteristics can be large. Maybe `ZZRingElem` (or alike?) makes more sense? Maybe you know @lgoettgens ?

lgoettgens: You should annotate here what really is saved in the attributes. Other functions for things like this take a IntegerUnion subtype (so eg Int, Int128, ZZRingElem) as the first argument to determine the return type. If something doesn't fit, the conversion in there will throw some error

HereAround: Ok. On 2nd thought, I think `Int` (`Int64`) is large enough for now. So a change to `::Int` should suffice (at least for now). But definitely not dropping the cast.

emikelsons: I have now changed it to ::Int

emikelsons: I also had to convert it to ::Int as initially it was of type QQFieldElem.

HereAround:
Thank you for working on this @emikelsons . This looks good to me!

experimental/FTheoryTools/src/G4Fluxes/properties.jl
```diff
@@ -63,7 +63,7 @@ end
 @doc raw"""
     passes_verticality_checks(gf::G4Flux)
 
-G4-fluxes are subject to verticality conditions described in [Lin16](@cite).
+G4-fluxes are subject to verticality conditions described in [Lin16](@cite) and [KMW12](@cite).
```
HereAround: I would suggest to link to Timo Weigand F-theory TASI lectures instead. Maybe in fact, replace those two references with this?

HereAround: @apturner also suggests to reference instead of Ling's thesis the historically first paper that ever talked about these conditions. For this, just check which papers the above Tasi lectures reference.

docs/oscar_references.bib
```diff
@@ -1490,6 +1490,19 @@ @Article{KMSS11
...
+  journal       = {Journal of High Energy Physics},
+  volume        = {2012},
+  number        = {8},
+  publisher     = {Springer Science and Business Media LLC},
+  year          = {2012},
+  month         = {8},
+  doi           = {10.1007/jhep08(2012)119},
+  eprint        = {1202.3138}
+}
+
```
HereAround: Presumably we do not need this reference? See comment below.

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -1245,7 +1248,7 @@ function euler_characteristic(m::AbstractFTheoryModel; check::Bool = true)
 
   # Check if the answer is known
   if has_attribute(m, :euler_characteristic)
-    return get_attribute(m, :euler_characteristic)::CohomologyClass
+    return get_attribute(m, :euler_characteristic)::Int
```
HereAround: This is a great change. We need this!

HereAround (changes requested):
I approve this PR modulo the above changes to the used references.

@emikelsons please adjust these references and ping me once you have done. Then this can be merged.

HereAround:
Another point raised just now is about the global naming scheme within `FTheoryTools`:
* `zero_section`: Returns a vector with coordinates (currently strings, but at some point polynomials, rational functions or wilder beasts yet...).
* `zero_section_class`: Return a cohomology class S in the toric ambient space s.t. S|Y4 is the cohomology class in Y4 of the zero section locus.

Bonus points:
When calling `zero_section` on a (partially) resolved model, we should get the zero section of the partially resolved model. Maybe we can obtain this by keeping track of it during the sequence of blowups? (If cannot be done easily within this PR, then let us open an issue/add this to our ToDo-List, so we do not forget about it.)

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
```diff
@@ -1212,6 +1234,9 @@ error.
 julia> qsm_model = literature_model(arxiv_id = "1903.00009", model_parameters = Dict("k" => 4))
 Hypersurface model over a concrete base
 
+julia> h = euler_characteristic(qsm_model; check = false)
+378
+
```
lgoettgens, line 1239:
```suggestion
```
these three lines are already there right below

emikelsons: I added these lines twice on purpose to function as an additional test

emikelsons:
I have changed the references though I was not able to find the historically first paper that mentions the verticality conditions. I also reverted back to zero_section for the original zero_section and introduced the method zero_section_class that return the cohomology class. @HereAround @apturner

experimental/FTheoryTools/src/AbstractFTheoryModels/attributes.jl
HereAround: Just to elaborate a bit: Once `euler_characteristic` is called for the 2nd time, it fetches the saved value of the attribute. This process failed in the past, because of the wrong type cast. Now this will be tested.

HereAround:
The historically first paper regarding the verticality conditions seems to be https://arxiv.org/abs/1111.1232 (cf. https://arxiv.org/pdf/1806.01854, page 86, below equation 9.8).

HereAround (approved):
Thank you for working on this @emikelsons . This looks good to me now.

PR #4046, opened 2024-08-26
TropicalGeometry: Vector to AbstractVector + general cleanup in groebner_basis.jl

YueRen:
general cleanup = removing named arguments when unused

### DIFF
--- a/src/TropicalGeometry/groebner_basis.jl
+++ b/src/TropicalGeometry/groebner_basis.jl
@@ -299,23 +299,24 @@ julia> desimulate_valuation(wSim, nuMax; perturbation=uSim)
 function desimulate_valuation(w::AbstractVector{QQFieldElem}, ::TropicalSemiringMap{K,p,typeof(min)}; perturbation::Union{Nothing,AbstractVector}=nothing) where {K,p}
     @req w[1]<0 "invalid weight vector"
     # scale the vector so that first entry is 1, then remove first entry
-    w = w ./ w[1]
+    w = w[2:end] ./ w[1]
     if !isnothing(perturbation)
         # negate vector, then remove first entry
-        perturbation *= -1
-        return w[2:end],perturbation[2:end]
+        perturbation = perturbation[2:end] * -1
+        return w,perturbation
     end
-    return w[2:end]
+    return w
 end
 function desimulate_valuation(w::AbstractVector{QQFieldElem}, ::TropicalSemiringMap{K,p,typeof(max)}; perturbation::Union{Nothing,AbstractVector}=nothing) where {K,p}
     @req w[1]<0 "invalid weight vector"
     # scale the vector so that first entry is -1, then remove first entry
-    w = w ./ -w[1]
+    w = w[2:end] ./ -w[1]
     if !isnothing(perturbation)
         # remove first entry
-        return w[2:end],perturbation[2:end]
+        perturbation = perturbation[2:end]
+        return w,perturbation
     end
-    return w[2:end]
+    return w
 end
 # if trivial valuation, just flip sign depending on convention
 function desimulate_valuation(w::AbstractVector{QQFieldElem}, ::TropicalSemiringMap{K,Nothing,typeof(min)}; perturbation::Union{Nothing,AbstractVector}=nothing) where {K}
--- a/src/TropicalGeometry/groebner_basis.jl
+++ b/src/TropicalGeometry/groebner_basis.jl
@@ -81,7 +81,7 @@ function clear_coefficient_denominators(f::MPolyRingElem)
     return lcm(denominator.(coefficients(f)))*f
 end
 # if valuation trivial, do nothing
-function simulate_valuation(I::MPolyIdeal, nu::TropicalSemiringMap{K,Nothing,<:Union{typeof(min),typeof(max)}}) where {K}
+function simulate_valuation(I::MPolyIdeal, ::TropicalSemiringMap{K,Nothing,<:Union{typeof(min),typeof(max)}}) where {K}
     return I
 end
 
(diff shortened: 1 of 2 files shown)


YueRen:
Note that in a lot of functions the input `TropicalSemiringMap` merely serves to select the right function using type dispatch (and for consistency).

src/TropicalGeometry/groebner_basis.jl
```diff
@@ -296,7 +296,7 @@ julia> desimulate_valuation(wSim, nuMax; perturbation=uSim)
 
 ```
 """
-function desimulate_valuation(w::Vector{QQFieldElem}, nu::TropicalSemiringMap{K,p,typeof(min)}; perturbation::Union{Nothing,Vector}=nothing) where {K,p}
+function desimulate_valuation(w::AbstractVector{QQFieldElem}, ::TropicalSemiringMap{K,p,typeof(min)}; perturbation::Union{Nothing,AbstractVector}=nothing) where {K,p}
     @req w[1]<0 "invalid weight vector"
     # scale the vector so that first entry is 1, then remove first entry
     w ./= w[1]
```
joschmitt: This does not work for (every) `AbstractVector`, as far as I can see:
```
julia> w = 1:2
1:2

julia> w isa AbstractVector
true

julia> w ./= w[1]
ERROR: CanonicalIndexError: setindex! not defined for UnitRange{Int64}
Stacktrace:
[...]
```
(If the elements of `w` are `QQFieldElem`s, the error is weirder...)

YueRen:
Hm, okay. I was under the wrong impression that any AbstractVector could do
what Vector can do. :-/

I just need a simple way to write a function that takes as input a Vector,
PointVector, or RayVector. (basically anything that can arise from
polyhedral geometry).

Should I create an alias for their union?

benlorenz:
> This does not work for (every) `AbstractVector`, as far as I can see:
> 
> ```
> julia> w = 1:2
> 1:2
> 
> julia> w isa AbstractVector
> true
> 
> julia> w ./= w[1]
> ERROR: CanonicalIndexError: setindex! not defined for UnitRange{Int64}
> Stacktrace:
> [...]
> ```
> 

`1:2` is not a subtype of `AbstractVector{QQFieldElem}` so this wont run into this code.

Modifying the input vector is very bad style anyway, so doing something like
```
w = w[2:end] ./ w[1]
```
and using just `w` further down would be better and should work with all (reasonable) vector types.
(Julia does support vectors with custom indices which still might break and we could add more checks for stuff like this but I don't think this is sustainable)

> (If the elements of `w` are `QQFieldElem`s, the error is weirder...)

What is the error for `QQFieldElem` ?

> Hm, okay. I was under the wrong impression that any AbstractVector could do what Vector can do. :-/ I just need a simple way to write a function that takes as input a Vector, PointVector, or RayVector. (basically anything that can arise from polyhedral geometry). Should I create an alias for their union?

We cannot really guarantee compatibility with any AbstractVector but I would still recommend just keeping AbstractVector and not some Union here.

joschmitt:
> `1:2` is not a subtype of `AbstractVector{QQFieldElem}` so this wont run into this code.

Yes, sorry, I felt like the error one gets for `QQFieldElem` was a bit of a red herring, but here it is:
```
julia> w = QQ(1):QQ(2)
1:1:2

julia> typeof(w)
StepRange{QQFieldElem, QQFieldElem}

julia> w isa AbstractVector{QQFieldElem}
true

julia> w ./= w[1]
ERROR: MethodError: no method matching Integer(::QQFieldElem)

Closest candidates are:
  (::Type{T})(::ArbFieldElem) where T<:Integer
   @ Nemo ~/.julia/packages/Nemo/QNyoV/src/arb/arb.jl:161
  Integer(::ZZRingElem)
   @ Nemo ~/.julia/packages/Nemo/QNyoV/src/HeckeMiscInteger.jl:10
  (::Type{T})(::BigFloat) where T<:Integer
   @ Base mpfr.jl:378
  ...

Stacktrace:
 [1] length(r::StepRange{QQFieldElem, QQFieldElem})
   @ Base ./range.jl:753
 [2] axes(r::StepRange{QQFieldElem, QQFieldElem})
   @ Base ./range.jl:706
 [3] materialize!
   @ ./broadcast.jl:914 [inlined]
 [4] materialize!(dest::StepRange{…}, bc::Base.Broadcast.Broadcasted{…})
   @ Base.Broadcast ./broadcast.jl:911
 [5] top-level scope
   @ REPL[7]:1
```
So, `length(::StepRange{QQFieldElem, QQFieldElem})` appears to be missing.

EDIT: This is not a surprise because it just uses the Julia step ranges. If one wants to have more functionality, one would need to implement dedicated `QQFieldElem` step ranges like we have for `ZZRingElem`, I suppose.

benlorenz:
> > `1:2` is not a subtype of `AbstractVector{QQFieldElem}` so this wont run into this code.
> 
> Yes, sorry, I felt like the error one gets for `QQFieldElem` was a bit of a red herring, but here it is:
> ...
> 
> So, `length(::StepRange{QQFieldElem, QQFieldElem})` appears to be missing.
> 
> EDIT: This is not a surprise because it just uses the Julia step ranges. If one wants to have more functionality, one would need to implement dedicated `QQFieldElem` step ranges like we have for `ZZRingElem`, I suppose.

The error shows that this problem is totally unrelated to supporting `AbstractVector` types here, since this type seems very broken, even `collect(QQ(1):QQ(2))` fails.

_Edit:_ 
This type is also not supported according to the description of `StepRange{T, S} <: OrdinalRange{T, S}`:
```julia
julia> w = QQ(1):QQ(2);

julia> w isa OrdinalRange{QQFieldElem,QQFieldElem}
true

help?> OrdinalRange
search: OrdinalRange

  OrdinalRange{T, S} <: AbstractRange{T}

  Supertype for ordinal ranges with elements of type T with spacing(s) of type S. The steps
  should be always-exact multiples of oneunit, and T should be a "discrete" type, which
  cannot have values smaller than oneunit. For example, Integer or Date types would qualify,
  whereas Float64 would not (since this type can represent values smaller than
  oneunit(Float64). UnitRange, StepRange, and other types are subtypes of this.
```

joschmitt:
Okay. What I wanted to point out was that `AbstractVector` only supports `size` and `getindex` in general (according to https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array). Everything else is optional.

I have no stakes in this; if it works, it works.

src/TropicalGeometry/groebner_basis.jl
```diff
@@ -307,7 +307,7 @@ function desimulate_valuation(w::Vector{QQFieldElem}, nu::TropicalSemiringMap{K,
     end
     return w[2:end]
 end
-function desimulate_valuation(w::Vector{QQFieldElem}, nu::TropicalSemiringMap{K,p,typeof(max)}; perturbation::Union{Nothing,Vector}=nothing) where {K,p}
+function desimulate_valuation(w::AbstractVector{QQFieldElem}, ::TropicalSemiringMap{K,p,typeof(max)}; perturbation::Union{Nothing,AbstractVector}=nothing) where {K,p}
     @req w[1]<0 "invalid weight vector"
     # scale the vector so that first entry is -1, then remove first entry
     w ./= -w[1]
```
benlorenz:
```suggestion
    w = w ./ -w[1]
```
Or even better I guess directly `w = w[2:end] ./ -w[1]`, but I cannot change some of the lines below in this suggestion to remove the `[2:end]`.

This does not modify the original vector unlike the broadcasted assignment. (And doesn't require `setindex!`)

src/TropicalGeometry/groebner_basis.jl
benlorenz:
```suggestion
    w = w ./ w[1]
```
see below

YueRen:
I just need the code to work regardless whether the input is a `Vector`, `PointVector`, or `RayVector`, and using `AbstractVector` seems like the simplest solution for now.  

Maybe having an alias for the union of all three is a cleaner solution (disadvantage: more bloat with structs), or maybe actually restricting oneself to the limited features of `AbstractVector` (disadvantage: less readable code), but I think that is a longer discussion with low priority.

I've just pushed a commit and all immediate issues should be accounted for.

benlorenz (approved):

joschmitt:
Is it possible to add a small test that calls these functions with a `PointVector` or `RayVector`?

YueRen:
@joschmitt Sorry it took so long, tests should be added now, hope it goes through.

joschmitt:
> @joschmitt Sorry it took so long, tests should be added now, hope it goes through.

Thank you!

PR #4049, opened 2024-08-27
add warning to load

antonydellavecchia:
Warns users when trying to load a file serialized with a newer version of Oscar (if there was an error during load)


HereAround (approved):
Thank you @antonydellavecchia . Looks good to me and - as I see it - can be merged whenever you give the call @antonydellavecchia .

PR #3996, opened 2024-08-06
[ToricVarieties] Introduce Chern classes of tangent bundle

HereAround:
This PR is inspired by developments in `FTheoryTools`. It aims to provide functionality to compute the Chern classes of the tangent bundle of a simplicial and complete toric variety.

Issue #922, opened 2021-12-22
What to do for object clashes?


lkastner:
This discussion already popped up a while ago and is also reflected in #769 as an example. The problem is the following: Some mathematical objects appear in different corners of Oscar, so who gets this object? Does anyone?

For example: Many mathematical disciplines know a `projective_space`, for example algebraic geometry and toric geometry. These lead to two different descriptions. So what should happen if someone writes `projective_space(n)`? Currently this is solved very awkwardly with `toric_projective_space`.

Another new example is `torus(...)`.

I don't think introducing a super object capturing all the aspects will solve this. Maybe a module solution could work, as I think that we can at least expect the user to know what kind of mathematics they want to do. Nevertheless if you arrive at another solution, even the superobject, I will happy that you arrived at a solution.

micjoswig:
I see essentially two ways of resolving this.  Either such a function gets a traits class as an optional argument, similar to, e.g.,
```
facets(as::Type{T} = AffineHalfspace, P::Polyhedron)
```
Or we introduce modules for separation:
```
SimplicialComplexes.torus()
AlgebraicGroups.torus(QQ, 3)
```
Don't be fooled here that you could tell these apart by their signatures. First, there are other examples (like the projective spaces that Lars mentioned).  Second, this would be totally intuitive for the user. I would consider this very bad style.

I am slightly leaning toward the modules.  To me this looks like the cleaner solution.

micjoswig:
Of course, there is a third solution:
```
torus_as_simplicial_complex()
torus_as_algebraic_group(QQ, 3)
```
But I don't see any advantage here.  So that would be my least favored.

fieker:
On Wed, Dec 22, 2021 at 04:49:30AM -0800, Michael Joswig wrote:
> I see essentially two ways of resolving this.  Either such a function gets a traits class as an optional argument, similar to, e.g.,
> ```
> facets(as::Type{T} = AffineHalfspace, P::Polyhedron)
> ```
We're using this version in several places already, julia does it e.g.
for ceil, round, floor, ...
ceil(x) has the same type as x
ceil(Int, x) yields an Int
...
We're using this e.g. for cylotomic_field...

Here, as above, we do not have any decent module to associate with.

> Or we introduce modules for separation:
> ```
> SimplicialComplexes.torus()
> AlgebraicGroups.torus(QQ, 3)
> ```
For BinomialIdeals this is happening. Supposed to be happening or so.

[...]

PR #4051, opened 2024-08-28
[ToricVarieties] Shorthands for hirzebruch surface and del Pezzo surface

HereAround:
As pointed out by @micjoswig in https://github.com/oscar-system/Oscar.jl/pull/3996, for the current constructor `hirzebruch_surface(::Type{NormalToricVariety}, r::Int)`, the first argument is currently obsolete. This PR introduces a shorthand `hirzebruch_surface(r::Int)`.

A similar situation exists for `del_pezzo_surface`, and I have taken the liberty to propose a similar short hand.

We also have the following specialized constructors:
* `affine_space(::Type{NormalToricVariety}, var_names::AbstractVector{<:VarName})`
* `affine_space(::Type{NormalToricVariety}, d::Int)`
* `projective_space(::Type{NormalToricVariety}, d::Int)`
* `weighted_projective_space(::Type{NormalToricVariety}, w::Vector{T}) where {T <: IntegerUnion}`

I could not find instances, where those constructors are called with a first argument different from `NormalToricVariety`. (But of course, I could have overlooked those instances...) In https://github.com/oscar-system/Oscar.jl/pull/3996, I was therefore carried away to propose short hands for all of these constructors. This is (at least not yet) done in this PR, as this seems to require discussion and so I removed this proposed change after being alerted to this by @lgoettgens .

This PR aims to clarify the situation and to resolve it.

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/NormalToricVarieties/standard_constructions.jl
@@ -115,6 +115,7 @@ function hirzebruch_surface(::Type{NormalToricVariety}, r::Int)
   return variety
 end
 
+hirzebruch_surface(r::Int) = hirzebruch_surface(NormalToricVariety, r)
 
 @doc raw"""
     del_pezzo_surface(::Type{NormalToricVariety}, b::Int)
@@ -169,6 +170,8 @@ function del_pezzo_surface(::Type{NormalToricVariety}, b::Int)
   set_coordinate_names(variety, vars[1:(3 + b)])
   return variety
 end
+
+del_pezzo_surface(r::Int) = del_pezzo_surface(NormalToricVariety, r)
 
 @doc raw"""
     Base.:*(v::NormalToricVarietyType, w::NormalToricVarietyType)


lgoettgens:
Just to note: The current format was decided on in https://github.com/oscar-system/Oscar.jl/issues/922.

The other area in Oscar that came to my mind with these function names was geometry/schemes. Thus pinging some people to get opinions on this @jankoboehm @afkafkafk13 @HechtiDerLachs

afkafkafk13:
We should postpone the discussion of this point, until @simonbrandhorst is back from his holidays (1.9.).

simonbrandhorst:
Agreed, we should add constructors for 

- `affine_space(::Type{AffineScheme}, base_ring::Ring, var_names::AbstractVector{<:VarName})`
-  `affine_space(::Type{AffineScheme}, base_ring::Ring, d::Int)`
-  `projective_space(::Type{ProjectiveScheme},  base_ring::Ring, d::Int)`

So far the shorthands for `affine_space` etc. just return the above. 
One could think about disabling the current shortcuts (would break the interface) or introduce other default 
... for me getting an affine space as a scheme always seemed a reasonable default. But of course I am biased here :-).

afkafkafk13:
> Agreed, we should add constructors for
> 
>     * `affine_space(::Type{AffineScheme}, var_names::AbstractVector{<:VarName})`
> 
>     * `affine_space(::Type{AffineScheme}, d::Int)`
> 
>     * `projective_space(::Type{ProjectiveScheme}, d::Int)`
> 
> 
> So far the shorthands for `affine_space` etc. just return the above. One could think about disabling the current shortcuts (would break the interface) or introduce other default ... for me getting an affine space as a scheme always seemed a reasonable default. But of course I am biased here :-).

Given the existence of different perspectives on such objects in different parts of Oscar and its community, we should probably deprecate the current shorthands and not introduce shorthands for Hirzebruch and DelPezzo surfaces. What do you think?

HereAround:
We just discussed this in triage. As we see it, this is waiting for a PI decision @micjoswig @fingolfin @fieker @thofma @simonbrandhorst @wdecker . Once decided, I will be happy to execute your decision.

micjoswig:
> We just discussed this in triage. As we see it, this is waiting for a PI decision @micjoswig @fingolfin @fieker @thofma @simonbrandhorst @wdecker . Once decided, I will be happy to execute your decision.

I would say we keep `projective_space` as is (maybe adding more types further down the road) but we also keep the possibility to have other functions (like, e.g., for weighted projective spaces and Hirzebruch surfaces) which follow a different strategy.

Let me explain my reasoning.

There is no canonical solution here, but we need to make choices (that's why this thing is pending). For me the overarching principle to derive decisions from is the user's perspective.  And then there is an immediate difference: projective spaces do occur everywhere in math, and thus in various categories; so the user should pick.  It is debatable if there should be a default or not.  I slightly tend to no default here.

Of course, I can also view a Hirzebruch surface from a scheme theoretic point of view, but this is somehow missing the point.  In toric geometry they arise somehow canonically, as the "simplest" toric varieties which are neither projective spaces nor products.  So I would claim that a Hirzebruch surface is primarily an object of toric geometry, and so its natural incarnation is as a `NormalToricVariety`.  So, the corresponding type parameter should be unnecessary (or optional).

Long story short, these decisions need to be done case by case (and properly documented).

simonbrandhorst:
A `NormalToricVariety` has as base ring `QQ`. 
What would `hirzebruch_surface(QQ, n)` or `hirzebruch_surface(GF(p), n)` return (in a possible future)? 
How would that relate to a possible default of `hirzebruch_surface(n)`?

I would strive towards: 
1) If there are several implementations of the same object in the same category, we have a default without type parameter. Based (hopefully) on usefulness/performance of the type in the most common use cases. 
2) If there are several competing categories an object with the same name can live in, do not pick a default and make the type parameter the first argument (probably the abstract type). 

If in doubt do not pick a default. This has the advantage of limiting friction amongst different parts of Oscar and among us as developers [Is object X primarily an object of category A or primarily in category B?]. My rationale would be that consistency of the system is more important than short names. 

The `projective_space` and `hirzebruch` issue is something in between, since a `NormalToricVariety` is also a scheme.
But it has as default the base ring `QQ` and it does not (cannot?) have a method called `homogeneous_coordinates`. On the other hand there is also a category of toric varieties. 

Thus "if in doubt" applies and I would not impose a default for  `projective_space` and `hirzebruch_surface`.

`projective_space(Scheme, GF(3), 2)`
`projective_space(ProjectiveVariety, GF(3), 2)`
`projective_space(ProjectiveScheme, GF(3), 2)`
Do not look too bad to me.

micjoswig:
>  What would `hirzebruch_surface(QQ, n)` or `hirzebruch_surface(GF(p), n)` return (in a possible future)? 

I don't think that either of these two makes much sense.

> How would that relate to a possible default of `hirzebruch_surface(n)`?

So there is no issue here.

> If in doubt do not pick a default.

Here I agree.

wdecker:
> Long story short, these decisions need to be done case by case (and properly documented).

I agree. And certainly Delpezzo  and Hirzebruch surfaces will show up also in classical algebraic geometry (that is, without considering their toric structure). So please, no shortcuts here.

simonbrandhorst:
> > What would `hirzebruch_surface(QQ, n)` or `hirzebruch_surface(GF(p), n)` return (in a possible future)?
> 
> I don't think that either of these two makes much sense.

Maybe they do not in the toric setting. There are however Hirzebruch surfaces in algebraic geometry over any base field. Quoting wikipedia:
"In mathematics, a Hirzebruch surface is a ruled surface over the projective line. They were studied by Friedrich Hirzebruch (1951). "

HechtiDerLachs:
> > What would `hirzebruch_surface(QQ, n)` or `hirzebruch_surface(GF(p), n)` return (in a possible future)?
> 
> I don't think that either of these two makes much sense.

Here's a way to make sense of it straight from their definition:
```julia
function hirzebruch_surface(kk::Ring, n::Int)
  PP1 = projective_space(kk, 1)
  X = covered_scheme(PP1)
  E = Oscar.DirectSumSheaf(X, [twisting_sheaf(PP1, 0), twisting_sheaf(PP1, n)])
  PE = projectivization(E)
  return covered_scheme(PE)
end
```
Probably this should go under 
```julia
function hirzebruch_surface(::Type{T}, kk::Ring, n::Int) where T<:AbsCoveredScheme
```
but that's not for me to decide. The point is: We could already add it in a `10`-liner and people could use it to study e.g. intersection numbers, subvarieties, ...

wdecker:
> intersection numbers

We have, by the way, already the function `abstract_hirzebruch_surface(n::Int)` in abstract intersection theory.

afkafkafk13:
@micjoswig  You already stated it yourself:
> For me the overarching principle to derive decisions from is the user's perspective. 

With as broad a system as Oscar, we encounter different communities of users and hence different scopes in which different return values from shorthands will be expected as natural. This is unavoidable. Each time the question will be  whether to give a specific one of them precedence over the others or not -- or maybe even in the long run how to encode such scopes  to allow for shorthands without confusing a part of the user base.

micjoswig:
> > > What would `hirzebruch_surface(QQ, n)` or `hirzebruch_surface(GF(p), n)` return (in a possible future)?
> > 
> > 
> > I don't think that either of these two makes much sense.
> 
> Maybe they do not in the toric setting. There are however Hirzebruch surfaces in algebraic geometry over any base field. Quoting wikipedia: "In mathematics, a Hirzebruch surface is a ruled surface over the projective line. They were studied by Friedrich Hirzebruch (1951). "

OK, fine.

Does it make sense then to stick to the pattern
```
projective_space(Scheme, GF(3), 2)
projective_space(ProjectiveVariety, GF(3), 2)
projective_space(ProjectiveScheme, GF(3), 2)
```
thoughout then, i.e., also for Hirzebruch surfaces (as suggested by @HechtiDerLachs)? OK with me.

wdecker:
> Does it make sense then to stick to the pattern
> 
> ```
> projective_space(Scheme, GF(3), 2)
> projective_space(ProjectiveVariety, GF(3), 2)
> projective_space(ProjectiveScheme, GF(3), 2)
> ```

Good point! We have

julia> ProjectiveVariety<:ProjectiveScheme
false

julia> ProjectiveScheme<:Scheme
true

@simonbrandhorst:  Why do I get false?  In the docu I find e.g.: 

`AbsProjectiveAlgebraicSet <: AbsProjectiveScheme`
A projective, geometrically reduced scheme of finite type over a field.

Otherwise, with regard to the above pattern, `projective_space(ProjectiveVariety, GF(3), 2)` would be enough. In any case, `projective_space(Scheme, GF(3), 2)` seems to be superfluous?

simonbrandhorst:
> > Does it make sense then to stick to the pattern
> > ```
> > projective_space(Scheme, GF(3), 2)
> > projective_space(ProjectiveVariety, GF(3), 2)
> > projective_space(ProjectiveScheme, GF(3), 2)
> > ```
> 
> Good point! We have
> 
> julia> ProjectiveVariety<:ProjectiveScheme false
> 
> julia> ProjectiveScheme<:Scheme true
> 
> @simonbrandhorst: Why do I get false? In the docu I find e.g.:
> 
> `AbsProjectiveAlgebraicSet <: AbsProjectiveScheme` A projective, geometrically reduced scheme of finite type over a field.
> 
> Otherwise, with regard to the above pattern, `projective_space(ProjectiveVariety, GF(3), 2)` would be enough. In any case, `projective_space(Scheme, GF(3), 2)` seems to be superfluous?

The reason is that both `ProjectiveVariety` and `ProjectiveAlgebraicSet` are concrete types and 
concrete types may not subtype each other bin Julia. 
We still need `Scheme` in any case, because an AlgebraicSet and a Variety must be defined over a field. But projective space can have any base.

fingolfin:
My view point is that requiring the desired "kind" of output (via a type or some other way) is the way to go, i.e. the suggestion in the last couple comments seems sane.

If there is agreement on this, we can proceed. Or is there something else that warrants a decision? What?

fieker:
On Wed, Dec 18, 2024 at 02:44:57AM -0800, Max Horn wrote:
> My view point is that requiring the desired "kind" of output (via a type or some other way) is the way to go, i.e. the suggestion in the last couple comments seems sane.
> 
> If there is agreement on this, we can proceed. Or is there something else that warrants a decision? What?
> 
I'd add that the interface should be able to lie...
eg. we support a "type" CyclotomicField - which does not exist.
It was done to have a "sane" interface to users, not to make Julia
happy.
an irred. group rep is returned over the QQab (the abelian closure) by
default. If you want ot over a cyclo. you do
gmodule(CyclotomicField, ...)
to get the obvious.

fingolfin:
We just discussed it again and it seems the consensus was that we use the first argument to return the desired return type. @HereAround tells me that in this case this PR can be closed.

HereAround:
Side note: This decision can actually already be found in the developer guide:
* https://docs.oscar-system.org/stable/DeveloperDocumentation/styleguide/#General-styleguide
* 4th bullet point:

PR #4052, opened 2024-08-28
Make GSets better inferreable

lgoettgens:
By adding a type parameter which represents the element type of the G-set (derived from the eltype of the seed iterator).
This then allows to concretize the return types of `orbits` and `elements` to something concretely that is statically known, i.e. in these two functions there is no dynamic dispatch happening, and they infer their concrete return type.

I noticed this issue while profiling some new functionality in https://github.com/oscar-system/Oscar.jl/pull/4053, since half of the flame graph was filled with dynamic dispatch stuff.

### DIFF
--- a/src/Groups/group_characters.jl
+++ b/src/Groups/group_characters.jl
@@ -1233,6 +1233,8 @@ end
 
 Base.iterate(tbl::GAPGroupCharacterTable, state = 1) = state > nrows(tbl) ? nothing : (tbl[state], state+1)
 
+Base.eltype(::Type{GAPGroupCharacterTable}) = GAPGroupClassFunction
+
 """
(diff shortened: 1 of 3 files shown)


ThomasBreuer (commented):
Thanks.
Just for curiosity: If we introduce a type parameter that describes the elements of the G-set then wouldn't it be more logical to take the element type itself and not the type of the seeds?

lgoettgens:
> Thanks. Just for curiosity: If we introduce a type parameter that describes the elements of the G-set then wouldn't it be more logical to take the element type itself and not the type of the seeds?

That is indeed a good question. I'll play around with it tomorrow to see if I can get it working like that

ThomasBreuer (approved):
Looks good. Thanks.

PR #4053, opened 2024-08-28
Lie algebras: more progress

lgoettgens:
Contains the following changes/additions:
- new: `isomorphism(FPGroup, ::WeylGroup)` via Coxeter relations. Already known information like `is_finite` and `order` gets set in the result as well to allow for better handling in GAP.
- new: conversions and arithmetic between RootSpaceElem and WeightLatticeElem
- new: `dot(::WeightLatticeElem, ::WeightLatticeElem)`
- change: `dominant_character`, `character`, and `tensor_product_decomposition` now use the root system, if available, and only fall back to GAP otherwise. (@benlorenz this should resolve the rest of cases of https://github.com/oscar-system/Oscar.jl/pull/3913#issuecomment-2209374911)

~~This is still a draft since I would like to wait for https://github.com/oscar-system/Oscar.jl/pull/4036 to get merged first to avoid conflicts.~~


experimental/LieAlgebras/src/CoxeterGroup.jl
```diff
@@ -7,26 +7,31 @@ Return the Coxeter matrix $m$ associated to the Cartan matrix `gcm`. If there is
...
   end
 
-  return cm
+  if gcm[i, j] * gcm[j, i] == 0
+    return 2
+  elseif gcm[i, j] * gcm[j, i] == 1
+    return 3
+  elseif gcm[i, j] * gcm[j, i] == 2
+    return 4
+  elseif gcm[i, j] * gcm[j, i] == 3
```
fingolfin: Maybe apply this micro optimization?
```suggestion
  d = gcm[i, j] * gcm[j, i]
  if d == 0
    return 2
  elseif d == 1
    return 3
  elseif d == 2
    return 4
  elseif d == 3
```

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1431,11 +1431,65 @@ function dim_of_simple_module(L::LieAlgebra, hw::Vector{<:IntegerUnion})
...
+          T(w) => d for (w, d) in
+          zip(
+            GAP.Globals.DominantWeights(
+              GAP.Globals.RootSystem(codomain(Oscar.iso_oscar_gap(L))),
+              GAP.Obj(hw; recursive=true),
+            )...,
+          )
+        ); by=last)
+    )
+  end
```
fingolfin, line 1476: Maybe use `partialsort!`? E.g. like this
```suggestion
    return partialsort!(
        collect(
          T(w) => d for (w, d) in
          zip(
            GAP.Globals.DominantWeights(
              GAP.Globals.RootSystem(codomain(Oscar.iso_oscar_gap(L))),
              GAP.Obj(hw; recursive=true),
            )...,
          )
        ), 1; by=last)
  end
```

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1476,29 +1542,37 @@ Dict{Vector{Int64}, Int64} with 10 entries:
...
-  return c
 end
 
 @doc raw"""
-    tensor_product_decomposition(L::LieAlgebra, hw1::Vector{Int}, hw2::Vector{Int}) -> MSet{Vector{Int}}
+    tensor_product_decomposition(L::LieAlgebra, hw1::Vector{<:IntegerUnion}, hw2::Vector{<:IntegerUnion}) -> MSet{Vector{Int}}
 
 Computes the decomposition of the tensor product of the simple modules of the Lie algebra `L` with highest weights `hw1` and `hw2`
 into simple modules with their multiplicities.
+This function uses Klymik's formula.
```
fingolfin: Never heard that (which of course just means I am clueless), perhaps you could add a `cite`?

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens: I think this does something different. Please note that I use `first.` (i.e. broadcast) and not only return the first element after sorting.

fingolfin: Indeed, I missed the `.` for broadcasting, sorry!

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens:
```suggestion
This function uses Klimyk's formula (see [Hum72; Exercise 24.9](@cite)).
```
The problem with references here is that different papers attribute it to different papers by different people (some say Brauer was earlier than Klimyk), but all of these references are not available online for me to check. So I instead decided to just reference the (in this part of Oscar standard reference) Hum72 (Humphreys' Introduction to Lie algebras) here, who give this formula as an exercise.

experimental/LieAlgebras/src/WeylGroup.jl
```diff
@@ -637,7 +669,7 @@ end
 function Base.iterate(iter::WeylOrbitIterator)
   (wt, _), data = iterate(iter.nocopy)
   # wt is already a copy, so here we don't need to make one
```
felix-roehrich: This line is wrong now.
```suggestion
```

experimental/LieAlgebras/src/RootSystem.jl
```diff
@@ -855,23 +951,43 @@ end
 ###############################################################################
 # more functions
 
+function Base.:+(r::RootSpaceElem, w::WeightLatticeElem)
```
felix-roehrich: I am not really happy with this. There is no clear reason why the return type should be `WeightLatticeElem` and not `RootSystemElem`. Is there a reason this function should exist?
@fingolfin What is your opinion?

In the case we go forward with this, there should be at least a test to verify the return type.

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1398,10 +1398,10 @@ function simple_module(L::LieAlgebra, hw::Vector{Int})
 end
 
 @doc raw"""
-    dim_of_simple_module([T = Int], L::LieAlgebra{C}, hw::Vector{<:IntegerUnion}) -> T
+    dim_of_simple_module([T = Int,] L::LieAlgebra{C}, hw::Vector{<:IntegerUnion}) -> T
```
felix-roehrich:
```suggestion
    dim_of_simple_module([T = Int], L::LieAlgebra{C}, hw::Vector{<:IntegerUnion}) -> T
```

experimental/LieAlgebras/src/LieAlgebraModule.jl
```diff
@@ -1431,11 +1431,65 @@ function dim_of_simple_module(L::LieAlgebra, hw::Vector{<:IntegerUnion})
 end
 
 @doc raw"""
-    dominant_character(L::LieAlgebra{C}, hw::Vector{Int}) -> Dict{Vector{Int}, Int}
+    dominant_weights([T,] L::LieAlgebra{C}, hw::Vector{<:IntegerUnion}) -> Vector{T}
```
felix-roehrich, line 1434:
```suggestion
    dominant_weights([T = Vector{Int}], L::LieAlgebra{C}, hw::Vector{<:IntegerUnion}) -> Vector{T}
```

felix-roehrich (changes requested):

experimental/LieAlgebras/src/LieAlgebraModule.jl
lgoettgens: I did not want to specify this at this point, as I want to straight out all of the return types in the near future. That's also the reason for the many `The return type will change in the future" in docstrings.

experimental/LieAlgebras/src/RootSystem.jl
lgoettgens: I think this is no longer needed. I'll try to remove it.

fingolfin (approved):
Looks good to me

PR #4056, opened 2024-08-29
cache Galois results (in AbsSimpleField)

fieker:


thofma (approved):
I added a simple test

PR #4057, opened 2024-08-29
make sure char values are in the correct field

fieker:
could/ chould fix a Heisenbug (cannot convert Rational to MultGroupElem) and others....

### DIFF
--- a/experimental/GModule/src/GModule.jl
+++ b/experimental/GModule/src/GModule.jl
@@ -807,7 +807,7 @@ gmodule(k::fpField, C::GModule{<:Any, <:AbstractAlgebra.FPModule{fpFieldElem}}) 
       continue
     end
     T = action(C, r)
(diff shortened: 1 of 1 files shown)


experimental/GModule/src/GModule.jl
```diff
@@ -807,7 +807,7 @@ gmodule(k::fpField, C::GModule{<:Any, <:AbstractAlgebra.FPModule{fpFieldElem}})
       continue
     end
     T = action(C, r)
-    push!(chr, (c, trace(matrix(T))))
+    push!(chr, (c, K(trace(matrix(T)))))
```
thofma, line 810: Out of curiosity, how could that not be in `K`?

fieker: Not sure.. I am scared about the default field(s), the GAP <-> Oscar FF conversion and algebraic closures...

PR #4059, opened 2024-08-29
Add `@attr` types in simple cases

lgoettgens:
If one only adds `@attr` to a function without giving a type, one shadows all inference of this function for callers of this function as the `@attr` code changes the inference result to `Any`. This PR thus adds types there, where it was obvious to me what the return type is, or where the compiler was able to prove a concrete return type (if one removes the `@attr` to allow for real inference).

The commits are partitioned by the area they touch. Let me ping some people in the different areas here to make them aware / get a review.
- Lie algebras -> myself, no review needed
- lattices -> @StevellM or maybe @ThomasBreuer ?
- schemes -> @afkafkafk13 @HechtiDerLachs 
- rings -> @jankoboehm @HechtiDerLachs 
- groups -> @ThomasBreuer

### DIFF
--- a/experimental/LieAlgebras/src/AbstractLieAlgebra.jl
+++ b/experimental/LieAlgebras/src/AbstractLieAlgebra.jl
@@ -82,7 +82,7 @@ end
 #
 ###############################################################################
 
-@attr function is_abelian(L::AbstractLieAlgebra)
+@attr Bool function is_abelian(L::AbstractLieAlgebra)
   return all(iszero, _struct_consts(L))
 end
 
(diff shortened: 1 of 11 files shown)


afkafkafk13 (commented):
Those in my courtyard are fine. ( src/AlgebraicGeometry/Schemes/CoveredSchemes/Objects/Properties.jl )
ToricVarieties is in @HereAround 's courtyard.

afkafkafk13 (commented):
I'm not @jankoboehm , but from my point of view the changes around rings are spot-on.

StevellM:
For `lattices_with_isometry.jl` and `spinor_norms.jl`, if the tests pass, then I am good.

ThomasBreuer:
Yes, this is o.k.

afkafkafk13 (approved):
As everyone answered in the meantime, I now put this on approve.

PR #4060, opened 2024-08-29
Speed up loading containers of polymake types

antonydellavecchia:
using the json function was slowing down deserialization of vectors of polymake types.
We use JSON3's write function instead.


src/Serialization/main.jl
```diff
@@ -653,9 +653,9 @@ function load(io::IO; params::Any = nothing, type::Any = nothing,
   end
 
   if file_version < VERSION_NUMBER
-    jsondict = JSON.parse(json(s.obj), dicttype=Dict{Symbol, Any})
+    jsondict = JSON.parse(JSON3.write(s.obj), dicttype=Dict{Symbol, Any})
     jsondict = upgrade(file_version, jsondict)
-    s.obj = JSON3.read(json(jsondict))
+    s.obj = JSON3.read(JSON2.write(jsondict))
```
benlorenz:
```suggestion
    s.obj = JSON3.read(JSON3.write(jsondict))
```

src/Serialization/main.jl
```diff
@@ -653,13 +653,12 @@ function load(io::IO; params::Any = nothing, type::Any = nothing,
   end
 
   if file_version < VERSION_NUMBER
-    jsondict = JSON.parse(json(s.obj), dicttype=Dict{Symbol, Any})
+    jsondict = JSON.parse(JSON3.write(s.obj), dicttype=Dict{Symbol, Any})
```
benlorenz:
```suggestion
    # we need a mutable dictionary
    jsondict = copy(s.obj)
```
I would prefer to do it this way but that might not work as expected if `s.obj` is not from `JSON3`. Maybe we need some helper method to extract the json as proper dict which dispatches on the different `s.obj` types ?

src/Serialization/serializers.jl
```diff
@@ -224,7 +224,7 @@ function serializer_open(
 end
 
 function deserializer_open(
-  io::IO,
+  io::Union{IO, String},
```
lgoettgens: wouldn't it be nicer for the interface to wrap string arguments in an `IOBuffer` before calling this function? That way, everything would be of type `IO`

lkastner (approved):

PR #4061, opened 2024-08-30
Overhaul tropical varieties, add various new options

YueRen:
New feature: Interface to `tropicalVariety` in Singular for tropicalization of ideals.


src/TropicalGeometry/variety_binomial.jl
```diff
@@ -0,0 +1,42 @@
...
+################################################################################
+
+function tropical_variety_binomial(I::MPolyIdeal,nu::TropicalSemiringMap; weighted_polyhedral_complex_only::Bool=false)
+    ###
+    # Construct matrix of exponent vector differences
+    # and vector of coefficient valuation differences
+    ###
+    G = gens(I)
+    A = matrix(ZZ,[first(collect(expv))-last(collect(expv)) for expv in exponents.(G)])
+    b = [ QQ(nu(last(collect(coeff))/first(collect(coeff)))) for coeff in coefficients.(G)]
```
fingolfin: There should be no need to "collect" the exponent resp. coefficient vectors, that just introduced extra allocations. Or am I missing something?
```suggestion
    A = matrix(ZZ,[first(expv)-last(expv) for expv in exponents.(G)])
    b = [ QQ(nu(last(coeff)/first(coeff))) for coeff in coefficients.(G)]
```

src/TropicalGeometry/variety_binomial.jl
```diff
@@ -0,0 +1,42 @@
...
+    ###
+    # Compute tropical variety multiplicity
+    ###
+    snfAdiag = elementary_divisors(A)
+    weight = abs(prod([m for m in snfAdiag if !iszero(m)]))
+
+    ###
+    # Constructing tropical variety set-theoretically
+    ###
+    A = QQMatrix(A)
```
fingolfin:
```suggestion
    A = matrix(QQ, A)
```
Note that this automatically makes your code type unstable, at least with the current Julia which doesn't deal well with a variable switching from one type to another. This can be solved by using a new variable name for the new matrix.

src/TropicalGeometry/variety_binomial.jl
```diff
@@ -0,0 +1,42 @@
...
+    L = transpose(kernel(A, side = :right))
+    can_solve, V = can_solve_with_solution(transpose(A),matrix(QQ,[b]),side=:left)
```
fingolfin: Better use `can_solve_with_solution_and_kernel` (so you solve only once, not twice)
```suggestion
    can_solve, V, L = can_solve_with_solution_and_kernel(transpose(A), matrix(QQ,[b]); side=:left)
```
(But please verify I got the right amount of transposing here ;-)

src/TropicalGeometry/variety_prime.jl
```diff
@@ -0,0 +1,135 @@
...
+    R = base_ring(I)
+    singularCommand = join(["ring r=0,("*join(string.(symbols(R)),",")*"),dp;",
+                            "ideal I = "*join(string.(gens(I)), ",")*";",
+                            "if (!defined(tropicalVariety)) { LIB \"tropical.lib\"; };",
+                            "fan TropI = tropicalVariety(I);",
+                            "string TropIString = string(TropI);"])
+    Singular.call_interpreter(singularCommand)
+    TropIString = Singular.lookup_library_symbol("Top", "TropIString")
+    Sigma = gfan_fan_string_to_oscar_complex(TropIString,convention(nu)==max,false)
+    TropI = compute_weights_and_construct_tropical_variety(Sigma,I,nu)
```
fingolfin, line 33: Hmm, do you really need to go through the interpreter? can't you do something like this instead?
```suggestion
    I_sing = singular_generators(I)
    TropI = Singular.LibTropical.tropicalVariety(I_sing)
```
OK this probably is missing some steps, `TropI` maybe needs additional processing -- but I hope the gist is clear?

fingolfin: Ah, sorry, I only now discovered you already tried this and submitted https://github.com/oscar-system/Singular.jl/issues/807

I'll comment there.

src/TropicalGeometry/variety_prime.jl
```diff
@@ -0,0 +1,135 @@
...
+#
+#  WARNING: assumes without test that `I` is prime
+#
+################################################################################
+
+function tropical_variety_prime(I::MPolyIdeal, nu::TropicalSemiringMap; weighted_polyhedral_complex_only::Bool=false)
+    # compute a reduced GB to check whether I is homogeneous
+    G = groebner_basis(I,complete_reduction=true)
+
+    if all(Oscar._is_homogeneous.(G))
```
fingolfin: This code first computes a `Vector{Bool}` via broadcast and then invokes `all` on it. In particular it allocates and can't do any short circuit.

Better then to use this:
```suggestion
    if all(Oscar._is_homogeneous, G)
```

Here is a toy benchmark demonstrating the difference:
```
julia> v = collect(1:100);

julia> @benchmark all(iseven.(v))
BenchmarkTools.Trial: 10000 samples with 209 evaluations.
 Range (min … max):  360.048 ns … 855.263 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     374.402 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   378.889 ns ±  17.451 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▅▆▇▆█▅▄▁▂▁
  ▁▂▅██████████▇▆▅▄▄▄▄▃▃▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  360 ns           Histogram: frequency by time          456 ns <

 Memory estimate: 160 bytes, allocs estimate: 4.

julia> @benchmark all(iseven, v)
BenchmarkTools.Trial: 10000 samples with 870 evaluations.
 Range (min … max):  135.345 ns … 245.210 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     140.566 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   140.761 ns ±   4.574 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▂▂          ▄█▂  ▁
  ▂▅██▅▅▇█▄▂▂▂▃█████▇█▇▃▂▃▃▃▃▃▄▃▃▂▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  135 ns           Histogram: frequency by time          156 ns <

```

YueRen: Good idea, thanks for the tip!

src/TropicalGeometry/variety_affine_linear.jl
```diff
@@ -0,0 +1,22 @@
...
+#
+#  WARNING: assumes without test that `gens(I)` has degree one
+#
+################################################################################
+
+function tropical_variety_affine_linear(I::MPolyIdeal,nu::TropicalSemiringMap; weighted_polyhedral_complex_only::Bool=false)
+    # compute a reduced GB to check whether I is homogeneous
+    G = groebner_basis(I,complete_reduction=true)
+
+    if all(Oscar._is_homogeneous.(G))
```
fingolfin: Also in other places :-)
```suggestion
    if all(Oscar._is_homogeneous,G)
```

src/TropicalGeometry/variety_principal.jl
```diff
@@ -0,0 +1,20 @@
...
+#
+#  WARNING: assumes without test that `gens(I)` has cardinality one
+#
+################################################################################
+
+function tropical_variety_principal(I::MPolyIdeal,nu::TropicalSemiringMap; weighted_polyhedral_complex_only::Bool=false)
+    ###
+    # Construct TropicalVariety from TropicalHypersurface
+    ###
+    g = first(gens(I))
```
fingolfin:
```suggestion
    g = gen(I,1)
```
or
```suggestion
    g = I[1]
```

src/TropicalGeometry/variety_zerodimensional.jl
```diff
@@ -0,0 +1,98 @@
...
+
+function tropical_variety_zerodimensional_eigenvalue(I::MPolyIdeal,nu::TropicalSemiringMap{QQField,QQFieldElem,<:Union{typeof(min),typeof(max)}})
+    k,(a,_) = number_field(I)
+    zk = maximal_order(k)
+    p = uniformizer(nu)
+    lp = [x[1] for x = prime_decomposition(zk,p)]
+    ma = representation_matrix(a)
+    mb = representation_matrix(k(lp[1].gen_two*lp[2].gen_two^2))
+    @assert iszero(ma*mb - mb*ma)
+    Qp = PadicField(p, 10)
```
fingolfin:
```suggestion
    Qp = padic_field(p; precision=10)
```

src/TropicalGeometry/variety_binomial.jl
```diff
@@ -0,0 +1,41 @@
...
+################################################################################
+
+function tropical_variety_binomial(I::MPolyIdeal,nu::TropicalSemiringMap; weighted_polyhedral_complex_only::Bool=false)
+    ###
+    # Construct matrix of exponent vector differences
+    # and vector of coefficient valuation differences
+    ###
+    G = gens(I)
+    A = matrix(ZZ,[first(expv)-last(expv) for expv in exponents.(G)])
+    b = [ QQ(nu(last(coeff)/first(coeff))) for coeff in coefficients.(G)]
```
fingolfin: Unfortunately it turns out my suggestion was wrong, `last` is not supported by these iterators. As a quick fix (with a result that still is IMHO a bit more legible than the original code), one can do this:
```suggestion
    A = matrix(ZZ,[first(expv)-last(expv) for expv in collect.(exponents.(G))])
    b = [ QQ(nu(last(coeff)/first(coeff))) for coeff in collect.(coefficients.(G))]
```
Luckily as long as the polynomials are binomials it shouldn't matter much.

fingolfin: I am going to apply this now to hopefully unbreak the tests.

fingolfin:
Unfortunately now has a merged conflict in `src/TropicalGeometry/variety.jl`

YueRen:
A book test (from my chapter) is failing due to a syntax change in `tropical_variety` (it used to return a list of `TropicalVariety`, not it simply returns one `TropicalVariety`), is it okay to ignore book tests or do we still want to adhere to them strictly?

joschmitt:
> it used to return a list of `TropicalVariety`, not it simply returns one `TropicalVariety`

This sounds like a breaking change to me which we cannot do in 1.*?

benlorenz:
> A book test (from my chapter) is failing due to a syntax change in `tropical_variety` (it used to return a list of `TropicalVariety`, not it simply returns one `TropicalVariety`), is it okay to ignore book tests or do we still want to adhere to them strictly?

Unless it is a clear bugfix, the input in the booktests should keep working for all 1.x versions. Only the printing is allowed to change.

Looking at the log there is also an [error](https://github.com/oscar-system/Oscar.jl/actions/runs/11883149213/job/33109510934?pr=4061#step:10:378) and not an output change:
```diff
  julia> TropV = tropical_variety(I,nu)
- 2-element Vector{TropicalVariety}:
-  Min tropical variety
-  Min tropical variety
+ ERROR: MethodError: no method matching tropical_variety_prime_singular(::MPolyIdeal{…}, ::TropicalSemiringMap{…}; weighted_polyhedral_complex_only::Bool)
+ 
+ Closest candidates are:
+   tropical_variety_prime_singular(::MPolyIdeal, !Matched::TropicalSemiringMap{QQField, ZZRingElem, <:Union{typeof(max), typeof(min)}}; weighted_polyhedral_complex_only)
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety_prime.jl:42
+   tropical_variety_prime_singular(::MPolyIdeal, !Matched::TropicalSemiringMap{QQField, Nothing, <:Union{typeof(max), typeof(min)}}; weighted_polyhedral_complex_only)
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety_prime.jl:23
+ 
+ Stacktrace:
+  [1] tropical_variety_prime(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…}; weighted_polyhedral_complex_only::Bool)
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety_prime.jl:18
+  [2] tropical_variety(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…}; weighted_polyhedral_complex_only::Bool)
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety.jl:209
+  [3] tropical_variety(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…})
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety.jl:189
+  [4] top-level scope
+    @ REPL[51]:1
+ Some type information was truncated. Use `show(err)` to see complete types.
```

YueRen:
> > it used to return a list of `TropicalVariety`, not it simply returns one `TropicalVariety`
> 
> This sounds like a breaking change to me which we cannot do in 1.*?

Yes, I agree that it is a breaking change.  So it will have to wait until 2.x then?

joschmitt:
> > > it used to return a list of `TropicalVariety`, not it simply returns one `TropicalVariety`
> > 
> > 
> > This sounds like a breaking change to me which we cannot do in 1.*?
> 
> Yes, I agree that it is a breaking change. So it will have to wait until 2.x then?

Sounds like it, but in the end this is not for me to decide. Could you provide a reasonable version of this pull request without the breaking changes? By "reasonable", I mean both in terms of functionality but also the work that needs to be done.

YueRen:
Yeah, the error is my bad.  I'll try to fix it and produce something mergable this week, whether it is for 1.x or 2.x.

YueRen:
There is still an [error](https://github.com/oscar-system/Oscar.jl/actions/runs/11896996464/job/33150243661?pr=4061#step:10:6926) with a book chapter, is that due to an output change?  The way `tropical_variety` was used in the book chapter should now trigger a custom error.  

 That being said, as this is a breaking change, I should probably add suggestions to the error message what the user should be using instead of `tropical_variety`.

joschmitt:
The actual error is in line 374 of the log (https://github.com/oscar-system/Oscar.jl/actions/runs/11896996464/job/33150243661?pr=4061#step:10:378):
```
  julia> TropV = tropical_variety(I,nu)
- 2-element Vector{TropicalVariety}:
-  Min tropical variety
-  Min tropical variety
+ ERROR: ArgumentError: Only primary ideals supported.
+ Stacktrace:
+  [1] macro expansion
+    @  AbstractAlgebra [inlined]
+  [2] tropical_variety(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…}; weighted_polyhedral_complex_only::Bool, check::Bool)
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety.jl:211
+  [3] tropical_variety(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…})
+    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety.jl:191
+  [4] top-level scope
+    @ REPL[51]:1
+ Some type information was truncated. Use `show(err)` to see complete types.
  
  julia> dim.(TropV)
- 2-element Vector{Int64}:
-  2
-  1
+ ERROR: UndefVarError: `TropV` not defined
+ Stacktrace:
+  [1] top-level scope
+    @ REPL[52]:1"""
```
`ERROR: ArgumentError: Only primary ideals supported.` doesn't sound good to me...

YueRen:
> The actual error is in line 374 of the log (https://github.com/oscar-system/Oscar.jl/actions/runs/11896996464/job/33150243661?pr=4061#step:10:378):
> 
> ```
>   julia> TropV = tropical_variety(I,nu)
> - 2-element Vector{TropicalVariety}:
> -  Min tropical variety
> -  Min tropical variety
> + ERROR: ArgumentError: Only primary ideals supported.
> + Stacktrace:
> +  [1] macro expansion
> +    @  AbstractAlgebra [inlined]
> +  [2] tropical_variety(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…}; weighted_polyhedral_complex_only::Bool, check::Bool)
> +    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety.jl:211
> +  [3] tropical_variety(I::MPolyIdeal{…}, nu::TropicalSemiringMap{…})
> +    @ Oscar ~/work/Oscar.jl/Oscar.jl/src/TropicalGeometry/variety.jl:191
> +  [4] top-level scope
> +    @ REPL[51]:1
> + Some type information was truncated. Use `show(err)` to see complete types.
>   
>   julia> dim.(TropV)
> - 2-element Vector{Int64}:
> -  2
> -  1
> + ERROR: UndefVarError: `TropV` not defined
> + Stacktrace:
> +  [1] top-level scope
> +    @ REPL[52]:1"""
> ```
> 
> `ERROR: ArgumentError: Only primary ideals supported.` doesn't sound good to me...

Right, I have fixed the error message.  It should now tell users to use `tropical_varieties` instead (or disable the check if they know what they are doing).

fingolfin:
I think the point @joschmitt wanted to make is this: would it be possible for you with reasonable effort (i.e. we are not asking you to spend days on this; hopefully this is something that can be done quickly) to change this PR so that it is *not* breaking the booktests? Presumably that would mean letting `tropical_variety` again return a vector with multiple results in that case. You could still also print a warning (printing changes are allowed) stating that this is deprecated and what should be called instead.

This way this PR could already be merged and the bulk of the changes would be in there. And then you could open a separate PR which probably would be very short and which turns the warning into an error again. That PR then could be added to a milestone for Oscar 2.0.

YueRen:
I see, that is what @joschmitt meant.  I can do that, but I'll take care of my two workshop PRs first.

YueRen:
Instead of waiting for 2.0, how about the I do the following to make it mergeable now:

1. The in- and output of the current command `tropical_variety` stay as they are (consistent with the book), however the user will be warned that the function will change in the future and is encouraged to use function `tropical_varieties` instead.
2. There will be a new function `tropical_varieties` that does what `tropical_variety` currently does and the function will stay in 2.0 and beyond.
3. There will be a new function `tropical_variety_new` (alternative name suggestions welcome) that does what `tropical_variety` will do from 2.0 onwards.

Just to re-iterate, the current version has two problems:

1. The functions `tropical_hypersurface`, `tropical_curve`, `tropical_linear_space` all return `TropicalHypersurface`, `TropicalCurve`, `TropicalLinearSpace` respectively.  The function `tropical_variety` returning a list of `TropicalVariety` is not very consistent.   It is also not consistent with gfan, Singular, and M2.
2. More importantly, `tropical_variety` is the only user-facing function for constructing the tropical variety from an ideal, and it computes saturation and primary decomposition first.  This is not always a good idea.  There are for example ideals that get more complicated post-saturation without changing their tropical variety.

YueRen:
Quick question: I've tried to make the PR pass the book test, but one test in my chapter is still failing.  Is there anyway to find out which jlcon is the offending one?  I've looked into the usual suspects and - apart from one minor discrepancy in the output addressed in https://github.com/oscar-system/Oscar.jl/pull/4061/commits/479a780a52da90616ae81de4cba5842a40753274 - they all seem good to me.

The new `tropical_variety` routine should behave like the old one by default.  When `skip_decomposition=true` is passed, it skips the primary decomposition and returns a single `TropicalVariety`.

test/book/specialized/holt-ren-tropical-geometry/variety.jlcon
```diff
@@ -8,7 +8,7 @@ julia> I = intersect(ideal([x+y+z+1,2*x+11*y+23*z+31]),
                      ideal([t^3*x*y*z-1]));
 
 julia> TropV = tropical_variety(I,nu)
-2-element Vector{TropicalVariety}:
+2-element Vector{TropicalVariety{typeof(min),true}}:
```
joschmitt: I think the problem is a missing whitespace:
```suggestion
2-element Vector{TropicalVariety{typeof(min), true}}:
```
See here: https://github.com/oscar-system/Oscar.jl/actions/runs/14125235041/job/39572795213?pr=4061#step:7:476

joschmitt:
> Quick question: I've tried to make the PR pass the book test, but one test in my chapter is still failing. Is there anyway to find out which jlcon is the offending one?

The way I know is to look through the CI log -- which is very long because of the gc information printed all the time. It helps to click `view raw log` and then search for "failed" or something.

test/book/specialized/holt-ren-tropical-geometry/variety.jlcon
YueRen: Thanks, that's fixed now!

YueRen:
@aaruni96 Could you take a look for me regarding why the tests fail?  For example: https://github.com/oscar-system/Oscar.jl/actions/runs/14127183363/job/39578783650?pr=4061

aaruni96:
These are failing because they run for too long, and then exceed the timeout value we set for tests (60 minutes for jobs running on RPTU runners, 80 minutes for jobs running on github runners).

I don't know why these tests don't finish in time, though. Maybe @benlorenz knows of a problem which is causing this again ?

YueRen:
I will rerun the book tests, in case this happens due to some fluke.  I know for sure that the book tests ran through previously, and I believe I only changed docs similar things afterwards.  Should the book tests run through, then I will rerun the rest also.

joschmitt:
If I see this correctly, this pull request is not breaking 1.* anymore, so I remove those labels. I add the triage label to find somebody who can review it.

fieker:
@YueRen will resolve conflicts and ask @micjoswig for a final go ahead

fingolfin:
The conflict has been resolved. Is this ready for a review then, @YueRen ?

YueRen:
@fingolfin I've double checked everything and I'm happy to merge this PR.

Just to re-iterate:  The command `tropical_variety` has two new optional parameters `skip_saturation::Bool=false` and `skip_decomposition::Bool=false`.  It works as before by default, returning a list of `TropicalVariety`, hence all book tests are passing.  However, setting `skip_decomposition=true` will skip a primary decomposition and return a single `TropicalVariety`, which is closer to the behaviour of that of other software.

YueRen:
There were some failing test were that are unrelated to my code. I've rebased the PR to the latest master, hopefully that fixes the issue.

benlorenz:
Everything related to julia nightly is currently expected to fail.

lgoettgens:
This now has a conflict due to https://github.com/oscar-system/Oscar.jl/pull/4838. @YueRen could you please give resolving this a try?

YueRen:
I've rebased the changes now and the only book tests that are failing are pertaining to areas that this PR is not touching.  Is there anything else left to do before this PR can be merged?

joschmitt:
> I've rebased the changes now and the only book tests that are failing are pertaining to areas that this PR is not touching. Is there anything else left to do before this PR can be merged?

Someone needs to review it :)

YueRen:
@micjoswig Would you be able to review the PR?  Here are the major changes:

* Dispatching to various specialized tropicalization algorithms was moved from the function `tropical_variety` to a new function `tropical_variety_dispatch`.

*  The function `tropical_variety` is now only responsible for some pre-processing and calling `tropical_variety_dispatch`.  Depending on the optional parameters, it may call `tropical_variety_dispatch` once or multiple times (e.g., once per prime factor).  It does the latter by default, as in the OSCAR book.

* All specialized tropicalization algorithms were moved to their own `variety_*.jl` file.  This is because `variety.jl` became too difficult to navigate and it will only get worse.  

* There is a new specialized tropicalization algorithm `tropical_variety_prime_singular` in `variety_prime.jl` (because it pertains prime ideals).  It essentially calls the `tropicalVariety` function in Singular via `Singular.call_interpreter` and uses string manipulation to reconstruct the resulting polyhedral complex in Oscar.  This is of course far from ideal, but wrapping `tropicalVariety` properly in `Singular.jl` would require some work as it returns a pointer to a Gfan fan.  I consider this a temporary bandaid until Oscar gets its own native tropicalization routines.

micjoswig:
> > I've rebased the changes now and the only book tests that are failing are pertaining to areas that this PR is not touching. Is there anything else left to do before this PR can be merged?
> 
> Someone needs to review it :)

@ooinaruhugh : you had a closer look.  Happy?

ooinaruhugh:
> > > I've rebased the changes now and the only book tests that are failing are pertaining to areas that this PR is not touching. Is there anything else left to do before this PR can be merged?
> > 
> > 
> > Someone needs to review it :)
> 
> @ooinaruhugh : you had a closer look. Happy?

@micjoswig @YueRen I tried some examples and looks good.

ooinaruhugh (approved):

lgoettgens:
This PR had a `release notes: to be added` label. I just replaced this be a `release notes: use title` and updated the title.
If somewants wants to have some more detailed release notes for this PR, then please change the label back and write approx two sentences in a comment below, and then I'll make sure they end up in the changelog.

fingolfin:
Seems fine to me, perhaps @YueRen has another wish?

YueRen:
Sounds good, thanks @lgoettgens @fingolfin

PR #4062, opened 2024-08-30
[FTheoryTools] Improve computation of refined Tate fiber types

apturner:
This adds Mohab's proposed calculation method, except in the case of Type I_0^* fibers, addressing the issue that models over concrete bases are often too complicated and seem "accidentally nonsplit". However, models over arbitrary bases are typically too simple and appear "accidentally split", so the code currently checks if the base is concrete are not and then decides whether to use the old or the new version of the calculation. This is not ideal.

Tests should also be added to more thoroughly check this.

@HereAround, @emikelsons

### DIFF
--- a/experimental/FTheoryTools/src/HypersurfaceModels/attributes.jl
+++ b/experimental/FTheoryTools/src/HypersurfaceModels/attributes.jl
@@ -65,7 +65,7 @@ hypersurface_equation_parametrization(h::HypersurfaceModel) = h.hypersurface_equ
     weierstrass_model(h::HypersurfaceModel)
 
 Return the Weierstrass model corresponding to the
-hypersurface model, provided that the latter is known.
+hypersurface model, provided that the former is known.
 
 ```jldoctest
 julia> t = literature_model(14)
--- a/experimental/FTheoryTools/src/HypersurfaceModels/attributes.jl
+++ b/experimental/FTheoryTools/src/HypersurfaceModels/attributes.jl
@@ -176,6 +176,6 @@ at each locus. Also the refined Tate fiber type is returned.
 """
 @attr Vector{<:Tuple{<:MPolyIdeal{<:MPolyRingElem}, Tuple{Int64, Int64, Int64}, String}} function singular_loci(h::HypersurfaceModel)
   @req base_space(h) isa NormalToricVariety "Singular loci currently only supported for toric varieties as base space"
-  @req has_attribute(h, :weierstrass_model) "No corresponding Weierstrass model is known"
-  return singular_loci(weierstrass_model(h))
+  @req has_attribute(h, :weierstrass_model) || has_attribute(h, :global_tate_model) "No corresponding Weierstrass model or global Tate model is known"
+  return has_attribute(h, :weierstrass_model) ? singular_loci(weierstrass_model(h)) : singular_loci(global_tate_model(h))
 end
--- a/experimental/FTheoryTools/src/auxiliary.jl
+++ b/experimental/FTheoryTools/src/auxiliary.jl
@@ -149,7 +149,7 @@ _count_factors(poly::QQMPolyRingElem) = mapreduce(p -> p[end], +, absolute_prima
 
 _string_from_factor_count(poly::QQMPolyRingElem, string_list::Vector{String}) = string_list[_count_factors(poly)]
 
-function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, Int64, Int64}, concrete_base::Bool) where {T<:MPolyRingElem}
+function _kodaira_type(id::MPolyIdeal{<:MPolyRingElem}, ords::Tuple{Int64, Int64, Int64}, w::WeierstrassModel; rand_seed::Union{Int64, Nothing} = nothing)
   f_ord = ords[1]
   g_ord = ords[2]
   d_ord = ords[3]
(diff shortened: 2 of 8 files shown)


experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -170,51 +170,115 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
-
-      monodromy_poly = _psi^2 - g_quotient
-      kod_type = _string_from_factor_count(monodromy_poly, ["Non-split IV^*", "Split IV^*"])
+    if concrete_base
+      # Over concrete bases, we reduce to only two variables
+      # so that the computation is faster
+      num_gens = length(gens(parent(f)))
+      first_coord_inds = collect(1:num_gens - 2)
+      last_coord_inds = collect(3:num_gens)
+      rand_ints_ab2 = rand(-100:100, num_gens - 2)
```
HereAround: Are we guaranteed that the method always works? Or could these random numbers sometimes fail us? If so, we would have random failures, which is likely not good. 

(There were discussions elsewhere regarding fixed seeds etc., so as to make tests/failures reproducible. If random numbers are truly needed, then we should investigate more.)

experimental/FTheoryTools/src/auxiliary.jl
```diff
@@ -170,51 +170,115 @@ function _kodaira_type(id::MPolyIdeal{T}, f::T, g::T, d::T, ords::Tuple{Int64, I
...
+      w_last_2 = evaluate(forget_decoration(gens(id)[1]), first_coord_inds, rand_ints_ab2)
+      f_last_2 = evaluate(forget_decoration(f), first_coord_inds, rand_ints_ab2)
+      g_last_2 = evaluate(forget_decoration(g), first_coord_inds, rand_ints_ab2)
+      d_last_2 = evaluate(forget_decoration(d), first_coord_inds, rand_ints_ab2)
+
+      # Check monodromy conditions for remaining cases
+      if f_ord == 0 && g_ord == 0
+        q_first = quotient(ideal([9 * g_first_2, w_first_2]), ideal([2 * f_first_2, w_first_2]))
+        q_last = quotient(ideal([9 * g_last_2, w_last_2]), ideal([2 * f_last_2, w_last_2]))
+
```
HereAround: Just for my curiosity: Why is there a line break here (and in similar placed below)? I am not sure if it enhances readability.

HereAround:
Thank you for working on this @apturner !

Generally speaking, this looks great. However, I believe all of the following points should be addressed before we merge this PR:
1. Could the random numbers (see above) lead to random failures? If yes, then we need to make this reproducible (e.g. by remembering the seed). This was discussed elsewhere. I do not recall at the top of my head where.
2. Let us add tests to this PR. If I am not mistaken, @emikelsons brought forward examples, where the current implementation yields the wrong Kodaira type. I strongly suggest to add those examples as tests (be it CI-tests or doc-tests). Thereby, we can truly appreciate this improvement and we have non-trivial tests.
3. I recall that we wanted to add a text to the documentation, in which we thank Mohab. However, this is currently missing in this PR.

HereAround (changes requested):
See above.

apturner:
1. You are correct that the random numbers could result in a random failure. Are you proposing that we always use the same seed, or that we have the ability to provide a seed? If we fix a seed, that's just as likely to randomly fail on any given model as any other seed is. In that sense, the ability to rerun the code several times and use it as a probabilistic method is actually something of a boon. Let me know your thoughts.
2. I will add tests, and the toric hypersurface models provide perfect examples, so I'll probably run through those.
3. I can add the acknowledgement to Mohad, good call.

HereAround:
> 1. You are correct that the random numbers could result in a random failure. Are you proposing that we always use the same seed, or that we have the ability to provide a seed? If we fix a seed, that's just as likely to randomly fail on any given model as any other seed is. In that sense, the ability to rerun the code several times and use it as a probabilistic method is actually something of a boon. Let me know your thoughts.
> 

Let me ping in @benlorenz. Maybe @lgoettgens knows as well.

@apturner wants to add a function, the outcome of which depends on random numbers. In the worst case scenario, this leads to random failures. I am certain similar situations did arise already and have been discussed. Suggestions?

experimental/FTheoryTools/src/auxiliary.jl
apturner: Only for readability. I generally prefer to break up the logical steps of the calculation rather than have a solid block of code. For some of these cases, the first step is many lines, and for consistency I line break even the shorter cases

experimental/FTheoryTools/src/auxiliary.jl
fingolfin: Discussed this in triage.

If you really need randomness here, I would strongly suggest to add a comment that explains the background.

Also: are the given ranges really always sufficient? If so this should be justified. If not then one should at least describe what might go wrong if the random value is "insufficient" (whatever that means). You could also add checks or loop to try multiple random values until some criterion proves it is correct; or until a consistent result value has been produced several times for different random choices...

More was said and discussed with @HereAround but that's the gist.

HereAround: @apturner @emikelsons and I just discussed more on this. Here is the summary:

What this function is actually deciding is if certain ideals are radical or not. However, for the geometries encountered in FTheory, these computations are too hard/time consuming. The random numbers serve as shortcut as follows:

1. Keep exactly two variable (in the polynomial generators of the ideals in question), and evaluate all remaining variables at integer values. Then check whether the result is radical or not.
2. Currently, this step is executed twice - once by keeping the very first two variables as symbols (and evaluating all other variables to random integer values), and once by keeping the very last two variables as symbols.

We just agreed on the following:
* Instead of always keeping the first two/last two variables as symbols, do so for random pairs of variables.
* Repeat the evaluation and subsequent radical check five times (instead of currently only two times).
* Add tests, to gauge if this algorithm succeeds.
* The current implementation corrects for a tendency of this algorithm to favor non-split over split. This could be made stricter. But then, we need to tests if such a stricter test/algorithm still yields correct answers, or tends to fail.
* @apturner and @emikelsons seemed to favor the choice of a seed. So I understand @apturner will work this in as an optional argument to this method. Thereby, the tests should become reproducible.
* We were also thinking of using `put_over_concrete_base(P3, some_model_over_a_non_concrete_base)`. Then we can run the new algorithm also for models that are defined over non-concrete bases. However, there is unfortunately always a chance that this step would introduce a gauge enhancement, and so the results computed from this method would disagree with the "generic model". This is to be played with/tested. Our feeling is that the projective space P3 minimized this potential of failure though.

I believe this is all the points that we came to think of and @apturner will work on this. Let us know once this is ready to be reviewed again. Thank you for working on this!

HereAround:
I just realized that my response/status of this PR is somewhat hidden, as I added it to an above conversation. So, for convenience, here is the current (Friday, September 6) status:

@apturner @emikelsons and I just discussed more on this. Here is the summary:

What this function is actually deciding is if certain ideals are radical or not. However, for the geometries encountered in FTheory, these computations are too hard/time consuming. The random numbers serve as shortcut as follows:

1. Keep exactly two variable (in the polynomial generators of the ideals in question), and evaluate all remaining variables at integer values. Then check whether the result is radical or not.
2. Currently, this step is executed twice - once by keeping the very first two variables as symbols (and evaluating all other variables to random integer values), and once by keeping the very last two variables as symbols.

We just agreed on the following:
* Instead of always keeping the first two/last two variables as symbols, do so for random pairs of variables.
* Repeat the evaluation and subsequent radical check five times (instead of currently only two times).
* Add tests, to gauge if this algorithm succeeds.
* The current implementation corrects for a tendency of this algorithm to favor non-split over split. This could be made stricter. But then, we need to tests if such a stricter test/algorithm still yields correct answers, or tends to fail.
* @apturner and @emikelsons seemed to favor the choice of a seed. So I understand @apturner will work this in as an optional argument to this method. Thereby, the tests should become reproducible.
* We were also thinking of using `put_over_concrete_base(P3, some_model_over_a_non_concrete_base)`. Then we can run the new algorithm also for models that are defined over non-concrete bases. However, there is unfortunately always a chance that this step would introduce a gauge enhancement, and so the results computed from this method would disagree with the "generic model". This is to be played with/tested. Our feeling is that the projective space P3 minimized this potential of failure though.

I believe this is all the points that we came to think of and @apturner will work on this. Let us know once this is ready to be reviewed again. Thank you for working on this!

apturner:
I believe I have addressed everything in the recent commit.

1. I have added further comments to explain the use of random numbers.
2. I now run the random process five times and compare the results.
3. Each random run now picks a random two variables to keep unevaluated.
4. I still default to "split" if there is disagreement in the results, to counteract the tendency for the method to fail toward "non-split".
5. I put all models over a concrete base (projective space, if it's not already over a concrete base) so that the same method can always be used.
6. I have included an optional seed argument for debugging purposes.
7. With the change to always do the computation over concrete bases, the existing tests all test the new code now, but I have also included new tests that try the code out on all of the Weierstrass models from the F-theory on All Toric Hypersurface models.
8. I added the acknowledgement of Mohab for his help devising this algorithm.
9. I added some minor fixes to the `put_over_concrete_base` function.
10. In the process of working on this commit, I discovered errors in the existing tests of Tate models over arbitrary bases, which I have fixed in this commit.

@HereAround

apturner:
Currently, the optional keyword argument to provide a seed is only available in the internal function `_kodaira_type`, which is normally only called by other user-facing functions. Testing this function using a given seed can be performed as in the following example:
```julia
julia> B3 = projective_space(NormalToricVariety, 3)
Normal toric variety

julia> Kbar = anticanonical_divisor(B3)
Torus-invariant, non-prime divisor on a normal toric variety

julia> foah16_B3_weier = literature_model(arxiv_id = "1408.4808", equation = "3.203", type = "weierstrass", base_space = B3, defining_classes = Dict("s7" => Kbar, "s9" => Kbar), completeness_check = false)
Construction over concrete base may lead to singularity enhancement. Consider computing singular_loci. However, this may take time!

Weierstrass model over a concrete base -- F-theory weierstrass model dual to hypersurface model with fiber ambient space F_16 based on arXiv paper 1408.4808 Eq. (3.203)

julia> sl = singular_loci(foah16_B3_weier)
4-element Vector{Tuple{MPolyIdeal{<:MPolyRingElem}, Tuple{Int64, Int64, Int64}, String}}:
 (Ideal with 1 generator, (0, 0, 1), "I_1")
 (Ideal with 1 generator, (0, 0, 3), "Split I_3")
 (Ideal with 1 generator, (0, 0, 3), "Split I_3")
 (Ideal with 1 generator, (0, 0, 3), "Split I_3")

julia> Oscar._kodaira_type(sl[2][1], sl[2][2], foah16_B3_weier, rand_seed = 1234)
"Split I_3"
```

experimental/FTheoryTools/test/literature_models.jl
```diff
@@ -783,4 +783,44 @@ foah16_B3_weier = literature_model(arxiv_id = "1408.4808", equation = "3.203", t
...
+  @test length(singular_loci(foah7_B3_weier)) == 1
+  @test length(singular_loci(foah8_B3_weier)) == 3
+  @test length(singular_loci(foah9_B3_weier)) == 2
+  @test length(singular_loci(foah10_B3_weier)) == 3
+  @test length(singular_loci(foah11_B3_weier)) == 3
+  @test length(singular_loci(foah12_B3_weier)) == 3
+  @test length(singular_loci(foah13_B3_weier)) == 4
+  @test length(singular_loci(foah14_B3_weier)) == 4
+  @test length(singular_loci(foah15_B3_weier)) == 5
+  @test length(singular_loci(foah16_B3_weier)) == 4
```
HereAround, line 801: For my curiosity - how long do you think those tests take now?

I believe we even raise a warning somewhere about `singular_loci` being potentially super slow. So I just want to gauge how much this slows down the OSCAR tests.

The tests are green - at least for now. That of course is a good indication.

experimental/FTheoryTools/test/tate_models.jl
```diff
@@ -128,39 +128,39 @@ istar0_s_auxiliary_base_ring, (a1pp, a2pp, a3pp, a4pp, a6pp, vp, wp, mp) = QQ["a
 tate_auxiliary_base_ring, (a1p, a2p, a3p, a4p, a6p, v, w) = QQ["a1p", "a2p", "a3p", "a4p", "a6p", "v", "w"];
 
 # construct Tate models over arbitrary base
-t_istar0_s = global_tate_model(istar0_s_auxiliary_base_ring, [1 2 3 4 6 0 2; -1 -2 -2 -3 -4 1 -1], 3, [a1pp * (vp^2 + wp)^1, mp * (vp^2 + wp)^1 + a2pp * (vp^2 + wp)^2, a3pp * (vp^2 + wp)^2, mp^2 * (vp^2 + wp)^2 + a4pp * (vp^2 + wp)^3, a6pp * (vp^2 + wp)^4]);
```
HereAround, line 131: Why have the gradings changed? (At least it looks so to me right now.)

experimental/FTheoryTools/test/tate_models.jl
```diff
@@ -177,9 +177,9 @@ t_nm = global_tate_model(tate_auxiliary_base_ring, [1 2 3 4 6 0; -1 -2 -3 -4 -6
 end
 
 @testset "Error messages in global Tate models over generic base space" begin
-  @test_throws ArgumentError global_tate_model(tate_auxiliary_base_ring, [1 2 3 4 6 0; 0 -1 -2 -3 -5 1], 3, [a1p])
-  @test_throws ArgumentError global_tate_model(tate_auxiliary_base_ring, [1 2 3 4 6 0; 0 -1 -2 -3 -5 1], 3, [a1p * v^0, a2p * v^1, a3p * v^2, a4p * v^3, sec_a6])
-  @test_throws ArgumentError global_tate_model(tate_auxiliary_base_ring, [1 2 3 4 6 0; 0 -1 -2 -3 -5 1], -1, [a1p * v^0, a2p * v^1, a3p * v^2, a4p * v^3, a6p * v^5])
+  @test_throws ArgumentError global_tate_model(tate_auxiliary_base_ring, [1 2 3 4 6 0 0; 0 -1 -2 -3 -5 1 2], 3, [a1p])
```
HereAround, line 180: Here the gradings also did change. Why?

HereAround (approved):
Thank you for working on this delicate matter @apturner . This looks very good now!

I left a few minor questions (see above).

For which models will we now be able to work out the (refined) singularity type? For instance, will it work for a hypersurface model without known associated Weierstrass/Tate model? What about hypersurface models with an associated Tate model? Do we maybe want to add more req statements to catch cases which are currently not (yet) covered?

Another question: Will the "identify the singularity type" functionality benefit from `put_over_concrete_base` working for hypersurface models also?

HereAround:
(I will not merge this PR just yet, because I am curious about the responses of @apturner . Other than that, there is no reason - as I see it - to hold this PR.)

HereAround:
(Another question: Are you using https://github.com/oscar-system/Oscar.jl/pull/4075 in this PR? I assume not/cannot see it. But just wanted to double check.)

experimental/FTheoryTools/test/tate_models.jl
apturner: All of the gradings in these tests were incorrect in multiple ways:

1. The number of columns in the grading matrices did not match the number of coordinates of the base. This also indicates an "error" of the functions that construct arbitrary bases, as they should probably require that this be the case. I didn't fix that issue in this PR, but I did update all the gradings in these tests.
2. Even setting that aside, the gradings were not updated when we changed the tests to use gauge divisor `v^2 + w = 0` rather than just `w = 0`. Where previously a given variable had grade `n` under `[w]`, now it really should have grade `2 n` under `[v]`, and `w` itself should have grade `2` under `[v]`. I've updated the gradings to make this the case.

experimental/FTheoryTools/test/tate_models.jl
apturner: See above.

apturner:
In regards to your top-level questions:

1. This algorithm only works for Weierstrass models. By extension, we can use it for any Tate model, because we know how to construct the Weierstrass model from a Tate model. Thus, we can use this algorithm for any model for which we know a Tate or Weierstrass model. If we have a hypersurface model for which we don't know either, then we cannot use this method. The use of `put_over_concrete_base` does not improve this fact, unfortunately.
2. I am not using #4075 currently. The function `put_over_concrete_base` uses `cox_ring` throughout, which originally led me to bring this up to you, but I have not changed it there because the use of `degree` in the same function (instead of `degrees` or some other related function, I'm not exactly sure) does not generalize to non-toric cases properly, from my experimentation. So instead, I included a `@req` call that requires the concrete base be toric, for the time being.

experimental/FTheoryTools/test/literature_models.jl
apturner: On my laptop, running all of the tests for FTheoryTools, including these new ones, takes 6.5 minutes.

HereAround:
> Currently, the optional keyword argument to provide a seed is only available in the internal function `_kodaira_type`, which is normally only called by other user-facing functions. Testing this function using a given seed can be performed as in the following example:
> 
> ```julia
> julia> B3 = projective_space(NormalToricVariety, 3)
> Normal toric variety
> 
> julia> Kbar = anticanonical_divisor(B3)
> Torus-invariant, non-prime divisor on a normal toric variety
> 
> julia> foah16_B3_weier = literature_model(arxiv_id = "1408.4808", equation = "3.203", type = "weierstrass", base_space = B3, defining_classes = Dict("s7" => Kbar, "s9" => Kbar), completeness_check = false)
> Construction over concrete base may lead to singularity enhancement. Consider computing singular_loci. However, this may take time!
> 
> Weierstrass model over a concrete base -- F-theory weierstrass model dual to hypersurface model with fiber ambient space F_16 based on arXiv paper 1408.4808 Eq. (3.203)
> 
> julia> sl = singular_loci(foah16_B3_weier)
> 4-element Vector{Tuple{MPolyIdeal{<:MPolyRingElem}, Tuple{Int64, Int64, Int64}, String}}:
>  (Ideal with 1 generator, (0, 0, 1), "I_1")
>  (Ideal with 1 generator, (0, 0, 3), "Split I_3")
>  (Ideal with 1 generator, (0, 0, 3), "Split I_3")
>  (Ideal with 1 generator, (0, 0, 3), "Split I_3")
> 
> julia> Oscar._kodaira_type(sl[2][1], sl[2][2], foah16_B3_weier, rand_seed = 1234)
> "Split I_3"
> ```

I wonder if we might want to add this as comment? Or even make it part of the documentation?

HereAround:
> In regards to your top-level questions:
> 
>     1. This algorithm only works for Weierstrass models. By extension, we can use it for any Tate model, because we know how to construct the Weierstrass model from a Tate model. Thus, we can use this algorithm for any model for which we know a Tate or Weierstrass model. If we have a hypersurface model for which we don't know either, then we cannot use this method. The use of `put_over_concrete_base` does not improve this fact, unfortunately.
 
Is this something to be mentioned in the documentation of `singular_loci` (maybe I overlooked it)? Maybe it is not too hard to extend to Tate models and hypersurface models with known Weierstrass and/or Tate model. To be reserved for another PR?

>
>     2. I am not using [[ToricVariety] Support coordinate_ring as reference to cox_ring #4075](https://github.com/oscar-system/Oscar.jl/pull/4075) currently. The function `put_over_concrete_base` uses `cox_ring` throughout, which originally led me to bring this up to you, but I have not changed it there because the use of `degree` in the same function (instead of `degrees` or some other related function, I'm not exactly sure) does not generalize to non-toric cases properly, from my experimentation. So instead, I included a `@req` call that requires the concrete base be toric, for the time being.

Ok.

experimental/FTheoryTools/test/tate_models.jl
HereAround: I see. Thx!

HereAround:
@apturner It seems there are quite a few other points that you found need attention. Maybe you could add those to https://github.com/oscar-system/Oscar.jl/issues/2699, so we do not forget about them after merging this PR?

experimental/FTheoryTools/test/literature_models.jl
HereAround: That sounds good to me.

apturner:
> Is this something to be mentioned in the documentation of `singular_loci` (maybe I overlooked it)? Maybe it is not too hard to extend to Tate models and hypersurface models with known Weierstrass and/or Tate model. To be reserved for another PR?

This already works for any model that has a known Weierstrass model, but doesn't automatically work if it has a known Tate model but no directly known Weierstrass model. That is a simple fix I can add here.

apturner:
> @apturner It seems there are quite a few other points that you found need attention. Maybe you could add those to #2699, so we do not forget about them after merging this PR?

Yes, I can add all the relevant points there.

experimental/FTheoryTools/src/HypersurfaceModels/attributes.jl
```diff
@@ -176,6 +176,6 @@ at each locus. Also the refined Tate fiber type is returned.
 """
 @attr Vector{<:Tuple{<:MPolyIdeal{<:MPolyRingElem}, Tuple{Int64, Int64, Int64}, String}} function singular_loci(h::HypersurfaceModel)
   @req base_space(h) isa NormalToricVariety "Singular loci currently only supported for toric varieties as base space"
-  @req has_attribute(h, :weierstrass_model) "No corresponding Weierstrass model is known"
-  return singular_loci(weierstrass_model(h))
+  @req has_attribute(h, :weierstrass_model) || has_attribute(h, :global_tate_model) "No corresponding Weierstrass model or global Tate model is known"
+  return has_attribute(h, :weierstrass_model) ? singular_loci(weierstrass_model(h)) : singular_loci(global_tate_model(h))
```
apturner, line 180: This change allows for hypersurface models with known Weierstrass or global Tate models to utilize `singular_loci`. I think this is all ready to merge, once the tests pass again.

@HereAround

HereAround: Thank you @apturner . Awesome!

HereAround (approved):

PR #4065, opened 2024-09-02
LieAlgebras: Serialization for modules

lgoettgens:
Some context for the review:
Most of the module stuff is pretty straightforward to serialize.
If I want to (de)serialize attributes, for a lot of them, I have some problems: They encode how the module was created, e.g. as a tensor product of some other modules. And this includes some not-serializable data, e.g. the pure tensor function that takes a tuple of elements of the tensor factors and returns the corresponding pure tensor in the module. To be able to re-create these things, I just de-serialize the tensor factors and call `tensor_product` in the load function with these, instead of de-serializing the whole module, as this sets up all necessary attributes and properties.
However, this means that I cannot use the usual attribute saving/loading, so I added the functions `{save,load}_construction_data` that deal with that. There are currently some hacks to detect in there if `with_attrs` was set in the surrounding `save/load` call. It would be great if we could refactor this at some time in the future to use a better interface to this, maybe something along the lines of https://github.com/oscar-system/Oscar.jl/pull/4068#issuecomment-2326128105. But I think that we can leave that to some future point.

### DIFF
--- a/experimental/LieAlgebras/src/serialization.jl
+++ b/experimental/LieAlgebras/src/serialization.jl
@@ -136,7 +136,7 @@ end
 #
 ###############################################################################
 
-@register_serialization_type LieAlgebraModule uses_id
+@register_serialization_type LieAlgebraModule uses_id [:_dummy_attribute]
 
 function save_object(s::SerializerState, V::LieAlgebraModule)
   save_data_dict(s) do
@@ -144,11 +144,12 @@ function save_object(s::SerializerState, V::LieAlgebraModule)
     save_object(s, dim(V), :dim)
     save_object(s, transformation_matrices(V), :transformation_matrices)
     save_object(s, symbols(V), :symbols)
+    save_construction_data(s, V)
     save_attrs(s, V)
   end
 end
 
-function load_object(s::DeserializerState, ::Type{LieAlgebraModule})
+function load_object(s::DeserializerState, T::Type{<:LieAlgebraModule})
   L = load_typed_object(s, :lie_algebra)
   R = coefficient_ring(L)
   dim = load_object(s, Int, :dim)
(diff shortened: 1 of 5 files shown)


experimental/LieAlgebras/src/serialization.jl
```diff
@@ -130,6 +130,116 @@ function load_type_params(s::DeserializerState, ::Type{<:LieAlgebraElem})
   return load_typed_object(s)
 end
 
+###############################################################################
+#
+#   Lie algebra modules
+#
+###############################################################################
+
+@register_serialization_type LieAlgebraModule uses_id [:_dummy_attribute]
```
antonydellavecchia: for now maybe just a more meaningful name?
```suggestion
@register_serialization_type LieAlgebraModule uses_id [:__serialize_construction]
```

experimental/LieAlgebras/src/serialization.jl
```diff
@@ -130,6 +130,116 @@ function load_type_params(s::DeserializerState, ::Type{<:LieAlgebraElem})
...
+  V = load_construction_data(s, T)
+  if isnothing(V)
+    V = abstract_module(L, dim, transformation_matrices, symbs; check=false)
+  end
+  load_attrs(s, V)
+  return V
+end
+
+function save_construction_data(s::SerializerState, V::LieAlgebraModule)
+  :_dummy_attribute in Oscar.attrs_list(s, typeof(V)) || return nothing # attribute saving disabled
```
antonydellavecchia:
```suggestion
   :__serialize_construction in Oscar.attrs_list(s, typeof(V)) || return nothing # attribute saving disabled
```

antonydellavecchia (approved):
works for now

PR #4068, opened 2024-09-03
small fix for loading without attrs

antonydellavecchia:
can load without attributes if desired

### DIFF
--- a/src/Serialization/serializers.jl
+++ b/src/Serialization/serializers.jl
@@ -249,7 +249,6 @@ function deserializer_open(
   return T(DeserializerState(obj, nothing, nothing, type_attr_map))
 end
 
-const state_types = Union{SerializerState, DeserializerState}
 
 function attrs_list(s::SerializerState, T::Type) 
   return get(s.type_attr_map, encode_type(T), Symbol[])
--- a/src/Serialization/serializers.jl
+++ b/src/Serialization/serializers.jl
@@ -223,33 +223,24 @@ function serializer_open(
   return T(SerializerState(true, UUID[], io, nothing, type_attr_map))
 end
 
-function deserializer_open(
-  io::IO,
-  T::Type{JSONSerializer},
-  type_attr_map::S) where S <:Union{Dict{String, Vector{Symbol}}, Nothing}
-
+function deserializer_open(io::IO, T::Type{JSONSerializer}, with_attrs::Bool)
   obj = JSON3.read(io)
   refs = nothing
   if haskey(obj, refs_key)
     refs = obj[refs_key]
   end
   
-  return T(DeserializerState(obj, nothing, refs, type_attr_map))
+  return T(DeserializerState(obj, nothing, refs, with_attrs))
 end
 
-function deserializer_open(
-  io::IO,
-  T::Type{IPCSerializer},
-  type_attr_map::S) where S <:Union{Dict{String, Vector{Symbol}}, Nothing}
+function deserializer_open(io::IO, T::Type{IPCSerializer}, with_attrs::Bool) 
   # Using a JSON3.Object from JSON3 version 1.13.2 causes
   # @everywhere using Oscar
   # to hang. So we use a Dict here for now.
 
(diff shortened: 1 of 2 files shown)


src/Serialization/serializers.jl
```diff
@@ -143,7 +143,7 @@ mutable struct DeserializerState
   obj::Union{Dict{Symbol, Any}, Vector, JSON3.Object, JSON3.Array, BasicTypeUnion}
   key::Union{Symbol, Int, Nothing}
   refs::Union{Dict{Symbol, Any}, JSON3.Object, Nothing}
-  type_attr_map::Dict{String, Vector{Symbol}}
+  type_attr_map::Bool
```
antonydellavecchia:
```suggestion
  type_attr::Bool
```

src/Serialization/serializers.jl
```diff
@@ -143,7 +143,7 @@ mutable struct DeserializerState
   obj::Union{Dict{Symbol, Any}, Vector, JSON3.Object, JSON3.Array, BasicTypeUnion}
   key::Union{Symbol, Int, Nothing}
   refs::Union{Dict{Symbol, Any}, JSON3.Object, Nothing}
-  type_attr_map::Dict{String, Vector{Symbol}}
+  type_attr::Bool
```
antonydellavecchia:
```suggestion
  with_attrs::Bool
```

lgoettgens:
The changes look great!
I think it would be great if we could have some dedicated accessor to `DeserializerState.with_attrs`. Furthermore, I would then like it if we would have the same accessor for `SerializerState` as well. This would enable the following refactoring of `SerializerState`: Don't save the `type_attr_map` in `SerializerState`, instead just `with_attrs::Bool`. The only place where this then needs a change is `attrs_list`, which would return `get(Oscar.type_attr_map, encode_type(T), Symbol[])` if `s.with_attrs` and `Symbol[]` otherwise. So there would be no need to pass the type_attr_map dict down through all the functions. What do you think about this?

src/Serialization/serializers.jl
```diff
@@ -251,7 +251,7 @@ end
 
 const state_types = Union{SerializerState, DeserializerState}
```
lgoettgens:
```suggestion
```
This union is no longer used anywhere

antonydellavecchia:
> The changes look great! I think it would be great if we could have some dedicated accessor to `DeserializerState.with_attrs`. Furthermore, I would then like it if we would have the same accessor for `SerializerState` as well. This would enable the following refactoring of `SerializerState`: Don't save the `type_attr_map` in `SerializerState`, instead just `with_attrs::Bool`. The only place where this then needs a change is `attrs_list`, which would return `get(Oscar.type_attr_map, encode_type(T), Symbol[])` if `s.with_attrs` and `Symbol[]` otherwise. So there would be no need to pass the type_attr_map dict down through all the functions. What do you think about this?

the reason it was like this, was to at some point in the future allow overrides from the user, possible something to discuss at the meeting in kaiserslautern

antonydellavecchia:
> > The changes look great! I think it would be great if we could have some dedicated accessor to `DeserializerState.with_attrs`. Furthermore, I would then like it if we would have the same accessor for `SerializerState` as well. This would enable the following refactoring of `SerializerState`: Don't save the `type_attr_map` in `SerializerState`, instead just `with_attrs::Bool`. The only place where this then needs a change is `attrs_list`, which would return `get(Oscar.type_attr_map, encode_type(T), Symbol[])` if `s.with_attrs` and `Symbol[]` otherwise. So there would be no need to pass the type_attr_map dict down through all the functions. What do you think about this?
> 
> the reason it was like this, was to at some point in the future allow overrides from the user, possible something to discuss at the meeting in kaiserslautern.

I could add an access-er but this may change in the future. For example if the user would like to have some attributes but not all, then we will go back to some attribute map

lgoettgens (approved):
This is already an improvement to current master, as it brings the functionality to the semantics described in the docstring. Let's postpone my other points to the meeting in KL

PR #4073, opened 2024-09-04
[FTheoryTools] Better type inference

lgoettgens:
While looking at `get_attribute` uses all over Oscar, I found a lot of places in the FTheoryTools that could easily get enhanced by explicit type annotations.

Futhermore, I found some things around `journal_name` missing, which are added here. And I also included the change of `euler_characteristic` from QQFieldElem to Int from https://github.com/oscar-system/Oscar.jl/pull/4044 in here, as this might produce ugly conflicts otherwise.


HereAround (approved):
Thank you very much @lgoettgens for catching and fixing all of this. Truly appreciated! Great!

PR #4078, opened 2024-09-10
Use Root system ref in weyl group serialization

antonydellavecchia:
remove unnecessary key


lgoettgens:
I would prefer to keep it with a key. In the future, we will add more properties to be serialized and by keeping the key here, this will no additional work for backwards compatibility like an upgrade script.

lgoettgens (changes requested):
I would prefer to keep it with a key. In the future, we will add more properties to be serialized and by keeping the key here, this will no additional work for backwards compatibility like an upgrade script.

PR #3688, opened 2024-05-04
Update to GAP.jl 0.11.3 resp. GAP 4.13.1

fingolfin:
Draft as there is a failing test I had to disable for now. We already have a fix for that in GAP, but I still need to apply it to the GAP_lib_jll . We might also just wait for GAP 4.13.1 which is due soon. But I would like to get full CI tests run now to make sure there aren't other regressions hiding.

Resolves #2341

PR #4081, opened 2024-09-11
fix `GAP.julia_to_gap(T::SubgroupTransversal)`

ThomasBreuer:
The problem arose from the improvement in https://github.com/oscar-system/GAP.jl/pull/989:
Right transversal objects are `AbstractVector`s, and now a three argument method for these objects is available in GAP.jl, which is not the method one wants.

The fix is preliminary, things will become better again as soon as `julia_to_gap` gets reworked, see https://github.com/oscar-system/GAP.jl/issues/1025.


lgoettgens:
This has been cherry-picked into https://github.com/oscar-system/Oscar.jl/pull/3688.

PR #4084, opened 2024-09-12
Lie algebras: Add root system detection

lgoettgens:
Add `root_system(L::LieAlgebra)` that either returns a cached root system or tries to compute the root system of `L`.
The remaining changes are either subroutines needed to implement the former, or some few places where using this new functionality made the code a lot cleaner.

~~For my convenience and to limit potential conflicts, this currently contains https://github.com/oscar-system/Oscar.jl/pull/4053. Once https://github.com/oscar-system/Oscar.jl/pull/4053 is merged, I'll rebase this PR.~~


fingolfin (approved):
Had a cursory looks. Seems sensible (I am not surprised). No objections from me.

lgoettgens:
Let's wait for another few days with merging for @felix-roehrich to have the chance to have a look as well

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -432,6 +432,321 @@ Return `true` if `L` is solvable, i.e. the derived series of `L` terminates in $
...
+    [
+      CartanInt(roots, alpha_i, alpha_j) for alpha_i in roots_simple,
+      alpha_j in roots_simple
+    ],
+  )
+  R = root_system(cm; check=true) # TODO: disable check
+
+  # reorder the simple roots according to the Dynkin diagram.
+  # users usually expect this ordering, but it is not necessary for the root system.
+  _, ordering = root_system_type_with_ordering(R)
```
felix-roehrich:
```suggestion
  _, ordering = cartan_type_with_ordering(cm)
```

This should also allow you to permute the matrix in place, we also don't compute the positive roots twice.

felix-roehrich: You might also want to store the type and then set it manually after the root system was constructed.

lgoettgens: Good catch, thanks! I didn't think about this. Let's try it out

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -432,6 +432,313 @@ Return `true` if `L` is solvable, i.e. the derived series of `L` terminates in $
...
+      CartanInt(roots, alpha_i, alpha_j) for alpha_i in roots_simple,
+      alpha_j in roots_simple
+    ],
+  )
+  _, ordering = cartan_type_with_ordering(cm)
+  permute!(roots_simple, ordering)
+  p = inv(Perm(ordering))
+  cm = p * cm * p
+  R = root_system(cm; check=false)
+  @assert root_system_type_with_ordering(R)[2] == 1:rank(R)
```
lgoettgens: I now replaced the second cartan matrix computation by this. Local testing showed that one has to invert the permutation here, because reasons. Is this what you thought of @felix-roehrich ?

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -432,6 +432,313 @@ Return `true` if `L` is solvable, i.e. the derived series of `L` terminates in $
...
+    [
+      CartanInt(roots, alpha_i, alpha_j) for alpha_i in roots_simple,
+      alpha_j in roots_simple
+    ],
+  )
+  _, ordering = cartan_type_with_ordering(cm)
+  permute!(roots_simple, ordering)
+  p = inv(Perm(ordering))
+  cm = p * cm * p
+  R = root_system(cm; check=false)
```
felix-roehrich: ~I think we can do it even easier, since `ordering` isn't used anywhere else.~ In the first version you had `check=true` on purpose, is this still needed?

```suggestion
  type, ordering = cartan_type_with_ordering(cm; check=false)
  permute!(roots_simple, ordering)
  R = root_system(type)
```

lgoettgens: The line `permute!(roots_simple, ordering)` needs to stay here and needs the ordering

felix-roehrich: I got confused about what is permuted. My updated comment should work, though.

lgoettgens: Yes, this is exactly what I was trying to type as well, but you were faster

experimental/LieAlgebras/src/LieAlgebra.jl
```diff
@@ -699,11 +699,9 @@ function _root_system_and_chevalley_basis(
...
     ],
   )
-  _, ordering = cartan_type_with_ordering(cm)
+  type, ordering = cartan_type_with_ordering(cm; check=false)
   permute!(roots_simple, ordering)
-  p = inv(Perm(ordering))
-  cm = p * cm * p
-  R = root_system(cm; check=false)
+  R = root_system(type)
   @assert root_system_type_with_ordering(R)[2] == 1:rank(R)
```
felix-roehrich: The assertion is no longer necessary now.
```suggestion
```

felix-roehrich (approved):

PR #4086, opened 2024-09-13
switch to a better Julia to GAP conversion

ThomasBreuer:
The changes rely on those in GAP.jl from oscar-system/GAP.jl/pull/1029, see the introductory comment there for the ideas.

The details are as follows.

- change the installation of `GapObj` and `GAP.julia_to_gap` methods to calls of `GAP.@install GapObj`
- change the `GAP.julia_to_gap` calls to `GapObj` calls
- move the Oscar to GAP conversion methods for `Set{T}` and `JSON3.Array` to GAP.jl

As soon as Oscar.jl includes GAP.jl containing the code from oscar-system/GAP.jl/pull/1029, the current changes (or something better) must be added to Oscar.jl.


fingolfin:
@ThomasBreuer this has conflicts now?

lgoettgens:
With the new Hecke release available, CI seems to be happy now

fingolfin (approved):
Looks good to me. It would be good if GAP.jl still provided `julia_to_gap` as fallback for existing code. I.e. we don't want people to install methods it, but code that just calls it should keep working. So something like `julia_to_gap(x) = GapObj` in GAP.jl might work (perhaps `recursive` also needs to be passed on)

PR #4091, opened 2024-09-16
Disable caching in a restriction_map method

fingolfin:
Extracted from #3865. On the surface it looks to me as if this one is harmless. We'll see whether the tests pass.


src/AlgebraicGeometry/Schemes/AffineSchemeOpenSubscheme/Rings/Methods.jl
```diff
@@ -243,7 +243,7 @@ function restriction_map(
   # the terms accordingly, we derive the desired expressions for the cᵢ's.
   #W = localized_ring(OO(Y))
   W = OO(Y)
-  S, t = polynomial_ring(W, ["t$i" for i in 1:r])
+  S, t = polynomial_ring(W, ["t$i" for i in 1:r]; cached=false)
```
fingolfin, line 246: This is the only important change in this PR. The rest is just rewriting to make things more terse but otherwise has no semantic effect whatsoever and could also be dropped from this PR.

afkafkafk13 (approved):
Looks fine.

PR #4094, opened 2024-09-16
Disable polynomial ring caching in `projective_space` and `affine_space` methods

fingolfin:
Another part of PR #3865 (and this part of it was already discussed extensively there)

I hope it'll either just pass fine, or if it has problems, then at least just a subset of those in PR #3865.

### DIFF
--- a/test/AlgebraicGeometry/Schemes/AffineRationalPoint.jl
+++ b/test/AlgebraicGeometry/Schemes/AffineRationalPoint.jl
@@ -6,15 +6,13 @@
   pX = X([1,0])
   pA = A2(pX)
   @test pX == pA
+  @test A2([1,0]) == pA
   # ring around the rosie once
   @test pX == X(rational_point_coordinates(defining_ideal(scheme(pX))))
   Oscar.closed_embedding(pA)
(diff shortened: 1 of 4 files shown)


simonbrandhorst (approved):

afkafkafk13 (approved):
Looks fine.

test/AlgebraicGeometry/Schemes/AffineRationalPoint.jl
```diff
@@ -6,16 +6,14 @@
...
+  @test A2([1,0]) == pA
   # ring around the rosie once
   @test pX == X(rational_point_coordinates(defining_ideal(scheme(pX))))
   Oscar.closed_embedding(pA)
   Oscar.closed_embedding(pX)
   @test dim(tangent_space(pX))==1
   @test dim(tangent_space(pA))==2
 
-  A2a = affine_space(GF(2), [:x, :y]);
-  @test A2a([1,0]) == pA
```
fingolfin, line 17: This test failed because now `A2` and `A2a` are not using the same polynomial ring anymore (note that `A2 != A2a` holds now and also before)

PR #4097, opened 2024-09-16
generic exterior algebra

HechtiDerLachs:
This PR adds a generic implementation for exterior algebras over arbitrary commutative `base_ring`s. The application is computation of relative BGG. To this end, we also introduce the Cartan-Eilenberg resolution of a complex of modules.


experimental/DoubleAndHyperComplexes/src/Objects/cartan_eilenberg_resolution.jl
```diff
@@ -0,0 +1,279 @@
...
+    return new{chain_type(c)}(c, is_exact, Dict{Int, AbsHyperComplex}(), Dict{Int, AbsHyperComplex}(), Dict{Int, AbsHyperComplexMorphism}())
+  end
+end
+
+function kernel_resolution(fac::CEChainFactory, i::Int)
+  if !haskey(fac.kernel_resolutions, i)
+    Z, _ = kernel(fac.c, i)
+    fac.kernel_resolutions[i] = free_resolution(SimpleFreeResolution, Z)[1]
+  end
+  return fac.kernel_resolutions[i]
```
fingolfin, line 57: This is the kind of code for which `get!` exists to avoid performing and hashing three times (for `haskey`, for storing, and then for looking it up again)
```suggestion
  return get!(fac.kernel_resolutions, i) do
    Z, _ = kernel(fac.c, i)
    return free_resolution(SimpleFreeResolution, Z)[1]
  end
```
Same applies to `boundary_resolution` and `induced_map`

HechtiDerLachs:
This has been taken up on in #4360 .

PR #4100, opened 2024-09-16
Injective and irreducible resolutions of Q-graded modules

tom111:
This is a draft pull request as recommended [here](https://docs.oscar-system.org/stable/DeveloperDocumentation/new_developers/).

Anna Hofer @annahofer00 has been working on an implementation of the [algorithms of Helm/Miller](https://arxiv.org/abs/math/0309256) and here we are now trying to put this into OSCAR. At the moment this code can compute some irreducible resolutions, but polyhedral pre-computation is done by hand. This needs to be automated.

The code also relies on the computation of certain colon-modules that is not yet available in OSCAR but maybe it should be.

Specifically, if I is an ideal in a ring R and M an R-Module then we need 0 :_M I = {m \in M : mI = 0}. As a module this is isomorphic to Hom(R/I, M) but one has to be careful with degree information (in the graded case). I think it would be good to implement a general purpose "module quotient" function like [this one in Macaualy2](https://www.macaulay2.com/doc/Macaulay2/share/doc/Macaulay2/Saturation/html/_quotient_lp__Module_cm__Module_rp.html), which has the very practical operator `:` for this purpose. Many of these colons also have a saturation, where this is repeated to stabilization.

Any comments are welcome of course. We'll keep adding to this branch for now.

-----

To briefly summarize the state of affairs: 

- [x] `is_normal` should be implemented and used for `MonoidAlgebra`s. Why does the current implementation not treat graded rings? 
- [x] the method `codim(::ModuleFP)` should be deleted;
- [ ] `dim(::ModuleFP)` should be optimized via Hilbert functions etc. -> Functionality does not seem to be available for `Singular.smodule`s in Singular.jl! Maybe we should make this a feature request? Ping @hannes14 .
- [x] `rational_to_integer_vector`: Replace any call to this function by the method linked by @lgoettgens below.
- [ ] ~~`mod_quotient` has to be streamlined with the existing functionality for the colon operator;~~ can now be postponed as everything is wrapped in its own module and this functionality is not exported. 
- [ ] docstrings have to be overhauled and streamlined. Moreover, the `docs` section has to be adjusted as pointed out by @lgoettgens 
- [x] Can we fuse `IrrSum` and `InjMod`?
- [x] The implementation for `+` of `MonoidAlgebraModule`s should become a method of `direct_sum`.
- [ ] printing has to be overhauled and adjusted according to the guidelines.
- [x] `_f_k` should be replaced by an appropriate `evaluate` and deleted. `evaluate` probably needs to be extended to `MonoidAlgebraElem`s and friends for this.
- [x] `fix_module`: ~~Why is it necessary?~~ Does not seem to be required anymore.
- [ ] Introduce some kind of context object to cache intermediate results for the computation of local cohomology.
- [x] wrap up the whole experimental package in a julia-module and only export the absolute necessary things.
- [ ] Functionality for `ModuleFP`s over `MonoidAlgebra`s has to be cleaned up; we will start using that in followup pull requests, but nothing using `MonoidAlgebraModule` should be exported in the meantime! 
- [x] Turn the example files into actual tests.
- [x] Rename `irreducible_dec` into `irreducible_decomposition` and vice versa for `irreducible_res`.
- [x] Replace `lc_zero` by some appropriate use of `is_zero`.
- [ ] use Polymake's `normaliz` (if present) to check normality.


experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+            push!(L_F[2], degree(Vector{Int},g))
+        end
+    end
+    
+    #get corresponding face defining it by a system of inequalities
+    F_poly = []
+    m = length(gens(grading_group(R_Q)))
+    C = convex_hull(zeros(Int, m))
+    A = Nothing
+    b = Nothing
```
fingolfin:
```suggestion
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+    end
+    return A, b
+end
+
+@doc raw"
+    get_face_to_prime(P_F::Ideal)
+
+    Given a homogeneous prime ideal, return the corresponding face F
+
+    INPUT:    prime ideal p_F
```
fingolfin: prime ideal in what kind of ring?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+    OUTPUT:   list [face F, matrix A, vector b] such that F is defined by Ax <= b
+"
+function get_face_to_prime(P_F::Ideal)
+    R_Q = base_ring(P_F)
+    #get generators of k[F] = k[Q]/P_F
+    Q_F,_ = quo(R_Q,P_F)
+    G_F = gens(Q_F)
+    L_F = (P_F,[])
```
fingolfin: Hmm, I am confused. You never access `L_F[1]`. Could you perhaps do this
```suggestion
    L_F = []
```
and then below replace `L_F[2]` by just `L_F`?

In addition `[]` is a `Vector{Any}` which isn't great for type stability.

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+    return F_poly
+end
+
+##get all prime ideals P_F associated to M
+##INPUT:    ideal I
+##OUTPUT:   list of all minimal associated primes + the maximal ideal and the corresponding face F as a polyhedron  
+function get_all_ass_primes(I)
+    R_Q = base_ring(I)
+    I_m = ideal(R_Q,gens(R_Q)) #maximal ideal
+    F_m = get_face_to_prime(I_m)[1]
```
fingolfin: Why does `get_all_ass_primes` return multiple values if you only use the first? Perhaps it should also return a `FaceQ` as its name suggests?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+export quotient_by_ideal
+
+## Data 
+export FaceQ 
+
+#########################
+# some composite types
+#########################
+
+struct FaceQ # face of semigroup
```
fingolfin: What is a "face of semigroup" ? And what does the `Q` stand for here?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+function compute_shift(I,i,c,P_Q)
+    n_bass = compute_bass_numbers(I,i)
+    j = 0
+    while !points_in_Q(n_bass,P_Q)
+        bass_ = []
+        for a_bass in n_bass
+            push!(bass_,a_bass+c)
+        end
+        n_bass = bass_
+        j = j +1
```
fingolfin:
```suggestion
        j += 1
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+##OUTPUT:   ZZ^d degree of shift
+function compute_shift(I,i,c,P_Q)
+    n_bass = compute_bass_numbers(I,i)
+    j = 0
+    while !points_in_Q(n_bass,P_Q)
+        bass_ = []
+        for a_bass in n_bass
+            push!(bass_,a_bass+c)
+        end
+        n_bass = bass_
```
fingolfin:
```suggestion
        n_bass = [a+c for a in n_bass]
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+    
+    D = []
+    for j=0:i in 
+        E = Nothing
+        try 
+            E = ext(R_k,M_mod,j)
+        catch e
+            if isa(e,AssertionError)
+                E = Nothing
+            end
```
fingolfin: That looks like a really nasty hack. At the very least it needs a comment justifying it. Better would be excise it with fire, though ;-).

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+    end
+
+    # residue_field
+    I_m = ideal(S,gens(S))
+    R_k,_ = quotient_by_ideal(I_m)
+
+    # module R_Q/I as a subquotient of S^1
+    M_mod = quotient_ring_as_module(quo(R_Q,I)[1])
+    
+    D = []
```
fingolfin: That's a `Vector{Any}`. Don't you know a bit more about the return value? Then you should use that here, e.g. for a `Vector{Vector{Int}}` you'd write
```suggestion
    D = Vector{Int}[]
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+#           homomorphism (inclusion) Mxy -> M
+#           k[ZF]-basis computed as in ZF_basis(_,_)
+function coefficients_incl(N::SubquoModule,Np::SubquoModule,incl_Np::SubQuoHom,Bp::Vector,pr::Ideal)
+    R_N = base_ring(N)
+    if is_zero(modulus(R_N))
+        R_poly = R_N
+    else
+        R_poly = base_ring(R_N)
+    end
+    T = typeof(zero(R_N))
```
fingolfin: This really should be
```suggestion
    T = elem_type(R_N)
```
also elsewhere in similar situations

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+
+#Returns ideal of underlying polynomial ring
+#INPUT:     ideal I of quotient ring
+#OUTPUT:    ideal I as an ideal of polynomial ring
+function ideal_in_poly_ring(I::Ideal)
+    R_I = base_ring(I)
+    J = modulus(R_I)
+    if is_zero(J)
+        return I
+    else
```
fingolfin: More idiomatic:
```suggestion
    is_zero(J) && return I
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+        S = base_ring(R_I)
+        G = []
+        for g in gens(I) 
+            push!(G,monomial_basis(S,degree(g))[1])
+        end
```
fingolfin: More idiomatic and also more type stable (i.e., the vector this produces not always a `Vector{Any}`) is to use a list comprehension:
```suggestion
        G = [monomial_basis(S,degree(g))[1] for g in gens(I)]
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+#OUTPUT:    quotient S/I as SubquoModule (S = base_ring(I))
+function quotient_by_ideal(I::Ideal)
+    R_I = base_ring(I)
+    F = graded_free_module(R_I,1)
+    f = F[1]
+    T = typeof(f)
+    V = Vector{T}()
+    for g in gens(I) 
+        push!(V,g*f)
+    end
```
fingolfin:
```suggestion
    V = [g*f for g in gens(I)]
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+        end
+        return ideal(S,G)
+    end
+end
+
+#get quotient ring S/I as module (S quotient ring) 
+#alternative to quotient_ring_as_module(I), which gives wrong result when base_ring(I) = quotient ring 
+#INPUT:     ideal I 
+#OUTPUT:    quotient S/I as SubquoModule (S = base_ring(I))
+function quotient_by_ideal(I::Ideal)
```
fingolfin: The name of this function and the comment leave me really confused. This is not something that should be reachable as `Oscar.quotient_by_ideal`. Might be acceptable if hidden in a submodule.

fingolfin:
Last night I was a bit terse in what I wrote here, so I'd like to clarify that overall I think this looks like a very valuable pull request, thank you for starting on it! My comments are all on a relatively low-level technical level, and I think most of them are trivial to address, but also not urgent, i.e. it is fine to only address them gradually. Also feel free to ask resp. push back if something is unclear or seems unreasonable to you. 

As an example: the name `FaceQ` of course is fine if you use it internally and know what you do, but long term others may have to interact with this code, and then either a more descriptive name, or else a comment that explains the name, can go a long way.

Anyway, thanks again for making this PR in the first place and I look forward to seeing how it evolves.

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
fingolfin: I assume there is no alternative to catching this assertion right now, and we need to improve something in the existing OSCAR code base. Hence my request for a comment elucidating on why this is here -- or even better, feel free to open an issue requesting a change to `ext` that allows you to deal with whatever you are dealing with here in a cleaner and more sustainable way.

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,573 @@
...
+        push!(P_F,FaceQ(mp,p_f[1],p_f[2],p_f[3]))
+    end
+    return P_F
+end
+
+
+##Let W = k{Q\(a + F - Q)} and let H be a hyperplane containing F as a subset.
+##Calculates generators of k{(a + H_+^°)\cap Q}.
+##INPUT:    semigroup Q_P as a polyhedron
+#           zonotope G_Q of Q_P as in Lemma 3.10
```
fingolfin: Excellent: I really appreciate these references to specific lemmas and theorems in the code, that will surely be a great asset to people in the future!

One tiny request, though: it is not immediately clear to me to which book or article these reference -- presumably https://arxiv.org/abs/math/0309256 from the PR description? If so, it would be great if e.g. at the top of this file, or in the README (or both places) this link could appear to remove any ambiguity (also, once this PR is merged, someone might read the code w/o knowing which PR it came from, so they wouldn't about that arXiv link at all.

(long term we'd put this paper into our bib file, too, but no need to do that now unless you really want to for some reason.)

wdecker:
Dear Anna and Thomas,

thanks a lot for this PR which is very welcome! I agree that OSCAR should have column operations and saturation also for modules and not only for ideals. This is on my todo list and I will do it asap. Let us keep in touch on this and your PR and feel free to suggest other features to be added to OSCAR.

Best wishes,

Wolfram

tom111:
Thanks everyone for the comments so far. I understand that there are many issues with the code as it stands now. We have a limited understanding of both stylistic conventions and inner workings of OSCAR. I think the detailed analysis, naming conventions, etc. can be deferred a little. Let me ask on a more strategic level:

Most of the time, we will work with modules over affine semigroup rings. Such a ring can be presented a quotient of a polynomial ring modulo a toric ideal and comes with different polyhedral and algebraic data attached to it. For example, it is graded by a finitely generated subsemigroup of ZZ^d (we always refer to this as 'Q'). This monoid lives inside a polyhedral cone, and the faces of that cone correspond 1:1 to Q-graded prime ideals. The data structures of our injective modules and local cohomlogy are often polyhedral in nature, which is why we are really hopeful that OSCAR is the right system to do this in.

Does it make sense to abstract such monoid algebras (some people say affine semigroup rings) in OSCAR before getting started?

At the moment, such a ring would be represented in OSCAR as a normal quotient ring. We will have R a polynomial ring, I a toric ideal and then our base_ring for most of what we do will be S = R/I (together with its multigrading, which is very important). In fact, the same question arises for the grading. Mathematically it is by a monoid Q, but we will (have to) use equivalence classes of vectors in ZZ^d instead, i.e. work in a presentation.

Now If J is an ideal of S and we construct the quotient S/J, then this is automatically turned into R/(I+J). I understand that this is what is needed _internally_, but for what I'm modeling, S/J is an S-module. That's the structure we want, e.g. compute the injective resolution of S/J as an S-module. Everything will be Q-graded S-modules from here on. So, for example, `quotient_ring_as_module` at the moment does not work in such a situation and Anna implemented it the way we need it (in the criticized function `quotient_by_ideal`). But part of me thinks that this should maybe be done in OSCAR. But then maybe we are just doing it wrongly and a different paradigm solves this.

Is this pull request the correct place to discuss such questions?

Another basic question: In this case, there is this one paper where we take all our notation from. Is it fair to assume (in the docs and comments) that people have read that paper?

mjrodgers:
@micjoswig or other commutative algebra people, do you think you can say something about this?

micjoswig:
> @micjoswig or other commutative algebra people, do you think you can say something about this?

Will do next Monday.

lkastner:
Answer after discussing with @micjoswig and @flenzen:

> Does it make sense to abstract such monoid algebras (some people say affine semigroup rings) in OSCAR before getting started?
> 
Yes, because we have affine algebras, we suggest something like `AffineSemigroupAlgebra` as a wrapper around our existing affine algebra type together with the auxiliary polyhedral data you need.

Eventually this should tie in with our [standard graded rings and modules](https://docs.oscar-system.org/stable/CommutativeAlgebra/rings/#Gradings). Once you reach this stage, please talk to @HechtiDerLachs .


> At the moment, such a ring would be represented in OSCAR as a normal quotient ring. We will have R a polynomial ring, I a toric ideal and then our base_ring for most of what we do will be S = R/I (together with its multigrading, which is very important). In fact, the same question arises for the grading. Mathematically it is by a monoid Q, but we will (have to) use equivalence classes of vectors in ZZ^d instead, i.e. work in a presentation.
> 
That should take care of this issue.

> Now If J is an ideal of S and we construct the quotient S/J, then this is automatically turned into R/(I+J). I understand that this is what is needed _internally_, but for what I'm modeling, S/J is an S-module. That's the structure we want, e.g. compute the injective resolution of S/J as an S-module. Everything will be Q-graded S-modules from here on. So, for example, `quotient_ring_as_module` at the moment does not work in such a situation and Anna implemented it the way we need it (in the criticized function `quotient_by_ideal`). But part of me thinks that this should maybe be done in OSCAR. But then maybe we are just doing it wrongly and a different paradigm solves this.

We just looked at the documentation of `quotient_ring_as_module` and it seems currently impossible. Of course this suggestion makes sense.
Ping @thofma : Do you have a suggestion?

> Is this pull request the correct place to discuss such questions?

Yes. You can also ask questions in the slack.

> Another basic question: In this case, there is this one paper where we take all our notation from. Is it fair to assume (in the docs and comments) that people have read that paper?

Yes, please add an bib entry as described in [our reference docs](https://docs.oscar-system.org/stable/DeveloperDocumentation/documentation/#Updating-the-bibliography) if necessary, and use this to references notation. Please note that the Miller Sturmfels book is already citable as MS05. Textbook references are preferred if viable.

micjoswig:
> > @micjoswig or other
> Will do next Monday.

Sorry for the delay.

tom111:
Thanks everybody. We will get to work as soon as possible and proceed as suggested here.

thofma:
> > Now If J is an ideal of S and we construct the quotient S/J, then this is automatically turned into R/(I+J). I understand that this is what is needed _internally_, but for what I'm modeling, S/J is an S-module. That's the structure we want, e.g. compute the injective resolution of S/J as an S-module. Everything will be Q-graded S-modules from here on. So, for example, `quotient_ring_as_module` at the moment does not work in such a situation and Anna implemented it the way we need it (in the criticized function `quotient_by_ideal`). But part of me thinks that this should maybe be done in OSCAR. But then maybe we are just doing it wrongly and a different paradigm solves this.
> 
> We just looked at the documentation of `quotient_ring_as_module` and it seems currently impossible. Of course this suggestion makes sense. Ping @thofma : Do you have a suggestion?

I had a quick glance at the examples in the PR here, but could not find an instance where this is used yet (maybe I looked at the wrong place). But from what you are saying, I think it is about the following situation (correct me if I am wrong):
```
julia> begin
         R_Q,(x,y) = graded_polynomial_ring(QQ,["x","y"])
         I = ideal(R_Q,[x^4,x^2*y^2,y^4])
         S, = quo(R_Q, I)
         J = ideal(S, [])
       end;

julia> quotient_by_ideal(J) # the method added here
(Graded submodule of S^1
[...]

julia> quotient_ring_as_module(J) # the method in Oscar
ERROR: MethodError: no method matching quotient_ring_as_module(::MPolyQuoIdeal{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}})

Closest candidates are:
  quotient_ring_as_module(::MPolyIdeal)
   @ Oscar ~/software/Oscar.jl/src/Modules/ModulesGraded.jl:2923
[...]
```
Is this correct? (I could not reproduce the "automatically turned into R/(I + J)" part.)

(Another difference would be that `quotient_by_ideal` also returns a map (but which is not used, at least in the examples yet).)

wdecker:
May be I misunderstand the discussion here, but should we not extend the function `quotient_ring_as_module` by just allowing another type:

function quotient_ring_as_module(I::Union{MPolyIdeal, MPolyQuoIdeal})
  R = base_ring(I)
  F = is_graded(R) ? graded_free_module(R, 1) : free_module(R, 1)
  e1 = F[1]
  return quo_object(F, [x * e1 for x = gens(I)])
end

annahofer00:
Thanks everyone! I just started working on this project again. I have introduced a type MonoidAlgebra as a struct with different fields for the auxiliary data. Is this the right way to go? 

The rest of the code I am still working on and I don’t need any comments on it at the moment.

lkastner:
> Thanks everyone! I just started working on this project again. I have introduced a type MonoidAlgebra as a struct with different fields for the auxiliary data. Is this the right way to go?
> 

It looks good to me. I assume that you need this auxiliary data like `faces` and `hyperplanes` since the order is important? Or is this something you could also recover from the cone?

Also I wonder whether there is documentation on our ring interface, maybe @thofma can help me out.

thofma:
In case you need it, there is a (I think) nice example for what a ring should implement here: https://nemocas.github.io/AbstractAlgebra.jl/stable/ring_interface/#Minimal-example-of-ring-implementation. Also feel free to drop me a message on slack if there are some specific questions.

annahofer00:
Hello everyone,

I've made significant updates and cleaned up the code. The main file, _InjectiveResolutions.jl_, now contains all functions for computing injective resolutions. Additionally, I’ve created a new file, _LocalCohomology.jl_, which provides functions for computing local cohomology of modules over monoid algebras.

This version now includes:
- Improved data types for monoid algebras, injective modules, sector partitions, and related structures.
- Several examples and tests, along with documentation for the provided functions. However, additional tests and more comprehensive documentation would still be beneficial.

I have not used the newly implemented quotient(M::SubquoModule, J::Ideal) function, as it returns a submodule of the ambient free module. Instead, I have implemented my own versions: mod_quotient and mod_saturate. Would it make sense to incorporate these directly into OSCAR?

I would appreciate any feedback or suggestions for further improvements. Hopefully, this PR can be merged soon!

Thanks!
Anna

micjoswig:
Dear @annahofer00 ,

your code is most welcome and useful!

If you should be available today (Friday, 4 April 2025), you could join the OSCAR zoom at 15:00, where you could meet with the developers.  Would that work for you?

Afterwards, you could maybe e-meet with @HechtiDerLachs and @flenzen to go through the code once more, to prepare the merge. 

Best,
Michael

annahofer00:
Thanks for the quick response @micjoswig!
> If you should be available today (Friday, 4 April 2025), you could join the OSCAR zoom at 15:00, where you could meet with the developers. Would that work for you?

Unfortunately, I am only available until 15:45 today so I would prefer next week. It would be great to meet with the developers!

micjoswig:
> Thanks for the quick response @micjoswig!
> 
> > If you should be available today (Friday, 4 April 2025), you could join the OSCAR zoom at 15:00, where you could meet with the developers. Would that work for you?
> 
> Unfortunately, I am only available until 15:45 today so I would prefer next week. It would be great to meet with the developers!

The meeting takes 30mins only.  So, if you can, please join today, just to say hello. Then you can meet with @HechtiDerLachs and @flenzen later.

annahofer00:
> The meeting takes 30mins only. So, if you can, please join today, just to say hello. Then you can meet with @HechtiDerLachs and @flenzen later.

Alright, I will join. Could you share the zoom link?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1460 @@
+## Functions visible on the outside
```
lgoettgens, line 92:
```suggestion
module InjectiveResolutions

using ..Oscar
using ..Oscar: IntegerUnion # add other things that are not exported from Oscar here

# functions with new methods
import ..Oscar:
  coefficient_ring,
  gens # add more things here

import Base:
  +,
  -,
  *,
  == # add more things here

## Functions visible on the outside
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1460 @@
...
+    d = dim(Jac)                   # Get dimension of the Jacobian
+    d < 0 && (d = -Inf)            # Handle negative dimensions
+    return (dim(R) - d >= 2)       # Check the condition
+  else
+    return false                    # S2 condition not satisfied
+  end
+end
+
+# import local cohomology functions
+include("LocalCohomology.jl")
```
lgoettgens, line 1070:
```suggestion
include("LocalCohomology.jl")

end # module InjectiveResolutions

using .InjectiveResolutions

# place to eventually add exports
```

lgoettgens (commented):
@HechtiDerLachs the two suggestions below add the submodule shenanigans we talk about during triage today.

All exports that are in the current list, will then only be available in the REPL once one types `using Oscar.InjectiveResolutions`. To make something already available after `using Oscar`, those exports must be *duplicated* to the second export list at the bottom of the file (currently empty, but that's how you described you want it for this first merge).
Other changes needed for this to work:
- the `import ..Oscar:` and `import Base:` lists at the top needs to be extended to contain all functions from outside the new submodule where the submodule adds a new method. I added some obvious ones as breadcrumbs.
- the test files should call `using Oscar.InjectiveResolutions` at the top, as otherwise you would need to qualify basically every function you want to call which clutters the code a lot

experimental/InjectiveResolutions/test/ModuleFunctionality.jl
```diff
@@ -0,0 +1,21 @@
...
+  (x, y, z) = gens(kQ)
+  I, inc_I = sub(F, [x*F[1], y*F[2]])
+  u, v = gens(I)
+  @test repres(v) in I
+  c = coordinates(repres(u + v), I)
+  @test c[1] == 1 && c[2] == 1
+
+  phi = hom(F, F, [F[1], F[1]])
+  @test F[1] - F[2] in kernel(phi)[1]
+end
```
HechtiDerLachs: @jankoboehm : 

I could make these tests run for modules over `MonoidAlgebra`s by implementing translations between the Singular and the Oscar side for the `ModuleGens`. The required code is in `experimental/InjectiveResolutions/src/ModuleFunctionality.jl` above. It's still a bit hacky and needs some cleanup. But I wanted to showcase it here, so that you can have a look and say whether this is going towards the right direction. 

@annahofer00, @flenzen : If this goes through, then we should be able to replace `MonoidAlgebraModule` by a `SubquoModule` over the `MonoidAlgebra` directly.

HechtiDerLachs:
To briefly summarize the state of affairs: 

- [x] `is_normal` should be implemented and used for `MonoidAlgebra`s. Why does the current implementation not treat graded rings? 
- [x] the method `codim(::ModuleFP)` should be deleted;
- [ ] `dim(::ModuleFP)` should be optimized via Hilbert functions etc. -> Functionality does not seem to be available for `Singular.smodule`s in Singular.jl! Maybe we should make this a feature request? Ping @hannes14 .
- [ ] `rational_to_integer_vector`: Replace any call to this function by the method linked by @lgoettgens below.
- [ ] `mod_quotient` has to be streamlined with the existing functionality for the colon operator;
- [ ] docstrings have to be overhauled and streamlined. Moreover, the `docs` section has to be adjusted as pointed out by @lgoettgens 
- [ ] Can we fuse `IrrSum` and `InjMod`?
- [ ] The implementation for `+` of `MonoidAlgebraModule`s should become a method of `direct_sum`.
- [ ] printing has to be overhauled and adjusted according to the guidelines.
- [x] `_f_k` should be replaced by an appropriate `evaluate` and deleted. `evaluate` probably needs to be extended to `MonoidAlgebraElem`s and friends for this.
- [ ] `fix_module`: Why is it necessary?
- [ ] Introduce some kind of context object to cache intermediate results for the computation of local cohomology.
- [ ] wrap up the whole experimental package in a julia-module and only export the absolute necessary things.
- [ ] Functionality for `ModuleFP`s over `MonoidAlgebra`s has to be cleaned up; we will start using that in followup pull requests, but nothing using `MonoidAlgebraModule` should be exported in the meantime! 
- [ ] Turn the example files into actual tests.

lgoettgens:
> * `rational_to_integer_vector`: Does this already exist somewhere?

Should be the same as https://docs.oscar-system.org/dev/PolyhedralGeometry/intro/#primitive_generator-Union{Tuple{AbstractVector{T}},%20Tuple{T}}%20where%20T%3C:Union{Integer,%20QQFieldElem,%20ZZRingElem,%20Rational

HechtiDerLachs:
I added a hotfix for issue #4833 which hopefully allows us to proceed for the moment while that issue is pending. @annahofer00 : Let me know if the problems encountered on Friday still prevail; then I can have another look. 

Edit: `irreducible_res` works for me now. `injective_res` still seems to need some work, but within `InjectiveResolutions` and it looks like this can be proceeded with.

fingolfin:
@hannes14 will add the missing `dim(::smodule)` to Singular.jl

hannes14:
https://github.com/oscar-system/Singular.jl/pull/861 implements dimension(smodule) in Singular.jl

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1680 @@
...
+export IndecInj
+#export underlying_element
+#export underlying_ideal
+export irreducible_hull
+#export _get_irreducible_ideal
+#export compute_shift
+#export Q_graded_part # should be renamed?
+#export compute_Q_graded_part
+export local_cohomology
+export zeroth_local_cohomology
```
HechtiDerLachs:
```suggestion
export local_cohomology
export zeroth_local_cohomology
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,652 @@
+export local_cohomology
+export local_cohomology_all
+export zeroth_local_cohomology
+export SectorPartitionLC
+export get_scalar_matrix
```
HechtiDerLachs:
```suggestion
export SectorPartitionLC
export get_scalar_matrix 
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
wdecker: @tom111 Out of curiosity: How does this method for `is_normal` compare with the other one we have in Oscar which computes the first step in the Greuel/Laplagne/Seelisch algorithm?

tom111: I have not checked tbh. The algorithm here works for general equidimensional rings and checks S2 and R1 conditions. S2 requires a bunch of Ext computations and R1 with the Jacobian ideal (which can be costly). I don't know about the performance of the GLS algorithm, but given that it is much younger than checking Serre conditions it's likely to be better?

In any case: For the new data type `MonoidAlgebra` both normalization (not currently implemented?) and `is_normal` should call upon `normaliz` or at least some polyedral check if the semigroup is saturated. We have not looked into this kind of interface work, though.

Edit: I don't know how to call the GLS is_normal properly:
```
S = monoid_algebra_from_lattice([[4,0],[3,1],[1,3],[0,4]], QQ)
is_normal(S) # works and gives false
S.algebra
is_normal(S.algebra) # calls default implementation which is not implemented for decorated rings
```

HechtiDerLachs: To give some context on how we worked on this: @annahofer00 found that the `is_normal` method for graded rings throws an error in Oscar. @jankoboehm suggested to me that this might be due to some decisison that `is_normal` was probably thought of as a check on a projective variety represented by a graded ring, so whatever happens at the irrelevant ideal should be ignored. But this is just speculation. 

Since we did not want to mess with the Oscar internals and we have now introduced an honest ring of its own type here (`MonoidAlgebra`), we simply rerouted the implementation in this PR as a custom method for `MonoidAlgebra`s for the time being. 

Comparison of performance is not that easy with the user-facing functionality provided by Oscar. One would first have to strip off the grading, so that the agnostic `is_normal` implementation is triggered. This can be done by calling `is_normal(forget_grading(A.algebra))` on a `MonoidAlgebra` `A` (modulo syntax mistakes). 

I also have not come around to compare performances as of yet.

tom111: Yes, this is good as it works right now. For the future of `MonoidAlgebra` the best thing is to use a polyedral test, i.e. what is done in Section 2.11 of the normaliz manual:
https://www.normaliz.uni-osnabrueck.de/wp-content/uploads/2017/01/Normaliz3.2.0Documentation.pdf
Is normaliz included in OSCAR, maybe via Polymake?

HechtiDerLachs: > Is normaliz included in OSCAR, maybe via Polymake?

If so, we should put this in the TODO-list for a follow-up PR. 

Ping @lkastner ?

wdecker: > To give some context on how we worked on this: @annahofer00 found that the `is_normal` method for graded rings throws an error in Oscar. @jankoboehm suggested to me that this might be due to some decisison that `is_normal` was probably thought of as a check on a projective variety represented by a graded ring, so whatever happens at the irrelevant ideal should be ignored. But this is just speculation.

@HechtiDerLachs No, geometrically, the normalization algorithms are meant for affine varieties. Technically, as the algorithms create further quotient rings on their way, we would mess up with gradings. The error thrown is just a not-implemented error. Should we add something like *use `forget_grading` first*?

jankoboehm: What I speculated was that the graded case was excluded to avoid missunderstandings thinking that this is a projective normality test.

HechtiDerLachs:
@lgoettgens : Do you know more about the cause of the failing tests here? 
```
Unbound type parameters detected:
      From worker 2:	[1] sector_partition(kQ::MonoidAlgebra, phi::Union{MatElem{<:T}, Vector{Any}}, psi::Union{MatElem{<:T}, Vector{Any}}, j::Integer, k::Integer, J::Oscar.InjectiveResolutions.IndecInj...) where T<:FieldElem @ Oscar.InjectiveResolutions ~/oscar-runners/oscar-runner-03/_work/Oscar.jl/Oscar.jl/experimental/InjectiveResolutions/src/LocalCohomology.jl:336
```
I do not really understand what's going on. Thanks for any help!

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,639 @@
...
+and j = |J^0|, k = |J^0| + |J^1|. This function computes a sector partition of the local cohomology module ker(psi)/im(phi). 
+"""
+function sector_partition(
+  kQ::MonoidAlgebra,
+  phi::Union{Vector{Any},MatElem{<:T}},
+  psi::Union{Vector{Any},MatElem{<:T}},
+  j::Integer,
+  k::Integer,
+  J::IndecInj...,
+) where {T<:FieldElem}
```
lgoettgens, line 336: @HechtiDerLachs the failure is due to the following: If phi and psi are both of type Vector{Any}, then there is nothing restricting the value of the type variable T.
This is usually a symptom of some bad code design somewhere. Without looking any deeper into the code in this PR (I could do that tomorrow if still required), I am confused why phi and psi can be a matrix (sounds reasonable for the internal representation of a morphism) *or a Vector{Any}*. Should the *Vector{Any}* encode some special case like zero or identity? If yes, why not use such a matrix instead.
If you can point me to the line of code where such a Vector{Any} is created, I may have a suggestion how to change it to something that doesn't complain

HechtiDerLachs: Thanks! I think, we fixed it.

experimental/InjectiveResolutions/README.md
```diff
@@ -0,0 +1,10 @@
+# Injective Resolutions over monoid algebras
+
+## Aims
+
+Computing injective resolutions and local cohomology over monoid algebras
+
+## Status
+
+Just setting up my Oscar.
+
```
HechtiDerLachs: Update this?

experimental/InjectiveResolutions/docs/src/introduction.md
```diff
@@ -0,0 +1,182 @@
...
+Any questions or comments about injective resolutions and local cohomology in
+OSCAR can be sent to:
+
+- [Anna Hofer](https://annahofer00.github.io)
+- [Thomas Kahle](https://www.thomas-kahle.de)
+
+## Acknowledgements
+
+We appreciate extensive support on integrating this code into OSCAR provided by
+[Mathias Zach](https://hechtiderlachs.github.io/).
```
HechtiDerLachs: Haha! Thanks. But:
```suggestion
[Matthias Zach](https://hechtiderlachs.github.io/).
```
Edit: *denying any responsibility!

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1327 @@
...
+end
+
+function Base.show(io::IO, ::MIME"text/plain", kQ::MonoidAlgebra)
+  println(io, "monoid algebra over ", lowercase(string(coefficient_ring(kQ))),
+    " with cone of dimension $(dim(cone(kQ)))",
+  )
+  #=
+  println(io,"Monoid algebra k[Q] over ", lowercase(string(coefficient_ring(kQ.algebra))), " generated by ", join(gens(kQ.algebra), ", "), " quotient by ideal (", join(gens(modulus(kQ.algebra)),", "),"), ")
+  print(io, "where Q is a subset of ZZ^", ambient_dim(kQ.cone)," and the cone RR_{>= 0}Q has dimension ", dim(kQ.cone), ".")
+  =#
```
HechtiDerLachs: Can this go?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1327 @@
...
+end
+
+function Base.show(io::IO,I::MonoidAlgebraIdeal)
+  print(io,underlying_ideal(I))
+end
+
+function Base.show(io::IO,J::InjMod)
+print(
+  io, "injective module given by direct sum of ", length(J.indec_injectives)," indecomposable injectives"
+)
```
HechtiDerLachs: Adjust indentation?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1327 @@
...
+
+function Base.show(io::IO, ::MIME"text/plain", J::InjMod)
+  println(io, "injective module given by direct sum of indecomposable injectives")
+  for Ji in J.indec_injectives
+    println(io, "  k{", Ji.vector, " + F - Q}, where p_F = ", Ji.face.prime)
+  end
+  print(
+    io,
+    "over ",
+    J.monoid_algebra
```
HechtiDerLachs, line 180:
```suggestion
    base_ring(J)
```
...and probably elsewhere?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1327 @@
...
+  irreducible_hull(Mi::SubquoModule, kQ::MonoidAlgebra, j=0)
+
+Returns an irreducible hull of a $\mathbb{Z}^d$-graded $k[Q]$-module M. It consists of indecomposable injectives $J_1,\dots,J_k$ and a $k$-matrix $\Lamdba$
+such that $Mi \xhookrightarrow{\Lambda} \sum{i=1}^k J_i$. 
+"""
+function irreducible_hull(Mi::SubquoModule{<:MonoidAlgebraElem}, j=0)
+  kQ = base_ring(Mi)
+  T = elem_type(kQ)
+
+  #initilalize
```
HechtiDerLachs:
```suggestion
  #initialize
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1327 @@
...
+    end
+
+    push!(irreducible_sums, Ji)
+    # TODO: Remove once bugs are gone!
+    if !isempty(cochain_maps) && !is_zero(compose(last(cochain_maps), hi))
+      @show length(cochain_maps)
+      @show last(cochain_maps)
+      @show hi
+      error()
+    end
```
HechtiDerLachs:
```suggestion
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,648 @@
...
+  kQ = base_ring(M)
+  k = coefficient_ring(kQ)
+
+  @req kQ == I.monoid_algebra "module and ideal must be over same monoid algebra"
+  @req is_normal(kQ) "monoid algebra must be normal"
+
+  #compute injective resolution
+  inj_res = injective_resolution(M, i+1)
+
+  #initilalize sector partition
```
HechtiDerLachs:
```suggestion
  #initialize sector partition
```

experimental/InjectiveResolutions/test/MonoidAlgebra.jl
```diff
@@ -0,0 +1,39 @@
...
+    return A(rand(A.algebra, -a:a, 0:a, 0:a))
+  end
+
+  # get MonoidAlgebra
+  kQ = monoid_algebra([[0, 1], [1, 1], [2, 1]],QQ)
+  Oscar.ConformanceTests.test_Ring_interface(kQ)
+  #Oscar.ConformanceTests.test_Ring_interface_recursive(kQ) # Needs AA update
+end
+
+@testset "constuct MonoidAlgebras" begin
```
HechtiDerLachs, line 17: TODO: Adjust indentation in tests!

HechtiDerLachs:
Tests pass!!1!!11!

HechtiDerLachs:
@lgoettgens : You mentioned something about adjustments of the structure of the documentation. Are you happy with the current state?

lgoettgens:
> @lgoettgens : You mentioned something about adjustments of the structure of the documentation. Are you happy with the current state?

Looks fine from what I can see on github. If you tried to build the docs locally and are satisfied with it, that's good.

experimental/InjectiveResolutions/docs/src/introduction.md
```diff
@@ -0,0 +1,180 @@
...
+points in a cone over a square.  This is also the coordinate ring of the 
+Segre embedding $\mathbb{P}^1\times \mathbb{P}^1 \to \mathbb{P}^3$.
+
+```julia
+kQ = monoid_algebra([[1, 0, 0], [1, 1, 0], [1, 1, 1], [1, 0, 1]], QQ)
+# monoid algebra over rational field with cone of dimension 3
+
+
+a, b, c, d = gens(kQ)
+# 4-element Vector{MonoidAlgebraElem{QQFieldElem, MonoidAlgebra{QQFieldElem, MPolyQuoRing{MPolyDecRingElem{QQFieldElem, QQMPolyRingElem}}}}}:
```
fingolfin: The following is *not* meant to block this PR, it can be addressed in a follow-up PR.

The drawback of this running example (which otherwise seems great to me!) is that it is not a `jldoctest`, and therefore won't get tested automatically. Unfortunately based on experience, the typical outcome of this is that at some point in the future, this example will stop working without anybody noticing, except for users trying to follow it manually, who then get confused by it.

I would therefore like to encourage you to consider rewriting this using `jldoctest`.

To be able to use multiple doctests blocks which "share definitions", you can label them as explained in the Documenter.jl documentation, [see here](https://documenter.juliadocs.org/stable/man/doctests/#Preserving-Definitions-Between-Blocks).

annahofer00: I actually wrote it as a `jldoctest` first, but I didn't know how to have multiple blocks with shared definitions... So thanks for the comment! I just rewrote and pushed some other minor adjustments.

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+using ..Oscar
+using ..Oscar: IntegerUnion # add other things that are not exported from Oscar here
+# functions with new methods
+import ..Oscar:
+  coefficient_ring,
+  gens,
+  ModuleGens,
+  SubModuleOfFreeModule,
+  is_zm_graded,
+grading_group,
```
fingolfin:
```suggestion
  grading_group,
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+function Base.show(io::IO, ::MIME"text/plain", SP::SectorPartitionLC)
+  println(
+    io,
+    "sector partition of ",
+    SP.i,
+    "-th local cohomology module supported on ideal (",
+    join(gens(SP.I), ", "),
+    ") of ",
+  )
+  println(io, " ", SP.M)
```
fingolfin: Found thanks to your new `jldoctest` which revealed the printing issue ;-)
```suggestion
  print(io, " ", SP.M)
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  inj_mods::Vector{InjMod}
+  cochain_maps::Vector{MatElem}
+  upto::Int
+  Q_graded_part::IrrRes
+  shift::Vector{Int} #not needed
+end
+
+function Base.show(io::IO,J::InjMod)
+  print(
+    io, "injective module given by direct sum of ", length(J.indec_injectives)," indecomposable injectives"
```
fingolfin, line 168:
```suggestion
    io, "Injective module given by direct sum of ", length(J.indec_injectives)," indecomposable injectives"
```
and similar elsewhere.

Note that to get this turned into lowercase for recursive printing, one can use `Lowercase()` as described in https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/#Advanced-printing-functionality

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  )
+end
+
+function Base.show(io::IO, ::MIME"text/plain", J::InjMod)
+  println(io, "injective module given by direct sum of indecomposable injectives")
```
fingolfin, line 173:
```suggestion
  println(io, "Injective module given by direct sum of indecomposable injectives")
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  end
+  print(
+    io,
+    "over ",
+    J.monoid_algebra
+  )
+end
+
+function Base.show(io::IO, ::MIME"text/plain", res::InjRes)
+  println(io, "injective resolution ")
```
fingolfin, line 185:
```suggestion
  println(io, "Injective resolution ")
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  println(io, "  ", join(["J^$i" for i in 0:res.upto], " -> "))
```
fingolfin, line 186: This looks as if it should use `Indent()` instead of the initial "  ", see https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/#Advanced-printing-functionality

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  println(io, "where ")
+  for i in eachindex(res.inj_mods)
+    j = i-1
+    println(io, " J^$j = direct sum of")
+    for Ji in res.inj_mods[i].indec_injectives
+      println(io, "    k{", Ji.vector, " + F - Q}, where p_F = ", Ji.face.prime)
+    end
+  end
+  println(io, "of ", res.mod)
+  println(
```
fingolfin:
```suggestion
  print(
```
and similar elsewhere: your `show` methods should never end with a `println` / a trailing newline.

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+end
+
+function Base.show(io::IO, Ji::IndecInj)
+  print(
+      io, "indecomposable injective k{", Ji.vector, " + F - Q}, where p_F = ", Ji.face.prime
+    )
+end
+
+@doc raw"""
+  generators_W_H(kQ::MonoidAlgebra, H::HyperplaneQ, a::Vector{Int})
```
fingolfin: Per convention these signature descriptions should be indented with 4 spaces -- this is because the docstring is parsed as Markdown, and the 4 space indent marks a code block. 
```suggestion
    generators_W_H(kQ::MonoidAlgebra, H::HyperplaneQ, a::Vector{Int})
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+@doc raw"""
+    is_zero(S::SectorPartitionLC)
+
+Given a sector partition of a local cohomology module, check if the local cohomology module is zero.
+"""
+function is_zero(S::SectorPartitionLC)
+  return all([dim(s.H) == 0 for s in S.sectors])
+end
+
+function print_A_vector(io::IO, A::Vector{Vector{Int}})
```
fingolfin, line 44: Why you do you need this function? Isn't it equivalent to `show(io, A)` ?

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+  return dim(s.H) == 0
+end
+
+@doc raw"""
+    is_zero(S::SectorPartitionLC)
+
+Given a sector partition of a local cohomology module, check if the local cohomology module is zero.
+"""
+function is_zero(S::SectorPartitionLC)
+  return all([dim(s.H) == 0 for s in S.sectors])
```
fingolfin:
```suggestion
  return all(dim(s.H) == 0 for s in S.sectors)
```
... avoids an allocation of a vector for storing the intermediate result, can perform short circuiting (if the check fails for one `s` it can stop immediately)

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+This function computes a vector $\tau^j \in (\mathbb{Z}\cup \infty)^n$ for every indecomposable injective $J^j$.
+See [HM05](@cite) (Section 6).
+"""
+function compute_taus(kQ::MonoidAlgebra, J::IndecInj...)
+  # get linear functionals tau_i for every hyperplane bounding Q, i.e. for every facet
+  A, _ = halfspace_matrix_pair(facets(kQ.cone))
+  H = []
+  for i in 1:length(kQ.hyperplanes)
+    push!(H, [kQ.hyperplanes[i].hyperplane, A[i, :]])
+  end
```
fingolfin: It may not be too relevant in this function (I can't judge it), but be warned that whenever you use `[]` you are creating a `Vector{Any}` which tends to prevent type stability later in the function. Here at least it is easy enough to fix:
```suggestion
  H = [[kQ.hyperplanes[i].hyperplane, A[i, :]] for i in 1:length(kQ.hyperplanes)]
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+
+  # for every indecomposable injectives J_j = k{a_j + F_j - Q} compute the vector \tau^j used in Section 6 of HM05  
+  _tau = []
+  for J_i in J
+    in
```
fingolfin: Huh? this looks wrong?
```suggestion
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
fingolfin: (Thought it might be better to just call `get_halfspace_eq` here?)

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+  _A = [A_0, A_1, A_2]
+
+  # define vector spaces J_{S_A0}, J_{S_A1} and J_{S_A2}
+  F_0 = free_module(field, length(A_0))
+  F_1 = free_module(field, length(A_1))
+  F_2 = free_module(field, length(A_2))
+
+  #compute the maps by deleting rows and columns in phi and psi
+  #phi
+  phi_del = matrix(field, zeros(field, length(A_0), length(A_1)))
```
fingolfin:
```suggestion
  phi_del = zero_matrix(field, length(A_0), length(A_1))
```
and similarly elsewhere when you use `matrix` + `zeros`.

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+
+  if !is_empty(phi) # `is_empty` is true for the the m x 0-matrix
+    rows = []
+    for i in A_0 #delete i-th row of phi for i \notin A_0
+      push!(rows, phi[i, :])
+    end
+    _phi_del = transpose(hcat(rows...))
```
fingolfin: Perhaps this also works?
```suggestion
    _phi_del = transpose(phi[A_0, :])
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+
+    if all(>(0), size(_phi_del)) #not of size 0xn or nx0
+      columns = []
+      for i in A_1 #delete i-th column of phi for i \notin A_1
+        push!(columns, _phi_del[:, i - j])
+      end
+      if length(columns) > 0
+        phi_del = matrix(field, hcat(columns...))
+      end
```
fingolfin: Again, I may have missed something, but perhaps this also works?
```suggestion
      if length(A_1) > 0
        phi_del = matrix(field, _phi_del[:, A_1 - j])
      end
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+            push!(V, 0*zero(H_B))
+          end
+        end
+
+        # define map H_A -> H_B
+        f_AB = hom(H_A, H_B, V)
+
+        if is_zero(f_AB)
+          continue
+        else
```
fingolfin:
```suggestion
        if !is_zero(f_AB)
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+
+This follows Algorithm 6.4. in [HM05](@cite) and computes all needed maps (as described in the proof of [HM05](@cite) (Proposition 5.1)).
+"""
+function maps_needed(kQ::MonoidAlgebra, S_A::Vector{SectorLC})
+  k = coefficient_ring(kQ)
+  maps = []
+  for s1 in S_A, s2 in S_A # loop over all pairs
+    if s1 == s2
+      continue
+    end
```
fingolfin: This is fine, but more idiomatic would be this:
```suggestion
    s1 == s2 && continue
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,642 @@
...
+    if issubset(s2.A, s1.A) && s1.index_vector >= s2.index_vector # check if B \subseteq A and (l_1, ..., l_n) <= (l_1', ..., l_n')
```
fingolfin: Rule of thumb: the deeper code is nested, the harder it is to understand.

In this case the check here is followed by `... else continue end` almost 30 lines later. You could change this to do the `continue` here and then de-indent the code afterwards one level (or more if you also do this for the `if dim(...` right after this)
```suggestion
    issubset(s2.A, s1.A) || continue # check if B \subseteq A
    s1.index_vector >= s2.index_vector || continue # check if (l_1, ..., l_n) <= (l_1', ..., l_n')
```

experimental/InjectiveResolutions/src/ModuleFunctionality.jl
```diff
@@ -0,0 +1,301 @@
...
+ return SubquoModule(FN, shift.(ambient_representatives_generators(M)), shift.(relations(M)))
+ # The original code below was giving strange errors from somewhere in AA.
+ GN = free_module(R, ngens(M))
+ HN = free_module(R, length(relations(M)))
+ a = hom(GN, F, ambient_representatives_generators(M); check=false)
+ b = hom(HN, F, relations(M); check=false)
+ A = matrix(a)
+ B = matrix(b)
+ N = subquotient(FN, A, B)
+ return N
```
fingolfin, line 159: I'd recommend to either not keep this, or keep it commented out, else it'll be very easy to miss the early abort and be confused when studying the code.
```suggestion
```

Of course it would also be nice to know what the "strange errors" where and to debug them in AA but that's a separate issue of course :-)

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+
+is_graded(A::MonoidAlgebra) = true
+is_zm_graded(A::MonoidAlgebra) = true
+
+function polyhedral_cone(A::MonoidAlgebra)
+  if !isdefined(A, :polyhedral_cone)
+    A.polyhedral_cone = get_polyhedral_cone(A.algebra)
+  end
+  return A.polyhedral_cone
+end
```
fingolfin, line 54: If you don't need every nanosecond, you could also handle this with an attribute and subsequently also get rid of the `MonoidAlgebra::polyhedral_cone` field:
```suggestion
@attr Cone{QQFieldElem} polyhedral_cone(A::MonoidAlgebra) = get_polyhedral_cone(A.algebra)
```
The same of course holds for `cone`, `faces`, `hyperplanes`, ... (for each this could be asked separately of course)

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+
+coefficient_ring(A::MonoidAlgebra) = coefficient_ring(A.algebra)
+number_of_variables(A::MonoidAlgebra) = ngens(A.algebra)
+
+### Elements of MonoidAlgebras
+mutable struct MonoidAlgebraElem{CoeffType,ParentType} <: RingElem
+  parent::ParentType
+  elem::RingElem
+  # TODO: Do we want to store additional information on the elements here?
```
fingolfin, line 110: If you don't need to store anything else it might be nice if this could be a non-mutable type, as that can be optimized much better.
```suggestion
struct MonoidAlgebraElem{CoeffType,ParentType} <: RingElem
  parent::ParentType
  elem::RingElem
```
Alas then you'd need to get rid of the single-arg constructor. Which might be a good idea anyway, as it has unclear semantics?

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+
+  function MonoidAlgebraElem(
+    A::ParentType
+  ) where {CoeffType,ParentType<:MonoidAlgebra{CoeffType}}
+    return new{CoeffType,ParentType}(A)
```
fingolfin, line 115: Where do you need this? Couldn't it be
```suggestion
    return new{CoeffType,ParentType}(A, zero(A.algebra))
```
Then you could remove a ton of `isdefined` special cases below?

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+  return A()
+end
+
+function one(A::MonoidAlgebra)
+  return MonoidAlgebraElem(A, one(A.algebra); check=false)
+end
+
+function +(a::MonoidAlgebraElem, b::MonoidAlgebraElem)
+  A = parent(a)
+  @assert A === parent(b)
```
fingolfin, line 170:
```suggestion
  check_parent(a, b)
  A = parent(a)
```
also below

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+AbstractAlgebra.promote_rule(
+                             ::Type{MPolyQuoRingElem{PolyType}}, ::Type{T}
+                            ) where {CoeffType, PolyType<:MPolyDecRingElem{CoeffType}, T<:MonoidAlgebraElem{CoeffType}} = T
+
+AbstractAlgebra.promote_rule(::Type{Int}, ::Type{T}) where {T<:MonoidAlgebraElem} = T
+AbstractAlgebra.promote_rule(::Type{ZZRingElem}, ::Type{T}) where {T<:MonoidAlgebraElem} = T
+
+gens(A::MonoidAlgebra) = [MonoidAlgebraElem(A, x) for x in gens(A.algebra)]
+number_of_generators(A::MonoidAlgebra) = ngens(A.algebra)
+getindex(A::MonoidAlgebra, i::Int) = MonoidAlgebraElem(A, A.algebra[i])
```
fingolfin, line 286: This method is actually implemented generically; the actual function to implement is `gen`:
```suggestion
gen(A::MonoidAlgebra, i::Int) = MonoidAlgebraElem(A, gen(A.algebra, i))
```

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+function prime_to_face(
+  kQ::Union{MPolyRing,MPolyQuoRing}, zonotope::Polyhedron, F::Polyhedron
+)
+  gens = Vector{MPolyDecRingElem}()
+  for lp in lattice_points(zonotope)
+    if !(lp in F) #check if lattice point is in F
+      a_v = Vector{ZZRingElem}()
+      for a_p in lp # PointVector -> Vector
+        push!(a_v, a_p)
+      end
```
fingolfin:
```suggestion
      a_v = [a_p for a_p in lp] # PointVector -> Vector
```

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+# (works more general for polyhedra)
+# given a polyhedral cone, this function returns its faces
+# INPUT:    polyhedral cone C
+# OUTPUT:   set of faces of C
+function get_faces_of_polyhedral_cone(
+  kQ::Union{MPolyRing,MPolyQuoRing}, zonotope::Polyhedron, P::Polyhedron
+)
+  P_faces = Vector{Polyhedron}()
+  for i in 0:dim(P)
+    in
```
fingolfin:
```suggestion
```

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+julia> monoid_algebra(M_Q,QQ)
+monoid algebra over rational field with cone of dimension 2
+```
+"""
+function monoid_algebra(M_Q::Matrix{Int}, k::Field)
+  d = size(M_Q, 1)
+
+  # construct k[t_1,...,t_d]
+  t_vars = [Symbol("t_$i") for i in 1:d]
+  T, t = graded_polynomial_ring(k, t_vars)
```
fingolfin: See https://docs.oscar-system.org/stable/DeveloperDocumentation/caching/
```suggestion
  T, t = graded_polynomial_ring(k, t_vars; cached = false)
```

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+
+  # construct k[x_1,...,x_n] where n is the number of columns/generators
+  x_vars = [Symbol("x_$i") for i in 1:size(M_Q, 2)]
+  R, _ = graded_polynomial_ring(
+    k, x_vars; weights=[Vector(row) for row in eachcol(M_Q)]
```
fingolfin:
```suggestion
    k, x_vars; weights=[Vector(row) for row in eachcol(M_Q)], cached = false
```

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+function monoid_algebra(V_Q::Vector{Vector{Int}}, k::Field)
+  return monoid_algebra(Matrix{Int}(transpose(matrix(V_Q))), k)
+end
+
+function Base.show(io::IO,F::FaceQ)
+  print(io,"face corresponding to homogeneous prime ",lowercase(string(F.prime)))
+end
+
+function Base.show(io::IO, ::MIME"text/plain", kQ::MonoidAlgebra)
+  println(io, "monoid algebra over ", lowercase(string(coefficient_ring(kQ))),
```
fingolfin:
```suggestion
  print(pretty(io), "Monoid algebra over ", Lowercase(), coefficient_ring(kQ),
```

experimental/InjectiveResolutions/src/MonoidAlgebra.jl
```diff
@@ -0,0 +1,611 @@
...
+  R = base_ring(I)
+  F = graded_free_module(R,1)
+  e1 = F[1]
+  return quo_object(F, [x * e1 for x = gens(I)]) 
+end
+
+function Base.show(io::IO, ::MIME"text/plain", I::MonoidAlgebraIdeal)
+  println(
+    io, "ideal over $(base_ring(I)) generated by "*join(["$(x)" for x in gens(I)], ", ")
+  )
```
fingolfin: `join` can target an `io` object which is more efficient then first printing into a string. So this should do it:
```suggestion
  print(pretty(io), "Ideal over ", Lowercase(), base_ring(I), " generated by "
  join(io, gens(I), ", ")
```

fingolfin (commented):
I just had a quick look at this PR and left some feedback. None of it is mathematical, as I am not qualified for this (in particular, an *approving review* should come from the @oscar-system/algebraic-geometry team).

Rather all my comments are about "best practices" for Julia resp. OSCAR code. I'd be happy to provide more details or answer questions. Since this code meant for "experimental", it would be OK to address this post-merge (though ideally it should not be forgotten).

Based on the last triage meeting, @HechtiDerLachs would like to see this code merged in time for OSCAR 1.4.0 which would kinda mean by this Wednesday, which is very close. As I wrote, my comments could be addressed in a future PR. 

But note that there are at least two instances of `in` floating around freely in the code ; I don't know why this doesn't cause errors, but these two should likely be addressed before the merge?

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  free_resolution,
+  _reduce, 
+  singular_assure,
+  oscar_assure, 
+  _graded_kernel,
+  oscar_generators,
+  singular_poly_ring,
+  _simple_kernel,
+  _extend_free_resolution, 
+  free_show
```
fingolfin, line 54:
```suggestion
  free_show,
  irreducible_decomposition
```

experimental/InjectiveResolutions/src/InjectiveResolutions.jl
```diff
@@ -0,0 +1,1098 @@
...
+  @eval import Oscar.Orderings: $i
+  #@eval export $i
+end
+
+#=
+for i in names(Oscar; all=true)
+  !isdefined(Oscar, i) && continue
+  @eval import Hecke: $i
+end
+=#
```
fingolfin, line 82:
```suggestion

import ..Oscar.Singular: 
  FreeModule, 
  has_global_ordering,
  svector,
  Module
 


for i in names(Oscar)
  !isdefined(Oscar, i) && continue
  @eval import Oscar: $i
  #@eval export $i
end

for i in names(Oscar.Orderings)
  !isdefined(Oscar.Orderings, i) && continue
  @eval import Oscar.Orderings: $i
  #@eval export $i
end

#=
for i in names(Oscar; all=true)
  !isdefined(Oscar, i) && continue
  @eval import Hecke: $i
end
=#
```

HechtiDerLachs: I don't understand this last bit. What's the difference between before and after?

experimental/InjectiveResolutions/src/LocalCohomology.jl
annahofer00: I want that empty lists are printed as `[]` and not `Int64[]`.

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,638 @@
...
+  _A = [A_0, A_1, A_2]
+
+  # define vector spaces J_{S_A0}, J_{S_A1} and J_{S_A2}
+  F_0 = free_module(field, length(A_0))
+  F_1 = free_module(field, length(A_1))
+  F_2 = free_module(field, length(A_2))
+
+  #compute the maps by deleting rows and columns in phi and psi
+  #phi
+  phi_del = zero_matrix(field,length(A0),length(A1))
```
joschmitt:
```suggestion
  phi_del = zero_matrix(field,length(A_0),length(A_1))
```
Typo?

experimental/InjectiveResolutions/src/LocalCohomology.jl
```diff
@@ -0,0 +1,638 @@
...
+        push!(columns, _phi_del[:, i - j])
+      end
+      if length(columns) > 0
+        phi_del = matrix(field, hcat(columns...))
+      end
+    end
+  end
+
+  #psi
+  psi_del = zero_matrix(field,length(A1),length(A2))
```
joschmitt:
```suggestion
  psi_del = zero_matrix(field,length(A_1),length(A_2))
```

experimental/InjectiveResolutions/src/LocalCohomology.jl
annahofer00: yes... thanks!

lkastner:
@flenzen Could you please have a final go for a review?

wdecker (approved):
I suggest that if the authors and @flenzen agree, then we merge this now and implement further improvements with future PRs.

lgoettgens:
With https://github.com/oscar-system/Oscar.jl/pull/4909 merged, there are now some errors in CI. @HechtiDerLachs said he will have a look into this.

fingolfin:
Just to say, we discussed this in triage today, and here is the general verdict:
- we want this code, we will merge it, thank you for your patience in applying review
- e.g. once we resolve the CI issues (I think we can do that on our side, no need for @annahofer00 who is on a conference or @tom111 to get active)
- however it still needs further work (but that's fine, as it is going into experimental) e.g. regarding user interfaces and tests -- this can be done in follow-up PRs in the coming weeks
  - if @annahofer00 (or @tom111 or both) would be interested in visiting us for some days to jointly discuss and work on things (sometimes much more efficient than remote discussion), we'd be happy to host you (funding for travel & hotel is there)

lkastner (approved):

fingolfin:
@flenzen did not yet have time to review this, but @lkastner had a look and approves. So will merges this now, and if @flenzen or anyone else has additional comments, we can log them in issues (resp. add them in comments on the existing issue #4913) and they can be addressed in future PRs.

Thanks to @annahofer00 and @tom111 for sticking through with this!

tom111:
Great news! Thanks everyone! 🎉

PR #4105, opened 2024-09-17
Fix is_very_ample

paemurru:
The function `is_very_ample` was giving `false` when the variety was not complete.

```
julia> X = affine_space(NormalToricVariety, 2)
Normal toric variety

julia> D = divisor_of_character(X, [1, 0])
Torus-invariant, prime divisor on a normal toric variety

julia> is_very_ample(D)
false
```

### DIFF
--- a/src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
@@ -155,7 +155,7 @@ false
 """
 @attr Bool function is_very_ample(td::ToricDivisor)
   @req is_complete(toric_variety(td)) "Very ampleness test is only implemented for complete toric varieties. ([Def 6.1.9 CLS11])"
-  pm_object(td).VERY_AMPLE
+  return pm_object(td).VERY_AMPLE
 end
 
 
--- a/src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
+++ b/src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
@@ -155,7 +155,7 @@ false
 """
 @attr Bool function is_very_ample(td::ToricDivisor)
   @req is_complete(toric_variety(td)) "Very ampleness test is only implemented for complete toric varieties. ([Def 6.1.9 CLS11])"
-  return pm_object(td).VERY_AMPLE
+  return pm_object(td).VERY_AMPLE::Bool
 end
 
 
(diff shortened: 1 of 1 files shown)


src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
```diff
@@ -153,7 +153,10 @@ julia> is_very_ample(td)
 false
 ```
 """
-@attr Bool is_very_ample(td::ToricDivisor) = pm_object(td).VERY_AMPLE
+@attr Bool function is_very_ample(td::ToricDivisor)
+  @req is_complete(toric_variety(td)) "Very ampleness test is only implemented for complete toric varieties. ([Def 6.1.9 CLS11])"
+  pm_object(td).VERY_AMPLE
```
lgoettgens:
```suggestion
  return pm_object(td).VERY_AMPLE
```
we usually add the explicit `return` in multiline functions

benlorenz:
```suggestion
  return pm_object(td).VERY_AMPLE::Bool
```
Polymake properties should usually get a type annotation (as it cannot be inferred).

src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
```diff
@@ -153,7 +153,10 @@ julia> is_very_ample(td)
 false
 ```
 """
-@attr Bool is_very_ample(td::ToricDivisor) = pm_object(td).VERY_AMPLE
+@attr Bool function is_very_ample(td::ToricDivisor)
+  @req is_complete(toric_variety(td)) "Very ampleness test is only implemented for complete toric varieties. ([Def 6.1.9 CLS11])"
+  return pm_object(td).VERY_AMPLE
```
HereAround: @paemurru, can you please add the type assertion `::Bool`, as suggested by @benlorenz. Then this can be merged.

lkastner:
```suggestion
  return pm_object(td).VERY_AMPLE::Bool
```

lkastner (approved):

src/AlgebraicGeometry/ToricVarieties/ToricDivisors/properties.jl
paemurru: Hmm, I thought I committed the change of @benlorenz , but apparently not.

PR #4110, opened 2024-09-18
Some serialization of schemes

HechtiDerLachs:
Intermediate state of the art of our joint effort with @antonydellavecchia this week.

### DIFF
--- a/test/AlgebraicGeometry/Schemes/smoothness_test.jl
+++ b/test/AlgebraicGeometry/Schemes/smoothness_test.jl
@@ -1,0 +1,47 @@
+@testset "_is_smooth" begin
+  R, (x, y, z) = QQ[:x, :y, :z]
+
+  I1 = ideal(R, x*y)
+  IA3 = spec(R)
+  X1, _ = sub(IA3, I1)
+  @test !Oscar._is_smooth(X1)
+  @test !Oscar._is_smooth(X1; jacobian_cut_off=0)
+
+  I2 = ideal(R, x*y^2)
+  X2, _ = sub(IA3, I2)
+  @test !Oscar._is_smooth(X2)
+  @test !Oscar._is_smooth(X2; jacobian_cut_off=0)
+
+  A = R[x y 0; 0 y z]
+  I3 = ideal(R, minors(A, 2))
+  X3, _ = sub(IA3, I3)
+  @test !Oscar._is_smooth(X3)
+  @test !Oscar._is_smooth(X3; jacobian_cut_off=0)
+
+  A = R[x y 0; 0 y-1 z]
+  I4 = ideal(R, minors(A, 2))
+  X4, _ = sub(IA3, I4)
+  @test !Oscar._is_smooth(X4)
+  @test !Oscar._is_smooth(X4; jacobian_cut_off=0)
+
+  I5 = ideal(R, [x, y]) * ideal(R, [x, y-1, z])
+  X5, _ = sub(IA3, I5)
+  @test Oscar._is_smooth(X5)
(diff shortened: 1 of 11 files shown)


src/AlgebraicGeometry/Schemes/PrincipalOpenSubset/Objects/Types.jl
```diff
@@ -33,8 +33,21 @@
...
       check::Bool=true
     ) where {T<:RingElem}
-    d = prod(length(f) > 0 ? f : one(OO(X)))
-    return PrincipalOpenSubset(X, R, lifted_numerator.(f), d, check=check)
+    @check begin
+        d = prod(length(f) > 0 ? f : one(OO(X)))
+        U = spec(R)
+        U == hypersurface_complement(X, d)
+    end
+    return new{base_ring_type(X), ring_type(U), typeof(X)}(X, U, lifted_numerator.(f))
```
simonbrandhorst: this looks fishy to me. `U` is not defined if check=false

HechtiDerLachs: You're right. I'll look into this.

test/Serialization/IPC.jl
```diff
@@ -24,4 +24,28 @@ end
...
+  X, _ = sub(IP2, I)
+  X_cov = covered_scheme(X)
+  U = affine_charts(X_cov)
+  for a in U
+    Oscar.put_params(channels, ambient_coordinate_ring(a))
+    Oscar.put_params(channels, OO(a))
+  end
+  results = pmap(is_smooth, U)
+  @test results == [1, 1, 0]
+end
```
HechtiDerLachs: ping @simonbrandhorst , @afkafkafk13 .

afkafkafk13:
Can we merge #4121 first and then get a better view on what's going on here? It seems that this is built on top of the other.

HechtiDerLachs:
@antonydellavecchia : I did another implementation which broadcasts all rings to all workers using `put_params`. That seems to work. With the new smoothness test implementation we verify smoothness of @wdecker's surface in roughly 35 seconds on my machine with three workers.

test/AlgebraicGeometry/Schemes/smoothness_test.jl
```diff
@@ -0,0 +1,47 @@
+@testset "_is_smooth" begin
+  R, (x, y, z) = QQ[:x, :y, :z]
+
+  I1 = ideal(R, x*y)
+  IA3 = spec(R)
+  X1, _ = sub(IA3, I1)
+  @test !Oscar._is_smooth(X1)
+  @test !Oscar._is_smooth(X1; jacobian_cut_off=0)
```
HechtiDerLachs, line 8: @afkafkafk13 : I think I have now implemented a truly generic smoothness test here which can also handle non-reduced, non-equidimensional, and non-connected cases. I'm still a bit skeptic, since all the classical implementations seem to solve this via some primary/equidimensional decomposition which I seem to be able to fully avoid. If you have the time: Could you have a look and tell me what you think? Maybe you have some further exotic example in mind for which my implementation fails. If so, it would be good to know about it.

HechtiDerLachs:
ping @danteluber : Check out this PR and try the following.
```julia
using Distributed
process_ids = addprocs(4)
@everywhere using Oscar
```
That starts the workers. Then you can do
```
Oscar._is_smooth_parallel(realization_space(fano_matroid(); ground_ring=QQ))
```
Replace `fano_matroid()` by anything more fancy which you want to tackle.

HechtiDerLachs:
This has been integrated into Oscar via #4773 and #4780.

PR #4111, opened 2024-09-18
Disable anonymous functions in gluings

HechtiDerLachs:
This is some independent work carried out on #4110 .

It removes the anonymous functions inside the `LazyGluing`s. These are not necessary and they don't serialize nicely. So I took them out. 

In the course of this I disabled the possibility to "cheaply" compute only the gluing domains. This turned out not to be useful and it just made things more complicated.

### DIFF
--- a/experimental/Schemes/src/Auxiliary.jl
+++ b/experimental/Schemes/src/Auxiliary.jl
@@ -254,7 +254,7 @@ struct InheritGluingData
   Y::AbsAffineScheme
 end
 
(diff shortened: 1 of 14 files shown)


HechtiDerLachs:
Tests seem to pass, but the booktests. But that looks rather like a hickup, doesn't it?

joschmitt (approved):

afkafkafk13 (approved):
Thank you! Makes sense and does the job!

PR #4114, opened 2024-09-19
Potential switch to sparse matrices for monomial orderings

HechtiDerLachs:
This is a draft for taking some first steps towards using sparse matrices for the `canonical_matrix` of a `ModuleOrdering`.

Today, @jankoboehm and myself investigated the caching of module gbs a little bit. We found that the production of the `canonical_matrix` for the `ModuleOrdering`s is quite expensive, when the module has many generators (in our example 172). This is probably due to the matrix being dense. I tried to take some steps towards replacing it by a sparse matrix, but reached a point where I couldn't figure out any more how to proceed. However, I thought it might be useful to report on what I did so far and hence I make this draft PR. Let us eventually look into this together.

CC: @ederc @RafaelDavidMohr


thofma:
The old code has a lot of root for improvements (for example, not using `is_zero` but `is_zero_entry`). Do you have an example for benchmarking?

P.S.: I am asking, because 200x200 is not a size where I would expect large improvements from a sparse representation.

HechtiDerLachs:
> Do you have an example for benchmarking?

Sure:

```julia
R, (x, y, z) = QQ[:x, :y, :z]
F = FreeMod(R, 500)
I, _ = sub(F, gens(F))
@time standard_basis(I)
```
```
julia> @time standard_basis(I);
  2.762075 seconds (22.04 M allocations: 841.855 MiB, 36.02% gc time)
```

joschmitt:
I now have this timing:
```
julia> @time standard_basis(I);
  0.042802 seconds (320.76 k allocations: 15.418 MiB)
```

joschmitt:
If we want this, we would have to (also) allow sparse matrices in `MatrixOrdering`, so probably add a type parameter to the struct etc. I do not know whether it makes sense to store these matrices sparse; in my use cases they never get big.
The reason why it was so slow in the example wasn't necessarily the size of the matrix, but more the communication with the flint side see #4117.

fingolfin:
Let's merge PR #4117 first, as it is the more conservative (= safer? maybe?) change. Then we can consider whether and how to get this one here over the finishing line. Thank you for starting it in any case!!!

fingolfin:
@ederc will talk to @RafaelDavidMohr about this later (but Rafael is somewhat busy right now with defense prep)

lgoettgens:
This has some non-trivial conflicts. Someone (maybe @HechtiDerLachs) should try to resolve them

fingolfin:
What the status of this PR?

PR #4118, opened 2024-09-19
More bib cleanup

lgoettgens:
DocumenterCitations.jl prefers to get unicode. Additional braces to keep casing are not needed as well.

Preview: https://docs.oscar-system.org/previews/PR4118/references/


benlorenz (approved):
preview looks good

PR #4119, opened 2024-09-19
Speed up `is_homogeneous`

fingolfin:
Some measurements taking the example from the `is_homogeneous` docstring:

    R, (x, y, z) = graded_polynomial_ring(QQ, ["x", "y", "z"], [1, 2, 3])
    f = x^2+y*z
    W = [1 2 1 0; 3 4 0 1]
    S, (w, x, y, z) = graded_polynomial_ring(QQ, ["w", "x", "y", "z"], W)
    F = w^3*y^3*z^3 + w^2*x*y^2*z^2 + w*x^2*y*z + x^3

Before:

    julia> @btime is_homogeneous(f);
      1.533 μs (62 allocations: 2.28 KiB)

    julia> @btime is_homogeneous(F);
      3.932 μs (152 allocations: 5.53 KiB)

After:

    julia> @btime is_homogeneous(f);
      334.071 ns (10 allocations: 416 bytes)

    julia> @btime is_homogeneous(F);
      556.150 ns (12 allocations: 656 bytes)

I will make a separate PR for Hecke which provides `addmul!`, `zero!` and more for `FinGenAbGroupElem`.

Also I think there are more places that could be similarly improved, e.g. `homogeneous_components`. However, unlike `is_homogeneous` they did not stick out in some timing profiles on CI tests we made this week.


HechtiDerLachs (approved):
If it works, it works. Thanks!

lgoettgens (approved):

joschmitt:
I now get 10 respectively 12 allocations in the example.

src/Rings/mpoly-graded.jl
```diff
@@ -990,7 +990,7 @@ function is_homogeneous(F::MPolyDecRingElem)
       #addmul!(u, d[i], c[i])
     end
     if S === nothing
-      S = u
+      S = deepcopy(u)
```
fingolfin, line 982: Ah indeed, thanks! I didn't use `zero!` initially

src/Rings/mpoly-graded.jl
```diff
@@ -978,14 +978,20 @@ true
...
   for c = MPolyExponentVectors(forget_decoration(F))
-    u = parent(F).D[0]
+    zero!(u.coeff)
+    # TODO: once Hecke supports zero! on FinGenAbGroupElem, switch to this:
+    #zero!(u)
     for i=1:length(c)
-      u += c[i]*d[i]
+      addmul!(u.coeff, d[i].coeff, c[i])
+      # TODO: once Hecke supports addmul! on FinGenAbGroupElem, switch to this:
+      #addmul!(u, d[i], c[i])
```
joschmitt: This is not correct:
```
julia> G = abelian_group([2, 2])
(Z/2)^2

julia> u = G([0, 0])
Abelian group element [0, 0]

julia> d = G([0, 1])
Abelian group element [0, 1]

julia> addmul!(u.coeff, d.coeff, ZZ(2))
[0   2]

julia> u
Abelian group element [0, 2]
```
In `G` there shouldn't be an element with coefficients `[0, 2]`.

fingolfin: Indeed, a final reduction is necessary. See also my Hecke PR adding a helper for that.

fingolfin: Assuming https://github.com/thofma/Hecke.jl/pull/1619 is merged and released soonish, can switch to that, too...

fingolfin: This is now resolved. I've also updated the measurements in the PR description (we allocate a tad more and things got slightly slower again, but it's still a clear win)

lgoettgens:
Maybe we wait for a release of https://github.com/thofma/Hecke.jl/pull/1619 and then can simplify the comments away?

fingolfin:
But that then also needs to go into a new Hecke release, and then more benchmarking should be done to verify it doesn't get slower again etc. I'd rather have this in fact sooner. IMHO the "simplification" is not as important as the speed up.

lgoettgens (approved):
Alright. Then let's get this in

lgoettgens:
Hecke 0.34.3 is now released with the required changes.

lgoettgens:
I now adapted to the new Hecke functions, and verified that this does not increase the allocations nor the runtime of the example above.

@fingolfin if you are happy with this, please merge

PR #4120, opened 2024-09-19
group attribute saving

antonydellavecchia:
Storing Gap groups with their attributes


src/Serialization/Groups.jl
```diff
@@ -154,7 +187,7 @@ end
 # We do the same for full free groups, subgroups of free groups,
 # full f.p. groups, and subgroups of f.p. groups.
 
-@register_serialization_type FPGroup uses_id
+@register_serialization_type FPGroup uses_id
```
lgoettgens:
```suggestion
@register_serialization_type FPGroup uses_id
```

lgoettgens (approved):

PR #4121, opened 2024-09-19
Remove anonymous functions from sheaves

HechtiDerLachs:
Also fixes the issue pointed out by @simonbrandhorst on #4110 .


afkafkafk13 (commented):
I had always been confused by your production_function and restriction_function arguments. This is now a lot clearer (thanks to the needs of serialization efforts).

Overall this looks good, but moving stuff at the same time as making this change was not overly helpful for reviewing.

afkafkafk13:
Let me know, wether you are still adding stuff to this or whether you would like a complete review.

HechtiDerLachs:
Thanks for having a look. Unfortunately, moving the functions was unavoidable. Yes, a full review would be appreciated. Nothing changed about the internal mathematics, though.

src/AlgebraicGeometry/Schemes/Sheaves/Sheaves.jl
```diff
@@ -59,7 +59,7 @@ function produce_object(F::AbsPreSheaf, U)
 end
 
 ### temporary workaround
-produce_object(F::AbsPreSheaf, U::AbsAffineScheme) = production_func(F)(F, U)
+produce_object(F::AbsPreSheaf, U::AbsAffineScheme) = produce_object(underlying_presheaf(F), U)
```
afkafkafk13: Is the comment about the temporary workaround still valid? If not, please change.

src/AlgebraicGeometry/Schemes/Sheaves/Types.jl
```diff
@@ -44,19 +44,15 @@ A basic minimal implementation of the interface for `AbsPreSheaf`; to be used in
 
   # production functions for new objects
   is_open_func::Function # To check whether one set is open in the other
-  production_func::Union{Function, Nothing} # To produce ℱ(U) for U ⊂ X
-  restriction_func::Union{Function, Nothing} # To produce the restriction maps ℱ(U) → ℱ(V) for V ⊂ U ⊂ X open
```
afkafkafk13: here again: please check whether the comment still reflects the current situation and change, if necessary. We do not want to confuse programmers who take a look at this as a model.

afkafkafk13 (approved):
I marked two places where comments should be changed to reflect the current implementation. Otherwise this is good to go.

PR #4124, opened 2024-09-19
Speed up polynomial mappings

HechtiDerLachs:
@fingolfin mentioned in his course this week that in the `RingFlattening`s there was lots of potential improvements available via the use of `MPolyBuildCtx`. I tried a little bit, but could not achieve any better performance. I'm wondering why, though, so I put it up here for discussion.


src/Rings/MPolyMap/MPolyRing.jl
```diff
@@ -132,12 +132,47 @@ end
...
+is_gen(x::MPolyQuoRingElem) = is_gen(lift(x))
+is_gen(x::MPolyDecRingElem) = is_gen(forget_grading(x))
+
+function _evaluate_plain(F::MPolyAnyMap{<:MPolyRing, <:MPolyRing}, u)
+  if isdefined(F, :variable_indices)
+    S = codomain(F)::MPolyRing
+    r = ngens(S)
+    ctx = MPolyBuildCtx(S)
+    for (c, e) in zip(AbstractAlgebra.coefficients(u), AbstractAlgebra.exponent_vectors(u))
+      ee = [0 for _ in 1:r]
```
simonbrandhorst:
```suggestion
      ee = Vector{Int}(undef, r)
```
should be an epsilon faster

joschmitt: One should be able to use the "same" vector `ee` for the whole loop as it gets copied anyway. I try to optimize this further.

HechtiDerLachs: An undefined vector will not work. We need the zeroes.

src/Rings/MPolyMap/MPolyRing.jl
```diff
@@ -132,12 +132,47 @@ end
 #
 ################################################################################
 
+# Some additional methods needed for the test in the constructor for MPolyAnyMap
+is_gen(x::MPolyQuoRingElem) = is_gen(lift(x))
+is_gen(x::MPolyDecRingElem) = is_gen(forget_grading(x))
```
HechtiDerLachs: Memo to self: This is not legitimate. We need to switch to some internal method like `_is_gen` for that.

src/Rings/MPolyMap/MPolyAnyMap.jl
```diff
@@ -42,17 +42,37 @@ const _DomainTypes = Union{MPolyRing, MPolyQuoRing}
   coeff_map::U
   img_gens::Vector{V}
   temp_ring           # temporary ring used when evaluating maps
+  variable_indices::Vector{Int} # a table where the i-th entry contains the 
+                                # index of the variable where it's mapped to
```
fingolfin:
```suggestion
                                # index of the variable where it is mapped to
```

src/Rings/MPolyMap/MPolyAnyMap.jl
```diff
@@ -42,17 +42,37 @@ const _DomainTypes = Union{MPolyRing, MPolyQuoRing}
...
+                                   coeff_map::U,
+                                   img_gens::Vector{V}) where {D <: Union{<:MPolyRing, <:MPolyQuoRing}, 
+                                                               C <: Union{<:MPolyRing, <:MPolyQuoRing}, 
+                                                               U, V}
+    @assert V === elem_type(C)
+    for g in img_gens
+      @assert parent(g) === codomain "elements does not have the correct parent"
+    end
+    result = new{D, C, U, V}(domain, codomain, coeff_map, img_gens)
+    if all(is_gen(x) for x in img_gens) && allunique(img_gens)
```
fingolfin:
```suggestion
    if all(is_gen, img_gens) && allunique(img_gens)
```

src/Rings/MPolyMap/MPolyAnyMap.jl
```diff
@@ -42,17 +42,37 @@ const _DomainTypes = Union{MPolyRing, MPolyQuoRing}
...
+      result.variable_indices = [findfirst(x==y for y in gens(codomain)) for x in img_gens]
```
fingolfin:
```suggestion
      result.variable_indices = [findfirst(==(x), gens(codomain)) for x in img_gens]
```

HechtiDerLachs:
Unfortunately this still does not seem to make a difference. I was running the tests for elliptic surfaces with and without the changes made here (since @fingolfin pointed to that as a potential example). The result is

Without this PR:
```
500.418518 seconds (2.93 G allocations: 112.836 GiB, 12.64% gc time, 0.28% compilation time)
```
With this PR:
```
508.444454 seconds (2.90 G allocations: 111.645 GiB, 12.23% gc time, 0.16% compilation time)
```
The different timings might also be due to randomization of parts of the tests on the lattice theoretic side (@simonbrandhorst : Is this correct?), but the allocations do not go down significantly. 

One thing I noticed is that `finish` does a lot of sorting (which has already been pointed out to me by @lgoettgens ). This creates some overhead. But in general, another problem I witnessed with the `RingFlattening`s is that we have massive `is_zero` checks whenever we do arithmetic with `SRow`s. These can not be avoided. But when we do this with a quotient ring, then we always need to translate the representative to the `Singular`-side. If this quotient ring is put to work via flattenings, then additionally we first have to translate to the flattened ring. This kills performance for sure. 

In my opinion it would therefore be desirable to have a `singular_representative` in a `MPolyQuoRingElem` and to do the arithmetic with these by default whenever they are defined. Only memorizing whether an element is reduced does not help in these cases. 

Orthogonal to that, the question remains: Why does this PR not help speeding things up for mappings?

HechtiDerLachs:
So: In the end this is useful. Just not for all the flattenings used in the elliptic surfaces (yet). I created the following [test polynomial](https://github.com/user-attachments/files/17129328/poly.txt).

Run the following code:
```julia
g = load("poly.txt")
S = parent(g)
flat = Oscar.flatten(S)
f = copy(g);
@time flat(f);
```
With this PR:
```

julia> f = copy(g);

julia> @time flat(f);
  0.335777 seconds (686.51 k allocations: 49.636 MiB)

julia> f = copy(g);

julia> @time flat(f);
  0.334014 seconds (690.14 k allocations: 48.630 MiB)

julia> f = copy(g);

julia> @time flat(f);
  0.328450 seconds (705.82 k allocations: 49.011 MiB)

julia> f = copy(g);

julia> @time flat(f);
  0.330575 seconds (711.82 k allocations: 49.188 MiB)
```
without this PR:
```
julia> f = copy(g);

julia> @time flat(f);
  4.988692 seconds (975.34 k allocations: 654.598 MiB, 27.35% gc time)

julia> f = copy(g);

julia> @time flat(f);
  5.196591 seconds (750.61 k allocations: 652.844 MiB, 27.13% gc time)

julia> f = copy(g);

julia> @time flat(f);
  5.379392 seconds (673.50 k allocations: 652.256 MiB, 26.06% gc time)

julia> f = copy(g);

julia> @time flat(f);
  5.556177 seconds (659.54 k allocations: 652.183 MiB, 25.55% gc time)

```

thofma:
Looks good so far. I think this is going in the right direction. (It reminds of the `MPolyHom_vars` we once had and some ideas of Dan).

The logic for the evaluation was already hard to follow before (blame on me), and it seems it won't be getting better. But it is easer to talk in persons about this.

HechtiDerLachs:
I added a keyword argument to the internal constructor for the maps to be able to avoid the internal checks whether you have a mapping of variables to variables. These are not yet forwarded to the user-facing constructors, but should this ever become a performance issue, this possibility exists.

HechtiDerLachs:
Looks like we're slowly getting somewhere here. I ran the `elliptic_surfaces.jl` test and got the following:
Before:
```
560.589288 seconds (3.01 G allocations: 119.270 GiB, 10.71% gc time, 21.51% compilation time)
```
after:
```
484.957044 seconds (2.48 G allocations: 92.375 GiB, 11.66% gc time, 0.02% compilation time)
```
ping: @fingolfin

src/AlgebraicGeometry/Schemes/AffineSchemes/Morphisms/Attributes.jl
```diff
@@ -46,7 +46,7 @@ Affine scheme morphism
   from [x1, x2, x3]  scheme(x1)
   to   [x1, x2, x3]  affine 3-space over QQ
 given by the pullback function
-  x1 -> 0
+  x1 -> x1
```
lgoettgens, line 49: Why do these doctests change? There is no change in the code above that would explain it

HechtiDerLachs: We don't reduce the generators anymore when there's a "variable-to-variable" map. Mathematically all these are the same.

HechtiDerLachs:
@HereAround : Might it be the case that your test in the `FTheoryTools` has some randomness? See [this commit](https://github.com/oscar-system/Oscar.jl/pull/4124/commits/0c65e56d996201f535647cb7d578c2742d38426a) with the local changes from my machine. 

Anyways: If the tests pass, then from my side, this is good to go.

lgoettgens:
> @HereAround : Might it be the case that your test in the `FTheoryTools` has some randomness? See [this commit](https://github.com/oscar-system/Oscar.jl/pull/4124/commits/0c65e56d996201f535647cb7d578c2742d38426a) with the local changes from my machine.
> 
> Anyways: If the tests pass, then from my side, this is good to go.

Unfortunately these kind of doctests (printing a large vector) are dependent on the windows size of the terminal you run the doctest_fix from

src/Rings/MPolyMap/MPolyRing.jl
```diff
@@ -132,17 +132,122 @@ end
...
+    S = base_ring(codomain(F))::MPolyRing
+    r = ngens(S)
+    ctx = MPolyBuildCtx(S)
+    for (c, e) in zip(coefficients(u), exponents(u))
+      ee = [0 for _ in 1:r]
+      for (i, k) in enumerate(e)
+        ee[F.variable_indices[i]] = k
+      end
+      push_term!(ctx, c, ee)
+    end
```
thofma: Any chance we can deduplicate this part? It seems to be repeated a few times.

Should probably be something like
```
function _bla_bla(S, variable_indices, u)
...
```
?

thofma:
> I added a keyword argument to the internal constructor for the maps to be able to avoid the internal checks whether you have a mapping of variables to variables. These are not yet forwarded to the user-facing constructors, but should this ever become a performance issue, this possibility exists.

Why does this exist if it is not used here?

src/Rings/MPolyMap/MPolyRing.jl
HechtiDerLachs: Probably, yes. I'll look into it.

HechtiDerLachs:
> Why does this exist if it is not used here?

I thought, it would be nice to give future developers a hint that they can safely disable this feature by using the keyword argument. If you think, it's causing more confusion than insight, then I can remove it. Also, it seems that we can reunite both internal constructors into one constructor again, now that I have allowed more rings in the codomain.

thofma:
> I thought, it would be nice to give future developers a hint that they can safely disable this feature by using the keyword argument. If you think, it's causing more confusion than insight, then I can remove it. Also, it seems that we can reunite both internal constructors into one constructor again, now that I have allowed more rings in the codomain.

I had hoped that a no one would ever create a map by calling `MPolyAnyMap` directly :) I am happy to keep it. I just wanted to ask why this unused (and not tested) feature exists.

HereAround:
> @HereAround : Might it be the case that your test in the `FTheoryTools` has some randomness? See [this commit](https://github.com/oscar-system/Oscar.jl/pull/4124/commits/0c65e56d996201f535647cb7d578c2742d38426a) with the local changes from my machine. 
> 
> Anyways: If the tests pass, then from my side, this is good to go. 

The computation of generic_section of line bundles picks random coefficients. This is the element of randomness in the FTheoryTools test. Everything else is (or so I hope at least) deterministic.

src/Rings/mpolyquo-localizations.jl
```diff
@@ -2717,3 +2717,90 @@ function dim(R::MPolyQuoLocRing)
...
+      push_term!(ctx, kk(c), ee+d)
+    end
+  end
+  return Q(finish(ctx))
+end
+
+# The following methods are only safe, because they are called 
+# exclusively in a setup where variables map to variables 
+# and no denominators are introduced. 
+_coefficients(x::MPolyRingElem) = coefficients(x)
```
thofma: Not sure, but maybe you want the following here?
```suggestion
_coefficients(x::MPolyRingElem) = AbstractAlgebra.coefficients(x)
```
And also below with `exponent_vectors`.

HechtiDerLachs: I saw the `AA`-qualifier elsewhere, too. But I thought, it was redundant. Isn't it?

thofma: No. It is the difference between painfully slow and fast when iterating over coefficients/exponents. There used to be an explanation in the documentation, but it was removed for reasons I don't know.

thofma: See https://github.com/oscar-system/Oscar.jl/issues/4168

HechtiDerLachs:
Again: Ready for merge from my side.

thofma (approved):

thofma:
Thanks, it looks good.

src/Rings/MPolyMap/MPolyAnyMap.jl
```diff
@@ -42,16 +42,31 @@ const _DomainTypes = Union{MPolyRing, MPolyQuoRing}
...
+      @assert parent(g) === codomain "elements does not have the correct parent"
+    end
+    result = new{D, C, U, V}(domain, codomain, coeff_map, img_gens)
+    # If it ever turned out that doing the checks within the following if-block
+    # is a bottleneck, consider passing on the `check_for_mapping_of_vars` kw 
+    # argument to the outer constructors or make the outer constructors 
+    # call the inner one with this argument set to `false`. This way the check 
+    # can safely be disabled.
+    if check_for_mapping_of_vars && all(_is_gen, img_gens) && _allunique(img_gens)
+      result.variable_indices = [findfirst(==(x), gens(codomain)) for x in img_gens]
```
thofma: This recomputes `gens(codomain)` over and over again. Feel free to ignore this. I just noticed it.

simonbrandhorst:
```suggestion
      gens_codomain = gens(codomain)
      result.variable_indices = [findfirst(==(x), gens_codomain) for x in img_gens]
```

PR #4128, opened 2024-09-20
Get rid of trivial anonymous functions (part 2)

lgoettgens:
See https://github.com/oscar-system/Oscar.jl/pull/4126 for part 1 and the reasoning behind these changes.

### DIFF
--- a/src/Modules/ExteriorPowers/FreeModules.jl
+++ b/src/Modules/ExteriorPowers/FreeModules.jl
@@ -45,7 +45,7 @@ function exterior_power(F::FreeMod, p::Int; cached::Bool=true)
 
   function my_decomp(u::FreeModElem)
     @req parent(u) === result "element does not belong to the correct module"
-    k = findfirst(x -> x == u, gens(result))
+    k = findfirst(==(u), gens(result))
     @req !isnothing(k) "element must be a generator of the module"
     ind = ordered_multi_index(k, p, n)
     e = gens(F)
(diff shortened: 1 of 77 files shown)


src/Modules/ExteriorPowers/FreeModules.jl
```diff
@@ -173,8 +173,8 @@ function koszul_duals(v::Vector{T}; cached::Bool=true) where {T<:FreeModElem}
   success, M, p = _is_exterior_power(F)
   n = rank(M)
   success || error("element must be an exterior product")
-  k = [findfirst(x->x==u, gens(F)) for u in v]
-  any(x->x===nothing, k) && error("elements must be generators of the module")
+  k = [findfirst(==(u), gens(F)) for u in v]
+  any(is_nothing, k) && error("elements must be generators of the module")
```
lgoettgens:
```suggestion
  any(isnothing, k) && error("elements must be generators of the module")
```

src/Rings/mpoly-graded.jl
```diff
@@ -1453,7 +1453,7 @@ end
 #########################################
 function add_relshp(R, S, h)
   #this assumes that h is essentially a canonical map from R -> S
-  D = get_attribute!(() -> Dict{Any, Any}(), R, :relshp)::Dict{Any, Any}
+  D = get_attribute!(R, :relshp, Dict{Any, Any}())::Dict{Any, Any}
```
fingolfin, line 1456: The new code always allocates a new dictionary even if it is not needed. But you can do this:
```suggestion
  D = get_attribute!(Dict{Any, Any}, R, :relshp)::Dict{Any, Any}
```

fingolfin (approved):

src/Rings/mpoly-graded.jl
fingolfin: All tests just passed, this can be done later

lgoettgens: Oooohhhh, that's sooo smart 😅

PR #4132, opened 2024-09-23
Use BinaryWrapper to point GAP packages to singular/polymake executables

fingolfin:
Now that we have GAP.jl 0.11, I can finally submit this as a PR. But only as draft as I didn't finish the Polymake support on macOS (@benlorenz gave me a hint how to do it, which I record in this patch, but I just didn't get around to following up on it).

Posting this so it doesn't get forgotten. Also now @ThomasBreuer can access it easily.


lgoettgens:
Just for future reference:
- the singular part of this was resolved in https://github.com/oscar-system/GAP.jl/pull/1099 and is available in Oscar since https://github.com/oscar-system/Oscar.jl/pull/4421
- the polymake part gets tracked in https://github.com/oscar-system/GAP.jl/issues/1122

PR #4133, opened 2024-09-23
Prefer symbols over strings in polynomial_ring

lgoettgens:
See https://github.com/Nemocas/AbstractAlgebra.jl/pull/1804


lkastner:
Please add a test to ensure this is not done wrong in `src` again. We don't want to repeat this procedure over and over. Also add a style guide entry that symbols should be used instead of strings.

fingolfin:
> Please add a test to ensure this is not done wrong in `src` again.

Not sure what you mean here? Using strings is not "wrong". Adding a test that reliably prevents new code from using this would be super hard to do, I don't think it is reasonable to expect this from this PR.

> We don't want to repeat this procedure over and over. Also add a style guide entry that symbols should be used instead of strings.

I agree a style guide entry suggesting (and explaining!) that would be a good idea. But again I don't think it is necessary to do it in this PR. Of course if @lgoettgens has the time for it, great. But otherwise either a follow-up PR, or an issue reminding us to do this should be fine?

lgoettgens:
I can add a style guide entry, but would prefer to do that in a follow-up PR. It would be great to get this in ASAP before we get any conflicts.

fingolfin (approved):
This is huge, so I did not look at every changed line. But I looked at many samples across varies files and they all seemed reasonable. So if the tests pass, we should just merge it.

lkastner:
> Not sure what you mean here? Using strings is not "wrong". Adding a test that reliably prevents new code from using this would be super hard to do, I don't think it is reasonable to expect this from this PR.
> 
It was just a jab at this PR being super large without significantly improving performance. As you said in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1804: 
```
Of course code where this noticeable would be code that calls polynomial_ring
far too many times, and should be changed to not do that.
```
So since we are approaching this bureaucratically I commented bureaucratically. ;)

PR #4134, opened 2024-09-23
Adds free resolutions over quotient rings by using `Singular.sres`

RafaelDavidMohr:
Adds free resolutions over quotient rings by using `Singular.sres` (with `fres` currently not working for quotient rings in Singular).

Some Macaulay2 people told me that their users were confused by the fact that they could compute free resolutions over quotient rings without specifying a length, so I force the user to set a length for the free resolution when computing over quotient rings.

CC: @ederc


wdecker:
> Some Macaulay2 people told me that their users were confused by the fact that they could compute free resolutions over quotient rings without specifying a length, so I force the user to set a length for the free resolution when computing over quotient rings.

That is o.k. Just for the record: entering zero would also be o.k. since then Singular stops after number of variables +1 steps.

wdecker (approved):

src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -263,19 +263,19 @@ Convert a Singular vector to a free module element.
...
-  R = base_ring(Rx)
-  for (i, e, c) = s
-    f = Base.findfirst(==(i), pos)
+  R = coefficient_ring(Rx)
+  values = elem_type(Rx)[]
+  for (i, e, c) in s
+    f = Base.findfirst(x->x==i, pos)
     if f === nothing
-      push!(values, MPolyBuildCtx(base_ring(F)))
+      push!(values, zero(Rx))
```
HechtiDerLachs: Were these deliberately reverted because we do not have an `MPolyBuildCtx` over quotient rings? In that case, I suggest to refactor this part of the code via type dispatch. We talked about it during last week's workshop how to speed this up, eventually. Also, I think @Lax202 's work in progress on non-commutative groebner bases might touch this part again. So we should eventually decide how to split this up here. 

Edit: This does not need to hold this PR back for now. But before anyone merges, we should open an issue to remind us to clean this bit up.

HechtiDerLachs:
Thanks for doing this! Actually, I have a big computation going on right now which could potentially benefit from this.

src/Modules/UngradedModules/FreeResolutions.jl
```diff
@@ -392,16 +396,20 @@ julia> matrix(map(FM3, 1))
...
+                         length::Int=0,
+                         algorithm::Symbol = T <:MPolyRingElem ? :fres : :sres) where {T <: Union{MPolyRingElem, MPolyQuoRingElem}}
 
   coefficient_ring(base_ring(M)) isa AbstractAlgebra.Field ||
       error("Must be defined over a field.")
 
+  if T <: MPolyQuoRingElem
+    !iszero(length) || error("Specify a length up to which a free resolution should be computed")
+  end
+
```
afkafkafk13, line 412: Singular's sres sets a default of nvars(basering) modules to be computed, if no length is specified.  It might be better for compatibility to document this fallback instead of throwing an error.

src/Modules/UngradedModules/ModuleGens.jl
HechtiDerLachs: Follow up: I tried do speed up some of our computations yesterday and made a patch for this part here in #4147 . Would it be desirable to push it directly to this PR? If so, please let me know.

RafaelDavidMohr: Thanks for taking a look and fixing this! Merging this in unison would be good for me.

src/Modules/UngradedModules/FreeResolutions.jl
RafaelDavidMohr: Thank you for taking a look. Having this default nvars(basering) length is the same they had in Macaulay2, they told me that they removed this at some point for clarity for the users.

But if you prefer to have this default, I will of course make the appropriate changes.

wdecker:
I think you can leave it as it is now.

src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -261,22 +261,49 @@ end
...
-      push!(pos, i)
+  R = coefficient_ring(Rx)
+  ctx = MPolyBuildCtx(Rx)
+  
+  # shortcut in order not to allocate the dictionary
+# if isone(length(s)) # TODO: length doesn't work!
+#   (i, e, c) = first(s)
+#   push_term!(ctx, R(c), e)
+#   return FreeModElem(sparse_row(Qx, [(i, finish(ctx))]))
+# end
```
HechtiDerLachs: Does anyone know what is the correct way to check whether a `Singular.svector` has only one term? `length` throws an error, even though it should be part of the iterator interface.

hannes14: length(p::svector) = Int(libSingular.pLength(p.ptr))

should do it

hannes14: see https://github.com/oscar-system/Singular.jl/pull/833

src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -283,18 +283,21 @@ function (F::FreeMod{<:MPolyRingElem})(s::Singular.svector)
   return FreeModElem(sparse_row(Rx, [(i, finish(ctx)) for (i, ctx) in cache]), F)
 end
 
+# TODO: move this eventually
+length(s::Singular.svector) = Int(Singular.libSingular.pLength(s.ptr))
```
lgoettgens: Please move this to Singular.jl before merging this PR. Sorting this out afterwards is more of a hassle than doing it right the first time

src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -261,22 +261,52 @@ end
...
-      f = length(values)
-      push!(pos, i)
+  R = coefficient_ring(Rx)
+  ctx = MPolyBuildCtx(Rx)
+  
+  # shortcut in order not to allocate the dictionary
+  if isone(length(s))
+    (i, e, c) = first(s)
+    push_term!(ctx, R(c), e)
+    return FreeModElem(sparse_row(Qx, [(i, finish(ctx))]))
```
HechtiDerLachs:
```suggestion
    return FreeModElem(sparse_row(Rx, [(i, finish(ctx))]))
```

HechtiDerLachs:
My changes seem to work for now. It remains to clean up some of the tests.

test/Modules/MPolyQuo.jl
```diff
@@ -96,7 +96,7 @@ end
   A2 = FreeMod(A, 2)
   v = [x*A2[1] + y*A2[2], z*A2[1] + (x-1)*A2[2]]
   M, _ = quo(A2, v)
-  p = free_resolution(M)
+  p = free_resolution(M, length = 11)
   @test !iszero(p[10])
```
HechtiDerLachs:
```suggestion
  @test p[10] isa FreeMod
```

test/Modules/MPolyQuo.jl
```diff
@@ -106,7 +106,7 @@ end
   A, _ = quo(R, I)
   A1 = FreeMod(A, 1)
   M, _ = quo(A1, [y*A1[1], z*A1[1]])
-  p = free_resolution(M)
+  p = free_resolution(M, length = 11)
   @test iszero(p[10])
```
HechtiDerLachs:
```suggestion
  @test p[10] isa FreeMod
```
being zero is not a mathematical necessity, but depends on the chosen algorithm. The key point is to see that the resolution actually works. 

But: I think, I have made the resolution over quotient rings lazy and infinite by default. Since for the previous implementation we had nothing better than iterative kernel computation, there was no trade-off from that. This seems to have changed now, so there might be further things to correct down the road. We'll see.

RafaelDavidMohr:
Thanks @HechtiDerLachs for fixing the last tests. Is there any interest in keeping the option to compute free resolutions over quotients by iterative kernel computations in addition to using `sres`?

HechtiDerLachs:
> Is there any interest in keeping the option to compute free resolutions over quotients by iterative kernel computations in addition to using `sres`?

I think it's always good to have different options for the user and a reasonable default. Eventually it would be nice to have the computation lazy, but using `sres`. But this is probably to be decided elsewhere and not on this PR. If the tests pass, then from my side I'm ok with the iterative kernel computation being disabled for the moment. We still have the code to put it back in and I am probably using the hypercomplexes, anyway. 

@jankoboehm : Any opinion on this matter?

src/Modules/UngradedModules/ModuleGens.jl
HechtiDerLachs: Update: @hannes14 's PR on Singular.jl just got merged. Do we have to wait for a release there? Or can we remove the method here already? 

ping @lgoettgens

lgoettgens: You will need to wait for a new release of Singular.jl, and then adapt the `compat` entry in Oscar's `Project.toml` to this new version number

HechtiDerLachs:
Some timings for `@time Oscar.test_module("test/Modules/", new=false)`

 On current master:
 37.206091 seconds (185.12 M allocations: 9.750 GiB, 24.01% gc time, 20.60% compilation time)
 37.702440 seconds (184.58 M allocations: 9.738 GiB, 22.81% gc time, 21.73% compilation time)

On the head of this PR:
 37.933323 seconds (189.73 M allocations: 9.926 GiB, 23.15% gc time, 23.39% compilation time)
 38.951809 seconds (190.67 M allocations: 9.939 GiB, 22.77% gc time, 19.89% compilation time)


On 3d35d0a363d8b19a592d172ad773ae1ec3feab90, i.e. not using `MPolyBuildCtx` for the conversion
 42.981092 seconds (215.30 M allocations: 10.906 GiB, 21.85% gc time, 17.19% compilation time)
 42.981092 seconds (215.30 M allocations: 10.906 GiB, 21.85% gc time, 17.19% compilation time)
 41.036220 seconds (214.36 M allocations: 10.897 GiB, 22.19% gc time, 18.05% compilation time)
 43.827199 seconds (215.27 M allocations: 10.907 GiB, 21.35% gc time, 20.50% compilation time)

I had to disable two lines in the tests for the modules over `MPolyQuoRing` for the latter.

HechtiDerLachs:
> You will need to wait for a new release of Singular.jl, and then adapt the `compat` entry in Oscar's `Project.toml` to this new version number

This is still pending. Other than that, this seems good to go, I think.

lgoettgens:
> > You will need to wait for a new release of Singular.jl, and then adapt the `compat` entry in Oscar's `Project.toml` to this new version number
> 
> This is still pending. Other than that, this seems good to go, I think.

Singular.jl v0.23.8 is now available

src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -261,22 +261,52 @@ end
...
+    ctx = get!(cache, i) do
+      MPolyBuildCtx(Rx)
+    end
+    push_term!(ctx, R(c), e)
+  end
+  return FreeModElem(sparse_row(Rx, [(i, finish(ctx)) for (i, ctx) in cache]), F)
+end
+
+# TODO: move this eventually
+length(s::Singular.svector) = Int(Singular.libSingular.pLength(s.ptr))
```
lgoettgens:
```suggestion
```

lgoettgens:
I added the compat bump myself. No objections from me anymore

Issue #4136, opened 2024-09-23
GAP package `sglppow` causes stuck subprocess and `fgets() failed with errno 5`


benlorenz:
**Describe the bug**
As reported on slack by @paemurru, since today when exiting julia we get a weird error message:
```julia
julia> exit()
fgets() failed with errno 5
Input/output error
```
It does not happen every time but quite often.

This seems to be caused by https://github.com/oscar-system/Oscar.jl/pull/3517, once I comment this line
```julia
     "sglppow",  # extend the small groups library
```
the issue disappears.

This error print is not from the julia process but from a (stuck?) `singular` subprocess that is started when Oscar is loaded:
```julia
julia> using Oscar
  ___   ____   ____    _    ____
 / _ \ / ___| / ___|  / \  |  _ \   |  Combining ANTIC, GAP, Polymake, Singular
| | | |\___ \| |     / _ \ | |_) |  |  Type "?Oscar" for more information
| |_| | ___) | |___ / ___ \|  _ <   |  Manual: https://docs.oscar-system.org
 \___/ |____/ \____/_/   \_\_| \_\  |  1.2.0-DEV #HEAD 55ea957 2024-09-22

shell> ps -f
UID        PID  PPID  C STIME TTY          TIME CMD
lorenz    4147  4146  0 Sep20 pts/18   00:00:00 bash
lorenz   22361  4147 26 17:55 pts/18   00:00:12 julia-latest --project=.
lorenz   22475 22361  0 17:55 pts/18   00:00:00 /usr/bin/singular -t /tmp/gaptempdirQdsGt8/sing.in
lorenz   22776 22361  0 17:55 pts/18   00:00:00 /bin/bash -c (ps -f) && true
lorenz   22777 22776  0 17:55 pts/18   00:00:00 ps -f

shell> cat /tmp/gaptempdirQdsGt8/sing.in
proc GAP_Done () { return ( "@" ) };
proc GAP_Apostrophe () { return ( "'" ) };
GAP_Done();

julia> exit()
fgets() failed with errno 5
Input/output error
```

I do not see this subprocess once I remove the `sglppow` package from `Oscar.jl`.

Attaching `strace` to the `singular` subprocess shows that this is indeed the source of this message:
```
...
17:46:35.462538 read(0, 0x563fb8d25810, 1024) = -1 EIO (Input/output error)
17:46:38.640469 write(2, "fgets() failed with errno 5\nInput/output error\n", 47) = 47
[...]

PR #4141, opened 2024-09-24
Revert "load the GAP packages sotgrps and sglppow"

fingolfin:
This reverts PR #3517 / commit 55ea9572354c55fec4d74b4b1a09067620f7f1d3.

We can reland it as soon as the underlying issue is fixed, but for now this brings relieve to people who are affected by issue #4136.


benlorenz (approved):
Disabling just `sglppow` would suffice to avoid the error.

PR #4071, opened 2024-09-04
Fix #4039

joschmitt:
The fix for the Singular bug in #4039 arrived in OSCAR with the latest Singular.jll

This closes #4039 by handling the empty dictionary that is now returned by Singular in the example.

PR #4142, opened 2024-09-24
Some more fixes

HechtiDerLachs:
Today's harvest from working on the elliptic surfaces.


experimental/DoubleAndHyperComplexes/src/Morphisms/simplified_complexes.jl
```diff
@@ -228,7 +228,7 @@ function (fac::SimplifiedChainFactory)(d::AbsHyperComplex, Ind::Tuple)
       w = Tinv[i]
       new_entries = Vector{Tuple{Int, elem_type(base_ring(w))}}()
       for (real_j, b) in w
-        j = findfirst(==(real_j), J)
+        j = findfirst(k->k==real_j, J)
```
lgoettgens:
```suggestion
        j = findfirst(==(real_j), J)
```
why was this changed back (compared to https://github.com/oscar-system/Oscar.jl/pull/4126)?

src/Rings/mpoly-ideals.jl
```diff
@@ -11,7 +11,7 @@ Given a vector `V` of polynomials in `R`, return the ideal of `R` generated by t
 
 # Examples
 ```jldoctest
-julia> R, (x, y) = polynomial_ring(QQ, [:x, :y])
+julia> R, (x, y) = polynomial_ring(QQ, ["x", "y"])
```
lgoettgens: this one and the similar changes should not get changed back (compared to https://github.com/oscar-system/Oscar.jl/pull/4133).

src/Rings/mpoly-ideals.jl
```diff
@@ -642,11 +654,8 @@ Return whether `I` is a radical ideal.
 Computes the radical.
 """
 @attr Bool function is_radical(I::MPolyIdeal)
-  if has_attribute(I, :is_prime) && is_prime(I)
-    return true
-  end
-
-  return I == radical(I)
+  has_attribute(I, :is_prime) && return is_prime(I)
```
lgoettgens: this change looks wrong to my eyes.
Consider a non-prime, radical ideal. This new version would return `false`, although the ideal is indeed radical

src/Rings/mpoly-ideals.jl
```diff
@@ -898,31 +907,26 @@ Multivariate polynomial ring in 2 variables over number field graded by
   y -> [1]
 ```
 """
-@attr Vector{Tuple{typeof(I), typeof(I), ideal_type(change_base_ring(rationals_as_number_field()[1], base_ring(I))[1]), Int64}} function absolute_primary_decomposition(I::MPolyIdeal{<:MPolyRingElem{QQFieldElem}})
```
lgoettgens: all of the below has been changed by https://github.com/oscar-system/Oscar.jl/pull/4071. Why is it changed back?

thofma:
I think a lot of changes from #4133 and #4126 are undone here. I guess this was not on purpose.

lgoettgens (dismissed):

HechtiDerLachs:
Yes, I'm already working on it. It's a bit annoying, because I had to work on the compute-server today, but there I can't put my login-credentials for GH. Thus, I manually copied the files via scp and this is what you get...

src/Rings/mpoly-ideals.jl
HechtiDerLachs: True. Thanks for the catch.

src/Rings/mpoly-ideals.jl
```diff
@@ -642,12 +654,10 @@ Return whether `I` is a radical ideal.
 Computes the radical.
 """
 @attr Bool function is_radical(I::MPolyIdeal)
-  if has_attribute(I, :is_prime) && is_prime(I)
-    return true
-  end
-
-  return I == radical(I)
+  has_attribute(I, :is_prime) && return is_prime(I)
```
HechtiDerLachs:
```suggestion
  has_attribute(I, :is_prime) && is_prime(I) && return true
```

src/Rings/mpoly-ideals.jl
```diff
@@ -2249,3 +2260,10 @@ function flag_pluecker_ideal(ring::MPolyRing{<: FieldElem}, dimensions::Vector{I
                    isReduced=true,
                    isGB=true))
 end
+
+# Since most ideals implement `==`, they have to implement the hash function.
+# See issue #4143 for problems entailed. This should fix it. 
+function hash(I::Ideal, c::UInt)
+  return UInt(0)
+end
```
HechtiDerLachs: Any objections? @lgoettgens , @fingolfin ?

src/Rings/mpoly-ideals.jl
```diff
@@ -1149,7 +1170,14 @@ function minimal_primes(
       end
       J = K
     end
-    result = unique!(filter!(!is_one, vcat([minimal_primes(j; algorithm, factor_generators=false) for j in J]...)))
+    # unique! seems to fail here. We have to do it manually.
+    pre_result = filter!(!is_one, vcat([minimal_primes(j; algorithm, factor_generators=false) for j in J]...))
```
lgoettgens, line 1206:
```suggestion
    pre_result = filter!(!is_one, reduce(vcat, [minimal_primes(j; algorithm, factor_generators=false) for j in J]))
```
save some allocations

src/Rings/mpoly-ideals.jl
```diff
@@ -2249,3 +2277,11 @@ function flag_pluecker_ideal(ring::MPolyRing{<: FieldElem}, dimensions::Vector{I
                    isReduced=true,
                    isGB=true))
 end
+
+# Since most ideals implement `==`, they have to implement the hash function.
+# See issue #4143 for problems entailed. Interestingly, this does not yet fix 
+# the failure of unique! on lists of ideals.
+function hash(I::Ideal, c::UInt)
+  return UInt(0)
+end
```
```suggestion
function hash(I::Ideal, h::UInt)
  return h
end
```
returning the second value is save and a small optimisation.

benlorenz: Hashing the concrete type and maybe even the parent ring should also be fast and help avoid a few comparisons.

src/Rings/mpoly-ideals.jl
```diff
@@ -2249,3 +2278,11 @@ function flag_pluecker_ideal(ring::MPolyRing{<: FieldElem}, dimensions::Vector{I
                    isReduced=true,
                    isGB=true))
 end
+
+# Since most ideals implement `==`, they have to implement the hash function.
+# See issue #4143 for problems entailed. Interestingly, this does not yet fix 
+# the failure of unique! on lists of ideals.
+function hash(I::Ideal, c::UInt)
+  return hash(typeof(I), hash(base_ring(I), c))
+end
```
HechtiDerLachs: I advocate for this solution now. It is a matter of fact that ideals can not be hashed in general and even for polynomial rings this is probably expensive if not impossible (we need to choose an ordering which might be a dead end for a specific ideal). Yet, the `==` function is vital to how we use ideals and we need to be able to use them in vectors/lists/dictionaries.

simonbrandhorst: Is it safe to hash the type? After all two mathematically equal objects may have different types. 
(Thinking e.g. of IdealSheaf, RadicalIdealSheaf, PrimeIdealSheaf etc.)

lgoettgens: > Is it safe to hash the type? After all two mathematically equal objects may have different types. (Thinking e.g. of IdealSheaf, RadicalIdealSheaf, PrimeIdealSheaf etc.)

It is only safe if you don't implement `==` between this type and other types. In your example, you shouldn't include the type in hashes.

HechtiDerLachs: Ok, good to know. I was only following the suggestion by @benlorenz . But you're probably right.

fieker:
On Fri, Sep 27, 2024 at 01:30:28AM -0700, Simon Brandhorst wrote:
> 
> 
> 
> > +# Since most ideals implement `==`, they have to implement the hash function.
> +# See issue #4143 for problems entailed. Interestingly, this does not yet fix 
> +# the failure of unique! on lists of ideals.
> +function hash(I::Ideal, c::UInt)
> +  return hash(typeof(I), hash(base_ring(I), c))
> +end
> 
> Is it safe to hash the type? After all two mathematically equal objects may have different types. 
> (Thinking e.g. of IdealSheaf, RadicalIdealSheaf, PrimeIdealSheaf etc.)
Tough one: Julia tries/ tried to make sure any number comparing equal
have the same hash. We don't - way too expensive.

Is there any situation where you want a dict or a set where the keys/
elements are different types and then catch that 2 objects of different
types are "the same"?

HechtiDerLachs:
> Is there any situation where you want a dict or a set where the keys/ elements are different types and then catch that 2 objects of different types are "the same"?

Probably yes. For instance, it would make sense to introduce a lazy type for radical ideals, say `RadicalIdeal`. This would have a special dispatch for ideal membership so that the computation of the radical can be avoided. Yet, comparison with the standard types of ideals for the respective ring could give that two ideals are equal, even though their type differs.

simonbrandhorst (approved):

src/Rings/mpoly-ideals.jl
```diff
@@ -572,7 +571,31 @@ Ideal generated by
...
+      S, iso, iso_inv = simplify(Q)
+      is_zero(ngens(S)) && return I # This only happens when all variables can be eliminated. 
+      # Then the ideal defines a reduced point.
+      J = modulus(S)
+      pre_res = _compute_radical(J)
+      pre_res = ideal(R, elem_type(R)[lift(iso_inv(S(g))) for g in gens(pre_res)])
+      res = ideal(R, small_generating_set(pre_res + I))
+      set_attribute!(res, :is_radical=>true)
+      return res
+    end
```
afkafkafk13, line 616: As you specifically asked my opinion on this code block: Yes, this definitely makes sense, as it prepares the objects better for the use of Singular's primdec.lib.

test/Rings/MPolyAnyMap/MPolyRing.jl
```diff
@@ -124,7 +124,7 @@
     Kxz, (z1, z2) = Kx[:z1, :z2];
     f = hom(Kxz, Kxz, hom(Kx, Kxz, [z1, z2]), [z2, z1])
     g = hom(Kxz, Kxz, hom(Kx, Kx, [y, x]), [z1 + 1, z2 + 1])
-    fg = @inferred f * g
+    fg = f * g
```
HechtiDerLachs, line 127: @thofma : I had reasons to not allow `CompositeMap`s at some points within the schemes. To achieve this, I had to overwrite composition for some constellations. But given the genererality of allowable coefficient maps, I did not see how to make this type stable anymore. Do you strongly object to this regression here?

HechtiDerLachs:
I am sorry about the messy state that this is in. But it turned out to be a really hard nut to crack here. 

Some things to be discussed:

1) I introduced a "simultaneous blowup strategy" for the elliptic surfaces. Something which is vital to what we try to achieve these days. This should actually be consensus.

2) Because of 1) I needed to restructure the implementation of `restrict(::AbsCoveredSchemeMorhpism, ::CoveredClosedEmbedding, ::CoveredClosedEmbedding)`. This lead to a lot of issues down the road. All code which uses refinements of coverings for morphisms of covered schemes is still rather brittle. Unfortunately, this is a bigger task to clean this up. A lot has already been done, but many things still only work in very particular constellations. One thing which did not work anymore was the `isomorphism_on_open_subset` which I had once added in order to be able to identify function fields along birational maps like blowups. This feature turned out to not be that useful and we actually don't use it these days. So I took the liberty to disable it. Let's see whether the tests pass. Ping @simonbrandhorst 

3) Something went wrong with some composition of maps. I suddenly got `CompositeMap`s where earlier the composition had been carried out. I still could not pin down the original cause of this, but I made some fixes. In the course of these events, I found it necessary to move the type declaration for `MPolyAnyMap` to a file of its own. Otherwise, debugging would not have been possible. I would appreciate it, if it could stay this way. Ping @thofma .  

Once this reaches a state where tests actually pass, I will remove the code which for the time being has just been commented out. But if you have any objections about where parts of this are going, please let me know already.

afkafkafk13:
Can you clarify what you mean by 'simultaneous blowup strategy'?

HechtiDerLachs:
> Can you clarify what you mean by 'simultaneous blowup strategy'?

In the elliptic surfaces we exclusively encounter A-D-E singularities. Up to now we blew them up one by one. But this produces a lot of charts and gluings. In the particular example we need for our paper we only have 8 ordinary A1 singularities. They are resolved simultaneously by one single blowup. This is achieved with with a keyword argument to the constructor for elliptic surfaces which then selects this strategy: Just blow up the ideal sheaf of the reduced singular locus, until you're done.

afkafkafk13:
> > Can you clarify what you mean by 'simultaneous blowup strategy'?
> 
> In the elliptic surfaces we exclusively encounter A-D-E singularities. Up to now we blew them up one by one. But this produces a lot of charts and gluings. In the particular example we need for our paper we only have 8 ordinary A1 singularities. They are resolved simultaneously by one single blowup. This is achieved with with a keyword argument to the constructor for elliptic surfaces which then selects this strategy: Just blow up the ideal sheaf of the reduced singular locus, until you're done.

oh. I see. This makes a lot of sense.

experimental/Schemes/src/BlowupMorphism.jl
```diff
@@ -494,29 +532,35 @@ function restrict(f::AbsCoveredSchemeMorphism,
...
   inc_cod_cov = covering_morphism(inc_cod)
 
-  # We need to do the following.
-  # - Pass to a common refinement ref_cod in X that both
-  #   f and inc_cod can restrict to.
-  # - Pass to a common refinement in Y
-  ref_cod, a, b = _register!(common_refinement(codomain(f_cov), codomain(inc_cod_cov)), codomain(f))
-  inc_cod_ref = restrict(inc_cod, ref_cod)
-  f_res = restrict(f, ref_cod)
-  ref_dom, aa, bb = _register!(common_refinement(domain(f_res), codomain(inc_dom_cov)), domain(f))
```
simonbrandhorst, line 504: Before the refinements were registered. Now this does not seem to be the case anymore. 
Is this intentional? And if yes what is the rationale here?

simonbrandhorst:
I am fine with discontinuing `isomorphism_on_open_subset` it is in experimental and not exported.

simonbrandhorst (changes requested):
The changes in the doctests in unrelated code such as algebraic statistics seem odd.
Can you explain the change in behavior? Are the results still correct?

experimental/Schemes/src/BlowupMorphism.jl
HechtiDerLachs: Originally we were planning to keep track of all the refinements and the morphisms between them inside the covered scheme. But this is not very garbage-collector-friendly, so we eventually refrained from that and switched to the implicit tree structure for refinements. The call to `_register!` was a deprecated artefact.

HechtiDerLachs:
> Can you explain the change in behavior?

See #4171 and my comment on slack.

simonbrandhorst:
O.K. when you checkout the files I reapprove.

simonbrandhorst (approved):

PR #4145, opened 2024-09-24
LieAlgebras: more optimizations regarding type stability

lgoettgens:


test/Serialization/setup_tests.jl
```diff
@@ -4,9 +4,11 @@ using JSONSchema, Oscar.JSON
 
 # we only need to define this once
 
-if !isdefined(Main, :test_save_load_roundtrip) || isinteractive()
+if !isdefined(Main, :mrdi_schema)
   mrdi_schema = Schema(JSON.parsefile(joinpath(Oscar.oscardir, "data", "schema.json")))
+end
 
+if !isdefined(Main, :test_save_load_roundtrip) || isinteractive()
```
fingolfin, line 11: This seems to be an unrelated change?

fingolfin (approved):

test/Serialization/setup_tests.jl
lgoettgens: Yeah, this should have been included in https://github.com/oscar-system/Oscar.jl/pull/4065, but I forgot it somehow.

Without this change, running `Oscar.test_experimental_module("LieAlgebras")` from the REPL parses the schema multiple times, including some network request and some logging.

PR #4147, opened 2024-09-24
Attempt to clear up mpolyquo modules

HechtiDerLachs:
Building on #4134 we should consider replacing the magic three functions for modules over `MPolyQuoRing`s. I hope to speed up some computations with this which we need together with @simon-felten.

This is just a first try to get an idea about whether the tests fail. For `coordinates` we should look into the caching, etc. All this I would prefer to discuss in person on the next occasion.


afkafkafk13:
@HechtiDerLachs : What is the status here (relative to the fact that #4134 has been merged in the meantime)?

HechtiDerLachs:
This would need to be cleaned up. But I will do it another time. I was hoping that this helps with a specific computation, but it didnt. So priority has become lower again.

PR #4152, opened 2024-09-26
Vinbergs Algorithm

simonbrandhorst:
- *Add Vinberg's algorithm*


src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,336 @@
...
+function _check_v0(Q::ZZMatrix, v0::ZZMatrix)
+  if iszero(v0) # generate `v0` by finding the negative eigenvalue `neg_ev` of `Q`. We set `v0` as a primitive generator of the eigenspace of `neg_ev`
+    V = quadratic_space(QQ,Q)
+    diag, trafo = diagonal_with_transform(V)
+    @req count(is_positive, diag) == 1 "Q must be of signature (1,n)"
+    @req count(is_zero, diag) == 0 "Q must be of signature (1,n)"
+    @req count(is_negative, diag) > 0 "Q must be of signature (1,n)"
+    i = findfirst(is_positive, diag)
+    v0 = trafo[i:i,:]
+    v0 = ZZ.(denominator(v0)*v0)
```
simonbrandhorst, line 19: @ElMerkel 
here I moved from eigenvalues to to an implementation of Gram-Schmidt orthogonalisation since I am not convinced that `Q` is diagonalisable over the rationals.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,336 @@
...
+function _has_non_obtuse_angles!(tmp::ZZMatrix, Qv::ZZMatrix, roots::Vector{ZZMatrix})
+  for r in roots
+    #tmp = r*Qv
+    mul!(tmp, r, Qv)
+    if !is_zero_entry(tmp,1, 1) && !is_positive_entry(tmp, 1, 1)
+      return false
+    end
+  end
+  return true
+end
```
simonbrandhorst, line 153: @ElMerkel  
This is the non-allocating version I was referring to. Note however that using `mul!` is unsafe and can lead to segmentation faults if it gets bad input, e.g. matrices of a wrong size.  
Hence it is only ever worth using in time critical places.

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,336 @@
...
+  Qv = zero_matrix(ZZ, ncols(Q), 1)
+  tmp2 = zero_matrix(ZZ, 1, 1)
+  for (n, k) in iteration # search for vectors which solve $n = v.v_0$ and $k = v^2$
+    @vprintln :Vinberg 1 "computing roots of squared length v^2=$(k) and v.v0 = $(n)"
+    possible_Vec = short_vectors_affine(Q, v0, QQ(n), k)
+    for v in possible_Vec
+      if v in roots
+        continue
+      end
+      v = _rescale_primitive(v)
```
simonbrandhorst: @ElMerkel  
it seems impossible to me that `v` is in roots already. After all `(n,k)` should be different. 
... unless `v` is not primitive... but then this check is too late anyways.

src/NumberTheory/vinberg.jl
fingolfin: Note that you can still make it perfectly safe by adding an input validation check yourself, i.e. in this case: that the number of rows resp. columns of the two matrices fit together.

simonbrandhorst:
For my part this is good to go. Thank you @ElMerkel .

src/NumberTheory/vinberg.jl
```diff
@@ -0,0 +1,341 @@
...
+  end
+  iteration = _distance_indices(upper_bound, real_root_lengths) # find the right order to iterate through $v.v_0$ and $v^2$
+  roots = _distance_0(Q, v0, real_root_lengths, direction_vector) # special case $v.v_0 = 0$
+  Qv = zero_matrix(ZZ, ncols(Q), 1)
+  tmp2 = zero_matrix(ZZ, 1, 1)
+  for (n, k) in iteration # search for vectors which solve $n = v.v_0$ and $k = v^2$
+    @vprintln :Vinberg 1 "computing roots of squared length v^2=$(k) and v.v0 = $(n)"
+    possible_Vec = short_vectors_affine(Q, v0, QQ(n), k)
+    for v in possible_Vec
+      if !isone(reduce(gcd, v))
```
fingolfin, line 292: I think this `reduce(gcd, v)` pops up multiple times here, and it could be made more efficient. Maybe move it into its own helper function?

Then you can make it potentially faster / allocate less by doing `reduce(gcd!, v; init=one(v[1]))`.

Alas, that's not a requirement for this PR, just a suggestion for a future tweak.

fingolfin (approved):
Since I was asked for a review: I don't know the maths here so i can't evaluate that. But code wise this seems reasonable enough.

PR #4154, opened 2024-09-26
Support computing strict transform of an ideal in Cox ring

paemurru:
Done.


experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -44,3 +44,60 @@ end
 function total_transform(f::AbsBlowupMorphism, II::AbsIdealSheaf)
   return pullback(f, II)
 end
+
+function _cox_ring_homomorphism(f::ToricBlowupMorphism)
+  Y = domain(f)
+  @assert grading_group(cox_ring(Y)) === class_group(Y)
+  G = class_group(Y)
```
afkafkafk13: First define G as the class_group(Y) and then use G in the assert, since you need G afterwards anyway and in the case of failure of the assert, you did not loose anything.

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -44,3 +44,60 @@ end
...
+  n = number_of_generators(G)
+  new_var = cox_ring(Y)[index_of_new_ray(f)]
+  M = generator_degrees(cox_ring(Y))
+  @assert M[index_of_new_ray(f)][n] < 0 "Assuming the blowup adds a new column vector to `Oscar.generator_degrees(cox_ring(codomain(f)))`, the entry of that vector corresponding to the new ray should be negative. If not, multiply by -1."
```
afkafkafk13: Your error messages in case of failing asserts seem pretty long to me.  Please try to make them shorter and maybe add appropriate notes to the documentation instead.

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -44,3 +44,60 @@ end
...
+  ind = Int64(abs(M[index_of_new_ray(f)][n]))
+  for i in 1:n_rays(Y)
+    i == index_of_new_ray(f) || @assert M[i][n] >= 0 "Assuming the blowup adds a new column vector to `Oscar.generator_degrees(cox_ring(codomain(f)))` for which the entry corresponding to the new ray is negative, the entries corresponding to all the other rays should be nonnegative. If not, add integer multiples of the other columns until it is."
```
afkafkafk13: Do you want to say 
`@assert findfirst(M[i][n] < 0 && i != index_of_new_ray(f) , 1:n_rays(Y)) === nothing`

afkafkafk13 (commented):
I leave you a few comments on the code. 

sideremark: I am not convinced (theorywise) that you can just set one entry to -1 in the grading change, but I am not a Cox-ring specialist. So I leave the reviewing of the content to someone else.

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -44,3 +44,60 @@ end
...
+    S = grade(S_ungraded, M_new)[1]
+    inj = hom(cox_ring(Y), S, gens(S))
+  end
+  F = hom(cox_ring(codomain(f)), S, [inj(new_var^(M[i][n])*S[i]) for i in 1:n_rays(Y) if i != index_of_new_ray(f)])
+  return (F, ind)
+end
+
+function strict_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+  F, ind = _cox_ring_homomorphism(f)
+  Y = domain(f)
```
HechtiDerLachs: Thanks for looking into this! 

This is to compute the strict transform of an ideal in the cox ring of the codomain, correct? I would probably insert an `@assert` here, or something. 

Once this is done, I would be curious to compare performance with the chart-based approach!

afkafkafk13:
@paemurru : Just a question, because this has been sitting here for 2 weeks now. What is the status here? Are you making progress on this? Are you waiting for comments from someone else? If so, please ping this person.

paemurru:
@afkafkafk13 I am working on this, this will probably stay as a draft pull request for quite some time. The current code only handles simple examples, the general case is a bit more complicated. The proof that it works uses a lot of notation, so is not suitable for a docstring or comment and must be typed in LaTeX. So it probably should be included in a research paper, unless an existing reference is found.

Essentially, we have an abelian group homomorphism, in general not a C-algebra homomorphism and in general not unique, between the Cox rings that takes homogeneous ideals to homogeneous ideals. And the image of a homogeneous ideal is its total transform.

docs/src/AlgebraicGeometry/ToricVarieties/BlowupMorphisms.md
```diff
@@ -79,7 +79,21 @@ morphisms are supported for toric blowups:
...
+
+## Attributes
+
+We can compute the total and strict transforms of homogeneous ideals in Cox rings under star subdivisions along a ray.
+```@docs
+total_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+strict_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+strict_transform_with_index(f::ToricBlowupMorphism, I::MPolyIdeal)
+```
+The above functions are implemented using a group homomorphism between the Cox rings that takes homogeneous ideals to homogeneous ideals:
```
HechtiDerLachs: I'm slightly confused: Is it intentional that we speak of a "group homomorphism" here? I would expect a ring homomorphism. Maybe this is because the ring homomorphism is induced from some morphism of lattices/groups?

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+julia> X = affine_space(NormalToricVariety, 2)
+Normal toric variety
+
+julia> f = blow_up(X, [2, 3])
+Toric blowup morphism
+
+julia> R = cox_ring(X)
+Multivariate polynomial ring in 2 variables over QQ graded by
+  x1 -> 1 by 0 empty matrix
+  x2 -> 1 by 0 empty matrix
```
HechtiDerLachs: This seems odd. Sorry to bring this up, since this is probably not the main point of this PR, but does everyone agree that this is what we want here?

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+Ideal generated by
+  x1 + x2
+
+julia> total_transform(f, I)
+Ideal generated by
+  x1*e^2 + x2*e^3
+```
+"""
+function total_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+  return cox_ring_group_homomorphism(f, I)
```
HechtiDerLachs: The example above suggests that the return value of this is indeed an ideal. Looking at the code, my naive guess would have been that we get a homomorphism whose construction depends on `f` and `I`. But it seems that in fact we apply a homomorphism associated to `f` to `I`. To make this a bit more streamlined with existing functionality, may I suggest to make the current implementation for `cox_ring_homomorphism` for this signature a method for `pullback(f, I)`? Unless this signature is ambiguous, it should be more or less what we want here, right?

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+julia> strict_transform(f, I)
+Ideal generated by
+  x1 + x2*e
+```
+"""
+function strict_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+  S = cox_ring(domain(f))
+  new_var = S[index_of_new_ray(f)]
+  J = total_transform(f, I)
+  return saturation(J, ideal(new_var))
```
HechtiDerLachs:
```suggestion
  new_var = S[index_of_new_ray(f)]
  J = total_transform(f, I)
  return saturation(J, ideal(S, new_var))
```
Just a nitpick, but to avoid potential future problems.

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+    strict_transform(f::ToricBlowupMorphism, I::MPolyIdeal) -> MPolyIdeal
+
+Let $f\colon Y \to X$ be the toric blowup corresponding to a star subdivision along a ray. Let $R$ and $S$ be the Cox rings of $X$ and $Y$, respectively.
+Here "strict transform" means the "scheme-theoretic closure of the complement of the exceptional divisor in the scheme-theoretic inverse image".
+This function returns a homogeneous ideal in $S$ corresponding to the strict transform under $f$ of the closed subscheme of $X$ defined by the homogeneous ideal $I$ in $R$.
+
+This is implemented under the following assumptions:
+  * the star subdivision adds a ray (instead of subdividing along an existing ray),
+  * $X$ is an orbifold (meaning its fan is simplicial), and
+  * $X$ has no torus factors (meaning the rays span $N_{\mathbb{R}}$).
```
HechtiDerLachs: I think it's great that you worked out the mathematics behind this, thank you for that! However, on a first glance the implementation below does not seem to have checks for these requirements. Is this because there is no algorithm for these checks? 

We do have some of these cases; for instance there is no feasible way to assure that a covered scheme is proper over a point. But unless this is the case here, I suggest to introduce the common optional `check::Bool=true` argument in the function's signature and some internal checks which make sure that these three things hold true.

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+"""
+function strict_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+  S = cox_ring(domain(f))
+  new_var = S[index_of_new_ray(f)]
+  J = total_transform(f, I)
+  return saturation(J, ideal(new_var))
+end
+
+@doc raw"""
+    strict_transform_with_index(f::ToricBlowupMorphism, I::MPolyIdeal) -> MPolyIdeal
```
HechtiDerLachs:
```suggestion
    strict_transform_with_index(f::ToricBlowupMorphism, I::MPolyIdeal) -> (MPolyIdeal, Int)
```

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+  J = total_transform(f, I)
+  return saturation_with_index(J, ideal(new_var))
+end
+
+@doc raw"""
+    cox_ring_group_homomorphism(f::ToricBlowupMorphism, g::MPolyDecRingElem) -> MPolyDecRingElem
+    cox_ring_group_homomorphism(f::ToricBlowupMorphism, I::MPolyIdeal) -> MPolyIdeal
+
+Let $f\colon Y \to X$ be the toric blowup corresponding to a star subdivision along a ray. Let $R$ and $S$ be the Cox rings of $X$ and $Y$, respectively.
+Considering $R$ and $S$ with their abelian group structures, there is a group homomorphism $\Phi\colon R \to S$ sending homogeneous ideals to homogeneous ideals with the defining property that the the total transform of a subscheme of $X$, given by a homogeneous ideal $I$ in $R$, is the subscheme of $Y$ given by the homogeneous ideal $\Phi(I)$ in $S$.
```
HechtiDerLachs:
```suggestion
Considering $R$ and $S$ with their abelian group structures, there is a group homomorphism $\Phi\colon R \to S$ sending homogeneous ideals to homogeneous ideals with the defining property that the total transform of a subscheme of $X$, given by a homogeneous ideal $I$ in $R$, is the subscheme of $Y$ given by the homogeneous ideal $\Phi(I)$ in $S$.
```

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+ x1
+ x2
+
+julia> cox_ring_group_homomorphism(f, x1 + x2)
+x1*e^2 + x2*e^3
+```
+"""
+cox_ring_group_homomorphism
+
+function cox_ring_group_homomorphism(f::ToricBlowupMorphism, g::MPolyDecRingElem)
```
HechtiDerLachs: Make these methods of `pullback(f::ToricBlowupMorphism, ...)`? See my above comment.

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+  return ideal(S, [cox_ring_group_homomorphism(f, g) for g in gens(I)])
+end
+
+@attr Tuple{ZZRingElem, Vector{ZZRingElem}} function _cox_ring_group_homomorphism_data(f::ToricBlowupMorphism)
+  # f: Y -> X
+  X = codomain(f)
+  Y = domain(f)
+  @req is_normal(X) "Only implemented when the variety is normal"
+  @req !has_torusfactor(X) "Only implemented when there are no torus factors"
+  @req n_rays(Y) == n_rays(X) + 1 "Only implemented when the blowup adds a ray"
```
HechtiDerLachs: Referring to one of my above comments: It seems that here the sanity of the input is checked. Maybe these checks are then also triggered by the above code. Either way: It looks like some of these tests could be rather expensive, so I would like to point out the option to use `@check` instead of `@req` here so that turning of these tests internally is possible. What do you think?

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+  b = transpose(matrix(ZZ, b_iterator))
+
+  # Multiplying by `d` to assure that a solution exists
+  b *= d
+
+  vs = solve_mixed(A, b, C)
+  n = nvars(cox_ring(X))
+  phi_exponents = zeros(ZZ, n)
+  for i in 1:n
+    ray = Oscar._ray_variable_correspondence(X)[cox_ring(X)[i]]
```
HechtiDerLachs:
```suggestion
    ray = _ray_variable_correspondence(X)[cox_ring(X)[i]]
```

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,49 @@ function cones(PF::_FanLikeType)
   return IncidenceMatrix([Vector{Int}(x) for x in cones])
 end
 
+@doc raw"""
+    minimal_supercone(X::NormalToricVariety, p::AbstractVector{<:IntegerUnion})
+    -> safksjafld
```
HechtiDerLachs:
```suggestion
    -> ???
```

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,49 @@ function cones(PF::_FanLikeType)
...
+  X::NormalToricVariety, r::AbstractVector{<:Union{IntegerUnion,QQFieldElem}}
+)
+  m_cone_indices = Oscar._get_maximal_cones_containing_vector(X, r)
+  @req !is_empty(m_cone_indices) "Ray `r` should be in the support of the fan of `X`."
+  m_cones = [maximal_cones(X)[i] for i in m_cone_indices]
+  m_cone = intersect(m_cones)
+  m_fan = polyhedral_fan(m_cone)
+
+  # By Proposition 1.2.8(c) in [CLS11](@cite), every proper face of a cone
+  # is the intersection of the facets containing it.
```
HechtiDerLachs: I think comments like this are great; thank you for that!

HechtiDerLachs (commented):
Thank you for your work here @paemurru. As I understand, these things have not yet been written up or implemented properly and it looks like you took care of working out the details. That said, I have to admit that I am probably not in the position to check all the mathematical details of this implementation, but I feel that this is also why we have people like you on board: You probably know best what's correct here and I would like to trust you with this. 

From what I can see, the contributions to the code base seem to fit with the existing code. I left some comments to indicate potential for streamlining a bit more. Let me know what you think.

experimental/Schemes/src/ToricBlowups/methods.jl
paemurru: Made the pull request https://github.com/oscar-system/Oscar.jl/pull/4383

src/PolyhedralGeometry/PolyhedralFan/properties.jl
paemurru: Fixed in [ff5df80](https://github.com/oscar-system/Oscar.jl/pull/4154/commits/ff5df80f61c91bdc1790dbd913388526fad0685e)

experimental/Schemes/src/ToricBlowups/methods.jl
paemurru: Only the `has_torusfactor` check is expensive. Related pull request: https://github.com/oscar-system/Oscar.jl/pull/4385

src/PolyhedralGeometry/PolyhedralFan/standard_constructions.jl
```diff
@@ -204,7 +204,7 @@ function _check_containment_via_facet_signs(smaller::Vector{Int}, bigger::Vector
   return true
 end
 function _get_maximal_cones_containing_vector(
-  Sigma::_FanLikeType, v::AbstractVector{<:IntegerUnion}
+  Sigma::_FanLikeType, v::AbstractVector{<:Union{Integer,ZZRingElem,QQFieldElem}}
```
HereAround: I do not yet understand why `QQFieldElem` is needed here. Can you explain? Also, maybe write this as union of `integerUnion` and `QQFieldElem`, so that future updates to IntegerUnion affect/arrive at this line.

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,49 @@ function cones(PF::_FanLikeType)
   return IncidenceMatrix([Vector{Int}(x) for x in cones])
 end
 
+@doc raw"""
+    minimal_supercone(X::NormalToricVariety, p::AbstractVector{<:IntegerUnion})
```
HereAround: This is a method in polyhedral fans, so the input should be a polyhedral fan. Right?

If you allow for polyhedral fans as input, then they can have lineality. Right? Where is that case treated in your code?

Final question: Why not just return a cone? Why is an iterator needed/better here? At least from the name of the method, I expect a cone as output.

HereAround: If you really want a toric variety as first argument, then probably this method should be moved to the toric code.

test/PolyhedralGeometry/polyhedral_fan.jl
```diff
@@ -119,6 +119,36 @@
   end
 end
 
+@testset "minimal_supercone" begin
```
HereAround: The following tests should be moved to the toric tests, as they test toric stuff. (Or you accept a polyhedral fan as first argument.)

HereAround (changes requested):
Some of the code in this PR is about toric varieties, but in the polyhedral folder. Two options:
1. Allow polyhedral fans as input and leave the code where it is.
2. Move the code to the toric varieties code/tests.

@lkastner will eyeball the `minimal_supercone` method. As currently is, this is likely not as efficient/performant as it could be.

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,49 @@ function cones(PF::_FanLikeType)
...
+  m_cone = intersect(m_cones)
+  m_fan = polyhedral_fan(m_cone)
+
+  # By Proposition 1.2.8(c) in [CLS11](@cite), every proper face of a cone
+  # is the intersection of the facets containing it.
+  m_facets = cones(m_fan, dim(m_cone) - 1)
+  cs = [c for c in m_facets if r in c]
+  isempty(cs) && return m_cone
+  return intersect(cs)
+end
```
lkastner: This is too expensive. Why are you intersecting cones that belong to the same fan? You should intersect the corresponding index sets. Why construct a fan out of `m_cone`? The intersection of cones `m_cone` is already the minimal cone containing `r`. I am lost for the remainder of this function. My suggestion is the following:

```suggestion
function minimal_supercone(
  X::NormalToricVariety, r::AbstractVector{<:Union{IntegerUnion,QQFieldElem}}
)
  m_cone_indices = Oscar._get_maximal_cones_containing_vector(X, r)
  @req !is_empty(m_cone_indices) "Ray `r` should be in the support of the fan of `X`."
  IX = maximal_cones(IncidenceMatrix, X)
  m_cone = intersect([row(IX, i) for i in m_cone_indices]...)
  RX = matrix(coefficient_field(X), rays(X))
  return positive_hull(RX[[m_cone...], :], lineality_space(X))
end
```

Probably @benlorenz has further improvements.

src/exports.jl
```diff
@@ -1068,6 +1068,7 @@ export minimal_block_reps
 export minimal_faces
 export minimal_generating_set
 export minimal_size_generating_set, has_minimal_size_generating_set, set_minimal_size_generating_set
+export minimal_supercone
```
lkastner: Do we need to export this?

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,49 @@ function cones(PF::_FanLikeType)
   return IncidenceMatrix([Vector{Int}(x) for x in cones])
 end
 
+@doc raw"""
+    minimal_supercone(X::NormalToricVariety, p::AbstractVector{<:IntegerUnion})
+    -> SubObjectIterator{RayVector}
```
lkastner:
```suggestion
```

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,49 @@ function cones(PF::_FanLikeType)
...
+
+Given an point $p$ with integer coordinates inside the support of the fan $Σ$ of a normal toric variety~$X$, return the unique cone $σ$ in $Σ$ such that $p$ is in the relative interior of $σ$.
```
lkastner: Please wrap

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+  J = total_transform(f, I)
+  return saturation_with_index(J, ideal(new_var))
+end
+
+@doc raw"""
+    cox_ring_group_homomorphism(f::ToricBlowupMorphism, g::MPolyDecRingElem) -> MPolyDecRingElem
+    cox_ring_group_homomorphism(f::ToricBlowupMorphism, I::MPolyIdeal) -> MPolyIdeal
+
+Let $f\colon Y \to X$ be the toric blowup corresponding to a star subdivision along a ray. Let $R$ and $S$ be the Cox rings of $X$ and $Y$, respectively.
+Considering $R$ and $S$ with their abelian group structures, there is a group homomorphism $\Phi\colon R \to S$ sending homogeneous ideals to homogeneous ideals with the defining property that the total transform of a subscheme of $X$, given by a homogeneous ideal $I$ in $R$, is the subscheme of $Y$ given by the homogeneous ideal $\Phi(I)$ in $S$.
```
lkastner: Please wrap

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+  return saturation(J, ideal(S, new_var))
+end
+
+@doc raw"""
+    strict_transform_with_index(f::ToricBlowupMorphism, I::MPolyIdeal) -> (MPolyIdeal, Int)
+
+Returns the pair $(J, k)$, where $J$ coincides with `strict_transform(f, I)` and where $k$ is the multiplicity of the total transform along the exceptional prime divisor.
+
+!!! note
+    If the multiplicity $k$ is not needed, we recommend to use `strict_transform(f, I)` which is typically faster.
```
lkastner: Please wrap

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
...
+function total_transform(f::ToricBlowupMorphism, I::MPolyIdeal)
+  return cox_ring_group_homomorphism(f, I)
+end
+
+@doc raw"""
+    strict_transform(f::ToricBlowupMorphism, I::MPolyIdeal) -> MPolyIdeal
+
+Let $f\colon Y \to X$ be the toric blowup corresponding to a star subdivision along a ray. Let $R$ and $S$ be the Cox rings of $X$ and $Y$, respectively.
+Here "strict transform" means the "scheme-theoretic closure of the complement of the exceptional divisor in the scheme-theoretic inverse image".
+This function returns a homogeneous ideal in $S$ corresponding to the strict transform under $f$ of the closed subscheme of $X$ defined by the homogeneous ideal $I$ in $R$.
```
lkastner: Please wrap

experimental/Schemes/src/ToricBlowups/methods.jl
```diff
@@ -1,3 +1,231 @@
+@doc raw"""
+    total_transform(f::ToricBlowupMorphism, I::MPolyIdeal) -> MPolyIdeal
+
+Let $f\colon Y \to X$ be the toric blowup corresponding to a star subdivision along a ray. Let $R$ and $S$ be the Cox rings of $X$ and $Y$, respectively.
+This function returns a homogeneous ideal in $S$ corresponding to the total transform (meaning the scheme-theoretic inverse image) under $f$ of the closed subscheme of $X$ defined by the homogeneous ideal $I$ in $R$.
```
lkastner: Please wrap

lkastner (changes requested):
Please update the PR description. Are these TODOs done now?

src/PolyhedralGeometry/PolyhedralFan/properties.jl
paemurru: > The intersection of cones m_cone is already the minimal cone containing r

No. If the fan has one maximal cone, then `m_cone` is just that one cone, which is not necessarily the minimal cone in the fan containing the point. In the attached picture, the point `(1, 0, 1)` is in the relative interior of a two-dimensional cone in the fan. That is the cone that this function returns.

Screenshot from 2024-12-06 15-13-45

src/exports.jl
paemurru: @simonbrandhorst recommended to me to export it since the code is documented, general and reasonably fast

src/PolyhedralGeometry/PolyhedralFan/properties.jl
paemurru: Why not show the output type?

src/PolyhedralGeometry/PolyhedralFan/properties.jl
paemurru: It would be good if we can only work with indices and perhaps just return the index of the cone. I was not sure how to do this

experimental/Schemes/src/ToricBlowups/methods.jl
paemurru: Since it is not actually a pullback and does not satisfy some expected properties of pullback (since the homomorphism is not generally a ring homomorphism), I would rather not call it pullback

experimental/Schemes/src/ToricBlowups/methods.jl
paemurru: It is not actually a ring homomorphism, only a homomorphism of modules. Also see https://github.com/oscar-system/Oscar.jl/pull/4154#discussion_r1873526484

docs/src/AlgebraicGeometry/ToricVarieties/BlowupMorphisms.md
paemurru: https://github.com/oscar-system/Oscar.jl/pull/4154#discussion_r1873528692

experimental/Schemes/src/ToricBlowups/methods.jl
afkafkafk13: I understand that this is not a pullback, but you do not return a group homomorphism either. Can you come up with any way to clearly phrase the name of this method so that users are not misled by the name? Is there a specific name for the returned object in literature?

paemurru: My initial thought was to actually return a group homomorphism `Phi`, and then let the user call `Phi(g)`. And store the group homomorphism as an attribute. But storing the homomorphism had some difficulties with function closures. So I instead turned it into taking two arguments, `f` and `g`

src/PolyhedralGeometry/PolyhedralFan/properties.jl
lkastner: You're right. My claim is even wrong when there are more maximal cones. I apologize for that.

lkastner: I will work something out using the Hasse diagram.

lkastner: In the end this is the same as your approach, but I managed to avoid expensive intersections:
```
julia> m_cone = positive_hull([1 0 0; 0 1 0; 0 0 1])

julia> r = [1,1, 0]
3-element Vector{Int64}:
 1
 1
 0

julia> r in m_cone
true

julia> Fm = facets(m_cone)
3-element SubObjectIterator{LinearHalfspace{QQFieldElem}} over the halfspaces of R^3 described by:
-x_1 <= 0
-x_2 <= 0
-x_3 <= 0


julia> zero_facets = findall(f -> f.a*r == [0], Fm)
1-element Vector{Int64}:
 3

julia> Rm = matrix(coefficient_field(m_cone), rays(m_cone))
[1   0   0]
[0   1   0]
[0   0   1]

julia> RIF = IncidenceMatrix(m_cone.pm_cone.RAYS_IN_FACETS)
3×3 IncidenceMatrix
[2, 3]
[1, 3]
[1, 2]


julia> result_ray_indices = intersect([row(RIF, i) for i in zero_facets]...)
Set{Int64} with 2 elements:
  2
  1

julia> result = positive_hull(Rm[[result_ray_indices...], :], lineality_space(m_cone))
```
So the general thing could look like:
```suggestion
function minimal_supercone(
  X::NormalToricVariety, r::AbstractVector{<:Union{IntegerUnion,QQFieldElem}}
)
  m_cone_indices = Oscar._get_maximal_cones_containing_vector(X, r)
  @req !is_empty(m_cone_indices) "Ray `r` should be in the support of the fan of `X`."
  IX = maximal_cones(IncidenceMatrix, X)
  m_cone = intersect([row(IX, i) for i in m_cone_indices]...)
  RX = matrix(coefficient_field(X), rays(X))
  m_cone = positive_hull(RX[[m_cone...], :], lineality_space(X))
  Fm = facets(m_cone)
  zero_facets = findall(f -> f.a*r == [0], Fm)
  Rm = matrix(coefficient_field(m_cone), rays(m_cone))
  RIF = IncidenceMatrix(m_cone.pm_cone.RAYS_IN_FACETS)
  result_ray_indices = intersect([row(RIF, i) for i in zero_facets]...)
  result = positive_hull(Rm[[result_ray_indices...], :], lineality_space(m_cone))
  return result
end
```
This avoids the expensive intersections, nevertheless it requires a convex hull computation. Might be more if the fan did not have rays before. Hope this works for you.

test/PolyhedralGeometry/polyhedral_fan.jl
```diff
@@ -119,6 +119,36 @@
...
+  X = projective_space(NormalToricVariety, 3)
+  c = Oscar.minimal_supercone(X, [1, 1, 1])
+  @test rays(c) isa SubObjectIterator{RayVector{QQFieldElem}}
+  @test sort(collect(rays(c))) == sort([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
```
lgoettgens:
```suggestion
  @test issetequal(rays(c)), [[1, 0, 0], [0, 1, 0], [0, 0, 1]])
```
and similar for the other tests below. Sorting needs resources that can be saved this way

paemurru: I don't think it works

```
julia>   X = projective_space(NormalToricVariety, 3)
Normal toric variety

julia>   c = Oscar.minimal_supercone(X, [1, 1, 1])

julia> sort(rays(c)) == sort([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
true

julia> issetequal(rays(c), [[1, 0, 0], [0, 1, 0], [0, 0, 1]])
false
```

lgoettgens: Hm, weird. I thought this was how @benlorenz adapted the other tests recently.

src/PolyhedralGeometry/PolyhedralFan/properties.jl
paemurru: The tests failed when some of the defined sets were empty. I pushed a commit that dealt with three such special cases in the above. Now the tests seem to pass, so I hope it is correct.

Returning the index of the resulting cone would also be OK if that is faster.

One side-comment that I am less familiar with polymake, so the variable names RIF, Fm, RX, IX, etc are very hard to read for me.

paemurru: Instead of tests with affine_space and projective_space, I guess the tests can be replaced by fan tests, so it can be entirely under PolyhedralGeometry

test/PolyhedralGeometry/polyhedral_fan.jl
paemurru: Perhaps the difference of types explains the output?

```
julia> typeof(rays(c))
SubObjectIterator{RayVector{QQFieldElem}}

julia> typeof([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
Vector{Vector{Int64}}
```

benlorenz: `issetequal` wont work if the hash-functions are not compatible which is not really feasible when comparing different types with different semantics.
This should work:
```suggestion
  @test issetequal(rays(c)), ray_vector.([[1, 0, 0], [0, 1, 0], [0, 0, 1]]))
```

paemurru: Updated the tests to use `issetequal` as @benlorenz suggested.

> which is not really feasible

It looks feasible to me:

```
julia> issetequal(Int64[1, 2], Int8[1, 2])
true

julia> issetequal(Int64[1, 2], ZZRingElem[1, 2])
true
```

It is not for this pull request, but perhaps the hash function of QQFieldElem should be altered.

```
julia> issetequal(Int64[1, 2], QQFieldElem[1, 2])
false
```

experimental/Schemes/src/ToricBlowups/methods.jl
paemurru: All of the polyhedral geometry checks will be much faster than saturation, so I think it is OK to always do them

experimental/Schemes/src/ToricBlowups/methods.jl
paemurru: https://github.com/oscar-system/Oscar.jl/pull/4154#discussion_r1884136347

src/PolyhedralGeometry/PolyhedralFan/standard_constructions.jl
paemurru: Changed to `RationalUnion`

src/PolyhedralGeometry/PolyhedralFan/properties.jl
paemurru: Moved to PolyhedralGeometry

test/PolyhedralGeometry/polyhedral_fan.jl
paemurru: Moved to PolyhedralGeometry

src/PolyhedralGeometry/PolyhedralFan/properties.jl
```diff
@@ -281,6 +281,163 @@ function cones(PF::_FanLikeType)
...
+
+julia> primitive_generator(r)
+2-element Vector{ZZRingElem}:
+ 2
+ -1
+```
+"""
+function primitive_generator(r::RayVector)
+  return vec(collect(matrix(ZZ, rays(positive_hull(r)))))
+end
```
lkastner:
```suggestion
function primitive_generator(r::RayVector)
  result = numerator.(lcm([denominator(i) for i in r]) * r)
  result = 1/gcd(result) * result
  convert(Vector{ZZRingElem}, result)
end
```

lkastner (changes requested):

src/PolyhedralGeometry/PolyhedralFan/properties.jl
lkastner: `positive_hull` and then asking for `rays` triggers a convex hull computation.
```
julia> t
3-element RayVector{QQFieldElem}:
 1
 5//4
 3//2

julia> @benchmark vec(collect(matrix(ZZ, rays(positive_hull(t)))))
BenchmarkTools.Trial: 242 samples with 1 evaluation.
 Range (min … max):  19.107 ms … 35.539 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     20.441 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   20.674 ms ±  1.326 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

          ▁ ▁██▁                                               
  ▂▂▄▄▄▄▄▅█▇████▇▆▅▅▅▄▃▃▄▃▄▃▁▁▁▁▁▁▂▂▂▁▁▁▁▁▁▁▁▂▁▁▂▁▁▁▁▁▁▁▁▂▂▁▂ ▃
  19.1 ms         Histogram: frequency by time        25.6 ms <

 Memory estimate: 10.81 KiB, allocs estimate: 932.

julia> @benchmark convert(Vector{ZZRingElem}, 1/gcd(numerator.(lcm([denominator(i) for i in t]) * t)) * (numerator.(lcm([denominator(i) for i in t]) * t)))
BenchmarkTools.Trial: 10000 samples with 8 evaluations.
 Range (min … max):  3.585 μs … 105.043 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.280 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.788 μs ±   1.813 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▄ ▁█▆▃▅▆▄▃▂▂▁▁▁           ▁▁                                ▂
  ██████████████████▇▆▇▆▇█▇███▇▇▇▇▇▆▅▅▅▃▄▁▁▄▄▃▄▁▁▁▃▁▄▁▁▃▁▁▁▄▇ █
  3.58 μs      Histogram: log(frequency) by time      13.7 μs <

 Memory estimate: 2.88 KiB, allocs estimate: 107.

julia> convert(Vector{ZZRingElem}, 1/gcd(numerator.(lcm([denominator(i) for i in t]) * t)) * (numerator.(lcm([denominator(i) for i in t]) * t))) == vec(collect(matrix(ZZ, rays(positive_hull(t)))))
true
```

paemurru: Great improvement, thanks!

lkastner (approved):

HereAround (approved):
Thank you for your hard work on this PR @paemurru and @lkastner . Looks great!

PR #4157, opened 2024-09-27
Improve pc group collector objects, add `get_conjugate`, `get_power`, `get_relative_order`, `get_relative_orders`

ThomasBreuer:
create an Oscar collector from a GAP collector

(suggested by @fieker)

### DIFF
--- a/src/Groups/pcgroup.jl
+++ b/src/Groups/pcgroup.jl
@@ -51,7 +51,7 @@ true
 collector(n::Int, ::Type{T} = ZZRingElem) where T <: IntegerUnion = GAP_Collector{T}(n)
 
 
-# Provide functions for entering data into the collector.
+# Provide functions for entering data into the collector and accessing data.
 
 # utility:
 # Convert a vector of pairs to a generator-exponent vector in GAP.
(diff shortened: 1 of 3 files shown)


test/Groups/pcgroup.jl
```diff
@@ -82,3 +82,13 @@ end
   @test GAP.Globals.IsMutable(cgg)
   @test cgg !== c.X
 end
+
+@testset "create collectors from polycyclic groups" begin
+  for i in rand(1:number_small_groups(96), 10)
```
lgoettgens:
```suggestion
  for i in rand(1:number_of_small_groups(96), 10)
```
the old version is deprecated

ThomasBreuer: o.k.

src/Groups/pcgroup.jl
```diff
@@ -199,6 +265,31 @@ function set_conjugate!(c::Collector{T}, j::Int, i::Int, rhs::Vector{Pair{Int, T
...
+
+julia> get_conjugate(c, 2, 1)
+1-element Vector{Pair{Int64, Int64}}:
+ 2 => 2
+```
+"""
+function get_conjugate(c::Collector{T}, j::Int, i::Int) where T <: IntegerUnion
+  @req 0 < i <= c.ngens "the collector has only $(c.ngens) generators not $i"
+  @req i < j "only for i < j, but i = $i, j = $j"
+  return c.conjugates[i,j]
```
fieker: This will fail if the entries is trivial, ie. for abelian groups:
```
G = abelian_group(PcGroup, [2,3]);
c = collector(G);
julia> c.conjugates
2×2 Matrix{Vector{Pair{Int64, ZZRingElem}}}:
 #undef  #undef
 #undef  #undef
```
so every access (`get_conjugate`) will fail

ThomasBreuer: Yes, thanks. I have fixed this.

ThomasBreuer:
restart the tests ...

fieker (approved):

Issue #3983, opened 2024-07-30
Serialization of empty julia containers forgets type parameters


lgoettgens:
**Describe the bug**
Consider an empty vector of type `Vector{Foo{Bar}}`, i.e. the eltype of the vector itself has a type parameter. If I save this vector, and load it again (it does not matter if loading happens in the same session or a new one), it will be the empty vector of type `Vector{Foo}`.
In particular, running `test_save_load_rountrip` from https://github.com/oscar-system/Oscar.jl/blob/master/test/Serialization/setup_tests.jl with such a vector fails.
The same problem arises for 0x0 matrices and empty dicts instead of vectors

**To Reproduce**
Steps to reproduce the behavior, please provide a code snippet that triggers
the bug.
```
using Oscar, Test

x1 = Vector{Vector{Int}}() # empty vector
save("tmp.mrdi", x1)
x2 = load("tmp.mrdi")
@test x2 isa typeof(x1)

using JSONSchema
include("test/Serialization/setup_tests.jl")
mktempdir() do path
  test_save_load_roundtrip(path, x1) do loaded
    # nothing here
  end
end
```
the same problem arises for e.g. 
`x1 = Vector{MPolyRingElem{QQFieldElem}}()`
`x1 = Dict{MPolyRingElem{QQFieldElem}, Int}()`
`x1 = Dict{Int, MPolyRingElem{QQFieldElem}}()`

For the following two empty containers, even trying to load raises a `KeyError`.
`x1 = Matrix{QQMatrix}(undef, 0, 0)` (will no longer error but still have the problem once #3980 is merged)
`x1 = Dict{QQMatrix, Int}()`
`x1 = Dict{Int, QQMatrix}()`

**Expected behavior**
All of the above containers should remember its eltype when saving and loading, as they already do if they contain some elements.

**System (please complete the following information):**
```
julia> Oscar.versioninfo(full=true)
OSCAR version 1.2.0-DEV - #master, bfcfee360a -- 2024-07-28 10:10:11 +0000
  combining:
    AbstractAlgebra.jl   v0.42.0
    GAP.jl               v0.10.4
    Hecke.jl             v0.33.0
    Nemo.jl              v0.46.0
    Polymake.jl          v0.11.19
    Singular.jl          v0.23.4
  building on:
    FLINT_jll               v300.100.300+0
    GAP_jll                 v400.1200.200+9
    Singular_jll            v404.0.301+0
[...]

PR #4162, opened 2024-09-28
Serialization Refactor

### LINKED ISSUE #4567: Serialization: problem while saving a list of tuples of matrices with varying sizes
**Describe the bug**
It seems that there is a bug while saving a list of tuples of matrices. In fact, it seems that saving or loading assumes wrong size of matrices which makes data unable to be loaded.

**To Reproduce**
```
using Oscar
julia> m = (matrix(QQ, 1, 1, [2]), matrix(QQ, 2, 2, [1//2 0; 0 1//2]))
([2], [1//2 0; 0 1//2])

julia> mm = [m, reverse(m)]
2-element Vector{Tuple{QQMatrix, QQMatrix}}:
 ([2], [1//2 0; 0 1//2])
 ([1//2 0; 0 1//2], [2])

julia> save("Schreibtisch/tmp.mrdi", mm)

julia> load("Schreibtisch/tmp.mrdi")
┌ Warning: Attempted loading file stored using a DEV version with commit 8fc7969f373c61cb9bab612fe608e89c9d2d7d0c
└ @ Oscar ~/.julia/dev/Oscar/src/Serialization/main.jl:712
ERROR: Expected dimension 1 x 1, got 2 x 2
Stacktrace:
  [1] _check_dim
    @ ~/.julia/packages/AbstractAlgebra/bL4Gd/src/Matrix.jl:46 [inlined]
  [2] _check_dim
    @ ~/.julia/packages/AbstractAlgebra/bL4Gd/src/Matrix.jl:44 [inlined]
  [3] (::MatSpace{QQFieldElem})(arr::Matrix{QQFieldElem})
    @ Nemo ~/.julia/packages/Nemo/E8ekQ/src/flint/fmpq_mat.jl:718
  [4] load_object(s::Oscar.DeserializerState{Oscar.JSONSerializer}, ::Type{MatElem}, parents::Vector{Union{Ring, FreeAssociativeAlgebra, MatSpace, SMatSpace}})
    @ Oscar ~/.julia/dev/Oscar/src/Serialization/Rings.jl:417
[...]
 [31] load(filename::String)
    @ Oscar ~/.julia/dev/Oscar/src/Serialization/main.jl:718
 [32] top-level scope
    @ REPL[13]:1
```

**Expected behavior**
Since we know how to serialize lists, tuples and matrices, I would expect to be able to serialize a list of tuples of matrices of various sizes.

**System (please complete the following information):**
Please paste the output of `Oscar.versioninfo(full=true)` below. If this does
not work, please paste the output of Julia's `versioninfo()` and your Oscar
version.
```
OSCAR version 1.3.0-DEV - #main, 8fc7969f37 -- 2025-02-08 18:29:40 +0000
  combining:
    AbstractAlgebra.jl   v0.44.5
    GAP.jl               v0.13.1
    Hecke.jl             v0.35.9
    Nemo.jl              v0.48.2
    Polymake.jl          v0.11.25
    Singular.jl          v0.24.1
  building on:
    FLINT_jll               v300.100.301+0
    GAP_jll                 v400.1400.3+0
    Singular_jll            v404.1.700+0
    libpolymake_julia_jll   v0.13.1+0
    libsingular_julia_jll   v0.46.1+0
    polymake_jll            v400.1300.2+0

Julia Version 1.10.1
Commit 7790d6f0641 (2024-02-13 20:41 UTC)
Build Info:
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, skylake)
Threads: 1 default, 0 interactive, 1 GC (on 8 virtual cores)
...
```

antonydellavecchia:
Refactor serialization to use a new function called `type_params` which helps build a stack of type dependencies which will allow for serializing type information before the data.

### Release notes:
manually copy from https://github.com/oscar-system/Oscar.jl/pull/4162#issuecomment-2893356751


lkastner (dismissed):

lgoettgens:
This PR seems to have diverged a lot from current master due to other changes to the serialization code in the meantime. Could you try to resolve the conflicts / rebase this onto a more current master, so that one can more easily follow what's happening and CI runs again?

I am a bit afraid that this just continues to diverge even further in the future, which makes it harder to resolve conflicts once you want to merge it.

src/Combinatorics/SimplicialComplexes.jl
```diff
@@ -670,7 +670,7 @@ function automorphism_group(K::SimplicialComplex; action=:on_vertices)
 end
 
 @doc raw"""
-    on_simplicial_complex(K::SimplicialComplex, g::PermGroupElem)
+    on_simplicial_complex(K::SimplicialComplex, g::PermGroupElem; action=:on_vertices)
```
lgoettgens: is this maybe a conflict resolution artifact? it does not match the signature of the function this docstring is attached to

src/Serialization/GAP.jl
```diff
@@ -125,43 +143,47 @@ install_GAP_deserialization(
...
+          GapObj(names; recursive = true)
+        end
+        G = GAP.Globals.FreeGroup(wfilt, init)::GapObj
       end
-      return G
     end
+    return G
+  end)
+
+install_GAP_deserialization(
```
antonydellavecchia, line 187: @ThomasBreuer  I think I need your help here, and it also makes sense for you to understand what I am doing here anyway.

I am currently try to get the serialization refactor working for gap objects, and the first one that causes an issue is when we try and save subgroups of free groups


If you see above I have a new function called `type_params` and this handles when params need to be added to the type description for example saving a subgroup of a free group currently looks like this

```
{
  "_ns": {
    "Oscar": [
      "https://github.com/oscar-system/Oscar.jl",
      "1.3.0-DEV-0324e87a589a2c2c5b6f4b0db4634b14477250ae"
    ]
  },
  "_refs": {
    "fef8f069-d2db-4c88-9118-f2b7ba1c3999": {
      "_type": "GapObj",
      "data": {
        "GapType": "IsFreeGroup",
        "names": [
          "f1",
          "f2"
        ],
        "wfilt": "IsSyllableWordsFamily"
      }
    }
  },
  "_type": {
    "name": "GapObj",
    "params": "fef8f069-d2db-4c88-9118-f2b7ba1c3999"
  },
  "data": {
    "GapType": "IsFreeGroup",
    "gens": [
      [
        "1",
        "1"
      ]
    ]
  },
  "id": "ab56be5f-1bf6-4fec-a3d5-35cb59b6096b"
}

```

Note how there is no longer a freeGroup key in the data section and that the type information of the GapObj now has a params key.

The issue I am having now is that I need to pass the free group down, but this means I need two different function signatures. 

But I am guessing these cannot share the same namespace in GAP?

please let me know if you have any questions about some of the changes here

ThomasBreuer: @antonydellavecchia I am not sure whether I understand the aim of the proposed changes.
A subgroup of a full free group is best described in terms of generators relative to the full free group, hence the "old" approach first serializes the full free group and then the generators of the subgroup in question, and the deserialization first reconstructs the full free group and then the subgroup generators w.r.t. the full group.

The idea of the "new" approach seems to be that the function `type_params` shall describe the object(s) to be serialized together with the data of the given object.
The problem is that `type_params` for a `GapObj` has to deal with a lot of different `GapObj`s.
In the "old" serialization and deserialization of GAP objects, the approach using `install_GAP_serialization` and `install_GAP_deserialization` yields small Julia functions for different kinds of GAP objects. One could think of an analogous `install_GAP_type_params` in order to achieve this for `type_params` iin the "new" approach, but what is the advantage of using `type_params`, compared to the "old" (recursive) serialization approach?

Just for curiosity: How does the "new" approach handle objects that need more than one dependent object to be serialized?
(For example, the most natural way to serialize a map seems to be to serialize domain and codomain and information that defines the function in question.)
These dependent objects are stored in `_refs`, their keys in `_refs` shall be referenced somewhere in `_type`, but how does the deserialization code know which is which? In the "old" approach, the references to the keys in `_refs` are stored in fields inside `data` (like the `freeGroup` field), whose meanings are defined by the serialization.

antonydellavecchia: The idea is that type params will do a shallow pass through the object and obtain all necessary type information for the object. This has many benefits, the main one being that this would allow us to put all type information at the top of the file and hence we would be able to read in all params for the object before we get to the data in a linear way. 

here is another example of how a the new format looks for a MPolyAnyMap.
You can see that the params section can also have keys, i.e. we can have params that are dictionnaries if necessary
(this does not yet include the linear ordering)

```
{
  "_ns": {
    "Oscar": [
      "https://github.com/oscar-system/Oscar.jl",
      "1.3.0-DEV-98fc56d6f49a6eacecd525b01c0e543857608f58"
    ]
  },
  "_refs": {
    "bede6628-38bf-4803-88df-2ed7487fc8c8": {
      "_type": {
        "name": "MPolyRing",
        "params": {
          "_type": "QQField"
        }
      },
      "data": {
        "symbols": [
          "s",
          "t",
          "m"
        ]
      }
    },
    "fc49a9a8-91af-4b54-888b-58b4f2dd1958": {
      "_type": {
        "name": "MPolyRing",
        "params": {
          "_type": "QQField"
        }
      },
      "data": {
        "symbols": [
          "x",
          "y"
        ]
      }
    }
  },
  "_type": {
    "name": "MPolyAnyMap",
    "params": {
      "codomain": {
        "name": "MPolyRing",
        "params": "bede6628-38bf-4803-88df-2ed7487fc8c8"
      },
      "domain": {
        "name": "MPolyRing",
        "params": "fc49a9a8-91af-4b54-888b-58b4f2dd1958"
      }
    }
  },
  "data": {
    "images": [
      [
        [
          [
            "1",
            "1",
            "0"
          ],
          "1"
        ]
      ],
      [
        [
          [
            "0",
            "0",
            "1"
          ],
          "1"
        ]
      ]
    ]
  }
}
```

ThomasBreuer: Thanks @antonydellavecchia .
Thus the rule is that it is recommended (or mandatory?) to "announce" the dependent objects via `type_params`.

Then the interface for the (de)serialization for a given object type `T` (here: `GapObj`) consists of
- a `save_object` method with second argument `T`,
- a `load_object` method with two or three arguments, the second argument being `T`,
- a `type_params` method with one argument of type `T`, returning `S, X` where `X` is either `nothing` or an object of type `S` (a dictionary or a `GapObj`); in the latter case, there must be a three argument method for `load_object` that takes `X` as its third argument.

If this is correct then I propose a function `install_GAP_type_params` as sketched above for the case of `GapObj`. I can provide the code for that.

antonydellavecchia: The rule will become mandatory.

the load object third parameters will exactly be the second entry of the output from `type_params`

the output of `type_params` should be a tuple where the first entry is the type and the second entry are the params for that type, for example above type_params with return the type  
```
PolyAnyMap, Dict(:codomain => QQ[:s, :t, :m], :domain => QQ[:x, :y])
```

for a  group element g It would be some thing like 

```
GroupElem, parent(g)
```

and yes exactly, if the second output of `type_params` is nothing, then that particular load_object method only needs two parameters

ThomasBreuer:
The tests for pc groups are still disabled. Here the problem is that `load_object(s, Vector, (Tuple, [Int, Int, [Vector, Int]]), :comm_rels)` does not work anymore.

antonydellavecchia:
> The tests for pc groups are still disabled. Here the problem is that `load_object(s, Vector, (Tuple, [Int, Int, [Vector, Int]]), :comm_rels)` does not work anymore.

Thanks for that. I've made the loading of vectors a little more intuitive. 
Please see the changes I've just pushed to get the GAP object serialization working

ThomasBreuer:
> Please see the changes I've just pushed to get the GAP object serialization working

Thanks @antonydellavecchia.
~~I am going to adjust the serialization for GAP pc groups.~~
Now the tests for GAP pc groups pass.

src/Serialization/LieTheory.jl
```diff
@@ -112,36 +93,27 @@ end
 
 @register_serialization_type WeylGroup uses_id
 
+type_params(W::WeylGroup) = WeylGroup, root_system(W)
```
lgoettgens: Stupid question maybe: It seems to me that the first return value of `type_params` is always the argument type. Is this a correct observation?
If yes, wouldn't it be possible to refactor that away into the calling function, so that `type_params` loses the first return value? Imo, this could make the code a bit more readable. If there is a reason for this, I am happy to learn about it

antonydellavecchia: maybe there is a way to get a round it, but it was the only way I found I could get the new `save_type_params` to work. See main.jl

src/Serialization/main.jl
```diff
@@ -188,57 +171,133 @@ function save_typed_object(s::SerializerState, x::T, key::Symbol) where T
...
   save_type_params(s, obj)
 end
 
-function save_attrs(s::SerializerState, obj::T) where T
-  !with_attrs(s) && return 
-  if any(attr -> has_attribute(obj, attr), attrs_list(s, T))
-    save_data_dict(s, :attrs) do
-      for attr in attrs_list(s, T)
-        has_attribute(obj, attr) && save_typed_object(s, get_attribute(obj, attr), attr)
+save_type_params(s::SerializerState, obj::Any) = save_type_params(s, type_params(obj)...)
```
lgoettgens:
```suggestion
save_type_params(s::SerializerState, obj::Any) = save_type_params(s, typeof(obj), type_params(obj)...)
```

I think this could do the trick

antonydellavecchia: yes this is probably fine

antonydellavecchia: actually, a little trickier than expected, there is an issue with the containers. 

I'll have to dig deeper, but I dont have time now

antonydellavecchia: I've pushed a fix for this now

src/Serialization/Groups.jl
```diff
@@ -157,7 +139,7 @@ end
 
 function load_object(s::DeserializerState, ::Type{PermGroup})
   n = load_object(s, Int, :degree)
-  generators = load_object(s, Vector, (Vector{Int}, Int), :gens)
+  generators = load_object(s, Vector{Vector{Int}}, :gens)
   G = permutation_group(n, [perm(x) for x in generators])
   load_attrs(s, G)
```
lgoettgens:
```suggestion
```
this can go, right? And similarly the `save_attrs` 6 lines above

antonydellavecchia: yes

lgoettgens:
In #4563, I encountered more instances of https://github.com/oscar-system/Oscar.jl/issues/3983. It would be great if you could have a look if the new loading code allows for an easy fix of this. (at some point in the refactor, not right now)

antonydellavecchia:
closes #4567

src/Serialization/Upgrades/1.3.0.jl
```diff
@@ -1,16 +1,105 @@
+function upgrade_NamedTuple(d::Dict)
```
lgoettgens:
```suggestion
module SerializationUpgrade-1-3-0

function upgrade_NamedTuple(d::Dict)
```
It would be good to put all of these `upgrade_*` functions into their own shbmodule to not clutter the global namespace. And the next time someone needs to write an upgrade script, it could otherwise lead to name clashes.

antonydellavecchia: This was only temporary because I thought that revise couldn't handle updates inside the function that got pushed into the function list.
Also, it was unsafe because I was evalutating strings from files

antonydellavecchia:
files of all types in src that have serialization tests.
[files_to_update.zip](https://github.com/user-attachments/files/18980076/files_to_update.zip)

lgoettgens:
Would it be possible to split this PR into two separate ones before merging? The github UI doesn't quite like >300 changed files, thus it is not possible to look at the changes that this introduces.
If I understand it correctly, it should be possible to split into the following two parts:
1. All changes to the serialization, upgrade scripts etc.
2. Upgrading everything in `data/` and the artifacts

As the upgrade scripts are contained in 1, the testsuite should even succeed in-between the two PRs. And then one would only need to look at 1 to see the "real" changes. But all of the many completely changed files that github trips over would be part of 2.

I imagine this not being too difficult, but I could be wrong.

src/Combinatorics/SimplicialComplexes.jl
```diff
@@ -637,7 +637,7 @@ function deletion(K::SimplicialComplex, face::Union{<:AbstractSet{Int},<:Abstrac
 end
 
 @doc raw"""
-    automorphism_group(K::SimplicialComplex; action=:on_vertices)
+    automorphism_group(K::SimplicialComplex)
```
benlorenz:
```suggestion
    automorphism_group(K::SimplicialComplex; action=:on_vertices)
```

src/Combinatorics/SimplicialComplexes.jl
```diff
@@ -670,7 +670,7 @@ function automorphism_group(K::SimplicialComplex; action=:on_vertices)
 end
 
 @doc raw"""
-    on_simplicial_complex(K::SimplicialComplex, g::PermGroupElem)
+    on_simplicial_complex(K::SimplicialComplex, g::PermGroupElem;)
```
benlorenz:
```suggestion
    on_simplicial_complex(K::SimplicialComplex, g::PermGroupElem)
```

src/PolyhedralGeometry/helpers.jl
```diff
@@ -317,7 +317,11 @@ function (NF::Hecke.EmbeddedNumField)(x::Polymake.QuadraticExtension{Polymake.Ra
 end
 
 (F::Field)(x::Polymake.Rational) = F(QQ(x))
-(F::Field)(x::Polymake.OscarNumber) = F(Polymake.unwrap(x))
+function (F::Field)(x::Polymake.OscarNumber)
+  #println(Oscar.global_serializer_state.obj_to_id[F])
+  #println(Oscar.global_serializer_state.obj_to_id[parent(Polymake.unwrap(x))])
+  F(Polymake.unwrap(x))
+end
```
antonydellavecchia:
```suggestion
(F::Field)(x::Polymake.OscarNumber) = F(Polymake.unwrap(x))
```

benlorenz:
```suggestion
(F::Field)(x::Polymake.OscarNumber) = F(Polymake.unwrap(x))
```

src/Serialization/Upgrades/1.2.2.jl
benlorenz, line 1: please change this file back to 1.3.0 (filename + content) to make it consistent with the 1.3.0 release

src/Serialization/Upgrades/main.jl
```diff
@@ -141,7 +141,8 @@ include("0.13.0.jl")
 include("0.15.0.jl")
 include("1.1.0.jl")
 include("1.2.0.jl")
-include("1.3.0.jl")
+include("1.2.2.jl")
```
benlorenz:
```suggestion
include("1.3.0.jl")
```

lgoettgens:
The macos failures are probably due to https://github.com/oscar-system/Oscar.jl/issues/4714

fingolfin (approved):
Fine by me.

benlorenz (approved):

lkastner (approved):

lgoettgens:
@antonydellavecchia could you please come up with approx. 2 sentences about this PR that we can put into the v1.4 release notes?

antonydellavecchia:
Cleaner handling of type parameter serialization. This update forces entries of container types to share the same output of Oscar.type_params when serializing. Deserialization speed improvements.

PR #4163, opened 2024-09-30
Update book example for GenericCharacterTables

fingolfin:
This syncs it with the revised version in the book repo.

I've not tested this locally yet and will fix it up later today. So while I won't complain if someone wants to jump in and help out with any issues in here, please don't feel obliged. Ideally I'll complete this and then afterwards @benlorenz and @lkastner can have a final sanity check / point out anything I also should adjust.

I also plan to update https://github.com/oscar-system/book.oscar-system.org later. Let me know if there are other places that need to be adjusted.

### DIFF
--- a/test/book/cornerstones/groups/auxiliary_code/main.jl
+++ b/test/book/cornerstones/groups/auxiliary_code/main.jl
@@ -1,5 +1,5 @@
 import Pkg
-Pkg.add(name="GenericCharacterTables", version="0.2"; io=devnull)
+Pkg.add(name="GenericCharacterTables", version="0.4"; io=devnull)
 using GenericCharacterTables
 # for nicer printing
 using GenericCharacterTables: ParameterException
--- a/test/book/cornerstones/groups/genchar.jlcon
+++ b/test/book/cornerstones/groups/genchar.jlcon
@@ -1,63 +1,66 @@
-julia> T = genchartab("SL3.n1")
-Generic character table
+julia> T = generic_character_table("SL3.n1")
+Generic character table SL3.n1
   of order q^8 - q^6 - q^5 + q^3
   with 8 irreducible character types
   with 8 class types
   with parameters (a, b, m, n)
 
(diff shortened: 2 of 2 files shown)


fingolfin:
Oops, there is a warning
```
WARNING: importing deprecated binding AbstractAlgebra.addeq! into GenericCharacterTables.
, use add! instead.
```
we'll take care of it.

@SoongNoonien can you please remove the `addeq!` import & replace the `addeq!` method? Something like this should do: replace
```
add!(x::GenericCyclo, y::GenericCyclo, z::GenericCyclo) = y+z

function addeq!(x::GenericCyclo, y::GenericCyclo)
	mergewith!(+, x.f, y.f)
	strip_zeros!(x.f)
	return x
end
```
by
```
function add!(x::GenericCyclo, y::GenericCyclo, z::GenericCyclo)
        if x !== y
		x.f = deepcopy(y.f)
	end
	mergewith!(+, x.f, z.f)
	strip_zeros!(x.f)
	return x
end
```
(untested, and I may have missed something else, so please double check it and ideally add a test case?)

fingolfin:
Apparently there are deeper issues running this against Oscar master (it works against the latest Oscar release). We'll resolve this, too

lgoettgens:
I don't think your fix works here. Two reasons:
1. In the case of `x !== y`, do you really wanna copy `y.f`, or should it rather just set `x.f` to something empty?
2. without a release of Oscar.jl, GenericCharacterTables.jl will not get hold of the new AbstractAlgebra version, that does these changes. (This of course doesn't mean you cannot work on a fix right now, but the CI in GCT will probably fail until we get an Oscar release)

SoongNoonien:
> In the case of x !== y, do you really wanna copy y.f, or should it rather just set x.f to something empty?

Setting `x.f` to something empty would mean that `add!` just returns something with the value of `z`.

benlorenz (approved):
Looks reasonable and seems to work fine.

PR #4164, opened 2024-09-30
Don't collect iterators in iteration

lgoettgens:
the old code first used the iterator exhaustively while saving all elements into a vector and then afterwards do a for-each thing over the vector. the new code directly does a for-each over the iterator.
Thus the only change is that we do not need to allocate memory for the intermediate vector.


afkafkafk13 (approved):
Yes, indeed. This should save memory.

PR #4166, opened 2024-09-30
Implement more shortcuts and remove code duplication.

HechtiDerLachs:
I wanted to push some more things for the conversion which I had in mind. 

Running `@time Oscar.test_module("test/Modules/", new=false)`
Before:
```
41.459414 seconds (185.33 M allocations: 9.765 GiB, 20.17% gc time, 22.57% compilation time)
```
after:
```
 36.911680 seconds (184.97 M allocations: 9.754 GiB, 20.81% gc time, 21.16% compilation time)
```
Not too significant, but at least we get rid of the code duplication. ~If there was some in-place version of `map_entries`, for `SRow`s, then we could probably get some more gain.~

### DIFF
--- a/src/Modules/UngradedModules/ModuleGens.jl
+++ b/src/Modules/UngradedModules/ModuleGens.jl
@@ -268,22 +268,25 @@ function (F::FreeMod{<:MPolyRingElem})(s::Singular.svector)
 end
 
 function (F::FreeMod{<:MPolyQuoRingElem})(s::Singular.svector)
-  is_zero(s) && return zero(F) # `map_entries` throws for empty vectors
   Qx = base_ring(F)::MPolyQuoRing
   Rx = base_ring(Qx)::MPolyRing
-  row = _build_sparse_row(Rx, s)
-  return FreeModElem(map_entries(Qx, row), F)
+  row = _build_sparse_row(Rx, s; cast=Qx)
+  return FreeModElem(row, F)
 end
 
-function _build_sparse_row(Rx::Ring, s::Singular.svector)
+function _build_sparse_row(
+    Rx::MPolyRing, s::Singular.svector; 
+    cast::Ring=Rx
+  )
+  is_zero(length(s)) && return sparse_row(cast)
   R = coefficient_ring(Rx)
-  ctx = MPolyBuildCtx(Rx)
 
   # shortcuts in order not to allocate the dictionary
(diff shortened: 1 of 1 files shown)


src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -263,47 +263,44 @@ Convert a Singular vector to a free module element.
...
+  return FreeModElem(map_entries(Qx, row), F)
+end
+
+function _build_sparse_row(Rx::Ring, s::Singular.svector)
   R = coefficient_ring(Rx)
   ctx = MPolyBuildCtx(Rx)
 
-  # shortcut in order not to allocate the dictionary
+  # shortcuts in order not to allocate the dictionary
+  is_zero(length(s)) && return sparse_row(Rx)
```
fingolfin: If you move it up some more you can also avoid creating `ctx` :-)

src/Modules/UngradedModules/ModuleGens.jl
```diff
@@ -263,47 +263,44 @@ Convert a Singular vector to a free module element.
...
   if isone(length(s))
     (i, e, c) = first(s)
     push_term!(ctx, R(c), e)
-    return FreeModElem(sparse_row(Qx, [(i, Qx(finish(ctx)))]), F)
+    return sparse_row(Rx, [(i, finish(ctx))])
   end
 
   cache = IdDict{Int, typeof(ctx)}()
```
thofma, line 293:
```suggestion
  cache = Dict{Int, typeof(ctx)}()
```
A `Dict` is faster if the keys are of type `Int`.

HechtiDerLachs: Oh! I didn't know. Thanks.

Edit: My measurements indicate that this is not the case here. But there was a lot of noise.

src/Modules/UngradedModules/ModuleGens.jl
HechtiDerLachs: True. It was late.

HechtiDerLachs:
No big improvement in runtime, but removes some code duplication. Good to go from my side.

afkafkafk13 (commented):
Looks fine to me. Merge at the triage meeting?

fingolfin (approved):

PR #4173, opened 2024-10-02
Change `eltype` methods for objects to types

lgoettgens:
The docstring of `eltype` states
> However the form that accepts a type argument should be defined for new types. 

Chaning this should help julia infer types of iterator constructions like `Iterators.reverse(...)` for these types.

### DIFF
--- a/src/InvariantTheory/iterators.jl
+++ b/src/InvariantTheory/iterators.jl
@@ -57,7 +57,7 @@ function AllMonomials(R::MPolyDecRing, d::Int, vars::Vector{Int})
   return AllMonomials{typeof(R)}(R, d, vars)
 end
 
-Base.eltype(AM::AllMonomials) = elem_type(AM.R)
+Base.eltype(::Type{AllMonomials{PolyRingT}}) where {PolyRingT} = elem_type(PolyRingT)
 
 Base.length(AM::AllMonomials) = length(AM.weak_comp_iter)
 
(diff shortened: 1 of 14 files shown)


src/InvariantTheory/iterators.jl
```diff
@@ -391,6 +391,7 @@ function iterate_basis_linear_algebra(IR::FinGroupInvarRing, d::Int)
 end
 
 Base.eltype(BI::FinGroupInvarRingBasisIterator) = elem_type(polynomial_ring(BI.R))
+# TODO: change this to Base.eltype(::Type{<:FinGroupInvarRingBasisIterator{FinGroupInvarRingT}}) where {FinGroupInvarRingT} = ...
```
lgoettgens: @joschmitt do you maybe have an idea how to do this more cleanly? I am a bit overwhelmed by the amount of type vars and what to use here

joschmitt: Should be possible, I'll have a look tomorrow.

test/PolyhedralGeometry/subdivision_of_points.jl
```diff
@@ -14,8 +14,8 @@
     SubdivisionOfPoints
 
   @testset "alternative inputs" begin
-    @test collect(maximal_cells(square_by_incidence)) ==
-      collect(maximal_cells(square_by_weights))
+    @test issetequal(collect(maximal_cells(square_by_incidence)),
+      collect(maximal_cells(square_by_weights)))
```
lgoettgens, line 18: @benlorenz just fyi I patched another one (or the same, cannot remember) of these flaky tests, as it failed here

benlorenz: I think this is the same one.
I am a bit surprised why this is now happening again and again without any related changes (and only on the macos runner?). @alexej-jordan is working on fixing this and more similar tests.

src/InvariantTheory/iterators.jl
joschmitt: Should be resolved.

fingolfin:
Code formatting CI test fails

joschmitt:
> Code formatting CI test fails

Yeah, my fault, fixed.

fingolfin (approved):

src/InvariantTheory/iterators.jl
lgoettgens: Thanks,!

PR #4177, opened 2024-10-04
Elliptic surfaces fixes

HechtiDerLachs:
* Fixes a bug in `inherit_decomposition_info!`
* Repairs in `has_dim_leq_zero` 
* Disables some superfluous internal check for the elliptic surfaces.

### DIFF
--- a/experimental/Schemes/src/elliptic_surface.jl
+++ b/experimental/Schemes/src/elliptic_surface.jl
@@ -1649,6 +1649,19 @@ function extended_ade(ADE::Symbol, n::Int)
   return -G, kernel(G; side = :left)
 end
 
+# This function allows to store a reduction map to positive characteristic,
+# e.g. for computing intersection numbers.
+function reduction_to_pos_char(X::EllipticSurface, red_map::Map)
+  return get_attribute!(X, :reduction_to_pos_char) do
+    kk0 = base_ring(X)
+    @assert domain(red_map) === kk0
+    kkp = codomain(red_map)
+    @assert characteristic(kkp) > 0
+    _, result = base_change(red_map, X)
+    return red_map, result
+  end::Tuple{<:Map, <:Map}
+end
+
 @doc raw"""
     basis_representation(X::EllipticSurface, D::WeilDivisor)
 
@@ -1661,9 +1674,21 @@ function basis_representation(X::EllipticSurface, D::WeilDivisor)
   n = length(basis_ambient)
   v = zeros(ZZRingElem, n)
   @vprint :EllipticSurface 3 "computing basis representation of $D\n"
-  for i in 1:n
-    @vprintln :EllipticSurface 4 "intersecting with $(i): $(basis_ambient[i])"
-    v[i] = intersect(basis_ambient[i], D)
+  kk = base_ring(X)
+  if iszero(characteristic(kk)) && has_attribute(X, :reduction_to_pos_char)
+    red_map, bc = get_attribute(X, :reduction_to_pos_char)
+    for i in 1:n
+      @vprintln :EllipticSurface 4 "intersecting with $(i): $(basis_ambient[i])"
+      
+      v[i] = intersect(base_change(red_map, basis_ambient[i]; scheme_base_change=bc), 
+                       base_change(red_map, D; scheme_base_change=bc))
+    end
+  else
+    for i in 1:n
+      @vprintln :EllipticSurface 4 "intersecting with $(i): $(basis_ambient[i])"
+
+      v[i] = intersect(basis_ambient[i], D)
+    end
   end
   @vprint :EllipticSurface 3 "done computing basis representation\n"
   return v*inv(G)
--- a/experimental/Schemes/src/elliptic_surface.jl
+++ b/experimental/Schemes/src/elliptic_surface.jl
@@ -1213,7 +1213,7 @@ function _section_on_weierstrass_ambient_space(X::EllipticSurface, P::EllipticCu
   U = X0[1][1]
   (x,y,t) = coordinates(U)
   b = P
-  return ideal_sheaf(X0,U,[OO(U)(i) for i in [x*denominator(b[1])(t)-numerator(b[1])(t),y*denominator(b[2])(t)-numerator(b[2])(t)]])
+  return ideal_sheaf(X0,U,[OO(U)(i) for i in [x*denominator(b[1])(t)-numerator(b[1])(t),y*denominator(b[2])(t)-numerator(b[2])(t)]]; check=false)
 end
 
 function _section(X::EllipticSurface, P::EllipticCurvePoint)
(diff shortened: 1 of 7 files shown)


src/AlgebraicGeometry/Schemes/Sheaves/IdealSheaves.jl
```diff
@@ -345,7 +345,7 @@ end
 end
   
 @attr Bool function has_dimension_leq_zero(I::Ideal)
-  is_one(I) && return true
+  #is_one(I) && return true
```
fingolfin: I am sure you have good reasons for this change and I appreciate your efforts to improve things, thank you!

However, whenever I see code being commented out and still being committed, I wonder: why is that?

As a rule of thumb, it almost always seems to be down to one of the following two:
1. it is wrong or for some reason not needed anymore.
2. it is desirable on the long run (e.g. an optimization) but for some reason has to be disabled for the time being (e.g. some other code needs a fix)

In the first case, we are better off just deleting the code.

In the second case, I really think a comment should be added explaining what's going on. Because right now, *you* certainly have the answer to this in your mind. But anyone looking at this code in a year will not, and they'll be thankful for any included hints...

Thus it would be super nice if you could do one or the other, and likewise in line 357 when the same happens :-)

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
```diff
@@ -107,18 +107,22 @@ function inherit_decomposition_info!(
 
   for V in patches(orig_cov)
     # Collect the patches in `ref_cov` refining V
-    V_ref = [U for U in patches(ref_cov) if decomp_dict[U][1] === V]
-    # Collect the corresponding complement equations
-    comp_eqns = [decomp_dict[U][2] for U in V_ref]
+    V_ref = [(U, (UV, h)) for (U, (UV, h)) in decomp_dict if UV === V]
+    _compl(a) = sum(length(lifted_numerator(b)) + length(lifted_denominator(b)) for b in a[2][2]; init=0)
```
fingolfin: A minor and purely optional nitpick (feel free to ignore):

So `UV` is always `V`. So why not do this?
```suggestion
    V_ref = [(U, h) for (U, (UV, h)) in decomp_dict if UV === V]
    _compl(a) = sum(length(lifted_numerator(b)) + length(lifted_denominator(b)) for b in a[2]; init=0)
```
and then the `for` loop in line 115 can be slightly simplified, and also line 120.


This avoids a little bit of storage space for those tuples.

Anyway, perhaps you have a reason for having it as it is (e.g. plans for further changes, keeping the code similar to something else etc.) -- so as I said, feel free to ignore.

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
```diff
@@ -107,18 +107,22 @@ function inherit_decomposition_info!(
...
+    V_ref = sort!(V_ref, by=_compl)
     # Start out from the original decomposition info
     dec_inf = copy(decomposition_info(orig_cov)[V])
-    for (i, U) in enumerate(V_ref)
+    for (i, (U, (_, h))) in enumerate(V_ref)
```
fingolfin: With my suggested change above, this would be:
```suggestion
    for (i, (U, h)) in enumerate(V_ref)
```

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
```diff
@@ -107,18 +107,22 @@ function inherit_decomposition_info!(
...
       # Cast the already made decomposition info down to U
       tmp = elem_type(OO(U))[OX(V, U)(i) for i in dec_inf] # help the compiler
       # Append the equations for the previously covered patches
       for j in 1:i-1
-        W = V_ref[j]
-        surplus = OX(V, U).(decomp_dict[W][2])
+        (_, (_, hh)) = V_ref[j]
```
fingolfin:
```suggestion
        _, hh = V_ref[j]
```
or
```suggestion
        hh = V_ref[j][2]
```

src/AlgebraicGeometry/Schemes/Covering/Objects/Attributes.jl
HechtiDerLachs: This makes sense. I just cluttered it together so that it would work for the moment. Thanks for the close look.

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1649,6 +1649,19 @@ function extended_ade(ADE::Symbol, n::Int)
   return -G, kernel(G; side = :left)
 end
 
+# This function allows to store a reduction map to positive characteristic,
+# e.g. for computing intersection numbers.
+function reduction_to_pos_char(X::EllipticSurface, red_map::Map)
```
simonbrandhorst, line 1654: Should we turn this into a field of `EllipticSurface`?

src/AlgebraicGeometry/Schemes/Sheaves/CoherentSheaves.jl
```diff
@@ -784,19 +784,6 @@ end
   return C
 end
 
-function inherit_decomposition_info!(C::Covering, X::AbsCoveredScheme)
```
simonbrandhorst, line 787: why remove this? Was it a duplicate?

src/AlgebraicGeometry/Schemes/Sheaves/IdealSheaves.jl
```diff
@@ -345,7 +345,6 @@ end
 end
   
 @attr Bool function has_dimension_leq_zero(I::Ideal)
-  is_one(I) && return true
```
simonbrandhorst, line 348: Why remove this line?

simonbrandhorst (approved):

src/AlgebraicGeometry/Schemes/Sheaves/CoherentSheaves.jl
HechtiDerLachs: Yes. And it was wrong. Fortunately, it seemed to be used nowhere.

src/AlgebraicGeometry/Schemes/Sheaves/IdealSheaves.jl
HechtiDerLachs: It's superfluous. It's usually not easier to compute this than the dimension. Both are the same Groebner basis computation.

experimental/Schemes/src/elliptic_surface.jl
HechtiDerLachs: Let's discuss this in #4183 .

src/AlgebraicGeometry/Schemes/Sheaves/IdealSheaves.jl
simonbrandhorst: Yep. Is the GB stored? Or could it be that it is discarded and just the `is_one` attribute is true?

PR #4178, opened 2024-10-04
Avoid bottlenecks in variable matching.

HechtiDerLachs:
Unfortunately I already ran into the first encounter where my recent "speedup" actually led to a regression. This should fix it.


fingolfin (approved):
Looks sensible to me

PR #4182, opened 2024-10-07
[FTheoryTools] Fixed some bugs that were missed before

emikelsons:
Some small bug fixes that were missed in some recents pull requests relating to model indexes and zero section classes of FTheory models:

- model_index doesn't result in errors anymore(there might have been a defunct method that was used before)
- Model constructors now no longer attempt to set a zero_section_class for models constructed over an arbitrary base
- For Tate and Weierstrass models the zero section class is now always set to be 'z'


emikelsons:
@apturner @HereAround

HereAround:
Looks good to me. Unless you are still working on this @emikelsons , it can be merged.

Regarding the failed book-tests (cf. https://github.com/oscar-system/Oscar.jl/actions/runs/11269198738/job/31337353429), which prevent the merge at this point: I could not see exactly why they failed. The error message "The operation was cancelled" was not very enlightening to me. So I just restarted those tests.

One possibly very naive idea: The book tests failed after about 2.5 hours. Could it maybe be a time-out error?

emikelsons:
The error that was raised previously was not a time-out error, but that one doesn't seem to appear any longer

HereAround:
> The error that was raised previously was not a time-out error, but that one doesn't seem to appear any longer

It seems the error has disappeared. As such, I will merge this now.

HereAround (approved):
Thank you for working on this @emikelsons

PR #4183, opened 2024-10-07
Add tweaks for enabling a 2-neighbor-step in characteristic 0

HechtiDerLachs:
Building on #4177.

@simonbrandhorst : I managed to perform our two neighbor step in characteristic zero with this (modulo mistakes). 

It would therefore be great if we could decide on how to integrate this code properly. I will leave some comments below.


experimental/Schemes/src/BlowupMorphism.jl
```diff
@@ -644,13 +644,17 @@ function produce_object_on_affine_chart(I::StrictTransformIdealSheaf, U::AbsAffi
...
   #return saturation_with_index(tot, IE_loc)
   # It is usually better to pass to the simplified covering to do the computations
   simp_cov = simplified_covering(X)
   U_simp = first([V for V in patches(simp_cov) if original(V) === U])
   a, b = identification_maps(U_simp)
-  result, _ = saturation_with_index(pullback(a)(tot), pullback(a)(IE_loc))
+  # This used to be the following line. But we don't use the index, so we 
+  # switch to the more performant version
+  # result, _ = saturation_with_index(pullback(a)(tot), pullback(a)(IE_loc))
+  result = _iterative_saturation(pullback(a)(tot), elem_type(OO(U_simp))[pullback(a)(u) for (u, _) in factor(lifted_numerator(first(gens(IE_loc))))])
```
HechtiDerLachs: @afkafkafk13 : The use of  `saturation_with_index` here goes back to your original code for computing transforms. Do you see a reason why we should not switch to `_iterative_saturation` here? It is probably more performant in many cases.

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1213,13 +1213,24 @@ function _section_on_weierstrass_ambient_space(X::EllipticSurface, P::EllipticCu
...
+
+  if !is_zero(P)
+    WC = weierstrass_chart_on_minimal_model(X)
+    U = original_chart(PX)
+    @assert OO(U) === ambient_coordinate_ring(WC)
+    PY = PrimeIdealSheafFromChart(X, WC, ideal(OO(WC), PX(U)))
+    set_attribute!(PY, :name, string("section: (",P[1]," : ",P[2]," : ",P[3],")"))
+    set_attribute!(PY, :_self_intersection, -euler_characteristic(X))
+    return WeilDivisor(PY, check=false)
+  end
```
HechtiDerLachs: I switched to the creation of sections from points to come from the weierstrass chart directly. This has the advantage that the strict transforms do not have to be computed to realize this divisor on that chart. This computation was killing us in char. 0, so I hope this is ok.

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1314,7 +1325,11 @@ function _prop217(E::EllipticCurve, P::EllipticCurvePoint, k)
 
   # collect the equations as a matrix
   cc = [[coeff(j, abi) for abi in ab] for j in eqns]
-  M = matrix(B, length(eqns), length(ab), reduce(vcat,cc, init=elem_type(base)[]))
+  mat = reduce(vcat,cc, init=elem_type(base)[])
+  @assert all(is_one(denominator(x)) for x in mat)
+  @assert all(is_constant(numerator(x)) for x in mat)
+  mat2 = [constant_coefficient(numerator(x)) for x in mat]
+  M = matrix(B, length(eqns), length(ab), mat2)
```
HechtiDerLachs, line 1344: This was making trouble over number fields:
```
ERROR: MethodError: no method matching (::AbsSimpleNumField)(::AbstractAlgebra.Generic.FracFieldElem{AbstractAlgebra.Generic.Poly{AbsSimpleNumFieldElem}})

Closest candidates are:
  (::AbsSimpleNumField)(::Singular.n_FieldElem{AbsSimpleNumFieldElem})
   @ Singular ~/.julia/packages/Singular/lsYSW/src/MessyHacks.jl:119
  (::AbsSimpleNumField)(::Vector{QQFieldElem})
   @ Hecke ~/.julia/packages/Hecke/7wI0D/src/NumField/Elem.jl:284
  (::AbsSimpleNumField)(::ZZRingElem)
   @ Nemo ~/.julia/packages/Nemo/tzyHK/src/antic/nf_elem.jl:1095
  ...
```
I guess the implicit conversion via `B` was not very clean, anyway, so I tried to clear this up. @simonbrandhorst : Could you have a look whether this is now reasonable?

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1368,7 +1383,8 @@ function linear_system(X::EllipticSurface, P::EllipticCurvePoint, k::Int64)
 
     I = saturated_ideal(defining_ideal(U))
     IP = ideal([x*xd(t)-xn(t),y*yd(t)-yn(t)])
-    issubset(I, IP) || error("P does not define a point on the Weierstrasschart")
+    # Test is disabled for the moment; should eventually be put behind @check!
```
HechtiDerLachs: See the comment. This was another bottleneck killing us. So we should allow to switch it off.

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1649,6 +1664,28 @@ function extended_ade(ADE::Symbol, n::Int)
...
+function reduction_to_pos_char(X::EllipticSurface, red_map::Map)
+  return get_attribute!(X, :reduction_to_pos_char) do
+    kk0 = base_ring(X)
+    @assert domain(red_map) === kk0
+    kkp = codomain(red_map)
+    @assert characteristic(kkp) > 0
+    _, result = base_change(red_map, X)
+    return red_map, result
+  end::Tuple{<:Map, <:Map}
+end
```
HechtiDerLachs: This is what we need to discuss. For elliptic surfaces it is sometimes reasonable to do computations of intersection numbers in a reduction modulo primes (as I hear). We probably do not want to allow this in general but have to hide it a bit from the user so that they use it at their own risk when they know what they're doing. 

Instead of giving it as a second argument, I propose to have a field/attribute `good_reduction` which holds the reduction of the `base_ring`s to positive characteristic, if it is set. Then the rationale is that a model of `X` in positive characteristic is held in the background, together with a list of counterparts for the generators of the `algebraic_lattice`. The methods to do intersection theory have to be customized in order to allow for several assumptions which can not be made unless we deal with elliptic surfaces. This is carried out in the code below. Someone with more background knowledge in elliptic surfaces (@simonbrandhorst ?) should check whether the code is safe to use, decide which precautions we require the user to take and how we want to communicate them.

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1661,14 +1698,53 @@ function basis_representation(X::EllipticSurface, D::WeilDivisor)
   n = length(basis_ambient)
   v = zeros(ZZRingElem, n)
   @vprint :EllipticSurface 3 "computing basis representation of $D\n"
-  for i in 1:n
-    @vprintln :EllipticSurface 4 "intersecting with $(i): $(basis_ambient[i])"
-    v[i] = intersect(basis_ambient[i], D)
+  kk = base_ring(X)
+  if iszero(characteristic(kk)) && has_attribute(X, :reduction_to_pos_char)
+    red_map, bc = get_attribute(X, :reduction_to_pos_char)
+    red_dict = IdDict{AbsWeilDivisor, AbsWeilDivisor}(D=>_reduce_as_prime_divisor(bc, D) for D in basis_ambient)
```
HechtiDerLachs, line 1808:
```suggestion
    red_dict = reduction_of_algebraic_lattice(X)
```

experimental/Schemes/src/elliptic_surface.jl
```diff
@@ -1661,14 +1698,53 @@ function basis_representation(X::EllipticSurface, D::WeilDivisor)
...
+function _reduce_as_prime_divisor(bc::AbsCoveredSchemeMorphism, I::PrimeIdealSheafFromChart)
+  U = original_chart(I)
+  bc_cov = covering_morphism(bc)
+  V = __find_chart(U, codomain(bc_cov))
+  IV = I(V)
+  bc_loc = first(maps_with_given_codomain(bc_cov, V))
+  J = pullback(bc_loc)(IV)
+  set_attribute!(J, :is_prime=>true)
+  return PrimeIdealSheafFromChart(domain(bc), domain(bc_loc), J)
+end
```
HechtiDerLachs, line 1874: This method assumes that the prime ideal sheaf `I` stays prime when reduced via `bc`. This is not automatically the case, so we should probably warn the user to only use "sufficiently good reduction" here.

experimental/Schemes/src/BlowupMorphism.jl
afkafkafk13: You are perfectly right here. In the very first version of the code, I assumed a single blow up at an irreducible component of the center -- for which the saturation_with_index was fine (and allowed extra debugging information). In sequences of blow ups this should indeed be done iteratively. The overall structure has evolved very far over time.

experimental/Schemes/src/elliptic_surface.jl
simonbrandhorst: Previously, it was a `PrimeIdealSheafFromChart`. But I deliberately changed that to `StrictTransformIdealSheaf` because it was more performant (and in principle it allows the algorithms to take advantage of the blowup structure). 
Unless you can provide evidence that your change improves performance in a wide range of examples please leave it as is. Optionally you could include another dispatch with a type as first argument.

experimental/Schemes/src/BlowupMorphism.jl
afkafkafk13: Just one more thought: Is the exceptional divisor reliably stored as principal on the given chart? We do not have full control on this.
I would not put the condition in an `assert` up there, but rather use the old variant as a fallback for exceptional divisors which happen to be stored non-principal.

experimental/Schemes/src/elliptic_surface.jl
simonbrandhorst: One case where a `PrimeIdealSheafFromChart` will be faster is when we do only computations in the weierstrass chart e.g. for computing a linear system.

experimental/Schemes/src/BlowupMorphism.jl
afkafkafk13: Why do you factor explicitly here? You have the factorization already built into the iterative_saturation in IdealSheaves.jl and this would make the code here more readable.

HechtiDerLachs: Ok, sounds reasonable. Thx.

experimental/Schemes/src/elliptic_surface.jl
HechtiDerLachs: Exactly. The computations just didn't go through with the strict transform. Having a second argument as the return type should be OK, but we would still have to decide for a default.

experimental/Schemes/src/BlowupMorphism.jl
HechtiDerLachs: I factor before passing to the "simplified scheme". The structure of the defining equations is usually destroyed by elim_part.

afkafkafk13: I am aware of the effects of simplified_scheme.

I am refering to `_iterative_saturation(I::Ideal, f::RingElem)` starting in line 562 of IdealSheaves.jl, where this factorization is taken care of explicitly by calling factor and then `_iterative_saturation` with the signature with a vector as second argument.

HechtiDerLachs: I introduced a new signature for `_iterative_saturation` which has a list of elements as second argument. That one does not do a factorization again.

afkafkafk13: ok. that fits.

src/AlgebraicGeometry/Schemes/Sheaves/IdealSheaves.jl
```diff
@@ -345,6 +345,7 @@ end
 end
   
 @attr Bool function has_dimension_leq_zero(I::Ideal)
+  is_one(I) && return true
```
lgoettgens, line 348: This was just removed in https://github.com/oscar-system/Oscar.jl/pull/4177/files#diff-0fb0ac851cccf4c4419c89353ba64ef43364db228c65357487fe512e76917062L348. Maybe check if you re-added more things by accident

HechtiDerLachs: I know... It happened in the rebase. But then I thought: "Well, maybe @simonbrandhorst was right and there was a shortcut in some cases where the `:is_one` attribute was set?" So it also doesn't hurt to keep it.

afkafkafk13: Keep it. In the worst case, it computes and stores the GB properly implying that the dimension further below is just a lightweight combinatorial computation on the leading terms.

experimental/Schemes/src/elliptic_surface.jl
simonbrandhorst: Huh, how can this possibly kill us? Either `P` is extremely large, or something else is way more slow than it should be. 
In this particular case this is better in an `@hassert` and then there should be a free check that 
`parent(P) == generic_fiber(X)`

afkafkafk13: There could always be a bad choice of monomial ordering somewhere in the background -- or some ignored caching of groebner bases.

simonbrandhorst: alright ... it is just trivial over `k(t)[x,y]` but the groebner basis does not see this and works over `k[x,y,t]` ... then degrevlex is a terrible choice. O.k. lets just disable this test.

experimental/Schemes/src/elliptic_surface.jl
simonbrandhorst: Your change looks like it should work. 

Alternatively in line 1302 it could be 
`R,ab = polynomial_ring(Bt,vcat([Symbol(:a,i) for i in 0:dega],[Symbol(:b,i) for i in 0:degb]),cached=false)`
instead (`base` was replaced by `Bt`)
because we don't get any denominators anyway.

simonbrandhorst:
@HechtiDerLachs it seems that some of your merges went wrong. The diff is completely off.

src/AlgebraicGeometry/Schemes/Divisors/WeilDivisor.jl
```diff
@@ -338,7 +338,7 @@ function intersect(D::AbsWeilDivisor, E::AbsWeilDivisor;
         I = c1 + c2
         if !has_dimension_leq_zero(I) # potentially faster for localized ideals
           if c1 == c2
-            result = result + a1*a2*_self_intersection(c1)
+            result = result + a1*a2* (has_attribute(c1, :_self_intersection) ? _self_intersection(c1) : _self_intersection(c2))
```
simonbrandhorst, line 341: why not put this in the function `_self_intersection` itself?

simonbrandhorst (approved):

src/AlgebraicGeometry/Schemes/Divisors/WeilDivisor.jl
HechtiDerLachs: Because the function is called with one argument and does not know about the possibility that another object which is `==` to the given one might hold the information.

PR #4184, opened 2024-10-08
Add missing Abelian groups docs

paemurru:
Similarly to the update in Hecke:
https://github.com/thofma/Hecke.jl/pull/1631


joschmitt:
Did you check whether this works as expected? I don't expect any problems, but since the pull request is from your fork there is no preview (I think).

lgoettgens (changes requested):
image
the navigaton bar on the left is broken

lgoettgens:
@paemurru if you have then access rights to do so, could you please open this PR (with changes to the issue above) as a new PR from a branch in the oscar-system/Oscar.jl project (and not from a fork)? This enables everyone here to look at the preview automatically published at https://docs.oscar-system.org/previews/PR4184 and make their own opinion, rather than having to build the docs on their own locally.

lgoettgens:
closing in favor of https://github.com/oscar-system/Oscar.jl/pull/4186

PR #4185, opened 2024-10-08
two_neighbor_step without computing intersection numbers

simonbrandhorst:


simonbrandhorst:
Once #4189 is merged we should get a green test here also.

HechtiDerLachs (approved):

PR #4193, opened 2024-10-11
serialization of more group attributes

ThomasBreuer:
The idea is to implement serializations directly for the underlying GAP objects, instead of implementing them for the Oscar objects.

For boolean and numeric attributes, the two alternatives do not really differ, but for attributes whose values are themselves groups, for example `center(g)`, we can simply serialize `GAP.Globals.Center(GapObj(g))`, together with its attributes, instead of first creating `center(g)` anew and then serializing its first component.

Another advantage is that this way, it will be easier to deal with "Oscar attributes" of groups that do not rely on GAP's attribute mechanism.

(The serialization of permutation groups gets changed with this approach; if we proceed like this then a fallback must be provided, in order to deal with the deserialization of files containing permutation groups that were serialized with the old method.)

Currently there is a problem with cyclic references, and the handling of attribute values that are GAP lists is not yet supported.

Here is an example for the cyclic reference problem:
```
julia> g = small_group(4, 1);

julia> path = mktempdir();  filenamev = joinpath(path, "v");

julia> save(filenamev, g)

julia> Oscar.reset_global_serializer_state();

julia> loadedv = load(filenamev);

julia> has_center(g)
false

julia> center(g);

julia> save(filenamev, g)

julia> Oscar.reset_global_serializer_state()
Dict{Base.UUID, Any}()

julia> loadedv = load(filenamev)
Internal error: encountered unexpected error during compilation of showerror:
StackOverflowError()
```

### DIFF
--- a/src/Serialization/GAP.jl
+++ b/src/Serialization/GAP.jl
@@ -29,9 +29,60 @@ end
 
 #############################################################################
 #
+# attributes handling
+#
+const GapObj_attributes = [
+  # boolean
+  :IsAbelian, :IsAlmostSimpleGroup, :IsCyclic, :IsElementaryAbelian,
+  :IsFinite, :IsNilpotentGroup, :IsPerfectGroup, :IsQuasisimpleGroup,
+  :IsSimpleGroup, :IsSolvableGroup, :IsSporadicSimpleGroup,
+  :IsSupersolvableGroup,
+  # numeric Int
+  :DerivedLength, :NilpotencyClassOfGroup, :NrConjugacyClasses, :NrMovedPoints,
+  # numeric ZZRingElem
+  :Exponent,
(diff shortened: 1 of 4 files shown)


fingolfin:
What the status of this PR?

ThomasBreuer:
> What the status of this PR?

This was an experiment from the times before the big refactoring of serialization.
The starting point was the idea to provide a mechanism for serializing GAP objects together with their attributes (and possible cyclic references between them), and to use this when serializing Oscar objects that rely on these GAP objects.
Meanwhile it is clear that this approach has the disadvantage that much more information gets serialized than is necessary. (But note that the behaviour of a GAP object can change significantly if one removes its non-defining attributes.)

Thus I will close this pull request.

PR #4194, opened 2024-10-11
More type stability for GSets

lgoettgens:
...by adding a second type variable for the element type.

In particular, this parameterized `ElementOfGSet` and makes the field access inferrable.


ThomasBreuer (approved):
Looks good. Thanks.

Issue #4191, opened 2024-10-10
some `==` issues for group elements and related objects


ThomasBreuer:
The change proposed in Nemocas/AbstractAlgebra.jl/pull/1853 has raised some questions about perhaps missing `==` methods in Oscar's code for groups.

- We have `==` methods for two `BasisGAPGroupElem`s and for two `MatrixGroupElem`s. Other cases were handled up to now by a generic method that returns `false`, but with Nemocas/AbstractAlgebra.jl/pull/1853 we will have to deal with more cases. One addition can be a `==` method for two `GAPGroupElem`s that returns `false`.
- A more subtle question (which is interesting independent of Nemocas/AbstractAlgebra.jl/pull/1853) is about the relation between elements of automorphism groups and the corresponding group homomorphisms.
  - It is likely that users want to compare an automorphism group element and a group homomorphism via `==` --our tests contain already such cases. We want to return `false` (this happens now) or throw an exception (but then with a meaningful error message). Thus we need special methods for this situation.
  - It is documented that `h in A` shall return `false` for a group homomorphism `h` and a group `A` of automorphisms, also if there is an element of `A` that is equal to `h` as a map on `A`.
    If we want to keep this definition then what is a good way to decide if `A(h)` will throw an exception or return the corresponding element of `A`?
    (In the related situation `g in H` where `H` is a subgroup of the parent of `g`, we had decided that `g in H` should return `true` if `H(g)` would be successful, i.e., that `g in H` is the natural way to ask this question.)
  - A related question about elements of automorphism groups:
    They can be used as maps in the sense that they have a `domain`, and one can ask for images and preimages. But they aren't `Map`s (and they do not have a `codomain`) -- is there a good reason why?

fingolfin:
> One addition can be a `==` method for two `GAPGroupElem`s that returns `false`.

But why do we need that, where do we compare group elems like that? The one legit case I can think of are PcGroupElems vs. SubPcGroupElems (and similar for fp groups) but for those we could provide dedicated methods?

fieker:
On Thu, Oct 10, 2024 at 07:24:02AM -0700, Thomas Breuer wrote:
> The change proposed in Nemocas/AbstractAlgebra.jl/pull/1853 has raised some questions about perhaps missing `==` methods in Oscar's code for groups.
> 
[...]

PR #4196, opened 2024-10-11
allow `==` for two groups/group elements in fewer cases

ThomasBreuer:
If `_check_compatible` for the two groups in question returns `false` then throw an exception.
Note that the result is not based only on the types of the arguments but also on the groups.

This addresses one of the questions from #4191.

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -373,7 +373,7 @@ end
 Base.:*(x::GAPGroupElem, y::GAPGroupElem) = _prod(x, y)
 
 function ==(x::GAPGroup, y::GAPGroup)
-  _check_compatible(x, y; error = false) || throw(ArgumentError("x and y are not compatible"))
+  _check_compatible(x, y)
   return GapObj(x) == GapObj(y)
 end
 
@@ -382,7 +382,7 @@ end
 # in the sense of `_check_compatible`,
 # and compare the `GapObj`s if this is the case.
 function ==(x::BasicGAPGroupElem, y::BasicGAPGroupElem)
-  _check_compatible(parent(x), parent(y); error = false) || throw(ArgumentError("parents of x and y are not compatible"))
+  _check_compatible(parent(x), parent(y))
   return GapObj(x) == GapObj(y)
 end
 
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -372,9 +372,26 @@ end
 
 Base.:*(x::GAPGroupElem, y::GAPGroupElem) = _prod(x, y)
 
-==(x::GAPGroup, y::GAPGroup) = GapObj(x) == GapObj(y)
+function ==(x::GAPGroup, y::GAPGroup)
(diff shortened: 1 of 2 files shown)


src/Groups/GAPGroups.jl
```diff
@@ -372,9 +372,26 @@ end
 
 Base.:*(x::GAPGroupElem, y::GAPGroupElem) = _prod(x, y)
 
-==(x::GAPGroup, y::GAPGroup) = GapObj(x) == GapObj(y)
+function ==(x::GAPGroup, y::GAPGroup)
+  _check_compatible(x, y; error = false) || throw(ArgumentError("x and y are not compatible"))
```
```suggestion
  _check_compatible(x, y; error = true)
```
wouldn't this be easier? (similar below)
I don't see a reason to first suppress an error and then throw basically the same one right afterwards

src/Groups/GAPGroups.jl
```diff
@@ -372,9 +372,26 @@ end
...
+  return GapObj(x) == GapObj(y)
+end
 
-==(x::BasicGAPGroupElem, y::BasicGAPGroupElem ) = GapObj(x) == GapObj(y)
+# For two `GAPGroupElem`s,
+# if no specialized method is applicable then no `==` comparison is allowed.
+function ==(x::GAPGroupElem, y::GAPGroupElem)
+  _check_compatible(parent(x), parent(y); error = false) || throw(ArgumentError("parents of x and y are not compatible"))
+  throw(ArgumentError("== is not implemented for the given types"))
+end
```
lgoettgens, line 394: once https://github.com/Nemocas/AbstractAlgebra.jl/pull/1853 is available, this method can be removed. maybe add this as a comment

src/Groups/GAPGroups.jl
ThomasBreuer: Yes, we do not get a better error message in this case.

src/Groups/GAPGroups.jl
ThomasBreuer: My point of view is:
If the compatibility check returns `false` then we are sure that the question for `==` is not allowed,
but if the compatibility check returns `true` then likely there is a new subtype of `GAPGroupElem` for which a `==` method should be provided.
(Currently we have just the subtypes `BasicGAPGroupElem` and `MatrixGroupElem`.)

src/Groups/GAPGroups.jl
```diff
@@ -372,9 +372,26 @@ end
...
+function ==(x::BasicGAPGroupElem, y::BasicGAPGroupElem)
+  _check_compatible(parent(x), parent(y))
+  return GapObj(x) == GapObj(y)
+end
 
-==(x::BasicGAPGroupElem, y::BasicGAPGroupElem ) = GapObj(x) == GapObj(y)
+# For two `GAPGroupElem`s,
+# if no specialized method is applicable then no `==` comparison is allowed.
+function ==(x::GAPGroupElem, y::GAPGroupElem)
+  _check_compatible(parent(x), parent(y); error = false) || throw(ArgumentError("parents of x and y are not compatible"))
```
lgoettgens, line 392: Same here as above

fingolfin (approved):

PR #4197, opened 2024-10-12
Improve stability of tests in PolyhedralGeometry

alexej-jordan:
Prevent tests from randomly failing if the order of some properties changes or rays are represented differently.

### DIFF
--- a/src/PolyhedralGeometry/iterators.jl
+++ b/src/PolyhedralGeometry/iterators.jl
@@ -104,8 +104,10 @@ end
 
 Base.:(==)(x::AbstractVector, y::RayVector) = y == x
 
-Base.:(==)(::PointVector, ::RayVector) = throw(ArgumentError("Cannot compare PointVector to RayVector"))
-Base.:(==)(::RayVector, ::PointVector) = throw(ArgumentError("Cannot compare PointVector to RayVector"))
+Base.:(==)(::PointVector, ::RayVector) =
+  throw(ArgumentError("Cannot compare PointVector to RayVector"))
+Base.:(==)(::RayVector, ::PointVector) =
+  throw(ArgumentError("Cannot compare PointVector to RayVector"))
 
 ################################################################################
 ######## Halfspaces and Hyperplanes
--- a/src/PolyhedralGeometry/iterators.jl
+++ b/src/PolyhedralGeometry/iterators.jl
@@ -90,12 +90,8 @@ function Base.:(==)(x::RayVector, y::RayVector)
   iy = findfirst(!is_zero, y)
   ix == iy || return false
   isnothing(ix) && return true
-  r = y[iy]//x[ix]
-  is_negative(r) && return false
-  for i in 1:length(x)
-    r * x[i] == y[i] || return false
-  end
-  return true
+  sign(x[ix]) == sign(y[iy]) || return false
+  return y[iy] * x.p == x[ix] * y.p
 end
 
 function Base.:(==)(x::RayVector, y::AbstractVector)
--- a/src/PolyhedralGeometry/iterators.jl
+++ b/src/PolyhedralGeometry/iterators.jl
@@ -85,6 +85,31 @@ Return a `RayVector` resembling a ray from the origin through the point whose co
 """
 ray_vector
 
+function Base.:(==)(x::RayVector, y::RayVector)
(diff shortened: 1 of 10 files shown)


lgoettgens:
This unfortunately has a conflict

src/PolyhedralGeometry/iterators.jl
```diff
@@ -85,6 +85,30 @@ Return a `RayVector` resembling a ray from the origin through the point whose co
...
+function Base.:(==)(x::RayVector, y::RayVector)
+  ax = collect(x)
+  ay = collect(y)
+  ix = findfirst(!is_zero, ax)
+  iy = findfirst(!is_zero, ay)
+  ix == iy || return false
+  isnothing(ix) && return true
+  r = ay[iy]//ax[ix]
+  r < zero(r) && return false
+  return (r .* ax == ay)
```
benlorenz:
```suggestion
  ix = findfirst(!is_zero, x)
  iy = findfirst(!is_zero, y)
  ix == iy || return false
  isnothing(ix) && return true
  r = y[iy]//x[ix]
  is_negative(r) && return false
  return r * x.p == y.p
```
This should be a bit faster and avoid some allocations.

test/PolyhedralGeometry/cone.jl
```diff
@@ -28,43 +28,35 @@
...
     @test n_rays(Cone1) == 2
     @test rays(RayVector{T}, Cone1) isa SubObjectIterator{RayVector{T}}
     @test rays(Cone1) isa SubObjectIterator{RayVector{T}}
     @test rays(RayVector, Cone1) isa SubObjectIterator{RayVector{T}}
-    @test vector_matrix(rays(Cone1)) == matrix(f, [1 0; 0 1])
+    @test issetequal(rays(Cone1), ray_vector.(Ref(f), [[1, 0], [0, 1]]))
+    @test vector_matrix(rays(Cone1)) == _oscar_matrix_from_property(f, rays(Cone1))
     if T == QQFieldElem
-      @test matrix(QQ, rays(Cone1)) == matrix(QQ, [1 0; 0 1])
-      @test matrix(ZZ, rays(Cone6)) == matrix(ZZ, [2 3; 2 5])
```
benlorenz, line 41: It seems that the test for `Cone6` was lost?

test/PolyhedralGeometry/subdivision_of_points.jl
```diff
@@ -14,9 +14,9 @@
     SubdivisionOfPoints
 
   @testset "alternative inputs" begin
-    @test issetequal(collect(maximal_cells(square_by_incidence)),
-      collect(maximal_cells(square_by_weights)))
-    @test min_weights(square_by_cells) == min_weights(square_by_weights)
+    @test issetequal(maximal_cells(square_by_incidence),
+                     maximal_cells(square_by_weights))
+    @test issetequal(min_weights(square_by_cells), min_weights(square_by_weights))
```
benlorenz:
```suggestion
    @test issetequal(square_max_cells, maximal_cells(subdivision_of_points(C,min_weights(square_by_cells))))
```
We should either remove this test or compare via maximal cells. But comparing the weights is still fragile.

src/PolyhedralGeometry/iterators.jl
```diff
@@ -85,6 +85,31 @@ Return a `RayVector` resembling a ray from the origin through the point whose co
...
+function Base.:(==)(x::RayVector, y::RayVector)
+  ix = findfirst(!is_zero, x)
+  iy = findfirst(!is_zero, y)
+  ix == iy || return false
+  isnothing(ix) && return true
+  r = y[iy]//x[ix]
+  is_negative(r) && return false
+  for i in 1:length(x)
+    r * x[i] == y[i] || return false
+  end
```
benlorenz: If we do it like this we need to make sure the dimensions match, i.e. currently extra elements at the end of `y` would be ignored, for `x` it would throw an error.

And whether this returns false or throw an error should probably be coordinated with https://github.com/Nemocas/AbstractAlgebra.jl/pull/1854 or we keep comparing them as vectors for now (possibly using `x.p` and `y.p` directly).

benlorenz (approved):

lkastner (approved):

PR #4108, opened 2024-09-18
Mjrodgers ow g modules

fieker:

PR #4202, opened 2024-10-14
Add `letters` function for `PcGroupElem`

jamesnohilly:
As a starting point for this PR, I've borrowed the `syllable` functions for PcGroup from PR #4108 to use as a reference for planned work.

Planned work for this PR:
- [x] New function `letters` for (Sub)PcGroupElem.
- [x] A check in `_exponent_vector` for whether the input is really "canonical".
- [x] Test cases in `test/Groups/pcgroups.jl` for the new functions.
- [x] Add doc strings for new functions.

Suggestions from @fingolfin


fingolfin:
Thanks @jamesnohilly to clarify: what you posted in the description is not (yet) what you did, but a description of the work programme, right? Also, should mention that the code here was taken from PR #4108 by @fieker and is only the starting point.

jamesnohilly:
Yes, that's correct! I've now updated the description to make it clearer and also included a note that the code originates from PR #4108 as the starting point.

src/Groups/pcgroup.jl
```diff
@@ -365,3 +365,57 @@ function pc_group(c::GAP_Collector)
...
+ 2
+ 2
+```
+"""
+function letters(g::PcGroupElem)
+  w = GAPWrap.UnderlyingElement(GapObj(g))
+  return Vector{Int}(GAPWrap.LetterRepAssocWord(w))
+end 
+
+function Oscar.syllables(g::Union{PcGroupElem, SubPcGroupElem})
```
lgoettgens:
```suggestion
function syllables(g::Union{PcGroupElem, SubPcGroupElem})
```
qualification not needed

src/Groups/pcgroup.jl
```diff
@@ -365,3 +365,57 @@ function pc_group(c::GAP_Collector)
...
+  l = GAPWrap.ExtRepOfObj(GapObj(g))
+  @assert iseven(length(l))
+  return Pair{Int, ZZRingElem}[l[i-1] => l[i] for i = 2:2:length(l)]
+end
+
+# Convert syllables in canonical form into exponent vector
+#Thomas
+function exponent_vector(sylls::Vector{Pair{Int64, ZZRingElem}}, n)
```
lgoettgens: I don't like that this adds a method with basic data types to an exported function that has a higher-level meaning for data structures like polynomials.
```julia
julia> methods(exponent_vector)
# 12 methods for generic function "exponent_vector" from AbstractAlgebra.Generic:
  [1] exponent_vector(::Type{ZZRingElem}, a::Union{FpMPolyRingElem, QQMPolyRingElem, ZZMPolyRingElem, fpMPolyRingElem, fqPolyRepMPolyRingElem, zzModMPolyRingElem}, i::Int64)
     @ Nemo ~/.julia/packages/Nemo/0X3AB/src/flint/mpoly.jl:28
  [2] exponent_vector(a::FqMPolyRingElem, i::Int64)
     @ Nemo ~/.julia/packages/Nemo/0X3AB/src/flint/fq_default_mpoly.jl:486
  [3] exponent_vector(::Type{T}, a::Union{FpMPolyRingElem, QQMPolyRingElem, ZZMPolyRingElem, fpMPolyRingElem, fqPolyRepMPolyRingElem, zzModMPolyRingElem}, i::Int64) where T<:Union{Int64, UInt64}
     @ Nemo ~/.julia/packages/Nemo/0X3AB/src/flint/mpoly.jl:18
  [4] exponent_vector(p::AbstractAlgebra.Generic.UnivPoly, i::Int64)
     @ ~/.julia/packages/AbstractAlgebra/DRky3/src/generic/UnivPoly.jl:59
  [5] exponent_vector(a::MPolyDecRingElem, i::Int64)
     @ Oscar ~/code/julia/Oscar.jl/src/Rings/mpoly-graded.jl:739
  [6] exponent_vector(a::MPolyDecRingElem, i::Int64, ::Type{T}) where T
     @ Oscar ~/code/julia/Oscar.jl/src/Rings/mpoly-graded.jl:740
  [7] exponent_vector(a::AbstractAlgebra.Generic.MPoly{T}, i::Int64, ::Val{:lex}) where T<:RingElement
     @ ~/.julia/packages/AbstractAlgebra/DRky3/src/generic/MPoly.jl:120
  [8] exponent_vector(a::AbstractAlgebra.Generic.MPoly{T}, i::Int64, ::Val{:deglex}) where T<:RingElement
     @ ~/.julia/packages/AbstractAlgebra/DRky3/src/generic/MPoly.jl:130
  [9] exponent_vector(a::AbstractAlgebra.Generic.MPoly{T}, i::Int64, ::Val{:degrevlex}) where T<:RingElement
     @ ~/.julia/packages/AbstractAlgebra/DRky3/src/generic/MPoly.jl:140
 [10] exponent_vector(a::AbstractAlgebra.Generic.MPoly{T}, i::Int64) where T<:RingElement
     @ ~/.julia/packages/AbstractAlgebra/DRky3/src/generic/MPoly.jl:158
 [11] exponent_vector(a::AbstractAlgebra.Generic.MPoly{T}, i::Int64, ::Type{Val{ord}}) where {T<:RingElement, ord}
     @ AbstractAlgebra deprecated.jl:103
 [12] exponent_vector(a::Union{FpMPolyRingElem, QQMPolyRingElem, ZZMPolyRingElem, fpMPolyRingElem, fqPolyRepMPolyRingElem, zzModMPolyRingElem}, i::Int64)
     @ Nemo ~/.julia/packages/Nemo/0X3AB/src/flint/mpoly.jl:38
```
maybe move it to an unexported helper function under a different name? (like e.g. `_exponent_vector`)

src/Groups/pcgroup.jl
```diff
@@ -365,3 +365,57 @@ function pc_group(c::GAP_Collector)
...
+end
+
+# Convert syllables in canonical form into group element
+#Thomas
+function (G::PcGroup)(sylls::Vector{Pair{Int64, ZZRingElem}})
+  e = exponent_vector(sylls, ngens(G))
+  pcgs = Oscar.GAPWrap.FamilyPcgs(GapObj(G))
+  x = Oscar.GAPWrap.PcElementByExponentsNC(pcgs, GapObj(e, true))
+  return Oscar.group_element(G, x)
+end
```
lgoettgens:
```suggestion
end

```

src/Groups/pcgroup.jl
```diff
@@ -365,3 +365,57 @@ function pc_group(c::GAP_Collector)
...
+function exponent_vector(sylls::Vector{Pair{Int64, ZZRingElem}}, n)
+  res = zeros(ZZRingElem, n)
+  for pair in sylls
+    @assert res[pair.first] == 0 #just to make sure 
+    res[pair.first] = pair.second
+  end
+  return res
+end
+
+# Convert syllables in canonical form into group element
```
lgoettgens, line 605: why does it only work for syllables in canonical form? I would at least expect the function to test this requirement.
And I think this requirement makes it impossible to construct certain elements in certain groups. Consider `< a, b | a^2 = 1, b^2 = 1, aba = bab >` (Weyl group of type A2). The element `aba` has no representation in "syllables in canonical form" (or maybe I misunderstood what canonical form means here)

lgoettgens (commented):
some comments. Maybe more directed towards @fingolfin and @ThomasBreuer instead of @jamesnohilly

src/Groups/pcgroup.jl
jamesnohilly: Removed unneeded qualification in latest commit

src/Groups/pcgroup.jl
jamesnohilly: Added the newline in the latest commit

src/Groups/pcgroup.jl
ThomasBreuer: The idea is to manipulate low level data that describe group elements, in order to create some other group elements. I think this is not intended as a user function, thus it need not be exported.

lgoettgens: Thanks for the reply @ThomasBreuer. The problem is that the same name as some existing exported function automatically exports this as well. @jamesnohilly can you take care of renaming this function to e.g. `_exponent_vector` or something similar?

src/Groups/pcgroup.jl
ThomasBreuer: The point is that there are different useful ways to encode a group element that is given by some kind of word. For general elements in finitely presented groups, there is no normal form, and an element is represented by a sequence of letters or syllables.
For elements in polycyclic groups, there is a normal form which corresponds to an exponent vector, and many low level functions deal just with these exponent vectors. In situations where one needs a word-like description of an element, it may be necessary to switch from the exponent vector to a syllable description and back. Often it is not needed to do actual computations (such as normalizing the data) in such a case. Testing whether the data are admissible can make sense; in this case, there should be a `check = false` option that allows one to skip these tests.

lgoettgens: Thanks for the explanation. I might have mixed up FP and PcGroups when thinking about this.

src/Groups/pcgroup.jl
jamesnohilly: No problem! I've renamed the function and its usages to `_exponent_vector` as of the latest commit.

test/Groups/pcgroup.jl
```diff
@@ -82,3 +82,24 @@ end
...
+  @test letters(gg[1]^-5*gg[2]^-7) == [1, 2, 2] # all negative exp
+  @test letters(gg[1]^2*gg[2]^3) == [] # both identity elements
+
+  # finite polycyclic subgroup
+  gg = pc_group(symmetric_group(4))
+  G = derived_subgroup(gg)[1]
+  @test letters(G[1]^2) == [2, 2]
+  @test letters(G[1]^2*G[2]^3*G[3]^3) == [2, 2, 3, 4]
+  @test letters(G[1]^-2*G[2]^-3*G[3]^-3) == [2, 3, 4]
+end
```
fingolfin:
```suggestion
end

```

fingolfin: @jamesnohilly you can apply this suggestion via the web interface by using the "Commit suggestions" button. Though you then should also "fetch" the changes from git into your local branch.

fingolfin:
@jamesnohilly tests are still failing. are on onto that, or do you need assistance?

test/Groups/pcgroup.jl
jamesnohilly: Ah, I never noticed this, thanks! I've applied this suggestion in the latest commit.

src/Groups/pcgroup.jl
jamesnohilly: I've added this check (enabled by default) to verify if the syllables are in the required canonical form in the latest commit.

test/Groups/pcgroup.jl
```diff
@@ -82,3 +82,48 @@ end
...
+
+  sylls = [1 => ZZ(1), 2 => ZZ(2), 1 => ZZ(3)]
+  @test_throws ArgumentError gg(sylls) # repeating generators
+
+  sylls = [2 => ZZ(1), 1 => ZZ(2)]
+  @test_throws ArgumentError gg(sylls) # not in ascending order
+
+  sylls = [2 => ZZ(1), 1 => ZZ(2), 1 => ZZ(3)] # both conditions
+  @test_throws ArgumentError gg(sylls)
+end
```
lgoettgens:
```suggestion
end

```
the final newline is missing again

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,65 @@ function pc_group(c::GAP_Collector)
...
+#Thomas
+function (G::PcGroup)(sylls::Vector{Pair{Int64, ZZRingElem}}, check::Bool=true)
+  # check if the syllables are in canonical form
+  if check
+    indices = map(p -> p.first, sylls)
+    unq_indices = unique(indices) # maintains order
+    @req length(indices) == length(unq_indices) "given syllables have repeating generators"
+    @req issorted(unq_indices) "given syllables must be in ascending order"
```
lgoettgens:
```suggestion
    @req allunique(indices) "given syllables have repeating generators"
    @req issorted(indices) "given syllables must be in ascending order"
```
this should be semantically the same but a but more readable

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,65 @@ function pc_group(c::GAP_Collector)
...
+  for pair in sylls
+    @assert res[pair.first] == 0 #just to make sure 
+    res[pair.first] = pair.second
+  end
+  return res
+end
+
+# Convert syllables in canonical form into group element
+#Thomas
+function (G::PcGroup)(sylls::Vector{Pair{Int64, ZZRingElem}}, check::Bool=true)
```
lgoettgens:
```suggestion
function (G::PcGroup)(sylls::Vector{Pair{Int64, ZZRingElem}}; check::Bool=true)
```
we usually have `check` as a keyword argument

test/Groups/pcgroup.jl
jamesnohilly: Added the newline back, thanks!

src/Groups/pcgroup.jl
jamesnohilly: Thanks, this is much clearer and concise!

src/Groups/pcgroup.jl
jamesnohilly: Good to know, thanks!

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+julia> c = collector(2, Int);
+
+julia> Oscar.set_relative_orders!(c, [2, 3])
+
+julia> Oscar.set_conjugate!(c, 2, 1, [2 => 2])
+
+julia> gg = pc_group(c)
+Pc group of order 6
+
+julia> letters(gg[1]^5*gg[2]^-4)
```
fingolfin: This example is not quite obvious for a reader. Better to show the intermediate result, that should make the connection to the output of `letters` "obvious" :-)
```suggestion
julia> x = gg[1]^5*gg[2]^-4
f1*f2^2

julia> letters(x)
```

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+julia> c = collector(2, Int);
+
+julia> Oscar.set_relative_orders!(c, [2, 3])
+
+julia> Oscar.set_conjugate!(c, 2, 1, [2 => 2])
+
+julia> gg = pc_group(c)
+Pc group of order 6
+
+julia> syllables(gg[1]^5*gg[2]^-4)
```
fingolfin:
```suggestion
julia> x = gg[1]^5*gg[2]^-4
f1*f2^2

julia> syllables(x)
```

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+
+# Examples
+```jldoctest
+julia> c = collector(2, Int);
+
+julia> Oscar.set_relative_orders!(c, [2, 3])
+
+julia> Oscar.set_conjugate!(c, 2, 1, [2 => 2])
+
+julia> gg = pc_group(c)
```
fingolfin: The interface for construction `collectors` is quite advanced and currently not even fully documented, so using it for an example is maybe not optimal. I suggest to use the small groups library to get a suitable example group instead, i.e. here we can just do this:
```suggestion
julia> gg = small_group(6,1)
```
and then I think the rest of the example stays unchanged.

Same remark applies below.

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+  res = zeros(ZZRingElem, n)
+  for pair in sylls
+    @assert res[pair.first] == 0 #just to make sure 
+    res[pair.first] = pair.second
+  end
+  return res
+end
+
+# Convert syllables in canonical form into group element
+#Thomas
```
fingolfin:
```suggestion
```

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+function letters(g::Union{PcGroupElem, SubPcGroupElem})
+  w = GAPWrap.UnderlyingElement(GapObj(g))
+  return Vector{Int}(GAPWrap.LetterRepAssocWord(w))
+end
+
+"""
+    syllables(g::Union{PcGroupElem, SubPcGroupElem})
+
+Return the syllables of `g` as a list of pairs of integers, each entry corresponding to
+a group generator and its exponent.
```
fingolfin, line 541: Maybe also mention and/or demonstrate in the example how to go back? My understanding is that `gg(syllables(x)) == x`, correct?

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+  for pair in sylls
+    @assert res[pair.first] == 0 #just to make sure 
+    res[pair.first] = pair.second
+  end
+  return res
+end
+
+# Convert syllables in canonical form into group element
+#Thomas
+function (G::PcGroup)(sylls::Vector{Pair{Int64, ZZRingElem}}; check::Bool=true)
```
fingolfin, line 606: Can you also add a similar constructor which takes an exponent vector, i.e., an inverse to `letters`?

lgoettgens: One needs to watch out for the semantic difference to the already existing function for FPGroups in https://github.com/oscar-system/Oscar.jl/blob/24711eeecbe977de2bd7ba4a0bba8d7decc86464/src/Groups/GAPGroups.jl#L2348, which expects a flattened list of syllable pairs instead.

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
   end
 end
 
+"""
+    letters(g::Union{PcGroupElem, SubPcGroupElem})
+
+Return the letters of `g` as a list of integers, each entry corresponding to
+a group generator.
```
fingolfin, line 479: Note that we can also produce negative numbers: e.g. -3 means "inverse of 3rd generator". This should be explained, and perhaps an example added showing that. E.g. based on this:

```
julia> x = (gg[1]*gg[2]*gg[3])^-2
g1*g2^-2*g3^3
```

Perhaps also add something like this (and then mirror it in the other function)
> ```See also [`syllables`](@ref).```

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,87 @@ function pc_group(c::GAP_Collector)
...
+```
+"""
+function syllables(g::Union{PcGroupElem, SubPcGroupElem})
+  l = GAPWrap.ExtRepOfObj(GapObj(g))
+  @assert iseven(length(l))
+  return Pair{Int, ZZRingElem}[l[i-1] => l[i] for i = 2:2:length(l)]
+end
+
+# Convert syllables in canonical form into exponent vector
+#Thomas
```
fingolfin:
```suggestion
```

src/Groups/pcgroup.jl
fingolfin: ugh, OK. perhaps we should kill that (is it documented?) first then. In GAP it made some sense to use such a flat list to avoid memory, as there are no tuples in GAP, only lists. But in Julia there is no real benefit of this over a `Vector{Pair{Int64, ZZRingElem}}`.


But that is way beyond this PR. So let's leave out the constructor I mentioned.

lgoettgens: No, not documented, but used for serialization. I haven't looked into how it is used there, so maybe we can just adapt the deserialization function, in the worst case it needs an upgrade script.

lgoettgens: > ugh, OK. perhaps we should kill that (is it documented?) first then. 

maybe @ThomasBreuer can look into that?

src/Groups/pcgroup.jl
```diff
@@ -365,3 +364,106 @@ function pc_group(c::GAP_Collector)
...
+2-element Vector{Pair{Int64, ZZRingElem}}:
+ 1 => 1
+ 2 => 2
+
+julia> gg(s)
+f1*f2^2
+
+julia> gg(s) == x
+true
+"""
```
lgoettgens, line 582: you are missing the three backticks here to end the doctest

jamesnohilly: good catch! I have fixed it now, thanks!

src/Groups/pcgroup.jl
jamesnohilly: I have added a small example with some brief explanation to `letters` for this. However I am unsure if the example is good as I was not able to get elements with negative exponents and test.

src/Groups/pcgroup.jl
jamesnohilly: Yes that's correct! I have added a small demonstration at the bottom of the example in the recent commit.

src/Groups/pcgroup.jl
jamesnohilly: Perfect, no worries! I have now updated the examples with this replacement.

src/Groups/pcgroup.jl
jamesnohilly: Good idea! I have updated the example in both `syllables` and `letters` to make this output clearer.

lgoettgens:
This PR now unfortunately has conflicts with https://github.com/oscar-system/Oscar.jl/pull/4157.

fingolfin:
@jamesnohilly please try to resolve the conflicts. If you run into troubles (this is something that causes pain to many people esp. at the beginning) drop by my office and we'll figure it out together (ideally with your laptop if you have one)

jamesnohilly:
I've now merged the changes from #4157 with the changes made in this PR. From a quick look it seems to have merged without any major issues.

src/Groups/pcgroup.jl
```diff
@@ -473,6 +472,110 @@ function pc_group(c::GAP_Collector)
...
+f1^-3
+
+julia> letters(x)
+3-element Vector{Int64}:
+ -1
+ -1
+ -1
+```
+
+# Examples
```
lgoettgens: This heading should be above the first code block, not between two code blocks

src/Groups/pcgroup.jl
```diff
@@ -473,6 +472,110 @@ function pc_group(c::GAP_Collector)
...
+
+Return the letters of `g` as a list of integers, each entry corresponding to
+a group generator.
+
+This method can produce letters represented by negative numbers. A negative number 
+indicates the inverse of the generator at the corresponding positive index.
+
+For example, as shown below, an output of -1 refers to the "inverse of the first generator".
+
+See also [`syllables`](@ref).
```
lgoettgens:
```suggestion
See also [`syllables(::Union{PcGroupElem, SubPcGroupElem})`](@ref).
```
otherwise this may link to any function called `syllables`, e.g. to `syllables(::FPGroupElem)` which is not what we want here

src/Groups/pcgroup.jl
```diff
@@ -473,6 +472,110 @@ function pc_group(c::GAP_Collector)
...
+"""
+    letters(g::Union{PcGroupElem, SubPcGroupElem})
+
+Return the letters of `g` as a list of integers, each entry corresponding to
+a group generator.
+
+This method can produce letters represented by negative numbers. A negative number 
+indicates the inverse of the generator at the corresponding positive index.
+
+For example, as shown below, an output of -1 refers to the "inverse of the first generator".
```
lgoettgens:
```suggestion
For example, as shown below, an output of ´-1´ refers to the "inverse of the first generator".
```

src/Groups/pcgroup.jl
fingolfin: For that you need an infinite group. E.g.
```
julia> g = dihedral_group(PosInf())
Pc group of infinite order

julia> g[1]^-3 * g[2]^-3
g1*g2^-3
```
or
```
julia> g = abelian_group(PcGroup, [5, 0])
Pc group of infinite order

julia> g[1]^-3 * g[2]^-3
g1^2*g2^-3
```

src/Groups/pcgroup.jl
```diff
@@ -473,6 +472,110 @@ function pc_group(c::GAP_Collector)
...
+
+julia> gg(s)
+f1*f2^2
+
+julia> gg(s) == x
+true
+```
+"""
+function syllables(g::Union{PcGroupElem, SubPcGroupElem})
+  l = GAPWrap.ExtRepOfObj(GapObj(g))
```
fingolfin: For infinite groups, GAp uses a completely different internal representation. You'll need something like this:
```julia
  if GAP.Globals.IsPcpElement(GapObj(g))
     expvec = GAP.Globals.Exponents(GapObj(g))
     ... do something with it....
  else
    ... current code
  end
```

src/Groups/pcgroup.jl
```diff
@@ -473,6 +472,110 @@ function pc_group(c::GAP_Collector)
...
+  # check if the syllables are in canonical form
+  if check
+    indices = map(p -> p.first, sylls)
+    @req allunique(indices) "given syllables have repeating generators"
+    @req issorted(indices) "given syllables must be in ascending order"
+  end
+
+  e = _exponent_vector(sylls, ngens(G))
+  pcgs = Oscar.GAPWrap.FamilyPcgs(GapObj(G))
+  x = Oscar.GAPWrap.PcElementByExponentsNC(pcgs, GapObj(e, true))
```
fingolfin: Here you also need to check for `if GAPWrap.IsPcGroup(GapObj(G))` and then do something different.

You'll need the function `PcpElementByExponents` or `PcpElementByExponentsNC`, see https://gap-packages.github.io/polycyclic/doc/chap4.html#X7882F0F57ABEB680

Also `GAP.Globals.Collector(GapObj(G))`

fingolfin (approved):
Looks good to me now!

Issue #4203, opened 2024-10-15
Wrap more cases for syzygies from Singular


thofma:
From a user on slack, I am just the messenger. 
```
julia> R, (x,y) = polynomial_ring(GF(2), ["x","y"]);

julia> A, _ = quo(R, ideal(R, [x^2+1, y^2+1]));

julia> F = free_module(A,2);

julia> M,_ = sub(F, [F([A(x+1),A(y+1)])]);

julia> FF = free_module(A, ngens(M)); phi = hom(FF, M, gens(M)); K, = kernel(phi); gens(K)
1-element Vector{SubquoModuleElem{MPolyQuoRingElem{FqMPolyRingElem}}}:
 (x*y + x + y + 1)*e[1]

julia> standard_basis(K)
ERROR: wrong rings
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] *(M::MonomialOrdering{FqMPolyRing}, N::ModuleOrdering{FreeMod{MPolyQuoRingElem{FqMPolyRingElem}}})
   @ Oscar.Orderings ~/.julia/packages/Oscar/38HwK/src/Rings/orderings.jl:1915
 [3] default_ordering(F::FreeMod{MPolyQuoRingElem{FqMPolyRingElem}})
   @ Oscar ~/.julia/packages/Oscar/38HwK/src/Modules/UngradedModules/Methods.jl:235
 [4] default_ordering(M::SubquoModule{MPolyQuoRingElem{FqMPolyRingElem}})
   @ Oscar ~/.julia/packages/Oscar/38HwK/src/Modules/UngradedModules/SubquoModule.jl:737
 [5] standard_basis(M::SubquoModule{MPolyQuoRingElem{FqMPolyRingElem}})
   @ Oscar ~/.julia/packages/Oscar/38HwK/src/Modules/UngradedModules/SubquoModule.jl:757
 [6] top-level scope
   @ REPL[8]:1
```
Maybe @wdecker @HechtiDerLachs have an idea or know the person that should have the idea?

Edit: This is with Oscar master version 534513fb17.

HechtiDerLachs:
It seems that the user wants to use `standard_basis` over quotient rings. I am not even sure, this exists. Recently @RafaelDavidMohr carried over some resolutions over quotient rings and I made some first steps to extend the groebner basis functionality for modules over quotient rings in Oscar. However, the latter only became a stub and was not polished to an honest PR. 

[...]

PR #4204, opened 2024-10-15
Implement `syzygy_generators` for lists of module elements

HechtiDerLachs:
Address #4203 until we have more standard basis methods implemented.

### DIFF
--- a/src/Modules/UngradedModules/FreeMod.jl
+++ b/src/Modules/UngradedModules/FreeMod.jl
@@ -357,3 +357,19 @@ function syzygy_generators(
   return syzygy_generators(ambient_representatives_generators(I); parent)
 end
 
+function syzygy_generators(
+    a::Vector{T};
+    parent::Union{FreeMod{T}, Nothing} = nothing
+  ) where {CT <: Union{<:FieldElem, ZZRingElem}, # Can be adjusted to whatever is digested by Singular
+           T<:MPolyRingElem{CT}}
+  isempty(a) && return Vector{FreeModElem{T}}()
+  R = Oscar.parent(first(a))
+  @assert all(Oscar.parent(x) === R for x in a) "parent mismatch"
+  I = ideal(R, a)
+  s = Singular.syz(singular_generators(I))
+  F = (parent === nothing ? FreeMod(R, length(a)) : parent)::FreeMod{T}
+  @assert ngens(F) == length(a) "parent does not have the correct number of generators"
+  @assert rank(s) == length(a)
+  return elem_type(F)[F(s[i]) for i=1:Singular.ngens(s)]
+end
+
--- a/src/Modules/UngradedModules/FreeMod.jl
+++ b/src/Modules/UngradedModules/FreeMod.jl
@@ -308,3 +308,20 @@ end
 
 rels(F::FreeMod) = elem_type(F)[]
 
+function syzygy_generators(
+    g::Vector{T};
+    parent::Union{<:FreeMod, Nothing}=nothing
+  ) where {T<:FreeModElem}
+  isempty(g) && return Vector{T}()
+  F = Oscar.parent(first(g))
+  @assert all(Oscar.parent(x) === F for x in g) "parent mismatch"
+  R = base_ring(F)
+  m = length(g)
+  G = (parent === nothing ?  FreeMod(R, m) : parent)::typeof(F)
+  @assert ngens(G) == m "given parent does not have the correct number of generators"
+  phi = hom(G, F, g)
+  K, _ = kernel(phi)
+  return ambient_representatives_generators(K)
+end
+
+
--- a/src/Modules/UngradedModules/FreeMod.jl
+++ b/src/Modules/UngradedModules/FreeMod.jl
@@ -353,14 +353,13 @@ function syzygy_generators(
   R = Oscar.parent(first(a))
   @assert all(Oscar.parent(x) === R for x in a) "parent mismatch"
   F = FreeMod(R, 1)
-  I, _ = sub(F, [x*F[1] for x in a])
-  return syzygy_generators(ambient_representatives_generators(I); parent)
+  return syzygy_generators(elem_type(F)[x*F[1] for x in a]; parent)
 end
 
 function syzygy_generators(
     a::Vector{T};
     parent::Union{FreeMod{T}, Nothing} = nothing
-  ) where {CT <: Union{<:FieldElem, ZZRingElem}, # Can be adjusted to whatever is digested by Singular
+  ) where {CT <: Union{<:FieldElem, ZZRingElem, QQPolyRingElem}, # Can be adjusted to whatever is digested by Singular
            T<:MPolyRingElem{CT}}
   isempty(a) && return Vector{FreeModElem{T}}()
   R = Oscar.parent(first(a))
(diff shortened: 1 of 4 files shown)


thofma:
Did you check this this gives a different error message? As far as I see it will still crash in the `default_ordering(M)` call.

wdecker (approved):
Thx!

src/Modules/UngradedModules/FreeMod.jl
```diff
@@ -308,3 +308,67 @@ end
...
+  R = Oscar.parent(first(a))
+  @assert all(Oscar.parent(x) === R for x in a) "parent mismatch"
+  F = FreeMod(R, 1)
+  return syzygy_generators(elem_type(F)[x*F[1] for x in a]; parent)
+end
+
+function syzygy_generators(
+    a::Vector{T};
+    parent::Union{FreeMod{T}, Nothing} = nothing
+  ) where {CT <: Union{<:FieldElem, ZZRingElem, QQPolyRingElem}, # Can be adjusted to whatever is digested by Singular
```
HechtiDerLachs, line 362: @wdecker , @joschmitt : Your booktests in `H3.jlcon` were failing because I restricted the signature of the original method for `syzygy_generators`. In your example there in line 59, the syzygies are computed in a ring which is of type `MPolyDecRing{QQPolyRingElem, AbstractAlgebra.Generic.MPolyRing{QQPolyRingElem}}`, so we have a tower of a multivariate, graded polynomial ring over a univariate polynomial ring over QQ. I was unaware that Singular can apparently handle this coefficient ring and do reasonable computations over it. 

Could you please double-check that this is really what you wanted and that this is OK? I.e. to also allow coefficients of type `QQPolyRingElem` here? Thanks!

HechtiDerLachs: For a bit more context: Starting from line 59 in `H3.jlcon` I get the following.
```julia
julia> cr = coefficient_ring(parent(F))
Univariate polynomial ring in t over QQ

julia> Oscar.singular_coeff_ring(cr)
Singular coefficient ring over Qt

julia> typeof(ans)
Singular.N_Ring{QQPolyRingElem}

julia> Oscar.singular_poly_ring(parent(F))
Singular polynomial ring (Coeffs(23)),(x,y,z),(dp(3),C)

julia> typeof(ans)
Singular.PolyRing{Singular.n_RingElem{QQPolyRingElem}}

```
I'm not 100% sure what happens here, but it looks to me as if Singular is just wrapping your coefficient ring. I think, we should make sure that this is really reasonable and you're not just lucky that your computation went through.

wdecker: > I.e. to also allow coefficients of type `QQPolyRingElem` here?

It worked correctly in the example, so I did not double check. The old code is here:

function syzygy_generators(a::Vector{<:MPolyRingElem})
  I = ideal(a)
  s = Singular.syz(singular_generators(I))
  F = free_module(parent(a[1]), length(a))
  @assert rank(s) == length(a)
  return [F(s[i]) for i=1:Singular.ngens(s)]
end

@ederc, @hannes14 What is happening here in singular.jl/Singular?

wdecker: P.S. If Singular is working with the fraction field of T, I could rewrite the example accordingly

HechtiDerLachs: Then I suppose that it's OK as is now. Everything is executed with the old code again.

src/Modules/UngradedModules/FreeMod.jl
```diff
@@ -308,3 +308,67 @@ end
 
 rels(F::FreeMod) = elem_type(F)[]
 
+function syzygy_generators(
+    g::Vector{T};
+    parent::Union{<:FreeMod, Nothing}=nothing
+  ) where {T<:FreeModElem}
+  isempty(g) && return Vector{T}()
+  F = Oscar.parent(first(g))
```
thofma, line 316:
```suggestion
  F = parent(first(g))
```
there are more

src/Modules/UngradedModules/FreeMod.jl
```diff
@@ -308,3 +308,67 @@ end
...
+  @assert all(Oscar.parent(x) === F for x in g) "parent mismatch"
+  R = base_ring(F)
+  m = length(g)
+  G = (parent === nothing ?  FreeMod(R, m) : parent)::typeof(F)
+  @assert ngens(G) == m "given parent does not have the correct number of generators"
```
thofma:
```suggestion
  @req ngens(G) != m "given parent does not have the correct number of generators"
```

src/Modules/UngradedModules/FreeMod.jl
HechtiDerLachs: The `Oscar.` qualifiers are necessary because of the shadowing from the keyword argument.

src/Modules/UngradedModules/FreeMod.jl
HechtiDerLachs: Do you happen to know by hard which type of error this throws so that I can adjust the tests?

thofma: `ArgumentError`

src/Modules/UngradedModules/FreeMod.jl
thofma: I see, thanks

src/Modules/UngradedModules/FreeMod.jl
```diff
@@ -308,3 +308,67 @@ end
...
+    parent::Union{FreeMod{T}, Nothing} = nothing
+  ) where {CT <: Union{<:FieldElem, ZZRingElem, QQPolyRingElem}, # Can be adjusted to whatever is digested by Singular
+           T<:MPolyRingElem{CT}}
+  isempty(a) && return Vector{FreeModElem{T}}()
+  R = Oscar.parent(first(a))
+  @assert all(Oscar.parent(x) === R for x in a) "parent mismatch"
+  I = ideal(R, a)
+  s = Singular.syz(singular_generators(I))
+  F = (parent === nothing ? FreeMod(R, length(a)) : parent)::FreeMod{T}
+  @assert ngens(F) == length(a) "parent does not have the correct number of generators"
```
HechtiDerLachs:
```suggestion
  @req ngens(F) != length(a) "parent does not have the correct number of generators"
```

test/Modules/UngradedModules.jl
```diff
@@ -1275,3 +1275,41 @@ end
...
+  K, = kernel(phi); 
+  @test ambient_representatives_generators(K) == [(x*y + x + y + 1)*FF[1]]
+  @test_throws ErrorException standard_basis(K)
+  s1 = syzygy_generators(ambient_representatives_generators(M))
+  G = parent(first(s1))
+  s2 = syzygy_generators(ambient_representatives_generators(M); parent=G)
+  @test s1 == s2
+  s3 = syzygy_generators(ambient_representatives_generators(M))
+  @test s3 != s2
+  @test_throws AssertionError syzygy_generators(ambient_representatives_generators(M); parent=F)
```
HechtiDerLachs:
```suggestion
  @test_throws ArgumentError syzygy_generators(ambient_representatives_generators(M); parent=F)
```

thofma (approved):

PR #4206, opened 2024-10-15
add `stabilizer` for G-sets

### LINKED ISSUE #3259: Implement `stabilizer` for (transitive?) G-sets
I would expect `stabilizer(gset)` to return the stabilizer of `representative(gset)` with respect to the action.

Moreover, I would expect it to cache that stabilizer.

This way it would be a very natural fit for e.g. representing conjugacy classes, or the result of orbit-stabilizer computations

This mirrors the GAP function `StabilizerOfExternalSet` 

One might perhaps restrict this to transitive G-sets, though technically there is no reason as long as we are clear what it means.

See also issue #108

ThomasBreuer:
resolves #3259

### DIFF
--- a/docs/src/Groups/action.md
+++ b/docs/src/Groups/action.md
@@ -70,5 +70,6 @@ orbits(Omega::T) where T <: GSetByElements{TG} where TG <: GAPGroup
 ## Stabilizers
 
 ```@docs
-stabilizer(G::Oscar.GAPGroup, pnt::Any, actfun::Function)
+stabilizer(G::GAPGroup, pnt::Any, actfun::Function)
+stabilizer(Omega::GSet)
 ```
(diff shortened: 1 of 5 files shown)


fingolfin (approved):

ThomasBreuer:
One test failed because it was canceled after 2 1/2 hours (test (nightly, book, ubuntu-latest).
Since this pull request introduced a new feature that cannot be used in these tests, I think that the proposed changes are not responsible for slowing down the test.

benlorenz:
> One test failed because it was canceled after 2 1/2 hours (test (nightly, book, ubuntu-latest). Since this pull request introduced a new feature that cannot be used in these tests, I think that the proposed changes are not responsible for slowing down the test.

I have seen this happen in the scheduled tests on master as well: https://github.com/oscar-system/Oscar.jl/actions/runs/11285680359/job/31388879146

joschmitt:
The failing test seems unrelated because it also fails on master, so I merge this now.

PR #4205, opened 2024-10-15
[1.0] Update book examples

lkastner:
Examples from modified book source migrated here.

PR #4209, opened 2024-10-16
booktest: add error checking for non-jlcon files also on master

benlorenz:
from #4205

### DIFF
--- a/src/Groups/GAPGroups.jl
+++ b/src/Groups/GAPGroups.jl
@@ -372,10 +372,15 @@ end
 
 Base.:*(x::GAPGroupElem, y::GAPGroupElem) = _prod(x, y)
 
+
+isequal(x::GAPGroup, y::GAPGroup) = GapObj(x) == GapObj(y)
+
 function ==(x::GAPGroup, y::GAPGroup)
   _check_compatible(x, y)
   return GapObj(x) == GapObj(y)
 end
+
+isequal(x::BasicGAPGroupElem, y::BasicGAPGroupElem) = GapObj(x) == GapObj(y)
 
 # For two `BasicGAPGroupElem`s,
 # we allow the question for equality if their parents fit together
--- a/test/book/cornerstones/groups/auxiliary_code/main.jl
+++ b/test/book/cornerstones/groups/auxiliary_code/main.jl
@@ -1,5 +1,3 @@
 import Pkg
 Pkg.add(name="GenericCharacterTables", version="0.4"; io=devnull)
 using GenericCharacterTables
-# for nicer printing
-using GenericCharacterTables: ParameterException
--- a/test/book/test.jl
+++ b/test/book/test.jl
@@ -166,7 +166,18 @@ isdefined(Main, :FakeTerminals) || include(joinpath(pkgdir(REPL),"test","FakeTer
     if jlcon_mode
       input_string = "\e[200~$s\e[201~"
     end
+    haderror = false
     REPL.activate(mockrepl.mockdule)
+    # this allows us to detect errors in the middle of non-jlcon files
+    if !jlcon_mode
+      od = Base.active_repl.interface.modes[1].on_done
+      Base.active_repl.interface.modes[1].on_done = function (x...)
+                if Base.active_repl.waserror
+                  haderror = true
+                end
+                od(x...)
+              end
+    end
     result = redirect_stdout(mockrepl.out_stream) do
       input_task = @async begin
         write(mockrepl.stdin_write, input_string)
@@ -175,8 +186,16 @@ isdefined(Main, :FakeTerminals) || include(joinpath(pkgdir(REPL),"test","FakeTer
       wait(input_task)
       readuntil(mockrepl.output.out, "\nEND_BLOCK")
     end
+    if !jlcon_mode
+      # restore on-done
+      Base.active_repl.interface.modes[1].on_done=od
+    end
     REPL.activate(Main)
-    return sanitize_output(result)
+    output = sanitize_output(result)
+    if !jlcon_mode && haderror
+      error("ERROR in jl-mode:\n", output)
+    end
+    return output
   end
 
   function test_chapter(chapter::String="")


benlorenz:
This uncovered a new error in `test/book/cornerstones/number-theory/galoismod_1.jlcon`:
```julia
julia> for i in 1:8, j in 1:number_of_small_groups(i)
         G = small_group(i, j)
         if is_abelian(G)
           continue
         end
         QG = QQ[G]
         ZG = integral_group_ring(QG)
         println(i, " ", j, " ", describe(G), ": ", locally_free_class_group(ZG))
       end
6 1 S3: Z/1
ERROR: ArgumentError: G1 and G2 are not compatible
Stacktrace:
  [1] #_check_compatible#1398
    @ /home/datastore/lorenz/software/julia/Oscar.jl/src/Groups/types.jl:683 [inlined]
  [2] _check_compatible
    @ /home/datastore/lorenz/software/julia/Oscar.jl/src/Groups/types.jl:681 [inlined]
  [3] ==
    @ /home/datastore/lorenz/software/julia/Oscar.jl/src/Groups/GAPGroups.jl:376 [inlined]
  [4] isequal
    @ ./operators.jl:134 [inlined]
  [5] _isequal (repeats 2 times)
    @ ./tuple.jl:536 [inlined]
  [6] isequal(t1::Tuple{QQField, PcGroup, typeof(*)}, t2::Tuple{QQField, PcGroup, typeof(*)})
    @ Base ./tuple.jl:533
  [7] ht_keyindex(h::Dict{Tuple{Ring, Any, Any}, WeakRef}, key::Tuple{QQField, PcGroup, typeof(*)})
    @ Base ./dict.jl:249
  [8] haskey
    @ ./dict.jl:548 [inlined]
  [9] (::AbstractAlgebra.var"#50#51"{Hecke.var"#155#158"{…}, AbstractAlgebra.WeakValueDict{…}, Tuple{…}})()
    @ AbstractAlgebra /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/DRky3/src/WeakValueDict.jl:544
 [10] lock(f::AbstractAlgebra.var"#50#51"{Hecke.var"#155#158"{…}, AbstractAlgebra.WeakValueDict{…}, Tuple{…}}, l::ReentrantLock)
    @ Base ./lock.jl:232
 [11] lock
    @ /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/DRky3/src/WeakValueDict.jl:512 [inlined]
 [12] get!
    @ /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/DRky3/src/WeakValueDict.jl:542 [inlined]
 [13] get_cached!
    @ /home/datastore/lorenz/software/julia/depot/packages/AbstractAlgebra/DRky3/src/AbstractAlgebra.jl:159 [inlined]
 [14] _
    @ /home/datastore/lorenz/software/julia/depot/packages/Hecke/7wI0D/src/AlgAss/Types.jl:156 [inlined]
 [15] GroupAlgebra
    @ /home/datastore/lorenz/software/julia/depot/packages/Hecke/7wI0D/src/AlgAss/Types.jl:155 [inlined]
 [16] #group_algebra#4266
    @ /home/datastore/lorenz/software/julia/depot/packages/Hecke/7wI0D/src/AlgAss/AlgGrp.jl:72 [inlined]
 [17] group_algebra
    @ /home/datastore/lorenz/software/julia/depot/packages/Hecke/7wI0D/src/AlgAss/AlgGrp.jl:71 [inlined]
 [18] getindex(K::QQField, G::PcGroup)
    @ Hecke /home/datastore/lorenz/software/julia/depot/packages/Hecke/7wI0D/src/AlgAss/AlgGrp.jl:93
 [19] top-level scope
    @ ./REPL[2]:6
```
The error (from #3166) is not in 1.0 which is why this only happens here and not on the other booktest PR.
Unfortunately before the change in this PR this error was silently ignored on master.

cc: @ThomasBreuer

thofma:
This is the reason why one might want to have a non-throwing `isequal`.

ThomasBreuer:
What happens here is that the constructor for the group algebra `R[G]`  from the ring `R` and the group `G` and the group operation `op` caches the already known results in a dictionary with key `(R, G, op)`.
In order to check whether a given key occurs already, the groups in question are compared for equality via `==`.
And since #3166, this comparison is restricted to pairs of groups that are compatible in the sense that multiplying their elements makes sense. (The groups need not have the same type.)

The fact that this caching mechanism tries to compare two groups that are completely unrelated seems to contradict @fieker's verdict that such a call of `==` indicates an error or a misunderstanding on the side of the user, and that the best way to avoid confusion is to throw an exception.
(Or should the comparison of the keys perhaps better check the group objects for identity?)

thofma:
The groups are compared using `isequal`, which falls back to `==`. The problem would be solved by having `isequal` being the non-throwing version of `==`.

ThomasBreuer:
Then we can recommend to write all `==` methods in the form `==(a, b; error::Bool = true)`, where an exception is thrown or `false` is returned (depending on `error`) if `a` and `b` are regarded as not comparable, and `isequal` methods for our objects can look like `isequal(a, b) = ==(a, b, error = false)`?

fingolfin:
I am not sure I'd want to recommend that style, it definitely is not what we do in AA and Nemo so far. In any case, we need a solution now, and the simple one is to add `isequal` methods for `GAPGroup` and `BasicGAPGroupElem` that don't include the parent checks.

I've pushed such methods to this PR now, let's see if they help.

ThomasBreuer:
> I've pushed such methods to this PR now, let's see if they help.

@fingolfin How is this change related to the attempt from #4196, which got merged earlier this week?

thofma:
Thanks @fingolfin. I think eventually the cache should use `===` as @ThomasBreuer suggests. I have to check whether we have some `IdDict` with "weak values" lying around.

lgoettgens:
> I have to check whether we have some `IdDict` with "weak values" lying around.

Julia Base has `WeakKeyDict`, AA has `WeakKeyIdDict` and `WeakValueDict`. Maybe you need to squash together the latter two to some `WeakValueIdDict` in AA?

fingolfin (approved):
Looks good to me now but it is still marked as draft (@benlorenz ?)

PR #4216, opened 2024-10-17
Fix JuliaFormatter version to 1.x to avoid breaking changes of 2.x for now

lkastner:
Basically @benlorenz  fixed this.


joschmitt (approved):
Thanks!

Should we also open an issue, so we don't forget about this?

benlorenz (approved):

fingolfin (approved):

Issue #4170, opened 2024-10-01
`galthy.jl` / `van_hoeij`: must never be zero


benlorenz:
**Describe the bug**
Rare error `must never be zero` in `test/NumberTheory/galthy.jl:109` in our CI, e.g.:
https://github.com/oscar-system/Oscar.jl/actions/runs/10960677682/job/30435848700#step:9:7123.
The issue might be in `fixed_field` or somewhere in Hecke?

**To Reproduce**
After trying some seeds I found that this code reproduces the error (on three different linux computers with julia 1.10.5 and an M1 mac with julia 1.10.4):
```julia
julia> R, s = QQ[:s];
julia> K, q = number_field(s^2 - 2, "q");
julia> Kw, w = polynomial_ring(K, :w);
julia> f = w^16 - 32*w^14 - 192*w^12 + 22720*w^10 + 23104*w^8 - 2580480*w^6 + 41287680*w^4 + 106168320*w^2 + 84934656;
julia> g, s = galois_group(f);
julia> ss = map(representative, subgroup_classes(g));
julia> H = ss[1000];

julia> Oscar.randseed!(15290);

julia> f = fixed_field(s, H)
ERROR: must never be zero
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] van_hoeij(f::AbstractAlgebra.Generic.Poly{AbsSimpleNumFieldElem}, P::AbsSimpleNumFieldOrderIdeal; prec_scale::Int64)
    @ Hecke ~/.julia/packages/Hecke/7wI0D/src/NumField/NfAbs/PolyFact.jl:759
  [3] van_hoeij(f::AbstractAlgebra.Generic.Poly{AbsSimpleNumFieldElem}, P::AbsSimpleNumFieldOrderIdeal)
    @ Hecke ~/.julia/packages/Hecke/7wI0D/src/NumField/NfAbs/PolyFact.jl:541
  [4] factor_new(f::AbstractAlgebra.Generic.Poly{AbsSimpleNumFieldElem})
    @ Hecke ~/.julia/packages/Hecke/7wI0D/src/NumField/NfAbs/PolyFact.jl:307
  [5] _factor(f::AbstractAlgebra.Generic.Poly{AbsSimpleNumFieldElem}; algo::Symbol)
    @ Hecke ~/.julia/packages/Hecke/7wI0D/src/NumField/NfAbs/Elem.jl:354
  [6] _factor
[...]

PR #4217, opened 2024-10-17
Module quotients I

wdecker:


src/Modules/UngradedModules/SubquoModule.jl
```diff
@@ -1619,6 +1619,130 @@ function intersect(M::SubquoModule{T}, N::SubquoModule{T}) where T
...
+"""
+function quotient(M::SubquoModule{T}, N::SubquoModule{T}) where T
+  @assert base_ring(M) == base_ring(N)
+  @assert ambient_module(M) == ambient_module(N)
+  MplusN, iM, _ = sum(M, N)
+  Q, _ = quo(MplusN, [iM(x) for x in gens(M)])
+  return annihilator(Q)
+end
+
+(::Colon)(M::SubquoModule{T}, N::SubquoModule{T}) where T = quotient(M, N)
```
lgoettgens, line 1775: I personally don't like overloading basic operators (like `:` to create `UnitRange`s) with more complex functions as this can make it really hard to understand code that uses these.
What do other people think of this?

wdecker: This is already in use for ideal quotients. It is the standard mathematical notation and, in interactve sessions, everybody would love to use it that way. Of course, when writing code, you can still use quotient.

fingolfin: While I am also in general wary of this, I believe it is fine here, and extremely convenient if one does a certain kind of math.

And as @wdecker points out, we already do it, this just adds another case (which I think is good, it is more confusing if we support this *sometimes*  but not in all cases in which it makes sense)

src/Modules/UngradedModules/SubquoModule.jl
```diff
@@ -1619,6 +1619,130 @@ function intersect(M::SubquoModule{T}, N::SubquoModule{T}) where T
...
+Return the annihilator of `N`.
+
+!!! note
+    By definition, the annihilator of $N$ is the ideal $0:N = \{a \in R \mid aN = 0\}$. 
+    Here, `R = base_ring(N)`.
+
+# Examples
+
+```jldoctest
+julia> R, (x, y, z) = polynomial_ring(QQ, [:x, :y, :z]);
```
fingolfin, line 1638: Just FYI these days you could also write
```suggestion
julia> R = @polynomial_ring(QQ, [:x, :y, :z]);
```

src/Modules/UngradedModules/SubquoModule.jl
lgoettgens: I didn't know we already have this for other types, then this is nothing I should complain to. Sorry for the inconvenience

src/Modules/UngradedModules/SubquoModule.jl
wdecker: Thx, the next time ...

wdecker:
@thofma The failing test here signals a problem in number theory. Any idea?

thofma:
Yes, this is https://github.com/oscar-system/Oscar.jl/issues/4170 can be ignored.

wdecker:
@HechtiDerLachs Can you please check whether you agree and if so, merge this PR. In the next PR I wish to do quotients of type M:J, M module, J ideal. Do you have already a generic method for this? Can we reasonably reduce this to `annihilator`?

HechtiDerLachs:
> In the next PR I wish to do quotients of type M:J, M module, J ideal. Do you have already a generic method for this?

No, I don't think I or anyone else has implemented a generic method for this, yet.

> Can we reasonably reduce this to `annihilator`?

I don't see this straight away. There are certainly generic implementations of this. But already my implementation for `annihilator` was probably not the most effective one. I just wanted to have something working at the time. Maybe it's best to implement this as a kernel of an appropriate map `M \to (F/M)^k` where `M \subset F` is a submodule of a free module,  `k` is the number of generators of `J` and the `j`-th component of the map is multiplication with that generator?

As you said the other day, we should wrap the established Singular routine for the polynomial case.

HechtiDerLachs (approved):

PR #4219, opened 2024-10-18
Minor improvements for FTheoryTools

HereAround:
If nothing unexpected happens, then this PR should:

* Remove wrong and unused final 3 blowups for the big model.
* Serialize the resolutions.

cc @apturner @emikelsons


apturner (approved):
This all looks good, thanks @HereAround.

PR #4225, opened 2024-10-21
Module quotients II

wdecker:


wdecker:
@HechtiDerLachs 
Our definition here is: $M:J = \{a \in A \mid Ja \subset M\}$. Here, $A$ is the ambient module of $M$.
Please check and merge if o.k..

Then, in a new PR, please replace general dummy function for this method of `quotient´ by general function:
1. Case J = <f> principal ideal: Syzygy computation in the ambient ring A of M and, thus, `kernel` computation.
2. Apply 1 to all generators of J and intersect results.

HechtiDerLachs (approved):

PR #4226, opened 2024-10-22
Extend `is_invertible_with_inverse` to rings of type `MPolyQuoRing{<:MPolyDecRingElem}`

wdecker:
…MPolyDecRingElem}`

### DIFF
--- a/src/Rings/MPolyQuo.jl
+++ b/src/Rings/MPolyQuo.jl
@@ -1172,7 +1172,7 @@ c[1]^2 - c[1] - c[2] + 1
 
 julia> inv(f)
 c[1] + c[2] + c[3] + 1
-a
+
 ```
 """
 function inv(a::MPolyQuoRingElem)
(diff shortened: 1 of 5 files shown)


wdecker:
I added an example from intersection theory where the extension is needed.

src/Rings/MPolyQuo.jl
```diff
@@ -1100,6 +1100,33 @@ end
...
+
+julia> f = A(c[1]^2 - c[1] - c[2] + 1)
+c[1]^2 - c[1] - c[2] + 1
+
+julia> inv(f)
+c[1] + c[2] + c[3] + 1
+
+julia> f*inv(f)
+1
+
```
joschmitt: Somehow the documented function is not mentioned in this example?

wdecker: Thx, I confused two versions of what I had. Should be o.k. now.I will add more of this to the docu a little later today.

wdecker:
O.k. from my side if lights are green

fieker (approved):

PR #4227, opened 2024-10-22
Some fixes and performance improvements for `iso_oscar_gap(::LieAlgebra)`

lgoettgens:
relevant changes:
- `_iso_oscar_gap(::AbstractLieAlgebra)` now accesses internals of `AbstractLieAlgebra`, but this has a >100x performance improvement
- `_iso_oscar_gap(::LinearLieAlgebra)` now also transfers the root system (if one is already known)
- `Oscar.iso_oscar_gap` now asserts the return type given in the docstring
- Remove some iso inv caching, that was an artifact from https://github.com/oscar-system/Oscar.jl/pull/2330


ThomasBreuer (approved):
Looks good. Thanks.

PR #4228, opened 2024-10-22
Improve performance of `bracket` for `AbstractLieAlgebraElem`

lgoettgens:
by being a bit smarter with handling sparse rows

### DIFF
--- a/experimental/LieAlgebras/src/AbstractLieAlgebra.jl
+++ b/experimental/LieAlgebras/src/AbstractLieAlgebra.jl
@@ -68,12 +68,15 @@ function bracket(
 ) where {C<:FieldElem}
   check_parent(x, y)
   L = parent(x)
-  mat = sum(
-    cxi * cyj * _struct_consts(L)[i, j] for (i, cxi) in enumerate(coefficients(x)),
-    (j, cyj) in enumerate(coefficients(y));
-    init=sparse_row(coefficient_ring(L)),
-  )
-  return L(mat)
+  vec = sparse_row(coefficient_ring(L))
+  for (i, cxi) in enumerate(coefficients(x))
+    iszero(cxi) && continue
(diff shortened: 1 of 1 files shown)


experimental/LieAlgebras/src/AbstractLieAlgebra.jl
```diff
@@ -68,12 +68,15 @@ function bracket(
...
-    (j, cyj) in enumerate(coefficients(y));
-    init=sparse_row(coefficient_ring(L)),
-  )
-  return L(mat)
+  vec = sparse_row(coefficient_ring(L))
+  for (i, cxi) in enumerate(coefficients(x))
+    iszero(cxi) && continue
+    for (j, cyj) in enumerate(coefficients(y))
+      iszero(cyj) && continue
+      Hecke.add_scaled_row!(_struct_consts(L)[i, j], vec, cxi * cyj)
```
fingolfin, line 76: I double checked and it seems this is indeed the correct argument order -- very confusing when one already knows `addmul!` :-(. 

On the up side, perhaps we can just simply define methods
```julia
addmul!(a::SRow{T}, b::SRow{T}, c::T) where T = add_right_scaled_row(b, a, c) 
addmul!(a::SRow{T}, b::T, c::SRow{T}) where T = add_left_scaled_row(c, a, b)
```
and migrate to those over time?


But of course that's not any reason to hold up this PR, just musings :-)

lgoettgens: See https://github.com/thofma/Hecke.jl/pull/1659. I would suggest to merge this PR here right now and then once https://github.com/thofma/Hecke.jl/pull/1659 is available go through Oscar and phase out the `add_scaled_row!` calls. (if you agree please approve and merge this)

fingolfin (approved):

PR #4235, opened 2024-10-23
Add `monomial_basis` for `MPolyQuoLocRing`

Syz-MS:
Adding monomial basis for MPolyQuoLocRing in preparation for versal_deformation(HypersurfaceGerm).

@afkafkafk13 please have a look at this.


src/Rings/mpolyquo-localizations.jl
```diff
@@ -1850,6 +1850,52 @@ function vector_space_dimension(R::MPolyQuoLocRing{<:Field, <:Any,<:Any, <:Any,
   return vector_space_dimension(quo(base_ring(R),ideal(base_ring(R),gens(LI)))[1])
 end
 
+@doc raw"""
+    monomial_basis(A::MPolyQuoLocRing{<:Field, <:Any, <:Any, <:Any, <:MPolyComplementOfKPointIdeal})
+
+If, say, `A = L/I`, where `L` is a localization of multivariate polynomial ring over a field
+`K` at a point 'p', and `I` is an ideal of `L`, return a vector of monomials of `L`
```
lgoettgens:
```suggestion
`K` at a point `p`, and `I` is an ideal of `L`, return a vector of monomials of `L`
```

afkafkafk13:
Could you also take care of localized rings? Up to now this seems only available for graded rings and affine algebras.

Sideremark: Do not shift the monomials back. They are still a basis after the shift (e.g. {1,x-3,x^2-6x+9 } spans the same vector space as {1,x,x^2}, but the latter is more human readable).

test/Rings/mpolyquo-localizations.jl
```diff
@@ -510,3 +510,20 @@ end
...
+  Q3, _ = quo(L, ideal(L, L.([x^2, y^3])))
+  @test isempty(monomial_basis(Q1))
+  @test_throws InfiniteDimensionError() monomial_basis(Q2)
+  @test monomial_basis(Q3) == L.([x*y^2, y^2, x*y, y, x, 1])
+  L1, _ = localization(R, complement_of_point_ideal(R, [1,2]))
+  Q4, _ = quo(L1, ideal(L1, [(x-1)^2, (y-2)^2]))
+  @test monomial_basis(Q4) == L1.([(x-1)*(y-2), y-2, x-1, 1])  
+  Q5, _ = quo(L1, ideal(L1, L1.([x*y, x, y])))
+  @test isempty(monomial_basis(Q5))
+end
```
fingolfin: Please add a trailing newline here
```suggestion
end

```

Syz-MS:
> Could you also take care of localized rings? Up to now this seems only available for graded rings and affine algebras.
> 
> Sideremark: Do not shift the monomials back. They are still a basis after the shift (e.g. {1,x-3,x^2-6x+9 } spans the same vector space as {1,x,x^2}, but the latter is more human readable).

@afkafkafk13 I just removed the backshift. Regarding the first part of your comment I am unsure which specific ring types in Oscar you are  refering to.

afkafkafk13:
> Could you also take care of localized rings? Up to now this seems only available for graded rings and affine algebras.
> 
> Sideremark: Do not shift the monomials back. They are still a basis after the shift (e.g. {1,x-3,x^2-6x+9 } spans the same vector space as {1,x,x^2}, but the latter is more human readable).

To clarify, I was envisioning a method monomial_basis(R,I) for R a MPolyLocalizedRing and I an Ideal thereof. The benefit would be that the quo does not need to be generated on user level. Note that the other method immediately passes from R/I to localized ring and modulus anyway!

wdecker:
> To clarify, I was envisioning a method monomial_basis(R,I) for R a MPolyLocalizedRing and I an Ideal thereof.

Do you mean any type of `MPolyLocalizedRing` or just localizations at a point?

afkafkafk13:
> > To clarify, I was envisioning a method monomial_basis(R,I) for R a MPolyLocalizedRing and I an Ideal thereof.
> 
> Do you mean any type of `MPolyLocalizedRing` or just localizations at a point?

Just for localizations at a point. I was already in contact with Marcel outside of github.

Syz-MS:
Should I also move the functionality of monomial_basis(A::MPolyQuoRing) into monomial_basis(R::MPolyRing, I::MPolyIdeal) like I did in the local case in the last commit and therefore not creating the quotient at all in monomial_basis(R::MPolyRing, I::MPolyIdeal)?

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1850,6 +1850,56 @@ function vector_space_dimension(R::MPolyQuoLocRing{<:Field, <:Any,<:Any, <:Any,
...
+```
+"""
+function monomial_basis(L::MPolyLocRing{<:Field, <:Any, <:Any, <:Any, <:MPolyComplementOfKPointIdeal}, I::MPolyLocalizedIdeal)
+  base_ring(I) == L || error("ideal does not belong to the correct ring")
+  G = numerator.(gens(I))
+  shift,_ = base_ring_shifts(L)
+  G_0 = shift.(G) 
+  R = base_ring(L)
+  S = standard_basis(ideal(R, G_0), ordering = negdeglex(R))
+  return L.(monomial_basis(R, ideal(S)))
```
ederc: As pointed out by @wdecker this is problematic: First a standard basis `S` for `G_0` w.r.t. to `negdeglex(R)` is computed, but then inside the following `monomial_basis()` call the `monomial_basis()` function from `mpoly-affine-algebras.jl` is called on the quotient `R/S`, which induces a Groebner basis computation on `S` w.r.t. to `degrevlex(R)`. So in the end we computed two bases for `S`, one w.r.t. the local ordering and another w.r.t. a global ordering, which I assume is not what is intended.

afkafkafk13: S should be `leading_ideal` and not `standard_basis`. Then S is a monomial ideal, for which the ordering is irrelevant.

test/Rings/mpolyquo-localizations.jl
```diff
@@ -510,3 +510,17 @@ end
...
+  Q1, _ = quo(L, ideal(L, L(x+1)))
+  @test isempty(monomial_basis(Q1))
+  Q2, _ = quo(L, ideal(L, L(x^2)))  
+  @test_throws InfiniteDimensionError() monomial_basis(Q2)
+  Q3, _ = quo(L, ideal(L, L.([x^2, y^3])))  
+  @test monomial_basis(Q3) == L.([x*y^2, y^2, x*y, y, x, 1])
+  L1, _ = localization(R, complement_of_point_ideal(R, [1,2]))
+  @test monomial_basis(L1, ideal(L1, [(x-1)^2, (y-2)^2])) == L1.([x*y, y, x, 1])  
+  @test isempty(monomial_basis(L1, ideal(L1, L1.([x, y]))))
+end
```
afkafkafk13: Please add one or two non-homogeneous cases to the tests, so that the difference between localization and non-localized ring will be detected and unexpected behaviour (by use of wrong ordering, as remarked by @ederc above) would pop up as an error.

src/Rings/mpolyquo-localizations.jl
Syz-MS: fixed

test/Rings/mpolyquo-localizations.jl
Syz-MS: just added the additional tests

src/Rings/mpoly-affine-algebras.jl
```diff
@@ -151,6 +152,11 @@ function monomial_basis(A::MPolyQuoRing)
   return gens(MPolyIdeal(base_ring(I), si))
 end
 
+function monomial_basis(R::MPolyRing, I::MPolyIdeal)
+  base_ring(I) == R || error("ideal does not belong to the correct ring")
```
fingolfin: Looking at this I was wondering "Why does one have to specify `R` at all? Why not just this:"
```suggestion
function monomial_basis(I::MPolyIdeal)
  R = base_ring(I)
```

Perhaps the answer is "because this is not really a "monomial basis of `I` so we want to avoid the impression that is by giving two arguments". I can see why one would want to avoid writing down a quotient "just" to get this monomial basis. But it still feels at odds with fundamental OSCAR principles (where we like to avoid such tricks). And in any case, you *still* compute `quo` in here, so there is nothing being saved here...

Mathematically I think this is about finding a monomial basis of a complement of `I` in `R`. So an alternatively would be to do something like this:
```suggestion
function monomial_basis_of_complement(I::MPolyIdeal)
  R = base_ring(I)
```

And then one could go a step further and move most of the body of the `monomial_basis(A::MPolyQuoRing)` method in here, as it only uses `I`, not the quotient ring?

src/Rings/mpolyquo-localizations.jl
```diff
@@ -1850,6 +1850,56 @@ function vector_space_dimension(R::MPolyQuoLocRing{<:Field, <:Any,<:Any, <:Any,
...
+julia> monomial_basis(A)
+4-element Vector{Oscar.MPolyLocRingElem{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem, Oscar.MPolyComplementOfKPointIdeal{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem}}}:
+ x*y
+ y
+ x
+ 1
+```
+"""
+function monomial_basis(L::MPolyLocRing{<:Field, <:Any, <:Any, <:Any, <:MPolyComplementOfKPointIdeal}, I::MPolyLocalizedIdeal)
+  base_ring(I) == L || error("ideal does not belong to the correct ring")
```
fingolfin: Same question as above: why not
```suggestion
function monomial_basis(I::MPolyLocalizedIdeal)
  L = base_ring(I)
```
(though perhaps with a different name)

src/Rings/mpoly-affine-algebras.jl
wdecker: What we compute here is a monomial basis of a quotient ring A = R/I, say. As does @fingolfin, I have already problems to understand why we should split this up like in `monomial_basis(R, I)` instead of `monomial_basis(A)` (this may already mislead the unexperienced user). Using just `monomial_basis(I)` makes it even worth. We do compute a monomial basis of A, not of (R, I), and certainly not of I.

Sorry, but `monomial_basis_of_complement(I)` does not make any mathematical sense.

In sum, I suggest to forget this idea of splitting up. In particular, I am against splitting up in cases where we already have a working `monomial_basis(A)`.

src/Rings/mpolyquo-localizations.jl
wdecker: See my remark above.

src/Rings/mpoly-affine-algebras.jl
afkafkafk13: > Perhaps the answer is "because this is not really a "monomial basis of `I` so we want to avoid the impression that is by giving two arguments". I can see why one would want to avoid writing down a quotient "just" to get this monomial basis. But it still feels at odds with fundamental OSCAR principles (where we like to avoid such tricks). And in any case, you _still_ compute `quo` in here, so there is nothing being saved here...

This is indeed the reason. `monomial_basis(I)` is absolutely not the way to go w.r.t. Oscar principles. But in particular in the context of germs and of ideal sheaves, you internally usually have an ideal  in your hands and not a quotient ring when you find yourself in a setting to call monomial_basis.

To the remark with the `quo`: The lines here are not the key contribution of this PR. This is only the compatibility change, which currently does not touch the old code (as I do not know whether we are allowed to change this). The new contribution is in the local case (other file below) and there the `quo` is not computed, but `monomial_basis(R,I)` is the inner worker, whereas `monomial_basis(quo...)` is the outside appearance. 
By the way, I would like to change the non-local case in the same way, if we are allowed to.

> 
> Mathematically I think this is about finding a monomial basis of a complement of `I` in `R`. So an alternatively would be to do something like this: 
> And then one could go a step further and move most of the body of the `monomial_basis(A::MPolyQuoRing)` method in here, as it only uses `I`, not the quotient ring?

NO, we are not looking at the complement. We are looking at the number of elements(=classes mod I) of the factorring R/I. I could live with monomial_basis_of_quo(I), but how do you type-dispatch on the type of the localization of R with this notation?

afkafkafk13: By the way: `quo(R,I)` also contains the R, even though it is not necessary. We simply tried to stay in tune with `quo` (thinking that it would certainly follow OSCAR's principles).

src/Rings/mpolyquo-localizations.jl
afkafkafk13: This is not about the name, but about the type dispatch for the localization of the ring.

src/Rings/mpoly-affine-algebras.jl
thofma: Since this is not returning a basis of $R/I$, but a lifted basis of $R/I$ to $R$, it is in fact a basis for the complement of $I$, complement as a $K$-subspace of $R$.

afkafkafk13: @wdecker: nobody wants to get rid of  `monomial_basis(A)`! This is the way the average user should use it! 

It is just a question of having a more internal, but documented variant for specialists (e.g. scripting, programming), which avoids creating the `quo` specifically to call monomial_basis which in turn immediately passes back to the modulus. This change is not done here yet. It is only done in the local case below. (But I would like to do it here as well.)

src/Rings/mpolyquo-localizations.jl
thofma: The type of the ring is the first parameter of the ideal, so it would be
```
blabla(I::MPolyLocalizedIdeal{MPolyLocRing{<:Field, <:Any, <:Any, <:Any, <:MPolyComplementOfKPointIdeal}})
```

src/Rings/mpoly-affine-algebras.jl
wdecker: If you document it, the user will find it. If it is internal, you may also put an _ in front. Also, you compute the quo anyway. So what do you gain?. Typing monomial_basis(R, I) instead of monomial_basis(quo(R, I)[1])? But, if it is important to you, I will not stand in your way.

afkafkafk13: I see this as an ok to put the work into the internal monomial_basis(R,I) and let the outside one monomial_basis(A) call it, like in the local case.

jankoboehm:
It should be sufficient to have a monomial basis command for the quotient ring.

afkafkafk13:
ok. I won't insist on the global case, but the local one I would like to keep as it is right now -- moving the inner syntax to _monomial_basis (with its own docstring).

wdecker:
How about adding a more meaningful example, that is, an example where the localization actually matters:

julia> R, (x,y) = QQ["x","y"];

julia> I = ideal(R, 2*x^2-y^3, 2*x^2-y^5) # two cusps
Ideal generated by
  2*x^2 - y^3
  2*x^2 - y^5

julia> A, _ = quo(R, I)
(Quotient of multivariate polynomial ring by ideal (2*x^2 - y^3, 2*x^2 - y^5), Map: R -> A)

julia> monomial_basis(A)
10-element Vector{QQMPolyRingElem}:
 x*y^2
 y^2
 x^3*y
 x^2*y
 x*y
 y
 x^3
 x^2
 x
 1

julia> L, _ = localization(A, complement_of_point_ideal(R, [0,0]));

julia> Iloc = ideal(L, 2*x^2-y^3, 2*x^2-y^5)
Ideal generated by
  2*x^2 - y^3
  2*x^2 - y^5

julia> Aloc,_ = quo(L, I)
(Localization of quotient of multivariate polynomial ring at complement of maximal ideal, hom: Localized quotient of multivariate polynomial ring -> Localized quotient of multivariate polynomial ring)

julia> monomial_basis(Aloc)
6-element Vector{Oscar.MPolyLocRingElem{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem, Oscar.MPolyComplementOfKPointIdeal{QQField, QQFieldElem, QQMPolyRing, QQMPolyRingElem}}}:
 x*y^2
 y^2
 x*y
 y
 x
 1

afkafkafk13 (approved):

Issue #4239, opened 2024-10-24
Missing `characteristic(R::MPolyQuoRing)` method


fingolfin:
Presumably it would return `characteristic(base_ring(R))` *unless* the ideal being factored out happens to be the the entire ring (then return 1). So that would require a GB computation (but hopefully that already is being cached?)

fieker:
On Thu, Oct 24, 2024 at 02:57:19AM -0700, Max Horn wrote:
> Presumably it would return `characteristic(base_ring(R))` *unless* the ideal being factored out happens to be the the entire ring (then return 1). So that would require a GB computation (but hopefully that already is being cached?)
> 
What is about Z[x,y]/<2>?
Or is this illegal by type?

fingolfin:
Dunno. But even if we only support it for the case were the coeffs are from a field it would already be useful.

joschmitt:
> What is about Z[x,y]/<2>? Or is this illegal by type?

One can construct the ring and do something with it:
```
julia> R, (x, y) = ZZ["x", "y"]
(Multivariate polynomial ring in 2 variables over ZZ, ZZMPolyRingElem[x, y])

julia> S, _ = quo(R, ideal(R, [2]))
(Quotient of multivariate polynomial ring by ideal (2), Map: R -> S)

julia> iszero(one(S))
false

julia> iszero(2*one(S))
true
```
Whether this works by design or by accident I don't know.

fingolfin:
Thanks to @joschmitt for addressing this over fields. I guess for integers we could handle it by computing a GB and checking if it contains any degree 0 entries; if so, compute their gcd to get the characteristic.

Beyond that I guess it can get arbitrarily complicated and I'd be happy to throw the towel in the situations (as in, throw an error "if you want it, go ahead and implement it")

thofma:
> Beyond that I guess it can get arbitrarily complicated and I'd be happy to throw the towel in the situations (as in, throw an error "if you want it, go ahead and implement it")

If the input is $I \subseteq R[x_1,...,x_n]$, one could do it generically by returning the characteristic of $R/I \cap R$. Generators for the intersection $R \cap I$ can be computed via a Gröbner basis as you described it.

fingolfin:
Indeed. What I meant by "arbitrary complicated" is that then next we have to compute a quotient of `R`, and get its characteristic, and that can of course be a nasty quotient.

[...]

PR #4241, opened 2024-10-24
Add `characteristic` for multivariate quotient rings over fields

joschmitt:
Addresses #4239 for fields. The `is_zero` call needs a Gröbner basis which uses a cached basis, if one was already computed.


fingolfin (approved):
Thank you!

PR #4243, opened 2024-10-24
Support computation of basis of H^(2,2) and H^4

HereAround:
cc @lkastner 

Motivated from F-theory applications.

(You might find this entertaining @apturner @emikelsons .)

### DIFF
--- a/docs/src/AlgebraicGeometry/ToricVarieties/CohomologyClasses.md
+++ b/docs/src/AlgebraicGeometry/ToricVarieties/CohomologyClasses.md
@@ -66,4 +66,5 @@ volume_form(v::NormalToricVariety)
 intersection_form(v::NormalToricVariety)
 chern_class(v::NormalToricVariety, k::Int; check::Bool = true)
 chern_classes(v::NormalToricVariety; check::Bool = true)
+basis_of_h4(v::NormalToricVariety; check::Bool = true)
 ```
--- a/experimental/FTheoryTools/docs/src/g4.md
+++ b/experimental/FTheoryTools/docs/src/g4.md
@@ -32,4 +32,9 @@ passes_elementary_quantization_checks(gf::G4Flux)
 ```
 
 
-## Methods
+## Auxiliary
+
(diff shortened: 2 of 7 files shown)


apturner:
At a glance, it looks like this is only computing the vertical part of the middle cohomology, is that right? Of course the others will be trickier, but I’m wondering based on the name of you are planning to compute the horizontal part as well.

HereAround:
> At a glance, it looks like this is only computing the vertical part of the middle cohomology, is that right? Of course the others will be trickier, but I’m wondering based on the name of you are planning to compute the horizontal part as well.

I only want to get the vertical part.

apturner (approved):
Glad to see this is working now. Thanks @HereAround, looks good!

HereAround:
@lkastner and @apturner I will merge this PR now, as I need this so that I can proceed with the draft of a paper on the `FTheoryTools`. If you have feedback, please let me know, and I will be happy to update/improve the methods.

Please note that I placed the method `basis_of_h22` in experimental (`FTheoryTools`), as it is relevant for and inspired by the related physics discussions. Still, from a math/algorithm perspective, it achieves the exact same (for simplicial, complete toric varieties) as `basis_of_h4`. As such, it would also seem reasonable to have both methods be called `basis_of_h4`, both be placed in the toric part of OSCAR, and call them with an optional "algorithm = ..." argument. The default algorithm could be the current method `basis_of_h4`, which relies on `monomial_basis`. Let me know your thoughts.

In case you wonder @lkastner , the algorithm with `monomial_basis` is painfully slow for big toric varieties, as it triggers a Groebner basis computation. One F-theory application that we are currently working on involves a simplicial, complete toric variety, whose cohomology ring is the quotient of a multivariate polynomial ring with 310 indeterminates and the ideal divided out is generated by about 47k generators. In this case, a GB computation is hopeless, which triggered the other algorithm `basis_of_h22`. For this involved variety, `basis_of_h22` completes in about 10 seconds on my computer.

(@wdecker Based on our conversation, you might find this entertaining. In particular, you were right - this problem could be cast into a linear equation system.)

PR #4244, opened 2024-10-24
Dimension of an ideal

ederc:
Allows AlgebraicSolving to compute the dimension of an ideal, if applicable.

### DIFF
--- a/src/Rings/mpoly-ideals.jl
+++ b/src/Rings/mpoly-ideals.jl
@@ -1884,6 +1884,7 @@ julia> dim(I)
   if I.dim > -1
     return I.dim
   end
+  is_one(I) && return -1 # Catch a boundary case
   is_zero(ngens(base_ring(I))) && return 0 # Catch a boundary case
   if is_f4_applicable(I, degrevlex(base_ring(I)))
     I.dim = AlgebraicSolving.dimension(AlgebraicSolving.Ideal(I.gens.O))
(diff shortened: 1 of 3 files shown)


src/Rings/mpoly-ideals.jl
```diff
@@ -1885,7 +1885,11 @@ julia> dim(I)
     return I.dim
   end
   is_zero(ngens(base_ring(I))) && return 0 # Catch a boundary case
-  I.dim = Singular.dimension(singular_groebner_generators(I, false, true))
+  if (is_f4_applicable(I, degrevlex(base_ring(I))))
```
```suggestion
  if is_f4_applicable(I, degrevlex(base_ring(I)))
```

src/Rings/mpoly-ideals.jl
```diff
@@ -1884,8 +1884,13 @@ julia> dim(I)
   if I.dim > -1
     return I.dim
   end
+  is_one(I) && return -1 # Catch a boundary case
```
lgoettgens, line 1887: Maybe stupid question: if I remember correctly, this is_one call already computes a groebner basis (possibly via singular). Wouldn't it then be more efficient to use this knowledge in Singular.dimension? Or does AlgebraicSolving.dimension make use of that as well?

ederc: You are right, `is_one()`  computes a GB via Singular. I will work on this tomorrow, there is more that can be optimized, e.g. caching GBs.

ederc:
Let me put this in draft mode, we first need to have #4246 and some discussions on default options before it makes sense to optimize this.

afkafkafk13:
@ederc In view of the changes to the file structure introduced in the meantime in  #4250 , how do you intend to proceed with this PR?

fingolfin:
@ederc what is the status of this PR?

ederc:
This depends on decisions to be made w.r.t. GB handling, I still have this on my list.

PR #4246, opened 2024-10-25
Adds algorithm option :modular for Gröbner basis computations

ederc:
This PR is a first step for enabling multi-modular Gröbner basis computations. It adds the algorithm option `:modular` and applies (if possible) either a multi-modular f4 from AlgebraicSolving or a multi-modular Buchberger based on Singular.

Following this PR, further implementations in other applicable settings for the multi-modular Buchberger need to be moved from `experimental`. Also, discussions about default options need to be discussed and are not part of this PR.


afkafkafk13 (approved):
Looks good! Thank you!

PR #4247, opened 2024-10-25
avoid calling `conductor`

ThomasBreuer:
Instead, call `Hecke.force_coerce_cyclo(Kg, data(x), Val(false))`, which finds out whether the conversion works.

### DIFF
--- a/src/Rings/AbelianClosure.jl
+++ b/src/Rings/AbelianClosure.jl
@@ -945,9 +945,9 @@ function _embedding(F::QQField, K::QQAbField{AbsSimpleNumField},
     return QQAbFieldElem(C1(x), 1)
   end
 
-  finv = function(x::QQAbFieldElem; check::Bool = false)
+  finv = function(x::QQAbFieldElem; throw_error::Bool = true)
     res = Hecke.force_coerce_cyclo(C1, data(x), Val(false))
-    !check && res === nothing && error("element has no preimage")
+    throw_error && res === nothing && error("element has no preimage")
     return res
   end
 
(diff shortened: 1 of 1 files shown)


src/Rings/AbelianClosure.jl
```diff
@@ -933,25 +933,22 @@ end
...
   finv = function(x::QQAbFieldElem; check::Bool = false)
-    if conductor(x) == 1
-      return Hecke.force_coerce_cyclo(C1, data(x))
-    elseif check
-      return
-    else
-      error("element has no preimage")
-    end
+    res = Hecke.force_coerce_cyclo(C1, data(x), Val(false))
+    !check && res === nothing && error("element has no preimage")
```
thofma: Hm, if `check == true` and `res === nothing`, will this return nothing? Sorry, I am not good with logic.

thofma:
might be better if @fieker has a look

ThomasBreuer:
Yes, the function `_embedding` returns `nothing` if the second argument suppresses the error, like `Hecke.force_coerce_cyclo` does.
(I admit that the meaning of `check` may be confusing; taking that `_embedding` is undocumented and used only in my code, we could change this.)

thofma:
Sorry, I missed that this is an internal function. Then forget my question. But @fieker should check.

src/Rings/AbelianClosure.jl
fingolfin: This code is rather confusing to me as well. But so is the old code: it indeed also returns `nothing` when `check` is `true` and `conductor(x) != 1` ?!

fingolfin: I guess what confuses me is that `check` apparently has a different meaning here than it has in e.g. our constructors. Since I don't know what the meaning of `check` is supposed to be here, I can't comment further.

Is it documented somewhere?

ThomasBreuer: @fingolfin See my comment below: The function in question is internal and not documented (as well as the functions called by it). I think we could change the meaning of `check` to that in other situations.

ThomasBreuer:
@fingolfin @fieker I have changed the name of the keyword argument from `check` to `throw_error`. Since this is (hopefully) used only by a `has_preimage_with_preimage` method, this change should not be visible outside.
And it does not affect the proposed improvement.

ThomasBreuer:
One failure due to a segmentation fault, restarted this test.

fingolfin (approved):
Looks good to me

fingolfin:
@fieker please merge if you are OK with this versio

PR #4249, opened 2024-10-27
Saturation for modules

wdecker:


wdecker:
The error which appears here in some tests stems from an example which is NOT NEW, but which I moved from a .jl to a .md file. The function called is in MPolyQuo.jl and reads a follows:

```julia
function intersect(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}...) where T
  for g in b
    @req base_ring(g) == base_ring(a) "base rings must match"
  end
  as = Singular.intersection(singular_generators(a.gens), [singular_generators(g.gens) for g in b]...)
  return MPolyQuoIdeal(base_ring(a), as)
end
```

The result is the ideal generated by `x*y` to which the Singular calculation NOW adds the generator `x*y^2` which is not wrong but superfluous. @hannes14, @ederc: Is there a recent change in Singular(.jl) which causes this?

docs/src/CommutativeAlgebra/affine_algebras.md
```diff
@@ -294,11 +294,38 @@ minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})
 
 #### Intersection of Ideals
 
-```@docs
+```@julia
```
lgoettgens: is this change intended? why don't you want to include the docstring here?

wdecker: Please see here:
https://docs.oscar-system.org/dev/CommutativeAlgebra/affine_algebras/#Ideals-in-Affine-Algebras
When I include the docstring, a whole Rattenschwanz of methods for intersection will be printed. These happens occasionally and so far, none of our gurus had a solution for this problem. Please note: I duplicated the example using jldoctest in the .md file, so the example will run anyway.

lgoettgens: I think @thofma has some tool that might help in these cases. Maybe he could try it on this instance of the problem.

thofma: There is no docstring for the signature
```
intersect(V::Vector{MPolyQuoIdeal{T}}) where T
```
so it just includes all of them. I think simply removing this should fix everything.

docs/src/CommutativeAlgebra/affine_algebras.md
```diff
@@ -294,11 +294,38 @@ minimal_generating_set(I::MPolyQuoIdeal{<:MPolyDecRingElem})
...
 intersect(a::MPolyQuoIdeal{T}, bs::MPolyQuoIdeal{T}...) where T
 intersect(V::Vector{MPolyQuoIdeal{T}}) where T
```
thofma:
```suggestion
```

docs/src/CommutativeAlgebra/affine_algebras.md
thofma: and keep the `@docs`

thofma:
@wdecker I wrote a little package to help with the docstrings (which is how I found out about the things mentioned earlier):
```
julia> using WhereIsMyDocstring, Oscar

julia> @docmatch intersect(V::Vector{MPolyQuoIdeal{T}}) where T
6-element Vector{WhereIsMyDocstring.DocStr}:
[...]
```
The `@docmatch bla` tells you which docstrings will be included, if you write this in a ```` ```@docs```` section.

You can also do
```
julia> @docmatch intersect
25-element Vector{WhereIsMyDocstring.DocStr}:
[...]
```
to list *all* docstrings for a specific function. This also gives you instructions on how to include a specific one in the documentation (which works most of the time).

fingolfin:
Huh, it's indeed odd: the docstring for `intersect(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}...) where T`  in `src/Rings/MPolyQuo.jl:505` contains a `jldoctest` which does not seem to be executed for some reason?

It looks like this:
````
@doc raw"""
    intersect(a::MPolyQuoIdeal{T}, bs::MPolyQuoIdeal{T}...) where T
    intersect(V::Vector{MPolyQuoIdeal{T}}) where T

Return the intersection of two or more ideals.

# Examples
```jldoctest
julia> R, (x, y) = polynomial_ring(QQ, [:x, :y]);

julia> A, _ = quo(R, ideal(R, [x^2-y^3, x-y]));

julia> a = ideal(A, [y^2])
Ideal generated by
  y^2

julia> b = ideal(A, [x])
Ideal generated by
  x

julia> intersect(a,b)
Ideal generated by
  x*y

julia> intersect([a,b])
Ideal generated by
  x*y
```
"""
function intersect(a::MPolyQuoIdeal{T}, b::MPolyQuoIdeal{T}...) where T
  for g in b
    @req base_ring(g) == base_ring(a) "base rings must match"
  end
  as = Singular.intersection(singular_generators(a.gens), [singular_generators(g.gens) for g in b]...)
  return MPolyQuoIdeal(base_ring(a), as)
end

````


And clearly CI tests pass. But if I run this test manually in the REPL, I get different output (just like @wdecker):
```
julia> R, (x, y) = polynomial_ring(QQ, [:x, :y]);

julia> A, _ = quo(R, ideal(R, [x^2-y^3, x-y]));

julia> a = ideal(A, [y^2])
Ideal generated by
  y^2

julia> b = ideal(A, [x])
Ideal generated by
  x

julia> intersect(a,b)
Ideal generated by
  x*y
  x*y^2

julia> intersect([a,b])
Ideal generated by
  x*y
  x*y^2
```

If I run
```
Oscar.doctest_fix("src/Rings/MPolyQuo.jl")
```
it modifies the docstring to the new output.

But CI does not complain. What's going on there?

thofma:
The following claims that this is run

https://github.com/oscar-system/Oscar.jl/actions/runs/11552275709/job/32151021946?pr=4249#step:8:2272

hannes14:
The problem with additional element in intersect is fixed in Singular with 
https://github.com/Singular/Singular/commit/de8b8820c8ed40e71333ebf2c90eb584954a12b9,
will be in Oscar with the next Singular(.jl) release

wdecker:
thx!

fieker (approved):

PR #4250, opened 2024-10-28
Splits groebner.jl into parts

ederc:
This PR splits `groebner.jl` into meaningful parts. It adds a global `groebner/groebner.jl` file that just includes all other parts like it is  done in `Groups`.


joschmitt (approved):
Thank you!

joschmitt:
This has some potential for merge conflicts, but right now the only pull request that edits this file is #4244 from what I can see.

fingolfin (approved):

PR #4256, opened 2024-10-30
Print Betti tables more nicely

fingolfin:
This makes it look a bit nicer (which is subjective) but also ensure that the Betti tables in jldoctests are formatted such that `doctest=:fix` works correctly (it didn't before because the first printed line had too many leading spaces, confusing Documenter's code for detecting were an input line ends).


src/Modules/ModulesGraded.jl
```diff
@@ -1426,29 +1426,32 @@ of `F` in form of a Betti table.
...
-2    : -  2  3  1
-------------------
-total: 1  4  4  1
+     |  0  1  2  3
+-----+------------
+0    |  1  -  -  -
+1    |  -  2  1  -
+2    |  -  2  3  1
+-----+------------
+total:  1  4  4  1
```
lgoettgens: this looks a bit off-centered. I suggest to remove one horizontal space between the `|` and the following entry, i.e. like this
```suggestion
     | 0  1  2  3
-----+-----------
0    | 1  -  -  -
1    | -  2  1  -
2    | -  2  3  1
-----+-----------
total: 1  4  4  1
```

jankoboehm:
The : is what is found in most books and papers (mostly due to M2), but if the causes trouble I can live with |, but we should discuss that.

src/Modules/ModulesGraded.jl
fingolfin: I changed it because to me it seemed weird that all entries had two space to their left, except those in the left most column *shrug*.

lgoettgens: Interesting to hear this pov. In my mind it was *obvious* that each column has the form space-elem-space
Dunno, I don't have any preference here

fingolfin:
I don't care about `:` vs. `|` much (although I find `:` weird for a table, but *shrug*), it is trivial to change.

My main interest is to get rid of the leading spaces in the first row of the table. We discussed this before, and as a temporary workaround the offending tests were changed to `jldoctest` to `julia`. 

So far nobody has shown any signs of implementing a better fix that allows us to re-enable those tests, and so I am doing that now. As you can see in the diff, there were already test outputs that had started to diverge from what actually happens in OSCAR, and we didn't notice.

fingolfin:
I've updated this now to make a *minimal* change, i.e. it just inserts `:` in the first row. 

As a side effect this also makes it even more obvious how much those examples have diverged since they were added

wdecker:
@fingolfin Thx for doing this. I would like to discuss details in person. I will be at the university tomorrow.

wdecker:
Here is a more extensive example for testing:

S, x = graded_polynomial_ring(QQ, :x => 1:14, [-1,-1,1,1,1,1,1,1,1,1,1,1,2,2]);
I = ideal(S, gens(S));
FI = free_resolution(I)
betti(FI)

fingolfin:
So that renders like this (if your terminal is wide enough):
```
julia> S, x = graded_polynomial_ring(QQ, :x => 1:14, [-1,-1,1,1,1,1,1,1,1,1,1,1,2,2]);

julia> I = ideal(S, gens(S));

julia> FI = free_resolution(I)
Free resolution of I
S^14 <---- S^91 <---- S^364 <---- S^1001 <---- S^2002 <---- S^3003 <---- S^3432 <---- S^3003 <---- S^2002 <---- S^1001 <---- S^364 <---- S^91 <---- S^14 <---- S^1 <---- 0
0          1          2           3            4            5            6            7            8            9            10          11         12         13        14

julia> betti(FI)
     :  0   1    2     3     4     5     6     7     8     9   10  11  12  13
-----------------------------------------------------------------------------
-3   :  -   1   10    45   120   210   252   210   120    45   10   1   -   -
-2   :  -   -    2    20    90   240   420   504   420   240   90  20   2   -
-1   :  2  20   90   241   430   549   540   450   342   230  122  45  10   1
0    :  -   4   40   180   480   840  1008   840   480   180   40   4   -   -
1    : 10  45  122   230   342   450   540   549   430   241   90  20   2   -
2    :  2  20   90   240   420   504   420   240    90    20    2   -   -   -
3    :  -   1   10    45   120   210   252   210   120    45   10   1   -   -
-----------------------------------------------------------------------------
total: 14  91  364  1001  2002  3003  3432  3003  2002  1001  364  91  14  1
```

fingolfin:
@jankoboehm and @wdecker and myself discussed this and we agreed on this layout:
```
julia> betti(FI)
degree:  0   1    2     3     4     5     6     7     8     9   10  11  12  13
------------------------------------------------------------------------------
    -3:  -   1   10    45   120   210   252   210   120    45   10   1   -   -
    -2:  -   -    2    20    90   240   420   504   420   240   90  20   2   -
    -1:  2  20   90   241   430   549   540   450   342   230  122  45  10   1
     0:  -   4   40   180   480   840  1008   840   480   180   40   4   -   -
     1: 10  45  122   230   342   450   540   549   430   241   90  20   2   -
     2:  2  20   90   240   420   504   420   240    90    20    2   -   -   -
     3:  -   1   10    45   120   210   252   210   120    45   10   1   -   -
------------------------------------------------------------------------------
 total: 14  91  364  1001  2002  3003  3432  3003  2002  1001  364  91  14   1
```

jankoboehm (approved):

wdecker (approved):
Fine with me as soon as tests are green. @fingolfin Thanks again for dong this.

fingolfin:
I've implemented the new styling and also changed on test to make it easier to maintain. Wolfram's example is now printed exactly as we discussed.

jankoboehm:
I guess auto-merge has a problem with the book tests since the printing changed.

benlorenz:
> I guess auto-merge has a problem with the book tests since the printing changed.

I don't see any errors from betti tables in the log, both book test jobs (1.10 and nightly) were stuck in `vinberg_2.jlcon` for about 2 hours and then killed. I restarted them now.

_Edit:_ Looking at the order of the chapters, I think the booktests will indeed fail later due to the changed printing.

fingolfin:
Ah I forgot about the book tests. Do we have something like `Oscar.doctest_fix` for the book test files to automatically update them?

Issue #277, opened 2021-02-12
Add an AUTHORS.md or CONTRIBUTORS.md or so...


fingolfin:
... which lists relevant names. The `authors` list in `Project.toml` is useless when it extends beyond more than 2-3 entries, and diffs are annoying to read. Thus, I'd also replace it by a single entry, e.g. the oscar-dev mailing list.

lkastner:
Is this done with adding corresponding authors in the docs?

fingolfin:
Seems we have this in `LICENSE.md` now:
```
The Oscar.jl package is licensed under the GNU Public License, Version 3.0+:

> Copyright (c) 2019-2025 The OSCAR Development Team
>
> See <https://www.oscar-system.org/contributors/> for a list of OSCAR
> Development Team members. Note that this list may occasionally be slightly
> outdated. Additional contributors may be found in the `git log` of the main
> OSCAR repository and other related repositories.
...
```

Which almost resolves the issue! Except <https://www.oscar-system.org/contributors> (so without the trailing slash) leads to a 404 error. So: we should make a redirect to the correct location and make sure it keeps working.

fingolfin:
@aaruni96 and @HereAround will take care of that redirect and then close this issue.

aaruni96:
Fixed by https://github.com/oscar-system/oscar-website/pull/519

PR #4257, opened 2024-10-30
Update license

paemurru:


LICENSE.md
```diff
@@ -1,6 +1,6 @@
 The Oscar.jl package is licensed under the GNU Public License, Version 3.0+:
 
-> Copyright (c) 2019 The OSCAR Development Team.
+> Copyright (c) 2019-2024 [The OSCAR Development Team](https://github.com/oscar-system/Oscar.jl/graphs/contributors)
```
fingolfin: https://www.oscar-system.org/contributors/

fingolfin: The year should be updated.

For legal purposes it would be important to "define" what "The OSCAR Development Team" means. A link to a list of contributors would be a good start but it is a bit vague.

In GAP we "solved" this by including an explicit explanation of what the "team" is, and in this case, a full list of people. See https://github.com/gap-system/gap/blob/master/COPYRIGHT . Note that we wrote "which include" on purpose to make it open to "we may have missed a few people". I think the GAP text could be even better if it explicitly mentioned that one should also check the `git log` or something.

So in our case that could be done by having an explicit text, like this:
```suggestion
> Copyright (c) 2019-2024 The OSCAR Development Team
>
> See <https://www.oscar-system.org/contributors/> for a list of OSCAR
> Development Team members. Note that this list may occasionally be slightly
> outdated. Additional contributors may be found in the `git log` of the main
> OSCAR repository and other related repositories.
```

fingolfin: And then we could perhaps use similar verbiage in AA and Nemo, though those don't have separate contributor lists, so perhaps there the analogous GitHub "contributors" list could be used. You are welcome to open PRs on those repositories with a proposal and we can fine tune them there :-)

LICENSE.md
```diff
@@ -1,6 +1,11 @@
 The Oscar.jl package is licensed under the GNU Public License, Version 3.0+:
 
-> Copyright (c) 2019 The OSCAR Development Team.
+> Copyright (c) 2019-2024 The OSCAR Development Team
+>
+> See <https://www.oscar-system.org/contributors/> for a list of OSCAR
+> Development Team members. Note that this list may occasionally be slightly
+> outdated. Additional contributors may be found in the `git log` of the main
+> OSCAR repository and other related repositories.
```
fingolfin, line 8: Also relevant: https://github.com/oscar-system/Oscar.jl/issues/277 -- perhaps that can be closed with this PR (and with the work by @HereAround and @aaruni96 on https://github.com/oscar-system/oscar-website/pull/373)

fingolfin (approved):
Sounds good to me, thank you.

fingolfin:
I'll wait a bit and see if @thofma or @fieker or someone else has further comments and will otherwise merge this on Friday or so

thofma (approved):

thofma:
if @fingolfin is happy, I am happy

